PR #131, opened 2015-05-15
Support for having file attributes in the State object

suvodeep-pyne:
Gobblin now preserves the owner, group, permissions and replication factor for files. The state object now supports the following configuration keys which may be used to capture these attributes and pass them on to the output file.

- HDFS_FILE_OWNER
- HDFS_FILE_GROUP
- HDFS_FILE_REPLICATION_FACTOR
- HDFS_FILE_PERMISSIONS

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -203,6 +203,10 @@ public class ConfigurationKeys {
   public static final String WRITER_FILE_NAME = WRITER_PREFIX + ".file.name";
   public static final String WRITER_FILE_PATH = WRITER_PREFIX + ".file.path";
   public static final String WRITER_FILE_PATH_TYPE = WRITER_PREFIX + ".file.path.type";
+  public static final String WRITER_FILE_OWNER = WRITER_PREFIX + ".file.owner";
+  public static final String WRITER_FILE_GROUP = WRITER_PREFIX + ".file.group";
+  public static final String WRITER_FILE_REPLICATION_FACTOR = WRITER_PREFIX + ".file.replication.factor";
+  public static final String WRITER_FILE_PERMISSIONS = WRITER_PREFIX + ".file.permissions";
   public static final String WRITER_BUFFER_SIZE = WRITER_PREFIX + ".buffer.size";
   public static final String WRITER_PRESERVE_FILE_NAME = WRITER_PREFIX + ".preserve.file.name";
   public static final String WRITER_DEFLATE_LEVEL = WRITER_PREFIX + ".deflate.level";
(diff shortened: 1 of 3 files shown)


gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
```diff
@@ -167,6 +168,24 @@ public void commit()
     if (!this.fs.rename(this.stagingFile, this.outputFile)) {
       throw new IOException("Failed to commit data from " + this.stagingFile + " to " + this.outputFile);
     }
+
```
liyinan926: I suggest abstracting the following code into a utility method into `gobblin.util.HadoopUtils`, say, `public static void keepSourceFileAttributes(FileSystem fs, Path file, State properties)`.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -409,6 +409,14 @@
   public static final String DEFAULT_JOB_HISTORY_STORE_USER = "gobblin";
   public static final String JOB_HISTORY_STORE_PASSWORD_KEY = "job.history.store.password";
   public static final String DEFAULT_JOB_HISTORY_STORE_PASSWORD = "gobblin";
+  /**
+   * HDFS file properties
```
liyinan926: Missing dot at the end of the javadoc.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -409,6 +409,14 @@
   public static final String DEFAULT_JOB_HISTORY_STORE_USER = "gobblin";
   public static final String JOB_HISTORY_STORE_PASSWORD_KEY = "job.history.store.password";
   public static final String DEFAULT_JOB_HISTORY_STORE_PASSWORD = "gobblin";
+  /**
```
liyinan926: Please add an empty line before the docs.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -409,6 +409,14 @@
...
+   * HDFS file properties
+   */
```
sahilTakiar: Can we move these configs to the "Writer configuration properties" section. They can be of the from "public static final String WRITER_FILE_OWNER = WRITER_PREFIX + ".file.owner"

This way we are consistent with the rest of the configuration keys.

sahilTakiar:
I believe we also discussed replacing `AvroHdfsDataWriter.CodecType` with `CodecFactory.fromString(String s)` since it will provide a safer option for setting the codec. Do you want to do that in this pull request, or a different one?

suvodeep-pyne:
I was planning to do the refactor as a different pull request.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
suvodeep-pyne: extracted a method. But I'm keeping it in the same class since it is very context specific. Do we have another use case in which this method may be useful?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
suvodeep-pyne: Fixed it.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
suvodeep-pyne: Updated the code.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
liyinan926: I still think it should be a utility method. A better place to put it is `WriterUtils`.

suvodeep-pyne: Noted. Moved the method to WriterUtils.

liyinan926:
LGTM other than the three comments above.

liyinan926:
LGTM.

sahilTakiar:
LGTM. In the next pull request, can we also add a unit test for AvroHdfsDataWriter to ensure that it is preserving all the properties properly.

suvodeep-pyne:
sure. I can do that. 
For the current changes though, the test will have to check the results of a bunch of set commands.

PR #132, opened 2015-05-15
Added support for optionally limiting data extraction in the Extractor

liyinan926:
This is the first version of work on adding support for optionally throttling data extraction in the Extractor. The changes are summarized as follows:
1. Added a new interface `Limiter` and three implementations: `RateBasedLimiter`, `TimeBasedLimiter`, `CountBasedLimiter`, and `PoolBasedLimiter` that limits the rate, total time elapsed, total permits available (with no refills), and total permits available in a pool with refills, respectively.
2. Added a `LimiterFactory` for building `Limiter`s.
3. Added some new configuration properties for the new `Limiter`s.
4. Added a new class `LimitingExtractorDecorator` that applies throttling using a given `Limiter` to the `Extractor` it wraps.
5. `TaskContext.getExtractor` now returns a `LimitingExtractorDecorator` if throttling is enabled.
6. Deleted the class `ExtractorDecorator` that is not doing any useful work in addition to what an `Extractor` is already doing.

[ci skip]


sahilTakiar:
Came across this, thought it was interesting / relevant: http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/RateLimiter.html

liyinan926:
Yeah, I didn't notice that. It seems more general and handles bursty requests. I changed `RateBasedLimiter` to use that.

gobblin-runtime/src/main/java/gobblin/runtime/LimiterFactory.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class LimiterFactory {
+
+  /**
+   * Create a new {@link Limiter} instance.
+   *
+   * @param state a {@link State} instance carrying configuration properties
+   * @return a new {@link Limiter} instance
+   */
+  public static Limiter newLimiter(State state) {
```
ibuenros: It seems this assumes a guarantee that a limiter is desired? I see that in TaskContext you are checking whether throttling is enabled before calling the factory, but in the future someone might not be checking that and assume LimiterFactory handles no-limiter graciously. Maybe check if limiter is enabled and return with no exception if it isn't? (either return null, or return a Default limiter with no limit, or an optional).

liyinan926: I think the natural contract is if no `Limiter` is needed, then this class should not be used.

gobblin-runtime/src/main/java/gobblin/runtime/LimitingExtractorDecorator.java
```diff
@@ -0,0 +1,86 @@
...
+ *   or each individual {@link LimitingExtractorDecorator} has its own {@link Limiter}.
+ *   The first case is useful for throttling at above the task level, e.g., at the job level.
+ * </p>
+ *
+ * @param <S> output schema type
+ * @param <D> output record type
+ *
+ * @author ynli
+ */
+public class LimitingExtractorDecorator<S, D> implements Extractor<S, D> {
```
ibuenros: The LimiterFactory allows for creation of poolBasedLimiters, however, this decorator will never release the resources, effectively making the poolBasedLimiter a count limiter. Maybe the factory should not allow pool based limiters for now?

gobblin-runtime/src/main/java/gobblin/runtime/LimiterFactory.java
ibuenros: That makes sense. Then, can we specify that contract in the javadoc?

gobblin-runtime/src/main/java/gobblin/runtime/LimitingExtractorDecorator.java
liyinan926: I think there's a clear distinction or boundary between the `LimiterFactory` and the entity who uses it. In this case, it does not make sense for `LimitingExtractorDecorator` to use a `PoolBasedLimiter` and we can definitely enforce that in `LimitingExtractorDecorator` by throwing an exception if the job configuration intends to use a `PoolBasedLimiter`. But I don't think the `LimiterFactory` should be enforcing that.

ibuenros: That is also an adequate solution for me.

liyinan926: I think it also makes sense to divide the `Limiter`s into two groups, `NonRefillableLimiter` and `RefillableLimiter`, and `LimitingExtractorDecorator` should only accept `NonRefillableLimiter`.

liyinan926:
Updated 1b7b809.

liyinan926:
@ibuenros can you review the updates?

ibuenros:
Just one comment.

liyinan926:
Updated 3858d96.

ibuenros:
LGTM

liyinan926:
@chavdar do you want to take a look?

gobblin-runtime/src/main/java/gobblin/runtime/Limiter.java
```diff
@@ -0,0 +1,119 @@
...
+   * <p>
+   *   Depending on the implementation, the caller of this method may be blocked.
+   *   It is also up to the caller to decide how to deal with the return value.
+   * </p>
+   *
+   * @param permits number of permits to get
+   * @return {@code true} if the permits have been acquired successfully or {@code false} otherwise
+   * @throws InterruptedException if the caller is interrupted while being blocked
+   */
+  public boolean acquirePermits(int permits) throws InterruptedException;
```
chavdar: Let's make this a long.

gobblin-runtime/src/main/java/gobblin/runtime/Limiter.java
```diff
@@ -0,0 +1,119 @@
...
+ * e.g., data record extraction using an {@link gobblin.source.extractor.Extractor}.
+ *
+ * @author ynli
+ */
+public interface Limiter {
+
+  /**
+   * Supported types of {@link Limiter}s.
+   */
+  public enum Type {
```
chavdar: I don't think the Limiter interface should be limiting the possible implementations :)

liyinan926: The purpose is not to limiting the implementations. I could not think of a better place to put this enum. I can move it out and maybe rename it to `LimiterType`.

gobblin-runtime/src/main/java/gobblin/runtime/LimiterFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+
+
+/**
+ * A factory class for {@link Limiter}s.
+ *
+ * @author ynli
+ */
+public class LimiterFactory {
```
chavdar: nit: can you rename to StandardLimiterFactory or something along those lines. LimiterFactory sounds a bit too generic. Someone else may want to implement their own limiter and limiter factory.

liyinan926: Would `BaseLimiterFactory` be better?

gobblin-runtime/src/main/java/gobblin/runtime/Limiter.java
liyinan926: That would have a huge impact on the implementations. Let's talk this in person.

chavdar: I am thinking that maybe the acquirePermits() method should return a Closeable which encapsulates the permits one has acquired. The semantics is that the close() method will release the permits. null is the same as false, acquisition failed.

I am afraid that the current API allows hard-to-track bugs:
* Acquire permits and forgetting to release
* Release less or more permits than acquired.

NonRefillableLimiter can return a singleton implementation where the close is a no-op.

What do you think?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -153,7 +153,17 @@
   public static final String EXTRACT_PRIMARY_KEY_FIELDS_KEY = "extract.primary.key.fields";
   public static final String EXTRACT_DELTA_FIELDS_KEY = "extract.delta.fields";
   public static final String EXTRACT_SCHEMA = "extract.schema";
-  public static final String EXTRACT_PULL_LIMIT = "extract.pull.limit";
+  public static final String EXTRACT_LIMIT_ENABLED_KEY = "extract.limit.enabled";
```
chavdar: How about putting the config names in the LimiterFactory class? Then it is very clear where they are used.

gobblin-runtime/src/main/java/gobblin/runtime/Limiter.java
liyinan926: Yeah, I am fine with this approach. This is definitely less error-prone.

liyinan926:
Updated 420ecef.

PR #133, opened 2015-05-18
Instrument runtime

ibuenros:
This PR:
* Adds topic and partition tags to Kafka extractor.
* Adds runtime timings to gobblin execution.
* Adds recursive metric reporting.
* Adds decorator interface to correctly resolve tags and instrumentation of decorated objects.


gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Finds the decorator lineage of the given decorator with the given directly underlying object.
+   *
+   * @param decorator decorator object.
+   * @param underlying directly underlying object to decorator.
+   * @return List of the non-decorator underlying object and all decorators on top of it.
+   */
+  public static List<Object> resolveDecoratorLineage(Decorator decorator, Object underlying) {
+    List<Object> lineage =  Decorator.class.isInstance(underlying) ?
```
zliu41: Is there a reason to use `isInstance` instead of `instanceof`?

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Finds the decorator lineage of the given decorator with the given directly underlying object.
+   *
+   * @param decorator decorator object.
+   * @param underlying directly underlying object to decorator.
+   * @return List of the non-decorator underlying object and all decorators on top of it.
+   */
+  public static List<Object> resolveDecoratorLineage(Decorator decorator, Object underlying) {
```
zliu41: I think this method should check whether `decorator` is indeed a decorator of `underlying`. Otherwise I can pass a `SourceDecorator` and a `Converter`.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -182,6 +232,11 @@ public Instrumented(State state, Class<?> klazz, List<Tag<?>> tags) {
   }
 
   @Override
+  public List<Tag<?>> generateTags(State state) {
+    return Lists.newArrayList();
```
zliu41: Add some comments to explain why.

gobblin-core/src/main/java/gobblin/instrumented/converter/InstrumentedConverterBase.java
```diff
@@ -69,6 +77,11 @@
   }
 
   @Override
+  public List<Tag<?>> generateTags(State state) {
+    return Lists.newArrayList();
```
zliu41: Ditto

gobblin-api/src/main/java/gobblin/util/Decorator.java
```diff
@@ -0,0 +1,61 @@
...
+
+package gobblin.util;
+
+import java.util.List;
+
+
+/**
+ * Interface for decorator patterns.
+ */
+public interface Decorator {
```
zliu41: My personal preference is that this interface just has a single method `getImmediateUnderlying()` or `getEnclosed()`. The two methods currently associated with it can be taken care of using static methods in DecoratorUtils. You can get another person's opinion.

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
ibuenros: I though about that, but it turns out things get very complicated due to type erasure. I decided to leave it up to the decorators themselves to handle the check if necessary. This decision was mostly because I found no simple way of doing it better.

I am open to suggestions on how to implement a more typesafe way of doing this logic.

gobblin-api/src/main/java/gobblin/util/Decorator.java
ibuenros: That makes sense! I'll just wait for second reviewer to make sure that there is no reason we could want the current logic.

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
ibuenros: I didn't know instanceof worked outside of if statements. Neat!

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -258,6 +259,15 @@ public void startMetricReporting(Properties properties) {
   }
 
   /**
+   * Immediately trigger metric reporting.
+   */
+  public void triggerMetricReporting() {
+    if (this.fileReporter.isPresent()) {
```
zliu41: Should something be done if fileReporter is absent?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
```diff
@@ -71,7 +75,15 @@ public KafkaExtractor(WorkUnitState state) {
   }
 
   @Override
-  public D readRecord(D reuse) throws DataRecordException, IOException {
+  public List<Tag<?>> generateTags(State state) {
+    List<Tag<?>> tags = super.generateTags(state);
+    tags.add(new Tag<String>("topic", state.getProp(KafkaSource.TOPIC_NAME)));
+    tags.add(new Tag<Integer>("partition", state.getPropAsInt(KafkaSource.PARTITION_ID)));
```
zliu41: `KafkaExtractor` has been modified in #135 to pull from multiple partitions of the same topic. I suggest waiting until that PR is merged to make the corresponding changes here.

liyinan926:
@ibuenros seems there's already some comments from @zliu41 and you may need to wait until #135 is merged. So I will look at this pull request after you address all comments from @zliu41.

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -140,6 +141,14 @@ public void addChildContext(String childContextName, MetricContext childContext)
   }
 
   /**
+   * Get child contexts.
+   * @return Map of child contexts.
+   */
+  public Map<String, MetricContext> getChildContexts() {
+    return Collections.unmodifiableMap(this.children);
```
zliu41: Use the superior `ImmutableMap`.

gobblin-metrics/src/main/java/gobblin/metrics/RecursiveScheduledReporter.java
```diff
@@ -0,0 +1,77 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricFilter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.ScheduledReporter;
+import com.codahale.metrics.Timer;
+import com.google.common.base.Function;
+import com.google.common.collect.Maps;
+
+
+public abstract class RecursiveScheduledReporter extends ScheduledReporter {
```
zliu41: Some javadoc is needed for this class.

gobblin-metrics/src/main/java/gobblin/metrics/Tagged.java
```diff
@@ -63,6 +65,14 @@ public void addTags(Collection<Tag<?>> tags) {
   }
 
   /**
+   * Return tags as a Map.
+   * @return map of tags.
+   */
+  public Map<String, Object> getTagMap() {
+    return Collections.unmodifiableMap(this.tags);
```
zliu41: Ditto

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
liyinan926: Yes, see https://code.google.com/p/guava-libraries/wiki/ImmutableCollectionsExplained#Why?.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
ibuenros: No. This will also change in the next PR.

gobblin-api/src/main/java/gobblin/util/Decorator.java
ibuenros: Fixed.

ibuenros:
@liyinan926 @zliu41 PR has been updated and merged with the new KafkaExtractor. Can you take a look?

gobblin-api/src/main/java/gobblin/util/Decorator.java
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Get directly underlying object.
+   * @return directly underlying object.
+   */
+  public Object getDirectlyUnderlying();
```
liyinan926: I suggest changing the name to `getDecoratedObject`.

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  public static Object resolveUnderlyingObject(Object obj) {
+    Object underlying = obj;
+    while(underlying instanceof Decorator) {
+      underlying = ((Decorator)underlying).getDirectlyUnderlying();
+    }
+    return underlying;
+  }
+
+  /**
+   * Finds the decorator lineage of the given decorator with the given directly underlying object.
```
liyinan926: This method may be called on object that is `instanceof Decorator`, so the Javadoc here may not apply.

gobblin-core/src/main/java/gobblin/instrumented/Instrumentable.java
```diff
@@ -33,4 +38,47 @@
...
   public boolean isInstrumentationEnabled();
+
+  /**
+   * Generate tags that should be added to the {@link gobblin.metrics.MetricContext}.
+   * @return List of tags to add to Metric Context.
+   */
+  public List<Tag<?>> generateTags(State state);
+
+  /**
+   * Generate a new metric context replacing old tags with input tags (only tags with the same keys will be replaced),
```
liyinan926: Can you use `{@link MetricContext}` and this kind of references for other classes whenever possible?

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +138,85 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
...
+   * @param tags New tags to apply to context. Repeated keys will override old tags.
+   * @param name Name of the new context. If absent or empty, will modify old name by adding a random integer at the end.
+   * @return Copied context.
+   */
+  public static MetricContext copyMetricContext(MetricContext context, List<Tag<?>> tags, Optional<String> name) {
+
+    String newName = name.orNull();
+
+    if (Strings.isNullOrEmpty(newName)) {
+      int randomId = RAND.nextInt();
```
liyinan926: You should do `RAND.nextInt() & Integer.MAX_VALUE` to make sure the random integers are non-negative.

gobblin-core/src/main/java/gobblin/instrumented/Instrumentable.java
```diff
@@ -33,4 +38,47 @@
...
+   * Switches the existing metric context with the supplied metric context and regenerates metrics.
+   *
+   * <p>
+   *   This method is useful when the state of the extractor changes and the user wants that state change to
+   *   be reflected in the tags of the extractor instrumentation.
+   * </p>
+   *
+   * <p>
+   *   This method is an alternative to {@link #switchMetricContext(List)} when metric context switching is done
+   *   often between a small set of contexts. The subclass should cache contexts, and call this method instead, saving
```
liyinan926: OK, what do you mean by `cache contexts`?

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -117,7 +124,7 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
     if (construct != null) {
       generatedTags.add(new Tag<String>("construct", construct.toString()));
     }
-    generatedTags.add(new Tag<String>("class", klazz.getCanonicalName()));
+    generatedTags.add(new Tag<String>("class", klazz.isAnonymousClass() ? "anonymous" : klazz.getCanonicalName()));
```
liyinan926: If a class is anonymous, why bother add the class name as a tag?

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +138,85 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
...
+  public static MetricContext copyMetricContext(MetricContext context, List<Tag<?>> tags, Optional<String> name) {
+
+    String newName = name.orNull();
+
+    if (Strings.isNullOrEmpty(newName)) {
+      int randomId = RAND.nextInt();
+
+      String oldName = context.getName();
+      List<String> splitName = Splitter.on(".").splitToList(oldName);
+      if (StringUtils.isNumeric(Iterables.getLast(splitName))) {
```
liyinan926: So you are pretty much assuming that if the last component is a number, then it is semantically the same as a random integer. I don't think this is a safe thing to do.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +138,85 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
...
+        splitName.set(splitName.size() - 1, Integer.toString(randomId));
+      } else {
+        splitName.add(Integer.toString(randomId));
+      }
+      newName = StringUtils.join(splitName, ".");
```
liyinan926: Please use `Joiner`.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +138,85 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
   }
 
   /**
+   * Copies a metric context and returns a new context with different name, and possibly different tags.
```
liyinan926: Please mention that the copy does not include any metrics associated with the old context.

gobblin-core/src/main/java/gobblin/instrumented/Instrumentable.java
```diff
@@ -33,4 +38,47 @@
...
+   *
+   * <p>
+   *   Notice that this method creates a brand new metric context and metrics every time it is called, with the
+   *   associated processing and memory overhead. Use sparingly only for state changes that MUST be visible in
+   *   emitted metrics.
+   * </p>
+   *
+   * @param tags additional tags.
+   */
+  public void switchMetricContext(List<Tag<?>> tags);
```
liyinan926: It seems to be both this and the method below are for very special use cases. I don't think they should be part of the `Instrumentable` interface. @chavdar what do you think?

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -200,6 +221,8 @@ public void launchJob(JobListener jobListener)
                 this.jobContext.getJobName()));
       }
 
+      Optional<Timer.Context> createWorkyunitsTimer = Instrumented.timerContext(metricContext,
```
liyinan926: `metricContext` is an `Optional` so it may be `absent`.

ibuenros: That is taken care of by Instrumented.timerContext

gobblin-runtime/src/main/java/gobblin/runtime/util/MetricNames.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Metric names for Gobblin runtime.
+ */
+public class MetricNames {
+  public static class LauncherTimings {
+    public static String BASE_NAME = "gobblin.launcher.timings";
+    public final static String CREATE_WORK_UNITS = MetricRegistry.name(BASE_NAME, "create.work.units");
+    public final static String ADD_WORK_UNITS = MetricRegistry.name(BASE_NAME, "add.work.units");
+    public final static String WRITE_JOB_HISTORY = MetricRegistry.name(BASE_NAME, "write.job.history");
+    public final static String RUN_JOB = MetricRegistry.name(BASE_NAME, "job.run");
```
liyinan926: Can you rename this to `JOB_RUN` to be consistent with the name? Ditto for other names.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +138,85 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
...
+
+  /**
+   * Determines whether an object or, if it is a decorator, any object on its lineage, is already instrumented.
+   * @param obj Object to analyze.
+   * @return Whether the object is instrumented.
+   */
+  public static boolean isLineageInstrumented(Object obj) {
+    List<Object> lineage = Lists.newArrayList(obj);
+    if(obj instanceof Decorator) {
+      lineage = DecoratorUtils.getDecoratorLineage(obj);
```
liyinan926: Seems `DecoratorUtils.getDecoratorLineage` already handles both cases. So you don't need to do the `if`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -81,6 +86,9 @@
   // A single-thread executor for executing job cancellation
   protected final ExecutorService cancellationExecutor;
 
+  // An MetricContext to track runtime metrics only if metrics are enabled.
+  protected final Optional<MetricContext> runtimeMetrics;
```
liyinan926: Please rename this to `runtimeMetricContext`

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -192,6 +207,12 @@ public void launchJob(JobListener jobListener)
       throws JobException {
     String jobId = this.jobContext.getJobId();
     JobState jobState = this.jobContext.getJobState();
+    Optional<MetricContext> metricContext = this.jobContext.getJobMetricsOptional().transform(new Function<JobMetrics, MetricContext>() {
```
liyinan926: Seems you should use `this.runtimeMetricContext`.

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -0,0 +1,56 @@
...
+public class DecoratorUtils {
+
+  /**
+   * Resolves the truly underlying object in a possible chain of {@link gobblin.util.Decorator}.
+   *
+   * @param obj object to resolve.
+   * @return the true non-decorator underlying object.
+   */
+  public static Object resolveUnderlyingObject(Object obj) {
+    Object underlying = obj;
```
zliu41: There's no need to define `underlying`.

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+    return underlying;
+  }
+
+  /**
+   * Finds the decorator lineage of the given decorator with the given directly underlying object.
+   *
+   * @param obj decorator object.
+   * @return List of the non-decorator underlying object and all decorators on top of it,
+   *  starting with underlying object and ending with decorator.
```
zliu41: It should be indicated that whether the interval is open or closed.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
zliu41: Do you need to guarantee generating a unique Id every time? If so you may want to use UUID.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
```diff
@@ -152,6 +165,12 @@ private boolean currentPartitionFinished() {
   private void moveToNextPartition() {
     this.currentPartitionIdx++;
     this.messageIterator = null;
+    switchMetricContextToCurrentPartition();
+  }
+
+  private void switchMetricContextToCurrentPartition() {
+    int currentPartitionId = this.partitions.get(this.currentPartitionIdx).getId();
```
zliu41: you can use `getCurrentPartition().getId()`

ibuenros:
Addressed comments.

gobblin-core/src/main/java/gobblin/instrumented/Instrumentable.java
ibuenros: I see your point, but I think this is a use case that could apply to any Instrumentable (anything that changes state might want to change its tags), which is why I put it here. Having said that, I am not opposed to removing it.

liyinan926: OK, I am fine leaving those here given `Instrumented` has default implementations for them.

ibuenros:
Addressed more comments.

zliu41:
LGTM. There are too many commits in this PR, please squash some of them before merging.

gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,6 +142,92 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
...
+    String newName = name.orNull();
+
+    if (Strings.isNullOrEmpty(newName)) {
+      UUID uuid = UUID.randomUUID();
+      String randomIdPrefix = "uuid:";
+
+      String oldName = context.getName();
+      List<String> splitName = Strings.isNullOrEmpty(oldName) ?
+          Lists.<String>newArrayList() :
+          Splitter.on(".").splitToList(oldName);
```
zliu41: `splitToList` returns an immutable list. Can you change this? 
You can push to upstream without creating a new PR.

ibuenros: Thanks! I fixed this in a new PR.

PR #134, opened 2015-05-18
Handled the Blob type of data in Mysql db

tuGithub:


gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -943,6 +970,19 @@ public long getCount(CommandOutput<?, ?> response) throws RecordCountException {
       throw new DataRecordException("Failed to get records from MySql; error - " + e.getMessage(), e);
     }
   }
+  
+  private void initialBlobTypeHashSet(Set<String> input, ResultSetMetaData resultsetMetadata) throws SQLException{
```
chavdar: nit: it looks more natural for the function to just return the set rather than have it passed as a parameter.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -914,15 +919,37 @@ public long getCount(CommandOutput<?, ?> response) throws RecordCountException {
...
+      Base64.Encoder encoder = Base64.getEncoder();
+      
       int batchSize = this.workUnit.getPropAsInt(ConfigurationKeys.SOURCE_QUERYBASED_FETCH_SIZE, 0);
       batchSize = (batchSize == 0 ? ConfigurationKeys.DEFAULT_SOURCE_FETCH_SIZE : batchSize);
 
       int recordCount = 0;
+      Set<String> blobDataNames = new HashSet<String>();
+      boolean initialed_blobDataNames = false;
       while (resultset.next()) {
+        if (!initialed_blobDataNames){
```
chavdar: Why do we have to initialize the set of blob columns inside the loop? Isn't it possible to do it outside the loop. In that case, the initialed_blobDataNames should not be needed.

nit: I assume you mean initializedBlobDataNames not initialed_blobDataNames .

chavdar:
This is not completely related but I think the current implementation of the Extractor needs a major overhaul. I've opened a ticket to follow-up ETL-3070 .

tuGithub:
Chavdar: 

I have addressed the comments you gave, please review it.
Thanks
Min

chavdar:
LGTM

PR #135, opened 2015-05-19
Refactor Kafka adapter for complex watermark

zliu41:
- Refactored `KafkaSource` and `KafkaExtractor` for the complex watermark:

Now each WorkUnit in a MultiWorkUnit returned by `KafkaSource.getWorkUnits()` is
responsible for pulling multiple Kafka partitions of the same topic, so that
these partitions can share output files rather than creating its own output files.

To this end, a method `KafkaSource.squeezeMultiWorkUnits()` is added to combine all WorkUnits in a MultiWorkUnit into a single WorkUnit.

`KafkaExtractor.readRecord()` is rewritten to pull from multiple partitions.

- Small modification to complex watermark:

  - Changed the implementation of `WorkUnit.getLowWatermark` and
`WorkUnit.getExpectedHighWatermark`. The original implementation doesn't work and
throws "Not a JsonObject" exceptions. 
  - Changed return type of
`WorkUnitState.getActualHighWatermark()` from `JsonElement` to `String`. In the
original implementation, `GSON.toJsonTree(String)` will convert `"` in the input
string to `\"` in the JsonElement, which will break if we pass this JsonElement to GSON.fromJson.
  - Added method `setWatermarkInterval` to `SourceState`, since we need a way to assign watermarks to workunits.

- Modified the best fit implementation in KafkaSource, the original one was incorrect.

- Modified the worst fit implementation in KafkaSource, using a MinMaxPriorityQueue to improve efficiency.

- Use `com.google.common.net.HostAndPort`

- Other small refactoring for better readability.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
@@ -27,6 +27,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
+import gobblin.source.extractor.WatermarkInterval;
 import gobblin.source.workunit.WorkUnit;
 import gobblin.source.workunit.Extract;
 
@@ -47,8 +48,8 @@ public class SourceState extends State {
 
   private final List<WorkUnitState> previousTaskStates = Lists.newArrayList();
   private static final Set<Extract> extractSet = Sets.newConcurrentHashSet();
-  private static final DateTimeFormatter DTF =
-      DateTimeFormat.forPattern("yyyyMMddHHmmss").withLocale(Locale.US).withZone(DateTimeZone.UTC);
+  private static final DateTimeFormatter DTF = DateTimeFormat.forPattern("yyyyMMddHHmmss").withLocale(Locale.US)
+      .withZone(DateTimeZone.UTC);
 
   /**
    * Default constructor.
@@ -67,13 +68,17 @@ public class SourceState extends State {
     this.previousTaskStates.addAll(previousTaskStates);
   }
 
+  public void setWatermarkInterval(WatermarkInterval interval) {
+    setProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY, interval.toJson());
+  }
+
   /**
    * Get a (possibly empty) list of {@link WorkUnitState}s from the previous job run.
    *
    * @return (possibly empty) list of {@link WorkUnitState}s from the previous job run
    */
   public List<WorkUnitState> getPreviousWorkUnitStates() {
-    return ImmutableList.<WorkUnitState>builder().addAll(this.previousTaskStates).build();
+    return ImmutableList.<WorkUnitState> builder().addAll(this.previousTaskStates).build();
   }
 
   /**
@@ -109,8 +114,7 @@ public class SourceState extends State {
   }
 
   @Override
-  public void write(DataOutput out)
-      throws IOException {
+  public void write(DataOutput out) throws IOException {
     out.writeInt(this.previousTaskStates.size());
     for (WorkUnitState state : this.previousTaskStates) {
       state.write(out);
@@ -119,8 +123,7 @@ public class SourceState extends State {
   }
 
   @Override
-  public void readFields(DataInput in)
-      throws IOException {
+  public void readFields(DataInput in) throws IOException {
     int size = in.readInt();
     for (int i = 0; i < size; i++) {
       WorkUnitState state = new WorkUnitState();
--- a/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
@@ -17,7 +17,6 @@ import java.io.IOException;
 import java.util.Set;
 
 import com.google.common.collect.Sets;
-import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 
 import gobblin.source.extractor.Watermark;
@@ -43,8 +42,6 @@ public class WorkUnitState extends State {
 
   private Watermark actualHighWatermark;
 
-  private static final Gson GSON = new Gson();
-
   /**
    * Runtime state of the {@link WorkUnit}.
    *
@@ -55,7 +52,12 @@ public class WorkUnitState extends State {
    * </p>
    */
   public enum WorkingState {
-    PENDING, RUNNING, SUCCESSFUL, COMMITTED, FAILED, CANCELLED
+    PENDING,
+    RUNNING,
+    SUCCESSFUL,
+    COMMITTED,
+    FAILED,
+    CANCELLED
   }
 
   private WorkUnit workunit;
@@ -107,10 +109,10 @@ public class WorkUnitState extends State {
   /**
    * Get the actual high {@link Watermark} as a {@link JsonElement}.
    *
-   * @return a {@link JsonElement} representing the actual high {@link Watermark}.
+   * @return a JSON string representing the actual high {@link Watermark}.
    */
-  public JsonElement getActualHighWatermark() {
-    return GSON.toJsonTree(getProp(ConfigurationKeys.WORK_UNIT_STATE_ACTUAL_HIGH_WATER_MARK_KEY));
+  public String getActualHighWatermark() {
+    return getProp(ConfigurationKeys.WORK_UNIT_STATE_ACTUAL_HIGH_WATER_MARK_KEY);
   }
 
   /**
@@ -224,15 +226,13 @@ public class WorkUnitState extends State {
   }
 
   @Override
-  public void readFields(DataInput in)
-      throws IOException {
+  public void readFields(DataInput in) throws IOException {
     this.workunit.readFields(in);
     super.readFields(in);
   }
 
   @Override
-  public void write(DataOutput out)
-      throws IOException {
+  public void write(DataOutput out) throws IOException {
     this.workunit.write(out);
     super.write(out);
   }
--- a/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
+++ b/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
@@ -19,8 +19,9 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 
-import com.google.gson.Gson;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 
 import gobblin.source.extractor.Extractor;
 import gobblin.source.extractor.Watermark;
@@ -40,8 +41,6 @@ import gobblin.source.extractor.WatermarkInterval;
 public class WorkUnit extends State {
 
   private Extract extract;
-
-  private static final Gson GSON = new Gson();
 
   /**
    * Default constructor.
@@ -115,7 +114,7 @@ public class WorkUnit extends State {
    * @return a {@link JsonElement} representing the low {@link Watermark}.
    */
   public JsonElement getLowWatermark() {
-    return GSON.toJsonTree(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY)).getAsJsonObject()
+    return ((JsonObject) (new JsonParser().parse(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY))))
         .get(WatermarkInterval.LOW_WATERMARK_TO_JSON_KEY);
   }
 
@@ -125,7 +124,7 @@ public class WorkUnit extends State {
    * @return a {@link JsonElement} representing the expected high {@link Watermark}.
    */
   public JsonElement getExpectedHighWatermark() {
-    return GSON.toJsonTree(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY)).getAsJsonObject()
+    return ((JsonObject) (new JsonParser().parse(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY))))
         .get(WatermarkInterval.EXPECTED_HIGH_WATERMARK_TO_JSON_KEY);
   }
 
@@ -174,15 +173,13 @@ public class WorkUnit extends State {
   }
 
   @Override
-  public void readFields(DataInput in)
-      throws IOException {
+  public void readFields(DataInput in) throws IOException {
     super.readFields(in);
     this.extract.readFields(in);
   }
 
   @Override
-  public void write(DataOutput out)
-      throws IOException {
+  public void write(DataOutput out) throws IOException {
     super.write(out);
     this.extract.write(out);
   }
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -18,7 +18,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.PriorityQueue;
-import java.util.Queue;
 import java.util.Set;
 
 import org.slf4j.Logger;
@@ -26,21 +25,24 @@ import org.slf4j.LoggerFactory;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.SourceState;
-import gobblin.configuration.State;
 import gobblin.configuration.WorkUnitState;
+import gobblin.source.extractor.WatermarkInterval;
 import gobblin.source.extractor.extract.EventBasedSource;
 import gobblin.source.workunit.Extract;
 import gobblin.source.workunit.MultiWorkUnit;
 import gobblin.source.workunit.WorkUnit;
+import gobblin.util.KafkaUtils;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.MinMaxPriorityQueue;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closer;
 import com.google.common.primitives.Longs;
+import com.google.gson.Gson;
 
 
 /**
@@ -51,6 +53,7 @@ import com.google.common.primitives.Longs;
 public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
 
   private static final Logger LOG = LoggerFactory.getLogger(KafkaSource.class);
+  private static final Gson GSON = new Gson();
 
   public static final String TOPIC_BLACKLIST = "topic.blacklist";
   public static final String TOPIC_WHITELIST = "topic.whitelist";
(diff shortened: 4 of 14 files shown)


gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -67,13 +68,17 @@ public SourceState(State properties, List<WorkUnitState> previousTaskStates) {
     this.previousTaskStates.addAll(previousTaskStates);
   }
 
+  public void setWatermarkInterval(WatermarkInterval interval) {
```
sahilTakiar: You should be able to set the `WatermarkInterval` using the `WorkUnit(SourceState state, Extract extract, WatermarkInterval watermarkInterval)` constructor. I think logically it makes more sense to specify the `WatermarkInterval` on a `WorkUnit` level than a `SourceState` level because a `WorkUnit` a single `WorkUnit` is responsible for pulling data from a low `Watermark` to a high `Watermark`.

akshayn1107:
I haven't looked too deeply into this PR but had a few questions based on the PR summary:

Does this mean there will be one extractor per topic no matter how many partitions there are? Before, I believe there was 1 task per partition per topic, so the job was parallelized based on partitions, i.e. if there are 10 partitions, there could be 10 different threads pulling from Kafka. Is this now a sequential process?

In the old way (where there was one extractor per partition per topic), we basically wrote this module: https://github.com/linkedin/gobblin/blob/c71a32875a97386d0214dbece53559ccdf8cd875/gobblin-core/src/main/java/gobblin/publisher/BatchKafkaData.java
which can get called from the publisher and basically batches together all kafka data for a given topic. This way, the extractors can still run in parallel, but the data gets batched at the publish step.

zliu41:
@akshayn1107 Not all partitions of a topic go to the same workunit/extractor, it depends on the number of mappers (`mr.job.max.mappers`). If that number is more than a third of the total number of partitions, then most likely each workunit will still have only one partition.

If the number of mappers is very small compared to the number of partitions (in our use case we use 200 mappers for 5000+ partitions), then we try to assign partitions of the same topic to the same workunit, so that these partitions can share output files. This helps reduce the number of small files generated.

The number of workunits/extractors is no less than 3x number of mappers. We use `workUnitMaxSize = total size of all partitions / num of mappers / 3`
If the total size of all partitions of a topic is smaller than workUnitMaxSize, all of them go to the same workunit. Otherwise, they go to different workunits via bin packing.

Let me know if this works for you.

akshayn1107:
@zliu41 Thanks for clarifying, that makes sense.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -107,10 +109,10 @@ public void setWorkingState(WorkingState state) {
   /**
    * Get the actual high {@link Watermark} as a {@link JsonElement}.
    *
-   * @return a {@link JsonElement} representing the actual high {@link Watermark}.
+   * @return a JSON string representing the actual high {@link Watermark}.
    */
-  public JsonElement getActualHighWatermark() {
-    return GSON.toJsonTree(getProp(ConfigurationKeys.WORK_UNIT_STATE_ACTUAL_HIGH_WATER_MARK_KEY));
+  public String getActualHighWatermark() {
```
sahilTakiar: Can you keep the return type as `JsonElement` and use a `JsonParser` like you did in the other classes.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -115,7 +114,7 @@ public Extract getExtract() {
    * @return a {@link JsonElement} representing the low {@link Watermark}.
    */
   public JsonElement getLowWatermark() {
-    return GSON.toJsonTree(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY)).getAsJsonObject()
+    return ((JsonObject) (new JsonParser().parse(getProp(ConfigurationKeys.WATERMARK_INTERVAL_VALUE_KEY))))
```
sahilTakiar: Can you make the `JsonParser` an instance variable?

sahilTakiar:
@zliu41 I fixed the bugs you brought up in a new PR, and also added a unit test: https://github.com/linkedin/gobblin/pull/141

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * A MultiLongWatermark w1 is less than w2, if there exists i, such that
+   * the 0th to (i-1)th values of w1 are equivalent as those of w2, and the ith
+   * value of w1 is less than the that of w2.
+   *
+   * @param wm must be an instance of MultiLongWatermark, and this.size() must
+   * equal wm.size().
+   */
+  @Override
+  public int compareTo(Watermark wm) {
```
ibuenros: Do we use anywhere the fact that watermark is Comparable? MultiLongWatermark does not actually respect the Comparable contract, which requires a total ordering of objects. IMO we should remove comparable, and not call this method compareTo (which would suggest that watermarks are sortable).

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,159 @@
...
+    Preconditions.checkArgument(wm instanceof MultiLongWatermark);
```
ibuenros: The compareTo contract requires this to throw a ClassCastException.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,159 @@
...
+   */
+  @Override
+  public short calculatePercentCompletion(Watermark lowWatermark, Watermark highWatermark) {
+    Preconditions.checkArgument(lowWatermark instanceof MultiLongWatermark, String.format(
+        "Arguments of %s.%s must be of type %s", MultiLongWatermark.class.getSimpleName(), Thread.currentThread()
+            .getStackTrace()[1].getMethodName(), MultiLongWatermark.class.getSimpleName()));
+
+    long pulled = ((MultiLongWatermark) lowWatermark).getGap(this);
+    long all = ((MultiLongWatermark) lowWatermark).getGap((MultiLongWatermark) highWatermark);
+    Preconditions.checkArgument(all > 0);
```
ibuenros: all is not an argument. this should be checkState instead.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,159 @@
...
+   */
+  public long set(int idx, long value) {
+    Preconditions.checkElementIndex(idx, this.values.size());
+    return this.values.set(idx, value);
+  }
+
+  /**
+   * Add a value to the watermark. This will increase size() by 1.
+   */
+  public void add(long value) {
```
ibuenros: should the watermark be allowed to grow when it is already in use?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,159 @@
...
+
+import gobblin.source.extractor.Watermark;
+
+
+/**
+ * A {@link gobblin.source.extractor.Watermark} that holds multiple long values.
+ *
+ * @author ziliu
+ */
+public class MultiLongWatermark implements Watermark {
```
ibuenros: Just to confirm. Say lowWatermark is (0,0) and highWatermark is (10,10). Say we pull partition 1 until it times out at current watermark (5,0). What happens next, do we start pulling partition 2? So then current watermark might be (5,3)?
If so, this would raise issues when comparing watermarks, and when computing completion.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -400,32 +449,55 @@ private boolean shouldMoveToLatestOffset(KafkaPartition partition, SourceState s
...
+    partitionsState.addAll(state);
+    partitionsState.setProp(TOPIC_NAME, partitions.get(0).getTopicName());
+    partitionsState.setProp(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY, partitions.get(0).getTopicName());
+    for (int i = 0; i < partitions.size(); i++) {
+      partitionsState.setProp(KafkaUtils.getPartitionPropName(KafkaSource.PARTITION_ID, i), partitions.get(i).getId());
+      partitionsState.setProp(KafkaUtils.getPartitionPropName(KafkaSource.LEADER_ID, i), partitions.get(i).getLeader()
+          .getId());
+      partitionsState.setProp(KafkaUtils.getPartitionPropName(KafkaSource.LEADER_HOSTANDPORT, i), partitions.get(i)
+          .getLeader().getHostAndPort());
+      partitionsState.setWatermarkInterval(interval);
```
ibuenros: this could be done only once

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: Not used by KafkaExtractor, but I suppose some extractors may want to compare current watermark with high watermark, to decide whether to continue pulling. @sahilTakiar may have a better answer.

For MultiLongWatermark, I can allow comparing two instances with different sizes, which makes it total order.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: Timeout is applied on the extractor, not individual partitions. So extractor will stop and the watermark will be (5, 0).

ibuenros: Not necessarily a timeout then, but any event that makes the extractor stop consuming a partition prematurely and move to the next one.

zliu41: Yes, if can no longer consume a partition (e.g., broker went down), then it'll move to the next partition. Hence KafkaExtractor doesn't use compareTo to determine whether it is finished.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: Good point, there's no need to allow this.

zliu41:
Merged with #141 and addressed comments.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
sahilTakiar: The fact that `Watermark` currently uses `Comparable` isn't leveraged anywhere. In fact the `calculatePercentCompletion` method isn't invoked anywhere either. It was deemed too much additional work to modify the framework as this point in time in order to integrate with these methods. However, I do think that keeping the `Watermark` interface `Comparable` has advantages. I think that `Watermark`s should actually be sortable. In most data ingestion use cases, one typically has a low watermark and a high watermark, and a single execution pulls data from the low watermark to the high watermark. The next execution takes the high watermark of the previous run, and uses it to calculate a new low and high watermark. For example, a MySQL puller may pull data based on the value of a SystemModifiedTimestamp column. Its first execution may pull data from [0, x = current epoch time), the next execution will pull from [x, current epoch time), etc; where the watermark is a timestamp.

I think the actual `compareTo` method may come in handy in the `WatermarkInterval` class, where a user constructs a `WatermarkInterval` that spans from a low watermark to a high watermark. Each `WorkUnit` has a single `WatermarkInterval` assigned to it. The framework can compare the low watermark and the high watermark using the `compareTo` method in order to check that one is actually lower than the other.

Here is another use case: the `Extractor` wants to pull data from a low watermark to a high watermark, but doesn't want to worry about tracking when its current watermark is >= the high watermark. By using the `compareTo` method the framework can handle this logic for the `Extractor`. This way the `Extractor` is just concerned about pulling data, but doesn't need to worry about tracking when it needs to stop pulling data.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private static final Gson GSON = new Gson();
+
+  private final List<Long> values;
```
sahilTakiar: I probably don't have time to go through this PR anytime soon, so it may best for someone else to take another look. Just one question about the `Watermark` implementation. Does the each index of this list refer to a specific partition, and then each value correspond to some offset?

zliu41: Yes.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
ibuenros: Yes, watermarks need to be comparable in some way, but there's no reason they need to have a total order as is implied by the comparable interface. In the multilong watermark, for example, the watermarks (5,5,5) and (5,6,4) should not be considered sortable. 

Instead, I suggest that watermarks should implement a less restrictive interface that merely requires a "isLargerThan" method, then (5,5,5).isLargerThan(5,5,4) is true, but (5,5,5).isLargerThan(5,6,4) is false, and also (5,6,4).isLargerThan(5,5,5) is false.

The requirement would be that isLargerThan is transitive, and that every time the watermark is incremented it is true that low_watermark < old_watermark < new_watermark < high_watermark, otherwise throw an IllegalStateException. This offers all the requirements explained above, without implying or requiring total order guarantees.

sahilTakiar: Ok, I'm probably not super well versed on the complete guarantees of the `Comparable` interface, can you talk with @chavdar to discuss this further. I probably won't have time until next week to deep dive into this.

zliu41: @chavdar what do you think?

I agree that forcing total order is inappropriate. There may be an extractor that allows both [lw=555, hw=564] (meaning it needs to pull the 2nd dimension) and [lw=564, hw=555] (meaning it needs to pull the 3rd dimension). In this case there's no total order.

If Gobblin wants to check the validity of WatermarkInterval, then we need a `isLargerThanOrEqual` or `isSmallerThanOrEqual`. But this doesn't seem necessary to me, since lw > hw, may not necessarily be an error case.

gobblin-utility/src/main/java/gobblin/util/KafkaUtils.java
```diff
@@ -0,0 +1,106 @@
...
+  public static String getPartitionPropName(String baseName, int idx) {
+    return baseName + "." + idx;
+  }
+
+  /**
+   * Get the average event size of a partition, which is stored in property "[topicname].[partitionid].avg.event.size".
+   * If state doesn't contain this property, it returns defaultSize.
+   */
+  public static long getPartitionAvgEventSize(State state, KafkaPartition partition, long defaultSize) {
+    return state.getPropAsLong(KafkaSource.TOPIC_NAME + "." + KafkaSource.PARTITION_ID + "."
```
tuGithub: Other functions have precondition check:
Preconditions.checkArgument(state.contains(KafkaSource.TOPIC_NAME) && state.contains(KafkaSource.PARTITION_ID));

In order to maintain the same style in this class, should add this check ?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: Talked to @chavdar , will remove Comparable from Watermark, and leave it to each extractor to interpret and validate lw and hw.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Copy constructor.
+   */
+  public MultiLongWatermark(MultiLongWatermark other) {
+    this.values = Lists.newArrayList(other.values);
```
tuGithub: Consider use java native:

this.values =new ArrayList<Long>(other.values);

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  public MultiLongWatermark(List<Long> values) {
+    this.values = Lists.newArrayList(values);
```
tuGithub: this.values =new ArrayList(other.values);

Then do not need to import jcommand.Lists

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,150 @@
...
+   */
+  @Override
+  public int compareTo(Watermark wm) {
+    if (!(wm instanceof MultiLongWatermark)) {
+      throw new ClassCastException("Cannot compare MultiLongWatermark with " + wm.getClass().getSimpleName());
+    }
+    MultiLongWatermark other = (MultiLongWatermark) wm;
+
+    ComparisonChain cc = ComparisonChain.start().compare(this.size(), other.size());
+    for (int i = 0; i < this.values.size(); i++) {
```
tuGithub: if the size are different for values, the java.lang.IndexOutOfBoundsException: will be thrown

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
```diff
@@ -0,0 +1,150 @@
...
+   * of events pulled.
+   *
+   * @return a percentage value between 0 and 100.
+   */
+  @Override
+  public short calculatePercentCompletion(Watermark lowWatermark, Watermark highWatermark) {
+    Preconditions.checkArgument(lowWatermark instanceof MultiLongWatermark, String.format(
+        "Arguments of %s.%s must be of type %s", MultiLongWatermark.class.getSimpleName(), Thread.currentThread()
+            .getStackTrace()[1].getMethodName(), MultiLongWatermark.class.getSimpleName()));
+
```
tuGithub: Need to check highWatermark instance of MultiLongWaterMark as well

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: Should have used `com.google.common.collect.Lists` instead of `jcommand.Lists`. Updated the import.
`Lists.newArrayList()` is used throughout Gobblin in favor of `new ArrayList()`.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/MultiLongWatermark.java
zliu41: See above comments.

zliu41:
Updated.

ibuenros:
LGTM

tuGithub:
LGTM

PR #136, opened 2015-05-19
Fixed unit tests around JobLauncher classes after recent changes on writer file naming

liyinan926:
A recent change on writer file naming allows user to fully control the writer file name including its extension through the configuration property `writer.file.name`. This property used to be used as the prefix for writer file names that will also include the task IDs, which guarantees that each task writes to a different file. However, the test job configurations explicitly set `writer.file.name=gobblin-test`, which means every tasks of a job will write to the same file. This issue shows up in `LocalJobLauncherTest` and `MRJobLauncherTest` in the form of `WARN` messages that complained that a writer of a task is trying to write to an existing file named `gobblin-test`. This commit fixed the tests by removing `writer.file.name` from the test job configurations.


sahilTakiar:
LGTM

chavdar:
Let's chat in person about the writer.file.name change.

liyinan926:
Talked with @chavdar, will merge this first and discuss the semantic change later.

PR #120, opened 2015-05-04
Code refactoring around JobLauncher (see details below)

liyinan926:
1. Refactored `JobLauncher` to extend `Closeable` and code around its usage to properly close it like other `Closeable`s.
2. Made clear that `JobLauncher` should not be reused and refactored code to somehow reinforce it.
3. Refactored implementation classes of `JobLauncher` to follow the above changes.
4. In `MRJobLauncher`, the `close()` method will now kill the Hadoop job if it has not completed.
5. Other code refactoring.

PR #139, opened 2015-05-19
Fixed a bug that caused the Hudson build failures

liyinan926:
Recently, as part of pul request #120, the method `commitJob(JobContext jobContext)` in `OutputCommitter` was overriden in  `GobblinOutputCommitter`, which called `cleanUpWorkingDirectory` 
to delete the working directory of the MR job upon completion and committing of the MR job. However, there are two versions of the method `commitJob(JobContext jobContext)`, one using `org.apache.hadoop.mapred.JobContext`, and the other using `org.apache.hadoop.mapreduce.JobContext`. The one that was overriden is the latter, and this is the one called in Hadoop 2. The issue is that the Gobblin job needs to read the task states stored under the working directory after the Hadoop job finishes, at which point the working directory has already been deleted if run on Hadoop 2. The issue didn't show up in Hadoop 1 is because the version of that method for Hadoop 1 was not overriden.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinOutputCommitter.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinOutputCommitter.java
@@ -107,22 +107,6 @@ public class GobblinOutputCommitter extends OutputCommitter {
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


ibuenros:
LGTM

PR #142, opened 2015-05-20
Fixed a bug that caused the NoClassDefFoundError on Hadoop

liyinan926:
The problem is due to code refactoring as part of #120, an instance of the Hadoop `Job` class is created in the constructor of `MRJobLauncher` before the dependent jars are added to DistributedCache (that updates `conf`). So the `Job` is missing the info on jars added to the DistributedCache. 


ydailinkedin:
Yinan, I've tested and it worked. LGTM.

PR #143, opened 2015-05-21
ETL-3020: Enable nullify fields in JsonIntermediateToAvroConverter

ydailinkedin:
Added two configurations. CONVERTER_AVRO_NULLIFY_FIELDS_ENABLED is to enable nullifying field and CONVERTER_AVRO_NULLIFY_FIELDS_ORIGINAL_SCHEMA_PATH is to specify the path where we can get the original schema. If CONVERTER_AVRO_NULLIFY_FIELDS_ORIGINAL_SCHEMA_PATH is not specified, the behavior is to look for the previously published avro file to get the old schema.

When having two schema, the new one may have removed some fields that were previously existed in the old schema. The desired behavior is to retain those fields, and use "union" to merge with "null" as the default value. 

Added unit tests and also tested on Nertz.


gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -174,6 +174,10 @@
   public static final String CONVERTER_STRING_SPLITTER_DELIMITER = "converter.string.splitter.delimiter";
   public static final String CONVERTER_CSV_TO_JSON_ENCLOSEDCHAR = "converter.csv.to.json.enclosedchar";
   public static final String DEFAULT_CONVERTER_CSV_TO_JSON_ENCLOSEDCHAR = "\0";
+  public static final String CONVERTER_AVRO_NULLIFY_FIELDS_ENABLED = "converter.avro.nullify.fileds.enabled";
```
chavdar: s/fileds/fields/

Same thing for the other config.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -174,6 +174,10 @@
   public static final String CONVERTER_STRING_SPLITTER_DELIMITER = "converter.string.splitter.delimiter";
   public static final String CONVERTER_CSV_TO_JSON_ENCLOSEDCHAR = "converter.csv.to.json.enclosedchar";
   public static final String DEFAULT_CONVERTER_CSV_TO_JSON_ENCLOSEDCHAR = "\0";
+  public static final String CONVERTER_AVRO_NULLIFY_FIELDS_ENABLED = "converter.avro.nullify.fields.enabled";
```
chavdar: Please move these config options to the class where they are used: JsonIntermediateToAvroConverter

gobblin-core/src/main/java/gobblin/converter/avro/JsonIntermediateToAvroConverter.java
```diff
@@ -103,13 +110,50 @@ public Schema convertSchema(JsonArray schema, WorkUnitState workUnit)
...
+      originalSchemaPath =
+          WriterUtils.getDataPublisherFinalDir(workUnitState,
+              workUnitState.getPropAsInt(ConfigurationKeys.FORK_BRANCHES_KEY, 1),
+              workUnitState.getPropAsInt(ConfigurationKeys.FORK_BRANCH_ID_KEY, 0)).getParent();
+    }
+    try {
+      Schema prevSchema = AvroUtils.getDirectorySchema(originalSchemaPath, conf, false);
+      Schema mergedSchema = AvroUtils.nullifyFiledsForSchemaMerge(prevSchema, currentAvroSchema);
+      return mergedSchema;
+    } catch (Exception e) {
```
chavdar: 1. Please avoid blanket catch (Exception e) statements
2. Please log the exception.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+   * @param latest true to return latest schema, false to return oldest schema
+   * @return the latest/oldest schema in the directory
+   * @throws IOException
+   */
+  public static Schema getDirectorySchema(Path directory, Configuration conf, boolean latest) throws IOException {
+    Schema schema = null;
+    Closer closer = Closer.create();
+    try {
+      List<FileStatus> files = getDirectorySchemaHelper(directory, FileSystem.get(conf));
+      FileStatus file = latest ? files.get(0) : files.get(files.size() - 1);
```
chavdar: Is it possible that the list is empty? Please be defensive.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+      LOG.info("Path to get the avro schema: " + file);
+      FsInput fi = new FsInput(file.getPath(), conf);
+      GenericDatumReader<GenericRecord> genReader = new GenericDatumReader<GenericRecord>();
+      schema = new DataFileReader<GenericRecord>(fi, genReader).getSchema();
+    } catch (IOException e) {
+      throw new IOException("Cannot get the schema for directory " + directory);
```
chavdar: please add the parent exception so we can preserve the stack trace: throw new IOException("Cannot get the schema for directory " + directory, e);

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
       closer.close();
     }
   }
+
+  /**
+   * Get the latest avro schema for a directory
+   * @param directory the input dir that contains avro files
+   * @param conf configuration
+   * @param latest true to return latest schema, false to return oldest schema
+   * @return the latest/oldest schema in the directory
```
chavdar: What is the return result if there are no data files.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+    if (oldSchema == null) {
+      return newSchema;
+    }
+
+    List<Field> combinedFields = new ArrayList<Field>();
+    for (Field newFld : newSchema.getFields()) {
+      combinedFields.add(new Field(newFld.name(), newFld.schema(), newFld.doc(), newFld.defaultValue()));
+    }
+
+    List<Schema> union = new ArrayList<Schema>();
```
chavdar: Shouldn't this be after line 246, i.e. only if oldFldSchema is a union.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+   * @param oldSchema
+   * @param newSchema
+   * @return schema that contains all the fields in both old and new schema.
+   */
+  public static Schema nullifyFiledsForSchemaMerge(Schema oldSchema, Schema newSchema) {
+    if (oldSchema == null) {
+      return newSchema;
+    }
+
+    List<Field> combinedFields = new ArrayList<Field>();
```
chavdar: Please make sure that the schema is a record schema.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +170,116 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+        Schema.createRecord(newSchema.getName(), newSchema.getDoc(), newSchema.getNamespace(), newSchema.isError());
+    mergedSchema.setFields(combinedFields);
+    return mergedSchema;
+  }
+
+  /**
+   * This method is to filter out the .avro files that need to be processed.
+   * @return pathFilter
+   */
+  protected static PathFilter getAvroFileFilter() {
```
chavdar: This does not seems to be used.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
ydailinkedin: The return result would be null. In method `nullifyFieldsForSchemaMerge`, if the oldSchema is null, we will retain the new schema.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
ydailinkedin: Yes, you are right. The error will be triggered when multiple fields are missing in the new schema. Fixed and new unit test added.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+        LOG.warn("There is no previous avro file in the directory: " + directory);
+      } else {
+        FileStatus file = latest ? files.get(0) : files.get(files.size() - 1);
+        LOG.info("Path to get the avro schema: " + file);
+        FsInput fi = new FsInput(file.getPath(), conf);
+        GenericDatumReader<GenericRecord> genReader = new GenericDatumReader<GenericRecord>();
+        schema = new DataFileReader<GenericRecord>(fi, genReader).getSchema();
+      }
+    } catch (IOException ioe) {
+      throw new IOException("Cannot get the schema for directory " + directory, ioe);
```
liyinan926: Please do:

```
catch (IOException ioe) {
  ...
} catch (Throwable t) {
  throw closer.rethrow(t);
} finally {
  closer.close();
}
```

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+
+  private static List<FileStatus> getDirectorySchemaHelper(Path directory, FileSystem fs) throws IOException {
+    List<FileStatus> files = new ArrayList<FileStatus>();
+    if (fs.exists(directory)) {
+      getAllNestedAvroFiles(fs.getFileStatus(directory), files, fs);
+      if (files.size() > 0) {
+        Collections.sort(files, new Comparator<FileStatus>() {
+          @Override
+          public int compare(FileStatus file1, FileStatus file2) {
+            return Long.valueOf(file2.getModificationTime()).compareTo(Long.valueOf(file1.getModificationTime()));
```
liyinan926: You can use `Longs.compare`.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+    } catch (IOException ioe) {
+      throw new IOException("Cannot get the schema for directory " + directory, ioe);
+    } finally {
+      closer.close();
+    }
+    return schema;
+  }
+
+  private static List<FileStatus> getDirectorySchemaHelper(Path directory, FileSystem fs) throws IOException {
+    List<FileStatus> files = new ArrayList<FileStatus>();
```
liyinan926: Please use `Lists.newArrayList`.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+          }
+        });
+      }
+    }
+    return files;
+  }
+
+  private static void getAllNestedAvroFiles(FileStatus dir, List<FileStatus> files, FileSystem fs) throws IOException {
+    if (dir.isDir()) {
+      for (FileStatus f : fs.listStatus(dir.getPath())) {
```
liyinan926: `listStatus` could potentially return null, so better to be defensive here.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+      LOG.warn("No previous schema available, use the new schema instead.");
+      return newSchema;
+    }
+
+    if (!(oldSchema.getType().equals(Type.RECORD) && newSchema.getType().equals(Type.RECORD))) {
+      LOG.warn("Both previous schema and new schema need to be record type. Quit merging schema.");
+      return newSchema;
+    }
+
+    List<Field> combinedFields = new ArrayList<Field>();
```
liyinan926: ditto. Please use `Lists.newArrayList`.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -156,4 +169,110 @@ public static GenericRecord convertRecordSchema(GenericRecord record, Schema new
...
+    for (Field newFld : newSchema.getFields()) {
+      combinedFields.add(new Field(newFld.name(), newFld.schema(), newFld.doc(), newFld.defaultValue()));
+    }
+
+    for (Field oldFld : oldSchema.getFields()) {
+      if (newSchema.getField(oldFld.name()) == null) {
+        List<Schema> union = new ArrayList<Schema>();
```
liyinan926: Ditto.

chavdar:
No more comments from me. Once Yinan's comments have been addressed, you can merge.

ydailinkedin:
Updated.

ydailinkedin:
Updated.

PR #146, opened 2015-05-27
Fixed new high priority FindBugs warnings

liyinan926:
Signed-off-by: Yinan Li <liyinan926@gmail.com>

PR #147, opened 2015-05-27
Fix findbugs warnings.

ibuenros:


gobblin-runtime/src/main/java/gobblin/runtime/util/MetricNames.java
```diff
@@ -20,7 +20,7 @@
  */
 public class MetricNames {
   public static class LauncherTimings {
-    public static String BASE_NAME = "gobblin.launcher.timings";
+    public final static String BASE_NAME = "gobblin.launcher.timings";
     public final static String WORK_UNITS_CREATE = MetricRegistry.name(BASE_NAME, "work.units.create");
```
liyinan926: I changed all of them to `static final` to be consistent with everywhere elsein #146.

liyinan926:
LGTM.

PR #150, opened 2015-05-27
Refactored some metric names so the naming is more consistent

liyinan926:


ibuenros:
Closing the Timer.Context does nothing but record the time. I think ensuring a close does nothing but possibly confuse metrics.

liyinan926:
OK, I didn't realize that. I was just thinking based on the name that it returns a `MetricContext`, which should be properly closed.

liyinan926:
OK, updated.

ibuenros:
LGTM

PR #151, opened 2015-05-29
Fixed a bug with the generation of JobExecutionInfo regarding start and end dates

liyinan926:
I was testing the job execution history store and I found this bug. Basically if start or end time is not set in `JobState` or `TaskState`, the default zero will be used when constructing a `JobExecutionInfo` or `TaskExecutionInfo`. However, when `DataaseJobHistoryStore.put(JobExecutionInfo)` is called, a zero start/end time maps to an invalid TIMESTAMP and will be rejected by the database.

@ydailinkedin can you help review this?


ydailinkedin:
@liyinan926 , I just have one question. With this fix, if start or end time is not set, then JobExecutionInfo/TaskExecutionInfo basically would still be constructed but not contain the start/end time. Is that right?

liyinan926:
@ydailinkedin yes, and `JobExecutionInfo.hasStartTime()` will return `false`.

ydailinkedin:
Thanks! LGTM.

PR #152, opened 2015-06-01
Generifying convertJsonToWatermark method in WatermarkSerializerHelper class

suvodeep-pyne:
Generifying convertJsonToWatermark method in WatermarkSerializerHelper class to return objects typed with the parameter provided


sahilTakiar:
LGTM

PR #154, opened 2015-06-04
Kafka Adapter Bug Fix

zliu41:
Fixed the following bugs:

- `KafkaExtractor.shouldLogError()`, original logic was wrong. Also, originally it didn't increment the error count, which it should do.
- `KafkaExtractor.readRecordImpl()`: the `if (currentPartitionFinished())` block should be inside `while`. Otherwise, if `allPartitionsFinished()=true`, `currentPartitionFinished()` will fail because there's no current partition.
- `KafkaExtractor.readRecordImpl()`: LOG should print out stack trace.
- `KafkaExtractor.moveToNextPartition()`: added a `if` block around `switchMetricContextToCurrentPartition`, since if all partitions are finished, then there's no current partition.
- `KafkaAvroSchemaRegistry.getSchemaByID()`: throw `SchemaNotFoundException` instead of `RuntimeException`.


zliu41:
@ibuenros please take a look. thanks.

ibuenros:
LGTM

PR #155, opened 2015-06-04
Do not persist state of a job if its output data has not been committed

liyinan926:
This pull request is motivated by a case from the Kafka adapter that doesn't want to have job state persisted if the job failed (so to be consistent with the Camus behavior). This pull request has the following changes (UPDATED):
1. `AbstractJobLauncher` will no longer persist job state of a job if its output data has not been committed successfully.
2. Added a check in `MRJobLauncher` on the number of `TaskState`s collected from the mapper output. If this number is less than the expected number (the number of tasks launched), the job state is set to `FAILED`.
3. Code refactoring.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -532,9 +532,9 @@ public abstract class AbstractJobLauncher implements JobLauncher {
    */
   private void persistJobState(JobState jobState) throws IOException {
     for (TaskState taskState : jobState.getTaskStates()) {
-      // Restore the previous actual high watermark for each task that has not been committed
+      // Backoff the actual high watermark to the low watermark for each task that has not been committed
       if (taskState.getWorkingState() != WorkUnitState.WorkingState.COMMITTED) {
-        taskState.restoreActualHighWatermark();
+        taskState.backoffActualHighWatermark();
       }
     }
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -269,14 +269,11 @@ public abstract class AbstractJobLauncher implements JobLauncher {
 
       setFinalJobState(jobState);
       if (canCommit(this.jobContext.getJobCommitPolicy(), jobState)) {
-        commitJob(jobState);
-        // Persist job state only if and only if data has been successfully committed.
-        // WARNING: there's a case that may result in duplicated data given this way of
-        // doing commit and state persistence. Specifically, commitJob is not atomic, so
-        // it may succeed partially and fail in the middle. In this case, job state will
-        // not be persisted, but output data is partially committed. The next run of the
-        // job may pull the committed portion of data again, leading to duplicated data.
-        persistJobState(jobState);
+        try {
+          commitJob(jobState);
+        } finally {
+          persistJobState(jobState);
+        }
       }
 
       Instrumented.endTimer(jobCommitTimer);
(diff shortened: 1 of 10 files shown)


liyinan926:
@zliu41 @ydailinkedin can you review this?

zliu41:
LGTM.
So in each run, if any workunit fails, nothing will be persisted, this run will appear to have never been happened. This definitely works, but in this case is it possible to persist the data and actualHighWm of the successful workunits, while for the failed workunits, don't persist the data but persist their lowWm? This way those successful workunits won't be wasted and the next run can still start at the correct Wm for each workunit.

liyinan926:
It all depends on which `JobCommitPolicy` is used. If `JobCommitPolicy.COMMIT_ON_FULL_SUCCESS` is used, the job state of a failed job won't be persisted since commit will not happen. For a successful job, the job state will be persisted if the commit is successful. If `JobCommitPolicy.COMMIT_ON_PARTIAL_SUCCESS` is used, however, the job state will be persisted if commit is successful.

liyinan926:
Updated c1e8d1b to move `persistJobState` to after `commitJob` to make it clear that job state persistence happens only if job commit is successful.

zliu41:
LGTM

ydailinkedin:
LGTM.

liyinan926:
Updated with the following changes:

1. `persistJobState` is called in a finally clause after `commitJob` so it's always called. 
2. `persistJobState` checks the state of each task and restore the previous actual high watermark for a task if the task has not been successfully committed.
3. `Task` now backs up the previous actual high watermark before starting the extraction.
4. `WorkUnitState.getActualHighWatermark` now checks the existence of `workunit.state.actual.high.water.mark` and returns `null` if the property is not set. `KafkaSource` is updated accordingly to skip previous `WorkUnitState`s that do not have that property set.
5. `KafkaExtractor` is updated to not commit actual high watermark if the task fails.

liyinan926:
Updated e8c32c3.

zliu41:
LGTM

liyinan926:
@chavdar just a remainder please take a look at this PR when you get a chance. Thanks!

liyinan926:
@chavdar have you got a chance to look at this PR? Thanks!

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -415,17 +397,20 @@ public void run() {
...
-    workUnit.setProp(ConfigurationKeys.TASK_ID_KEY, taskId);
-    jobState.addTask();
-    // Pre-add a task state so if the task fails and no task state is written out,
-    // there is still task state for the task when job/task states are persisted.
-    jobState.addTaskState(new TaskState(new WorkUnitState(workUnit)));
+  private void prepareWorkUnits(List<WorkUnit> workUnits, JobState jobState) {
+    int taskIdSequence = 0;
+    for (WorkUnit workUnit : workUnits) {
+      workUnit.setProp(ConfigurationKeys.JOB_ID_KEY, this.jobContext.getJobId());
+      String taskId = JobLauncherUtils.newTaskId(this.jobContext.getJobId(), taskIdSequence++);
```
chavdar: Previously MultiWorkUnits were handled differently. That seems to have changed now. Why? Wouldn't this create problems with existing job states?

chavdar: Although the MultiWorkUnits are flattened, the task was previously generated in a different way.

liyinan926:
Answering your questions above: 
1. It's possible that a `MultiWorkUnit` further contains `MultiWorkUnit`s, so flattening should be recursive.
2. It is not changing the way tasks are created and handled. Each `WorkUnit` (flattened) is still corresponding to one task.

chavdar:
LGTM.

PR #156, opened 2015-06-05
Various fixes to metrics emitted to Kafka in MR mode.

ibuenros:
Summary of changes:
* Added getProperties method to State.
* File metrics reporting can add a suffix to the filename. This is useful in MR jobs where, before, all mappers would attempt to write to the same file.
* Added schema registry functionality to Kafka metrics reporting. Since this needed KafkaAvroSchemaRegistry, refactored this class from gobblin-core to gobblin-metrics.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -370,10 +370,11 @@ public class ConfigurationKeys {
   public static final String METRICS_CONFIGURATIONS_PREFIX = "metrics.";
   public static final String METRICS_LOG_DIR_KEY = METRICS_CONFIGURATIONS_PREFIX + "log.dir";
   public static final String METRICS_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "enabled";
+  public static final String DEFAULT_METRICS_ENABLED = Boolean.toString(true);
   public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
-  public static final String DEFAULT_METRICS_ENABLED = Boolean.toString(false);
+  public static final String DEFAULT_METRICS_FILE_SUFFIX = "";
   public static final String METRICS_REPORTING_FILE_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.enabled";
-  public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(true);
+  public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(false);
   public static final String METRICS_REPORTING_JMX_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.jmx.enabled";
   public static final String DEFAULT_METRICS_REPORTING_JMX_ENABLED = Boolean.toString(false);
   public static final String METRICS_REPORTING_KAFKA_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.enabled";
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -367,21 +367,26 @@ public class ConfigurationKeys {
   /**
    * Common metrics configuration properties.
    */
-  public static final String METRICS_LOG_DIR_KEY = "metrics.log.dir";
-  public static final String METRICS_ENABLED_KEY = "metrics.enabled";
+  public static final String METRICS_CONFIGURATIONS_PREFIX = "metrics.";
+  public static final String METRICS_LOG_DIR_KEY = METRICS_CONFIGURATIONS_PREFIX + "log.dir";
+  public static final String METRICS_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "enabled";
+  public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
   public static final String DEFAULT_METRICS_ENABLED = Boolean.toString(false);
-  public static final String METRICS_REPORTING_FILE_ENABLED_KEY = "metrics.reporting.file.enabled";
+  public static final String METRICS_REPORTING_FILE_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.enabled";
   public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(true);
-  public static final String METRICS_REPORTING_JMX_ENABLED_KEY = "metrics.reporting.jmx.enabled";
+  public static final String METRICS_REPORTING_JMX_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.jmx.enabled";
   public static final String DEFAULT_METRICS_REPORTING_JMX_ENABLED = Boolean.toString(false);
-  public static final String METRICS_REPORTING_KAFKA_ENABLED_KEY = "metrics.reporting.kafka.enabled";
+  public static final String METRICS_REPORTING_KAFKA_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.enabled";
   public static final String DEFAULT_METRICS_REPORTING_KAFKA_ENABLED = Boolean.toString(false);
-  public static final String METRICS_REPORTING_KAFKA_FORMAT = "metrics.reporting.kafka.format";
+  public static final String METRICS_REPORTING_KAFKA_FORMAT = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.format";
   public static final String DEFAULT_METRICS_REPORTING_KAFKA_FORMAT = "json";
-  public static final String METRICS_KAFKA_BROKERS = "metrics.reporting.kafka.brokers";
-  public static final String METRICS_KAFKA_TOPIC = "metrics.reporting.kafka.topic";
-  public static final String METRICS_CUSTOM_BUILDERS = "metrics.reporting.custom.builders";
-  public static final String METRICS_REPORT_INTERVAL_KEY = "metrics.report.interval";
+  public static final String METRICS_REPORTING_KAFKA_USE_SCHEMA_REGISTRY = METRICS_CONFIGURATIONS_PREFIX +
+      "reporting.kafka.avro.use.schema.registry";
+  public static final String DEFAULT_METRICS_REPORTING_KAFKA_USE_SCHEMA_REGISTRY = Boolean.toString(false);
+  public static final String METRICS_KAFKA_BROKERS = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.brokers";
+  public static final String METRICS_KAFKA_TOPIC = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.topic";
+  public static final String METRICS_CUSTOM_BUILDERS = METRICS_CONFIGURATIONS_PREFIX + "reporting.custom.builders";
+  public static final String METRICS_REPORT_INTERVAL_KEY = METRICS_CONFIGURATIONS_PREFIX + "report.interval";
   public static final String DEFAULT_METRICS_REPORT_INTERVAL = "30000";
 
   /**
--- a/gobblin-api/src/main/java/gobblin/configuration/State.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/State.java
@@ -39,7 +39,7 @@ public class State implements Writable {
   /**
    * Return a copy of the underlying {@link java.util.Properties} object.
    *
-   * @return Underlying {@link java.util.Properties} object.
+   * @return A copy of the underlying {@link java.util.Properties} object.
    */
   public Properties getProperties() {
     Properties props = new Properties();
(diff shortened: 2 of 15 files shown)


ibuenros:
@zliu41 @liyinan926 can you review?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -367,21 +367,26 @@
   /**
    * Common metrics configuration properties.
    */
-  public static final String METRICS_LOG_DIR_KEY = "metrics.log.dir";
-  public static final String METRICS_ENABLED_KEY = "metrics.enabled";
+  public static final String METRICS_CONFIGURATIONS_PREFIX = "metrics.";
+  public static final String METRICS_LOG_DIR_KEY = METRICS_CONFIGURATIONS_PREFIX + "log.dir";
+  public static final String METRICS_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX + "enabled";
+  public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
```
zliu41: Is it possible to add default values for these three properties (line 371-373)? Makes it easier to use.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -171,6 +172,14 @@ public void setHighWaterMark(long value) {
   }
 
   @Override
+  public Properties getProperties() {
+    Properties props = new Properties();
+    props.putAll(this.workunit.getProperties());
+    props.putAll(super.getProperties());
```
liyinan926: I suggest changing to `this.getProperties`. This make it clearer that it is adding properties belonging to this `WorkUnitState` instance.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -372,8 +397,27 @@ private void buidlKafkaMetricReporter(Properties properties) {
       formatEnum = KafkaReportingFormats.JSON;
     }
 
+    Optional<KafkaAvroSchemaRegistry> registry = Optional.absent();
+    try {
+      if ( formatEnum == KafkaReportingFormats.AVRO &&
```
liyinan926: Extra empty space after `(`.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -372,8 +397,27 @@ private void buidlKafkaMetricReporter(Properties properties) {
...
+    KafkaReporter.Builder<?> builder = formatEnum.reporterBuilder(this.metricContext);
+
+    if(builder instanceof KafkaAvroReporter.Builder<?> && registry.isPresent()) {
+      LOGGER.info("Using schema registry for Kafka avro metrics reporter.");
+      builder = ((KafkaAvroReporter.Builder<?>) builder).withSchemaRegistry(registry.get());
+    }
+
     this.scheduledReporters.add(
-        this.closer.register(formatEnum.reporterBuilder(this.metricContext).build(brokers, topic)));
+        this.closer.register(builder.build(brokers, topic)));
```
liyinan926: Extra empty line here.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -314,8 +333,14 @@ private void buildFileMetricReporter(Properties properties) {
         return;
       }
 
+      // Add a suffix to file name if specified in properties.
+      String metricsFileSuffix = properties.getProperty(ConfigurationKeys.METRICS_FILE_SUFFIX, "");
```
liyinan926: Since you check `ifNullOrEmpty` on `metricsFileSuffix`, there's no need to provide a default of empty string here.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
ibuenros: Metrics enabled already defaults to false, and metrics.reporting.file.suffix defaults to empty, although it is not in this file (I will add it here). I don't think giving a default to metrics.log.dir is a good idea, we could be messing up with the user's file system without them realizing it. On the other hand, metrics.reporting.file.enabled should probably be set to false.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
ibuenros: this.getProperties is an infinite recursion, and this.properties is private. We can either make this.properties protected, or leave it as it is right now. I have no preference.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -575,14 +576,27 @@ protected void cleanup(Context context) throws IOException, InterruptedException
...
+    private void runWorkUnits(List<WorkUnit> workUnits, Context context) throws IOException, InterruptedException {
       if (workUnits.isEmpty()) {
         LOG.warn("No work units to run");
         return;
       }
 
       String jobId = workUnits.get(0).getProp(ConfigurationKeys.JOB_ID_KEY);
+
+      // Setup and start metrics reporting
+      Properties metricReportingProperties = new WorkUnitState(workUnits.get(0)).getProperties();
```
liyinan926: I think you can just use `workUnits.get(0).getProperties()`.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -575,14 +576,27 @@ protected void cleanup(Context context) throws IOException, InterruptedException
...
       JobMetrics jobMetrics = JobMetrics.get(null, jobId);
+      String metricFileSuffix = metricReportingProperties.getProperty(ConfigurationKeys.METRICS_FILE_SUFFIX, "");
```
liyinan926: Ditto. No need to provide a default of empty string here.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/SchemaNotFoundException.java
```diff
@@ -9,7 +9,7 @@
  * CONDITIONS OF ANY KIND, either express or implied.
  */
 
-package gobblin.source.extractor.extract.kafka;
+package gobblin.metrics.kafka;
 
 @SuppressWarnings("serial")
 public class SchemaNotFoundException extends Exception {
```
liyinan926: Why this move?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
```diff
@@ -52,31 +51,34 @@
...
 
   /**
-   * @param state state should contain property "kafka.schema.registry.url", and optionally
+   * @param properties properties should contain property "kafka.schema.registry.url", and optionally
    * "kafka.schema.registry.max.cache.size" (default = 1000) and
    * "kafka.schema.registry.cache.expire.after.write.min" (default = 10).
    */
-  public KafkaAvroSchemaRegistry(State state) {
-    Preconditions.checkArgument(state.contains(KAFKA_SCHEMA_REGISTRY_URL), "Schema registry URL not provided.");
+  public KafkaAvroSchemaRegistry(Properties properties) {
```
liyinan926: What's the motivation to change from `State` to `Properties`?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ibuenros: I tried, but it was not getting all the properties. I don't quite understand why (I spend a long time trying to figure it out).

gobblin-metrics/src/main/java/gobblin/metrics/kafka/SchemaNotFoundException.java
ibuenros: It's explained in the description of the PR.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
ibuenros: State is not available in gobblin-metrics

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
liyinan926: Yeah, you are right. Didn't realize that.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
```diff
@@ -44,6 +42,7 @@
  */
 public class KafkaAvroSchemaRegistry {
 
+  public static final int SCHEMA_ID_LENGTH_BYTE = 16;
```
zliu41: put this together with MAGIC_BYTE since they are both public.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -575,14 +576,27 @@ protected void cleanup(Context context) throws IOException, InterruptedException
...
       if (workUnits.isEmpty()) {
         LOG.warn("No work units to run");
         return;
       }
 
       String jobId = workUnits.get(0).getProp(ConfigurationKeys.JOB_ID_KEY);
+
+      // Setup and start metrics reporting
+      Properties metricReportingProperties = new WorkUnitState(workUnits.get(0)).getProperties();
       JobMetrics jobMetrics = JobMetrics.get(null, jobId);
```
liyinan926: What if you use `GobblinMetrics.get(context.getTaskAttemptID().getTaskID().toString())`? This way, you don't need to append the suffix.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -228,6 +234,19 @@ public Timer getTimer(String prefix, String... suffixes) {
   /**
    * Start metric reporting.
    *
+   * @param configuration configuration properties
+   */
+  public void startMetricReporting(Configuration configuration) {
```
liyinan926: It seems that this method is not used.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
liyinan926: This is weird.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ibuenros: This will not generate job level tags correctly. Currently the only thing we would be missing is the job id, but if in the future we add more stuff to JobMetrics, it will be lost in MR executions of Gobblin.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroReporter.java
```diff
@@ -103,4 +119,24 @@ public KafkaAvroReporter build(String brokers, String topic) {
   protected Encoder getEncoder(OutputStream out) {
     return EncoderFactory.get().binaryEncoder(out, null);
   }
+
+  @Override
+  protected void writeSchemaVersioningInformation(DataOutputStream outputStream)
```
zliu41: In addition to doing what the javadoc of the super method says, this method also registers a schema to the registry. It's better to either explain it in the javadoc, or register the schema in another method.

gobblin-api/src/main/java/gobblin/configuration/State.java
```diff
@@ -37,6 +37,17 @@
   private final Properties properties = new Properties();
 
   /**
+   * Return a copy of the underlying {@link java.util.Properties} object.
+   *
+   * @return Underlying {@link java.util.Properties} object.
```
liyinan926: Should be `@return a copy of the underlying ...`.

liyinan926:
Also a typo `buidlKafkaMetricReporter`.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ibuenros: Fixed

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
ibuenros: It was used at some point when I was working on these changes, but not anymore. I figure I'd leave it because it might be useful in the future, although I don't mind removing it.

liyinan926: Then just leave it there.

ibuenros:
@zliu41 @liyinan926 addressed comments.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
```diff
@@ -52,31 +50,35 @@
   private static final String GET_RESOURCE_BY_TYPE = "/latest_with_type=";
   private static final String SCHEMA_ID_HEADER_NAME = "Location";
   private static final String SCHEMA_ID_HEADER_PREFIX = "/id=";
-  private static final String KAFKA_SCHEMA_REGISTRY_URL = "kafka.schema.registry.url";
+  public static final String KAFKA_SCHEMA_REGISTRY_URL = "kafka.schema.registry.url";
```
zliu41: Please also put this with the other public variables.

zliu41:
LGTM other than the above.

PR #157, opened 2015-06-05
Increase flexibility on which offset to start for kafka adapter, when previous offset not found or out of range

zliu41:
Before: if previous offset not found, it always starts at the earliest offset; if previous offset is out of range, either start at earliest offset or skip partition.

Now:
- If previous offset not found, may start at earliest or latest, or skip. Controlled by "bootstrap.with.offset".
- If previous offset out of range, may start at earliest or latest, or skip, or start at the nearest offset (if it's closer to earliest, start at earliest, otherwise start at latest). Controlled by "reset.on.offset.out.of.range".

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -56,9 +56,14 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
 
   public static final String TOPIC_BLACKLIST = "topic.blacklist";
   public static final String TOPIC_WHITELIST = "topic.whitelist";
+  public static final String LATEST_OFFSET = "latest";
+  public static final String EARLIEST_OFFSET = "earliest";
+  public static final String NEAREST_OFFSET = "nearest";
+  public static final String BOOTSTRAP_WITH_OFFSET = "bootstrap.with.offset";
+  public static final String DEFAULT_BOOTSTRAP_WITH_OFFSET = LATEST_OFFSET;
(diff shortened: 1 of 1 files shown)


zliu41:
@ibuenros please take a look. Thanks.

zliu41:
Updated, since `switch(string)` not supported in Java 6. Changed to if-else.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -377,31 +383,43 @@ private WorkUnit getWorkUnitForTopicPartition(KafkaWrapper kafkaWrapper, KafkaPa
...
 
     } catch (PreviousOffsetNotFoundException e) {
-      LOG.warn(String
-          .format(
-              "Previous offset for topic %s, partition %s does not exist. This partition will start from the earliest offset: %d",
-              partition.getTopicName(), partition.getId(), offsets.getEarliestOffset()));
-      offsets.startAtEarliestOffset();
+      String offsetNotFoundMsg = String.format("Previous offset for partition %s does not exist. ", partition);
+      String offsetOption = state.getProp(BOOTSTRAP_WITH_OFFSET, DEFAULT_BOOTSTRAP_WITH_OFFSET).toLowerCase();
+      if (offsetOption.equals(LATEST_OFFSET)) {
```
ibuenros: should this be equalsIgnoreCase? It makes it more user friendly.

ibuenros:
Just one comment.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
zliu41: Previous line has `.toLowerCase()`.

ibuenros:
LGTM

PR #158, opened 2015-06-05
Change the place that checks whether writer output dir exists

zliu41:
Changed `if (this.fss.get(branchId).exists(writerOutputDir))` from line 135 to line 111. Originally `addWriterOutputToExistingDir` (line 124) may throw FileNotFoundException if the workunit produced no data.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -11,7 +11,6 @@
 
 package gobblin.publisher;
 
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
@@ -71,13 +70,10 @@ public class BaseDataPublisher extends DataPublisher {
     this.numBranches = this.getState().getPropAsInt(ConfigurationKeys.FORK_BRANCHES_KEY, 1);
 
     // Get a FileSystem instance for each branch
-    URI uri;
     for (int i = 0; i < this.numBranches; i++) {
-      uri =
-          URI.create(this.getState()
-              .getProp(
-                  ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI,
-                      this.numBranches, i), ConfigurationKeys.LOCAL_FS_URI));
+      URI uri = URI.create(this.getState().getProp(
+          ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, this.numBranches, i),
+          ConfigurationKeys.LOCAL_FS_URI));
       this.fss.add(FileSystem.get(uri, conf));
     }
   }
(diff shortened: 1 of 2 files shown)


zliu41:
@liyinan926 please take a look. Thanks.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -110,29 +108,27 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
-                  ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISHER_REPLACE_FINAL_DIR,
-                      this.numBranches, branchId));
-
-          // If the final output directory is not configured to be replaced, then append the new data to the existing
-          // output folder
-          if (!replaceFinalOutputDir) {
-            addWriterOutputToExistingDir(writerOutputDir, publisherOutputDir, workUnitState, branchId);
-            writerOutputPathsMoved.add(writerOutputDir);
-            continue;
+        if (this.fss.get(branchId).exists(writerOutputDir)) {
```
liyinan926: Or you can do 
```
if (!this.fss.get(branchId).exists(writerOutputDir)) {
  LOG.warn("WorkUnit " + workUnitState.getId() + " produced no data");
  return;
}
```
right after `Path writerOutputDir = WriterUtils.getWriterOutputDir(workUnitState, this.numBranches, branchId);
`.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -73,11 +73,9 @@ public void initialize() throws IOException {
     // Get a FileSystem instance for each branch
     URI uri;
```
liyinan926: Seems you can also move this into the for loop.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
zliu41: But then `workUnitState.setWorkingState(WorkUnitState.WorkingState.COMMITTED)` will be skipped.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -153,18 +149,19 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
 
   protected void addWriterOutputToExistingDir(Path writerOutputDir, Path publisherOutputDir,
       WorkUnitState workUnitState, int branchId) throws FileNotFoundException, IOException {
```
liyinan926: You can remove `FileNotFoundException` here.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -110,29 +108,27 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
-          // output folder
-          if (!replaceFinalOutputDir) {
-            addWriterOutputToExistingDir(writerOutputDir, publisherOutputDir, workUnitState, branchId);
-            writerOutputPathsMoved.add(writerOutputDir);
-            continue;
+        if (this.fss.get(branchId).exists(writerOutputDir)) {
+          if (this.fss.get(branchId).exists(publisherOutputDir)) {
+
+            // The final output directory already exists, check if the job is configured to replace it
+            boolean replaceFinalOutputDir = this.getState().getPropAsBoolean(ForkOperatorUtils.getPropertyNameForBranch(
```
liyinan926: This can be moved out to the constructor and stored in a final member variable.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
liyinan926: You can add it in the if block before returning.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
zliu41: There's one `replaceFinalOutputDir` for each branch.

zliu41:
Updated.

PR #160, opened 2015-06-05
Refactored the code around JobMetrics in MRJobLauncher

liyinan926:
This pull request has the following changes:
1. `JobMetrics` is now wrapped in an `Optional`, just like in `JobContext`.
2. The setup happens and `JobMetrics.startMetricReporting` is called in the `setup` method.
3. `HobMetrics.stopMetricReporting` is called in the `cleanup` method.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -498,6 +498,7 @@ public class MRJobLauncher extends AbstractJobLauncher {
     private TaskExecutor taskExecutor;
     private TaskStateTracker taskStateTracker;
     private ServiceManager serviceManager;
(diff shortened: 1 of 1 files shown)


liyinan926:
@ibuenros can you review this? I just realized that `JobMetrics.stopMetricReporting` should be called somewhere soon after you merged in #156. Then I realized `setup` and `cleanup` are better places to start and stop metric reporting respectively. Also the `Configuration` object returned by `context.getConfiguration()` has all job configuration properties in it. So now the method `JobMetrics.startMetricReporting(Configuration)` gets used.

liyinan926:
Hey @ibuenros can you review this?

ibuenros:
@liyinan926 It looks good, but I would want to test that the expected properties actually get propagated to the mappers. It will take me some time to do that.

liyinan926:
Sure. Let me know once you are done with the testing.

ibuenros:
LGTM

PR #161, opened 2015-06-08
Support for running test harness on Azkaban

suvodeep-pyne:
Added a class to enable running the test harness on Azkaban. The class reads a properties file hardcoded to testharness.gobblin.azkaban.properties in the working directory. This file may be used to run different configurations of tests using the Test Harness.

### DIFF
--- a/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
+++ b/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
@@ -18,6 +18,7 @@ import org.apache.log4j.Logger;
 
 import azkaban.jobExecutor.AbstractJob;
 
+import gobblin.configuration.ConfigurationKeys;
 import gobblin.test.setup.config.TestHarnessLauncher;
 
 /**
@@ -40,6 +41,7 @@ public class AzkabanIntegrationTestLauncher extends AbstractJob {
   @Override
   public void run() throws Exception {
     // Get the test harness launcher instance
+    launcher = createTestHarnessInstance();
 
     // Read the properties file
     launcher.configure(properties);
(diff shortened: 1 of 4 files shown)


gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.File;
+
+import org.apache.log4j.Logger;
+
+import azkaban.jobExecutor.AbstractJob;
+
+import gobblin.test.setup.config.TestHarnessLauncher;
+
+/**
+ * This class launches the TestHarness framework in Gobblin and executes tests using TestNG
```
sahilTakiar: It should say Azkaban instead of TestNg right?

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Created by spyne on 6/8/15.
+ */
+public class AzkabanIntegrationTestLauncher extends AbstractJob {
+  private static final String CONFIG_PATH = "testharness.gobblin.azkaban.properties";
+  private static final Logger LOG = Logger.getLogger(AzkabanIntegrationTestLauncher.class);
+  private static final String id = "gobblin.idpc.integration.test";
```
sahilTakiar: Can we remove the idpc specific configs? As discussed you should be able to use the `AzkabanIntegrationTestLauncher(String jobId, Properties props)` to get all the relevant configuration properties. The Azkaban documentation should have that feature listed somewhere, if not I'm sure we can ask azkaban-dev.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public AzkabanIntegrationTestLauncher(String id, Properties properties) {
+    super(id, LOG);
+    this.properties = properties;
+  }
+
+  @Override
+  public void run() throws Exception {
+    // Get the test harness launcher instance
+
```
sahilTakiar: I think your missing the initialization of the launcher object.

sahilTakiar:
Minor comment, can you also do a git pull and merge. There were some updates to gobblin-test-harness to fix a bunch of the license headers.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java
suvodeep-pyne: Yes. Thats what I was pointing out yesterday that since the test harness implementation is outside Gobblin, we need to create an instance and send it across

sahilTakiar: Ah I see, in that case lets make the class name a configuration properties. Something like `test.harness.launcher.class=ClassThatImplementsTestHarnessLauncher`

suvodeep-pyne: updated the pull request.

sahilTakiar:
LGTM!

PR #163, opened 2015-06-09
Fixing NPE in countersToMetrics

sahilTakiar:
`countersToMetrics` throws a NPE if `job.getCounters()` returns `null`. This PR fixes the NPE.


gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -205,7 +205,7 @@ protected void runWorkUnits(List<WorkUnit> workUnits) throws Exception {
 
       // Create a metrics set for this job run from the Hadoop counters.
       // The metrics set is to be persisted to the metrics store later.
-      countersToMetrics(this.job.getCounters(),
+      countersToMetrics(Optional.of(this.job.getCounters()),
```
liyinan926: `Optional.of` will throw NPE if the input is `null`. Please use `fromNullable`.

liyinan926:
LGTM.

sahilTakiar:
Merged

PR #164, opened 2015-06-09
Fixed around reporting metrics before destroying metric contexts and …

ibuenros:
…reporters.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
+++ b/gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
@@ -12,9 +12,8 @@
 
 package gobblin.metrics;
 
-import java.util.concurrent.ConcurrentMap;
-
-import com.google.common.collect.MapMaker;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 
 
 /**
@@ -32,18 +31,18 @@ public class GobblinMetricsRegistry {
     return GLOBAL_INSTANCE;
   }
 
-  private final ConcurrentMap<String, GobblinMetrics> metricsMap = new MapMaker().weakValues().makeMap();
+  private final Cache<String, GobblinMetrics> metricsMap = CacheBuilder.newBuilder().softValues().build();
 
   public GobblinMetrics putIfAbsent(String id, GobblinMetrics gobblinMetrics) {
-    return this.metricsMap.putIfAbsent(id, gobblinMetrics);
+    return this.metricsMap.asMap().putIfAbsent(id, gobblinMetrics);
   }
 
   public boolean containsKey(String id) {
-    return this.metricsMap.containsKey(id);
+    return this.metricsMap.asMap().containsKey(id);
   }
 
   public GobblinMetrics get(String id) {
-    return this.metricsMap.get(id);
+    return this.metricsMap.getIfPresent(id);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ibuenros:
@liyinan926 can you review? Thanks.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
```diff
@@ -54,7 +53,7 @@ public GobblinMetrics get(String id) {
    *         instance for the given job is not found
    */
   public GobblinMetrics remove(String id) {
-    return this.metricsMap.remove(id);
+    return this.metricsMap.asMap().remove(id);
```
liyinan926: You can do `this.metricMap.invalidate(id);`.

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -138,10 +138,11 @@ public String getName() {
    *
    * @param childContext the child {@link MetricContext} to add
    */
-  public void addChildContext(String childContextName, MetricContext childContext) {
-    if (this.children.putIfAbsent(childContextName, childContext) != null) {
```
liyinan926: You can do `this.children.asMap().putIfAbsent()`, then you don't need to use `synchronized`.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -594,7 +594,9 @@ protected void cleanup(Context context) throws IOException, InterruptedException
         // Ignored
       } finally {
         if (this.jobMetrics.isPresent()) {
+          this.jobMetrics.get().triggerMetricReporting();
```
liyinan926: Better to switch the order between the two so scheduled reporting won't happen while manual reporting is in progress.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -594,7 +594,9 @@ protected void cleanup(Context context) throws IOException, InterruptedException
...
           this.jobMetrics.get().stopMetricReporting();
+          JobMetrics.remove(this.jobMetrics.get().getName());
```
liyinan926: Also suggesting put `remove` into a `finally` clause following a `catch` around the above two statements.

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -149,7 +150,7 @@ public void addChildContext(String childContextName, MetricContext childContext)
    * @return Map of child contexts.
```
liyinan926: I suggest renaming this to `getChildContextsAsMap`, and updating the Javadoc to `Get a view of the child {@link MetricContext}s as a {@link ImmutableMap}.`. Please also update `@return an {@link ImmutableMap} of child {@link MetricContext}s keyed on their names`.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
ibuenros: invalidate does not return the object upon removal, so we would have to change API of the method, or put additional logic here to first get the object to return.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ibuenros: Stop metric reporting kills the Kafka connections, etc. so the order cannot be switched. I don't think there is an issue with two reports happening at the same time, metrics will be duplicated, but there was never a guarantee otherwise.

ibuenros:
@liyinan926 addressed comments.

PR #166, opened 2015-06-10
Kafka metrics extended

ibuenros:
Adds task level metrics for Kafka sources.
Adds UUID to each metric context.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/State.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/State.java
@@ -24,9 +24,6 @@ import org.apache.hadoop.io.Writable;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableSortedSet;
-import com.google.common.collect.Lists;
-import com.google.common.escape.Escaper;
-import com.google.common.escape.Escapers;
 
 
 /**
@@ -103,10 +100,17 @@ public class State implements Writable {
 
   /**
    * Appends the input value to a list property that can be retrieved with {@link #getPropAsList}.
+   *
+   * <p>
+   *   List properties are internally stored as comma separated strings. Adding a value that contains commas (for
+   *   example "a,b,c") will essentially add multiple values to the property ("a", "b", and "c"). This is
+   *   similar to the way that {@link org.apache.hadoop.conf.Configuration} works.
+   * </p>
+   *
    * @param key property key
-   * @param value property value (should not include commas, which are used as separators)
+   * @param value property value (if it includes commas, it will be split by the commas).
    */
-  public void appendToListProp(String key, String value) {
+  public synchronized void appendToListProp(String key, String value) {
     if(contains(key)) {
       setProp(key, Joiner.on(",").join(getProp(key), value));
     } else {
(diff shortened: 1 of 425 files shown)


ibuenros:
@zliu41 @liyinan926 can you review?

ibuenros:
@liyinan926 addressed comments except for the one in State.java (still waiting for a reply).

ibuenros:
@liyinan926 addressed new comments. @zliu41 do you have some time to look at the PR? it's fairly short.

gobblin-api/src/main/java/gobblin/configuration/State.java
```diff
@@ -98,6 +102,19 @@ public void setProp(String key, Object value) {
   }
 
   /**
+   * Appends the input value to a list property that can be retrieved with {@link #getPropAsList}.
+   * @param key property key
+   * @param value property value (should not include commas, which are used as separators)
```
zliu41: value can include commas, if the user wants to pass a set of comma-separated values.
Also is it better to make the separator a parameter?

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -54,6 +54,8 @@
  */
 public class GobblinMetrics {
 
+  public static final String METRIC_STATE_CUSTOM_TAGS = "metrics.state.custom.tags";
```
zliu41: name and value are inconsistent.

zliu41:
only 2 comments.

gobblin-api/src/main/java/gobblin/configuration/State.java
ibuenros: The separator can't really be changed because getPropAsList is an existing API that relies on separator being always ",". I did think about the fact that values could have commas. Ideally we would want to escape and unescape the strings, but it is not particularly easy.
Two options: either spend the time doing the escaping, or just specify in the doc that if values have commas they will be parsed incorrectly.

zliu41: If you currently have `kafka.brokers=broker1,broker2`, you can certainly pass `broker3,broker4,broker5` as value to this method, so you know have 5 brokers, right?

It is true that if an individual broker's value has comma, then it will be problematic. However, the fact that comma is used as separator not only affects this method but also for other methods such as `setProp` and `getPropAsList`. So this should be explained in the Javadoc of this class, rather than here.

liyinan926: One thing we can do is to add a new method `getPropAsList(String separator, String key)`, which allows users to specify a separator, and simply call `getPropAsList(",", key)` in `getPropAsList(key)` as a default and make clear in the Javadoc for this default that it is expecting comma as separator.

ibuenros: Looking at Hadoop Configuration, which supports property lists as comma separated strings, it suffers from the same issues as the current implementation of State (see https://hadoop.apache.org/docs/current/api/org/apache/hadoop/conf/Configuration.html#setStrings(java.lang.String, java.lang.String...) ). 
Hadoop docs do not go extensively into the comma issue (but I looked at the source code). I think it is generally understood that values containing commas always have the risk of being interpreted as lists, and this is doing the same as a well known project as Hadoop is doing.
I don't think there's a benefit in adding a more complex API, or explaining to much in the doc (although I will add some more explanation). Making the separator configurable just adds more configurations that don't contribute much.

ibuenros:
@zliu41 addressed comments. The issue around the comma separated list might still be up for debate.

zliu41:
LGTM

PR #167, opened 2015-06-10
Fix bug in AvroUtils, add default schema to KafkaAvroExtractor, add logdir option to gobblin-standalone.sh

zliu41:
- Added an option to specify the log dir in gobblin-standalone.sh, which is needed for the Kafka adapter test harness.
- In KafkaAvroExtractor, if the topic schema of the workunit is not found, give it a default schema. Because currently even if a workunit extracts 0 record, a data writer is still created, and we need an Avro schema in order to create an Avro data writer.
- In `AvroUtils.getFieldSchemaHelper`, added a `if` block to avoid NPE.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaAvroExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaAvroExtractor.java
@@ -17,19 +17,22 @@ import java.util.Arrays;
 import kafka.message.MessageAndOffset;
 
 import org.apache.avro.Schema;
+import org.apache.avro.SchemaBuilder;
 import org.apache.avro.generic.GenericData.Record;
 import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.generic.GenericRecord;
-import org.apache.avro.io.DatumReader;
 import org.apache.avro.io.Decoder;
 import org.apache.avro.io.DecoderFactory;
 import org.apache.commons.codec.binary.Hex;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Optional;
+
 import gobblin.configuration.WorkUnitState;
 import gobblin.metrics.kafka.KafkaAvroSchemaRegistry;
 import gobblin.metrics.kafka.SchemaNotFoundException;
+import gobblin.source.extractor.DataRecordException;
 import gobblin.source.extractor.Extractor;
 import gobblin.util.AvroUtils;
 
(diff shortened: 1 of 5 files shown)


zliu41:
@ibuenros please take a look. Thanks.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaAvroExtractor.java
```diff
@@ -42,8 +46,11 @@
 public class KafkaAvroExtractor extends KafkaExtractor<Schema, GenericRecord> {
 
   private static final Logger LOG = LoggerFactory.getLogger(KafkaAvroExtractor.class);
+  private static final Schema DEFAULT_SCHEMA = SchemaBuilder.record("PageViewEvent").fields().name("header")
```
ibuenros: Could this be an even simpler schema, like SchemaBuilder.record("DefaultSchema").endRecord() ?
Also, why name it "PageViewEvent"?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaAvroExtractor.java
```diff
@@ -54,20 +61,33 @@
...
    * from the schema registry.
    */
-  public KafkaAvroExtractor(WorkUnitState state) throws SchemaNotFoundException {
+  public KafkaAvroExtractor(WorkUnitState state) {
     super(state);
     this.schemaRegistry = new KafkaAvroSchemaRegistry(state.getProperties());
-    this.schema = getLatestSchemaByTopic();
-    this.reader = new GenericDatumReader<Record>(this.schema);
+    this.schema = Optional.fromNullable(getLatestSchemaByTopic());
+    this.reader = this.schema.isPresent() ? new GenericDatumReader<Record>(this.schema.get()) : null;
```
ibuenros: Should this.reader also be optional?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaAvroExtractor.java
zliu41: The schema needs to have a `header.time` field, since this is needed to create `AvroHdfsTimePartitionedWriter`.

I can change the name to "DefaultSchema"

zliu41:
Updated.

liyinan926:
@zliu41 this PR has been failing build on both Java 7 & 8: https://travis-ci.org/linkedin/gobblin/builds/66302548.

zliu41:
Updated.

zliu41:
@ibuenros do you have more comments?

ibuenros:
LGTM

PR #170, opened 2015-06-12
Make Kafka adapter blacklist and whitelist work for regex

zliu41:


gobblin-utility/src/main/java/gobblin/util/DatasetFilterUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      return !stringInPatterns(topic, blacklist);
+    }
+  }
+
+  /**
+   * Determines whether a string matches one of the regex patterns.
+   */
+  public static boolean stringInPatterns(String s, List<Pattern> patterns) {
+    for (Pattern pattern : patterns) {
```
ydailinkedin: When blacklist is null, here needs check for null `patterns`.

zliu41: blacklist is obtained from `State.getPropAsList` which never returns null.

ydailinkedin: I see. LGTM.

PR #171, opened 2015-06-13
Create empty workunit when skipping a Kafka partition

zliu41:
Before: if a partition is skipped (e.g., unable to get offsets from Kafka or other reasons), no workunit will be created. But this means the previous pulled offset will be lost.

Now: create an empty workunit, so that the previous offset can be persisted in the state store and is available for the next run.


zliu41:
@liyinan926 please take a look. Thanks.

liyinan926:
LGTM.

PR #172, opened 2015-06-14
Improvement and bug fix for MR-based compaction

zliu41:
- (Improvement) Before: the compactor checks whether the outputdir exists. If exists, it checks the property `compaction.force.reprocess`. If true, it will reprocess inputdir. If false, will skip inputdir.

This is problematic if inputdir and outputdir are the same. Then the fact that outputdir exists doesn't indicate whether it has been compacted or not.

Now: after compaction is done, a file named `_COMPACTION_COMPLETE` will be created in outputdir. Property `compaction.force.reprocess` is no longer used. Now if outputdir exists and has `_COMPACTION_COMPLETE`, will skip inputdir. Otherwise, will process inputdir.

- (Bug) Before: the outputdir is deleted immediately when it decides to process inputdir.

Again, this doesn't work if inputdir and outputdir are the same. Now: delete outputdir after the MR job is done.

We don't have a use case where inputdir = outputdir. But I think it's good to support it.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
@@ -22,11 +22,11 @@ import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
+
 
 /**
  * This class creates the following properties for a single MapReduce job for compaction:
@@ -105,12 +105,7 @@ public class MRCompactorJobPropCreator {
       LOG.warn("Input folder " + this.topicInputDir + " does not exist. Skipping topic " + topic);
       return emptyJobProps;
     }
-    Optional<State> jobProps = createJobProps(this.topicInputDir, this.topicOutputDir, this.topicTmpDir);
-    if (jobProps.isPresent()) {
-      return Collections.singletonList(jobProps.get());
-    } else {
-      return emptyJobProps;
-    }
+    return Collections.singletonList(createJobProps(this.topicInputDir, this.topicOutputDir, this.topicTmpDir));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


zliu41:
@ibuenros please review, thanks.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -168,27 +167,12 @@ private void submit(Job job) throws ClassNotFoundException, IOException, Interru
 
   private void moveTmpPathToOutputPath() throws IOException {
     LOG.info(String.format("Moving %s to %s", this.tmpPath, this.outputPath));
+    this.fs.delete(this.outputPath, true);
```
ibuenros: is there some protection against the user giving wrong parameters and accidentally deleting something they shouldn't?
I suggest having the old logic for most cases, and just having the different logic if inputPath == outputPath.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
```diff
@@ -105,12 +105,7 @@ protected MRCompactorJobPropCreator(Builder<?> builder) {
       LOG.warn("Input folder " + this.topicInputDir + " does not exist. Skipping topic " + topic);
       return emptyJobProps;
     }
-    Optional<State> jobProps = createJobProps(this.topicInputDir, this.topicOutputDir, this.topicTmpDir);
-    if (jobProps.isPresent()) {
-      return Collections.singletonList(jobProps.get());
-    } else {
-      return emptyJobProps;
-    }
+    return Collections.singletonList(createJobProps(this.topicInputDir, this.topicOutputDir, this.topicTmpDir));
```
ibuenros: What is the reasoning behind this change?

ibuenros:
Why are we removing `compaction.force.reprocess`? It seems to me it is a reasonable use case that users want to re-compact the input every run, even if the output was already compacted.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
zliu41: The old logic deletes outputPath even earlier (before MR job is run), which is why the old logic didn't work if inputPath = outputPath.
The outputPath has to be deleted at some stage so that the compacted data can be moved to the outputPath.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
zliu41: Before: `createJobProps` was responsible for determining whether a folder should be compacted. If so, return a job prop file, if not, return `Optional.absent()`. Now this responsibility is moved to `shouldProcessFolder`. `createJobProps` is only called if the folder should be compacted.

zliu41:
One should only re-compact if there's a newer file (outlier) in the input folder. Otherwise, always re-compacting is a waste. I'll add the outlier handling logic later in a separate PR.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ibuenros: But earlier output path was only deleted if the option `compaction.force.reprocess` was true, it requires the user to explicitly ask for output path to be deleted. In this case, the output is deleted without user allowing it.

zliu41:
@ibuenros updated. Added property "compaction.overwrite.output.dir", default false. If inputdir = outputdir, this must be set to true.

ibuenros:
LGTM other than one comment.

zliu41:
FIxed.

PR #173, opened 2015-06-15
Encrypt passwords in config files (ETL-3148)

zliu41:
All password in config files should now be in the format of `ENC(.*)`.
Also there should be a file containing the encryption key, in order to decrypt the passwords in the config files.


zliu41:
@chavdar please review. Thanks.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -707,7 +707,8 @@ private void parseInputQuery(String query) {
   protected JdbcProvider createJdbcSource() {
     String driver = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_DRIVER);
     String userName = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_USERNAME);
-    String password = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PASSWORD);
+    String password =
+        PasswordUtils.decryptPassword(this.workUnit, this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PASSWORD));
```
chavdar: We still need to support the case where the password are un-encrypted fo backwards compatibility.  We probably need a readPassword() method in PasswordUtils that will check if the password matches ENC(.*) than it should be decrypted, otherwise return the plain text.

gobblin-utility/src/main/java/gobblin/util/PasswordUtils.java
```diff
@@ -0,0 +1,177 @@
...
+import com.google.common.io.Closer;
+import com.google.common.io.LineReader;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+
+/**
+ * A utility class for encrypting and decrypting passwords.
+ */
+public class PasswordUtils {
```
chavdar: The ENCRYPTOR object does not seem to be used.

I suggest considering a different design where encryptPassword/decryptPassword/readPassword methods are instance-level methods and PasswordUtils is an object with a encryptor/decryptor instance that gets initialized in the constructor . There can be different getInstance(...) methods which can read the master password from the specified location. Internally, getInstance() can use an instance cache keyed on the master password in case instantiating a new encryptor/decryptor is expensive.

gobblin-utility/src/test/java/gobblin/util/PasswordUtilsTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import com.google.common.io.Files;
+
+
```
chavdar: Can you also add a unit test which gets the password from a state/workunit since this seems to be the path used by most extractors.

gobblin-utility/src/main/java/gobblin/util/PasswordUtils.java
```diff
@@ -0,0 +1,177 @@
...
+    StrongTextEncryptor ec = new StrongTextEncryptor();
+    ec.setPassword(key);
+    try {
+      return ec.decrypt(matcher.group(1));
+    } catch (Exception e) {
+      throw new RuntimeException("Cannot decrypt password " + encrypted);
+    }
+  }
+
+  private static String getKey(State state) {
```
chavdar: Consider using masterPassword instead of "key" since "key" is very ambiguous.

gobblin-utility/src/main/java/gobblin/util/PasswordUtils.java
```diff
@@ -0,0 +1,177 @@
...
+    if (state.contains(ConfigurationKeys.ENCRYPT_KEY_FS_URI)) {
+      fsUri = state.getProp(ConfigurationKeys.ENCRYPT_KEY_FS_URI);
+    } else {
+      fsUri = state.getProp(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI);
+    }
+    try {
+      return getKey(FileSystem.get(new URI(fsUri), new Configuration()),
+          new Path(state.getProp(ConfigurationKeys.ENCRYPT_KEY_LOC)));
+    } catch (Exception e) {
+      throw Throwables.propagate(e);
```
chavdar: Please create a new RuntimeException with a message that explains that the master password cannot be read and "e" is the cause. This will make logging these kind of errors a lot easier to understand.

gobblin-utility/src/main/java/gobblin/util/PasswordUtils.java
```diff
@@ -0,0 +1,177 @@
...
+  private static String decryptPassword(String key, String encrypted) {
+    Matcher matcher = PASSWORD_PATTERN.matcher(encrypted);
+    Preconditions.checkArgument(matcher.matches(),
+        String.format("Encrypted password %s doesn't match pattern %s", encrypted, PASSWORD_PATTERN));
+    StrongTextEncryptor ec = new StrongTextEncryptor();
+    ec.setPassword(key);
+    try {
+      return ec.decrypt(matcher.group(1));
+    } catch (Exception e) {
+      throw new RuntimeException("Cannot decrypt password " + encrypted);
```
chavdar: Please add "e" as a cause to the RuntimeException.

gobblin-utility/src/main/java/gobblin/util/PasswordUtils.java
chavdar: Actually, that's probably better suited for the other form of getKey().

zliu41:
@chavdar updated. Replaced `PasswordUtils` with `PasswordManager`, which caches instances, and supports both encrypted and plain password.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -476,6 +479,12 @@
   public static final String DEFAULT_JOB_HISTORY_STORE_PASSWORD = "gobblin";
 
   /**
+   * Password encryption and decryption properties.
+   */
+  public static final String ENCRYPT_KEY_FS_URI = "encrypt.key.fs.uri";
```
chavdar: nit: can't we combine the two properties into a single URI?

chavdar:
We also need a simple command line tool that can encrypt/decrypt a password given a master password.

zliu41:
@chavdar updated. Added command line tool `CLIPasswordEncryptor`. After building gobblin there will be a `gobblin-utility.tar.gz` file in `build/gobblin-utility/distributions`. Unzip it and there will be a runnable jar.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
zliu41: I need to first get a `FileSystem` instance using fs uri, then call `FileSystem.exists()`, `FileSystem.open()`, etc. So I doubt I can combine them.

chavdar: You can use Path(URI aUri) and then Path.getFileSystem() to get the file system associated with the URI.

zliu41: I see. Fixed.

PR #123, opened 2015-05-12
Feature/simple data writer

akshayn1107:
A simple data writer which writes records as byte[] to a file. This also has some refactors around some of the common elements between writers. Adds a new configuration option to add the writer's final output path to the task properties. This can be used by a publisher to read the write file for a task.

PR #175, opened 2015-06-17
Refactored the code on how sys and job config properties are handled

liyinan926:
1. Refactored the code on how system and job config properties are handled and passed around. With the changes, `AbstractJobLauncher` nows takes only a single `Properties` object that includes both system and job config properties. 
2. Added a `JobConfigurationUtils` and used it.
3. Added `WriterUtils.setFileAttributesFromState(properties, fs, outputFile);` back to `AvroHdfsDataWriter` that was deleted by accident in #123.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.util.ToolRunner;
 import com.google.common.io.Closer;
 
 import gobblin.runtime.JobException;
+import gobblin.util.JobConfigurationUtils;
 
 
 /**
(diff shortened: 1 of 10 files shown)


liyinan926:
@sahilTakiar can you review this?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
```diff
@@ -52,15 +53,11 @@ public CliMRJobLauncher(Properties sysConfig, Properties jobConfig)
...
-    final Properties jobProps = new Properties();
-    // First load system configuration properties
-    jobProps.putAll(this.sysConfig);
-    // Then load job configuration properties that might overwrite system configuration properties
-    jobProps.putAll(this.jobConfig);
+    final Properties jobProps = JobConfigurationUtils.combineSysAndJobProperties(this.sysConfig, this.jobConfig);
 
     Closer closer = Closer.create();
     try {
-      closer.register(new MRJobLauncher(this.sysConfig, jobProps, getConf())).launchJob(null);
```
sahilTakiar: Its not really relevant to this PR, but any reason why this line doesn't use `JobLauncherFactory`

gobblin-runtime/src/test/java/gobblin/runtime/JobStateStoreTest.java
```diff
@@ -111,7 +109,7 @@ public void testLaunchSecondJob() throws Exception {
   public void testLaunchThirdJob() throws Exception {
     Closer closer = Closer.create();
     try {
-      closer.register(new LocalJobLauncher(this.properties, this.jobProps)).launchJob(null);
+      closer.register(new LocalJobLauncher(this.jobConfig)).launchJob(null);
```
sahilTakiar: Same comment as above, use `JobLauncherFactory` for the above three tests

sahilTakiar:
Minor comments, otherwise LGTM

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
liyinan926: It is using a different constructor than the one used by `JobLauncherFactory`.

gobblin-runtime/src/test/java/gobblin/runtime/JobStateStoreTest.java
liyinan926: Due to the refactoring above, the single `jobConfig` already includes everything so it makes more sense to use `LocalJobLauncher` directly.

PR #177, opened 2015-06-22
Addressing some build issues with the previous commit for password en…

chavdar:
Addressing some build issues with the previous commit for password encryption:

* The commit changed the name of the jar generated by gobblin-utility . I've reverted that change.
* I've removed the Main-Class from the manifest since it is conceivable that we have other
  CLI classes in the library.
* I have added a helper bash script to run the password encryptor utility


zliu41:
LGTM.

PR #179, opened 2015-06-22
Add hadoop cluster tag to metrics

pcadabam:
ibuenros@ can you take a look at this?
The pull request is to be able to tag metrics by hadoop cluster name.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
@@ -12,7 +12,14 @@
 
 package gobblin.runtime.util;
 
+import java.net.InetAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.UnknownHostException;
 import java.util.List;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.conf.Configuration;
 
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/gobblin/instrumented/Instrumented.java
```diff
@@ -131,7 +132,9 @@ public static MetricContext getMetricContext(State state, Class<?> klazz, List<T
     if (!klazz.isAnonymousClass()) {
       generatedTags.add(new Tag<String>("class", klazz.getCanonicalName()));
     }
-
+    if (state.contains(ConfigurationKeys.HADOOP_CLUSTER_NAME_KEY)) {
+      generatedTags.add(new Tag<String>("cluster", state.getProp(ConfigurationKeys.HADOOP_CLUSTER_NAME_KEY)));
```
ibuenros: I would prefer putting this at the job level. Look at JobMetrics class.

gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
```diff
@@ -87,6 +88,9 @@ private static String name(JobState jobState) {
     List<Tag<?>> tags = Lists.newArrayList();
     tags.add(new Tag<String>("jobName", jobState.getJobName() == null ? "" : jobState.getJobName()));
     tags.add(new Tag<String>("jobId", jobState.getJobId()));
+    if (jobState.contains(ConfigurationKeys.HADOOP_CLUSTER_NAME_KEY)) {
```
chavdar: It will be great if this can be automatically inferred, instead of relying on yet another config. Say, use the host name of the Name Node.

ibuenros:
Other than Chavdar's comment, LGTM.

ibuenros:
Can you fix tests before merging.
Also, just remembered, if the jobmanager / resourcemanager is localhost, we should resolve localhost.

PR #181, opened 2015-06-24
Copyright changes

swami-veeramani:
Changed Step to have a getStepName which will help in getting the name of the step 
Also added throws exception to LaunchStep

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,7 +170,7 @@ subprojects {
       }
 
       // Required to add JDK's tool jar, which is required to run byteman tests.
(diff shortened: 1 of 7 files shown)


sahilTakiar:
Looks like the build is failing because the class https://github.com/linkedin/gobblin/blob/master/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanIntegrationTestLauncher.java is throwing some build errors related to the API changes. Can you resolve the errors.

Other than that, the changes LGTM.

PR #183, opened 2015-06-25
Add azkaban flowId tag to metrics

pcadabam:


gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -96,6 +99,13 @@ public JobContext(Properties jobProps, Logger logger) throws Exception {
         previousJobState.getPropAsInt(ConfigurationKeys.JOB_FAILURES_KEY, 0));
 
     if (GobblinMetrics.isEnabled(jobProps)) {
+      Configuration configuration = new Configuration();
+      String flowId = configuration.get("azkaban.flow.flowid");
```
liyinan926: The key may not exist, so you should check first if it exists.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -96,6 +99,13 @@ public JobContext(Properties jobProps, Logger logger) throws Exception {
...
+
+      if (StringUtils.isNotBlank(flowId)) {
```
liyinan926: Please use Guava's `Strings.isNullOrEmpty` instead.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: Please define a String constant for `azkaban.flow.flowid` in `ConfigurationKeys`.

liyinan926: Actually I think it's better to put it into a class say called `TagNames` under `gobblin-core/src/main/java/gobblin/metrics/`.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
pcadabam: Strings.isNullOrEmpty does not check white spaces.
it returns false for " ", I want it to be true.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
pcadabam: 1. Defined a constant for azkaban.flow.flowid
2. Moved tag names to gobblin-core/src/main/java/gobblin/metrics/TagNames.java

liyinan926:
LGTM.

PR #185, opened 2015-06-25
Use current system time if record timestamp is not present

zliu41:
Before: `AvroHdfsTimePartitionedWriter` requires a partition field name (e.g., `header.time`), and the Avro schema must contain this field.

Now: if field name is not provided, or if Avro schema doesn't have this field, use current system time. This is consistent with Camus. There are some tracking topics whose schemas don't have a timestamp field.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
+import com.google.common.primitives.Longs;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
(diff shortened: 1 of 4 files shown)


zliu41:
@liyinan926 please take a look.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -187,6 +177,21 @@ public void write(GenericRecord record) throws IOException {
...
 
+  /**
+   * Checks whether the given value is present and can be cast to Long.
+   */
+  private boolean isPartitionValueUsable(Optional<Object> writerPartitionColumnValue) {
+    if (!writerPartitionColumnValue.isPresent()) {
+      return false;
+    }
+    try {
+      Long.parseLong(writerPartitionColumnValue.get().toString());
```
liyinan926: You can use `Longs.tryParse`, http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/primitives/Longs.html#tryParse(java.lang.String).

zliu41:
@liyinan926 updated.

PR #186, opened 2015-06-25
change blacklist and whitelist logic

zliu41:
Before: is whitelist is non-empty, a topic is selected if it matches the whitelist, regardless of blacklist.
Now: topics that match blacklist will be filtered, even if they match whitelist.

This makes it easier to select a set of topics.


zliu41:
@liyinan926 please take a look.

liyinan926:
LGTM.

PR #187, opened 2015-06-29
Count bytes read from FileBasedExtractor

pcadabam:
Moved  gobblin-metrics/src/main/java/gobblin/metrics/Counters.java  from gobblin-tracking-purger

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
@@ -27,6 +27,8 @@ import com.google.common.io.Closer;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
+import gobblin.instrumented.Instrumented;
+import gobblin.metrics.Counters;
 import gobblin.source.extractor.DataRecordException;
 import gobblin.source.extractor.Extractor;
 import gobblin.source.workunit.WorkUnit;
@@ -48,7 +50,7 @@ public abstract class FileBasedExtractor<S, D> implements Extractor<S, D> {
 
   protected final WorkUnit workUnit;
   protected final WorkUnitState workUnitState;
-  protected final FileBasedHelper fsHelper;
+  protected final SizeAwareFileBasedHelper fsHelper;
   protected final List<String> filesToPull;
 
   protected final Closer closer = Closer.create();
@@ -60,6 +62,12 @@ public abstract class FileBasedExtractor<S, D> implements Extractor<S, D> {
   private String currentFile;
   private boolean readRecordStart;
 
+  private enum CounterNames {
+    FileBytesRead;
+  }
+
+  private Counters<CounterNames> counters = new Counters<CounterNames>();
+
   public FileBasedExtractor(WorkUnitState workUnitState, FileBasedHelper fsHelper) {
     this.workUnitState = workUnitState;
     this.workUnit = workUnitState.getWorkunit();
@@ -67,12 +75,19 @@ public abstract class FileBasedExtractor<S, D> implements Extractor<S, D> {
         Lists.newArrayList(workUnitState.getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FILES_TO_PULL, ""));
     this.statusCount = this.workUnit.getPropAsInt(ConfigurationKeys.FILEBASED_REPORT_STATUS_ON_COUNT,
         ConfigurationKeys.DEFAULT_FILEBASED_REPORT_STATUS_ON_COUNT);
-    this.fsHelper = fsHelper;
+
+    if (fsHelper instanceof SizeAwareFileBasedHelper) {
+      this.fsHelper = (SizeAwareFileBasedHelper)fsHelper;
+    } else {
+      this.fsHelper = new SizeAwareFileBasedHelperDecorator(fsHelper);
+    }
     try {
       this.fsHelper.connect();
     } catch (FileBasedHelperException e) {
       Throwables.propagate(e);
     }
+
+    counters.initialize(Instrumented.getMetricContext(workUnitState, this.getClass()), CounterNames.class, this.getClass());
   }
 
   /**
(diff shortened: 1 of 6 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -192,4 +209,14 @@ public void close() {
       LOGGER.error("Could not successfully close file system helper due to error: " + e.getMessage(), e);
     }
   }
+
+  private void incrementBytesReadCounter() {
+    try {
+      counters.inc(CounterNames.FileBytesRead, fsHelper.getFileSize(currentFile));
+    } catch (FileBasedHelperException e) {
+      LOGGER.info("Unable to get file size. Will skip increment to bytes counter " + e.getMessage());
```
ibuenros: You can do `LOGGER.info("Unable to get file size. Will skip increment to bytes counter ", e);`. It will print the entire stacktrace, but it is more informative.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/SizeAwareFileBasedHelperDecorator.java
```diff
@@ -0,0 +1,56 @@
...
+package gobblin.source.extractor.filebased;
+
+import java.io.InputStream;
+import java.util.List;
+
+/**
+ * A decorator that delegates to inner {@link FileBasedHelper}.
+ * {@link #getFileSize(String)} is not implemented by this class.
+ */
+public class SizeAwareFileBasedHelperDecorator implements SizeAwareFileBasedHelper {
```
ibuenros: this should implement decorator interface

ibuenros:
Just two comments from me.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -94,6 +109,7 @@ public D readRecord(@Deprecated D reuse) throws DataRecordException, IOException
       if (!filesToPull.isEmpty()) {
         currentFile = filesToPull.remove(0);
         currentFileItr = downloadFile(currentFile);
+        incrementBytesReadCounter();
```
sahilTakiar: This approach seems to be incrementing the counter before it downloads the file, correct? It may be better to increment the counter after the file has been downloaded. I think it makes the metrics more accurate. For example, if the download of a file fails, the metrics will still report that the whole file was downloaded (even though it wasn't).

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -165,4 +165,15 @@ public Schema getAvroSchema(String file)
       throw new FileBasedHelperException("Failed to open avro file " + file + " due to error " + e.getMessage(), e);
     }
   }
+
+  @Override
+  public long getFileSize(String filePath) throws FileBasedHelperException {
+    try {
+      Path p = new Path(filePath);
+      return fs.getFileStatus(p).getLen();
```
sahilTakiar: You can probably combine lines 172 and 173

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -67,12 +75,19 @@ public FileBasedExtractor(WorkUnitState workUnitState, FileBasedHelper fsHelper)
         Lists.newArrayList(workUnitState.getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FILES_TO_PULL, ""));
     this.statusCount = this.workUnit.getPropAsInt(ConfigurationKeys.FILEBASED_REPORT_STATUS_ON_COUNT,
         ConfigurationKeys.DEFAULT_FILEBASED_REPORT_STATUS_ON_COUNT);
-    this.fsHelper = fsHelper;
+
+    if (fsHelper instanceof SizeAwareFileBasedHelper) {
+      this.fsHelper = (SizeAwareFileBasedHelper)fsHelper;
+    } else {
+      this.fsHelper = new SizeAwareFileBasedHelperDecorator(fsHelper);
```
sahilTakiar: You might want to also check if `fsHelper` is a `FileBasedHelper` before invoking new `SizeAwareFileBasedHelperDecorator(fsHelper)`. If it's not a `FileBasedHelper` probably want to throw some exception.

sahilTakiar:
Just a few comments

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
pcadabam: fsHelper is guaranteed to be a FileBasedHelper as it is passed in as a FileBasedHelper in the constructor (line 71)

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
pcadabam: Since it was an info I did not want to print the complete stacktrace. May be we can print it in a LOGGER.Warn but it may be an overkill for a missing counter. I can add a debug message with the stacktrace

gobblin-core/src/main/java/gobblin/source/extractor/filebased/SizeAwareFileBasedHelperDecorator.java
pcadabam: Done gobblin.util.decorator

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
pcadabam: Done

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
pcadabam: Moved it after closeCurrentFile()

sahilTakiar:
LGTM

ibuenros:
LGTM

PR #192, opened 2015-07-01
Publisher to add row counts to file names (for kafka adapter)

xkrogen:
Updated with a new publisher which is similar to TimePartitionedDataPublisher but which also updates the file names of the published files to include the row count (how many records the file contains) for use with the Kafka adapter


xkrogen:
@zliu41 @ibuenros Can you review?

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.publisher;
+
+import gobblin.configuration.ConfigurationKeys;
```
zliu41: The import order we use is  java, org, com, gobblin

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * Unlike {@link TimePartitionedDataPublisher}, moves each file individually rather than moving
+ * directories at a time. This is done so that each task/branch can move the file it created,
+ * allowing it to access the row count information and place it into the file name.
+ *
+ * Publisher output directory is $GOBBLIN_WORK_DIR/job-output/{tableName}/{partitionPath}
+ *
+ * @author ekrogen
+ */
+public class TimePartitionedDataWithRowCountsPublisher extends TimePartitionedDataPublisher {
```
zliu41: Does this class only work for jobs that write to time partitioned folders?
I guess it can extend `BaseDataPublisher` and be applied to all jobs, correct?

zliu41: You can see if it can be used for the Wikipedia job which uses `BaseDataPublisher`.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TimePartitionedDataWithRowCountsPublisher.class);
+
+  public TimePartitionedDataWithRowCountsPublisher(State state) {
+    super(state);
+  }
+
+  @Override
+  public void publishData(Collection<? extends WorkUnitState> states) throws IOException {
+
+    Set<Path> publisherDirsProcessed = new HashSet<Path>();
```
zliu41: Please use Sets.newHashSet() for consistency

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: Currently only works for TimePartitioned since it makes assumptions about the presence of date folders according to WRITER_PARTITION_PATTERN. I can probably change it to be more flexible if you want.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: fixed

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: fixed

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      // Upon successfully committing the data to the final output directory, set states
+      // of successful tasks to COMMITTED. leaving states of unsuccessful ones unchanged.
+      // This makes sense to the COMMIT_ON_PARTIAL_SUCCESS policy.
+      workUnitState.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
+    }
+  }
+
+  private Path getWriterOutputGlobMatcher(WorkUnitState workUnitState, int branchId, Path writerOutputDir,
```
zliu41: Please add javadoc to explain what this method does. Give an example.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+          LOG.warn("WorkUnit " + workUnitState.getId() + " produced no data");
+          workUnitState.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
+          break;
+        }
+
+        String fileExtension = WriterOutputFormat.valueOf(workUnitState.getProp(ForkOperatorUtils
+                .getPropertyNameForBranch(ConfigurationKeys.WRITER_OUTPUT_FORMAT_KEY, this.numBranches, branchId),
+            WriterOutputFormat.AVRO.name())).getExtension();
+
+        String writerOutputFileName = WriterUtils.getWriterFileName(workUnitState, this.numBranches,
```
zliu41: Is this assuming each workunit outputs to a single file? If a workunit pulls records in different hours, it will output to diffierent files.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+            writerOutputPathStr.indexOf(writerOutputDir.toString()) + writerOutputDir.toString().length() + 1);
+
+        int recordsWritten = workUnitState.getPropAsInt(ForkOperatorUtils
+            .getPropertyNameForBranch(ConfigurationKeys.WRITER_ROWS_WRITTEN, this.numBranches, branchId));
+
+        // Remove extension, insert number of records written, put extension back on the end
+        String outputFileName = writerOutputFileName.substring(0,
+            writerOutputFileName.length() - fileExtension.length()) + recordsWritten + "." + fileExtension;
+
+        Path outputPath = new Path(publisherOutputDir, pathSuffix + Path.SEPARATOR + outputFileName);
```
zliu41: This method is too long. You can separate some parts into smaller methods such as `getOutputPath()` etc.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+                .getPropertyNameForBranch(ConfigurationKeys.WRITER_OUTPUT_FORMAT_KEY, this.numBranches, branchId),
+            WriterOutputFormat.AVRO.name())).getExtension();
+
+        String writerOutputFileName = WriterUtils.getWriterFileName(workUnitState, this.numBranches,
+            branchId, workUnitState.getId(), fileExtension);
+
+        Path writerOutputFullGlob =
+            getWriterOutputGlobMatcher(workUnitState, branchId, writerOutputDir, writerOutputFileName);
+
+        Path writerOutputFilePath = null;
```
zliu41: Use `com.google.common.base.Optional` instead of null: https://code.google.com/p/guava-libraries/wiki/UsingAndAvoidingNullExplained

(don't use `java.util.Optional` since we need to support Java 6 & 7).

liyinan926:
@xkrogen I will take a look after you update this PR with the fixes.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: removed this assumption

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: ended up disappearing when I refactored to remove the single output file assumption - resolved

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: fixed

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRowCountsPublisher.java
xkrogen: Actually, more complicated than I originally thought. Needs more work.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -172,9 +170,11 @@ public void write(GenericRecord record) throws IOException {
     // If the path is in pathToWriterMap simply retrieve the writer, and write the record
     if (!this.pathToWriterMap.containsKey(writerOutputPath)) {
 
-      LOG.info("Creating a new DataWriter for path: " + new Path(this.baseFilePath, writerOutputPath));
+      Path writerOutputPathWithBase = new Path(this.baseFilePath, writerOutputPath);
+
+      LOG.info("Creating a new DataWriter for path: " + writerOutputPathWithBase);
```
xkrogen: Previously this log output was actually incorrect (outputted e.g. CaptchaEvent/CaptchaEvent/hourly/2015/06/29/01 - the baseFilePath was doubled). Having the writerOutputPath contain the baseFilePath also created an inconsistency since WriterUtils.getWriterOutputDir already appends the baseFilePath onto the end of the output directory so I refactored this a little. This is why getPathForColumnValue changed below (the javadoc comment for getPathForColumnValue didn't seem to imply that it would be adding the baseFilePath, anyway)

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataWithRecordCountsPublisher.java
```diff
@@ -0,0 +1,150 @@
...
+
+        if (filePathToRecordCountMappings.isEmpty()) {
+          LOG.warn("No record count mappings found for WorkUnit " + workUnitState.getId() + " branch " + branchId);
+        }
+        for (String filePathToRecordCountMap : filePathToRecordCountMappings) {
+          String filePath;
+          int recordsWritten;
+          try {
+            filePath = filePathToRecordCountMap.split(":")[0];
+            recordsWritten = Integer.parseInt(filePathToRecordCountMap.split(":")[1]);
```
xkrogen: Let me know if you have any suggestions on a better way to achieve this type of mapping - thought about JSON serializing but that seemed like overkill

xkrogen:
@zliu41 @liyinan926  Updated with an almost entirely new implementation that can handle the Writer outputting multiple files

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
zliu41: Good point. btw I don't think `this.baseFilePath` is needed in this class. You can remove it. In your code the LOG is still printing CaptchaEvent twice.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -172,9 +170,11 @@ public void write(GenericRecord record) throws IOException {
...
 
-      DataWriter<GenericRecord> avroHdfsDataWriter = createAvroHdfsDataWriterForPath(writerOutputPath);
+      DataWriter<GenericRecord> avroHdfsDataWriter = createAvroHdfsDataWriterForPath(writerOutputPathWithBase);
```
zliu41: Why change this line?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -247,16 +247,25 @@ public long bytesWritten() throws IOException {
...
   public void close() throws IOException {
+    String fileName = WriterUtils.getWriterFileName(this.properties, this.numBranches, this.branch, this.writerId,
+        this.writerOutputFormat.getExtension());
+    List<String> pathToRecordCountMapList = Lists.newArrayList();
+
     boolean closeFailed = false;
     for (Entry<Path, DataWriter<GenericRecord>> entry : this.pathToWriterMap.entrySet()) {
       try {
         entry.getValue().close();
+        pathToRecordCountMapList.add(new Path(entry.getKey(), fileName) + ":" + entry.getValue().recordsWritten());
```
zliu41: Seems it's easier to just rename the files here, rather than passing the record count to publishers. Are there benefits of doing so?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
xkrogen: No, it only has it once. writerOutputPath no longer contains baseFilePath - only writerOutputPathWithBase does

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
xkrogen: Since writerOutputPath no longer contains baseFilePath, which is needed in this argument

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
zliu41: I see. Then I think you can just have writerOutputPath contain baseFilePath. There doesn't seem to be any inconsistency since `WriterUtils.getWriterOutputDir` won't append baseFilePath to the end of the dir if property `writer.file.path` is already set, which is the case when creating the DataWriter. So you can simply remove the redundant `baseFilePath` from `LOG.info`.

Also, the name `baseFilePath` is very confusing, please rename it to `datasetName`.

xkrogen: Alright, sounds fine. baseFilePath is based off of the configuration key WRITER_FILE_PATH, though, are you sure we want to rename it?

zliu41: Yes. We probably should have had a better design of how different parts of the file path are stored in different places (in particular, redesign `WriterUtils.getWhateverPathOrNameOrDir` to make them more intuitive). But when looking at the name `baseFilePath` I have absolutely no idea what it refers to. So for now it's better to give it a better name.

xkrogen:
@zliu41 pushed a new version that renames the files within the writer so no information needs to be passed from writer to publisher

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
```diff
@@ -0,0 +1,92 @@
...
+  private final WriterOutputFormat writerOutputFormat;
+  private final State properties;
+  private final int numBranches;
+  private final int branch;
+
+  public AvroHdfsTimePartitionedWithRecordCountsWriter(Destination destination, String writerId, Schema schema,
+      WriterOutputFormat writerOutputFormat, int numBranches, int branch) {
+    super(destination, writerId, schema, writerOutputFormat, numBranches, branch);
+
+    this.writerId = writerId;
```
zliu41: Please change these variables to protected, and remove line 58-62.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
```diff
@@ -0,0 +1,92 @@
...
+    super.close();
+
+    String fileName = WriterUtils.getWriterFileName(this.properties, this.numBranches, this.branch, this.writerId,
+        this.writerOutputFormat.getExtension());
+
+    Configuration conf = new Configuration();
+    URI fileSystemURI = URI.create(this.properties.getProp(ForkOperatorUtils
+            .getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, this.numBranches, this.branch),
+        ConfigurationKeys.LOCAL_FS_URI));
+    FileSystem fileSystem = FileSystem.get(fileSystemURI, conf);
```
zliu41: This needs to be consistent with how `FsDataWriter` creates the filesystem (which involves proxy).
I suggest to have `fs` in `AvroHdfsDataWriter` accessible by adding `getFs()`. Then in this class, for each map entry, use `(AvroHdfsDataWriter) entry.getValue().getFs()`.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
```diff
@@ -0,0 +1,77 @@
...
+public class AvroHdfsTimePartitionedWithRecordCountsWriter extends AvroHdfsTimePartitionedWriter {
+
+  public AvroHdfsTimePartitionedWithRecordCountsWriter(Destination destination, String writerId, Schema schema,
+      WriterOutputFormat writerOutputFormat, int numBranches, int branch) {
+    super(destination, writerId, schema, writerOutputFormat, numBranches, branch);
+  }
+
+  @Override
+  public void close() throws IOException {
+    super.close();
```
liyinan926: `super.close()` should be called at the end in the `finally` part of a `try-finally` clause. The rest code should all be in the `try` part.

xkrogen: We have to do super.close() before we rename since the files get created upon close()

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
xkrogen: I copied that file system creation code from FsDataWriter (lines 53-56 of FsDataWriter), is that not correct?: 

    // Initialize file system
    String uri = properties.getProp(
        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, numBranches, branchId),
        ConfigurationKeys.LOCAL_FS_URI);
    this.fs = FileSystem.get(URI.create(uri), conf);

xkrogen: Actually regardless to keep the logic centralized I agree that having a getFs() would be better, I will do so.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
liyinan926: Got it.

zliu41:
LGTM.

liyinan926:
LGTM.

PR #193, opened 2015-07-01
Modify PasswordManager to allow BasicTextEncryptor

zliu41:
Also modified the encryption tool to allow reading master password from file.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/password/PasswordManager.java
+++ b/gobblin-api/src/main/java/gobblin/password/PasswordManager.java
@@ -15,6 +15,8 @@ package gobblin.password;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.AbstractMap;
+import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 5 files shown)


gobblin-api/src/main/java/gobblin/password/PasswordManager.java
```diff
@@ -63,12 +67,17 @@ public PasswordManager load(Optional<String> masterPassword) {
             }
           });
 
-  private final Optional<StrongTextEncryptor> encryptor;
+  private Optional<TextEncryptor> encryptor;
 
   private PasswordManager(Optional<String> masterPassword) {
```
chavdar: Please pass the encryptor class as an optional argument in the constructor. Using a static field creates all kinds of race conditions.

zliu41:
@chavdar updated. Changed key of instance cache from `Optional<String>` to `Map.Entry<Optional<String>, Boolean>`, where true means strong encryptor and false means basic encryptor.

chavdar:
Please add a unit test to check the new plain/strong encryption functionality.

zliu41:
@chavdar updated.

chavdar:
LGTM.

PR #194, opened 2015-07-01
Implement proxied file system operation

ydailinkedin:
Added one option in FsDataWriter for doing file system operations as a proxy user. Tested on spades.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -440,19 +440,19 @@ public class ConfigurationKeys {
   public static final String DEFAULT_METRICS_ENABLED = Boolean.toString(true);
   public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
   public static final String DEFAULT_METRICS_FILE_SUFFIX = "";
-  public static final String METRICS_REPORTING_FILE_ENABLED_KEY =
-      METRICS_CONFIGURATIONS_PREFIX + "reporting.file.enabled";
+  public static final String METRICS_REPORTING_FILE_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX
+      + "reporting.file.enabled";
   public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(false);
-  public static final String METRICS_REPORTING_JMX_ENABLED_KEY =
-      METRICS_CONFIGURATIONS_PREFIX + "reporting.jmx.enabled";
+  public static final String METRICS_REPORTING_JMX_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX
+      + "reporting.jmx.enabled";
   public static final String DEFAULT_METRICS_REPORTING_JMX_ENABLED = Boolean.toString(false);
-  public static final String METRICS_REPORTING_KAFKA_ENABLED_KEY =
-      METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.enabled";
+  public static final String METRICS_REPORTING_KAFKA_ENABLED_KEY = METRICS_CONFIGURATIONS_PREFIX
+      + "reporting.kafka.enabled";
   public static final String DEFAULT_METRICS_REPORTING_KAFKA_ENABLED = Boolean.toString(false);
   public static final String METRICS_REPORTING_KAFKA_FORMAT = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.format";
   public static final String DEFAULT_METRICS_REPORTING_KAFKA_FORMAT = "json";
-  public static final String METRICS_REPORTING_KAFKA_USE_SCHEMA_REGISTRY =
-      METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.avro.use.schema.registry";
+  public static final String METRICS_REPORTING_KAFKA_USE_SCHEMA_REGISTRY = METRICS_CONFIGURATIONS_PREFIX
+      + "reporting.kafka.avro.use.schema.registry";
   public static final String DEFAULT_METRICS_REPORTING_KAFKA_USE_SCHEMA_REGISTRY = Boolean.toString(false);
   public static final String METRICS_KAFKA_BROKERS = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.brokers";
   public static final String METRICS_KAFKA_TOPIC = METRICS_CONFIGURATIONS_PREFIX + "reporting.kafka.topic";
@@ -484,6 +484,15 @@ public class ConfigurationKeys {
    * Password encryption and decryption properties.
    */
   public static final String ENCRYPT_KEY_LOC = "encrypt.key.loc";
+
+  /**
+   * Proxy Filesystem operation properties.
+   */
+  public static final String SHOULD_FS_PROXY_AS_USER = "should.fs.proxy.as.user";
+  public static final boolean DEFAULT_SHOULD_FS_PROXY_AS_USER = false;
+  public static final String FS_PROXY_AS_USER_NAME = "fs.proxy.as.user.name";
+  public static final String FS_PROXY_AS_USER_TOKEN_FILE = "fs.proxy.as.user.token.file";
+  public static final String SUPER_USER_NAME_TO_PROXY_AS_OTHERS = "super.user.name.to.proxy.as.others";
 
   /**
    * Other configuration properties.
--- a/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
@@ -14,6 +14,7 @@ package gobblin.writer;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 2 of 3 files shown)


ydailinkedin:
@sahilTakiar can you review this? Thanks!

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -0,0 +1,148 @@
...
+   * @param logInUserName
+   * @param properties
+   * @param authType is either TOKEN or KEYTAB.
+   * @param authPath is the KEYTAB location if the authType is KEYTAB; otherwise, it is the token file.
+   * @param uri File system URI.
+   * @throws IOException
+   * @throws InterruptedException
+   * @throws URISyntaxException
+   */
+  public ProxiedFileSystemWrapper(String proxyUserName, String logInUserName, State properties, AuthType authType,
```
sahilTakiar: Why not add all the logic in this method into getProxiedFilesystem. All you have to do is create a `final FileSystem proxiedFileSystem` object before the `doAs` block, and then you can set `proxiedFileSystem` in the `doAs` block. For example: http://stackoverflow.com/questions/9545076/access-variables-of-outer-class-in-java

I think having this logic in `getProxiedFileSystem` makes more sense than having this logic in the constructor

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -0,0 +1,148 @@
...
+      String authPath, String uri) throws IOException, InterruptedException, URISyntaxException {
+    // If the authentication type is KEYTAB, log in first before creating a proxy user.
+    if (authType == AuthType.KEYTAB) {
```
sahilTakiar: Since `AuthType` is an enum, why not use a case, switch statement

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -0,0 +1,148 @@
...
+      LOG.info("Logging in for user " + logInUserName);
+      UserGroupInformation.loginUserFromKeytab(logInUserName, authPath);
+    }
+    UserGroupInformation proxyUser =
+        UserGroupInformation.createProxyUser(proxyUserName, UserGroupInformation.getLoginUser());
```
sahilTakiar: Should `UserGroupInformation.getLoginUser()` be `logIngUserName`?

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -49,17 +59,42 @@ public FsDataWriter(State properties, String fileName, int numBranches, int bran
...
+    String uri =
+        properties
+            .getProp(ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, numBranches,
+                branchId), ConfigurationKeys.LOCAL_FS_URI);
+
+    if (properties.getPropAsBoolean(ConfigurationKeys.FS_AS_PROXY_USER, ConfigurationKeys.DEFAULT_FS_AS_PROXY_USER)) {
+      // Initialize file system as a proxy user. Both properties PROXY_USER_NAME and TOKEN_FILE_NAME should be set in Source.
+      Preconditions.checkArgument(StringUtils.isNotBlank(properties.getProp(ConfigurationKeys.PROXY_USER_NAME)),
+          "State does not contain a proper proxy user name");
+      Preconditions.checkArgument(StringUtils.isNotBlank(properties.getProp(ConfigurationKeys.TOKEN_FILE_NAME)),
```
sahilTakiar: Can we move these precondition checks into `ProxiedFileSystemWrapper`, that way we don't have to re-implement these checks each place `ProxiedFileSystemWrapper` is used

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -486,6 +486,14 @@
   public static final String ENCRYPT_KEY_LOC = "encrypt.key.loc";
 
   /**
+   * Proxy Filesystem operation properties.
+   */
+  public static final String FS_AS_PROXY_USER = "fs.as.proxy.user";
```
sahilTakiar: Change to something like `writer.fs.proxy.as.user`

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -486,6 +486,14 @@
...
+  public static final boolean DEFAULT_FS_AS_PROXY_USER = false;
+  public static final String PROXY_USER_NAME = "proxy.user.name";
```
sahilTakiar: Change to something like `writer.fs.proxy.as.user.name`

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -486,6 +486,14 @@
...
+  public static final String TOKEN_FILE_NAME = "token.file.name";
```
sahilTakiar: Change to something like `writer.fs.proxy.as.user.token.file`

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
ydailinkedin: UserGroupInformation.getLoginUser returns an instance of UGI.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
ydailinkedin: I am using the same key for deciding if we need proxy in `publisher`. What do you think if I change it to `should.fs.proxy.as.user`?

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -49,17 +56,35 @@ public FsDataWriter(State properties, String fileName, int numBranches, int bran
...
+
+    if (properties.getPropAsBoolean(ConfigurationKeys.SHOULD_FS_PROXY_AS_USER,
+        ConfigurationKeys.DEFAULT_SHOULD_FS_PROXY_AS_USER)) {
+      // Initialize file system as a proxy user.
+      try {
+        this.fs =
+            new ProxiedFileSystemWrapper().getProxiedFileSystem(properties, ProxiedFileSystemWrapper.AuthType.TOKEN,
+                properties.getProp(ConfigurationKeys.FS_PROXY_AS_USER_TOKEN_FILE), uri);
+      } catch (InterruptedException e) {
+        Throwables.propagate(e);
```
liyinan926: Instead of propagating the exceptions, I suggest you simply wraps them into an `IOException` and throw that instead.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -37,11 +42,13 @@
 public abstract class FsDataWriter<D> implements DataWriter<D> {
   private static final Logger LOG = LoggerFactory.getLogger(FsDataWriter.class);
 
-  protected final FileSystem fs;
+  protected FileSystem fs;
```
liyinan926: You can still keep `final` if you throw `IOException` as commented below.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -37,11 +42,13 @@
...
   protected final Path stagingFile;
   protected final Path outputFile;
   protected final State properties;
 
+  // final UserGroupInformation proxyUserByToken;
```
liyinan926: If you don't need this, please remove it.

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -0,0 +1,151 @@
...
+      Token value = new Token();
+      while (tokenReader.next(key, value)) {
+        LOG.info("Find token for " + key);
+        if (key.toString().equals(proxyUserName)) {
+          return Optional.of(value);
+        }
+      }
+
+    } catch (IOException e) {
+      Throwables.propagate(e);
```
liyinan926: Ditto. I suggest allowing the method itself to throw at least `IOException` instead of using `propagate`.

ydailinkedin:
@liyinan926 Updated.

ydailinkedin:
Updated.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -486,6 +486,15 @@
...
 
   /**
+   * Proxy Filesystem operation properties.
+   */
+  public static final String SHOULD_FS_PROXY_AS_USER = "should.fs.proxy.as.user";
+  public static final boolean DEFAULT_SHOULD_FS_PROXY_AS_USER = false;
+  public static final String FS_PROXY_AS_USER_NAME = "fs.proxy.as.user.name";
+  public static final String FS_PROXY_AS_USER_TOKEN_FILE = "fs.proxy.as.user.token.file";
+  public static final String SUPER_USER_NAME_TO_PROXY_AS_OTHERS = "super.user.name.to.proxy.as.others";
+
```
sahilTakiar: Can you prefix this all with "writer." since the fs proxying is only happening in the writer.

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -0,0 +1,143 @@
...
+  private Optional<Token> getTokenFromSeqFile(String authPath, String proxyUserName) throws IOException {
+    Closer closer = Closer.create();
+    try {
+      FileSystem localFs = FileSystem.getLocal(new Configuration());
+      SequenceFile.Reader tokenReader =
+          closer.register(new SequenceFile.Reader(localFs, new Path(authPath), localFs.getConf()));
+      Text key = new Text();
+      Token value = new Token();
+      while (tokenReader.next(key, value)) {
+        LOG.info("Find token for " + key);
```
sahilTakiar: Find --> Found

sahilTakiar:
Minor comment, otherwise LGTM

ydailinkedin:
Merged.

PR #197, opened 2015-07-03
Fixed a bug that causes false-alarm warning log message when loading job configs

liyinan926:

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
@@ -32,9 +32,9 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


liyinan926:
@zliu41 can you review this?

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -226,18 +236,15 @@ private static void loadJobConfigsRecursive(List<Properties> jobConfigs, Propert
...
-    return Sets.newHashSet(Iterables.transform(jobConfigFileExtensionsIterable, new Function<String, String>() {
-          @Override
-          public String apply(String input) {
-            if (input == null) {
-              throw new NullPointerException("Given input is null: " + input);
-            }
-            return input.toLowerCase();
-          }
-        }));
+    return ImmutableSet.copyOf(Iterables.transform(jobConfigFileExtensionsIterable, new Function<String, String>() {
```
zliu41: Is it better to use case insensitive set here?

liyinan926: Yeah, we could use a case insensitive `TreeSet`, but to me that is more for sorting.

zliu41:
LGTM

PR #199, opened 2015-07-06
Fixed a bug that may cause a job to hang if its tasks fail to getExtractor

liyinan926:
This bug is affecting stand-alone jobs using `LocalJobLauncher` that may hang if any of its tasks fails to `getExtractor`. In this case, the `finally` part of the `Task.run` method will be executed, and `this.taskStateTracker.onTaskCompletion(this)` is called. In the implementation class `LocalTaskStateTracker2`, `this.taskExecutor.retry(task)` is called, which will throw an exception on `task.getTaskState().getPropAsInt(ConfigurationKeys.FORK_BRANCHES_KEY)` since the key `FORK_BRANCHES_KEY` is not set because the task failed before the number of branches is known and that key is set. This runtime exception is thrown in the `finally` clause and is not caught, causing the task to hang.

### DIFF
--- a/gobblin-runtime/src/test/resources/log4j.xml
+++ b/gobblin-runtime/src/test/resources/log4j.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+  <appender name="console" class="org.apache.log4j.ConsoleAppender">
+    <param name="Target" value="System.out" />
+    <layout class="org.apache.log4j.PatternLayout">
+      <param name="ConversionPattern"
+        value="%d{yyyy-MM-dd HH:mm:ss z} %-5p [%t] %C %L - %m%n" />
+    </layout>
+  </appender>
+
+  <root>
+    <level value="info" />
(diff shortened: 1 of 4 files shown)


liyinan926:
@zliu41 can you review this?

gobblin-runtime/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
zliu41: Why is this needed? There's a `log4j.xml` in `conf/`, is it possible to reuse that one?

zliu41:
In `LocalTaskStateTracker2.onTaskCompletion`, should `task.markTaskCompletion();` be moved to a finally block to make sure it is called?

gobblin-runtime/src/test/resources/log4j.xml
liyinan926: It is slightly different from that one as this is only used for testing.

liyinan926:
`task.markTaskCompletion` should only be called if all retries fail. So calling it in a finally clause is not appropriate. What we can do is to make sure it is called if any `Throwable` is thrown.

zliu41:
So should it be put in a `catch(Throwable t)` block?

liyinan926:
Yes, updated.

zliu41:
LGTM

PR #200, opened 2015-07-06
Add more parallelization during job finialization

zliu41:
Parallelize the following: (1) moving file from task output dir to final output dir; (2) cleaning task output dir and task staging dir.

1. Renamed `ParallelStateSerDeRunner` to `ParallelRunner`, and add two methods: `deletePath()` and `renamePath()`.
2. Moved `ParallelRunner` from `gobblin-runtime` to `gobblin-utility`, since `gobblin.util.JobLauncherUtils` depends on it, but `gobblin-utility` cannot depend on `gobblin-runtime`.
3. `ParallelRunnerTest` has to stay in `gobblin-runtime`, since it depends on `gobblin.runtime.TaskState`.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import org.apache.hadoop.conf.Configuration;
@@ -26,12 +27,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.io.Closer;
 
 import gobblin.configuration.State;
 import gobblin.configuration.WorkUnitState;
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.util.ForkOperatorUtils;
+import gobblin.util.ParallelRunner;
 import gobblin.util.WriterUtils;
 
 
@@ -53,14 +57,14 @@ public class BaseDataPublisher extends DataPublisher {
   private static final Logger LOG = LoggerFactory.getLogger(BaseDataPublisher.class);
 
   protected final List<FileSystem> fss = Lists.newArrayList();
-  protected int numBranches;
+  protected final Closer closer;
+  protected final int numBranches;
+  protected final int parallelRunnerThreads;
+  protected final Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();
 
-  public BaseDataPublisher(State state) {
+  public BaseDataPublisher(State state) throws IOException {
     super(state);
-  }
-
-  @Override
-  public void initialize() throws IOException {
+    this.closer = Closer.create();
     Configuration conf = new Configuration();
 
     // Add all job configuration properties so they are picked up by Hadoop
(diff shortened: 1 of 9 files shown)


zliu41:
@liyinan926 please review.

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -205,6 +198,37 @@ public Void call()
     }));
   }
 
+  /**
+   * Delete a path.
+   * @param path path to be deleted.
```
liyinan926: Please add `* <p>
    *   This method submits a task to serialize the {@link State} object and returns immediately
    *   after the task is submitted.
    * </p>`.

liyinan926:
For `ParallelRunnerTest`, you can change `TaskState` to `WorkUnitState` and move it to under `gobblin-utility`.

zliu41:
@liyinan926 updated javadoc. Couldn't change `TaskState` to `WorkUnitState` since `WorkUnitState` doesn't have `setStartTime` or `setEndTime`.

liyinan926:
It's fine to delete `setStartTime` and `setEndTime`.

zliu41:
@liyinan926 Updated. Now both `BaseDataPublisher` and `JobLauncherUtils` uses a `map` from Fs URI to `ParallelRunner`, which enables different tasks to share the same `ParallelRunner` if they use the same `FileSystem`. Also updated the unit test for `ParallelRunner`.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -53,10 +57,14 @@
   private static final Logger LOG = LoggerFactory.getLogger(BaseDataPublisher.class);
 
   protected final List<FileSystem> fss = Lists.newArrayList();
+  protected final Closer closer;
   protected int numBranches;
+  protected int parallelRunnerThreads;
+  protected Map<String, ParallelRunner> parallelRunners;
```
liyinan926: I suggest doing `protected final Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();`. By doing this, `parallelRunners` is guaranteed to be non-null when it's used.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -77,11 +85,14 @@ public void initialize() throws IOException {
           ConfigurationKeys.LOCAL_FS_URI));
       this.fss.add(FileSystem.get(uri, conf));
     }
+    this.parallelRunnerThreads =
```
liyinan926: Seems most code in `intialize` can be done in the constructor.  I suggest moving all of them to the constructor.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -133,12 +146,9 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
         }
 
-        if (this.fss.get(branchId).rename(writerOutputDir, publisherOutputDir)) {
-          LOG.info(String.format("Moved %s to %s", writerOutputDir, publisherOutputDir));
-          writerOutputPathsMoved.add(writerOutputDir);
-        } else {
-          throw new IOException("Failed to move from " + writerOutputDir + " to " + publisherOutputDir);
-        }
+        LOG.info(String.format("Moving %s to %s", writerOutputDir, publisherOutputDir));
+        writerOutputPathsMoved.add(writerOutputDir);
```
liyinan926: This is changing the semantics. I suggest doing something like `renamePath(Path, Path, Function<T, Void> callback)` and wrap `writerOutputPathsMoved.add(writerOutputDir)` into the callback, which will be called upon successful rename.

zliu41:
Updated.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
zliu41: I moved `writerOutputPathsMoved.add(writerOutputDir)` after `parallelRunner.renamePath(writerOutputDir, publisherOutputDir)`, which should have the same semantics as before.

liyinan926:
LGTM.

PR #203, opened 2015-07-07
Data Management

ibuenros:
This PR is not complete yet. It is here for reference.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.retention.dataset;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Lists;
+
+import azkaban.utils.Props;
+
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.policy.RetentionPolicy;
+import gobblin.data.management.retention.version.finder.VersionFinder;
+import gobblin.data.management.trash.Trash;
+
+
+/**
+ * Implementation of a {@link gobblin.data.management.retention.dataset.Dataset} that uses a
+ * {@link gobblin.data.management.retention.version.finder.VersionFinder} to find dataset versions, a
+ * {@link gobblin.data.management.retention.policy.RetentionPolicy} to figure out deletable versions, and then deletes
+ * those files and newly empty parent directories.
+ *
+ * <p>
+ *   Concrete subclasses should implement {@link #getVersionFinder} and {@link #getRetentionPolicy}.
+ * </p>
+ *
+ * <p>
+ * Datasets are directories in the filesystem containing data files organized in version-like directory structures.
+ * Example datasets:
+ * </p>
+ *
+ * <p>
+ *   For snapshot based datasets, with the directory structure:
+ *   <pre>
+ *   /path/to/table/
+ *      snapshot1/
+ *          dataFiles...
+ *      snapshot2/
+ *          dataFiles...
+ *   </pre>
+ *   each of snapshot1 and snapshot2 are dataset versions.
+ * </p>
+ *
+ * <p>
+ *   For tracking datasets, with the directory structure:
+ *   <pre>
+ *   /path/to/tracking/data/
+ *      2015/
+ *          06/
+ *              01/
+ *                  dataFiles...
+ *              02/
+ *                  dataFiles...
+ *   </pre>
+ *   each of 2015/06/01 and 2015/06/02 are dataset versions.
+ * </p>
+ *
+ * <p>
+ *  {@link DatasetBase} uses a {@link gobblin.data.management.retention.version.finder.DatasetVersionFinder} to find all
+ *  subdirectories that are versions of this dataset. After that, for each dataset, it uses a
+ *  {@link gobblin.data.management.retention.policy.RetentionPolicy} to decide which versions of the dataset should be
+ *  deleted. For each version deleted, if {@link #deleteEmptyDirectories} it will also look at all parent directories
+ *  and delete directories that are now empty, up to but not including the dataset root.
+ * </p>
+ *
+ * @param <T> type of {@link gobblin.data.management.retention.version.DatasetVersion} supported by this
+ *           {@link gobblin.data.management.retention.dataset.Dataset}.
+ */
+public abstract class DatasetBase<T extends DatasetVersion> implements Dataset {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DatasetBase.class);
+
+  public static final String CONFIGURATION_KEY_PREFIX = "gobblin.retention.";
+  public static final String SIMULATE_KEY = CONFIGURATION_KEY_PREFIX + "simulate";
+  public static final boolean SIMULATE_DEFAULT = false;
+  public static final String SKIP_TRASH_KEY = CONFIGURATION_KEY_PREFIX + "skip.trash";
+  public static final boolean SKIP_TRASH_DEFAULT = false;
+  public static final String DELETE_EMPTY_DIRECTORIES_KEY = CONFIGURATION_KEY_PREFIX + "delete.empty.directories";
+  public static final boolean DELETE_EMPTY_DIRECTORIES_DEFAULT = true;
+
+  protected final FileSystem fs;
+  protected final Trash trash;
+  protected final boolean simulate;
+  protected final boolean skipTrash;
+  protected final boolean deleteEmptyDirectories;
+
+  /**
+   * Get {@link gobblin.data.management.retention.version.finder.VersionFinder} to use.
+   */
+  public abstract VersionFinder<? extends T> getVersionFinder();
+
+  /**
+   * Get {@link gobblin.data.management.retention.policy.RetentionPolicy} to use.
+   * @return
+   */
+  public abstract RetentionPolicy<T> getRetentionPolicy();
+
+  public DatasetBase(FileSystem fs, Props props) throws IOException {
+    this.fs = fs;
+
+    this.simulate = props.getBoolean(SIMULATE_KEY, SIMULATE_DEFAULT);
+    this.skipTrash = props.getBoolean(SKIP_TRASH_KEY, SKIP_TRASH_DEFAULT);
+    this.deleteEmptyDirectories = props.getBoolean(DELETE_EMPTY_DIRECTORIES_KEY, DELETE_EMPTY_DIRECTORIES_DEFAULT);
+
+    if(this.simulate || this.skipTrash) {
+      this.trash = null;
+    } else {
+      this.trash = new Trash(fs, props);
+    }
+  }
+
+  public DatasetBase(FileSystem fs, Trash trash, boolean simulate, boolean skipTrash, boolean deleteEmptyDirectories) {
+    this.fs = fs;
+    this.trash = trash;
+    this.simulate = simulate;
+    this.skipTrash = skipTrash;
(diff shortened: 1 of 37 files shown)


ibuenros:
@chavdar @ydailinkedin can you review? Thanks.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
```diff
@@ -0,0 +1,215 @@
...
+   */
+  public abstract VersionFinder<? extends T> getVersionFinder();
+
+  /**
+   * Get {@link gobblin.data.management.retention.policy.RetentionPolicy} to use.
+   * @return
+   */
+  public abstract RetentionPolicy<T> getRetentionPolicy();
+
+  public DatasetBase(FileSystem fs, Props props) throws IOException {
```
chavdar: This constructor and the DatasetBase(FileSystem fs, Trash trash, boolean simulate, boolean skipTrash, boolean deleteEmptyDirectories) constructor should be sharing code or the former should be using the latter. Otherwise, we are risking having inconsistent logic in the constructor.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
```diff
@@ -0,0 +1,215 @@
...
+    for(DatasetVersion versionToDelete : deletableVersions) {
+      LOGGER.info("Deleting dataset version " + versionToDelete);
+
+      Set<Path> pathsToDelete = versionToDelete.getPathsToDelete();
+      LOGGER.info("Deleting paths: " + Arrays.toString(pathsToDelete.toArray()));
+
+      boolean deletedAllPaths = true;
+
+      for(Path path: pathsToDelete) {
+        if (!this.simulate) {
```
chavdar: We should log the directories that would have been deleted in the simulate mode.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/RetentionPolicy.java
```diff
@@ -0,0 +1,30 @@
...
+  public Class<? extends DatasetVersion> versionClass();
+
+  /**
+   * Logic to decide which dataset versions should be deleted. Only datasets returned will be deleted from filesystem.
+   *
+   * @param allVersions {@link java.util.SortedMap} of all dataset versions in the file system,
+   *                                               sorted from newest to oldest.
+   * @return Map of dataset versions that should be deleted.
+   */
+  public List<T> preserveDeletableVersions(List<T> allVersions);
```
chavdar: The naming of this method is confusing. It returns a list of versions to be deleted, not preserved. How about something along the lines of listDeletableVersions() ?

Please fix the javadoc is since the method does not return a map.

I think the order of the versions is not significant from an API point of view. In that case, it is better to use Collection<T>.

gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ConfigurableGlobDataset.java
```diff
@@ -0,0 +1,107 @@
...
+import gobblin.data.management.retention.dataset.Dataset;
+import gobblin.data.management.retention.dataset.finder.DatasetFinder;
+import gobblin.data.management.retention.version.DatasetVersion;
+
+
+/**
+ * A configurable {@link gobblin.data.management.retention.dataset.finder.DatasetFinder} that looks for
+ * {@link gobblin.data.management.retention.dataset.Dataset}s using a glob pattern.
+ */
+public class ConfigurableGlobDataset implements DatasetFinder {
```
chavdar: I think the class needs to be named ConfigurableGlobDatasetFinder since it implements the DatasetFinder interface, not the Dataset interface.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/VersionFinder.java
```diff
@@ -0,0 +1,45 @@
...
+   * Find dataset versions for the dataset at {@link org.apache.hadoop.fs.Path}. Dataset versions are subdirectories of the
+   * input {@link org.apache.hadoop.fs.Path} representing a single manageable unit in the dataset.
+   * See {@link gobblin.data.management.retention.DatasetCleaner} for more information.
+   *
+   * @param dataset {@link org.apache.hadoop.fs.Path} to directory containing all versions of a dataset.
+   * @return Map of {@link gobblin.data.management.retention.version.DatasetVersion} and {@link org.apache.hadoop.fs.FileStatus}
+   *        for each dataset version found.
+   * @throws IOException
+   */
+  public List<T> findDatasetVersions(Dataset dataset) throws IOException;
```
chavdar: nit: I would change the return type to Collection<T> since the order is not significant from an API point-of-view. Also, please fix the javadoc for @return.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
chavdar: Never mind. I missed the logging statement before that.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/DatasetBase.java
```diff
@@ -0,0 +1,215 @@
...
+ *  deleted. For each version deleted, if {@link #deleteEmptyDirectories} it will also look at all parent directories
+ *  and delete directories that are now empty, up to but not including the dataset root.
+ * </p>
+ *
+ * @param <T> type of {@link gobblin.data.management.retention.version.DatasetVersion} supported by this
+ *           {@link gobblin.data.management.retention.dataset.Dataset}.
+ */
+public abstract class DatasetBase<T extends DatasetVersion> implements Dataset {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DatasetBase.class);
```
chavdar: Please use a non-static logger. You may still use the static version as a default but allow for callers to pass their category name. This will allow use in parallel multiple cleaners for different datasets and still now which cleaner logged each message.

chavdar:
Your wiki is really nice. You should move it to github once the code gets merged.

ibuenros:
@chavdar addressed comments.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/finder/DatasetFinder.java
```diff
@@ -0,0 +1,38 @@
...
+ * </p>
+ */
+public interface DatasetFinder {
+
+  /**
+   * Find all {@link gobblin.data.management.retention.dataset.Dataset}s in the file system.
+   * @return List of {@link gobblin.data.management.retention.dataset.Dataset}s in the file system.
+   * @throws IOException
+   */
+  public List<Dataset> findDatasets() throws IOException;
```
ydailinkedin: Can you add one more method, `findDatasets(FileSystem fs)`, since it is not always the case that one fs has read permission to all directories in that dataset.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/NewestKRetentionPolicy.java
```diff
@@ -0,0 +1,46 @@
...
+public class NewestKRetentionPolicy implements RetentionPolicy<DatasetVersion> {
+
+  public static final String VERSIONS_RETAINED_KEY = DatasetCleaner.CONFIGURATION_KEY_PREFIX +
+      "versions.retained";
+  public static final int VERSIONS_RETAINED_DEFAULT = 2;
+
+  private final int versionsRetained;
+
+  public NewestKRetentionPolicy(Props props) {
+    this.versionsRetained = props.getInt(VERSIONS_RETAINED_KEY, VERSIONS_RETAINED_DEFAULT);
```
ydailinkedin: Can you add one log here to show how many versions will be retained, especially when the default value is used.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/TimestampedDatasetVersion.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public int compareTo(DatasetVersion other) {
+    TimestampedDatasetVersion otherAsDateTime = (TimestampedDatasetVersion)other;
+    return this.version.compareTo(otherAsDateTime.version);
+  }
+
+  @Override
+  public String toString() {
```
ydailinkedin: Can you also include `path` info in `toString()`.

gobblin-data-management/src/main/java/gobblin/data/management/trash/Trash.java
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Get location of Trash directory. Parsed from props at key {@link #TRASH_LOCATION_KEY}, defaulting to
+   * /home/directory/_GOBBLIN_TRASH.
+   * @param fs {@link org.apache.hadoop.fs.FileSystem} where trash should be found.
+   * @param props {@link azkaban.utils.Props} containing trash configuration.
+   * @return {@link org.apache.hadoop.fs.Path} for trash directory.
+   * @throws java.io.IOException
+   */
+  public static Path getTrashLocation(FileSystem fs, Props props) throws IOException {
+    Preconditions.checkArgument(props.containsKey(TRASH_LOCATION_KEY), "Trash location not provided.");
```
ydailinkedin: Is this check necessary, given that you have a default value for `TRASH_LOCATION_KEY`?

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/DatasetVersionFinder.java
```diff
@@ -0,0 +1,84 @@
...
+   */
+  public abstract Path globVersionPattern();
+
+  /**
+   * Parse {@link gobblin.data.management.retention.version.DatasetVersion} from the path of a dataset version.
+   * @param pathRelativeToDatasetRoot {@link org.apache.hadoop.fs.Path} of dataset version relative to dataset root.
+   * @param fullPath full {@link org.apache.hadoop.fs.Path} of the dataset version.
+   * @return {@link gobblin.data.management.retention.version.DatasetVersion} for that path.
+   */
+  public abstract T getDatasetVersion(Path pathRelativeToDatasetRoot, Path fullPath);
```
ydailinkedin: The default constraint here is that some comparable information needs to be contained in `pathRelativeToDatasetRoot`. This may not work well when we purely run retention based on modifiedTimeStamp (for example). I am fine with the current implementation. But it may be better to have a generic comparable object as the first parameter in this method, instead of `Path`.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/finder/DatasetFinder.java
ibuenros: Actually, after thinking about it, I would prefer if DatasetFinder handled its own filesystems. 
The way I see it, if wanting datasets for multipe users, the dataset finder would itself switch filesystems based on the user needed (in this case, the fs passed at constructor would be used only to get the URI of the file system).
The problem of the method `findDatasets(FileSystem fs)` is that this makes the driver program responsible for figuring out file systems, and I would prefer this responsibility to be delegated to the "MultiUserDatasetFinder". Any thoughts?

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/DatasetVersionFinder.java
ibuenros: I don't think this is limiting for two reasons:
1. getDatasetVersion is free to use the file system to do anything it wants (e.g. getting modified time, or even opening a file to get version metadata). It can also load an lsr and figure out information from there.
2. DatasetVersionFinder is just a particular implementation of the interface VersionFinder. If radically different logic is necessary, a new VersionFinder can be implemented. The only method in VersionFinder is a very generic `findDatasetVersions(Dataset)`.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/TimestampedDatasetVersion.java
ibuenros: fixed

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/NewestKRetentionPolicy.java
ibuenros: fixed

gobblin-data-management/src/main/java/gobblin/data/management/trash/Trash.java
ibuenros: true, forgot to remove this. fixed.

ibuenros:
Addressed comments.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/finder/DatasetFinder.java
ydailinkedin: Yes, it makes sense to me. We can keep DatasetFinder as it is.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/DatasetVersionFinder.java
ydailinkedin: True. I didn't notice it implements VersionFinder.

ydailinkedin:
LGTM.

chavdar:
LGTM.

PR #204, opened 2015-07-08
refactor JobLauncherUtils

zliu41:
created method `getFsWithProxy` so that it can be used by both `cleanStagingData` methods.


zliu41:
@ydailinkedin please review

ydailinkedin:
LGTM.

PR #205, opened 2015-07-08
several fixes and optimizations for compaction

zliu41:
1. Added code to add dependency jars to distributed cache.
2. In Mapper class, created method `populateRecord` to replace `AvroUtils.convertRecordSchema`, which is more efficient, since `AvroUtils.convertRecordSchema` needs to serialize the record to byte array and then deserialize it.
3. For Avro key schema, created method `removeUncomparableFields` to replace `getAllFieldsExceptMap`, since we not only need to remove MAP fields, but also need to remove ARRAY and ENUM fields, as well as UNIONS that contain one of those types. Also, `getAllFieldsExceptMap` was not recursive, which is now fixed.

The reason to remove those fields in Avro key schema is that a schema that contains these fields cannot be used as the dedup key.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyMapper.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyMapper.java
@@ -55,7 +55,7 @@ public class AvroKeyMapper extends Mapper<AvroKey<GenericRecord>, NullWritable, 
     if (context.getNumReduceTasks() == 0) {
       context.write(key, NullWritable.get());
     } else {
-      populateRecord(key.datum(), outKey.datum());
+      populateComparableKeyRecord(key.datum(), outKey.datum());
       outValue.datum(key.datum());
       context.write(outKey, outValue);
     }
@@ -64,9 +64,10 @@ public class AvroKeyMapper extends Mapper<AvroKey<GenericRecord>, NullWritable, 
   /**
    * Populate the target record, based on the field values in the source record.
    * Target record's schema should be a subset of source record's schema.
-   * Target record's schema cannot have MAP or ARRAY fields.
+   * Target record's schema cannot have MAP, ARRAY or ENUM fields, or UNION fields that
+   * contain these fields.
    */
-  private static void populateRecord(GenericRecord source, GenericRecord target) {
+  private static void populateComparableKeyRecord(GenericRecord source, GenericRecord target) {
     for (Field field : target.getSchema().getFields()) {
       if (field.schema().getType() == Schema.Type.UNION) {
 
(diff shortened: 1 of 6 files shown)


zliu41:
@ibuenros please review

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
```diff
@@ -87,7 +91,9 @@ private Schema getKeySchema(Job job, Schema newestSchema) throws IOException {
     Schema keySchema = null;
     if (useAllAttributesForCompaction()) {
       LOG.info("Using all attributes in the schema (except Map fields) for compaction");
-      keySchema = getAllFieldsExceptMap(newestSchema);
+      System.out.println("Original schema: " + newestSchema);
```
ibuenros: Please remove print statements.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
```diff
@@ -161,17 +167,60 @@ private boolean useAllAttributesForCompaction() {
         ConfigurationKeys.DEFAULT_COMPACTION_USE_ALL_ATTRIBUTES);
   }
 
-  private Schema getAllFieldsExceptMap(Schema schema) {
+  /**
+   * Remove map, array, enum fields, as well as union fields that contain map, array or enum,
+   * from an Avro schema.
+   */
+  private static Optional<Schema> removeUncomparableFields(Schema schema) {
```
ibuenros: This seems like a useful utility method. Can we move it to a public utility class?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyMapper.java
```diff
@@ -45,22 +45,60 @@
...
       context.write(outKey, outValue);
     }
   }
 
+  /**
+   * Populate the target record, based on the field values in the source record.
+   * Target record's schema should be a subset of source record's schema.
+   * Target record's schema cannot have MAP or ARRAY fields.
+   */
+  private static void populateRecord(GenericRecord source, GenericRecord target) {
```
ibuenros: Can you rename this to `populateComparableKeyRecord` or something similar? The current name does not suggest that target Schema needs to be comparable, and it might be confusing in the future. (I know it's in the comments, but renaming the method is very easy).

ibuenros:
I'm a bit confused, shouldn't gobblin-runtime itself handle the distributed cache?

zliu41:
@ibuenros Compaction does not use Gobblin runtime. It's something completely independent, similar as Camus daily sweeper..

zliu41:
@ibuenros updated

ibuenros:
LGTM

PR #207, opened 2015-07-08
Final task state

ibuenros:
This PR implements final construct states that get added to the Task state at the end of the task execution. This allows for recovering interesting construct states in the driver.


ibuenros:
@liyinan926 can you review? Thanks!

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -43,6 +44,8 @@
  */
 public class WorkUnitState extends State {
 
+  private final static String FINAL_CONSTRUCT_STATE_PREFIX = "construct.final.state.";
```
liyinan926: `static final` please to be consistent with everywhere else.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -261,4 +264,55 @@ public void write(DataOutput out)
   public String toString() {
     return super.toString() + "\nWorkUnit: " + getWorkunit().toString() + "\nExtract: " + getExtract().toString();
   }
+
+  /**
+   * Adds all properties from {@link gobblin.configuration.State} to this {@link gobblin.configuration.WorkUnitState}.
+   *
+   * <p>
+   *   A property with name "property" will be added to this object with the key
+   *   "{@link #FINAL_CONSTRUCT_STATE_PREFIX}[.<infix>].property
```
liyinan926: Missing `"` at the end.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -261,4 +264,55 @@ public void write(DataOutput out)
...
+   * to this {@link gobblin.configuration.WorkUnitState}. All such properties will be stripped of
+   * {@link #FINAL_CONSTRUCT_STATE_PREFIX} but not of any infixes.
+   *
+   * <p>
+   *   For example, if state={sample.property: sampleValue}
+   *   then
+   *   <pre>
+   *     {@code
+   *        this.addFinalConstructState("infix",state);
+   *        this.getFinalConstructState()
```
liyinan926: Missing `;` at the end.

gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
```diff
@@ -0,0 +1,51 @@
...
+   * Stop the timer and submit the event.
+   */
+  public void stop() {
+    if(this.stopped) {
+      throw new RuntimeException("Timer has already been stopped.");
+    }
+    this.stopped = true;
+    long endTime = System.currentTimeMillis();
+    long duration = endTime - this.startTime;
+    submitter.submit(this.name, EventSubmitter.EVENT_TYPE, TIMING_EVENT, START_TIME, Long.toString(this.startTime),
```
liyinan926: `this.submitter`.

gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
```diff
@@ -0,0 +1,51 @@
...
+    this.submitter = submitter;
+    this.startTime = System.currentTimeMillis();
+  }
+
+  /**
+   * Stop the timer and submit the event.
+   */
+  public void stop() {
+    if(this.stopped) {
+      throw new RuntimeException("Timer has already been stopped.");
```
liyinan926: I think it just needs to return here, and you can make clear in the Javadoc that it will be a no-op if it has already been stopped.

gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedDataWriterDecorator.java
```diff
@@ -89,6 +89,15 @@ public long bytesWritten()
   }
 
   @Override
+  public State getFinalState() {
```
liyinan926: Seems you can just do `this.getFinalState()` like you do for other `Decorator` classes.

gobblin-core/src/main/java/gobblin/instrumented/extractor/InstrumentedExtractorDecorator.java
```diff
@@ -80,6 +81,15 @@ public long getHighWatermark() {
   }
 
   @Override
+  public State getFinalState() {
+    if(this.embeddedExtractor instanceof FinalState) {
```
liyinan926: You can simply do `return this.getFinalState();` like other `Decorator` classes.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -156,7 +153,7 @@ public MetricContext apply(JobMetrics input) {
     }
 
     new EventSubmitter.Builder(JobMetrics.get(jobId).getMetricContext(), "gobblin.runtime").build().
-        submit(EventNames.TASKS_SUBMITTED, "tasksCount", Integer.toString(workUnits.size()));
+        submit(gobblin.metrics.event.EventNames.TASKS_SUBMITTED, "tasksCount", Integer.toString(workUnits.size()));
```
liyinan926: Why not import `gobblin.metrics.event.EventNames`?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -392,4 +396,22 @@ private boolean inMultipleBranches(List<Boolean> branches) {
...
+   * @param converter {@link gobblin.converter.Converter} used by this task.
+   */
+  private void addConstructsFinalStateToTaskState(InstrumentedExtractorBase<?, ?> extractor,
+      Converter<?, ?, ?, ?> converter) {
+    this.taskState.addFinalConstructState(Constructs.EXTRACTOR.toString().toLowerCase(), extractor.getFinalState());
+    this.taskState.addFinalConstructState(Constructs.CONVERTER.toString().toLowerCase(), converter.getFinalState());
+    int forkIdx = 0;
+    for(Optional<Fork> fork : this.forks) {
+      if(fork.isPresent()) {
+        this.taskState.addFinalConstructState("fork." + forkIdx, fork.get().getFinalState());
```
liyinan926: Please use `Constructs.FORK_OPERATOR`.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -94,4 +95,15 @@ public FsDataWriter(State properties, String fileName, int numBranches, int bran
...
     }
   }
+
+  @Override
+  public State getFinalState() {
+    State state = new State();
+    state.setProp("RecordsWritten", recordsWritten());
+    try {
+      state.setProp("BytesWritten", bytesWritten());
+    } catch(IOException ioe) {
```
liyinan926: Please add some comments within this catch block on why this is ignored.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
ibuenros: This is not a FORK_OPERATOR, it is a Fork, which is not part of gobblin-core. I will add the constant somewhere else though.

gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedDataWriterDecorator.java
ibuenros: If embedded writer implements FinalState, then we return the actual final state of the embedded object. However, DataWriter in general doesn't implement FinalState, so we return super in that case (which is just empty).

gobblin-core/src/main/java/gobblin/instrumented/extractor/InstrumentedExtractorDecorator.java
ibuenros: Same as above.

ibuenros:
@liyinan926  Addressed comments.

PR #208, opened 2015-07-10
Fixing bugs in MultiWorkUnit

sahilTakiar:
Since the `MultiWorkUnit` class extends `WorkUnit` it is possible for a user to execute the following code:

```
WorkUnit workUnit = new MultiWorkUnit(Lists.newArrayList(new WorkUnit(), new WorkUnit());
workUnit.setProp("key1", "value1");
```

In the current implementation of `MultiWorkUnit` the `setProp("key1", "value1")` method will add the k, v pair to the internal properties object of the `MultiWorkUnit`, but it won't add the property to the `WorkUnit`s in the internal list.

The current implementation / usage has a few issues:

1: The k, v pair gets dropped when a `MultiWorkUnit` gets serialized.
2: The `JobLauncherUtils.flattenWorkUnits` method will also drop all these properties.

The approach I took to remedy this is to override the `setProp` method to invoke `super.setProp()` and to invoke `setProp()` on each individual `WorkUnit` in the internal list. Also, to fix the serialization bug.

I find it a little confusing that now a `MultiWorkUnit` can hold state in addition to a list of `WorkUnit`s so I am open to suggestions for a better approach.

I consider (and briefly talked to Yinan about this today), just overriding `setProp` and `getProp` with implementations that throw `UnsupportedOperationException`, but realized that would require overriding at least 5 methods. It would also be against Object Oriented Design, as `MultiWorkUnit` extends `State` and thus should be capable of storing and retrieving k, v pairs.

@liyinan926 and @ibuenros any comments or suggestions?


liyinan926:
Travis build failed because line 408 `workUnit.setProp(ConfigurationKeys.TASK_ID_KEY, multiTaskId);` caused all wrapped `WorkUnit`s to have the same task ID, given the changes in this PR. So you need to fix that. You can remove that line and use `workUnit.getId()` in line 412 instead of reading the value of `ConfigurationKeys.TASK_ID_KEY`.

sahilTakiar:
I removed lines 405-411, I'm not sure if setting a property on a `MultiWorkUnit` actually does anything in the original implementation, since the `MultiWorkUnit` is serialized right in the next code block, the 'workUnit.setProp(ConfigurationKeys.TASK_ID_KEY, multiTaskId);' property will get dropped during serialization.

sahilTakiar:
Ignore the most recent changes please

sahilTakiar:
I messed up the commit history, will make a new PR.

PR #210, opened 2015-07-10
Added HadoopFileInputSource and HadoopFileInputExtractor

liyinan926:
1. `HadoopFileInputSource` and `HadoopFileInputExtractor` for new Hadoop APIs. Also added a `HadoopTextInputSource` mostly for example and testing.
2. `OldApiHadoopFileInputSource` and `OldApiHadoopFileInputExtractor` for old Hadoop APIs. Also added a `OldApiHadoopTextInputSource` mostly for example and testing.
3. Fixed a concurrency bug in `ParallelRunnerTest`.
 
### DIFF
--- a/gobblin-core/build.gradle
+++ b/gobblin-core/build.gradle
@@ -41,8 +41,11 @@ dependencies {
 
   if (project.hasProperty('useHadoop2')) {
     compile externalDependency.avroMapredH2
+    compile externalDependency.hadoopClientCore
+    compile externalDependency.hadoopAnnotations
   } else {
     compile externalDependency.avroMapredH1
+    compile externalDependency.hadoop
   }
 
   testCompile externalDependency.testng
--- a/gobblin-core/build.gradle
+++ b/gobblin-core/build.gradle
@@ -41,11 +41,8 @@ dependencies {
 
   if (project.hasProperty('useHadoop2')) {
     compile externalDependency.avroMapredH2
-    compile externalDependency.hadoopClientCore
-    compile externalDependency.hadoopAnnotations
   } else {
     compile externalDependency.avroMapredH1
-    compile externalDependency.hadoop
   }
 
   testCompile externalDependency.testng
--- a/gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputExtractor.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.source.extractor.hadoop;
+
+import java.io.IOException;
+
+import org.apache.hadoop.mapreduce.RecordReader;
+
+import gobblin.source.extractor.DataRecordException;
+import gobblin.source.extractor.Extractor;
+
+
+/**
+ * An implementation of {@link Extractor} that uses a Hadoop {@link RecordReader} to read records
+ * from a {@link org.apache.hadoop.mapreduce.lib.input.FileSplit}.
+ *
+ * <p>
+ *   This class can read either keys of type {@link #<K>} or values of type {@link #<V>} using the
+ *   given {@link RecordReader}, depending on the value of the second argument of the constructor
+ *   {@link #HadoopFileInputExtractor(RecordReader, boolean)}. It will read keys if the argument
+ *   is {@code true}, otherwise it will read values. Normally, this is specified using the property
+ *   {@link HadoopFileInputSource#FILE_INPUT_READ_KEYS_KEY}, which is {@code false} by default.
+ * </p>
+ *
+ * <p>
+ *   This class provides a default implementation of {@link #readRecord(Object)} that simply casts
+ *   the keys or values read by the {@link RecordReader} into type {@link #<D>}. It is required
+ *   that type {@link #<K>} or {@link #<V>} can be safely casted to type {@link #<D>}.
+ * </p>
+ *
+ * <p>
+ *   The Hadoop {@link RecordReader} is passed into this class, which is responsible for closing
+ *   it by calling {@link RecordReader#close()} in {@link #close()}.
+ * </p>
+ *
+ * <p>
+ *   A concrete implementation of this class should at least implement the {@link #getSchema()}
+ *   method.
+ * </p>
+ *
+ * @param <S> output schema type
+ * @param <D> output data record type that MUST be compatible with either {@link #<K>} or {@link #<V>}
+ * @param <K> key type expected by the {@link RecordReader}
+ * @param <V> value type expected by the {@link RecordReader}
+ *
+ * @author ynli
+ */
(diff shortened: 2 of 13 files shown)


liyinan926:

zliu41:
Why do we need to have a source and extractor for the old API?
If we must have them, I suggest using `OldApiHadoopFileInputSource` for old API, and use `HadoopFileInputSource` for the new API since new API should be used much more often.

liyinan926:
The old APIs are still widely used in the community so it's good to support them. Yes, I can make the naming changes.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputExtractor.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author ynli
+ */
+public abstract class HadoopFileInputExtractor<S, D, K, V> implements Extractor<S, D> {
+
+  private final RecordReader<K, V> recordReader;
+  private final boolean readKeys;
+
+  public HadoopFileInputExtractor(RecordReader<K, V> recordReader, boolean readKeys) {
+    this.recordReader = recordReader;
+    this.readKeys = readKeys;
```
zliu41: Should add a check here whether K or V is a subtype of D?

liyinan926:
Updated with the naming changes.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputExtractor.java
liyinan926: I believe there's no way to check that at compile time on type parameters. At runtime, an exception will be thrown if casting fails.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/OldApiHadoopFileInputSource.java
```diff
@@ -0,0 +1,212 @@
...
+      closer.close();
+    }
+  }
+
+  private FileSplit deserializeFileSplit(String fileSplitBytesStr) throws IOException  {
+    Closer closer = Closer.create();
+    try {
+      byte[] fileSplitBytes = BaseEncoding.base64().decode(fileSplitBytesStr);
+      ByteArrayInputStream byteArrayInputStream = closer.register(new ByteArrayInputStream(fileSplitBytes));
+      DataInputStream dataInputStream = new DataInputStream(byteArrayInputStream);
```
zliu41: `dataInputStream` is not registered to `closer`.

liyinan926:
Addressed comments and fixed build after the naming changes.

zliu41:
LGTM

gobblin-core/build.gradle
```diff
@@ -41,8 +41,11 @@ dependencies {
 
   if (project.hasProperty('useHadoop2')) {
     compile externalDependency.avroMapredH2
+    compile externalDependency.hadoopClientCore
+    compile externalDependency.hadoopAnnotations
   } else {
     compile externalDependency.avroMapredH1
+    compile externalDependency.hadoop
```
sahilTakiar: Are these hadoop additions necessary? The top level build.gradle all expresses these dependencies for each subproject

liyinan926: Yeah, you are right. Those are not needed.

liyinan926: Fixed.

liyinan926:
Updated. Also added a prefix to config key names so there's less chance of naming conflicts.

zliu41:
LGTM

liyinan926:
@sahilTakiar any more comments?

sahilTakiar:
Yes, sorry I will take a look today.

liyinan926:
@sahilTakiar it would be great if you can take a look asap, as I need this to be checked in for the Lynda source. Thanks!

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputSource.java
```diff
@@ -0,0 +1,258 @@
...
+public abstract class HadoopFileInputSource<S, D, K, V> implements Source<S, D> {
+
+  private static final String HADOOP_SOURCE_KEY_PREFIX = "source.hadoop.";
+  public static final String FILE_INPUT_FORMAT_CLASS_KEY = HADOOP_SOURCE_KEY_PREFIX + "file.input.format.class";
+  public static final String FILE_SPLITS_DESIRED_KEY = HADOOP_SOURCE_KEY_PREFIX + "file.splits.desired";
+  public static final int DEFAULT_FILE_SPLITS_DESIRED = 1;
+  public static final String FILE_INPUT_PATHS_KEY = HADOOP_SOURCE_KEY_PREFIX + "file.input.paths";
+  public static final String FILE_INPUT_READ_KEYS_KEY = HADOOP_SOURCE_KEY_PREFIX + "file.read.keys";
+  public static final boolean DEFAULT_FILE_INPUT_READ_KEYS = false;
+  static final String FILE_SPLIT_BYTES_STRING_KEY = HADOOP_SOURCE_KEY_PREFIX + "file.split.bytes.string";
```
sahilTakiar: Should this be public also?

liyinan926: This is only used internally and is not available for the users to set, and it is not used outside the package.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/OldApiHadoopFileInputSource.java
```diff
@@ -0,0 +1,214 @@
...
+   * @param recordReader a Hadoop {@link RecordReader} object used to read input records
+   * @param fileSplit the {@link FileSplit} to read input records from
+   * @param readKeys whether the {@link OldApiHadoopFileInputExtractor} should read keys of type {@link #<K>};
+   *                 by default values of type {@link #>V>} are read.
+   * @return a {@link OldApiHadoopFileInputExtractor} instance
+   */
+  protected abstract OldApiHadoopFileInputExtractor<S, D, K, V> getExtractor(WorkUnitState workUnitState,
+      RecordReader<K, V> recordReader, FileSplit fileSplit, boolean readKeys);
+
+  private String serializeFileSplit(FileSplit fileSplit) throws IOException {
```
sahilTakiar: Could you just move this method and `deserializeFileSplit` into a utility class, and have the parameter be of type `Writable`. These methods should work for any custom `Writable` class that needs to be serialized into a string.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputSource.java
```diff
@@ -0,0 +1,258 @@
...
+    } finally {
+      closer.close();
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private TaskAttemptContext getTaskAttemptContext(Configuration configuration, TaskAttemptID taskAttemptID) {
+    Class<?> taskAttemptContextClass;
+
+    // This is a method learned from Spark (See "org.apache.spark.mapreduce.SparkHadoopMapReduceUtil").
```
sahilTakiar: In this necessary given that the PR already has a hadoop 1.x and hadoop 2.x compatible version.

sahilTakiar:
Just a few comments, otherwise LGTM.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFileInputSource.java
liyinan926: Yes, this is necessary. The same new Hadoop API is available in Hadoop 1 and 2, but the implementation is different between Hadoop 1 and 2.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/OldApiHadoopFileInputSource.java
liyinan926: Yes, good point, will do.

liyinan926:
Updated.

sahilTakiar:
LGTM

PR #211, opened 2015-07-10
Fixing bugs in MultiWorkUnit

sahilTakiar:
Duplicate of https://github.com/linkedin/gobblin/pull/208

@liyinan926 I addressed your comments.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -402,16 +402,15 @@ public class MRJobLauncher extends AbstractJobLauncher {
       // Serialize each work unit into a file named after the task ID
       for (WorkUnit workUnit : workUnits) {
 
-        Path workUnitFileExt;
+        String workUnitFileName;
         if (workUnit instanceof MultiWorkUnit) {
-          workUnitFileExt =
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -401,15 +401,18 @@ private Path prepareJobInput(Path jobInputPath, List<WorkUnit> workUnits) throws
       int multiTaskIdSequence = 0;
       // Serialize each work unit into a file named after the task ID
       for (WorkUnit workUnit : workUnits) {
+
+        Path workUnitFileExt;
```
liyinan926: You don't need even to make a `Path`, a `String` is sufficient. Also please rename it to `workUnitFileName`.

liyinan926:
LGTM other than the only comment.

sahilTakiar:
@ibuenros any comments?

ibuenros:
It looks good. Is there any use case where we would want to set a property in the MultiWorkUnit, but not for it to spread to the WorkUnits? If so, we could add a separate method to allow for that, but I can't  think of any such use case.

sahilTakiar:
I can't think of a use case either, but it may be nice to have, I can add such a method.

sahilTakiar:
Updated

ibuenros:
LGTM

PR #214, opened 2015-07-14
Separate SourceState/JobState from workUnits

zliu41:
Summary:
- Deprecated all constructors of `WorkUnit`, and added some factory methods, which do not add job states to workunits. 
The reason to use factory methods is to avoid having too many constructors for `WorkUnit`, some deprecated and some not deprecated.
- Deprecated the constructor of `Extract` that takes `SourceState`, added a new constructor without `SourceState`.
- In `MRJobLauncher.prepareHadoopJob()`, serialize `jobState` into a file. Then in `TaskRunner.setup()`, read `jobState` from the file and add it to each `workUnit`.
- In `LocalJobLauncher.runWorkUnits()`, add `jobState` to each `workUnit`.
- In `MRJobLauncher.runWorkUnits()`, combine `jobState` with each `workUnit` into a `fatWorkUnit`, since properties from both `jobState` and `workUnit` are needed in order to invoke `JobLauncherUtils.cleanStagingData()`.
- Modified `KafkaSource` to use the new paradigm.

Tested with Kafka adapter. Existing source implementations still work.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -121,6 +121,7 @@ public class ConfigurationKeys {
   public static final String JOB_TRACKING_URL_KEY = "job.tracking.url";
   public static final String FORK_STATE_KEY = "fork.state";
   public static final String METRIC_CONTEXT_NAME_KEY = "metrics.context.name";
+  public static final String JOB_STATE_FILE_PATH_KEY = "job.state.file.path";
 
   /**
    * Work unit related configuration properties.
--- a/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
@@ -49,7 +49,7 @@ import gobblin.source.workunit.Extract;
  */
 public class SourceState extends State {
 
-  private static final Set<Extract> extractSet = Sets.newConcurrentHashSet();
+  private static final Set<Extract> EXTRACT_SET = Sets.newConcurrentHashSet();
   private static final DateTimeFormatter DTF =
       DateTimeFormat.forPattern("yyyyMMddHHmmss").withLocale(Locale.US).withZone(DateTimeZone.UTC);
 
@@ -107,7 +107,7 @@ public class SourceState extends State {
    * @return (possibly empty) list of {@link WorkUnitState}s from the previous job run
    */
   public List<WorkUnitState> getPreviousWorkUnitStates() {
-    return ImmutableList.<WorkUnitState>builder().addAll(this.previousWorkUnitStates).build();
+    return ImmutableList.<WorkUnitState> builder().addAll(this.previousWorkUnitStates).build();
   }
 
   /**
@@ -121,10 +121,13 @@ public class SourceState extends State {
    * @param namespace namespace of the table this extract belongs to
    * @param table name of the table this extract belongs to
    * @return a new unique {@link Extract} instance
+   *
+   * @Deprecated Use {@link gobblin.source.extractor.extract.AbstractSource#createExtract(gobblin.source.workunit.Extract.TableType, String, String)}
    */
+  @Deprecated
   public synchronized Extract createExtract(Extract.TableType type, String namespace, String table) {
     Extract extract = new Extract(this, type, namespace, table);
-    while (extractSet.contains(extract)) {
+    while (EXTRACT_SET.contains(extract)) {
       if (Strings.isNullOrEmpty(extract.getExtractId())) {
         extract.setExtractId(DTF.print(new DateTime()));
       } else {
@@ -132,7 +135,7 @@ public class SourceState extends State {
         extract.setExtractId(DTF.print(extractDateTime.plusSeconds(1)));
       }
     }
-    extractSet.add(extract);
+    EXTRACT_SET.add(extract);
     return extract;
   }
 
@@ -141,7 +144,11 @@ public class SourceState extends State {
    *
    * @param extract given {@link Extract}
    * @return a new {@link WorkUnit} instance
+   *
+   * @deprecated Properties in SourceState should not added to a WorkUnit. Having each WorkUnit contain a copy of
+   * SourceState is a waste of memory. Use {@link WorkUnit#create(Extract)}.
    */
+  @Deprecated
   public WorkUnit createWorkUnit(Extract extract) {
     return new WorkUnit(this, extract);
   }
@@ -173,8 +180,8 @@ public class SourceState extends State {
     }
 
     SourceState other = (SourceState) object;
-    return super.equals(other) && this.previousSourceState.equals(other.previousSourceState) &&
-        this.previousWorkUnitStates.equals(other.previousWorkUnitStates);
+    return super.equals(other) && this.previousSourceState.equals(other.previousSourceState)
+        && this.previousWorkUnitStates.equals(other.previousWorkUnitStates);
   }
 
   @Override
--- a/gobblin-api/src/main/java/gobblin/source/workunit/Extract.java
+++ b/gobblin-api/src/main/java/gobblin/source/workunit/Extract.java
@@ -62,7 +62,11 @@ public class Extract extends State {
    * @param namespace dot separated namespace path
    * @param type {@link TableType}
    * @param table table name
+   *
+   * @deprecated Extract does not use any property in {@link SourceState}.
+   * Use {@link #Extract(TableType, String, String)}
    */
+  @Deprecated
   public Extract(SourceState state, TableType type, String namespace, String table) {
     // Values should only be null for deserialization
     if (state != null && type != null && !Strings.isNullOrEmpty(namespace) && !Strings.isNullOrEmpty(table)) {
@@ -81,11 +85,22 @@ public class Extract extends State {
       }
 
       // Setting full drop date if not already specified, the value can still be overridden if required.
-      if (state.getPropAsBoolean(ConfigurationKeys.EXTRACT_IS_FULL_KEY) && !state
-          .contains(ConfigurationKeys.EXTRACT_FULL_RUN_TIME_KEY)) {
+      if (state.getPropAsBoolean(ConfigurationKeys.EXTRACT_IS_FULL_KEY)
+          && !state.contains(ConfigurationKeys.EXTRACT_FULL_RUN_TIME_KEY)) {
         super.setProp(ConfigurationKeys.EXTRACT_FULL_RUN_TIME_KEY, System.currentTimeMillis());
       }
     }
+  }
+
+  /**
+   * Constructor.
+   *
+   * @param type {@link TableType}
+   * @param namespace dot separated namespace path
+   * @param table table name
+   */
+  public Extract(TableType type, String namespace, String table) {
+    this(new SourceState(), type, namespace, table);
   }
 
   /**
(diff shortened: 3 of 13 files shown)


zliu41:
@liyinan926 @chavdar please review.

liyinan926:
Seems all `MRJobLauncherTest` tests are failing. @zliu41 can you first resolve those?

zliu41:
@liyinan926 updated. Since `WriterUtils.getDefaultWriterFilePath` requires the input `State` object to be either `WorkUnit` or `WorkUnitState`, I needed to change `State combinedState` to `WorkUnit fatWorkUnit`.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -74,7 +74,7 @@ public SourceState(State properties, List<WorkUnitState> previousTaskStates) {
    * @return (possibly empty) list of {@link WorkUnitState}s from the previous job run
    */
   public List<WorkUnitState> getPreviousWorkUnitStates() {
-    return ImmutableList.<WorkUnitState>builder().addAll(this.previousTaskStates).build();
+    return ImmutableList.<WorkUnitState> builder().addAll(this.previousTaskStates).build();
```
liyinan926: Empty space introduced.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -104,14 +104,17 @@ public synchronized Extract createExtract(Extract.TableType type, String namespa
    *
    * @param extract given {@link Extract}
    * @return a new {@link WorkUnit} instance
+   *
+   * @deprecated Properties in SourceState should not added to a WorkUnit. Having each WorkUnit contain a copy of
+   * SourceState is a waste of memory. Use WorkUnit.create(Extract extract).
```
liyinan926: Use `Use {@link WorkUnit#create(Extract)}`.

gobblin-api/src/main/java/gobblin/source/workunit/Extract.java
```diff
@@ -62,7 +62,10 @@
    * @param namespace dot separated namespace path
    * @param type {@link TableType}
    * @param table table name
+   *
+   * @deprecated Extract does not use any property in SourceState. Use Extract(TableType, namespace, table).
```
liyinan926: Ditto.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -76,7 +83,11 @@ public WorkUnit(SourceState state, Extract extract) {
    * @param state a {@link gobblin.configuration.SourceState} the properties of which will be copied into this {@link WorkUnit} instance.
    * @param extract an {@link Extract}.
    * @param watermarkInterval a {@link WatermarkInterval} which defines the range of data this {@link WorkUnit} will process.
+   *
+   * @deprecated Properties in SourceState should not be added to a WorkUnit. Having each WorkUnit contain a copy of
+   * SourceState is a waste of memory. Use WorkUnit.create(Extract extract, WatermarkInterval watermarkInterval).
```
liyinan926: Ditto.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -56,7 +59,11 @@ public WorkUnit() {
    *
    * @param state a {@link SourceState} the properties of which will be copied into this {@link WorkUnit} instance
    * @param extract an {@link Extract}
+   *
+   * @deprecated Properties in SourceState should not be added to a WorkUnit. Having each WorkUnit contain a copy of
+   * SourceState is a waste of memory. Use WorkUnit.create(Extract extract).
```
liyinan926: Ditto.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -95,13 +106,52 @@ public WorkUnit(SourceState state, Extract extract, WatermarkInterval watermarkI
...
+   * @return An empty WorkUnit.
+   */
+  public static WorkUnit createEmpty() {
+    return new WorkUnit();
+  }
+
+  /**
+   * Facotry method.
+   *
+   * @return A WorkUnit with the given Extract.
```
liyinan926: Please use `{@link ...}` for class references.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -557,6 +582,26 @@ protected void setup(Context context) {
       }
     }
 
+    private JobState readJobState(Context context) throws IOException {
+      JobState jobState = new JobState();
+      if (context.getConfiguration().get(ConfigurationKeys.JOB_STATE_FILE_KEY) == null) {
```
liyinan926: You can use a `Preconditions.checkArgument()` here.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -557,6 +582,26 @@ protected void setup(Context context) {
       }
     }
 
+    private JobState readJobState(Context context) throws IOException {
+      JobState jobState = new JobState();
```
liyinan926: You can move this to after the precondition check.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -121,6 +121,7 @@
   public static final String JOB_TRACKING_URL_KEY = "job.tracking.url";
   public static final String FORK_STATE_KEY = "fork.state";
   public static final String METRIC_CONTEXT_NAME_KEY = "metrics.context.name";
+  public static final String JOB_STATE_FILE_KEY = "job.state.file";
```
liyinan926: `job.state.file.path` is a better name.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
liyinan926: Incomplete Javadoc.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -95,13 +106,52 @@ public WorkUnit(SourceState state, Extract extract, WatermarkInterval watermarkI
...
+   */
+  public static WorkUnit create(Extract extract) {
+    return new WorkUnit(null, extract);
+  }
+
+  /**
+   * Facotry method.
+   *
```
liyinan926: Ditto. Incomplete Javadoc.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
liyinan926: You can actually change the member variable to `private final JobState jobState = new JobState()` and then just deserialize into `this.jobState`.

liyinan926:
There's still one potential issue with this approach. If someone sets anything in the `workUnitstate` that is supposed to overwrite the static config in the method `Source.getExtractor(WorkUnitState workUnitstate)`, the changes will be lost when `JobState` gets added back.

zliu41:
@liyinan926 updated.
- Added method `State.addAllIfNotExist()`, so that when combining `jobState` and `workUnit`, `jobState` doesn't override `workUnit`.
- Added method `Extract.getUniqueInstance()` which replaces the deprecated method `SourceState.createExtract()`.

sahilTakiar:
I think this may also require changes to `AbstractSource`. The method `getPreviousWorkUnitStatesForRetry` contains some code to use a config `ConfigurationKeys.OVERWRITE_CONFIGS_IN_STATESTORE`. If this config is set to true then all properties in the specified `SourceState` object are added to each `WorkUnitState`. I'm guessing this is something we would want to change too.

zliu41:
Updated.

liyinan926:
LGTM.

gobblin-api/src/main/java/gobblin/source/workunit/Extract.java
```diff
@@ -81,14 +88,47 @@ public Extract(SourceState state, TableType type, String namespace, String table
...
+
+  /**
+   * Factory method. Returns a unique {@link Extract} instance.
+   *
+   * @param type {@link TableType}
+   * @param namespace dot separated namespace path
+   * @param table table name
+   * @return a unique {@link Extract} instance
+   */
+  public static Extract getUniqueInstance(TableType type, String namespace, String table) {
```
chavdar: This looks like a copy/paste of gobblin.configuration.SourceState.createExtract(TableType, String, String) which is not good. We need to remove this function and that static CREATED_INSTANCES object. The extract creation should be owned by the Source.

1) Please move the CREATED_INSTANCES map and this method into a new interface/class ExtractFactory/ExtractFactoryImpl or something along those lines. 

2) Have AbstractSource create/own an instance of ExtractFactoryImpl.

3) Have the KafkaSource use the factory provided by AbstractSource

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -89,10 +89,13 @@ public SourceState(State properties, List<WorkUnitState> previousTaskStates) {
    * @param namespace namespace of the table this extract belongs to
    * @param table name of the table this extract belongs to
    * @return a new unique {@link Extract} instance
+   *
+   * @Deprecated Use {@link Extract#getUniqueInstance(Extract.TableType, String, String)}
```
chavdar: This should be deprecated in favor of the AbstractSource ExtractFactoryImpl.

chavdar: For now, you can also make EXTRACT_SET and ExtractFactoryImpl and reuse the implementation in createExtract to avoid code duplication.

zliu41:
@chavdar updated.

gobblin-api/src/main/java/gobblin/source/workunit/ExtractFactory.java
```diff
@@ -0,0 +1,49 @@
...
+   * Returns a unique {@link Extract} instance.
+   * Any two calls of this method from the same {@link ExtractFactory} instance guarantees to
+   * return {@link Extract}s with different IDs.
+   *
+   * @param type {@link TableType}
+   * @param namespace dot separated namespace path
+   * @param table table name
+   * @return a unique {@link Extract} instance
+   */
+  public Extract getUniqueExtract(TableType type, String namespace, String table) {
```
chavdar: This needs to be synchronized and the createdInstances need not. There are race conditions in the current implementation.

gobblin-api/src/main/java/gobblin/source/workunit/ExtractFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.common.collect.Sets;
+
+import gobblin.source.workunit.Extract.TableType;
+
+
+public class ExtractFactory {
+  private final Set<Extract> createdInstances;
+  private final DateTimeFormatter dtf;
+
+  public ExtractFactory(String dateTimeFormat) {
```
chavdar: Please add some unit tests for this new class.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -49,7 +49,7 @@
  */
 public class SourceState extends State {
 
-  private static final Set<Extract> extractSet = Sets.newConcurrentHashSet();
+  private static final Set<Extract> EXTRACT_SET = Sets.newConcurrentHashSet();
```
chavdar: Why don't make use ExtractFactory?

zliu41: `SourceState.createExtract()` invokes a deprecated constructor of `Extract` which adds properties in `SourceState` to the created `Extract`. `ExtractFactory` doesn't do this. Changing this may break existing jobs.

zliu41:
Updated

chavdar:
LGTM.

PR #216, opened 2015-07-14
Adding a DummyExtractor

sahilTakiar:
Adding a `DummyExtractor` which always returns 0 records. The use is for IDPC where we are trying to implement on exponential backoff for the retrying of `WorkUnit`s. There will be some executions where we want a `WorkUnit` to be serialized to the state-store, but we don't want the `WorkUnit` to actually do anything, in which case the `DummyExtractor` becomes useful.


liyinan926:
Missing open source header, otherwise LGTM.

PR #217, opened 2015-07-14
ETL-3057: fix for file not exist during mapper task

ydailinkedin:
Added checks for file existence in extractor.
Another change is in AvroFsHelper. Added method updateFileSystem, which basically makes the fs creation configurable when proxy is enabled. 
Tested on Nertz.
@sahilTakiar @liyinan926 can you review? Thanks!


gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFileExtractor.java
```diff
@@ -36,10 +37,12 @@ public AvroFileExtractor(WorkUnitState workUnitState) {
   }
 
   @Override
-  public Iterator<GenericRecord> downloadFile(String file)
-      throws IOException {
+  public Iterator<GenericRecord> downloadFile(String file) throws IOException {
     try {
-      return this.closer.register(((AvroFsHelper) this.fsHelper).getAvroFile(file));
```
sahilTakiar: I think `Closer.register` can handles `null`s

http://grepcode.com/file/repo1.maven.org/maven2/com.google.guava/guava/18.0/com/google/common/io/Closer.java#Closer.register%28java.io.Closeable%29

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -62,41 +62,54 @@ public FileSystem getFileSystem() {
     return this.fs;
   }
 
+  private void updateFileSystem(String uri) throws IOException, InterruptedException, URISyntaxException {
```
sahilTakiar: Consider changing the method name from `updateFileSystem` to `createFileSystem`

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -62,41 +62,54 @@ public FileSystem getFileSystem() {
...
+  public void close() throws FileBasedHelperException {
+    /*
+     * TODO
+     * Removing this for now, FileSystem.get() returns the same object each time it is called within a process
+     * If this method gets called in parallel across tasks it is going to cause problems
+     * Basically, close cannot be called multiple times within a process
+     * http://stackoverflow.com/questions/17421218/multiples-hadoop-filesystem-instances
+     */
+
+    //        try {
```
liyinan926: Please remove those lines as this is exactly what git provides: maintain code history so you won't lose it.

ydailinkedin:
Updated.

liyinan926:
LGTM.

sahilTakiar:
LGTM

PR #218, opened 2015-07-14
Create empty workunits for blacklisted topic partitions

zliu41:
Added a `HashSet` in `KafkaSource` that contains the processed partitions. If a partition has a previous offset from the state store, but is not processed, an empty workunit will be created for this partition to avoid losing the checkpoint.


zliu41:
@liyinan926 please review.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -99,6 +99,8 @@ public int compare(WorkUnit w1, WorkUnit w2) {
   private final Map<KafkaPartition, Long> previousOffsets = Maps.newHashMap();
   private final Map<KafkaPartition, Long> previousAvgSizes = Maps.newHashMap();
 
+  private final Set<KafkaPartition> processedPartitions = Sets.newHashSet();
```
liyinan926: I think it's more like `to-be-processed` partitions.

liyinan926:
LGTM other than the only comment.

PR #219, opened 2015-07-14
Fixed a bug in SourceState.createExtract

liyinan926:

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
@@ -24,6 +24,7 @@ import org.joda.time.DateTimeZone;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


liyinan926:
@zliu41 can you review?

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -92,8 +93,12 @@ public SourceState(State properties, List<WorkUnitState> previousTaskStates) {
   public synchronized Extract createExtract(Extract.TableType type, String namespace, String table) {
     Extract extract = new Extract(this, type, namespace, table);
     while (extractSet.contains(extract)) {
```
zliu41: Seems `extractSet` is not used anywhere. Why is this needed?
Also this method will be deprecated in #214 since properties in `SourceState` do not need to be added to `Extract`.

liyinan926: `extractSet` is used here to make sure that each `Extract` has a unique `id`. I need this to unblock the Lynda gobblin job testing.

zliu41:
LGTM

PR #220, opened 2015-07-14
ETL-3057: fix for file not exist during mapper task

ydailinkedin:
This is duplicate as #217 . I messed up with my git squash previously. Now fixed.


liyinan926:
LGTM.

PR #224, opened 2015-07-15
Fixing NPE when Extractor.getSchema returns null

sahilTakiar:
Also made some formatting changes to the State class.

@liyinan926 could you review this?


gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -328,7 +328,7 @@ public void close() throws IOException {
         .writeTo(Destination.of(this.taskContext.getDestinationType(this.branches, this.index), this.taskState))
         .writeInFormat(this.taskContext.getWriterOutputFormat(this.branches, this.index))
         .withWriterId(this.taskId)
-        .withSchema(this.convertedSchema)
+        .withSchema(this.convertedSchema.get())
```
liyinan926: This will throw an NPE if `convertedSchema` is absent.

sahilTakiar:
Updated

liyinan926:
LGTM.

PR #225, opened 2015-07-15
Added a per-partition TimeBasedLimiter to KafkaExtractor

xkrogen:
Added a per-partition TimeBasedLimiter to KafkaExtractor to allow it to limit how long it spends on any one partition, since a long-running partition will make all other partitions wait for its completion. This required moving the Limiters into the gobblin-core module.


xkrogen:
@zliu41 Can you review? I'm not 100% sure about the new location of the Limiter classes that had to be moved so let me know if they're incorrect

zliu41:
I just realized that we can simply turn-on a per-mapper time limiter by setting the following properties
`extract.limit.enabled`
`extract.limit.type`
`extract.limit.time.limit`
`extract.limit.time.limit.timeunit`
So I'll see how well the per-mapper time limiter works. If it's more desirable to use a per-partition time limiter, then I'll revisit this PR.

PR #226, opened 2015-07-15
Fixed a potential NPE in WorkUnit.getLowWatermark()

liyinan926:
`WorkUnit.getLowWatermark()` assumes that the key `watermark.interval.value` is set, which seems only happening in the Kafka adapter. If this key is not set, `getLowWatermark` will throw an NPE if it is called, which is called in `WorkUnitState.backoffActualHighWatermark`.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -164,7 +164,11 @@ public class MRJobLauncher extends AbstractJobLauncher {
         this.job.killJob();
       }
     } finally {
-      super.close();
+      try {
+        cleanUpWorkingDirectory();
+      } finally {
+        super.close();
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


liyinan926:
@zliu41 can you review?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -164,7 +164,11 @@ public void close() throws IOException {
         this.job.killJob();
       }
     } finally {
-      super.close();
+      try {
+        cleanUpWorkingDirectory();
```
zliu41: `runWorkUnits()` also calls `cleanUpWorkingDirectory()`, why need both?

liyinan926: `runWorkUnits` may not get a chance to be called if the logic fails before it. `close` is guaranteed to be called.

zliu41:
LGTM

PR #228, opened 2015-07-16
Made the previous SourceState available in the current run

liyinan926:
Currently only the previous `WorkUnitState`s are available in the current job run through `SourceState.getPreviousWorkUnitStates`. The list of previous `WorkUnitStates` are from the previous `SourceState` (`JobState` more precisely). The previous `SourceState` itself, however, is not available for the `Source` to use. This PR fixed this so the previous `SourceState` is also available through `SourceState.getPreviousSourceState`.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ImmutableWorkUnitState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ImmutableWorkUnitState.java
@@ -10,46 +10,45 @@
  * CONDITIONS OF ANY KIND, either express or implied.
  */
 
-package gobblin.source.workunit;
+package gobblin.configuration;
 
 import java.io.DataInput;
 import java.io.IOException;
 import java.util.Properties;
 
-import gobblin.configuration.State;
+import gobblin.source.extractor.Watermark;
 
 
 /**
- * An immutable version of {@link WorkUnit}.
+ * An immutable version of {@link WorkUnitState}.
  *
  * @author ynli
  */
-public class ImmutableWorkUnit extends WorkUnit {
+public class ImmutableWorkUnitState extends WorkUnitState {
 
-  public ImmutableWorkUnit(WorkUnit workUnit) {
-    super(workUnit.getExtract());
-    super.addAll(workUnit);
+  public ImmutableWorkUnitState(WorkUnitState workUnitState) {
+    super(workUnitState.getWorkunit());
+    super.addAll(workUnitState);
+  }
+
+  @Override
+  public void setWorkingState(WorkingState state) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void setActualHighWatermark(Watermark watermark) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Deprecated
+  @Override
+  public void setHighWaterMark(long value) {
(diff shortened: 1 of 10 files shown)


liyinan926:

sahilTakiar:
LGTM

It looks like the goal of `getPreviousSourceState` is to return a `SourceState` that cannot be modified, which I think is the correct approach. However, `getPreviousWorkUnitStates` only returns an `ImmutableList` of `WorkUnitState`s. While the list itself cannot be modified there is nothing stopping a user from modifying the `WorkUnitState`s inside the list. Perhaps we should consider creating an `ImmutableState` class, or just migrating to something like commons-configuration which already provides these types of things.

liyinan926:
Good point, I will create an `ImmutableWorkUnitState` class.

zliu41:
If every `SourceState` also contains a copy of the previous `SourceState`, then will the size of `SourceState` grow infinitely?

liyinan926:
No. Note that `previousSourceState` is not persisted. So when the previous `SourceState` is read from the state store, it does not contain the `SourceState` of the run before the previous run.

liyinan926:
Updated to add a `ImmutableWorkUnitState` class and use it in `SourceState`.

gobblin-api/src/main/java/gobblin/configuration/ImmutableWorkUnitState.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * An immutable version of {@link WorkUnitState}.
+ *
+ * @author ynli
+ */
+public class ImmutableWorkUnitState extends WorkUnitState {
+
+  public ImmutableWorkUnitState(WorkUnitState workUnitState) {
+    super(workUnitState.getWorkunit());
```
zliu41: Should be `super(new ImmutableWorkUnit(workUnitState.getWorkunit()))`?

gobblin-api/src/main/java/gobblin/source/workunit/ImmutableWorkUnit.java
```diff
@@ -20,11 +27,49 @@
 public class ImmutableWorkUnit extends WorkUnit {
 
   public ImmutableWorkUnit(WorkUnit workUnit) {
-    super(workUnit);
+    super(workUnit.getExtract());
```
zliu41: Seems the `Extract` object in an `ImmutableWorkUnit` is still mutable. Is this acceptable?

liyinan926: Yes, right. Seems we also need an `ImmutableExtract`.

gobblin-api/src/main/java/gobblin/configuration/ImmutableWorkUnitState.java
liyinan926: Not necessary since `WorkUnitState.getWorkUnit()` returns an `ImmutableWorkUnit`.

liyinan926:
Updated to add an `ImmutableExtract` class.

zliu41:
LGTM

liyinan926:
@sahilTakiar any additional comments?

sahilTakiar:
LGTM

PR #230, opened 2015-07-16
Fixed FindBugs warnings

liyinan926:

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
@@ -25,6 +25,7 @@ import org.joda.time.DateTimeZone;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Strings;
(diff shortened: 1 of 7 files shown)


liyinan926:
@zliu41 can you review?

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -99,7 +98,7 @@ public SourceState(State properties, SourceState previousSourceState, List<WorkU
    * @return {@link SourceState} of the previous job run
    */
   public SourceState getPreviousSourceState() {
-    return new ImmutableSourceState(this.previousSourceState);
+    return new ImmutableSourceState(this.previousSourceState.or(new SourceState()));
```
zliu41: It seems the original approach of using a static `EMPTY_SOURCE_STATE` is more efficient. Why the change?

liyinan926: It was actually doing pre-mature objection creation of the `SourceState` class even before the other members are properly initialized, which is not a good practice.

zliu41:
LGTM

PR #233, opened 2015-07-17
Fixing bug in WorkUnitState.getExtract()

sahilTakiar:
Hey @liyinan926 I noticed that the `getExtract()` method in `WorkUnitState` adds all the props for the current `WorkUnitState` to its `Extract` before returning it. Is this behavior necessary / expected?

It's causing some issues on the IDPC side because now each `Extract` is no longer unique. Properties such as `qualitychecker.rows.written` are showing up in each `Extract` which is a little weird. This only became an issue recently because the `equals()` method of `Extract` has recently been changed so that is does a comparison for all the `Extract`s props.


liyinan926:
LGTM. I am not sure if this is expected, but it is definitely not necessary.

sahilTakiar:
Ok sounds good. Will merge.

PR #234, opened 2015-07-17
Fixed the way addAll(State) is used in some State classes

liyinan926:
Currently, `addAll` is called to add all properties from one `State` object to another. This will cause a `UnsupportedOperationException` if the source `State` object is immutable because the `addAll` of the immutable `State` class will be called. Changed all such occurrences to `super.addAll`.
 

liyinan926:
@sahilTakiar can you help review?

sahilTakiar:
LGTM

PR #235, opened 2015-07-20
Add explicit dependency to commons-cli-1.3 to gobblin-api to override…

chavdar:
… picking up an older version (1.2)


liyinan926:
LGTM.

PR #241, opened 2015-07-21
Avoid duplciate keytab login in cleaning staging data

ydailinkedin:
@sahilTakiar Can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
@@ -23,7 +23,10 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.io.Closer;
 
 import gobblin.configuration.ConfigurationKeys;
@@ -38,6 +41,7 @@ import gobblin.source.workunit.WorkUnit;
  * @author ynli
  */
 public class JobLauncherUtils {
+  private static Map<String, FileSystem> ownerAndFs = Maps.newConcurrentMap();
 
   /**
    * Create a new job ID.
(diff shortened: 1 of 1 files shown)


gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
```diff
@@ -203,4 +218,12 @@ private static ParallelRunner getParallelRunner(FileSystem fs, Closer closer, in
...
     return parallelRunners.get(uri);
   }
+
+  private static synchronized void updateOwnerAndFs(String owner, FileSystem proxiedFs) {
+    ownerAndFs.put(owner, proxiedFs);
+  }
+
+  private static synchronized FileSystem getFsWithProxyOwner(String owner) {
+    return ownerAndFs.get(owner);
+  }
```
sahilTakiar: It may not be necessary to have these two methods since they are both just one line. Also, instead of making these methods synchronized, you can probably just use `Maps.newConcurrentMap`

ydailinkedin:
Updated.

gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
```diff
@@ -179,19 +183,30 @@ public static void cleanStagingData(State state, Logger logger, Closer closer,
...
-            state.getProp(ConfigurationKeys.SUPER_USER_KEY_TAB_LOCATION), writerFsUri);
-      } catch (InterruptedException e) {
-        throw new IOException(e);
-      } catch (URISyntaxException e) {
-        throw new IOException(e);
-      }
-    } else {
       return FileSystem.get(URI.create(writerFsUri), new Configuration());
+    } else {
+      Preconditions.checkArgument(StringUtils.isNotBlank(state.getProp(ConfigurationKeys.FS_PROXY_AS_USER_NAME)),
```
sahilTakiar: i think the standard in gobblin is to use Guava's `Strings.isNullOrEmpty` method.

sahilTakiar:
Minor comment, otherwise LGTM.

ydailinkedin:
Updated.

sahilTakiar:
LGTM

PR #242, opened 2015-07-22
Support for specifying the group for a file

sahilTakiar:
Modification to `AvroHdfsDataWriter` that allows a config to specify the group for a file written to HDFS.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -219,6 +219,7 @@ public class ConfigurationKeys {
   public static final String WRITER_PARTITION_LEVEL = WRITER_PREFIX + ".partition.level";
   public static final String WRITER_PARTITION_PATTERN = WRITER_PREFIX + ".partition.pattern";
   public static final String WRITER_PARTITION_TIMEZONE = WRITER_PREFIX + ".partition.timezone";
+  public static final String WRITER_GROUP_NAME = WRITER_PREFIX + ".group.name";
   public static final String DEFAULT_WRITER_FILE_BASE_NAME = "part";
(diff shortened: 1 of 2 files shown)


sahilTakiar:
@liyinan926 and @ydailinkedin can you take a look?

ydailinkedin:
LGTM!

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -219,6 +219,7 @@
   public static final String WRITER_PARTITION_LEVEL = WRITER_PREFIX + ".partition.level";
   public static final String WRITER_PARTITION_PATTERN = WRITER_PREFIX + ".partition.pattern";
   public static final String WRITER_PARTITION_TIMEZONE = WRITER_PREFIX + ".partition.timezone";
+  public static final String WRITER_GROUP_NAME = WRITER_PREFIX + ".owner.name";
```
liyinan926: Should it be named `writer.group.name`?

sahilTakiar: Yes, you are write. Fixed.

liyinan926:
LGTM.

dmoore247:
Interested observer here: The roadmap calls for ORC, Parquet and many other types of HDFS writers I presume. When would it make sense to abstract the file attribute setting code out from the file format code? Put another way, setting the permissions and group ownership is independent of Avro. At some point in the future, setting Hadoop ACLs to facilitate more advanced security needs being required, you would not want to couple that to Avro, vs. ORC vs. Parquet...
Respectfully,
Douglas

liyinan926:
@dmoore247 yes, code that deals with general file attributes should be moved to `FsDataWriter`. We will fix this.

PR #245, opened 2015-07-22
Gobblin metrics executors were closed, causing hanged JVM. Fixed.

ibuenros:


ibuenros:
@ydailinkedin can you review?

gobblin-metrics/src/main/java/gobblin/metrics/reporter/EventReporter.java
```diff
@@ -62,7 +65,10 @@ public EventReporter(Builder builder) {
     super(builder.context, builder.name, builder.filter, builder.rateUnit, builder.durationUnit);
 
     this.closer = Closer.create();
-    this.immediateReportExecutor = Executors.newSingleThreadExecutor();
+    this.immediateReportExecutor = MoreExecutors.
+        getExitingExecutorService((ThreadPoolExecutor) Executors.newFixedThreadPool(1,
+            new ThreadFactoryBuilder().setNameFormat("EventReporter-" + builder.name + "-%d").build()),
+            5, TimeUnit.MINUTES);
```
ydailinkedin: What does 5 mins here mean? Is it enough?

ibuenros: They should ideally shutdown immediately. 5 min is an arbitrary buffer in case some messages still need to be sent to Kafka.

sahilTakiar:
Since both these classes implement `Closeable` would it just be easier to shutdown the `ExecutorService` in the `close()` method. We could just use Guava's `MoreExecutors.shutdownAndAwaitTermination` method (http://docs.guava-libraries.googlecode.com/git-history/master/javadoc/com/google/common/util/concurrent/MoreExecutors.html#shutdownAndAwaitTermination(java.util.concurrent.ExecutorService,%20long,%20java.util.concurrent.TimeUnit))

While both approaches achieve the same thing, I think this is a little cleaner. It looks like `MoreExecutors. getExitingExecutorService` basically sets all the threads as daemon threads, and then adds a shutdown hook to cleanly shutdown all the daemon threads, which sounds a little like a hack.

Just a thought, I'm fine with either approach.

ibuenros:
There are two issues with that approach:
* The main problem here is that `MoreExecutors.shutdownAndAwaitTermination` was introduced in version 17.0 of Guava, and we are using version 15.0. I tried upgrading the version, but we would have to also update all downstream projects, or risk breaking things.
* Also, if a user forgets to close a MetricContext, their program will hang, and it will probably take a while for them to find the root cause. With this implementation, the executors close themselves, with minimal input from user.

sahilTakiar:
Ok, thats fine. Some other comments.

* It looks like we are on Guava 18.0
* Is it necessary for each `MetricContext` to have its own threadpool. I'm not that familiar with the metrics implementation, but my understanding is that each class you want to emit metrics from requires one to create a new `MetricContext`, which in turn requires the class to create a new thread pool. My guess is that this would add more overhead than necessary, since you could just have one, centralized `ExecutorService`.

ibuenros:
The thread pool is created lazily. Only `MetricContext` that are reporting events create one (almost always only one `MetricContext`).

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -141,7 +144,10 @@ private MetricContext(String name, MetricContext parent, List<Tag<?>> tags,
 
   private synchronized ExecutorService getExecutorService() {
     if(!this.executorServiceOptional.isPresent()) {
-      this.executorServiceOptional = Optional.of(Executors.newCachedThreadPool());
+      this.executorServiceOptional = Optional.of(MoreExecutors.getExitingExecutorService(
```
chavdar: Please use gobblin.util.ExecutorsUtils.newThreadFactory(Optional<Logger>, Optional<String>) or a variation of it as by default uncaught exceptions in thread pools will go completely undetected.

gobblin-metrics/src/main/java/gobblin/metrics/reporter/EventReporter.java
```diff
@@ -62,7 +65,10 @@ public EventReporter(Builder builder) {
     super(builder.context, builder.name, builder.filter, builder.rateUnit, builder.durationUnit);
 
     this.closer = Closer.create();
-    this.immediateReportExecutor = Executors.newSingleThreadExecutor();
+    this.immediateReportExecutor = MoreExecutors.
```
chavdar: See above.

ibuenros:
@chavdar addressed comments.

chavdar:
LGTM

PR #246, opened 2015-07-22
Refactor Writer and Publisher, and add a Util class for publishing Kafka audit counts

zliu41:
- Break `BaseDataPublisher.publishData()` and `AvroHdfsTimePartitionedWriter.write()` into smaller methods, so that they can be reused by subclasses.
- Added a class `SerializationUtils` which serializes and deserializes an object to/from Strings using charset ISO_8859_1. This is used to transfer Kafka audit count objects from Writers to Publishers as a config property.
- Added a method in `AvroUtils` for converting `GenericRecord` into byte array.
- Some other small changes.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
@@ -153,19 +153,35 @@ public class AvroHdfsTimePartitionedWriter implements DataWriter<GenericRecord> 
 
   @Override
   public void write(GenericRecord record) throws IOException {
+    long recordTimestamp = getRecordTimestamp(record);
+    write(record, recordTimestamp);
+  }
 
-    // Retrieve the value of the field specified by this.partitionColumnName
-    Optional<Object> writerPartitionColumnValue;
-    if (this.partitionColumnName.isPresent()) {
-      writerPartitionColumnValue = AvroUtils.getFieldValue(record, this.partitionColumnName.get());
-    } else {
-      writerPartitionColumnValue = Optional.absent();
-    }
+  protected long getRecordTimestamp(GenericRecord record) {
+    Optional<Object> writerPartitionColumnValue = getWriterPartitionColumnValue(record);
+    return getRecordTimestamp(writerPartitionColumnValue);
+  }
 
-    // Check if the partition column value is present and is a Long object. Otherwise, use current system time.
-    long recordTimestamp = writerPartitionColumnValue.orNull() instanceof Long ? (Long) writerPartitionColumnValue.get()
+  /**
+   *  Check if the partition column value is present and is a Long object. Otherwise, use current system time.
+   */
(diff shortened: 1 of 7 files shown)


zliu41:
@liyinan926 can you take a look?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -153,19 +153,35 @@ public AvroHdfsTimePartitionedWriter(Destination destination, String writerId, S
 
   @Override
   public void write(GenericRecord record) throws IOException {
+    long recordTimestamp = getRecordTimestamp(record);
+    write(record, recordTimestamp);
```
liyinan926: You can do `write(record, getRecordTimestamp(record));`.

gobblin-utility/src/main/java/gobblin/util/SerializationUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   * @return A String representing the input object
+   */
+  public static <T extends Serializable> String serialize(T obj) throws IOException {
+    Closer closer = Closer.create();
+    try {
+      ByteArrayOutputStream bos = closer.register(new ByteArrayOutputStream());
+      ObjectOutputStream oos = closer.register(new ObjectOutputStream(bos));
+      oos.writeObject(obj);
+      oos.flush();
+      return bos.toString(CHARSET.name());
```
liyinan926: I suggest you add an additional parameter of type `BaseEncoding` to this method and use that to encode to and decode from strings.

gobblin-utility/src/main/java/gobblin/util/SerializationUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   * Deserialize a String obtained via {@link #serialize(Serializable)} into an object.
+   *
+   * @param serialized The serialized String
+   * @param clazz The class the deserialized object should be cast to.
+   * @return The deserialized object
+   */
+  public static <T extends Serializable> T deserialize(String serialized, Class<T> clazz) throws IOException {
+    Closer closer = Closer.create();
+    try {
+      ByteArrayInputStream bis = closer.register(new ByteArrayInputStream(serialized.getBytes(CHARSET)));
```
liyinan926: Ditto.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -153,19 +153,35 @@ public AvroHdfsTimePartitionedWriter(Destination destination, String writerId, S
...
+  }
+
+  protected long getRecordTimestamp(GenericRecord record) {
+    Optional<Object> writerPartitionColumnValue = getWriterPartitionColumnValue(record);
+    return getRecordTimestamp(writerPartitionColumnValue);
```
liyinan926: Ditto, you can combine the two lines.

zliu41:
@liyinan926 updated.

liyinan926:
LGTM.

PR #248, opened 2015-07-23
Moved code dealing with general file level arrributes to FsDataWriter

liyinan926:

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
@@ -22,9 +22,7 @@ import org.apache.avro.generic.GenericDatumWriter;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.DatumWriter;
 
-import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
@@ -54,9 +52,6 @@ class AvroHdfsDataWriter extends FsDataWriter<GenericRecord> {
   // Number of records successfully written
   protected final AtomicLong count = new AtomicLong(0);
 
-  // Whether the writer has already been closed or not
-  private volatile boolean closed = false;
-
   public AvroHdfsDataWriter(State properties, String fileName, Schema schema, int numBranches, int branchId)
       throws IOException {
     super(properties, fileName, numBranches, branchId);
@@ -69,7 +64,7 @@ class AvroHdfsDataWriter extends FsDataWriter<GenericRecord> {
 
     this.schema = schema;
     this.datumWriter = new GenericDatumWriter<GenericRecord>();
-    this.writer = createDatumWriter(this.stagingFile, codecFactory);
+    this.writer = this.closer.register(createDataFileWriter(codecFactory));
   }
 
   public FileSystem getFileSystem() {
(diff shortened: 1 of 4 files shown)


liyinan926:
@zliu41 @sahilTakiar can you review?

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -75,12 +82,33 @@ public FsDataWriter(State properties, String fileName, int numBranches, int bran
...
+        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_BLOCK_SIZE, numBranches, branchId),
+        this.fs.getDefaultBlockSize(this.outputFile));
+
+    this.permission = new FsPermission(properties.getPropAsShort(
+        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_PERMISSIONS, numBranches, branchId),
+        FsPermission.getDefault().toShort()));
+
+    Optional<String> group = Optional.fromNullable(properties.getProp(ConfigurationKeys.WRITER_GROUP_NAME));
+    if (group.isPresent()) {
+      this.fs.setOwner(this.stagingFile, this.fs.getFileStatus(this.stagingFile).getOwner(), group.get());
```
sahilTakiar: I'm not sure if this will work, because technically `this.stagingFile` doesn't exist yet. It only gets created in the `createDatumWriter` method of `AvroHdfsDataWriter`.

liyinan926: Yes, good point. I will move the if block to `commit` after the check on whether the staging file exists.

liyinan926:
Updated.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
```diff
@@ -36,26 +32,27 @@
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
 import gobblin.util.ForkOperatorUtils;
-import gobblin.util.HadoopUtils;
 import gobblin.util.WriterUtils;
 
 
 /**
- * An implementation of {@link DataWriter} that writes directly to HDFS in Avro format.
+ * An extension to {@link DataWriter} that writes in Avro format.
```
zliu41: DataWriter -> FsDataWriter

zliu41:
LGTM with one small comment

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
liyinan926: Fixed.

liyinan926:
@sahilTakiar any more comments?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
```diff
@@ -170,26 +117,16 @@ public long bytesWritten() throws IOException {
...
       throw new IOException(String.format("File %s already exists", avroFile));
     }
 
     FSDataOutputStream outputStream =
-        this.fs.create(avroFile, permissions, true, bufferSize, replication, blockSize, null);
-
-    if (group.isPresent()) {
-      this.fs.setOwner(avroFile, this.fs.getFileStatus(avroFile).getOwner(), group.get());
-    }
+        this.fs.create(avroFile, this.permission, true, this.bufferSize, this.replicationFactor, this.blockSize, null);
```
sahilTakiar: Can this be initialized in `FsDataWriter`, and we can make `outputStream` an protected instance variable. This way we can set the group of the file in `FsDataWriter` constructor rather than in the `commit()` method. I think this approach may be cleaner. Also, I think its important to set the group in the constructor because the `commit()` phase won't trigger until all the data is written, so there is a period of time where the file doesn't have the proper group. This could have security implications for the data purger job.

liyinan926:
Updated.

sahilTakiar:
LGTM

PR #249, opened 2015-07-27
Adding new data-management policies, updating proxying code, creating PublisherUtils

sahilTakiar:
Changes:

* Creating a few new data-management policies
* Creating new classes `ProxiedFileSystemUtils` and `ProxiedFileSystemCache` in order to replace `ProxiedFileSystemWrapper`. The main reason I did this was the clean up some of the code and to make a cache around all proxied filesystem objects created
* Creating a new class called `PublisherUtils` to store utilities for the `DataPublisher`s

### DIFF
--- a/gobblin-data-management/build.gradle
+++ b/gobblin-data-management/build.gradle
@@ -11,12 +11,16 @@
 apply plugin: 'java'
 
 dependencies {
+  compile project(":gobblin-api")
+  compile project(":gobblin-utility")
+
   compile externalDependency.guava
   compile externalDependency.slf4j
   compile externalDependency.log4j
   compile externalDependency.jodaTime
   compile externalDependency.findBugs
   compile externalDependency.azkaban
+  compile externalDependency.lombok
 
   testCompile externalDependency.testng
   testCompile externalDependency.mockito
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/DatasetCleaner.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/DatasetCleaner.java
@@ -1,3 +1,15 @@
+/*
+* Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+* this file except in compliance with the License. You may obtain a copy of the
+* License at  http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software distributed
+* under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+* CONDITIONS OF ANY KIND, either express or implied.
+*/
+
 package gobblin.data.management.retention;
 
 import java.io.IOException;
@@ -5,11 +17,9 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.Properties;
 
-import org.apache.hadoop.fs.FileSystem;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.google.common.base.Preconditions;
+
+import org.apache.hadoop.fs.FileSystem;
 
 import gobblin.data.management.retention.dataset.Dataset;
 import gobblin.data.management.retention.dataset.finder.DatasetFinder;
@@ -20,8 +30,6 @@ import gobblin.data.management.retention.dataset.finder.DatasetFinder;
  */
 public class DatasetCleaner {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(DatasetCleaner.class);
-
   public static final String CONFIGURATION_KEY_PREFIX = "gobblin.retention.";
   public static final String DATASET_PROFILE_CLASS_KEY = CONFIGURATION_KEY_PREFIX + "dataset.profile.class";
 
@@ -31,7 +39,7 @@ public class DatasetCleaner {
 
     Preconditions.checkArgument(props.containsKey(DATASET_PROFILE_CLASS_KEY));
 
-    try{
+    try {
       Class<?> datasetFinderClass = Class.forName(props.getProperty(DATASET_PROFILE_CLASS_KEY));
       this.datasetFinder = (DatasetFinder) datasetFinderClass.
           getConstructor(FileSystem.class, Properties.class).newInstance(fs, props);
@@ -55,10 +63,8 @@ public class DatasetCleaner {
   public void clean() throws IOException {
     List<Dataset> dataSets = this.datasetFinder.findDatasets();
 
-    for(Dataset dataset: dataSets) {
+    for (Dataset dataset: dataSets) {
       dataset.clean();
     }
-
   }
-
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ConfigurableGlobDatasetFinder.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ConfigurableGlobDatasetFinder.java
@@ -20,6 +20,7 @@ import java.util.regex.Pattern;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -33,6 +34,7 @@ import gobblin.data.management.retention.dataset.ConfigurableDataset;
 import gobblin.data.management.retention.dataset.Dataset;
 import gobblin.data.management.retention.dataset.finder.DatasetFinder;
 import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.util.PathUtils;
 
 
 /**
@@ -52,10 +54,10 @@ public class ConfigurableGlobDatasetFinder implements DatasetFinder {
   protected final Properties props;
 
   public ConfigurableGlobDatasetFinder(FileSystem fs, Properties props) throws IOException {
-    for(String property : requiredProperties()) {
+    for (String property : requiredProperties()) {
       Preconditions.checkArgument(props.containsKey(property));
     }
-    if(props.containsKey(DATASET_BLACKLIST_KEY) && !Strings.isNullOrEmpty(props.getProperty(DATASET_BLACKLIST_KEY))) {
+    if (props.containsKey(DATASET_BLACKLIST_KEY) && !Strings.isNullOrEmpty(props.getProperty(DATASET_BLACKLIST_KEY))) {
       this.blacklist = Optional.of(Pattern.compile(props.getProperty(DATASET_BLACKLIST_KEY)));
     } else {
       this.blacklist = Optional.absent();
@@ -83,8 +85,9 @@ public class ConfigurableGlobDatasetFinder implements DatasetFinder {
   @Override
   public List<Dataset> findDatasets() throws IOException {
     List<Dataset> datasets = Lists.newArrayList();
-    for(FileStatus fileStatus : this.fs.globStatus(datasetPattern)) {
-      if(this.blacklist.isPresent() && this.blacklist.get().matcher(fileStatus.getPath().toString()).matches()) {
+    for (FileStatus fileStatus : this.fs.globStatus(datasetPattern)) {
+      Path pathToMatch = PathUtils.getPathWithoutSchemeAndAuthority(fileStatus.getPath());
+      if (this.blacklist.isPresent() && this.blacklist.get().matcher(pathToMatch.toString()).matches()) {
         continue;
       }
       LOG.info("Found dataset at " + fileStatus.getPath());
--- a/gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
@@ -1,0 +1,94 @@
+/*
+* Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+* this file except in compliance with the License. You may obtain a copy of the
+* License at  http://www.apache.org/licenses/LICENSE-2.0
+*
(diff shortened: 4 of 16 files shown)


sahilTakiar:
@ydailinkedin and @ibuenros could you both please review?

gobblin-data-management/build.gradle
```diff
@@ -11,12 +11,16 @@
 apply plugin: 'java'
 
 dependencies {
+	compile project(":gobblin-api")
```
ydailinkedin: Extra spaces here.

gobblin-data-management/src/main/java/gobblin/data/management/retention/DatasetCleaner.java
```diff
@@ -5,11 +17,17 @@
 import java.util.List;
 import java.util.Properties;
 
+<<<<<<< HEAD
```
ydailinkedin: Incomplete merging?

sahilTakiar:
Updated

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
```diff
@@ -0,0 +1,79 @@
...
+    public boolean apply(StringDatasetVersion input) {
+      return Iterables.all(input.getPathsToDelete(), new ProxyablePath());
+    }
+  }
+
+  private class ProxyablePath implements Predicate<Path> {
+
+    @Override
+    public boolean apply(Path input) {
+      return ProxiedFileSystemUtils.canProxyAs(userNameToProxyAs, superUserName, superUserKeytabLocation);
```
ydailinkedin: For each `input`, this check will require a log-in action of the super user. I am not sure if there is any pre-check in filesystem cache for the fileowner of `input`. A potential problem is that for 1000 files of a same owner under a same hdfs dir, `apply` will log in 1000 times, which could cause performance issue.

gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ConfigurableGlobDatasetFinder.java
```diff
@@ -20,6 +20,7 @@
 import org.apache.hadoop.fs.FileStatus;
```
ydailinkedin: Missing license head.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
sahilTakiar: Yes, that is true. I think we can use https://hadoop.apache.org/docs/r1.2.1/api/org/apache/hadoop/security/UserGroupInformation.html#getLoginUser() in the `ProxiedFileSystemUtils.loginAndProxyUser` method to prevent the code from re-logging in as the same user. Do you know how to get the equivalent of `[user-name]/[cluster-name]@[realm]` from a UGI object?

ydailinkedin: Yes, we can getLoginUser() and compare  `[user-name] ` to determine if logging-in is necessary. But I am not sure if we need to compare `[cluster-name]@[realm]` besides  `[user-name] `?

gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ConfigurableGlobDatasetFinder.java
sahilTakiar: It's there.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
sahilTakiar: Updated. Turns out `UserGroupInformation.getLoginUser().getUserName()` is exactly equal to what we have been setting for `super.user.name.to.proxy.as.others`

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/NoOpRetentionPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+import azkaban.utils.Props;
+
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.StringDatasetVersion;
+
+/**
+ * Implementation of {@link RetentionPolicy} that uses a {@link StringDatasetVersion} and marks all
+ * {@link StringDatasetVersion}s as deletable.
+ */
+public class NoOpRetentionPolicy implements RetentionPolicy<StringDatasetVersion> {
```
ibuenros: I would not call this NoOp. To me NoOp means nothing will be deleted. Maybe call it DeleteAllRetentionPolicy instead?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/NoOpRetentionPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+
+import azkaban.utils.Props;
+
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.StringDatasetVersion;
+
+/**
+ * Implementation of {@link RetentionPolicy} that uses a {@link StringDatasetVersion} and marks all
+ * {@link StringDatasetVersion}s as deletable.
+ */
```
ibuenros: Why is this limited to StringDatasetVersion?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
ibuenros: This method is not using `input` for anything, am I wrong?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Implementation of {@link RetentionPolicy} that uses a {@link StringDatasetVersion} and checks if each name of a
+ * {@link StringDatasetVersion} corresponds to a proxyable user.
+ *
+ * <p>
+ *  A proxyable user is defined by the {@link ProxiedFileSystemUtils#canProxyAs(String, String, Path)} method.
+ * <p>
+ */
+@AllArgsConstructor
+public class ProxyableFolderRetentionPolicy implements RetentionPolicy<StringDatasetVersion> {
```
ibuenros: Why is this only applicable to StringDatasetVersion?

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/NoOpVersionFinder.java
```diff
@@ -0,0 +1,46 @@
...
+
+import gobblin.data.management.retention.dataset.Dataset;
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.StringDatasetVersion;
+
+/**
+ * Implementation of {@link VersionFinder} that uses a {@link StringDatasetVersion} and simply creates a single
+ * {@link StringDatasetVersion} for the given {@link Dataset}.
+ */
+public class NoOpVersionFinder implements VersionFinder<StringDatasetVersion> {
```
ibuenros: Should this be called something like SingletonVersionFinder?

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private static final int DEFAULT_MAX_CACHE_SIZE = 1000;
+
+  private static final Cache<String, FileSystem> USER_NAME_TO_FILESYSTEM_CACHE = CacheBuilder.newBuilder()
+      .maximumSize(DEFAULT_MAX_CACHE_SIZE).build();
+
+  /**
+   * Cached version of {@link ProxiedFileSystemUtils#getProxiedFileSystemUsingKeytab(String, String, Path, URI, Configuration)}.
+   */
+  public static FileSystem getProxiedFileSystemUsingKeytab(String userNameToProxyAs, String superUserName,
```
ibuenros: Guava cache has a pattern for adding elements using a Callable (see https://code.google.com/p/guava-libraries/wiki/CachesExplained under the heading "From a Callable"). It might be cleaner for this function.

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
```diff
@@ -0,0 +1,93 @@
...
+    USER_NAME_TO_FILESYSTEM_CACHE.put(userNameToProxyAs, proxiedFileSystem);
+    return proxiedFileSystem;
+  }
+
+  /**
+   * Cached version of {@link ProxiedFileSystemUtils#getProxiedFileSystemUsingKeytab(State, URI, Configuration)}.
+   */
+  public static FileSystem getProxiedFileSystemUsingKeytab(State state, URI fsURI, Configuration conf)
+      throws IOException, InterruptedException, URISyntaxException {
+
```
ibuenros: Maybe add a Preconditions check in here. Non-existing keys will return null, creating a NPE that is harder to debug.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private final String userNameToProxyAs;
+  private final String superUserName;
+  private final Path superUserKeytabLocation;
+
+  public ProxyableFolderRetentionPolicy(Props props) {
```
ibuenros: Maybe add preconditions check?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableFolderRetentionPolicy.java
sahilTakiar: Yes, you are right. That was a bug. Fixed it.

sahilTakiar:
@ibuenros updated, all comments addressed.

ydailinkedin:
LGTM!

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/WhiteListPredicateRetentionPolicy.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Properties;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+
+import gobblin.data.management.retention.version.DatasetVersion;
+
+
+public class WhiteListPredicateRetentionPolicy implements RetentionPolicy<DatasetVersion> {
```
ibuenros: Please add javadoc. Maybe rename to just PredicateRetentionPolicy?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/WhiteListPredicateRetentionPolicy.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public Class<? extends DatasetVersion> versionClass() {
+    return DatasetVersion.class;
+  }
+
+  @Override
+  public Collection<DatasetVersion> listDeletableVersions(List<DatasetVersion> allVersions) {
+    return Lists.newArrayList(Iterables.filter(allVersions, this.whitelistPredicate));
```
ibuenros: This will delete the versions that pass the predicate. IMO it is more natural to delete the versions that fail the predicate. Any thoughts?
This should be well documented either way.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableDatasetVersionPredicate.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.hadoop.fs.Path;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.StringDatasetVersion;
+import gobblin.util.ProxiedFileSystemUtils;
+
+
+@AllArgsConstructor
+public class ProxyableDatasetVersionPredicate implements Predicate<DatasetVersion> {
```
ibuenros: This class is hacky. `ProxiedFileSystemUtils.canProxyAs()` really should not be the way of figuring out whether a directory has the name of a user, nor does the name of the class imply that that's what it is doing.
Can we move either clean it up, or move the class to an internal repo?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/WhiteListPredicateRetentionPolicy.java
```diff
@@ -0,0 +1,52 @@
...
+
+import gobblin.data.management.retention.version.DatasetVersion;
+
+
+public class WhiteListPredicateRetentionPolicy implements RetentionPolicy<DatasetVersion> {
+
+  private final Predicate<DatasetVersion> whitelistPredicate;
+
+  private static final String WHITELIST_RETENTION_POLICY_PREDICATE_NAME =
+      "gobblin.retention.whitelist.retention.policy.predicate";
```
ibuenros: Maybe add `.class` to the property string?

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/ProxyableDatasetVersionPredicate.java
sahilTakiar: Agree is it hacky, unfortunately I don't see a better short term solution to the problem. Will move to an internal repo.

ibuenros:
LGTM

PR #253, opened 2015-07-29
make some changes for kafka adapter job-level metrics

zliu41:
- `AvroHdfsTimePartitionedWriter` now writes recordsWritten and bytesWritten to task state, so that the job launcher can aggregate them into job-level metrics.
- Added a noop method in `AbstractJobLauncher` so that subclasses can override it.
- `JobLauncherFactory` now supports creating job launcher based on class name, so I'll be able to write a job launcher class for Kafka.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -279,12 +279,12 @@ public abstract class AbstractJobLauncher implements JobLauncher {
       if (canCommit(this.jobContext.getJobCommitPolicy(), jobState)) {
         try {
           commitJob(jobState);
-          processTaskStates(jobState.getTaskStates());
         } finally {
           persistJobState(jobState);
         }
       }
 
+      postProcessTaskStates(jobState.getTaskStates());
       jobCommitTimer.stop();
     } catch (Throwable t) {
       jobState.setState(JobState.RunningState.FAILED);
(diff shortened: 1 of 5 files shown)


zliu41:
@liyinan926 please a take a look. Thanks.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -323,6 +324,14 @@ public void launchJob(JobListener jobListener) throws JobException {
     }
   }
 
+  /**
+   * Subclasses can override this method to do whatever processing on the taskStates,
```
liyinan926: Use `{@link TaskState}s`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -323,6 +324,14 @@ public void launchJob(JobListener jobListener) throws JobException {
...
+   * e.g., aggregate task-level metrics into job-level metrics.
+   */
+  protected void processTaskStates(List<TaskState> taskStates) {
```
liyinan926: `postProcessTaskStates` may be a better name? BTW: I will make big changes to the commit logic here to work on a per-dataset basis. I would suggest you move the method out of the `if` block.

zliu41:
@liyinan926 updated

zliu41:
Added some more metrics: (1) the earliest and average timestamps each writer writes. Used to measure latency; (2) the number of kafka partitions and messages that have errors. 187b89a

zliu41:
@liyinan926 do you have more comments?

liyinan926:
LGTM.

PR #254, opened 2015-07-29
Persist state and commit data on a per-dataset basis

liyinan926:
1. A `WorkUnit` can be specified to belong to a dataset group by using a new config property `dataset.urn`.
2. Job/task state will be persisted on a per-dataset basis into different state store files, one per dataset. All `current` dataset state store files will be loaded in the next job run. 
3. Commit of data will be done on a per-dataset basis, allowing datasets to be committed independently.
4. For existing job that does not have a dataset concept or only deals with a single dataset, the logic of job/task state persistence and data commit remains the same.
 
### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -126,6 +126,7 @@ public class ConfigurationKeys {
   /**
    * Dataset-related configuration properties;
    */
+  // This property is used to specify the URN of a dataset a job or WorkUnit extracts data for
   public static final String DATASET_URN_KEY = "dataset.urn";
   public static final String DEFAULT_DATASET_URN = "";
 
@@ -278,7 +279,8 @@ public class ConfigurationKeys {
   public static final String DATA_PUBLISHER_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".final.dir";
   public static final String DATA_PUBLISHER_REPLACE_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".replace.final.dir";
   public static final String DATA_PUBLISHER_FINAL_NAME = DATA_PUBLISHER_PREFIX + ".final.name";
-  public static final String DATA_PUBLISHER_GROUP_NAME = DATA_PUBLISHER_PREFIX + ".group.name";
+  // This property is used to specify the owner group of the data publisher final output directory
+  public static final String DATA_PUBLISHER_FINAL_DIR_GROUP = DATA_PUBLISHER_PREFIX + "final.dir.group";
 
   /**
    * Configuration properties used by the extractor.
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -124,6 +124,12 @@ public class ConfigurationKeys {
   public static final String JOB_STATE_FILE_PATH_KEY = "job.state.file.path";
 
   /**
+   * Dataset-related configuration properties;
+   */
+  public static final String DATASET_URN_KEY = "dataset.urn";
+  public static final String DEFAULT_DATASET_URN = "";
+
+  /**
    * Work unit related configuration properties.
    */
   public static final String WORK_UNIT_LOW_WATER_MARK_KEY = "workunit.low.water.mark";
@@ -264,6 +270,7 @@ public class ConfigurationKeys {
   public static final String DATA_PUBLISHER_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".final.dir";
   public static final String DATA_PUBLISHER_REPLACE_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".replace.final.dir";
   public static final String DATA_PUBLISHER_FINAL_NAME = DATA_PUBLISHER_PREFIX + ".final.name";
+  public static final String DATA_PUBLISHER_GROUP_NAME = DATA_PUBLISHER_PREFIX + ".group.name";
 
   /**
    * Configuration properties used by the extractor.
--- a/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/SourceState.java
@@ -17,6 +17,7 @@ import java.io.DataOutput;
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
@@ -25,10 +26,11 @@ import org.joda.time.DateTimeZone;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
-import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
 import gobblin.source.workunit.WorkUnit;
@@ -53,14 +55,14 @@ public class SourceState extends State {
   private static final DateTimeFormatter DTF =
       DateTimeFormat.forPattern("yyyyMMddHHmmss").withLocale(Locale.US).withZone(DateTimeZone.UTC);
 
-  private final Optional<SourceState> previousSourceState;
+  private final Map<String, SourceState> previousDatasetStatesByUrns;
   private final List<WorkUnitState> previousWorkUnitStates = Lists.newArrayList();
 
   /**
    * Default constructor.
    */
   public SourceState() {
-    this.previousSourceState = Optional.absent();
+    this.previousDatasetStatesByUrns = ImmutableMap.of();
   }
 
   /**
@@ -71,7 +73,7 @@ public class SourceState extends State {
    */
   public SourceState(State properties, List<WorkUnitState> previousWorkUnitStates) {
     super.addAll(properties);
-    this.previousSourceState = Optional.absent();
+    this.previousDatasetStatesByUrns = ImmutableMap.of();
     for (WorkUnitState workUnitState : previousWorkUnitStates) {
       this.previousWorkUnitStates.add(new ImmutableWorkUnitState(workUnitState));
     }
@@ -81,12 +83,13 @@ public class SourceState extends State {
    * Constructor.
    *
    * @param properties job configuration properties
-   * @param previousSourceState {@link SourceState} of the previous job run
+   * @param previousDatasetStatesByUrns {@link SourceState} of the previous job run
    * @param previousWorkUnitStates list of {@link WorkUnitState}s of the previous job run
    */
-  public SourceState(State properties, SourceState previousSourceState, List<WorkUnitState> previousWorkUnitStates) {
+  public SourceState(State properties, Map<String, ? extends SourceState> previousDatasetStatesByUrns,
+      List<WorkUnitState> previousWorkUnitStates) {
     super.addAll(properties);
-    this.previousSourceState = Optional.of(previousSourceState);
+    this.previousDatasetStatesByUrns = ImmutableMap.copyOf(previousDatasetStatesByUrns);
     for (WorkUnitState workUnitState : previousWorkUnitStates) {
       this.previousWorkUnitStates.add(new ImmutableWorkUnitState(workUnitState));
     }
@@ -95,10 +98,31 @@ public class SourceState extends State {
   /**
    * Get the {@link SourceState} of the previous job run.
    *
-   * @return {@link SourceState} of the previous job run
+   * <p>
+   *   This is a convenient method for existing jobs that do not use the new feature that allows output data to
+   *   be committed on a per-dataset basis. Use of this method assumes that the job deals with a single dataset,
+   *   which uses the default data URN defined by {@link ConfigurationKeys#DEFAULT_DATASET_URN}.
+   * </p>
+   *
+   * @return {@link SourceState} of the previous job run or {@code null} if no previous {@link SourceState} is found
    */
   public SourceState getPreviousSourceState() {
-    return new ImmutableSourceState(this.previousSourceState.or(new SourceState()));
+    return getPreviousDatasetState(ConfigurationKeys.DEFAULT_DATASET_URN);
+  }
+
+  /**
+   * Get the state (in the form of a {@link SourceState}) of a dataset identified by a dataset URN
+   * of the previous job run.
+   *
+   * @param datasetUrn the dataset URN
+   * @return the dataset state (in the form of a {@link SourceState}) of the previous job run
+   *         or {@code null} if no previous dataset state is found for the given dataset URN
+   */
+  public SourceState getPreviousDatasetState(String datasetUrn) {
+    if (!this.previousDatasetStatesByUrns.containsKey(datasetUrn)) {
+      return null;
+    }
+    return new ImmutableSourceState(this.previousDatasetStatesByUrns.get(datasetUrn));
   }
 
   /**
@@ -108,6 +132,32 @@ public class SourceState extends State {
    */
   public List<WorkUnitState> getPreviousWorkUnitStates() {
     return ImmutableList.<WorkUnitState> builder().addAll(this.previousWorkUnitStates).build();
+  }
+
+  /**
+   * Get a {@link Map} from dataset URNs (as being specified by {@link ConfigurationKeys#DATASET_URN_KEY}
+   * to the {@link List} of {@link WorkUnitState} with the dataset URNs.
+   *
+   * <p>
+   *   {@link WorkUnitState}s that do not have {@link ConfigurationKeys#DATASET_URN_KEY} set will be added
+   *   to the dataset state belonging to {@link ConfigurationKeys#DEFAULT_DATASET_URN}.
+   * </p>
+   *
+   * @return a {@link Map} from dataset URNs to the {@link List} of {@link WorkUnitState} with the dataset URNs
+   */
+  public Map<String, List<WorkUnitState>> getPreviousWorkUnitStatesByDatasetUrns() {
+    Map<String, List<WorkUnitState>> previousWorkUnitStatesByDatasetUrns = Maps.newHashMap();
+
+    for (WorkUnitState workUnitState : this.previousWorkUnitStates) {
+      String datasetUrn = workUnitState.getProp(ConfigurationKeys.DATASET_URN_KEY,
+          ConfigurationKeys.DEFAULT_DATASET_URN);
+      if (!previousWorkUnitStatesByDatasetUrns.containsKey(datasetUrn)) {
+        previousWorkUnitStatesByDatasetUrns.put(datasetUrn, Lists.<WorkUnitState>newArrayList());
+      }
+      previousWorkUnitStatesByDatasetUrns.get(datasetUrn).add(workUnitState);
+    }
+
+    return ImmutableMap.copyOf(previousWorkUnitStatesByDatasetUrns);
   }
 
   /**
@@ -122,7 +172,8 @@ public class SourceState extends State {
    * @param table name of the table this extract belongs to
    * @return a new unique {@link Extract} instance
    *
-   * @Deprecated Use {@link gobblin.source.extractor.extract.AbstractSource#createExtract(gobblin.source.workunit.Extract.TableType, String, String)}
+   * @Deprecated Use {@link gobblin.source.extractor.extract.AbstractSource#createExtract(
+   * gobblin.source.workunit.Extract.TableType, String, String)}
    */
   @Deprecated
   public synchronized Extract createExtract(Extract.TableType type, String namespace, String table) {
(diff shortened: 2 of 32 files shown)


liyinan926:
@zliu41 @chavdar can both of you review?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -19,6 +19,7 @@
 import java.io.IOException;
 import java.util.List;
 
+import java.util.Optional;
```
zliu41: use Guava's Optional

zliu41:
Is it possible that each workunit has multiple datasets? Currently each Kafka workunit contains multiple partitions, each of which should be a dataset.

liyinan926:
The problem with allowing a `WorkUnit` to have multiple dataset labels is there's no easy way to tell if each dataset in the `WorkUnit` has been successfully processed so there's no reliable way to determine if a dataset should be committed.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
liyinan926: It's an unused import, fixed by deleting it.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -57,6 +59,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(BaseDataPublisher.class);
 
   protected final List<FileSystem> fss = Lists.newArrayList();
+  protected final List<Optional<String>> groups = Lists.newArrayList();
```
zliu41: You can use `Lists.newArrayListWithCapacity(this.numBranches)`

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -190,18 +191,160 @@ public boolean isJobLockEnabled() {
...
-      if (!previousJobStateList.isEmpty()) {
-        // There should be a single job state on the list if the list is not empty
-        return previousJobStateList.get(0);
+  private Map<String, JobState> getPreviousDatasetStatesByUrns(String jobName) throws IOException {
+    Path stateStorePath = new Path(this.stateStoreRootDir, jobName);
+    if (!this.stateStoreFs.exists(stateStorePath)) {
+      return ImmutableMap.of();
+    }
+
+    FileStatus[] stateStoreFileStatuses = this.stateStoreFs.listStatus(stateStorePath, new PathFilter() {
```
zliu41: Seems `JobContext` heavily depends on the state store being an `FsStateStore`. Can you move the logic of getting previous dataset states to `StateStore` interface to make it more flexible?

liyinan926: I agree on this. Actually I think the entire `StateStore` interface needs some major refactoring to make it more usable and less-prone to misuse. I personally think it's better to work on that in a separate PR because this one is already pretty big. What do you think?

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -471,4 +501,40 @@ public JobExecutionInfo toJobExecutionInfo() {
...
+    copy.duration = other.duration;
+    copy.state = other.state;
+    if (copyTaskStates) {
+      copy.taskStates.putAll(other.taskStates);
+    }
+
+    return copy;
+  }
+
+  private static List<WorkUnitState> workUnitStatesFromDatasetStates(Collection<JobState> datasetStates) {
```
zliu41: You can change `Collection` to `Iterable` for more flexibility

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -190,18 +191,160 @@ public boolean isJobLockEnabled() {
...
-  private JobState getPreviousJobState(String jobName)
-      throws IOException {
-    if (this.jobStateStore.exists(jobName, "current" + AbstractJobLauncher.JOB_STATE_STORE_TABLE_SUFFIX)) {
-      // Read the job state of the most recent run of the job
-      List<JobState> previousJobStateList =
-          this.jobStateStore.getAll(jobName, "current" + AbstractJobLauncher.JOB_STATE_STORE_TABLE_SUFFIX);
-      if (!previousJobStateList.isEmpty()) {
-        // There should be a single job state on the list if the list is not empty
-        return previousJobStateList.get(0);
+  private Map<String, JobState> getPreviousDatasetStatesByUrns(String jobName) throws IOException {
```
zliu41: Is it better to rename `JobState` to `DataSet` state? A job run should only have a single job state. Having a map of job states is a little confusing

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
zliu41: makes sense

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -190,18 +191,160 @@ public boolean isJobLockEnabled() {
...
+      int failures = this.jobState.getPropAsInt(ConfigurationKeys.JOB_FAILURES_KEY, 0) + 1;
+      this.jobState.setProp(ConfigurationKeys.JOB_FAILURES_KEY, failures);
+    }
+  }
+
+  /**
+   * Check if it is OK to commit the output data of a dataset.
+   */
+  private boolean canCommitDataset(JobCommitPolicy commitPolicy, JobState datasetState) {
+    // Only commit job data if 1) COMMIT_ON_PARTIAL_SUCCESS is used,
```
zliu41: The comments here need to be updated

zliu41:
If multiple workunits have the same `dataset.urn`, then `JobContext.commit()` needs to check that all workunits of a urn succeeded before committing a dataset. Where is this check performed?

liyinan926:
Take a look at `commitDataset` and `canCommitDataset` in `JobContext`.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: It might be better to make `DatasetState` a subclass of `SourceClass` and use that instead of `JobState` here.

zliu41:
Got it

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -190,18 +191,160 @@ public boolean isJobLockEnabled() {
...
+    if (stateStoreFileStatuses == null || stateStoreFileStatuses.length == 0) {
+      return ImmutableMap.of();
+    }
+
+    Map<String, JobState> datasetStatesByUrns = Maps.newHashMap();
+    for (FileStatus stateStoreFileStatus : stateStoreFileStatuses) {
+      List<JobState> previousDatasetStates =
+          this.jobStateStore.getAll(jobName, stateStoreFileStatus.getPath().getName());
+      if (!previousDatasetStates.isEmpty()) {
+        // There should be a single dataset state on the list if the list is not empty
```
zliu41: Why is there a single dataset state, since there are multiple datasets in each job?

liyinan926: `getAll` returns a list even though the list itself only contains a single `JobState` object for the dataset. Each dataset state (an `JobState` object) is written to a separate file.

zliu41: I see

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: Actually just realized we need to use `JobState` for backward compatibility so persisted `JobState`s of existing jobs can be deserialized in the next run.

liyinan926: What I will do is to create an empty subclass called `DatasetState` that `extends JobState`. This new class is currently identical to `JobState` without any new stuffs. However, it has a more meaningful and less confusing name.

liyinan926:
Updated 2ff6961.

liyinan926:
Ignore update 2ff6961 as it is not backward compatible since it is impossible to use a `DatasetState` instance to deserialize a serialized `JobState` object. The `SequenceFile.Reader.next()` method will throw an `IOException` because the value classes are different.

liyinan926:
Updated f29a2b9.

liyinan926:
Updated d2cd673.

liyinan926:
Updated 638e846.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -471,4 +502,62 @@ public JobExecutionInfo toJobExecutionInfo() {
...
+  }
+
+  /**
+   * A subclass of {@link JobState} that is used to represent dataset states. This class is currently
+   * identical to {@link JobState} except that the name is more meaningful and less confusing.
+   */
+  public static class DatasetState extends JobState {
+
+    // For serialization/deserialization
+    @SuppressWarnings("unused")
```
zliu41: This seems unnecessary

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -471,4 +502,62 @@ public JobExecutionInfo toJobExecutionInfo() {
...
+    public void setDatasetUrn(String datasetUrn) {
+      setProp(ConfigurationKeys.DATASET_URN_KEY, datasetUrn);
+    }
+
+    public String getDatasetUrn() {
+      return getProp(ConfigurationKeys.DATASET_URN_KEY, ConfigurationKeys.DEFAULT_DATASET_URN);
+    }
+
+    @Override
+    public void write(DataOutput out) throws IOException {
```
zliu41: Do you need to override these two methods since they just call the super methods?

zliu41:
LGTM other than two minor comments

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
liyinan926: No, they should not be there. I will remove them.

liyinan926:
Updated.

liyinan926:
@chavdar can you review?

chavdar:
Could you create a wki for the state model? This is a very important aspect of Gobblin that should be understood particularly by people who develop new sources?

liyinan926:
Sure, will do.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -81,12 +83,13 @@ public SourceState(State properties, List<WorkUnitState> previousWorkUnitStates)
    * Constructor.
    *
    * @param properties job configuration properties
-   * @param previousSourceState {@link SourceState} of the previous job run
+   * @param previousDatasetStatesByUrns {@link SourceState} of the previous job run
    * @param previousWorkUnitStates list of {@link WorkUnitState}s of the previous job run
    */
-  public SourceState(State properties, SourceState previousSourceState, List<WorkUnitState> previousWorkUnitStates) {
+  public SourceState(State properties, Map<String, ? extends SourceState> previousDatasetStatesByUrns,
+      List<WorkUnitState> previousWorkUnitStates) {
```
chavdar: Please use Iterable<WorkUnitState> instead.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -111,6 +135,32 @@ public SourceState getPreviousSourceState() {
...
+   * to the {@link List} of {@link WorkUnitState} with the dataset URNs.
+   *
+   * <p>
+   *   {@link WorkUnitState}s that do not have {@link ConfigurationKeys#DATASET_URN_KEY} set will be added
+   *   to the dataset state belonging to {@link ConfigurationKeys#DEFAULT_DATASET_URN}.
+   * </p>
+   *
+   * @return a {@link Map} from dataset URNs to the {@link List} of {@link WorkUnitState} with the dataset URNs
+   */
+  public Map<String, List<WorkUnitState>> getPreviousWorkUnitStatesByDatasetUrns() {
```
chavdar: Please replace List<WorkUnitState> with Iterable<WorkUnitState>

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -56,9 +57,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(BaseDataPublisher.class);
 
-  protected final List<FileSystem> fss = Lists.newArrayList();
-  protected final Closer closer;
   protected final int numBranches;
+  protected final List<FileSystem> fss;
```
chavdar: Please use a more descriptive name.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -272,6 +278,7 @@
   public static final String DATA_PUBLISHER_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".final.dir";
   public static final String DATA_PUBLISHER_REPLACE_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".replace.final.dir";
   public static final String DATA_PUBLISHER_FINAL_NAME = DATA_PUBLISHER_PREFIX + ".final.name";
+  public static final String DATA_PUBLISHER_GROUP_NAME = DATA_PUBLISHER_PREFIX + ".group.name";
```
chavdar: What is the difference between DATASET_URN_KEY and DATA_PUBLISHER_GROUP_NAME ?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -56,9 +57,10 @@
...
+  protected final List<Optional<String>> groups;
```
chavdar: Please use a more descriptive name.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
liyinan926: Fixed.

gobblin-api/src/main/java/gobblin/configuration/SourceState.java
liyinan926: Fixed.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
liyinan926: Fixed.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
liyinan926: Fixed.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
liyinan926: You mean `DATA_PUBLISHER_FINAL_NAME` and `DATA_PUBLISHER_GROUP_NAME`, right?

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -170,38 +169,131 @@ public boolean isJobLockEnabled() {
...
+    return commitPolicy == JobCommitPolicy.COMMIT_ON_PARTIAL_SUCCESS
+        || (commitPolicy == JobCommitPolicy.COMMIT_ON_FULL_SUCCESS
+        && datasetState.getState() == JobState.RunningState.SUCCESSFUL);
+  }
+
+  /**
+   * Commit the output data of a dataset.
+   */
+  @SuppressWarnings("unchecked")
+  private void commitDataset(JobState.DatasetState datasetState) throws Exception {
```
chavdar: Can we use a more concrete exception class like IOException? Using Exception kind of defeats the purpose of static exception checking.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -170,38 +169,131 @@ public boolean isJobLockEnabled() {
...
+      // Increment the failure count by 1 if the job failed
+      int failures = this.jobState.getPropAsInt(ConfigurationKeys.JOB_FAILURES_KEY, 0) + 1;
+      this.jobState.setProp(ConfigurationKeys.JOB_FAILURES_KEY, failures);
+    }
+  }
+
+  /**
+   * Commit the job on a per-dataset basis.
+   */
+  void commit() throws Exception {
```
chavdar: See other comment about throwing Exception.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -170,38 +169,131 @@ public boolean isJobLockEnabled() {
...
+    Map<String, JobState.DatasetState> datasetStatesByUrns = this.jobState.getDatasetStatesByUrns();
+    for (Map.Entry<String, JobState.DatasetState> entry : datasetStatesByUrns.entrySet()) {
+      setFinalDatasetState(entry.getValue());
+      if (canCommitDataset(this.jobCommitPolicy, entry.getValue())) {
+        try {
+          commitDataset(entry.getValue());
+        } finally {
```
chavdar: I think you should catch the exceptions and log/save them. At the end, you can check if there was an exception thrown and rethrow it. This will allow us to commit as many dataset states even if there was an issue with some of them.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: Changed to `IOException`.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: Changed to `IOException`.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: Fixed.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
chavdar: I don't know. I am not completely sure what these 2 new properties mean: DATASET_URN_KEY and DATA_PUBLISHER_GROUP_NAME. The description of the PR talks about dataset groups which are defined by the dataset.urn property so I assumed a "dataset" and "dataset group" are the same thing.

liyinan926: OK, `DATASET_URN_KEY` is used to specify the URN of a dataset processed by a `WorkUnit`, whereas  `DATA_PUBLISHER_GROUP_NAME` is used to specify the own group of the final data publisher dir. I changed `DATA_PUBLISHER_GROUP_NAME` to `DATA_PUBLISHER_FINAL_DIR_GROUP`.

liyinan926: Also added comments as docs for both properties.

gobblin-runtime/src/main/java/gobblin/runtime/FsDatasetStateStore.java
```diff
@@ -0,0 +1,226 @@
...
+   * @param datasetUrn the dataset URN
+   * @param datasetState the {@link JobState.DatasetState} to persist
+   * @throws IOException if there's something wrong persisting the {@link JobState.DatasetState}
+   */
+  public void persistDatasetState(String datasetUrn, JobState.DatasetState datasetState) throws IOException {
+    String jobName = datasetState.getJobName();
+    String jobId = datasetState.getJobId();
+
+    String tableName = Strings.isNullOrEmpty(datasetUrn) ? jobId + DATASET_STATE_STORE_TABLE_SUFFIX
+        : datasetUrn + "-" + jobId + DATASET_STATE_STORE_TABLE_SUFFIX;
```
chavdar: We need some escaping here as the URNs may contain ':' and other characters which are not allowed. Overall, I think we have to enclose this logic in a separate function for easier managibility.

liyinan926: Added this:

```
private String getAliasName(String datasetUrn) {
    return Strings.isNullOrEmpty(datasetUrn) ?
        CURRENT_DATASET_STATE_FILE_SUFFIX + DATASET_STATE_STORE_TABLE_SUFFIX :
        CharMatcher.is(':').replaceFrom(datasetUrn, '.') + "-" + CURRENT_DATASET_STATE_FILE_SUFFIX +
            DATASET_STATE_STORE_TABLE_SUFFIX;
  }
```

liyinan926:
Updated ea208b5.

sahilTakiar:
Whats the difference between an `Extract` and a `dataset.urn`?

liyinan926:
A `dataset.urn` is simply a more general form of dataset identifiers than an `Extract`, which feels to me too closely tight to table-based datasets. I thought about using `Extract` for dataset identifiers, but decided to use something more general in the end. I also think `Extract` itself is not super clear and intuitive.

sahilTakiar:
Yes, I agree too. It's too closely tied to table based datasets. In that case, should we deprecate the `Extract` class in either this PR or a future one?

liyinan926:
As far as I know, data purger uses `Extract` for more than just getting the output path. @sahilTakiar you can give more details on that. So we cannot deprecate it as part of this PR. In the meanwhile, we probably also need to extract `dataset.urn` into a class that can hold more information about a dataset.

PR #256, opened 2015-07-29
ETL-3331 Add a helper class to add human-readable cluster names

chavdar:
We annotate metrics and tracking events with a cluster identifier ("clusterIdentifier"). Currently, we use the Hadoop RM or JT URL. We've got some feedback from the DWH team to have a more human- readable ID.

I've added a new class ClustersNames. It will look for a standard resource GobblinClustersNames.properties which contains a properties collection (map) from a resource manager/job tracker url to a human-readable string. If the resource is missing and/or there isn't an entry for a given URL, a normalized URL form will be used.


chavdar:
@zliu41 @pcadabam Could you review?

gobblin-runtime/src/main/java/gobblin/runtime/util/ClustersNames.java
```diff
@@ -0,0 +1,108 @@
...
+ * as name/value separators.
+ */
+public class ClustersNames {
+
+  public static final String URL_TO_NAME_MAP_RESOURCE_NAME = "GobblinClustersNames.properties";
+  private static final Logger LOG = LoggerFactory.getLogger(ClustersNames.class);
+
+  private static ClustersNames THE_INSTANCE;
+
+  private Properties _urlToNameMap = new Properties();
```
zliu41: No underscore in variable names.

pcadabam:
LGTM

gobblin-runtime/src/main/java/gobblin/runtime/util/ClustersNames.java
```diff
@@ -0,0 +1,108 @@
...
+  // Strip out the port number if it is a valid URI
+  private static String normalizeClusterUrl(String clusterIdentifier) {
+  try {
+    URI uri = new URI(clusterIdentifier.trim());
+    // URIs without protocol prefix
+    if (! uri.isOpaque() && null != uri.getHost()){
+      clusterIdentifier = uri.getHost();
+    }
+  } catch (URISyntaxException e) {
+    LOG.warn("Unable to parse cluster URL: " + e, e);
```
zliu41: I suggest removing this log warn. If a job calls this method 1000 times, this log message could appear 1000 times which is undesirable

gobblin-runtime/src/main/java/gobblin/runtime/util/ClustersNames.java
chavdar: Why not?

zliu41: Gobblin follows Google Java style which uses lower camel for field names. All other gobblin classes do so.

chavdar: LinkedIn's coding conventions call for using "_" in instance variables. Personally, I dislike having to constantly type "this." to distinguish instance variables.

zliu41:
LGTM

PR #257, opened 2015-07-31
ETL-2989: Add file level retention for fork operator

ydailinkedin:
1. Added ProxyableFileLevelDatasetProfile for finding file level datasets with proxy enabled.
2. Added SingleTimestampVersionFinder, which uses TimestampedDatasetVersion to find single versions.
3. Added FileListUtils, which is moved from internal idpc repo.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/DatasetCleaner.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/DatasetCleaner.java
@@ -38,20 +38,19 @@ public class DatasetCleaner {
   public DatasetCleaner(FileSystem fs, Properties props) throws IOException {
 
     Preconditions.checkArgument(props.containsKey(DATASET_PROFILE_CLASS_KEY));
-
     try {
       Class<?> datasetFinderClass = Class.forName(props.getProperty(DATASET_PROFILE_CLASS_KEY));
-      this.datasetFinder = (DatasetFinder) datasetFinderClass.
-          getConstructor(FileSystem.class, Properties.class).newInstance(fs, props);
-    } catch(ClassNotFoundException exception) {
+      this.datasetFinder =
+          (DatasetFinder) datasetFinderClass.getConstructor(FileSystem.class, Properties.class).newInstance(fs, props);
+    } catch (ClassNotFoundException exception) {
       throw new IOException(exception);
-    } catch(NoSuchMethodException exception) {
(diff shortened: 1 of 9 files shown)


ydailinkedin:
@ibuenros Can you review?

gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/ProxyableFileLevelDatasetProfile.java
```diff
@@ -0,0 +1,68 @@
...
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.data.management.retention.dataset.ConfigurableDataset;
+import gobblin.data.management.retention.dataset.Dataset;
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.util.FileListUtils;
+import gobblin.util.ProxiedFileSystemCache;
+
+
+public class ProxyableFileLevelDatasetProfile extends ConfigurableGlobDatasetFinder {
```
ibuenros: Please add javadoc

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/SingleTimestampVersionFinder.java
```diff
@@ -0,0 +1,57 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Lists;
+
+import gobblin.data.management.retention.dataset.Dataset;
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.TimestampedDatasetVersion;
+
+
+public class SingleTimestampVersionFinder implements VersionFinder<TimestampedDatasetVersion> {
```
ibuenros: javadoc

ibuenros:
This is just a suggestion:
Why not make a single dataset per glob match, and then make each file under the dataset a TimeStampedDatasetVersion? For example, if the glob is /user/*, each user will be a dataset, with one DatasetVersion per file owned by that user. You can do the proxying in the VersionFinder. Advantages:
* No need to modify ConfigurableGlobDatasetFinder
* Get empty directory cleanup for free
* (Debatable) cleaner implementation

ydailinkedin:
@ibuenros Updated.

ibuenros:
LGTM

PR #258, opened 2015-07-31
Add logs when quality check fails

pcadabam:
@ibuenros can you take a look at this.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/policies/count/RowCountRangePolicy.java
+++ b/gobblin-core/src/main/java/gobblin/policies/count/RowCountRangePolicy.java
@@ -41,7 +41,7 @@ public class RowCountRangePolicy extends TaskLevelPolicy {
       return Result.PASSED;
     } else {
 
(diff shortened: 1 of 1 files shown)


gobblin-core/src/main/java/gobblin/policies/count/RowCountRangePolicy.java
```diff
@@ -31,9 +36,16 @@ public RowCountRangePolicy(State state, Type type) {
 
   @Override
   public Result executePolicy() {
-    if (Math.abs((this.rowsWritten - this.rowsRead) / (double)this.rowsRead) <= this.range) {
+    double computedRange = Math.abs((this.rowsWritten - this.rowsRead) / (double)this.rowsRead);
+    if (computedRange <= this.range) {
       return Result.PASSED;
     } else {
+
```
chavdar: I think this should be error level.

ibuenros:
LGTM other than Chavdar's comment.

PR #261, opened 2015-08-03
Allow subclasses of BaseDataPublisher to use custom publisher output dir

liyinan926:
Currently the way `BaseDataPublisher` creates the final output directory path is fixed and cannot be customized. This PR opens this up so subclasses can create a custom final output directory, e.g., using a custom directory structure or naming pattern.


liyinan926:
@chavdar can you review?

chavdar:
LGTM.

PR #264, opened 2015-08-04
Fix a couple of FindBugs warnings

chavdar:
* Add proper closing of input streams to ClustersNames constructor to fix a FindBugs warning
* Rename TestClustersNames to ClustersNamesTest to be consistent with other tests.
* Fix a FindBugs warning in Scheduler utils which complains that "input" parameter might be null.

@pcadabam can you please review?


pcadabam:
LGTM

PR #266, opened 2015-08-04
Add Kafka offset-related metrics

zliu41:
Added counters for:
- Failed to get current available offset from Kafka
- persisted offset too early
- persisted offset too late


zliu41:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #270, opened 2015-08-05
Fix a bug in ProxiedFileSystemUtils

ydailinkedin:
@ibuenros Can you review?


chavdar:
Please provide a bug number and/or description of the bug and fix.

ydailinkedin:
`ProxiedFileSystemUtils` should be creating new proxiedFileSystem instead of calling `ProxiedFileSystemCache.getProxiedFileSystemUsingToken`, which would result in a recursive call on `ProxiedFileSystemCache.getProxiedFileSystemUsingToken`.

ibuenros:
LGTM

PR #274, opened 2015-08-05
Make superuser configuration keys consistent, and separate authpath for tokens/keytab

ydailinkedin:
SuperUser related configuration keys are duplicated. Even within the single class ProxiedFileSystemUtils, we have two places using two keys which have the same value.
Also, use AUTH_TOKEN_PATH for token path only.


ydailinkedin:
@ibuenros Can you review?

ibuenros:
LGTM

PR #275, opened 2015-08-05
Modified gobblin-metrics test to be less dependent on Kafka

ibuenros:


ibuenros:
@chavdar @pcadabam can you review?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroEventReporter.java
```diff
@@ -47,13 +47,15 @@ protected KafkaAvroEventReporter(Builder<?> builder) throws IOException {
    *
    * @param context the {@link gobblin.metrics.MetricContext} to report
    * @return KafkaAvroReporter builder
+   * @deprecated this method is bugged. Use {@link KafkaAvroEventReporter.Factory#forContext} instead.
```
chavdar: Could you elaborate what is broken?

ibuenros: On the javadoc?

The issue is that the return type of this method makes so that if you run
`KafkaEventReporter.forContext(context).modifierFunction1().modifierFunction2().build()`
where modifierFunction are anything like withSchemaRegistry, etc., then the last build() is no longer accesible, because the returned class is now of type MetricReport.Builder, which does not have a build() method (unless it gets recasted to BuilderImpl).

To fix this, the return type must be BuilderImpl instead, which is done in the Factory object.

chavdar:
LGTM.  When we move the Metric design wikis to github we should describe the pattern in the "How to write a reporter" section.

PR #279, opened 2015-08-06
KafkaSource sets actual high watermark in workunits

zliu41:
ETL-3352. High watermark was originally set into `WorkUnitState` by `KafkaExtractor`. However, if `KafkaExtractor` fails before the high watermark is set, and its `WorkUnitState` is still committed (which may happen if job.commit.policy is set to partial), the next run may lose checkpoint.

Now `KafkaSource` sets actual high watermark into `WorkUnit`s.

Also, not sure why `State` has both `getProp` and `getProperty`. Deprecated `getProperty`.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/State.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/State.java
@@ -341,10 +341,18 @@ public class State implements Writable {
     return Boolean.parseBoolean(getProperty(key, String.valueOf(def)));
   }
 
+  /**
+   * @deprecated Use {@link #getProp(String)}
+   */
+  @Deprecated
   protected String getProperty(String key) {
     return this.properties.getProperty(key);
   }
 
+  /**
+   * @deprecated Use {@link #getProp(String, String)}
+   */
+  @Deprecated
   protected String getProperty(String key, String def) {
     return this.properties.getProperty(key, def);
   }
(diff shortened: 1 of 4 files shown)


zliu41:

liyinan926:
@zliu41 Instead of setting the actual high watermark in the `WorkUnit`, which is not a good place for the `actual` high watermark, you can do so in `KafkaExtractor.close()`, which is guarantee to be called even if the task fails.

chavdar:
@liyinan926 I don't understand your concern. I think the invariant is that the WorkUnit should contain a low WM, high WM and actual WM. We should try to maintain the invariant at all times, including when the WU gets created. I don't thing we should wait till the finalization in KafkaExtractor.close() to restore the invariant.

LGTM.

liyinan926:
@chavdar the current API makes it clear that `WorkUnit` is supposed to contain static configuration and metadata only, and it should contain a `WatermarkInterval` that defines a low watermark and an expected high watermark. The `WorkUnitState`, however, is supposed to also contain runtime metadata, which includes the actual high watermark. Setting the actual high watermark violates this contract, and the Javadoc of the newly added method in `WorkUnit` in this PR also makes it clear that setting the actual high watermark in a `WorkUnit` is not a well-established pattern. It makes perfect sense to set the actual high watermark in the `close` method since at that point the actual high watermark should be well defined.

zliu41:
@chavdar I agree with Yinan. `Source` should set the low wm and the EXPECTED high wm. The ACTUAL high wm should be set by the `Extractor` depending on how many records it actually extracted.

chavdar:
@zliu41 In that case the source/extractor better not throw NPEs if they don't find the actual watermark.

chavdar:
Let's discuss this in person tomorrow morning. There is too much back-and-forth.

PR #280, opened 2015-08-10
Downgrade Guava to version 15.0 to work with Hadoop 2.6.0

liyinan926:
Due to the issue reported in HADOOP-10961, Hadoop 2.6.0 does not work with Guava version 16.0 or after because the deprecation of a constructor of Stopwatch. To make sure Gobblin supports Hadoop 2.6.0, we need to downgrade Guava to version 15.0.

### DIFF
--- a/gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java
+++ b/gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java
@@ -27,7 +27,7 @@ import org.testng.annotations.Test;
 
 public class AvroFsHelperTest {
 
-  @Test(groups = { "ignore" }, expectedExceptions = IllegalArgumentException.class)
+  @Test(expectedExceptions = IllegalArgumentException.class)
   public void testConnectFailsWithS3URLWithoutAWSCredentials() throws FileBasedHelperException {
     Configuration conf = new Configuration(); // plain conf, no S3 credentials
     SourceState sourceState = new SourceState();
--- a/gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java
+++ b/gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java
@@ -27,7 +27,7 @@ import org.testng.annotations.Test;
 
 public class AvroFsHelperTest {
 
-  @Test(expectedExceptions = IllegalArgumentException.class)
+  @Test(groups = { "ignore" }, expectedExceptions = IllegalArgumentException.class)
   public void testConnectFailsWithS3URLWithoutAWSCredentials() throws FileBasedHelperException {
     Configuration conf = new Configuration(); // plain conf, no S3 credentials
     SourceState sourceState = new SourceState();
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ ext.externalDependency = [
   "commonsHttpClient": "commons-httpclient:commons-httpclient:3.1",
   "datanucleusCore": "org.datanucleus:datanucleus-core:4.1.2",
   "datanucleusRdbms": "org.datanucleus:datanucleus-rdbms:4.1.2",
-  "guava": "com.google.guava:guava:18.0",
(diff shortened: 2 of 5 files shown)


liyinan926:
@chavdar can you review?

gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java
```diff
@@ -27,7 +27,7 @@
 
 public class AvroFsHelperTest {
 
-  @Test(expectedExceptions = IllegalArgumentException.class)
```
chavdar: why ignoring the test?

liyinan926: For some reason, Hadoop 2.6.0 seems not recognizing the `s3` scheme.

chavdar: What exception does it throw? Is it possible to add this to expectedExceptions instead of disabling the test?

liyinan926:
```
Caused by: java.io.IOException: No FileSystem for scheme: s3
	at org.apache.hadoop.fs.FileSystem.getFileSystemClass(FileSystem.java:2584)
	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:2591)
	at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:91)
	at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:2630)
	at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:2612)
	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:370)
	at gobblin.source.extractor.hadoop.AvroFsHelper.createFileSystem(AvroFsHelper.java:77)
	at gobblin.source.extractor.hadoop.AvroFsHelper.connect(AvroFsHelper.java:85)
```

liyinan926: Turns out the reason for this is: `For some reason, the jar hadoop-aws.jar which contains the implementation to NativeS3FileSystem is not present in the classpath of hadoop by default in the version 2.6.`.

liyinan926:
Updated.

chavdar:
LGTM

henryon:
@liyinan926  I found on master .
the file  gobblin/gobblin-core/src/test/java/gobblin/source/extractor/hadoop/AvroFsHelperTest.java line 30 
still old code, not latest

refer below:
https://github.com/liyinan926/gobblin/commit/e22a768d907bdc999df41578b6e69249c12ca9a1

mark1900:
Just a quick comment.  Not sure if it is related though it would seem that gobblin/bin/gobblin-mapreduce.sh still requires guava-18.0.jar.

```java
# Jars Gobblin runtime depends on
LIBJARS=$USER_JARS$separator$FWDIR_LIB/gobblin-metastore.jar,$FWDIR_LIB/gobblin-metrics.jar,\
$FWDIR_LIB/gobblin-core.jar,$FWDIR_LIB/gobblin-api.jar,$FWDIR_LIB/gobblin-utility.jar,\
$FWDIR_LIB/guava-18.0.jar,$FWDIR_LIB/avro-1.7.7.jar,$FWDIR_LIB/metrics-core-3.1.0.jar,\
$FWDIR_LIB/gson-2.3.1.jar,$FWDIR_LIB/joda-time-2.8.1.jar,$FWDIR_LIB/data-1.15.9.jar
```

henryon:
@liyinan926   passed on another box, the original box ,which running ambari master who's port is 8080. Thanks every !

sahilTakiar:
@mark1900 that does indeed look like a bug.

@liyinan926 is it possible to build `LIBJARS` in a similar fashion as `GOBBLIN_DEP_JARS`. The point being we probably shouldn't need to update `gobblin-mapreduce.sh` every time a dependency changes. Can we just use whatever is under `FWDIR_LIB`?

PR #281, opened 2015-08-10
Add record count and timestamp to Avro compaction output files

xkrogen:
Added logic to MRCompactorAvroKeyDedupJobRunner to rename the outputted files to include a timestamp and record count.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyCompactorOutputCommitter.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyCompactorOutputCommitter.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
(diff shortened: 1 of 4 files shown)


xkrogen:
@zliu41 Can you review?
Tested on Azkaban for daily and hourly compaction

zliu41:
Please add license headers to new files.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyCompactorOutputCommitter.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  @Override
+  public void commitTask(TaskAttemptContext context) throws IOException {
+
+    Path workPath = getWorkPath();
+    FileSystem fs = workPath.getFileSystem(context.getConfiguration());
+
+    if (fs.exists(workPath)) {
+      long recordCount = getRecordCountCounter(context);
+      String fileName = recordCount + "." + System.currentTimeMillis() + ".avro";
```
zliu41: please add "part-r-" as a prefix.
Also, please add a random number as another component, in order to prevent the unlikely case that two files happen to have the same record counts and same timestamp.

xkrogen:
Headers added, name updated.

xkrogen:
Whoops - meant to cast to an int first. Changed to use Random.nextInt()

zliu41:
LGTM.

PR #284, opened 2015-08-12
Added config for non-deduping compaction

xkrogen:
Added a config to have the ability to do compaction which is non-deduplicating, just copying files to the destination rather than going through all records and deduping.

Tested on Azkaban.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -460,9 +460,12 @@ public class ConfigurationKeys {
   public static final int DEFAULT_COMPACTION_MR_JOB_TIMEOUT_MINUTES = Integer.MAX_VALUE;
   public static final String COMPACTION_RECOMPACT_FOR_LATE_DATA = COMPACTION_PREFIX + "recompact.for.late.data";
   public static final boolean DEFAULT_COMPACTION_RECOMPACT_FOR_LATE_DATA = false;
+  public static final String COMPACTION_DEDUPLICATE = COMPACTION_PREFIX + "deduplicate";
+  public static final boolean DEFAULT_COMPACTION_DEDUPLICATE = true;
   public static final String COMPACTION_JOB_LATE_DATA_MOVEMENT_TASK = COMPACTION_PREFIX + "job.late.data.movement.task";
   public static final String COMPACTION_JOB_LATE_DATA_FILES = COMPACTION_PREFIX + "job.late.data.files";
   public static final String COMPACTION_COMPLETE_FILE_NAME = "_COMPACTION_COMPLETE";
+  public static final String COMPACTION_LATE_FILES_DIRECTORY = "late";
   public static final String COMPACTION_ENABLE_SUCCESS_FILE = "mapreduce.fileoutputcommitter.marksuccessfuljobs";
   public static final String COMPACTION_OVERWRITE_OUTPUT_DIR = COMPACTION_PREFIX + "overwrite.output.dir";
   public static final boolean DEFAULT_COMPACTION_OVERWRITE_OUTPUT_DIR = false;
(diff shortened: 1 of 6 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -89,28 +92,57 @@ public Void call() throws IOException, ClassNotFoundException, InterruptedExcept
...
     if (this.jobProps.getPropAsBoolean(ConfigurationKeys.COMPACTION_JOB_LATE_DATA_MOVEMENT_TASK, false)) {
-      this.copyLateDataFiles(new Path(this.outputPath, "late"), conf);
+      List<Path> lateFilePaths = Lists.newArrayList();
+      for (String filePathString : this.jobProps.getPropAsList(ConfigurationKeys.COMPACTION_JOB_LATE_DATA_FILES)) {
+        if (FilenameUtils.isExtension(filePathString, getApplicableFileExtensions())) {
+          lateFilePaths.add(new Path(filePathString));
+        }
+      }
+      this.copyDataFiles(
+          new Path(this.outputPath, ConfigurationKeys.COMPACTION_LATE_FILES_DIRECTORY), lateFilePaths, conf);
```
zliu41: If deduplication = false, late files should not be copied to the output dir instead of "late" dir.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -89,28 +92,57 @@ public Void call() throws IOException, ClassNotFoundException, InterruptedExcept
...
+  private void copyDataFiles(Path outputDirectory, List<Path> inputFilePaths, Configuration conf)
+      throws IOException {
+    for (Path filePath : inputFilePaths) {
+      String fileName = filePath.getName();
+      Path outPath;
+      int fileSuffix = 0;
+      do {
+        outPath = new Path(outputDirectory, String.format("%s.%d.%s",
+            FilenameUtils.getBaseName(fileName), fileSuffix++, FilenameUtils.getExtension(fileName)));
+      } while (this.fs.exists(outPath));
```
zliu41: Since each file has a unique fileSuffix, each file name should be unique. Do you really need this do-while?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -266,11 +266,13 @@ private void processTopic(String topic) throws IOException {
   MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Path topicOutputDir, Path topicTmpDir) {
     String builderClassName = this.state.getProp(ConfigurationKeys.COMPACTION_JOBPROPS_CREATOR_CLASS,
         ConfigurationKeys.DEFAULT_COMPACTION_JOBPROPS_CREATOR_CLASS) + "$Builder";
+    boolean deduplicate = this.state.getPropAsBoolean(ConfigurationKeys.COMPACTION_DEDUPLICATION,
```
zliu41: please use either "deduplication" or "deduplicate" consistently

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
xkrogen: The do-while is what makes sure that the file name is unique - otherwise the fileSuffix would always just be 0.

zliu41: Can you just declare `fileSuffix = 0` outside the for loop?

xkrogen: But what if there's already a file in the compaction output directory, then we copy over late data and it happens to have the same name?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -460,9 +460,12 @@
   public static final int DEFAULT_COMPACTION_MR_JOB_TIMEOUT_MINUTES = Integer.MAX_VALUE;
   public static final String COMPACTION_RECOMPACT_FOR_LATE_DATA = COMPACTION_PREFIX + "recompact.for.late.data";
   public static final boolean DEFAULT_COMPACTION_RECOMPACT_FOR_LATE_DATA = false;
+  public static final String COMPACTION_DEDUPLICATE = COMPACTION_PREFIX + "deduplicate";
+  public static final boolean DEFAULT_COMPACTION_DEDUPLICATION = true;
```
zliu41: deduplication -> deduplicate

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
zliu41: Makes sense.

zliu41:
LGTM

PR #285, opened 2015-08-12
Use jobState to get JobMetrics in TaskRunner

pcadabam:
@ibuenros can you review this?
This will make sure tags get populated from mapper tasks.


ibuenros:
LGTM

PR #287, opened 2015-08-13
Return filtered and non filtered results from PublisherUtils

pcadabam:
@sahilTakiar can you take a look at this?
Currently in the DataPurgerPublisher we filter out failed extracts. We want to also know the failed extracts to emit events

### DIFF
--- a/gobblin-utility/build.gradle
+++ b/gobblin-utility/build.gradle
@@ -25,6 +25,7 @@ dependencies {
   compile externalDependency.jacksonCore
   compile externalDependency.jasypt
   compile externalDependency.lombok
+
   if (project.hasProperty('useHadoop2')) {
     compile externalDependency.avroMapredH2
     runtime externalDependency.hadoopCommon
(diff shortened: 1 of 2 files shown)


sahilTakiar:
Can you re-factor the javadocs and the variables names to make it more generic. Since this method is splitting the WorkUnitStates by Predicate, it doesn't seem necessary to call one `MultiMap` `successful`, and the other one `failed`.

Also, what approach did you decide on with regards to the location of the event emission for failed `WorkUnitState`s.

pcadabam:
@sahilTakiar Please take a look
- I have updated the javadoc. 
- The approach I am going with is to submit the events from the DataPurgerPublisher. The reason for this is I want to emit only 1 event after a workunit commit is failed. 
If we want to emit the event from the framework, we need to make the framework aware of workunits that failed commit after the publisher has attempted to commit data. Also, each publisher (for Eg. DataPurgerPublisher) will have custom data in the event which may not be relevant to the framework.

sahilTakiar:
LGTM

PR #291, opened 2015-08-14
Moved AbstractJobLauncher.postProcessTaskStates to before commit

liyinan926:


liyinan926:
@zliu41 can you review?

zliu41:
Why put it before job commit? This was intended to be invoked after job committed successfully, e.g., publish job level metrics. If a job fails to commit, the metrics shouldn't be published

liyinan926:
I actually have a use case in Lynda data ingestion: the data publisher needs to know the value of a metric. What's the side effect if the Kafka ingestion job reports metrics in case of commit failure?

zliu41:
In Kafka ingestion, `postProcessTaskStates` will create counters such as number of records pulled. Will these counters be published if the job commit fails? If not then it's fine. It doesn't make sense to publish the counter if the job actually didn't commit those records.

liyinan926:
Makes sense. OK, I will try a different approach for the Lynda use case.

PR #293, opened 2015-08-17
Add compaction avro dedup key option

zliu41:
Previously there were two options for deduping: (1) set `compaction.use.all.attributes=true` and use all fields in the topic schema; (2) specify a custom dedup schema via `avro.key.schema.loc`.

Now adding a 3rd option suggested by @shirshanka : use all fields in the topic schema annotated by "primarykey". If there's no such fields, all fields will be used.

### DIFF
--- a/gobblin-compaction/test/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunnerTest.java
+++ b/gobblin-compaction/test/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunnerTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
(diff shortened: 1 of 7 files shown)


zliu41:
@liyinan926 can you take a look?

gobblin-compaction/test/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunnerTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.hadoop.mapreduce.Job;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+import gobblin.util.AvroUtils;
+
+
```
liyinan926: Missing `@Test{group = {}}`.

liyinan926:
LGTM.

PR #294, opened 2015-08-17
Adding Fork name tags to each Fork branch

sahilTakiar:
@ibuenros is this the correct approach?

Also, I see that `METRIC_CONTEXT_NAME_KEY` is set for `JobMetrics` on line 692 of `MRJobLauncher`, but I don't see an equivalent setting for `TaskMetrics`. Is this expected?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
@@ -14,6 +14,7 @@ package gobblin.runtime;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
@@ -23,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.Closer;
 
 import gobblin.configuration.ConfigurationKeys;
@@ -31,11 +33,14 @@ import gobblin.converter.Converter;
 import gobblin.converter.DataConversionException;
 import gobblin.converter.SchemaConversionException;
 import gobblin.metrics.GobblinMetrics;
+import gobblin.metrics.Tag;
+import gobblin.instrumented.Instrumented;
 import gobblin.instrumented.writer.InstrumentedDataWriterDecorator;
 import gobblin.publisher.TaskPublisher;
 import gobblin.qualitychecker.row.RowLevelPolicyCheckResults;
 import gobblin.qualitychecker.row.RowLevelPolicyChecker;
 import gobblin.qualitychecker.task.TaskLevelPolicyCheckResults;
+import gobblin.runtime.util.TaskMetrics;
 import gobblin.util.FinalState;
 import gobblin.util.ForkOperatorUtils;
 import gobblin.writer.DataWriter;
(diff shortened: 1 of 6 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -103,8 +107,10 @@
   private final AtomicReference<ForkState> forkState;
 
   public Fork(TaskContext taskContext, TaskState taskState, Object schema, int branches, int index,
-      CountDownLatch countDownLatch)
-      throws Exception {
+      CountDownLatch countDownLatch) throws Exception {
+
+    super(taskState, Fork.class.getClass(), createMetricsTags(taskState, index));
+    Instrumented.setMetricContextName(taskState, this.getMetricContext().getName());
```
ibuenros: You need to register the metric context before doing this (JobMetrics and TaskMetrics do it by default, but not Instrumented). You can do `GobblinMetricsRegistry.getInstance().putIfAbsent(context.getName(), context)`.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -60,7 +64,7 @@
  * @author ynli
  */
 @SuppressWarnings("unchecked")
-public class Fork implements Closeable, Runnable, FinalState {
+public class Fork extends Instrumented implements Closeable, Runnable, FinalState {
```
ibuenros: Can you make sure the method close() calls super.close()?

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -103,8 +107,10 @@
   private final AtomicReference<ForkState> forkState;
 
   public Fork(TaskContext taskContext, TaskState taskState, Object schema, int branches, int index,
-      CountDownLatch countDownLatch)
-      throws Exception {
+      CountDownLatch countDownLatch) throws Exception {
+
+    super(taskState, Fork.class.getClass(), createMetricsTags(taskState, index));
+    Instrumented.setMetricContextName(taskState, this.getMetricContext().getName());
```
ibuenros: Also, this will mutate the original input TaskState, which might not be a good idea if it is used elsewhere. Can we make a copy of the state instead?

ibuenros:
About the use of METRIC_CONTEXT_NAME_KEY. It seems that this constant is defined in two different places: in ConfigurationKeys, and also in Instrumented. That is a mistake, we should use only one of them.

sahilTakiar:
With the current code base, its a little weird to have a class extend `Instrumented` and to get it registered in `GobblinMetricsRegistry`. When one extends `Instrumented`, the super constructor invokes `getMetricContext` which creates a `MetricContext` properly. However, the `MetricContext` is not registered in `GobblinMetricsRegistry`. In order to get it registered, I tried to create a `GobblinMetrics` instance using the `GobblinMetrics.get()` method (which automatically registers the newly made `MetricContext` in `GobblinMetricsRegistry`). However the `get()` method requires a `parentContext` be specified. This is where the problem arises, the `Instrumented` constructor has already made a `metricContext` that is a child of `TaskMetrics`, and now `GobblinMetrics` requires a `parentContext` to be specified.

So instead I solely use `GobblinMetrics`.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
sahilTakiar: Created a copy constructor in `TaskState`

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
sahilTakiar: I hold a reference to `MetricContext` and close it in the `Fork.close()` method.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
sahilTakiar: Addressed, see comments below.

sahilTakiar:
@ibuenros can you take another look?

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -102,32 +107,33 @@
...
-      CountDownLatch countDownLatch)
-      throws Exception {
+      CountDownLatch countDownLatch) throws Exception {
 
     this.logger = LoggerFactory.getLogger(Fork.class.getName() + "-" + index);
 
     this.taskContext = taskContext;
-    this.taskState = taskState;
-    this.taskId = taskState.getTaskId();
+    this.taskStateCopy = new TaskState(taskState);
```
ibuenros: This could just be `this.taskState = new TaskState(taskState)`, given that we are not also storing the original. But I don't really have a preference either way.

ibuenros:
It looks good. Have you tested that the metric contexts do get generated as expected?

sahilTakiar:
I tested it, it is working as expected. The only change I had to make was to remove the copying of `TaskState` it seems that the `Fork` class is already mutating the `TaskState` object in a very other places, and there are some features that depend on the ability of mutating `TaskState`. Basically, a bunch of tests failing when I added the copying of `TaskState` so I removed it.

PR #296, opened 2015-08-18
Make some change regarding file permission

zliu41:
1. In `FsDataWriter`, when getting the writer output file permission from the state, `Short.parseShort` should be based on radix 8, not 10.
1. Originally `FsDataWriter` calls `this.fs.mkdirs`, which does not use `this.permission`, but uses default permission. Although there's another version of `FileSystem.mkdirs()` that takes a `FsPermission` object, it does not set permissions recursively. So I added a util method `mkdirsWithRecursivePermission()`.
1. In `BaseDataPublisher` add a `List<FsPermission>` field to correctly set publisher output file/folder permissions.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.permission.FsPermission;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -60,6 +61,7 @@ public class BaseDataPublisher extends DataPublisher {
   protected final int numBranches;
   protected final List<FileSystem> fileSystemByBranches;
   protected final List<Optional<String>> publisherFinalDirOwnerGroupsByBranches;
+  protected final List<FsPermission> permissions;
   protected final Closer closer;
   protected final int parallelRunnerThreads;
   protected final Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();
@@ -78,6 +80,7 @@ public class BaseDataPublisher extends DataPublisher {
 
     this.fileSystemByBranches = Lists.newArrayListWithCapacity(this.numBranches);
     this.publisherFinalDirOwnerGroupsByBranches = Lists.newArrayListWithCapacity(this.numBranches);
+    this.permissions = Lists.newArrayListWithCapacity(this.numBranches);
 
     // Get a FileSystem instance for each branch
     for (int i = 0; i < this.numBranches; i++) {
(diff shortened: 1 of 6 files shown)


zliu41:
@liyinan926 please take a look

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -89,6 +92,10 @@ public BaseDataPublisher(State state) throws IOException {
       // The group(s) will be applied to the final publisher output directory(ies)
       this.publisherFinalDirOwnerGroupsByBranches.add(Optional.fromNullable(this.getState().getProp(ForkOperatorUtils
           .getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISHER_FINAL_DIR_GROUP, this.numBranches, i))));
+
+      this.permissions.add(new FsPermission(state.getPropAsShortWithRadix(
```
liyinan926: Can you add a comment on the application scope of the `permissions`?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -89,6 +92,10 @@ public BaseDataPublisher(State state) throws IOException {
...
+          ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISHER_PERMISSION, numBranches, i),
+          FsPermission.getDefault().toShort(), 8)));
```
liyinan926: Seems `8` is used several time, so it deserves defining it as a constant.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -204,7 +212,7 @@ protected void addWriterOutputToExistingDir(Path writerOutputDir, Path publisher
           : new Path(publisherOutputDir, status.getPath().getName());
 
       LOG.info(String.format("Moving %s to %s", status.getPath(), finalOutputPath));
-      parallelRunner.renamePath(status.getPath(), finalOutputPath, Optional.<String>absent());
+      parallelRunner.renamePath(status.getPath(), finalOutputPath, Optional.<String> absent());
```
liyinan926: Can you remove the empty space after `>`?

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataPublisher.java
```diff
@@ -59,12 +60,11 @@ protected void addWriterOutputToExistingDir(Path writerOutput, Path publisherOut
...
 
-      if (!this.fileSystemByBranches.get(branchId).exists(outputPath.getParent())) {
-        this.fileSystemByBranches.get(branchId).mkdirs(outputPath.getParent());
-      }
+      WriterUtils.mkdirsWithRecursivePermission(this.fileSystemByBranches.get(branchId), outputPath.getParent(),
+          this.permissions.get(branchId));
 
       LOG.info(String.format("Moving %s to %s", status.getPath(), outputPath));
-      parallelRunner.renamePath(status.getPath(), outputPath, Optional.<String>absent());
+      parallelRunner.renamePath(status.getPath(), outputPath, Optional.<String> absent());
```
liyinan926: Ditto.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -162,7 +169,8 @@ protected void publishData(WorkUnitState workUnitState, int branchId, Set<Path>
     } else {
       Path parentDir = publisherOutputDir.getParent();
       // Create the parent directory of the final output directory if it does not exist
-      this.fileSystemByBranches.get(branchId).mkdirs(parentDir);
+      WriterUtils.mkdirsWithRecursivePermission(this.fileSystemByBranches.get(branchId), parentDir,
```
liyinan926: You can simply do `WriterUtils.mkdirsWithRecursivePermission(this.fileSystemByBranches.get(branchId), publisherOutputDir,
+          this.permissions.get(branchId));`.

zliu41:
@liyinan926 updated. Also added a new property `writer.dir.permissions` to be distinguished from `writer.file.permissions`.

zliu41:
If `publisherOutputDir` is created here, it may cause problem when the publisher moves `writerOutptuDir` to `publisherOutputDir` since `publisherOutputDir` already exists.

PR #297, opened 2015-08-19
Fixing FindBugs warnings

chavdar:
Issues fixed:
* gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java doing unnecessary boxing/unboxing of long
* gobblin-runtime/src/main/java/gobblin/runtime/Task.java using ++ on a volatile
* gobblin-runtime/src/main/java/gobblin/runtime/local/LocalJobManager.java - negating an integer in compare() (may cause integer overflow)
* gobblin-runtime/src/main/java/gobblin/runtime/local/LocalJobManager.java not closing a stream
* gobblin-utility/src/main/java/gobblin/util/AvroUtils.java - making FB shut up that input can be null

@liyinan926 @ibuenros can you have a look?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
@@ -12,9 +12,6 @@
 
 package gobblin.source.extractor.utils;
 
-import gobblin.configuration.ConfigurationKeys;
-import gobblin.source.extractor.watermark.WatermarkType;
-
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
@@ -38,6 +35,9 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.source.extractor.watermark.WatermarkType;
 
 
 public class Utils {
--- a/gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
@@ -12,6 +12,9 @@
 
 package gobblin.source.extractor.utils;
 
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.source.extractor.watermark.WatermarkType;
+
 import java.text.ParseException;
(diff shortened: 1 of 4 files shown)


sahilTakiar:
For `AvroUtils` I think you can just tag it with `javax.annotation.Nonnull`. That should make the FindBugs warnings go away.

gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
```diff
@@ -12,6 +12,9 @@
 
 package gobblin.source.extractor.utils;
 
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.source.extractor.watermark.WatermarkType;
```
liyinan926: The convention for import orders we use put gobblin imports at the end.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -38,6 +25,20 @@
 import gobblin.qualitychecker.row.RowLevelPolicyChecker;
 import gobblin.runtime.util.RuntimeConstructs;
 
+import java.io.IOException;
```
liyinan926: Ditto. We use the ordering throughout the code base.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -82,7 +83,7 @@
   private final List<Optional<Fork>> forks = Lists.newArrayList();
 
   // Number of task retries
-  private volatile int retryCount = 0;
+  private AtomicInteger retryCount = new AtomicInteger();
```
liyinan926: This can be final.

chavdar: Fixed.

chavdar:
@sahilTakiar It is defined as nullable in the Guava interface, this is the contract of the API.

gobblin-core/src/main/java/gobblin/source/extractor/utils/Utils.java
chavdar: Fixed.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
chavdar: Fixed.

liyinan926:
LGTM.

PR #298, opened 2015-08-20
use lowWm in KafkaSource if actualHighWm doesn't exist

zliu41:
also deprecated redundant methods `getProperty` in `State` and `WorkUnitState`.


zliu41:
@chavdar can you review?

chavdar:
LGTM

PR #301, opened 2015-08-21
ETL-3428: Fix for cleaning staging data using proxy file system

ydailinkedin:
The parallel runner will use the same filesystem for the ones that have same fs uri. But in proxy filesystem, besides comparing fs uri, the user of that proxy file system should also be compared.
Add fs.getHomeDirectory() to getParallelRunner. 
Tested on Nertz.
@sahilTakiar Can you review it?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
@@ -166,10 +166,10 @@ public class JobLauncherUtils {
           state.getProp(ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI,
               numBranches, branchId), ConfigurationKeys.LOCAL_FS_URI);
       FileSystem fs = getFsWithProxy(state, writerFsUri);
-
       ParallelRunner parallelRunner = getParallelRunner(fs, closer, parallelRunnerThreads, parallelRunners);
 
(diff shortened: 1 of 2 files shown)


gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
```diff
@@ -212,10 +212,10 @@ private static FileSystem getFsWithProxy(State state, String writerFsUri)
 
   private static ParallelRunner getParallelRunner(FileSystem fs, Closer closer, int parallelRunnerThreads,
       Map<String, ParallelRunner> parallelRunners) {
-    String uri = fs.getUri().toString();
-    if (!parallelRunners.containsKey(uri)) {
-      parallelRunners.put(uri, closer.register(new ParallelRunner(parallelRunnerThreads, fs)));
+    String uriAndHomeDir = fs.getUri().toString() + fs.getHomeDirectory();
```
sahilTakiar: We could probably changed this to `new Path(new Path(fs.getUri), fs.getHomeDirectory())`, and make the key for the `Map` a `Path` instead of a `String`.

sahilTakiar:
LGTM besides the minor comment. Unfortunately, I cannot find a cleaner way to do this than to use the home directory of the `FileSystem` object. Ideally `FileSystem.equals()` would be sufficient for this, the `equals()` method doesn't seem to be overloaded in the `FileSystem` class, or any of its subclasses.

Merging this for now since it is a blocker. I will address the comment in another PR.

PR #303, opened 2015-08-21
Allow AvroHdfsTimePartitionedWriter to take multiple partition columns

zliu41:
Before: `AvroHdfsTimePartitionedWriter` can only take a single partition column
Now: it can take multiple partition columns, which will be tried one by one.


zliu41:
@sahilTakiar can you take a look?

sahilTakiar:
LGTM

PR #304, opened 2015-08-21
Fixed the way completed task count is computed and some refactoring

liyinan926:

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
@@ -169,6 +169,16 @@ public class TaskState extends WorkUnitState {
   }
 
   /**
+   * Return whether the task has completed running or not.
+   *
+   * @return {@code true} if the task has completed or {@code false} otherwise
+   */
+  public boolean isCompleted() {
+    WorkingState state = getWorkingState();
+    return state == WorkingState.SUCCESSFUL || state == WorkingState.COMMITTED || state == WorkingState.FAILED;
+  }
+
+  /**
    * Update record-level metrics.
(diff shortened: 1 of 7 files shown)


liyinan926:
@zliu41 can you review?

gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
```diff
@@ -169,6 +169,16 @@ public void setTaskDuration(long duration) {
   }
 
   /**
+   * Return whether the task has completed running or not.
+   *
+   * @return {@code true} if the task has completed or {@code false} otherwise
+   */
+  public boolean isCompleted() {
+    WorkingState state = getWorkingState();
+    return state == WorkingState.SUCCESSFUL || state == WorkingState.COMMITTED || state == WorkingState.FAILED;
```
zliu41: Should CANCELLED be included?

liyinan926: CANCELLED tasks are semantically not completed so that's why it's not included here.

zliu41: In `LocalJobManager.onTaskCompletion()`, it won't commit the job unless `jobState.getCompletedTasks() == jobState.getTaskCount()`. If a CANCELLED task doesn't count for completed tasks but does count for total task count, will it prevent the job from being committed?

liyinan926: `LocalJobManager` has been long deprecated and shouldn't be used.

zliu41:
LGTM

PR #306, opened 2015-08-24
Add another option to KafkaReportingFormats

zliu41:
Add another option `CLASS` for `KafkaReportingFormats` so that I can pass in a custom reporter instead of using `KafkaAvroReporter`


zliu41:
@ibuenros please take a look, thanks

PR #307, opened 2015-08-25
Add classname and line number to azkaban logs

pcadabam:
@liyinan926 and @chavdar can you guys take a looks this?
- I spoke to the Azkaban team today as to why the azkaban logs do not have class names and log messages even though we have it in our log4j conifg files. It turns out the azkaban hard codes the pattern to just project name and message. In fact it seems like azkaban ignores all the log4j properties we set.
- This change is to add the class name and line number to all the appenders in the root logger.

### DIFF
--- a/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
+++ b/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
@@ -12,13 +12,16 @@
 
 package gobblin.azkaban;
 
+import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.log4j.Appender;
 import org.apache.log4j.Logger;
+import org.apache.log4j.PatternLayout;
 
 import azkaban.jobExecutor.AbstractJob;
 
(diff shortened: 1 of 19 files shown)


gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -126,6 +132,22 @@ public AzkabanJobLauncher(String jobId, Properties props)
...
+   */
+  private void updateLogger() {
+
+    @SuppressWarnings("unchecked")
+    Enumeration<Appender> appenders = Logger.getRootLogger().getAllAppenders();
+
+    if (appenders != null) {
+      while (appenders.hasMoreElements()) {
+        appenders.nextElement().setLayout(new PatternLayout(LOGGER_PATTERN_LAYOUT));
+
```
liyinan926: Additional empty line.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -126,6 +132,22 @@ public AzkabanJobLauncher(String jobId, Properties props)
     return tags;
   }
 
+  /**
+   * Azkaban ignores the log4j config provided in the artifacts. This method includes class name and line number to azkaban logs
```
liyinan926: Can you wrap the Javadoc and add dot at the end?

liyinan926:

chavdar:
LGTM

PR #308, opened 2015-08-25
Enable retention based on folder mod time

zliu41:
@ibuenros can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/ModDateTimeDatasetVersionFinder.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/ModDateTimeDatasetVersionFinder.java
@@ -1,0 +1,40 @@
+package gobblin.data.management.retention.version.finder;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Properties;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.joda.time.DateTime;
+
+import com.google.common.collect.Lists;
+
+import gobblin.data.management.retention.dataset.Dataset;
+import gobblin.data.management.retention.version.DatasetVersion;
+import gobblin.data.management.retention.version.TimestampedDatasetVersion;
+
+
+/**
+ * {@link VersionFinder} for datasets based on modification timestamps.
+ */
+public class ModDateTimeDatasetVersionFinder implements VersionFinder<TimestampedDatasetVersion> {
+
+  private final FileSystem fs;
+
+  public ModDateTimeDatasetVersionFinder(FileSystem fs, Properties props) {
+    this.fs = fs;
+  }
+
+  @Override
+  public Class<? extends DatasetVersion> versionClass() {
+    return TimestampedDatasetVersion.class;
+  }
+
+  @Override
+  public Collection<TimestampedDatasetVersion> findDatasetVersions(Dataset dataset) throws IOException {
+    FileStatus status = this.fs.getFileStatus(dataset.datasetRoot());
+    return Lists
+        .newArrayList(new TimestampedDatasetVersion(new DateTime(status.getModificationTime()), dataset.datasetRoot()));
+  }
+}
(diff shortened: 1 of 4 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/retention/profile/SimpleDatasetProfile.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * {@link gobblin.data.management.retention.dataset.finder.DatasetFinder} for simple datasets.
+ *
+ * Simple datasets will be cleaned by the modification timestamps of the datasets that match
+ * 'gobblin.retention.dataset.pattern'.
+ */
+public class SimpleDatasetProfile extends ConfigurableGlobDatasetFinder {
+  public SimpleDatasetProfile(FileSystem fs, Properties props) throws IOException {
+    super(fs, props);
+  }
```
ibuenros: You need to override the method `dataseAtPath(Path)`.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/SimpleDataset.java
```diff
@@ -0,0 +1,55 @@
...
+import gobblin.data.management.retention.version.finder.ModDateTimeDatasetVersionFinder;
+import gobblin.data.management.retention.version.finder.VersionFinder;
+
+
+/**
+ * {@link DatasetBase} for a simple dataset.
+ *
+ * Uses a {@link ModDateTimeDatasetVersionFinder} and a {@link TimeBasedRetentionPolicy}.
+ */
+public class SimpleDataset extends DatasetBase<TimestampedDatasetVersion> {
```
ibuenros: Can we call this something different from SimpleDataset? (not very informative). Maybe ModificationTimeDataset?

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/ModDateTimeDatasetVersionFinder.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * {@link DatasetVersionFinder} for datasets based on modification timestamps.
+ */
+public class ModDateTimeDatasetVersionFinder extends DatasetVersionFinder<TimestampedDatasetVersion> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ModDateTimeDatasetVersionFinder.class);
+
+  private static final Path GLOB_PATTERN = new Path(".");
```
ibuenros: Will this glob correctly? I don't understand exactly what your datasets are and what your dataset versions are, can you give an example?

zliu41: Yes. suppose we have
```
/root
    /a
    /b
```
If `gobblin.retention.dataset.pattern=/root/*` then datasets will be `a` and `b`, the version is their modification time.

zliu41:
Updated.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/ModDateTimeDatasetVersionFinder.java
ibuenros: You can actually use `SingleVersionFinder` for this. It uses FileStatusDatasetVersion which does not require an additional call to `fs.filestatus`

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/PathDateTimeDatasetVersionFinder.java
```diff
@@ -27,19 +27,20 @@
...
 
 /**
- * {@link gobblin.data.management.retention.version.finder.DatasetVersionFinder} for timestamped datasets. Uses a datetime pattern
- * to find dataset versions and parse the {@link org.joda.time.DateTime} representing the version.
+ * {@link gobblin.data.management.retention.version.finder.DatasetVersionFinder} for datasets based on path timestamps.
+ * Uses a datetime pattern to find dataset versions from the dataset path
+ * and parse the {@link org.joda.time.DateTime} representing the version.
  */
-public class DateTimeDatasetVersionFinder extends DatasetVersionFinder<TimestampedDatasetVersion> {
+public class PathDateTimeDatasetVersionFinder extends DatasetVersionFinder<TimestampedDatasetVersion> {
```
ibuenros: This name is certainly better, but this class might be already used in some running projects. Can we revert it to the original name? Thanks.

zliu41:
updated.

ibuenros:
LGTM

PR #312, opened 2015-08-30
Use MR jobs for non-deduping compaction

zliu41:
Previously if `deduplicate=false`, it uses the Azkaban job to copy the data from hourly folder to daily folder, which is too slow for large number of topics and data volumes.

Now using a map-only job similar as camus-sweeper.

Also add a counter in `AvroKepMapper` to get row count for map-only jobs.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
@@ -62,6 +62,7 @@ import gobblin.util.HadoopUtils;
 public abstract class MRCompactorJobRunner implements Callable<Void> {
 
   private static final Logger LOG = LoggerFactory.getLogger(MRCompactorJobRunner.class);
+  private static final String HADOOP_JOB_NAME = "Gobblin MR Compaction";
 
   protected final State jobProps;
   protected final String topic;
@@ -70,6 +71,7 @@ public abstract class MRCompactorJobRunner implements Callable<Void> {
   protected final Path outputPath;
   protected final FileSystem fs;
   protected final FsPermission perm;
+  protected final boolean deduplicate;
 
   protected MRCompactorJobRunner(State jobProps, FileSystem fs) {
     this.jobProps = jobProps;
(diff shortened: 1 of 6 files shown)


zliu41:
@liyinan926 can you have a look?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyCompactorOutputCommitter.java
```diff
@@ -60,8 +59,18 @@ public void commitTask(TaskAttemptContext context) throws IOException {
...
     if (fs.exists(workPath)) {
-      long recordCount = getRecordCountCounter(context);
-      String fileName = "part-r-" + recordCount + "." + System.currentTimeMillis() + "."
+      long recordCount = getRecordCountCounter(context, AvroKeyDedupReducer.EVENT_COUNTER.RECORD_COUNT);
+      String fileNamePrefix;
+      if (recordCount == 0) {
+
+        // recordCount == 0 indicates that it is a map-only, non-dedup job, and thus record count should
+        // be obtained from mapper counter.
+        fileNamePrefix = "part-m-";
```
liyinan926: It's better to define such string literals as constants.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyCompactorOutputCommitter.java
```diff
@@ -79,17 +88,16 @@ public boolean accept(Path path) {
     super.commitTask(context);
   }
 
-  private long getRecordCountCounter(TaskAttemptContext context) {
+  private long getRecordCountCounter(TaskAttemptContext context, Enum<?> counterName) {
```
liyinan926: Should this be renamed to `getRecordCountFromCounter`?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -102,44 +99,32 @@ public Void call() throws IOException, ClassNotFoundException, InterruptedExcept
...
-          if (FilenameUtils.isExtension(status.getPath().getName(), getApplicableFileExtensions())) {
-            filePaths.add(status.getPath());
-          }
-        }
-        this.copyDataFiles(this.tmpPath, filePaths, conf);
-      }
+      addJars(conf);
+      Job job = Job.getInstance(conf);
+      this.configureJob(job);
+      this.submit(job);
```
liyinan926: `submitAndWait` seems a better name as it indeed waits for the job to complete.

liyinan926:

PR #314, opened 2015-08-31
Add locks to ParallelRunner

zliu41:
Reason: https://github.com/linkedin/gobblin/issues/313
An alternative approach is to modify `AbstractJobLauncher` to avoid submitting the same directory for cleaning more than once. I'd prefer the lock approach since it is cleaner and makes it easier for future classes that use `ParallelRunner`.


zliu41:
@chavdar @sahilTakiar can you take a look?

chavdar:
LGTM.

PR #315, opened 2015-08-31
Adding FsPermission serialize / de-serialize methods

sahilTakiar:
@ibuenros and @pcadabam could you both take a look?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
@@ -115,13 +115,9 @@ public abstract class FsDataWriter<D> implements DataWriter<D>, FinalState {
         ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_BLOCK_SIZE, numBranches, branchId),
         this.fs.getDefaultBlockSize(this.outputFile));
 
-    this.filePermission = new FsPermission(properties.getPropAsShortWithRadix(
-        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_PERMISSIONS, numBranches, branchId),
-        FsPermission.getDefault().toShort(), ConfigurationKeys.PERMISSION_PARSING_RADIX));
+    this.filePermission = HadoopUtils.deserializeWriterFilePermissions(properties, numBranches, branchId);
 
-    this.dirPermission = new FsPermission(properties.getPropAsShortWithRadix(
-        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_DIR_PERMISSIONS, numBranches, branchId),
-        FsPermission.getDefault().toShort(), ConfigurationKeys.PERMISSION_PARSING_RADIX));
+    this.dirPermission = HadoopUtils.deserializeWriterDirPermissions(properties, numBranches, branchId);
 
     this.stagingFileOutputStream = this.closer.register(this.fs.create(this.stagingFile, this.filePermission, true,
         this.bufferSize, this.replicationFactor, this.blockSize, null));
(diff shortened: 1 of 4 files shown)


gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -168,4 +170,32 @@ public static Writable deserializeFromString(Class<? extends Writable> writableC
...
+  }
+
+  /**
+   * Deserializes a {@link FsPermission} object from the given {@link State}. This method is meant to be used in
+   * conjunction with {@link #serializeFsPermission(State, String, FsPermission)} or
+   * {@link #serializeOctalPermissions(State, String, String)}.
+   */
+  public static FsPermission deserializeFsPermission(State state, String key) {
+    return new FsPermission(state.getPropAsShortWithRadix(key,
+        FsPermission.getDefault().toShort(), ConfigurationKeys.PERMISSION_PARSING_RADIX));
```
ibuenros: This should not use `ConfigurationKeys.PERMISSION_PARSING_RADIX` as the serializer is hard coded to 8.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -168,4 +170,32 @@ public static Writable deserializeFromString(Class<? extends Writable> writableC
       closer.close();
     }
   }
+
+  /**
+   * Serializes a given {@link FsPermission} object into a {@link String} and stores it as a key, value pair in the
+   * given {@link State}. This method is meant to be used in conjunction with
+   * {@link #deserializeFsPermission(State, String)}.
+   */
+  public static void serializeFsPermission(State state, String key, FsPermission fsPermissions) {
```
pcadabam: Do you want to expose the key to the clients? This client 'may' misuse the key to store something else.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -168,4 +170,32 @@ public static Writable deserializeFromString(Class<? extends Writable> writableC
...
+    state.setProp(key, String.format("%04o", fsPermissions.toShort()));
+  }
+
+  /**
+   * Serializes a given {@link String} that represents the permissions of a file / directory in octal notation. The
+   * {@link String} is stored in the given {@link State}. This method is meant to be used in conjunction with
+   * {@link #deserializeFsPermission(State, String)}.
+   */
+  public static void serializeOctalPermissions(State state, String key, String octalPermissions) {
```
ibuenros: not: can this be renamed to `setOctalPermissions`? It is not actually serializing anything.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -115,13 +115,14 @@ public FsDataWriter(State properties, String fileName, int numBranches, int bran
...
-    this.dirPermission = new FsPermission(properties.getPropAsShortWithRadix(
-        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_DIR_PERMISSIONS, numBranches, branchId),
-        FsPermission.getDefault().toShort(), ConfigurationKeys.PERMISSION_PARSING_RADIX));
+    this.filePermission =
+        HadoopUtils.deserializeFsPermission(properties, ForkOperatorUtils.getPropertyNameForBranch(
+            ConfigurationKeys.WRITER_FILE_PERMISSIONS, numBranches, branchId));
+
+    this.dirPermission =
+        HadoopUtils
+            .deserializeFsPermission(properties, ForkOperatorUtils.getPropertyNameForBranch(
```
pcadabam: Same here. I like the clients to be unaware of the key

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: Thats fine though right, since the short is converted in the `String.format` code block. So radix = 8 should handle setting the short and setting in octal.

sahilTakiar:
Addressed comments.

pcadabam:
LGTM

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
ibuenros: radix 8 will work, yes. But someone in the future may decide to change the configuration in `ConfigurationKeys` and break everything.

PR #318, opened 2015-09-02
Create writer when there are no records

pcadabam:
Tested this change with IDPC testharness.
@liyinan926 @ibuenros  can you take a look at this?

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -220,6 +220,8 @@ public class ConfigurationKeys {
   public static final String WRITER_PRESERVE_FILE_NAME = WRITER_PREFIX + ".preserve.file.name";
   public static final String WRITER_DEFLATE_LEVEL = WRITER_PREFIX + ".deflate.level";
   public static final String WRITER_CODEC_TYPE = WRITER_PREFIX + ".codec.type";
+  public static final String WRITER_EAGER_INITIALIZATION_KEY = WRITER_PREFIX + ".eager.initialization";
+  public static final boolean DEFAULT_EAGER_WRITER_INITIALIZATION = false;
 
   // Deprecated. Use WRITER_PARTITION_COLUMNS
   @Deprecated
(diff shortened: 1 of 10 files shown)


liyinan926:
@pcadabam please hold on this until we have a thorough understanding the problems and requirements here.

pcadabam:
@liyinan926 I have updated the pr

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -355,10 +355,25 @@ public void close() throws IOException {
     return new InstrumentedDataWriterDecorator<Object>(writer, this.taskState);
   }
 
+  private void buildWriterIfNotPreset() throws IOException {
```
liyinan926: Can you rename this to `buildWriterIfNotPresent()`?

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -355,10 +355,25 @@ public void close() throws IOException {
...
+      }
+    }
+  }
+
   /**
    * Get new records off the record queue and process them.
    */
   private void processRecords() throws IOException, DataConversionException {
+
+    if (this.taskState.getPropAsBoolean(ConfigurationKeys.WRITER_CREATE_EMPTY_FILE_KEY)) {
```
liyinan926: I suggest defining the default (`false`) explicitly in `ConfigurationKeys` and use the default here.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -220,6 +220,7 @@
   public static final String WRITER_PRESERVE_FILE_NAME = WRITER_PREFIX + ".preserve.file.name";
   public static final String WRITER_DEFLATE_LEVEL = WRITER_PREFIX + ".deflate.level";
   public static final String WRITER_CODEC_TYPE = WRITER_PREFIX + ".codec.type";
+  public static final String WRITER_CREATE_EMPTY_FILE_KEY = WRITER_PREFIX + ".create.empty.file";
```
liyinan926: I suggest renaming this to `writer.lazy.initialization`

pcadabam:
addressed comments

liyinan926:
LGTM.

PR #319, opened 2015-09-04
Changed start_time & end_time from TIMESTAMP to DATETIME

omok:
To work around this error:
Data truncation: Incorrect datetime value: '1970-01-01 00:00:00' for column 'end_time'.

According to http://dev.mysql.com/doc/refman/5.6/en/datetime.html, The TIMESTAMP data type is used for values that contain both date and time parts. TIMESTAMP has a range of '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07' UTC.

So when DatabaseJobHistoryStore declares DEFAULT_TIMESTAMP = "1970-01-01 00:00:00", it bombs.

Does LinkedIn have an internal version of MySQL with a different TIMESTAMP range?

### DIFF
--- a/gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
+++ b/gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
@@ -14,8 +14,8 @@
 CREATE TABLE IF NOT EXISTS gobblin_job_executions (
 	job_name VARCHAR(128) NOT NULL,
 	job_id VARCHAR(128) NOT NULL,
-	start_time TIMESTAMP,
-	end_time TIMESTAMP,
+	start_time DATETIME,
+	end_time DATETIME,
 	duration BIGINT(21),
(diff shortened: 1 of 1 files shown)


liyinan926:
@omok have you tested this?

omok:
@liyinan926 Yes.  This schema change allows Gobblin writes to the gobblin_job_executions & gobblin_task_executions tables on my MySQL instance.

liyinan926:
@omok one benefit of using `TIMESTAMP` is that MySQL converts `TIMESTAMP` values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. It seems the zero value `0000-00-00 00:00:00` is acceptable and seems a good choice for a default value. Can you try using that for `DEFAULT_TIMESTAMP`?

liyinan926:
@omok any update on this?

lbendig:
@omok We also bumped into this issue. Now fixed in [pull-338](https://github.com/linkedin/gobblin/pull/338).

omok:
Thanks @lbendig !

PR #326, opened 2015-09-14
Include complete stacktrace in task state

pcadabam:
@ibuenros please take a look


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyMapper.java
```diff
@@ -24,6 +24,8 @@
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.mapreduce.Mapper;
 
+import gobblin.compaction.mapreduce.avro.AvroKeyDedupReducer.EVENT_COUNTER;
```
chavdar: I am not sure why this is showing up in the diff. Can you make sure that your branch is up-to-date with the master?

ibuenros:
LGTM

PR #327, opened 2015-09-15
Adding additional logging in case an ExecutorService cannot shutdown all its threads

sahilTakiar:
@liyinan926 can you take a look.

We're adding this because there was a recent Gobblin map task that hung for over 40 hours. ETL-3550 has more info, but long story short, the theory is that not all threads got shut down properly, causing the JVM to hang.

This PR should better help us debug this issue if it occurs again.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
@@ -84,27 +84,48 @@ public class ExecutorsUtils {
    * </p>
    *
    * @param executorService the {@link ExecutorService} to shutdown
+   * @param logger an {@link Optional} wrapping the {@link Logger} that is used to log metadata of the executorService
+   *               if it cannot shutdown all its threads
    * @param timeout the maximum time to wait for the {@code ExecutorService} to terminate
    * @param unit the time unit of the timeout argument
    */
-  public static void shutdownExecutorService(ExecutorService executorService, long timeout, TimeUnit unit) {
+  public static void shutdownExecutorService(ExecutorService executorService, Optional<Logger> logger, long timeout,
+      TimeUnit unit) {
     Preconditions.checkNotNull(unit);
     // Disable new tasks from being submitted
     executorService.shutdown();
+
+    if (logger.isPresent()) {
+      logger.get().info("Attempting to shutdown ExecutorSerivce: " + executorService);
+    }
+
     try {
       long halfTimeoutNanos = TimeUnit.NANOSECONDS.convert(timeout, unit) / 2;
       // Wait for half the duration of the timeout for existing tasks to terminate
       if (!executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS)) {
         // Cancel currently executing tasks
         executorService.shutdownNow();
+
+        if (logger.isPresent()) {
+          logger.get().info("Shutdown un-successful, attempting shutdownNow of ExecutorService: " + executorService);
+        }
+
         // Wait the other half of the timeout for tasks to respond to being cancelled
-        executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS);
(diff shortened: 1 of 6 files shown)


gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -98,7 +103,9 @@ public static void shutdownExecutorService(ExecutorService executorService, long
         // Cancel currently executing tasks
         executorService.shutdownNow();
         // Wait the other half of the timeout for tasks to respond to being cancelled
-        executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS);
+        if (!executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS)) {
+          log.info("Could not shutdown all threads in ExecutorService: " + executorService.toString());
```
chavdar: nit: no need for toString()

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -28,6 +30,7 @@
  *
  * @author ynli
  */
+@Slf4j
```
chavdar: Can you just declare a log field instead adding yet another external dependency?

liyinan926: Can we avoid using lombok for this? The savings it brings (one line of code) does not justify the amount of confusion it brings in at the same time. I prefer defining a constant `LOGGER` explicitly.

sahilTakiar:
Updated, removed use of lombok. Changed the method to take in an `Optional<Logger>` to be consistent with the other methods.

liyinan926:
LGTM.

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -98,7 +101,9 @@ public static void shutdownExecutorService(ExecutorService executorService, long
         // Cancel currently executing tasks
         executorService.shutdownNow();
         // Wait the other half of the timeout for tasks to respond to being cancelled
-        executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS);
+        if (!executorService.awaitTermination(halfTimeoutNanos, TimeUnit.NANOSECONDS) && logger.isPresent()) {
+          logger.get().info("Could not shutdown all threads in ExecutorService: " + executorService);
```
chavdar: actually, this should be an ERROR level.

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -84,10 +84,13 @@ public static ThreadFactory newThreadFactory(Optional<Logger> logger, Optional<S
...
+   *               if it cannot shutdown all its threads
    * @param timeout the maximum time to wait for the {@code ExecutorService} to terminate
    * @param unit the time unit of the timeout argument
    */
-  public static void shutdownExecutorService(ExecutorService executorService, long timeout, TimeUnit unit) {
+  public static void shutdownExecutorService(ExecutorService executorService, Optional<Logger> logger, long timeout,
+      TimeUnit unit) {
     Preconditions.checkNotNull(unit);
     // Disable new tasks from being submitted
     executorService.shutdown();
```
chavdar: White you are at it, you can probably add a logging statement here.

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -98,7 +101,9 @@ public static void shutdownExecutorService(ExecutorService executorService, long
         // Cancel currently executing tasks
         executorService.shutdownNow();
```
chavdar: and here.

sahilTakiar:
@chavdar and @liyinan926 addressed all comments

chavdar:
LGTM

liyinan926:
LGTM.

liyinan926:
LGTM.

PR #329, opened 2015-09-15
Hourly Compaction

zliu41:
How this works:

- There is a `FolderState` associated with each input folder. Possible values are `YES`, `NO`, `PENDING`, `COMPLETE`. Initially the status of all input folders is `NO`.
- When compaction starts, it submits data completeness verification requests in a thread pool, one request per input folder, in the order of their priorities (high priority = 3.0, normal = 2.0, low = 1.0). After data completeness verification for a folder starts, the folder status changes to `PENDING`.
- Meanwhile, a job submitter submits compaction MR jobs for the input folders to another thread pool. The compaction MR job of an input folder is only submitted if the folder status is `PENDING` or `YES`, and there is not already a compaction MR job running for this input folder. If the status is `NO`, there's no point to run the compaction MR job since we don't know if the data is complete or not. If the status is `COMPLETE`, it means compaction for this input folder has finished (which may have either succeeded or failed), and the job submitter will no longer submit MR jobs for this input folder. If there's already a compaction MR job running for this input folder, the job submitter will not submit it again until the current run finishes.
 - If, when a compaction finishes, the status of the input folder is still `PENDING`, it will not publish the data and will discard the work it has done.
- Once the data completeness verification for a folder finishes:
 - if the data is complete, the status of the folder is changed to `YES`, and the thread which is running the compaction job for this folder (if exists) is notified, which will proceed to publish the data. 
 - if the data is incomplete, the status of the folder is changed to `NO`, and the thread which is running the compaction job for this folder will be asked to terminate asap without publishing the data. Another data completeness verification for this folder will be submitted again with priority reduced by 1/3.
- Once the compaction MR job for a folder finishes, if the job successfully published the data, or failed for reasons other than data incompleteness, the status of the input folder will be changed to COMPLETE, and the job submitter will no longer submit jobs for this input folder. Otherwise, the job submitter will attempt to re-submit the compaction MR job for this input folder with priority reduced by 1/3 (it will not re-submit the job until the input folder status is changed to PENDING or YES, as mentioned before).


zliu41:
@liyinan926 @chavdar Could you please review?

gobblin-compaction/src/main/java/gobblin/compaction/Compactor.java
```diff
@@ -26,5 +26,5 @@
    * A method for data compaction, e.g., merge a snapshot table and a delta table.
    * @throws IOException
    */
-  public void compact() throws IOException;
+  public void compact() throws Exception;
```
chavdar: Please don't use Exception in the throws clause as it defeats the purpose of static exception checking. List the type of exceptions that can happent.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -46,16 +50,46 @@
...
+  private static final double PRIORITY_REDUCTION_FACTOR = 1.0 / 3.0;
+  private static final long COMPACTION_JOB_WAIT_INTERVAL_SECONDS = 10;
+  private static final Map<String, Job> RUNNING_MR_JOBS = Maps.newConcurrentMap();
+
+  public enum FolderState {
+
+    // An input folder with status NO indicates that the verification of its data completeness has not started.
+    // If data completeness verification is enabled, compaction should not start on an input folder if its status is
+    // NO, since we won't know whether its data is complete or not.
+    NO,
```
chavdar: Please rename to something more meaningful like VERIFICATION_PENDING.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -46,16 +50,46 @@
...
+
+    // PENDING indicates that the data completeness verification of a folder has started but not yet finished.
+    // Compaction may start on a folder with PENDING status. If, when the compaction job finishes, its status is still
+    // PENDING, the job won't publish data. If during the compaction job, the data verification finishes and the status
+    // changes to NO, the compaction job will be killed.
+    PENDING,
```
chavdar: How about VERIFICATION_STARTED

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -46,16 +50,46 @@
...
+
+    // YES indicates that data of an input folder is complete.
+    YES,
```
chavdar: How about VERIFICATION_PASSED

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -46,16 +50,46 @@
...
+
+    // COMPLETE indicates that compaction of an input folder has finished, which may have either succeeded or failed.
+    COMPLETE
```
chavdar: How about COMPACTION_COMPLETE

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+ * right after data completeness verification starts for the input folder, and
+ * {@link #afterVerificationFinishes(boolean)} should be called right after data completeness verification finishes
+ * for the input folder.
+ *
+ * @author ziliu
+ */
+public abstract class AbstractDataCompletenessVerifier
+    implements Runnable, Comparable<AbstractDataCompletenessVerifier> {
+
+  protected final MRCompactor mrCompactor;
```
chavdar: I don't think we have the right abstractions here. The verifier should not be modifying the state of the compactor directly. I suggest that you define a callback interface which can have the same methods. MRCompactor  can pass an implementation of that interface which will modify its state.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+  protected final String inputFolder;
+  protected final State jobProps;
+  protected final double priority;
+
+  public AbstractDataCompletenessVerifier(MRCompactor compactor, String inputFolder, State jobProps, Double priority) {
```
liyinan926: `Double` => `double`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+    this.mrCompactor = compactor;
+    this.inputFolder = inputFolder;
+    this.jobProps = jobProps;
```
chavdar: Why is that needed? I don't see it being used any where.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+  public String inputFolder() {
+    return this.inputFolder;
+  }
+
+  public double priority() {
+    return this.priority;
+  }
+
+  /**
+   * This method should be called by subclasses right after data completeness verification starts for the input folder.
```
liyinan926: I would suggest implementing the `run` like the following so this and the method below are guaranteed to be called. The actual logic of verification should be implemented by the subclasses in the method `verify`.

```
public void run() {
  afterVerificationStarts();
  boolean dataComplete = verify();
  afterVerificationFinishes(dataComplete);    
}

protected abstract boolean verify();
```

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
liyinan926: I would further suggest creating two enums, one named `VerificationState` and the other named `CompactionState`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -65,8 +99,14 @@
   private final String destSubDir;
   private final String tmpDir;
   private final FileSystem fs;
-  private final ExecutorService executorService;
-  private final List<Future<?>> futures;
+  private final Map<String, State> inputFolderJobProps;
```
chavdar: Please define a record type instead of having three different maps with the same key.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -78,8 +118,18 @@ public MRCompactor(Properties props) throws IOException {
...
     this.tmpDir = getTmpDir();
     this.fs = getFileSystem();
-    this.executorService = createExecutorService();
-    this.futures = Lists.newArrayList();
+    this.inputFolderJobProps = Maps.newConcurrentMap();
+    this.inputFolderPriorities = Maps.newConcurrentMap();
+    this.inputFolderStates = Maps.newConcurrentMap();
+    this.foldersBeingProcessed = Sets.newConcurrentHashSet();
+    this.jobExecutor = createJobExecutor();
+    if (shouldVerifyDataCompleteness()) {
```
chavdar: We should allow controlling the completeness checks on a per-dataset level.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -278,40 +319,286 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+    }
+    this.submitCompactionJobsAndWaitForCompletion();
+  }
+
+  private void verifyDataCompleteness(String inputFolder) {
+    this.dataComplVerifExecutor.get().execute(getDataComplVerifier(inputFolder, getInputFolderPriority(inputFolder)));
+  }
+
+  private boolean shouldVerifyDataCompleteness() {
+    return this.state.getPropAsBoolean(ConfigurationKeys.COMPACTION_SHOULD_VERIFY_DATA_COMPLETENESS,
```
chavdar: Please don't check at the configuration properties map in methods. Check only in the constructor and initialize the necessary member variables. Otherwise, figuring out all configuration that is used by a class is a major PITA. That is especially true when the configs have default values.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -475,6 +475,15 @@
   public static final String COMPACTION_OVERWRITE_OUTPUT_DIR = COMPACTION_PREFIX + "overwrite.output.dir";
   public static final boolean DEFAULT_COMPACTION_OVERWRITE_OUTPUT_DIR = false;
   public static final String COMPACTION_JARS = COMPACTION_PREFIX + "jars";
+  public static final String COMPACTION_DATA_COMPLETENESS_VERIFICATION_THREAD_POOL_SIZE =
```
chavdar: These configs are specific to the MRCompactor. Please declare them in the class that uses them.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
liyinan926: You can also take a look at Guava's `EventBus` class and possible use that as the channel for notification between a verifier and the compactor.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -278,40 +319,286 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+            setInputFolderState(jobRunner.getInputPath().toString(), COMPLETE);
+          } else {
+            reduceInputFolderPriority(jobRunner.getInputPath().toString());
+          }
+        } else {
+          setInputFolderState(jobRunner.getInputPath().toString(), COMPLETE);
+          MRCompactor.this.compactionJobThrowables.put(jobRunner.getInputPath().toString(), t);
+        }
+      } else {
+        LOG.error(String.format("Runnable expected to be instance of %s, actual %s",
```
liyinan926: You probably want to use a `Preconditions.checkArgument()` to make sure the right type is seen here.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -278,40 +319,286 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+  }
+
+  private void submitCompactionJobsAndWaitForCompletion() {
+    LOG.info("Submitting compaction jobs. Number of input folders: " + MRCompactor.this.foldersBeingProcessed.size());
+
+    long timeoutMinutes = getCompactionTimeoutMinutes();
+    Stopwatch sw = Stopwatch.createStarted();
+
+    while (!MRCompactor.this.foldersBeingProcessed.isEmpty()) {
+      for (Iterator<String> it = MRCompactor.this.foldersBeingProcessed.iterator(); it.hasNext();) {
```
liyinan926: You can use a `while` loop here that will look more natural.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -278,40 +319,286 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+      LOG.error("Error processing input folder " + entry.getKey(), entry.getValue());
+    }
+    if (!this.compactionJobThrowables.isEmpty()) {
+      throw new RuntimeException("Failed to process some input folders.");
+    }
+  }
+
+  private void shutdownExecutors() {
+    LOG.info("Shutting down Executors");
+    this.jobExecutor.shutdownNow();
```
liyinan926: You probably can use `ExecutorUtils.shutdownExecutorService()`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -262,4 +308,33 @@ private void moveTmpPathToOutputPath() throws IOException {
       throw new IOException(String.format("Unable to move %s to %s", this.tmpPath, this.outputPath));
     }
   }
+
+  /**
+   * Tell the MRCompactorJobRunner that it can go ahead and publish the data.
+   */
+  public void go() {
```
liyinan926: Will it possible for more than one threads to call `go` or `stop` below?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -46,16 +50,46 @@
...
 
   private static final Logger LOG = LoggerFactory.getLogger(MRCompactor.class);
-  private static final List<Job> RUNNING_MR_JOBS = Lists.newCopyOnWriteArrayList();
+
+  private static final double HIGH_PRIORITY = 3.0;
+  private static final double NORMAL_PRIORITY = 2.0;
+  private static final double LOW_PRIORITY = 1.0;
+  private static final double PRIORITY_REDUCTION_FACTOR = 1.0 / 3.0;
+  private static final long COMPACTION_JOB_WAIT_INTERVAL_SECONDS = 10;
+  private static final Map<String, Job> RUNNING_MR_JOBS = Maps.newConcurrentMap();
```
liyinan926: This probably should just be `final` instead of `static final`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -125,30 +176,23 @@ private int getThreadPoolSize() {
...
+      LOG.error("Caught throwable during compaction", t);
+      throw Throwables.propagate(t);
     } finally {
-      this.executorService.shutdown();
       try {
-        executorService.awaitTermination(getMRJobTimeoutValue(), TimeUnit.MINUTES);
+        shutdownExecutors();
+      } finally {
         deleteDependencyJars();
-        for (Future<?> future : this.futures) {
```
liyinan926: Instead of storing `Throwable`s in `compactionJobThrowables` and check them in `checkForThrowables`, you can probably use `ListeningFuture` and `ListenableFuture` and the callback mechanism around it. Take a look at https://github.com/google/guava/wiki/ListenableFutureExplained.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
chavdar: I don't think the verifier object should be on a per-folder/dataset level. This precludes the ability to pool multiple completeness checks to the audit backend.

I also think we should decouple the completeness read from the check if the completeness value meets the publish threshold. The former may also be useful for other purposes like monitoring.

I suggest you define a DatasetAuditCountReader interface or something along that lines. It should define a method like ListenableFuture<Long> readAuditCounts(dataset). ListenableFuture allows to add a callback on success or failure which can then modify the state of MRCompactor. Whether there is a threadpool to run the audit count requests there is some kind of elevator algorithm or so should be completely up to the reader implementation. MRCompactor shouldn't be concerned with thread pools and so on.

You can also have a DatasetAuditCountCompletenessChecker which has a checkCompleteness(dataset, count) method to verify the completeness. This should be fairly light-weight and should probably be done synchronously. 

With this model, I don't think you need afterVerificationStarts() and  afterVerificationFinishes(). The process should be driven by MRCompactor and it could provide the necessary callbacks to drive the state transitions.

I also suggest that you add some abstraction around the input folder. I am not sure what this folder is. Is it an HDFS folder? What is the namenode, etc. I think the interfaces should use a dataset interface instead. It can have the attributes that we discussed for the dataset publication. Also consider the use case of running compaction on multiple folders.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
zliu41: `afterVerificationStarts();` should be called after the `verify()` starts, not before.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
zliu41: No, only one thread verifies data completeness for a folder.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
zliu41: Then I'll need to pass an instance of MRCompactor to `MRCompactorJobRunner` which creates the Hadoop job.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
zliu41: `Future` is no longer used. I need to use `ThreadPoolExecutor.execute()` which does not return `Future`, since `submit()` cannot prioritize the tasks using a priority queue.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
zliu41: Some implementations of data completeness verifier may want to use properties in `jobProps` in verifying completeness..

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
zliu41: `AbstractDataCompletenessVerifier` doesn't contradict with decoupling read and check. It's just an abstract class; its subclasses can first read from kafka audit, then check. Those implementations will be in an RB since they are internal.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
zliu41: I liked YES and NO from aesthetical point of view but it's fine if you prefer VERIFICATION_PASSED

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
zliu41: `Double` is used instead of `double` since `getDeclaredConstructors()` doesn't support primitive types.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
liyinan926: Then can we move this `Map` to `MRCompactorJobRunner` which seems a better place to host it?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
chavdar: IMO, the problem with YES and NO  is that you can't really know what they mean unless you read the docs. I think it is better when the code is self-explanatory.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
chavdar: To me AbstractDataCompletenessVerifier does not define a meaningful API for classes to override/implement. There are just some conventions which are very much tied to the MRCompactor implementation. We need an interface independent of MRCompactor.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/AbstractDataCompletenessVerifier.java
chavdar: Then the implementations that may need it should just declare it.

zliu41:
@chavdar @liyinan926 updated (cbc38b6).
---Summary---

Big changes:
- Added `gobblin.compaction.Dataset` class. This is a temporary solution; after Gobblin dataset is merged, I'll put it into the class hierarchy.
- In addition to verifying data completeness using `DataCompletenessVerifier` (via Kafka audit), the `MRCompactorJobRunner` also performs a self-verification: when compaction of a dataset completes, if it finds new files that were added to the input folder after compaction started, it will discard the compacted data and re-do the compaction.
- The compaction MR job of a dataset is no longer required to start after the data completeness verification of the dataset starts. Now it submits compaction MR jobs to one thread pool, and the completeness verification jobs to another thread pool, with no particular order between them.
- `MRCompactor` no longer manages the thread pool for data completeness verification. That thread pool is now managed by `DataCompletenessVerifier`.
- `DataCompletenessVerifier` no longer holds a reference to `MRCompactor`. Instead, it returns a `ListenableFuture<Results>` object, which `MRCompactor` can use to add callbacks and modify its states. This also means that data verification is no longer performed in the order of dataset priorities, since ExecutorService.submit() doesn't support prioritization through a priority queue (compaction MR jobs are stilled ordered by dataset priorities).
- Added a `SimulatedDataCompletenessVerifier` class which randomly makes decisions.

Small changes:
- Now it is possible to verify the completeness of multiple datasets together by using `DataCompletenessVerifier.verify(Iterable<Dataset>)`. This enables multiple datasets to share a single SQL query against Kafka audit.
- Now it is possible to specify which datasets need completeness verification by using a whitelist and a blacklist.
- Now one can specify the timeout for verifying data completeness, e.g. 30min. If after 30 minutes from the time the compaction starts, there are still some datasets whose completeness are not verified, those datasets will be given up, and their running verification jobs (if exists) will be killed. For these datasets, it is configurable whether to not compact them, or compact them anyway without completeness guarantee.
- Addressed other comments in the review.

zliu41:
@liyinan926 updated 7edb31a

gobblin-compaction/src/main/java/gobblin/compaction/verify/DataCompletenessVerifier.java
```diff
@@ -0,0 +1,163 @@
...
+    this.exeSvc = getExecutorService();
+  }
+
+  private ListeningExecutorService getExecutorService() {
+    int threadPoolSize = getDataCompletenessVerificationThreadPoolSize();
+    return MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(threadPoolSize));
+  }
+
+  private int getDataCompletenessVerificationThreadPoolSize() {
+    return this.props.getPropAsInt(ConfigurationKeys.COMPACTION_DATA_COMPLETENESS_VERIFICATION_THREAD_POOL_SIZE,
```
chavdar: This should be a member variable in the constructor. Please avoid accessing the configuration object outside the constructor. It makes it very difficult to figure out all configuration properties. It is also error prone if there are default values involved.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -470,20 +469,32 @@
...
-  public static final String COMPACTION_JOB_LATE_DATA_MOVEMENT_TASK = COMPACTION_PREFIX + "job.late.data.movement.task";
-  public static final String COMPACTION_JOB_LATE_DATA_FILES = COMPACTION_PREFIX + "job.late.data.files";
-  public static final String COMPACTION_COMPLETE_FILE_NAME = "_COMPACTION_COMPLETE";
-  public static final String COMPACTION_LATE_FILES_DIRECTORY = "late";
-  public static final String COMPACTION_ENABLE_SUCCESS_FILE = "mapreduce.fileoutputcommitter.marksuccessfuljobs";
   public static final String COMPACTION_OVERWRITE_OUTPUT_DIR = COMPACTION_PREFIX + "overwrite.output.dir";
   public static final boolean DEFAULT_COMPACTION_OVERWRITE_OUTPUT_DIR = false;
-  public static final String COMPACTION_JARS = COMPACTION_PREFIX + "jars";
-  public static final String COMPACTION_TRACKING_EVENTS_NAMESPACE = COMPACTION_PREFIX + "tracking.events";
+  public static final String COMPACTION_DATA_COMPLETENESS_VERIFICATION_THREAD_POOL_SIZE =
```
chavdar: Please bundle configuration properties with the code that uses them rather than dumping them in ConfigurationKeys .

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanCompactionJobLauncher.java
```diff
@@ -42,15 +44,22 @@ public AzkabanCompactionJobLauncher(String jobId, Properties props) {
...
+    this.compactor = compactorClass.getDeclaredConstructor(Properties.class).newInstance(this.properties);
+    this.compactor.compact();
   }
 
   @SuppressWarnings("unchecked")
   private Class<? extends Compactor> getCompactorClass() throws ClassNotFoundException {
-    String compactorClassName =
-        this.properties.getProperty(ConfigurationKeys.COMPACTION_COMPACTOR_CLASS,
-            ConfigurationKeys.DEFAULT_COMPACTION_COMPACTOR_CLASS);
+    String compactorClassName = this.properties.getProperty(ConfigurationKeys.COMPACTION_COMPACTOR_CLASS,
```
chavdar: This logic should be in the constructor.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanCompactionJobLauncher.java
```diff
@@ -42,15 +44,22 @@ public AzkabanCompactionJobLauncher(String jobId, Properties props) {
   @Override
   public void run() throws Exception {
     Class<? extends Compactor> compactorClass = getCompactorClass();
-    compactorClass.getDeclaredConstructor(Properties.class).newInstance(this.properties).compact();
+    this.compactor = compactorClass.getDeclaredConstructor(Properties.class).newInstance(this.properties);
```
chavdar: I'd move this to the constructor and make it final. Constructor should not leave the object in semi-initialized state.

gobblin-compaction/src/main/java/gobblin/compaction/verify/SimulatedDataCompletenessVerifier.java
```diff
@@ -0,0 +1,96 @@
...
+
+import gobblin.compaction.Dataset;
+import gobblin.compaction.verify.DataCompletenessVerifier.Results;
+import gobblin.compaction.verify.DataCompletenessVerifier.Results.Result;
+import gobblin.configuration.State;
+
+
+/**
+ * A class that simulates data completeness verification for Gobblin compaction. It randomly decides
+ * whether the verification of a {@link Dataset} passes or fails.
```
chavdar: What is the purpose of the class? Is it for testing? If yes, In general, having test classes rely on random behavior is not ideal as they make reproducing errors hard. At the very least, there should be enough audit trail about all the random decisions made.

gobblin-compaction/src/main/java/gobblin/compaction/verify/SimulatedDataCompletenessVerifier.java
```diff
@@ -0,0 +1,96 @@
...
+    int maxTime = getMaxVerificationTime();
+    Preconditions.checkState(minTime >= 0, SIMULATED_VERIFIER_MIN_VERIFICATION_TIME_SEC + " should be nonnegative.");
+    Preconditions.checkState(minTime <= maxTime, SIMULATED_VERIFIER_MIN_VERIFICATION_TIME_SEC
+        + " should be no more than " + SIMULATED_VERIFIER_MAX_VERIFICATION_TIME_SEC);
+    long verifTime = RANDOM.nextInt(maxTime - minTime) + minTime;
+    LOG.info("Verifying data completeness for " + this.datasets + ", will take " + verifTime + " sec");
+    try {
+      Thread.sleep(verifTime * 1000);
+    } catch (InterruptedException e) {
+      LOG.error("Interrupted", e);
```
chavdar: Nit: if the wait was interrupted. It's probably best to propagate a runtime exception as something is being shutdown/stopped.

gobblin-compaction/src/main/java/gobblin/compaction/verify/SimulatedDataCompletenessVerifier.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Override
+  public Results call() throws Exception {
+    int minTime = getMinVerificationTime();
+    int maxTime = getMaxVerificationTime();
+    Preconditions.checkState(minTime >= 0, SIMULATED_VERIFIER_MIN_VERIFICATION_TIME_SEC + " should be nonnegative.");
+    Preconditions.checkState(minTime <= maxTime, SIMULATED_VERIFIER_MIN_VERIFICATION_TIME_SEC
+        + " should be no more than " + SIMULATED_VERIFIER_MAX_VERIFICATION_TIME_SEC);
+    long verifTime = RANDOM.nextInt(maxTime - minTime) + minTime;
```
chavdar: typo: s/verifTime/verifyTime/

gobblin-compaction/src/main/java/gobblin/compaction/verify/SimulatedDataCompletenessVerifier.java
liyinan926: Should also do `Thread.currentThread.interrupt()` to restore the interruption flag.

gobblin-compaction/src/main/java/gobblin/compaction/verify/DataCompletenessVerifier.java
```diff
@@ -0,0 +1,163 @@
...
+   * {@link ConfigurationKeys#COMPACTION_DATA_COMPLETENESS_VERIFICATION_THREAD_POOL_SIZE}.
+   */
+  public DataCompletenessVerifier(State props) {
+    this.props = props;
+    this.exeSvc = getExecutorService();
+  }
+
+  private ListeningExecutorService getExecutorService() {
+    int threadPoolSize = getDataCompletenessVerificationThreadPoolSize();
+    return MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(threadPoolSize));
```
chavdar: Using a fixed thread pools ties up resources (threads) when not needed. Have a look at gobblin.util.executors.ScalingThreadPoolExecutor . We should use a thread pool with sizes (0, threadPoolSize) so that the thread pool can grow and shrink as needed.

gobblin-compaction/src/main/java/gobblin/compaction/verify/DataCompletenessVerifier.java
```diff
@@ -0,0 +1,163 @@
...
+   */
+  public ListenableFuture<Results> verify(Iterable<Dataset> datasets) {
+    return this.exeSvc.submit(getRunner(datasets));
+  }
+
+  private AbstractRunner getRunner(Iterable<Dataset> datasets) {
+    Preconditions.checkArgument(this.props.contains(ConfigurationKeys.COMPACTION_DATA_COMPLETENESS_VERIFICATION_CLASS),
+        "Missing required property " + ConfigurationKeys.COMPACTION_DATA_COMPLETENESS_VERIFICATION_CLASS);
+    try {
+      @SuppressWarnings("unchecked")
```
chavdar: The class object creation should happen in the constructor.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -297,39 +368,355 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+        ConfigurationKeys.DEFAULT_COMPACTION_DATA_VERIFICATION_TIMEOUT_MINUTES);
+  }
+
+  private void throwExceptionsIfAnyDatasetCompactionFailed() {
+    Set<Dataset> datasetsWithThrowables = getDatasetsWithThrowables();
+    int numDatasetsWithThrowables = 0;
+    for (Dataset dataset : datasetsWithThrowables) {
+      numDatasetsWithThrowables++;
+      for (Throwable t : dataset.throwables()) {
+        LOG.error("Error processing dataset " + dataset, t);
```
chavdar: We should emit tracking events.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
zliu41: All config properties that can be set by the user are in ConfigurationKeys. All internal properties have been moved to MRCompactor.

zliu41:
@chavdar updated a5a11a7

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
chavdar: That's the part I don't get. What does it buy us? We have so many of these options (363 in my version)  that I doubt anyone will say "Let me see what knobs and levers are out there". More likely, the question is "I need to figure out how to configure component X". Given that the property naming is not very consistent, I think the starting point is always going to be the implementation. I think ConfigurationKeys should be only for keys that are shared across components. This is most likely state properties and not real configuration properties.

I also suggest at renaming the configuration properties so that they share a more specific prefix than compaction, e.g. compaction.completeness_verification or similar. Then, "compaction.should.verify.data.completeness" can become "compaction.completeness_verification.enabled", "compaction.data.completeness.verification.class" can become "compaction.completeness_verification.verificatier_class" and so on. I find the current "English phrase with dots" naming scheme confusing as the same concepts can be expressed in different ways.

This also reminds me of another issue -- please add javadocs for all configuration properties what they mean. Looking just at the configs I don't think I can guess what "compaction.num.datasets.verified.together" does.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -294,39 +373,357 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+        } else {
+
+          // Compaction job of a dataset has aborted because data completeness is not verified.
+          // Reduce priority and try again.
+          jobRunner.getDataset().reducePriority();
+        }
+      } else {
+
+        // Compaction job of a dataset has failed with a throwable.
+        jobRunner.getDataset().setState(COMPACTION_COMPLETE);
```
chavdar: For these kind of state transitions where multiple attributes need to be changed, it is better to have a helperMethod, e.g. completeWithError(Throwable) to make sure that the object is not in an inconsistent state.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -75,57 +98,73 @@
...
-    this.topic = jobProps.getProp(ConfigurationKeys.COMPACTION_TOPIC);
-    this.inputPath = new Path(jobProps.getProp(ConfigurationKeys.COMPACTION_JOB_INPUT_DIR));
-    this.outputPath = new Path(jobProps.getProp(ConfigurationKeys.COMPACTION_JOB_DEST_DIR));
-    this.tmpPath = new Path(jobProps.getProp(ConfigurationKeys.COMPACTION_JOB_TMP_DIR));
+  private volatile Policy policy = Policy.DO_NOT_PUBLISH_DATA;
+  private volatile Status status = Status.RUNNING;
+
+  protected MRCompactorJobRunner(Dataset dataset, FileSystem fs, Double priority) {
+    this.dataset = dataset;
+    this.jobProps = dataset.jobProps();
```
chavdar: This is probably going to change once you integrate with the generic dataset but I think it looks unusual that the dataset is coupled with the job properties.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -294,39 +373,357 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+       * (i.e., verification didn't pass and it didn't timeout).
+       */
+      @Override
+      public void onSuccess(Results results) {
+        List<Dataset> datasetsToBeVerifiedAgain = Lists.newArrayList();
+        for (Results.Result result : results) {
+          Optional<MRCompactorJobRunner> jobRunner =
+              Optional.fromNullable(MRCompactor.this.jobRunnables.get(result.dataset()));
+
+          if (result.status() == DataCompletenessVerifier.Results.Result.Status.PASSED) {
```
chavdar: I think it is better to have a switch statement here. If someone later adds a new state, there will be a warning that there is a case that has not been processed.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -294,39 +373,357 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+          jobRunner.abort();
+        }
+        break;
+      }
+
+      // Sleep for a few seconds before another round
+      try {
+        Thread.sleep(TimeUnit.SECONDS.toMillis(COMPACTION_JOB_WAIT_INTERVAL_SECONDS));
+      } catch (InterruptedException e) {
+        LOG.error("Interruppted while waiting", e);
```
chavdar: InterruptedExceptions should not be swallowed. Most likely someone is trying to shutdown the service. In that case the code should exit gracefully.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -294,39 +373,357 @@ MRCompactorJobPropCreator getJobPropCreator(String topic, Path topicInputDir, Pa
...
+
+  private void throwExceptionsIfAnyDatasetCompactionFailed() {
+    Set<Dataset> datasetsWithThrowables = getDatasetsWithThrowables();
+    int numDatasetsWithThrowables = 0;
+    for (Dataset dataset : datasetsWithThrowables) {
+      numDatasetsWithThrowables++;
+      for (Throwable t : dataset.throwables()) {
+        LOG.error("Error processing dataset " + dataset, t);
+        CompactionSlaEventHelper.populateState(dataset.jobProps(), Optional.<Job> absent(), this.fs);
+        new SlaEventSubmitter(eventSubmitter, "CompactionFailed", dataset.jobProps().getProperties()).submit();
```
chavdar: Actually, reviewing more in depth the rest of the code, I think we need the event emission to happen on certain dataset state transitions, like like VERIFIED, GIVEN_UP, etc.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
zliu41: Yes this will change. I think it makes sense that a compaction dataset has a compaction job props, and it's better than adding another Map in `MrCompactor` from dataset to jobprops.

zliu41:
@chavdar updated 980251b

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
chavdar: Ok. I'll wait to see for the new PR with the integrated datasets. In general, if those are associated with the dataset, they should be "datasetProps", not "jobProps". 

Further,  as an editorial comment, I think we have way too many property collections. It makes it hard to figure out what is the life-cycle of the collections and the properties, which are read/only and which are read/write, and all kinds of other data dependencies. I'd love to see more well-type java objects and interfaces and less property collections.  This will also make us think a little bit more what the API is instead of just dumping stuff into a collection.

chavdar:
LGTM

PR #330, opened 2015-09-15
Fix overflow bug in FileBased partitioning

ahollenbach:
In certain edge case combinations of number of files and partitions, you can run into an issue where the `fileOffset` will exceed the number of `filesToPull` and still attempt to create partitions even though there are no files to add to the partitions. Switched the for loop to loop by `fileOffset` rather than the number of partitions.

Example issue state: 100 partitions, 180 files


ahollenbach:
@sahilTakiar If you could take a look, that'd be great.

sahilTakiar:
LGTM

PR #333, opened 2015-09-16
Simple distcp POC

ibuenros:


gobblin-core/src/main/java/gobblin/source/extractor/hadoop/RawByteExtractor.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+
+/**
+ * Implementation of {@link gobblin.source.extractor.Extractor} that reads a file as a sequence of byte arrays.
+ * Uses a {@link gobblin.source.extractor.utils.BufferedRawByteIterator}.
+ */
+public class RawByteExtractor implements Extractor<String, byte[]> {
```
chavdar: We should consider using ByteBuffers (direct ByteBuffers in particular) to avoid repeated re-allocation of large byte arrays, 0 initialization and the cost of moving those during GC.

chavdar: Ignore that. See my comments to CopySource.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/CopySource.java
```diff
@@ -0,0 +1,206 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Lists;
+
+
+/**
+ * {@link gobblin.source.Source} that generates work units from {@link gobblin.data.management.copy.CopyableDataset}s.
+ * Used for doing byte-level copies of files.
+ */
+public class CopySource extends AbstractSource<String, byte[]> {
```
chavdar: I don't think the record type should be byte[]. I think it should be a custom object that contains:
* An InputStream or FSDataInputStream of the input file. Chances are that the input stream we are reading from is already buffered so we are adding double-buffering overhead. Further, having the InputStream data model will allow us to do some transformation like encrypting/decrypting, compressing/decompressing on-the-fly. I think that FSDataInputStream might be a better fit since it implements Seekable. I think we need that if we want to duplicate input streams (for fork).
* OwnerAndPermission
* Other metadata (e.g. support for extended attributes if we move to a version of hadoop that supports it).

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/CopySource.java
```diff
@@ -0,0 +1,206 @@
...
+      FileSystem originFs = getSourceFileSystem(state);
+      FileSystem targetFs = getTargetFileSystem(state);
+
+      DatasetFinder datasetFinder = DatasetUtils.instantiateDatasetFinder(state.getProperties(), originFs);
+      Collection<Dataset> datasets = datasetFinder.findDatasets();
+
+      for (Dataset dataset : datasets) {
+
+        Collection<Partition<CopyableFile>> partitions;
+        if(dataset instanceof CopyableDataset && dataset instanceof PartitionableDataset &&
```
pcadabam: Can we think of a way to avoid these casts?

- Since we know that this is a CopySource, may be we should use the CopyDatasetFinder instead of DatasetFinder which vends a bunch of CopyableDatasets and avoid cast from Dataset to CopyableDataset. Later implementations can extends CopyableDatasetFinder.

- Also in the current implementation if the dataset is not partitionable, we create a single partition with all the files. Can the CopyableDataset always extend PartitionableDataset? A copyableDataset will return Partitions instead of files. In case there are no partitions, the CopyableDataset will return a single partition will all files. This will avoid a cast to PartitionableDataset.

ibuenros:
Replaced by PR 411

PR #334, opened 2015-09-18
Adding option to submit additional metadata when stopping a timer event

sahilTakiar:
Modified `TimingEvent` and added a new method `stop(Map<String, String> additionalMetadata)` which specifies extra metadata that can be submitted along with the `TimingEvent`. The advantage being that a user may want to emit additional metadata specific to that timing event. I decided to add `additionalMetadata` to the `stop()` method instead of the constructor, because the metadata may not be known until the timing event has completed.

Updated `MRJobLauncher` to record the hadoop job id when it submits a timing event for the length of the MR job.

@ibuenros and @liyinan926 could you review?

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
@@ -12,6 +12,10 @@
 
 package gobblin.metrics.event;
 
+import java.util.Map;
+
+import com.google.common.collect.Maps;
+
 /**
  * Event to time actions in the program. Automatically reports start time, end time, and duration from the time
  * the {@link gobblin.metrics.event.TimingEvent} was created to the time {@link #stop} is called.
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -208,7 +210,8 @@ protected void runWorkUnits(List<WorkUnit> workUnits) throws Exception {
       TimingEvent mrJobRunTimer = this.eventSubmitter.getTimingEvent(TimingEventNames.RunJobTimings.MR_JOB_RUN);
       LOG.info(String.format("Waiting for Hadoop MR job %s to complete", this.job.getJobID()));
       this.job.waitForCompletion(true);
-      mrJobRunTimer.stop();
+      mrJobRunTimer.stop(new ImmutableMap.Builder<String, String>().put("hadoopMRJobId", this.job.getJobID().toString())
```
liyinan926: You can use `ImmutableMap.of("hadoopMRJobId", this.job.getJobID().toString())`.

liyinan926:
LGTM other than the minor comment.

gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
```diff
@@ -36,16 +40,33 @@ public TimingEvent(EventSubmitter submitter, String name) {
...
     long duration = endTime - this.startTime;
-    this.submitter.submit(this.name, EventSubmitter.EVENT_TYPE, TIMING_EVENT, START_TIME, Long.toString(this.startTime),
-        END_TIME, Long.toString(endTime), DURATION, Long.toString(duration));
+
+    Map<String, String> finalMetadata = Maps.newHashMap();
+    finalMetadata.put(EventSubmitter.EVENT_TYPE, TIMING_EVENT);
+    finalMetadata.put(START_TIME, Long.toString(this.startTime));
+    finalMetadata.put(END_TIME, Long.toString(endTime));
+    finalMetadata.put(DURATION, Long.toString(duration));
+    finalMetadata.putAll(additionalMetadata);
```
ibuenros: I would rather put this first, so that if the user sets START_TIME to something different, is is overriden to the correct value.

ibuenros:
Just one comment. LGTM otherwise.

gobblin-metrics/src/main/java/gobblin/metrics/event/TimingEvent.java
sahilTakiar: Yes, good point fixed.

PR #335, opened 2015-09-20
Enable tasks to publish data

zliu41:
Summary:

- Data are published by tasks if `job.commit.policy=partial` and `publish.data.in.tasks=true`.
- Added method `DataPublisher.publish(WorkUnitState)` to publish data for a single task.
- Added a flag in `BaseDataPublisher.publishData()` to control whether to publish the data for a single task. When publishing for a single task, it checks the files against `writer.final.output.path` in the `WorkUnitState` to make sure it doesn't publish data written by other tasks.
- In `AvroHdfsTimePartitionedWriter.close()`, it adds all files created by this writer to `writer.final.output.path`.
- `AvroHdfsTimePartitionedWithRecordCountsWriter.close()` will rewrite `writer.final.output.path` upon renaming output file names to include row counts. It also gets the file names from `FsDataWriter` instead of constructing it itself, which is more consistent.
- `Task` publishes data in the `finally` block since it needs to wait till `closer.close()` is called which closes all the forks and data writers.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
+++ b/gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
@@ -14,6 +14,8 @@ package gobblin.publisher;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 
 import gobblin.configuration.State;
@@ -31,21 +33,47 @@ public abstract class DataPublisher implements Closeable {
     this.state = state;
   }
 
-  public abstract void initialize()
-      throws IOException;
+  /**
+   * @deprecated {@link DataPublisher} initialization should be done in the constructor.
+   */
+  @Deprecated
+  public abstract void initialize() throws IOException;
 
   /**
-   * Returns true if it successfully publishes the data, false otherwise.
+   * Publish the data for the given task. The state must contains property "writer.final.output.file.paths"
+   * (or "writer.final.output.file.paths.[branchId]" if there are multiple branches),
+   * so that the publisher knows which files to publish.
    */
-  public abstract void publishData(Collection<? extends WorkUnitState> tasks)
-      throws IOException;
+  public abstract void publishData(WorkUnitState state) throws IOException;
 
   /**
-   * Returns true if it successfully publishes the metadata, false otherwise. Examples of publishing metadata include
-   * writing offset files, checkpoint files, etc.
+   * Publish the data for the given tasks.
    */
-  public abstract void publishMetadata(Collection<? extends WorkUnitState> tasks)
-      throws IOException;
+  public abstract void publishData(Collection<? extends WorkUnitState> states) throws IOException;
+
+  /**
+   * Publish the metadata (e.g., schema) for the given task. Checkpoints should not be published as part of metadata.
+   * They are published by Gobblin runtime after the metadata and data are published.
+   */
+  public abstract void publishMetadata(WorkUnitState state) throws IOException;
+
+  /**
+   * Publish the metadata (e.g., schema) for the given tasks. Checkpoints should not be published as part of metadata.
+   * They are published by Gobblin runtime after the metadata and data are published.
+   */
+  public abstract void publishMetadata(Collection<? extends WorkUnitState> states) throws IOException;
+
+  /**
+   * First publish the metadata via {@link DataPublisher#publishMetadata(WorkUnitState)}, and then publish the output data
+   * via the {@link DataPublisher#publishData(WorkUnitState)} method.
+   *
+   * @param state is a {@link WorkUnitState}.
+   * @throws IOException if there is a problem with publishing the metadata or the data.
+   */
+  public void publish(WorkUnitState state) throws IOException {
+    publishMetadata(state);
+    publishData(state);
+  }
 
   /**
    * First publish the metadata via {@link DataPublisher#publishMetadata(Collection)}, and then publish the output data
@@ -54,13 +82,26 @@ public abstract class DataPublisher implements Closeable {
    * @param states is a {@link Collection} of {@link WorkUnitState}s.
    * @throws IOException if there is a problem with publishing the metadata or the data.
    */
-  public void publish(Collection<? extends WorkUnitState> states)
-      throws IOException {
+  public void publish(Collection<? extends WorkUnitState> states) throws IOException {
     publishMetadata(states);
     publishData(states);
   }
 
   public State getState() {
     return this.state;
+  }
+
+  /**
+   * Get an instance of {@link DataPublisher}.
+   *
+   * @param dataPublisherClass A concrete class that extends {@link DataPublisher}.
+   * @param state A {@link State} used to instantiate the {@link DataPublisher}.
+   * @return A {@link DataPublisher} instance.
+   */
+  public static DataPublisher getInstance(Class<? extends DataPublisher> dataPublisherClass, State state)
+      throws NoSuchMethodException, SecurityException, InstantiationException, IllegalAccessException,
+      IllegalArgumentException, InvocationTargetException {
+    Constructor<? extends DataPublisher> dataPublisherConstructor = dataPublisherClass.getConstructor(State.class);
+    return dataPublisherConstructor.newInstance(state);
   }
 }
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -15,6 +15,7 @@ package gobblin.publisher;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -28,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -116,6 +118,21 @@ public class BaseDataPublisher extends DataPublisher {
   }
 
   @Override
+  public void publishData(WorkUnitState state) throws IOException {
+    for (int branchId = 0; branchId < this.numBranches; branchId++) {
+      publishSingleTaskData(state, branchId);
+    }
+  }
+
+  /**
+   * This method publishes output data for a single task based on the given {@link WorkUnitState}.
+   * Output data from other tasks won't be published even if they are in the same folder.
+   */
+  private void publishSingleTaskData(WorkUnitState state, int branchId) throws IOException {
+    publishData(state, branchId, true, new HashSet<Path>());
+  }
+
+  @Override
   public void publishData(Collection<? extends WorkUnitState> states) throws IOException {
 
     // We need a Set to collect unique writer output paths as multiple tasks may belong to the same extract. Tasks that
@@ -124,7 +141,7 @@ public class BaseDataPublisher extends DataPublisher {
 
     for (WorkUnitState workUnitState : states) {
       for (int branchId = 0; branchId < this.numBranches; branchId++) {
-        publishData(workUnitState, branchId, writerOutputPathsMoved);
+        publishMultiTaskData(workUnitState, branchId, writerOutputPathsMoved);
       }
 
       // Upon successfully committing the data to the final output directory, set states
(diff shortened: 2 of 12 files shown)


zliu41:
@liyinan926 @chavdar could you review

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -54,8 +82,7 @@ public abstract void publishMetadata(Collection<? extends WorkUnitState> tasks)
    * @param states is a {@link Collection} of {@link WorkUnitState}s.
    * @throws IOException if there is a problem with publishing the metadata or the data.
    */
-  public void publish(Collection<? extends WorkUnitState> states)
-      throws IOException {
+  public void publish(Collection<? extends WorkUnitState> states) throws IOException {
     publishMetadata(states);
```
chavdar: This may lead to data loss. If the metadata includes checkpoints and the publish of data fails, on the next run will use the new checkpoints.

We should publish data first. Generally, duplicate data is better than data loss.

zliu41: metadata doesn't include checkpoint. Checkpoints are published elsewhere (in `JobContext.persistDatasetState`.

I think metadata is published before data because in some implementation of data publisher, it needs to first publish the metadata (e.g., schema) before data can be published.

chavdar: The javadocs for publishMetadata() say "Examples of publishing metadata include writing offset files, checkpoint files, etc.". We need to very precise what constitutes metadata as it has important consequences. Which are these implementations and what metadata is being published?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -233,6 +244,55 @@ public void run() {
...
+
+  private void publishTaskData() throws IOException {
+    Closer closer = Closer.create();
+    try {
+      for (int i = 0; i < this.forks.size(); i++) {
+        if (this.forks.get(i).isPresent()) {
+          LOG.info("Publishing data from task " + this.taskId + " for fork " + i);
+        }
+      }
+      @SuppressWarnings("unchecked")
```
chavdar: nit: I suggest that we move the instantiation of the publisher before the logging above to avoid misleading messages in case that fails.

zliu41:
Updated.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
zliu41: Updated Javadoc.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -31,21 +33,47 @@ public DataPublisher(State state) {
     this.state = state;
   }
 
-  public abstract void initialize()
-      throws IOException;
+  /**
+   * @deprecated DataPublisher initialization should be done in the constructor of DataPublisher.
```
liyinan926: Please use `{@link DataPublisher}`.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -63,4 +90,18 @@ public void publish(Collection<? extends WorkUnitState> states)
   public State getState() {
     return this.state;
   }
+
+  /**
+   * Get an instance of {@link DataPublisher}.
+   *
+   * @param dataPublisherClass A concrete class that extends {@link DataPublisher}.
```
liyinan926: Should be `implements`.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -134,51 +147,65 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
     }
   }
 
-  protected void publishData(WorkUnitState workUnitState, int branchId, Set<Path> writerOutputPathsMoved)
-      throws IOException {
+  private void publishData(WorkUnitState state, int branchId, Set<Path> writerOutputPathsMoved) throws IOException {
+    publishData(state, branchId, false, writerOutputPathsMoved);
```
liyinan926: Why line 128 uses `true` while this uses `false` for the third argument? What's the deciding factor here?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -197,6 +224,29 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
+      WorkUnitState workUnitState, int branchId, ParallelRunner parallelRunner) throws IOException {
+    Preconditions.checkArgument(workUnitState.contains(ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH), "Missing property "
+        + ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH + ", which is needed for publishing data of a single task");
+
+    Iterable<String> taskOutputFiles = workUnitState.getPropAsList(ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH);
+
+    for (String taskOutputFile : taskOutputFiles) {
+      Path taskOutputPath = new Path(taskOutputFile);
+      if (!this.fileSystemByBranches.get(branchId).exists(taskOutputPath)) {
+        continue;
```
liyinan926: Better to add a warn log message here.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -214,7 +264,7 @@ protected void addWriterOutputToExistingDir(Path writerOutputDir, Path publisher
           : new Path(publisherOutputDir, status.getPath().getName());
 
       LOG.info(String.format("Moving %s to %s", status.getPath(), finalOutputPath));
-      parallelRunner.renamePath(status.getPath(), finalOutputPath, Optional.<String>absent());
+      parallelRunner.renamePath(status.getPath(), finalOutputPath, Optional.<String> absent());
```
liyinan926: Additional space.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -197,6 +224,29 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
+        continue;
+      }
+      String pathSuffix = taskOutputFile
+          .substring(taskOutputFile.indexOf(writerOutputDir.toString()) + writerOutputDir.toString().length() + 1);
+      Path publisherOutputPath = new Path(publisherOutputDir, pathSuffix);
+      WriterUtils.mkdirsWithRecursivePermission(this.fileSystemByBranches.get(branchId),
+          publisherOutputPath.getParent(), this.permissions.get(branchId));
+
+      LOG.info(String.format("Moving %s to %s", taskOutputFile, publisherOutputPath));
+      parallelRunner.renamePath(taskOutputPath, publisherOutputPath, Optional.<String> absent());
```
liyinan926: Additional space after `<String>`.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
```diff
@@ -48,21 +53,21 @@ public AvroHdfsTimePartitionedWithRecordCountsWriter(Destination destination, St
...
-    Path writerOutputDir = new Path(this.properties.getProp(ForkOperatorUtils
-        .getPropertyNameForBranch(ConfigurationKeys.WRITER_OUTPUT_DIR, this.numBranches, this.branch)));
+    for (Entry<Path, FsDataWriter<GenericRecord>> entry : this.pathToWriterMap.entrySet()) {
 
-    for (Entry<Path, DataWriter<GenericRecord>> entry : this.pathToWriterMap.entrySet()) {
-
-      String filePathOld = new Path(entry.getKey(), fileName).toString();
+      String filePathOld = entry.getValue().getOutputFilePath();
 
       String filePathNew = filePathOld.substring(0, filePathOld.lastIndexOf(".")) + "."
```
liyinan926: You can use `Files.getNameWithoutExtension(filePathOld)`.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -456,7 +456,6 @@ private boolean checkDataQuality(Optional<Object> schema)
    */
   private void commitData() throws IOException {
     if (this.writer.isPresent()) {
-      this.logger.info(String.format("Committing data of fork %d of task %s", this.index, this.taskId));
```
liyinan926: Why this is gone?

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -184,7 +183,7 @@ boolean isJobLockEnabled() {
    * @return a {@link Map} from dataset URNs to {@link JobState.DatasetState}s representing the dataset states
    */
   Map<String, JobState.DatasetState> getDatasetStatesByUrns() {
-    return ImmutableMap.copyOf(this.datasetStatesByUrns.or(Maps.<String, JobState.DatasetState>newHashMap()));
+    return ImmutableMap.copyOf(this.datasetStatesByUrns.or(Maps.<String, JobState.DatasetState> newHashMap()));
```
liyinan926: Ditto.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -220,6 +219,10 @@ void finalizeJobStateBeforeCommit() {
   void commit() throws IOException {
     this.datasetStatesByUrns = Optional.of(this.jobState.createDatasetStatesByUrns());
     boolean allDatasetsCommit = true;
+    boolean shouldCommitDataInJob = shouldCommitDataInJob();
+    if (!shouldCommitDataInJob) {
```
liyinan926: You can move this check to after line 240.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -146,15 +148,13 @@ public void run() {
...
-                  schema instanceof Copyable ? ((Copyable) schema).copy() : schema,
-                  branches, i, forkCountDownLatch));
+          Fork fork = closer.register(new Fork(this.taskContext, this.taskState,
+              schema instanceof Copyable ? ((Copyable) schema).copy() : schema, branches, i, forkCountDownLatch));
           // Schedule the fork to run
           this.taskExecutor.submit(fork);
           this.forks.add(Optional.of(fork));
         } else {
-          this.forks.add(Optional.<Fork>absent());
+          this.forks.add(Optional.<Fork> absent());
```
liyinan926: Ditto.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -233,6 +244,52 @@ public void run() {
...
+   * {@link ConfigurationKeys#PUBLISH_DATA_IN_TASKS} is set to true, and task state is SUCCESSFUL.
+   */
+  private boolean shouldPublishDataInTask() {
+    boolean jobCommitPolicyIsPartial =
+        JobCommitPolicy.getCommitPolicy(this.taskState.getProperties()) == JobCommitPolicy.COMMIT_ON_PARTIAL_SUCCESS;
+    boolean publishDataInTasksAllowed = this.taskState.getPropAsBoolean(ConfigurationKeys.PUBLISH_DATA_IN_TASKS,
+        ConfigurationKeys.DEFAULT_PUBLISH_DATA_IN_TASKS);
+
+    if (jobCommitPolicyIsPartial && !publishDataInTasksAllowed) {
+      LOG.info(String.format("%s is partial, but %s is disabled. Will publish data in driver.",
```
liyinan926: I think `will publish data at the job level` sounds better.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -233,6 +244,52 @@ public void run() {
...
+          ConfigurationKeys.JOB_COMMIT_POLICY_KEY, ConfigurationKeys.PUBLISH_DATA_IN_TASKS));
+    } else if (!jobCommitPolicyIsPartial && publishDataInTasksAllowed) {
+      LOG.info(String.format("%s is enabled, but %s is set to full. Will publish data in driver.",
+          ConfigurationKeys.PUBLISH_DATA_IN_TASKS, ConfigurationKeys.JOB_COMMIT_POLICY_KEY));
+    }
+
+    return jobCommitPolicyIsPartial && publishDataInTasksAllowed
+        && this.taskState.getWorkingState() == WorkUnitState.WorkingState.SUCCESSFUL;
```
liyinan926: Actually currently if partial commit is selected, failed tasks are also allowed to publish data.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -197,6 +224,29 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
     return WriterUtils.getDataPublisherFinalDir(workUnitState, this.numBranches, branchId);
   }
 
+  protected void addSingleTaskWriterOutputToExistingDir(Path writerOutputDir, Path publisherOutputDir,
+      WorkUnitState workUnitState, int branchId, ParallelRunner parallelRunner) throws IOException {
+    Preconditions.checkArgument(workUnitState.contains(ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH), "Missing property "
+        + ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH + ", which is needed for publishing data of a single task");
+
+    Iterable<String> taskOutputFiles = workUnitState.getPropAsList(ConfigurationKeys.WRITER_FINAL_OUTPUT_PATH);
```
liyinan926: Sounds like the config property should be named `WRITER_FINAL_OUTPUT_FILE_PATHS`.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -134,51 +147,65 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+        boolean replaceFinalOutputDir = this.getState().getPropAsBoolean(ForkOperatorUtils
+            .getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISHER_REPLACE_FINAL_DIR, this.numBranches, branchId));
+
+        // If the final output directory is not configured to be replaced, put new data to the existing directory.
+        if (!replaceFinalOutputDir) {
+          addWriterOutputToExistingDir(writerOutputDir, publisherOutputDir, state, branchId, parallelRunner);
+          writerOutputPathsMoved.add(writerOutputDir);
+          return;
+        }
+        // Delete the final output directory if it is configured to be replaced
```
liyinan926: Good to add a log message here printing the path to be deleted.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
zliu41: DataPublisher is an abstract class not interface.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
zliu41: This `publishData` method (with parameter `writerOutputPathsMoved`) is for publishing multiple-task data. Publishing single task data doesn't need `writerOutputPathsMoved`. The other one is for publishing single task data. I'll add some javadoc.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
zliu41: There's an identical log message in `Fork.commitData()`

liyinan926:
It's all good to update the Javadoc for `DataPublisher` to make it clear it may publish data at both the job and task levels.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
zliu41: This only return file names, but I also need the path (e.g., given `/a/b/c.txt` it returns `c` but I need `a/b/c`

zliu41:
Javadoc of `DataPublisher` already says it can publish at task level.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
liyinan926: It actually returns all but the extension of the file name.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
zliu41: Then the log message in this `if` block will be printed multiple times which is unnecessary

liyinan926: I mean moving this log to the `else` part of the `if ` block on line 240.

zliu41: That `if` block is inside the `for` loop, so this log message will be printed multiple times if there are multiple datasets.

chavdar:
LGTM

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWithRecordCountsWriter.java
zliu41: No it only returns the file name. But I'm now using Path.getParent() to get the parent folder, plus `Files.getNameWithoutExtension` to make it look cleaner.

zliu41:
Updated.

PR #336, opened 2015-09-22
Filter Hadoop counter groups in MRTaskStateTracker

lbendig:
```MRTaskStateTracker``` exposes metrics as Hadoop counters if **mr.report.metrics.as.counters** is set. However, all the counters are associated to the JOB group (pull #169) which results in an emtpy taskCounterGroup in ```MRJobLauncher#countersToMetrics```. As a result, the TASK level counters will be shown in the JOB Counter group on the JobHistory UI.

The patch tries to differentiate between the MetricGroup type  by applying a filter on the counters returned by ```JobMetrics.get(null, task.getJobId()).getMetricContext().getCounters()```
in ```MRTaskStateTracker```

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
@@ -132,13 +132,13 @@ public class MRTaskStateTracker extends AbstractTaskStateTracker {
     JOB() {
       @Override
       public String getGroupName() {
-        return MetricGroup.JOB.name();
+        return MetricGroup.JOB.toString();
       }
     },
     TASK() {
       @Override
       public String getGroupName() {
-        return MetricGroup.TASK.name();
+        return MetricGroup.TASK.toString();
       }
     };
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
@@ -23,6 +23,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.Counter;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricFilter;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.metrics.GobblinMetrics;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java [hunk 1]
```diff
@@ -70,10 +72,7 @@ public void onTaskCompletion(Task task) {
 
         if (workUnit.getPropAsBoolean(ConfigurationKeys.MR_REPORT_METRICS_AS_COUNTERS_KEY,
             ConfigurationKeys.DEFAULT_MR_REPORT_METRICS_AS_COUNTERS)) {
-          Map<String, Counter> counters = JobMetrics.get(null, task.getJobId()).getMetricContext().getCounters();
-          for (Map.Entry<String, Counter> entry : counters.entrySet()) {
-            this.context.getCounter(MetricGroup.JOB.name(), entry.getKey()).setValue(entry.getValue().getCount());
-          }
+          updateCounters(task);
```
liyinan926: `this.task`.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
```diff
@@ -117,4 +113,40 @@ protected void updateTaskMetrics() {
       this.context.progress();
     }
   }
+  
+  private void updateCounters(Task task) {
+    updateCounters(task, MetricGroupFilter.JOB);
```
liyinan926: Ditto.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
```diff
@@ -117,4 +113,40 @@ protected void updateTaskMetrics() {
...
+        this.context.getCounter(filter.getGroupName(), entry.getKey()).setValue(entry.getValue().getCount());
+      }
+    }
+  }
+  
+  private enum MetricGroupFilter implements MetricFilter {
+    JOB() {
+      @Override
+      public String getGroupName() {
+        return MetricGroup.JOB.name();
```
liyinan926: `toString` should be used in preference to `name` according to http://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#name().

liyinan926:

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
lbendig: There's no reference to this. inside onTaskCompletion()

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
lbendig: updateCounters() is outside MRTaskMetricsUpdater, there no reference to this.task.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
liyinan926: Ah, my bad. Didn't realize this.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRTaskStateTracker.java
lbendig: Fixed. I just followed the code style according to this.context.getCounter(MetricGroup.JOB.name().

liyinan926:
LGTM. Thanks for fixing this!

lbendig:
Thank you Yinan for the review!

PR #338, opened 2015-09-22
Fix default timestamp insertion to JobHistoryStore

lbendig:
When a job execution info is inserted into the jobhistory store, first a default timestamp is inserted into the end_time column of the tables. As the job finishes this timestamp is updated to the actual one.
However, the default timestamp is defined as **1970-01-01 00:00:00** which results in the following exception:
```
Failed to write job execution information to the job history store: java.io.IOException: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Incorrect datetime value: '1970-01-01 00:00:00' for column 'end_time' at row 1
java.io.IOException: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Incorrect datetime value: '1970-01-01 00:00:00' for column 'end_time' at row 1
	at gobblin.metastore.DatabaseJobHistoryStore.put(DatabaseJobHistoryStore.java:245)
```
According to http://dev.mysql.com/doc/refman/5.1/en/datetime.html the min value of the timestamp should be **1970-01-01 00:00:01**.
In MySQL 5 and above, TIMESTAMP values are converted from the current timezone to UTC for storage, and converted back from UTC to the current time zone for retrieval.
We need to ensure to insert '1970-01-01 00:00:01' with UTC timzone, othewise the minimum value might violate the MySQL timstamp constraint if we are using a different timezone.
E.g: 1970-01-01 00:00:01 inserted with UTC+4:30 hours would result in 1969-12-31 19:30:01 in UTC.

### DIFF
--- a/gobblin-metastore/src/main/java/gobblin/metastore/DatabaseJobHistoryStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/DatabaseJobHistoryStore.java
@@ -22,8 +22,11 @@ import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.AbstractMap;
+import java.util.Calendar;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
+
 import javax.sql.DataSource;
 
 import org.slf4j.Logger;
@@ -151,7 +154,7 @@ public class DatabaseJobHistoryStore implements JobHistoryStore {
   private static final String TASK_PROPERTY_QUERY_STATEMENT_TEMPLATE =
       "SELECT property_key, property_value FROM gobblin_task_properties WHERE task_id=?";
 
-  private static final String DEFAULT_TIMESTAMP = "1970-01-01 00:00:00";
+  private static final Timestamp DEFAULT_TIMESTAMP = new Timestamp(1000L);
 
   private final DataSource dataSource;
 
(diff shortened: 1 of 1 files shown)


gobblin-metastore/src/main/java/gobblin/metastore/DatabaseJobHistoryStore.java
```diff
@@ -781,4 +784,8 @@ private String constructTableFilter(Table table) {
 
     return sb.toString();
   }
+  
+  private static Calendar getCalendarUTCInstance() {
+    return Calendar.getInstance(TimeZone.getTimeZone("UTC"));
```
liyinan926: I think this can be defined as a constant.

lbendig: I think the UTC Timezone object can be put into a constant but I'm not sure about the Calendar object; I've checked some of the internals of ```PreparedStatement#setTimestamp``` and it seems to me that the Calendar object may be mutated which is up to the given driver implementation. E.g:
```com.mysql.jdbc.TimeUtil#jdbcCompliantZoneShift``` :
> // JDBC spec is not clear whether or not this calendar should be immutable, so let's treat it like it is, for safety

What do you think?

liyinan926: Got it.

liyinan926:
LGTM. Thanks!

lbendig:
Thank you Yinan, for the review!

PR #339, opened 2015-09-22
Gobblin on Yarn POC preview

liyinan926:
This is a preview PR for the Gobblin on Yarn POC. I am still testing this and particularly the part that handles security/authentication so this probably will be updated as I find and fix issues. I will schedule a meeting to go over this PR later this week.


liyinan926:
@chavdar @zliu41 can you do an initial review? I will schedule a meeting later this week to go over this PR.

zliu41:
Container management doesn't seem to be included, e.g., pulling workunits from priority queue in ZK, which container runs which workunits, general container pool vs specialized pool, etc. Are these going to be in a different PR?

zliu41:
Seems most of the classes in this PR are for batch jobs. It's better to rename them to be more clear, e.g., "GobblinHelixJobLauncher" -> "GobblinHelixBatchJobLauncher"

liyinan926:
@zliu41 there's been a huge change in the design due to the use of Helix's task execution framework. I will explain it during the meeting.

liyinan926:
Fixed build and updated with upstream/master. @chavdar @zliu41 @sahilTakiar can you take a look?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+      throw Throwables.propagate(e);
+    }
+
+    // Start all the services running in the ApplicationMaster
+    this.serviceManager.startAsync();
+
+    this.yarnClient.start();
+    this.applicationId = Optional.of(setupAndSubmitApplication());
+
+    ListenableFuture<ApplicationReport> appReportFuture =
```
sahilTakiar: Would it be simpler to just have a regular `Executors.newSingleThreadScheduledExecutor` that is scheduled the below `Callable` to run every 60 seconds. The `call()` method can check if the `appState` is `FINISHED`, `FAILED`, or `KILLED`; if it is then it executes `GobblinYarnAppLauncher.this.stop()` and sets `applicationCompleted = true`. Otherwise it does nothing.

If an exception is ever thrown, it should be accessible via a regular `Future` from the thrown `ExecutionException`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+    ApplicationSubmissionContext appSubmissionContext = gobblinYarnApp.getApplicationSubmissionContext();
+    ApplicationId applicationId = appSubmissionContext.getApplicationId();
+
+    GetNewApplicationResponse newApplicationResponse =gobblinYarnApp.getNewApplicationResponse();
+    // Set up resource type requirements for ApplicationMaster
+    Resource resource = prepareContainerResource(newApplicationResponse);
+
+    Map<String, LocalResource> appMasterLocalResources = Maps.newHashMap();
+    // Add lib jars, and jars and files that the ApplicationMaster need as LocalResources
+    addAppMasterLocalResources(applicationId, appMasterLocalResources);
```
sahilTakiar: Why not make this method return a `Map` instead of take in a `Map.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+    addContainerLocalResources(applicationId);
+
+    // Submit the application
+    LOGGER.info("Submitting application " + applicationId);
+    this.yarnClient.submitApplication(appSubmissionContext);
+
+    return applicationId;
+  }
+
+  private Resource prepareContainerResource(GetNewApplicationResponse newApplicationResponse) {
```
sahilTakiar: Suggest renaming this to something more specific to the AM, such as getApplicationMasterResource

sahilTakiar: Actually, I would think this method can be moved to a util class, since we would need similar code when specifying the resources for any new container.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+            environmentVariableMap, ApplicationConstants.Environment.CLASSPATH.key(), classpath.trim());
+      }
+    }
+
+    return environmentVariableMap;
+  }
+
+  private String buildApplicationMasterCommand(int memoryMbs) {
+    String appMasterClassName = GobblinApplicationMaster.class.getSimpleName();
+    return String.format(
```
sahilTakiar: I think using `StringBuilder` rather than `String.format` will make the code more readable.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+      addAppRemoteFiles(this.config.getString(ConfigurationConstants.APP_MASTER_FILES_REMOTE_KEY), resources);
+    }
+    if (this.config.hasPath(ConfigurationConstants.JOB_CONF_PACKAGE_PATH_KEY)) {
+      Path appFilesDestDir = new Path(appMasterWorkDir, ConfigurationConstants.APP_FILES_DIR_NAME);
+      addJobConfPackage(this.config.getString(ConfigurationConstants.JOB_CONF_PACKAGE_PATH_KEY),
+          appFilesDestDir, resources);
+    }
+  }
+
+  private void addContainerLocalResources(ApplicationId applicationId) throws IOException {
```
sahilTakiar: Whats the difference this method and `YarnClient. addContainerLocalResources`. Can the methods be combined. I would guess most of the utility methods for adding local resources for the AM is very similar to the methods that add resources for containers, right?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+    Map<String, String> environmentVariableMap = Maps.newHashMap();
+
+    Apps.addToEnvironment(environmentVariableMap, ApplicationConstants.Environment.JAVA_HOME.key(),
+        System.getenv(ApplicationConstants.Environment.JAVA_HOME.key()));
+
+    // Add jars/files in the working directory of the ApplicationMaster to the CLASSPATH
+    Apps.addToEnvironment(environmentVariableMap, ApplicationConstants.Environment.CLASSPATH.key(),
+        ApplicationConstants.Environment.PWD.$());
+    Apps.addToEnvironment(environmentVariableMap, ApplicationConstants.Environment.CLASSPATH.key(),
+        ApplicationConstants.Environment.PWD.$() + File.separator + "*");
```
sahilTakiar: Is it necessary to add both `PWD` and `PWD/*` to the classpath? Also, what is even present under the `PWD` dir that needs to be in the classpath?

Are local resources automatically added to the classpath?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+    fileResource.setSize(fileStatus.getLen());
+    fileResource.setTimestamp(fileStatus.getModificationTime());
+    fileResource.setType(resourceType);
+    fileResource.setVisibility(LocalResourceVisibility.APPLICATION);
+    LOGGER.debug(String.format("Created a LocalResource for file %s of type %s",
+        fileResource.getResource(), resourceType));
+    resourceMap.put(destFilePath.getName(), fileResource);
+  }
+
+  private Map<String, String> getEnvironmentVariables() {
```
sahilTakiar: Again, this seems similar to `YarnClient. getEnvironmentVariables`. I would suggest moving all these methods out into a separate class.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
```diff
@@ -0,0 +1,413 @@
...
+
+  private final HelixManager helixManager;
+
+  private final ServiceManager serviceManager;
+
+  private final EventBus eventBus;
+
+  private final TaskStateModelFactory taskStateModelFactory;
+
+  private final MetricRegistry metricRegistry = new MetricRegistry();
```
sahilTakiar: Any reason why we are directly using the codahale metrics APIs instead of the Gobblin metrics APIs

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+      ByteBuffer fsTokens = ByteBuffer.wrap(dataOutputBuffer.getData(), 0, dataOutputBuffer.getLength());
+      containerLaunchContext.setTokens(fsTokens);
+    } catch (Throwable t) {
+      throw closer.rethrow(t);
+    } finally {
+      closer.close();
+    }
+  }
+
+  private void sendShutdownRequest() {
```
sahilTakiar: It looks like this method sends a message to the AM via Helix. An the HelixManager in the AM receives that message, and sends a message on the EventBus. On the AM receives the message on the EventBus and then shutdowns. It sounds like a lot of hops just to execute a shutdown method.

Could this method simply perform `this.yarnClient.killApplication()`. Any cleanup that needs to be done the AM can be added to a shutdown hook. This avoids having to go through Helix, which I think decreases complexity.

liyinan926: I noticed that sometime `killApplication` does not properly kill the containers, and that's why I resorted to this approach for more graceful shutdown.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,453 @@
...
+     */
+    private class ControllerShutdownMessageHandler extends MessageHandler {
+
+      public ControllerShutdownMessageHandler(Message message, NotificationContext context) {
+        super(message, context);
+      }
+
+      @Override
+      public HelixTaskResult handleMessage() throws InterruptedException {
+        String messageSubType = this._message.getMsgSubType();
```
zliu41: References to member variables are inconsistent: some with underscore, some with `this`, some with both and some with neither.

gobblin-yarn/src/main/java/gobblin/yarn/ConfigurationConstants.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+package gobblin.yarn;
+
+/**
+ * A central place for configuration related constants of Gobblin on Yarn.
+ *
+ * @author ynli
+ */
+public class ConfigurationConstants {
```
zliu41: Can you rename to `YarnConfigurationKeys` so as to not be confused with `gobblin.configuration.ConfigurationKeys`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,453 @@
...
+  private volatile boolean isStopped = false;
+
+  public GobblinApplicationMaster(String applicationName, Config config) throws Exception {
+    // An EventBus used for communications between services running in the ApplicationMaster
+    this.eventBus = new EventBus(GobblinApplicationMaster.class.getSimpleName());
+    this.eventBus.register(this);
+
+    ContainerId containerId =
+        ConverterUtils.toContainerId(System.getenv().get(ApplicationConstants.Environment.CONTAINER_ID.key()));
+    ApplicationAttemptId applicationAttemptIdId = containerId.getApplicationAttemptId();
```
zliu41: 'Id' appears twice.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,453 @@
...
+    } finally {
+      // Stop metric reporting
+      this.jmxReporter.stop();
+
+      if (this.helixManager.isConnected()) {
+        this.helixManager.disconnect();
+      }
+    }
+
+    this.isStopped = true;
```
zliu41: What's the use of `isStopped`? If it is useful, should this be in a finally block?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,453 @@
...
+  }
+
+  @SuppressWarnings("unused")
```
zliu41: Not sure if it shows warnings in your IDE but I don't think you need to do this for public methods. Ditto for a few others.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: `_message` is defined in `MessageHandler` as part of Helix so I'm just using it. it's impossible to use `this.` in `run` below.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Actually with `killApplication`, the AM will not have a chance to unregister itself properly and there'll be exceptions thrown for not be able to do hearbeats between AM and the RM.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
liyinan926: Yes, we can use the Gobblin metrics APIs. But for this simple task of reporting JVM metrics to JMX, `MetricRegistry` is sufficient and simpler. This is also used in `SchedulerDeamon`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: I'm not 100% sure on this so I added both. I think for jars, `PWD/*` is necessary and for `.conf` files, `PWD` is necessary.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -0,0 +1,273 @@
...
+    Map<String, TaskConfig> taskConfigMap = Maps.newHashMap();
+
+    Closer closer = Closer.create();
+    try {
+      ParallelRunner stateSerDeRunner = closer.register(new ParallelRunner(this.stateSerDeRunnerThreads, this.fs));
+
+      for (WorkUnit workUnit : workUnits) {
+        if (workUnit instanceof MultiWorkUnit) {
+          // Flatten the MultiWorkUnit and add each individual WorkUnit
+          for (WorkUnit innerWorkUnit : JobLauncherUtils.flattenWorkUnits(((MultiWorkUnit) workUnit).getWorkUnits())) {
```
zliu41: Why do MultiWorkUnits need to be flattened to WorkUnits to be persisted?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -0,0 +1,273 @@
...
+  /**
+   * Submit a job to run.
+   */
+  private void submitJobToHelix(String jobName, String jobId, JobConfig.Builder jobConfigBuilder) throws Exception {
+    // Create one queue for each job with the job name being the queue name
+    JobQueue jobQueue = new JobQueue.Builder(jobName).build();
+    try {
+      this.helixTaskDriver.createQueue(jobQueue);
+    } catch (IllegalArgumentException iae) {
+      LOGGER.warn(String.format("Job queue %s already exists", jobQueue.getName()));
```
zliu41: This could be an `info` rather than `warn` since if a job runs multiple times then it's normal that from the 2nd time the jobQueue will already exist.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Yes, this is useful to prevent `stop` from being called more than once, which may happen if the AM receives a shutdown message and for some reason the shutdown hook is called due to some reason like running out of virtual memory limit.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Yes, it shows warning in IDEA.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -0,0 +1,273 @@
...
+    stateSerDeRunner.serializeToFile(workUnit, workUnitFile);
+    return workUnitFile.toString();
+  }
+
+  private void waitForJobCompletion() throws InterruptedException {
+    while (true) {
+      WorkflowContext workflowContext = TaskUtil.getWorkflowContext(this.helixManager, this.helixQueueName);
+      if (workflowContext != null) {
+        org.apache.helix.task.TaskState jobState = workflowContext.getJobState(this.jobResourceName);
+        if (jobState != null && jobState == org.apache.helix.task.TaskState.COMPLETED) {
```
zliu41: `jobState != null` is not needed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -0,0 +1,273 @@
...
+  private void addWorkUnit(WorkUnit workUnit, ParallelRunner stateSerDeRunner,
+      Map<String, TaskConfig> taskConfigMap, String jobName, String jobId) throws IOException {
+    String workUnitFilePath = persistWorkUnit(new Path(this.inputWorkUnitDir, jobId), workUnit, stateSerDeRunner);
+
+    Map<String, String> rawConfigMap = Maps.newHashMap();
+    rawConfigMap.put(ConfigurationConstants.WORK_UNIT_FILE_PATH, workUnitFilePath);
+    rawConfigMap.put(ConfigurationKeys.JOB_NAME_KEY, jobName);
+    rawConfigMap.put(ConfigurationKeys.JOB_ID_KEY, jobId);
+    rawConfigMap.put(ConfigurationKeys.TASK_ID_KEY, workUnit.getId());
+    rawConfigMap.put("TASK_SUCCESS_OPTIONAL", "true");
```
zliu41: should this be defined as constant?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -0,0 +1,273 @@
...
+    JobConfig.Builder jobConfigBuilder = new JobConfig.Builder();
+    jobConfigBuilder.addTaskConfigMap(taskConfigMap).setCommand(GobblinWorkUnitRunner.GOBBLIN_TASK_FACTORY_NAME);
+
+    return jobConfigBuilder;
+  }
+
+  /**
+   * Submit a job to run.
+   */
+  private void submitJobToHelix(String jobName, String jobId, JobConfig.Builder jobConfigBuilder) throws Exception {
```
zliu41: You don't need to pass jobName and jobId since they can be obtained from this.jobContext. Ditto for a few other methods.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobScheduler.java
```diff
@@ -0,0 +1,122 @@
...
+  @Subscribe
+  public void handleNewJobConfigArrival(NewJobConfigArrivalEvent newJobArrival) {
+    LOGGER.info("Received new job configuration of job " + newJobArrival.getJobName());
+    try {
+      Properties jobConfig = new Properties();
+      jobConfig.putAll(this.properties);
+      jobConfig.putAll(newJobArrival.getJobConfig());
+      if (jobConfig.containsKey(ConfigurationKeys.JOB_SCHEDULE_KEY)) {
+        LOGGER.info("Scheduling new job " + newJobArrival.getJobName());
+        scheduleJob(newJobArrival.getJobConfig(), null);
```
zliu41: should it check `this.scheduleJobs` to see if the job is already scheduled?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
```diff
@@ -0,0 +1,413 @@
...
+        new GobblinHelixTaskFactory(taskExecutor, taskStateTracker, fs, appWorkDir));
+    this.taskStateModelFactory = new TaskStateModelFactory(this.helixManager, taskFactoryMap);
+    this.helixManager.getStateMachineEngine().registerStateModelFactory(StateModelDefId.from("Task"),
+        this.taskStateModelFactory);
+  }
+
+  /**
+   * Start this {@link GobblinWorkUnitRunner} instance.
+   */
+  public void start() {
```
zliu41: This looks fairly similar to `GobblinApplicationMaster.start()`. Can this be refactored?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -0,0 +1,598 @@
...
+          GobblinYarnAppLauncher.this.stop();
+        } catch (IOException ioe) {
+          LOGGER.error("Failed to close the " + GobblinYarnAppLauncher.class.getSimpleName(), ioe);
+        }
+      }
+    });
+  }
+
+  public synchronized void stop() throws IOException {
+    if (this.stopped) {
```
zliu41: Why does this class not have `this.stopInProgress`?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: The naming is confusing, but this is the Helix's `JobState`, which may be `null` if `getJobState` returns `null`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: Updated.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobScheduler.java
liyinan926: Added a check in `JobScheduler.scheduleJob`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: `stopInProgress` is used in `GobblinApplicationMaster` to prevent the shutdown message handler from scheduling the `shutdownMessageHandlingCompletionWatcher` if shutdown is already in progress, for example, because the shutdown hook is triggered and in the meantime a shutdown message is received.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/ConfigurationConstants.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: I couldn't remember exactly why I did this initially but it had something to do with allowing Helix to have full control on the assignment/grouping of `WorkUnit`s. I think we can stick to this and see how Helix does in terms of load balancing.

liyinan926:
@zliu41 regarding your suggestion on renaming "GobblinHelixJobLauncher" -> "GobblinHelixBatchJobLauncher", I think we can keep the current name and use a new name like `GobblinHelixStreamingJobLauncher` for streaming jobs.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Yes, we can certainly do that instead.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: This method needs to add all the app-level dependencies added by the AM, which is already on HDFS instead of local disk.

liyinan926: `addFileAsLocalResource` is a good candidate to put into a utility class.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: Changed it to `helixJobState`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: A `ScheduledExecutorService` does not allow scheduling `Callable` to run at fixed rate. Only `Runnable` can be used which cannot return the `appReport`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
zliu41: If `jobState` is null then `jobState == org.apache.helix.task.TaskState.COMPLETED` should return false.

liyinan926: Yes, you are right. Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
zliu41: It seems `stopInProgress` alone is sufficient for this purpose.

liyinan926: Yes, fixed.

liyinan926:
@zliu41 @sahilTakiar updated 408a22f.

zliu41:
LGTM

liyinan926:
@sahilTakiar do you have more comments? @chavdar now it's in your plate.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
sahilTakiar: I see. In that case this approach is probably fine.

Although I would consider either (1) using an `EventBus` so that instead you can use `Runnable` instead of `Callable`, or (2) using something like Akka's Actor model, or (3) making `appReport` an instance variable, and have a single thread pool with two scheduled threads: the first to check the `YarnApplicationState` and set `appReport` accordingly, and the second to see if `appReport` becomes non-null and act accordingly.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
sahilTakiar: Ok, sounds reasonable to not use `killApplication` in this instance.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,452 @@
...
+        if (messageSubType.equalsIgnoreCase(HelixMessageSubTypes.TOKEN_FILE_UPDATED.toString())) {
+          LOGGER.info("Handling message " + HelixMessageSubTypes.TOKEN_FILE_UPDATED.toString());
+
+          eventBus.post(new DelegationTokenUpdatedEvent());
+          HelixTaskResult helixTaskResult = new HelixTaskResult();
+          helixTaskResult.setSuccess(true);
+          return helixTaskResult;
+        }
+
+        throw new RuntimeException(String.format("Unknown %s message subtype: %s",
```
sahilTakiar: Seems more like an `IllegalArgumentException` or `IllegalStateException`. Since we provide the mapping between message type and message handler, Helix should also perform the mapping properly.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,452 @@
...
+          }
+
+          LOGGER.info("Handling message " + HelixMessageSubTypes.APPLICATION_MASTER_SHUTDOWN.toString());
+
+          ScheduledExecutorService shutdownMessageHandlingCompletionWatcher =
+              MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(1));
+
+          // Schedule the task for watching on the removal of the shutdown message, which indicates that
+          // the message has been successfully processed and it's safe to disconnect the HelixManager.
+          shutdownMessageHandlingCompletionWatcher.scheduleAtFixedRate(new Runnable() {
```
sahilTakiar: Why do we need a scheduled thread to invoke the shutdown of the AM? It's not fully clear why `helixProperty == null` means that we should post a `ApplicationMasterShutdownRequest`.

liyinan926: It's a hacky way of making sure the message has been processed before shutting down and disconnecting the `helixManager`. Helix internally schedules a task to handle a message and it needs the ZooKeeper connection from the `HelixManager` to be active during this process. Unless Helix allows users to use a callback upon completion of the handling of a message, there's no way to know if a message has been processed or not. `helixProperty == null` means the message has been fully handled and deleted from ZooKeeper, at which time it's safe to disconnect.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Sure we can use `IllegalStateException`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: `EventBus` sounds a good solution.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Actually for `SHUTDOWN` messages, I will just use a `Preconditions.checkArgument` at the beginning as there shouldn't be a third message subtype other than the current two. But for `USER_DEFINED_MSG` I will keep the `if` block and throw an `IllegalArgumentException` if an unexpected message subtype is seen.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -0,0 +1,457 @@
...
+    // Register itself with the EventBus for container-related requests
+    this.eventBus.register(this);
+  }
+
+  @SuppressWarnings("unused")
+  @Subscribe
+  public void handleNewContainerRequest(NewContainerRequest newContainerRequest) {
+    int newContainersRequested = newContainerRequest.getNewContainersRequested();
+    if (newContainersRequested <= 0) {
+      LOGGER.error("Invalid number of new containers requested: " + newContainersRequested);
```
sahilTakiar: Change error message to "Number of containers requested me be positive and non-zero..."

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -0,0 +1,457 @@
...
+      return;
+    }
+
+    if (!this.maxResourceCapacity.isPresent()) {
+      LOGGER.error("Unable to handle new container request as maximum resource capacity is not available");
```
sahilTakiar: Probably useful to log error the the value of the max capacity, the current number of already requested containers, and the number of containers being requested.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -0,0 +1,457 @@
...
+
+    private volatile boolean done = false;
+
+    @Override
+    public void onContainersCompleted(List<ContainerStatus> statuses) {
+      for (ContainerStatus containerStatus : statuses) {
+        LOGGER.info(String.format("Container %s has completed with exit status %d",
+            containerStatus.getContainerId(), containerStatus.getExitStatus()));
+        containerMap.remove(containerStatus.getContainerId());
+        containerToParticipantMap.remove(containerStatus.getContainerId());
```
sahilTakiar: Are all the callbacks executed by a single thread? If not we may want to add a `synchronized` block around the two previous lines, so that `containerMap` and `containerToParticipantMap` are always consistent (e.g. they always contain the same `ContainerId`s.

Same goes for all other mutations of `containerMap` and `containerToParticipantMap`.

liyinan926: Fixed this by using a single map `containerMap` which will now stores `<String, Map.Entry<Container, ParticipantId>>`.

liyinan926:
@sahilTakiar Updated 36af44b.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -0,0 +1,457 @@
...
+  private final int requestedContainerCores;
+
+  private final String containerJvmArgs;
+
+  // Security tokens for accessing HDFS
+  private final ByteBuffer tokens;
+
+  private final Closer closer = Closer.create();
+
+  private final Object allContainersStopped = new Object();
```
sahilTakiar: Why not replace this with `Optional<CountDownLatch> containersToStop` - it can initially be set to `Optional.absent()` but then set to `Optional.of(new CountDownLatch(containerMap.size())` in the `shutdown()` method.

liyinan926: We probably don't want to do that because in the future the number of contains will change at runtime and `containerMap.size()` will no longer be a good indication.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
sahilTakiar: If possible, I think it may be better to just use the Gobblin Metrics API:

* I think it makes the code more consistent if we stick to just using Gobblin Metrics - only the gobblin-metrics submodule should be aware that we are using the codahale library - which also makes migration to a different metrics backend easier.
* Gobblin Metrics lets us take advantage of the child-parent aggregation
* Gobblin Metrics allows us to report metrics to Kafka also
* Gobblin Metrics has EventReporters that may be useful throughout the code

liyinan926:
@sahilTakiar updated e617b25.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
sahilTakiar: I see, interesting. Sounds like you opened a ticket for this too (HELIX-611) can you add it to the java docs?

liyinan926: Sure. But I still feel we should switch to Akka for messaging sometime in the future as Helix's way of handling messaging and the use of ZooKeeper for message queues are not so sounding.

liyinan926: Added to the Javadoc.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -0,0 +1,468 @@
...
+            String.format("Failed to handle message with exception %s, error code %s, error type %s", e, code, type));
+      }
+    }
+  }
+
+  /**
+   * A custom {@link MessageHandlerFactory} for {@link ControllerUserDefinedMessageHandler}s that
+   * handle messages of type {@link org.apache.helix.model.Message.MessageType#USER_DEFINE_MSG}.
+   */
+  private class ControllerUserDefinedMessageHandlerFactory implements MessageHandlerFactory {
```
sahilTakiar: This seems to be identical to the equivalent class in `GobblinWorkUnitRunner.java`. But I'm guessing we didn't merge the two in case other user defined messages need to be added. Would it be easier to use something like protobuf for this kind of communication. Helix doesn't seem to provide a generic and easy way to do this. Not a pressing issue, but may be useful down the line, especially when the containers + app master need to communicate more.

liyinan926: Yeah, we cannot merge them for exactly that reason. I am thinking Akka would be a great candidate for async messaging.

liyinan926:
@sahilTakiar Addressed your comments. Any more comments?

sahilTakiar:
LGTM

PR #341, opened 2015-09-24
Defining Sla Event metadata

pcadabam:
@ibuenros @chavdar  can you take a look at this?

Overview
- The change defines metadata required in an SLA event. An SLA event must have datasetUrn, partition, originTs and upstreamTs
- An SLA event can also have additional optional metadata like completenessPercentage, recordCount etc. See SlaEventKeys.java. Optional metadata is not applicable to all ETL flows.
- Any ETL flow can extend the abstract class SlaEvent and implement the getters for datasetUrn, partition, originTs, upstreamTs. The SlaEvent class does not expose the key names for any metadata. This promises a fixed schema for an SlaEvent.
- Next, we want to define and Arvo schema for the sla event. This change will still emit a GobblinTrackingEvent.

Compaction changes
- Enabled metrics/events for MRCompactor job
- emit an SLA event when daily compaction is completed. See CompactionSlaEvent.java

### DIFF
--- a/gobblin-compaction/build.gradle
+++ b/gobblin-compaction/build.gradle
@@ -14,6 +14,8 @@ apply plugin: 'eclipse'
 dependencies {
   compile project(":gobblin-api")
   compile project(":gobblin-utility")
+  compile project(":gobblin-metrics")
+  compile project(":gobblin-core")
   compile externalDependency.avro
   compile externalDependency.commonsLang
   compile externalDependency.hiveExec
--- a/gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
@@ -1,0 +1,97 @@
+package gobblin.compaction.event;
+
+import java.io.IOException;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.Counter;
+import org.apache.hadoop.mapreduce.Counters;
+import org.apache.hadoop.mapreduce.Job;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import gobblin.compaction.mapreduce.avro.AvroKeyDedupReducer;
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+import gobblin.metrics.event.sla.SlaEventKeys;
+
+/**
+ * Helper class to populate sla event metadata in state.
+ */
+public class CompactionSlaEventHelper {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CompactionSlaEventHelper.class);
+
+  public static void populateState(State state , Job job, FileSystem fs) {
+    setDatasetUrn(state);
+    setPartition(state);
+    setDedupeStatus(state);
+    setPreviousPublishTime(state, fs);
+    setRecordCount(state, job);
+    setUpstreamTimeStamp(state, fs);
+  }
+
+  private static void setDatasetUrn(State state) {
+    state.setProp(SlaEventKeys.DATASET_URN_KEY, new Path(state.getProp(ConfigurationKeys.COMPACTION_DEST_DIR),
+        state.getProp(ConfigurationKeys.COMPACTION_TOPIC)).toString());
+  }
+
+  private static void setPartition(State state) {
+    state.setProp(SlaEventKeys.PARTITION_KEY, state.getProp(ConfigurationKeys.COMPACTION_JOB_DEST_PARTITION));
+  }
+
+  private static void setUpstreamTimeStamp(State state, FileSystem fs) {
+
+    String inputDirectory = state.getProp(ConfigurationKeys.COMPACTION_JOB_INPUT_DIR);
+    try {
+      FileStatus fileStatus = fs.getFileStatus(new Path(inputDirectory));
+      state.setProp(SlaEventKeys.UPSTREAM_TS_IN_MILLI_SECS_KEY, Long.toString(fileStatus.getModificationTime()));
+    } catch (IOException e) {
+      LOG.debug("Failed to get upstream time.", e);
+    }
+  }
+
+  private static void setPreviousPublishTime(State state, FileSystem fs) {
+
+    Path compactionCompletePath =
+        new Path(state.getProp(ConfigurationKeys.COMPACTION_JOB_DEST_DIR),
+            ConfigurationKeys.COMPACTION_COMPLETE_FILE_NAME);
+
+    try {
+      FileStatus fileStatus = fs.getFileStatus(compactionCompletePath);
+      state.setProp(SlaEventKeys.PREVIOUS_PUBLISH_TS_IN_MILLI_SECS_KEY, Long.toString(fileStatus.getModificationTime()));
+    } catch (IOException e) {
+      LOG.debug("Failed to get previous publish time.", e);
+    }
+  }
+
+  private static void setDedupeStatus(State state) {
+    if (state.getPropAsBoolean(ConfigurationKeys.COMPACTION_DEDUPLICATE,
+        ConfigurationKeys.DEFAULT_COMPACTION_DEDUPLICATE)) {
+      state.setProp(SlaEventKeys.DEDUPE_STATUS_KEY, DedupeStatus.DEDUPED);
+
+    } else {
+      state.setProp(SlaEventKeys.DEDUPE_STATUS_KEY, DedupeStatus.NOT_DEDUPED);
+    }
+  }
+
+  private static void setRecordCount(State state, Job job) {
+
+    Counters counters = null;
+    try {
+      counters = job.getCounters();
+    } catch (IOException e) {
+      LOG.debug("Failed to get job counters. Record count will not be set. ", e);
+      return;
+    }
+
+    if (counters != null) {
+      Counter recordCounter = counters.findCounter(AvroKeyDedupReducer.EVENT_COUNTER.RECORD_COUNT);
+      if (recordCounter != null) {
+        state.setProp(SlaEventKeys.RECORD_COUNT_KEY, Long.toString(recordCounter.getValue()));
+      }
+    }
+  }
+
+}
--- a/gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventKeys.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventKeys.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.metrics.event.sla;
+
+public class SlaEventKeys {
+
+  static final String EVENT_GOBBLIN_STATE_PREFIX = "event.sla.";
+  public static final String DATASET_URN_KEY = EVENT_GOBBLIN_STATE_PREFIX + "datasetUrn";
+  public static final String PARTITION_KEY = EVENT_GOBBLIN_STATE_PREFIX + "partition";
+  public static final String ORIGIN_TS_IN_MILLI_SECS_KEY = EVENT_GOBBLIN_STATE_PREFIX + "originTimestamp";
+  public static final String UPSTREAM_TS_IN_MILLI_SECS_KEY = EVENT_GOBBLIN_STATE_PREFIX + "upstreamTimestamp";
+
+  public static final String COMPLETENESS_PERCENTAGE_KEY = EVENT_GOBBLIN_STATE_PREFIX + "completenessPercentage";
+  public static final String DEDUPE_STATUS_KEY = EVENT_GOBBLIN_STATE_PREFIX + "dedupeStatus";
+  public static final String RECORD_COUNT_KEY = EVENT_GOBBLIN_STATE_PREFIX + "recordCount";
+  public static final String PREVIOUS_PUBLISH_TS_IN_MILLI_SECS_KEY = EVENT_GOBBLIN_STATE_PREFIX + "previousPublishTs";
+
+}
--- a/gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
@@ -1,0 +1,129 @@
+package gobblin.metrics.event.sla;
+
+import java.util.Map;
+import java.util.Properties;
+
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Singular;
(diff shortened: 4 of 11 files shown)


gobblin-metrics/src/main/java/gobblin/metrics/event/GenericEvent.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package gobblin.metrics.event;
+
+import java.util.Map;
+
+/**
+ * A generic event with a name and metadata.
+ */
+public abstract class GenericEvent {
```
ibuenros: What is the point of GenericEvent?

pcadabam: Both event name and metadata are owned by the sub classes. If callers use eventSubmitter.submit(eventName, metadata) then the eventName need to be made public. I prefer sla eventName to be private to the SlaEvent class

chavdar:
Some high-level description of what's in the PR will be helpful.

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEvent.java
```diff
@@ -0,0 +1,129 @@
...
+import gobblin.metrics.event.sla.SlaEventKeys;
+
+
+public class CompactionSlaEvent extends SlaEvent {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CompactionSlaEvent.class);
+  private static final String EVENT_NAME = "CompactionCompleted";
+  private static final String DEFAULT = StringUtils.EMPTY;
+
+  private Job job;
```
chavdar: I don't see a reason for the even to keep references to any of these. It should get whatever data it needs in a constructor/factory method.

pcadabam: job - used to get record count
jobProps - used to get COMPACTION_DEST_DIR, COMPACTION_TOPIC, COMPACTION_JOB_DEST_PARTITION, COMPACTION_JOB_INPUT_DIR, COMPACTION_COMPLETE_FILE_NAME, COMPACTION_DEDUPLICATE
fs - used to get the modification time of the compaction file.

I see a few problems with passing all this data.
1. The list of arguments to be passed in the constructor is too long. It often leads to bugs and is unreadable
2. If this class needs more data from jobProps we have to change the method signature of the constructor.
3. To derive the data needed by this class we have to write some code. For example to get the record count we need to get all the counters from job and check if the record counter exists. I see this class as a place to collect that data.

chavdar: You can still use the Job/state objects to instantiate the Event. What I am saying is that event should not keep those pointers.

pcadabam:
@chavdar and @ibuenros 
I have made changes as per our discussion this morning. Please take a look.
Changes -
- Created an SlaEventSubmitter which is a wrapper around the existing EventSubmitter. This class allows clients to inject sla metadata through state or through a builder. The class is responsible for validating the required fields.
- In gobblin-compaction added a helper class to populate the sla metadata required by the SlaEventSumbitter

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,86 @@
...
+  private String partition;
+  private String originTimestamp;
+  private String upstreamTimestamp;
+  private String recordCount;
+  private String previousPublishTimestamp;
+  private String dedupeStatus;
+  private String completenessPercentage;
+
+  public static SlaEventSubmitter of(String eventName, EventSubmitter eventSubmitter, Properties props) {
+    return SlaEventSubmitter.builder().eventName(eventName).eventSubmitter(eventSubmitter)
```
ibuenros: Can we make event submitter part of the constructor of the builder to match the API of the `TimingEvent`?

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,86 @@
...
+        .datasetUrn(SlaEventKeys.DATASET_URN_KEY).partition(SlaEventKeys.PARTITION_KEY)
```
ibuenros: I don't quite understand what this method does. It seems its filling an SLA event with key names.

pcadabam: Good catch. I will change it. It should read values for those keys from props

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,86 @@
...
+        .originTimestamp(SlaEventKeys.ORIGIN_TS_KEY).upstreamTimestamp(SlaEventKeys.UPSTREAM_TS_KEY)
+        .completenessPercentage(SlaEventKeys.COMPLETENESS_PERCENTAGE_KEY).recordCount(SlaEventKeys.RECORD_COUNT_KEY)
+        .previousPublishTimestamp(SlaEventKeys.PREVIOUS_PUBLISH_TS_KEY).dedupeStatus(SlaEventKeys.DEDUPE_STATUS_KEY)
+        .build();
+  }
+
+  public void submit() {
```
ibuenros: I think `submit` should allow passing in custom metadata. Users might want to include an extra field or two without having to extend the whole class.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,86 @@
...
+    eventMetadataMap.put(withoutPropertiesPrefix(SlaEventKeys.UPSTREAM_TS_KEY), upstreamTimestamp);
+    eventMetadataMap.put(withoutPropertiesPrefix(SlaEventKeys.COMPLETENESS_PERCENTAGE_KEY), completenessPercentage);
+    eventMetadataMap.put(withoutPropertiesPrefix(SlaEventKeys.RECORD_COUNT_KEY), recordCount);
+    eventMetadataMap.put(withoutPropertiesPrefix(SlaEventKeys.PREVIOUS_PUBLISH_TS_KEY), previousPublishTimestamp);
+    eventMetadataMap.put(withoutPropertiesPrefix(SlaEventKeys.DEDUPE_STATUS_KEY), dedupeStatus);
+
+    return Maps.newHashMap(Maps.filterValues(eventMetadataMap, NOT_NULL_OR_EMPTY_PREDICATE));
+  }
+
+  private String withoutPropertiesPrefix(String key) {
```
ibuenros: Can you include javadoc with and example of what this does?

gobblin-compaction/build.gradle
```diff
@@ -14,6 +14,9 @@ apply plugin: 'eclipse'
 dependencies {
   compile project(":gobblin-api")
   compile project(":gobblin-utility")
+  compile project(":gobblin-metrics")
+  compile project(":gobblin-core")
+  compile project(":gobblin-runtime")
```
ibuenros: what are you using gobblin-runtime for?

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -0,0 +1,89 @@
...
+      state.setProp(SlaEventKeys.PREVIOUS_PUBLISH_TS_KEY, Long.toString(fileStatus.getModificationTime()));
+    } catch (IOException e) {
+    }
+  }
+
+  private static void setDedupeStatus(State state) {
+    if (state.getPropAsBoolean(ConfigurationKeys.COMPACTION_DEDUPLICATE,
+        ConfigurationKeys.DEFAULT_COMPACTION_DEDUPLICATE)) {
+      state.setProp(SlaEventKeys.DEDUPE_STATUS_KEY, DedupeStatus.DEDUPED);
+
```
ibuenros: are you supposed to return here?

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: I am trying to use lombok's standard builder. I have changed this static method to a constructor to match TimingEvent. I would like to avoid writing my own builder unless it is required. Let me know if you still think it is necessary, I will change it.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: For additional metadata I was thinking of adding a member to the SlaEventSubmitter. You can set it using the builder. This field can be reused by the constructor that uses props to build the SlaEventSubmitter. We can have some helper methods to serialize any map into props.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: Added java doc to all methods

gobblin-compaction/build.gradle
pcadabam: Removed

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
pcadabam: These methods set properties in the state that is passed in by reference.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventKeys.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package gobblin.metrics.event.sla;
+
+public class SlaEventKeys {
+
+  static final String EVENT_GOBBLIN_STATE_PREFIX = "event.sla.";
+  public static final String DATASET_URN_KEY = EVENT_GOBBLIN_STATE_PREFIX + "datasetUrn";
+  public static final String PARTITION_KEY = EVENT_GOBBLIN_STATE_PREFIX + "partition";
+  public static final String ORIGIN_TS_KEY = EVENT_GOBBLIN_STATE_PREFIX + "originTimestamp";
```
chavdar: Can you make the units obvious from the name ?

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -0,0 +1,89 @@
...
+      Counters counters = job.getCounters();
+
+      if (counters != null) {
+        Counter recordCounter = counters.findCounter(AvroKeyDedupReducer.EVENT_COUNTER.RECORD_COUNT);
+        if (recordCounter != null) {
+          state.setProp(SlaEventKeys.RECORD_COUNT_KEY, Long.toString(recordCounter.getValue()));
+        }
+      }
+    } catch (IOException e) {
+    }
```
chavdar: Who throws IOException? Could you log anything here?

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
ibuenros: What I meant is you probably want to return here, as line 70 overwrites the property you just set.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
ibuenros: That works, can you add the method though?

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
ibuenros: No, it's fine. The benefits of Lombok trump this detail.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: added

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
pcadabam: Done

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventKeys.java
pcadabam: Added units

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
pcadabam: job.getCounters throws IOException. Add logging

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,128 @@
...
+  private String eventName;
+  private String datasetUrn;
+  private String partition;
+  private String originTimestamp;
+  private String upstreamTimestamp;
+  private String recordCount;
+  private String previousPublishTimestamp;
+  private String dedupeStatus;
+  private String completenessPercentage;
+  private Map<String, String> additionalMetadata;
```
ibuenros: can you add `@singular` annotation to this? https://projectlombok.org/features/Builder.html

ibuenros:
just one last comment, otherwise LGTM

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: Done

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -0,0 +1,95 @@
...
+
+    Path compactionCompletePath =
+        new Path(state.getProp(ConfigurationKeys.COMPACTION_JOB_DEST_DIR),
+            ConfigurationKeys.COMPACTION_COMPLETE_FILE_NAME);
+
+    try {
+      FileStatus fileStatus = fs.getFileStatus(compactionCompletePath);
+      state.setProp(SlaEventKeys.PREVIOUS_PUBLISH_TS_IN_MILLI_SECS_KEY, Long.toString(fileStatus.getModificationTime()));
+    } catch (IOException e) {
+    }
```
chavdar: Log an error and/or rethrow an IOException

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -0,0 +1,129 @@
+package gobblin.metrics.event.sla;
+
+import java.util.Map;
+import java.util.Properties;
+
+import lombok.AccessLevel;
```
chavdar: Since there are more and more Lombok dependencies, please write a github wiki on how to setup Lombok.

chavdar:
Other than my comments about swallowing IOExceptions and lombok setup, LGTM.

gobblin-metrics/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
pcadabam: Sure. Currently we have this page https://github.com/linkedin/gobblin/wiki/IDE-setup that describes lombok at a high level. It points to lombok's IDE setup page. Do we want to add more?

PR #342, opened 2015-09-25
BUGFIX: number of branches ignored when using AvroHdfsTimePartitionedWriter

mwol:
Using Branches (fork.branches > 1) in combination with AvroHdfsTimePartitionedWriter ignores the number of branches property and therefore leads to error messages about missing properties like 'Missing required property writer.staging.dir' or thows Exceptions ...


zliu41:
LGTM. Thanks for the fix.

PR #343, opened 2015-09-25
Fix EOFException with Avro DataFileReader

sahilTakiar:
@ydailinkedin can you review?

Small fix so that `hasNext()` does not get called multiple times before `next()` is invoked. This should fix the `EOFException`s we are seeing when reading specific Avro files.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
@@ -18,10 +18,10 @@ import java.util.Iterator;
 import java.util.List;
 
 import org.apache.commons.io.IOUtils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
@@ -45,7 +45,7 @@ import gobblin.source.workunit.WorkUnit;
  */
 public abstract class FileBasedExtractor<S, D> extends InstrumentedExtractor<S, D> {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(FileBasedExtractor.class);
+  private static final Logger LOG = LoggerFactory.getLogger(FileBasedExtractor.class);
 
   protected final WorkUnit workUnit;
   protected final WorkUnitState workUnitState;
@@ -59,7 +59,8 @@ public abstract class FileBasedExtractor<S, D> extends InstrumentedExtractor<S, 
 
   private Iterator<D> currentFileItr;
   private String currentFile;
-  private boolean readRecordStart;
+  private boolean hasNext = false;
+  private final boolean shouldSkipFirstRecord;
 
   protected enum CounterNames {
     FileBytesRead;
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -118,7 +118,8 @@ public D readRecordImpl(@Deprecated D reuse) throws DataRecordException, IOExcep
       readRecordStart = true;
     }
 
-    while ((currentFileItr == null || !currentFileItr.hasNext()) && !filesToPull.isEmpty()) {
+    boolean hasNext = currentFileItr.hasNext();
+    while ((currentFileItr == null || !hasNext) && !filesToPull.isEmpty()) {
```
chavdar: That's not correct because you have taken currentFileItr.hasNext() out of the while loop.

sahilTakiar: Yes, you are right, that will not work. Addressed in new updated.

sahilTakiar:
Updated with new approach, writing a unit test to ensure logic is correct.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -59,7 +59,7 @@
 
   private Iterator<D> currentFileItr;
   private String currentFile;
-  private boolean readRecordStart;
+  private boolean hasNext;
```
chavdar: Please make sure the variable is properly initialized.

sahilTakiar:
@chavdar and @ydailinkedin updated with new approach. The logic is similar to the previous version, but there is a new instance variable called `hasNext` which basically allows the code to avoid duplicate calls to `hasNext()`. I added some unit tests in the file `FileBasedExtractorTest` which basically check that given different values for `ConfigurationsKeys.SOURCE_FILEBASED_FILES_TO_PULL` that the code correctly iterates through all the records.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -174,12 +186,12 @@ public long getHighWatermark() {
...
   @SuppressWarnings("unchecked")
   public Iterator<D> downloadFile(String file) throws IOException {
-    LOGGER.info("Beginning to download file: " + file);
+    LOG.info("Beginning to download file: " + file);
 
     try {
       InputStream inputStream = this.closer.register(this.fsHelper.getFileStream(file));
       Iterator<D> fileItr = (Iterator<D>) IOUtils.lineIterator(inputStream, ConfigurationKeys.DEFAULT_CHARSET_ENCODING);
-      if (workUnitState.getPropAsBoolean(ConfigurationKeys.SOURCE_SKIP_FIRST_RECORD, false) && fileItr.hasNext()) {
+      if (this.workUnitState.getPropAsBoolean(ConfigurationKeys.SOURCE_SKIP_FIRST_RECORD, false) && fileItr.hasNext()) {
```
chavdar: Please move the access to the config property to the constructor and initialize a member variable. Any idea why we event have a property like that? Seems very dangerous to me.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -59,7 +59,8 @@
 
   private Iterator<D> currentFileItr;
   private String currentFile;
-  private boolean readRecordStart;
+  private boolean readRecordStart = false;
```
chavdar: I don't think you need this variable. If closeCurrentFile() sets currentFile to null, then you can use null != currentFile instead of readRecordStart.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
sahilTakiar: I believe the use case was for ingestion of csv data from certain sources. Some sources have files where the first line of the file is a list of the column names in the file, and then each subsequent row would be an actual record.

sahilTakiar:
@chavdar updated

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -102,36 +105,45 @@ public D readRecordImpl(@Deprecated D reuse) throws DataRecordException, IOExcep
...
 
-    if (currentFileItr != null && currentFileItr.hasNext()) {
-      return currentFileItr.next();
-    } else {
-      LOGGER.info("Finished reading records from all files");
-      return null;
+    while (!this.hasNext && !this.filesToPull.isEmpty()) {
+      this.currentFile = this.filesToPull.remove(0);
+      this.currentFileItr = downloadFile(this.currentFile);
+      this.hasNext = this.currentFileItr.hasNext();
```
ydailinkedin: Do you need to check for currentFileItr == null here?

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedExtractor.java
```diff
@@ -59,7 +59,8 @@
 
   private Iterator<D> currentFileItr;
   private String currentFile;
-  private boolean readRecordStart;
+  private boolean hasNext = false;
+  private boolean shouldSkipFirstRecord;
```
chavdar: should be final

chavdar:
One minor comment. Other than that, LGTM

PR #344, opened 2015-09-27
Kafka Adapter Performance Enhancement

zliu41:
- Now using the avg time to pull a record for mapper load balancing, instead of avg record size. This is more accurate since the time spent on a record is not necessarily proportional to its size. This is done in the following way:
 - Each partition pulled in the previous run has an avg time per record in the workunitstate.
 - In the next run, the estimated avg time per record for each topic is the geometric mean of the avg time per record of all partitions. For example if a topic has two partitions whose avg time per record in the previous run are 2 and 8, the next run will use 4 as the estimated avg time per record. The reason to choose **geometric mean** over algebraic mean is because large numbers are likely outliers, e.g., a topic may have 5 partitions, and the avg time per record collected from the previous run could sometimes be [1.1, 1.2, 1.1, 1.3, 100].
 - If a topic was not pulled in the previous run, its estimated avg time per record is the geometric mean of the estimated avg time per record of all topics that were pulled in the previous run. If no topic was pulled in the previous run, then we use a default value 1.0.

- Previously `KafkaSource` uses a bi-level bin packing approach for assigning workunits to mappers. Now added an option to use single-level bin packing, which is the default option. Bi-level bin packing has two advantages: (1) reduce the number of small output files since it tends to pack partitions of the same topic together; (2) reduce the total number of workunits / tasks since multiple partitions of the same topic are assigned to the same task. A task has a non-trivial cost of initialization, tear down and task state persistence. **However**, bi-level bin packing has more mapper skew than single-level bin packing, because if we pack lots of partitions of the same topic to the same mapper, and we underestimate the avg time per record for this topic, then this mapper could be much slower than other mappers. In my observation, the advantages of single-level bin packing outweigh the disadvantages.
 - Note that for each skipped topic, we create an empty workunit for each partition to preserve its checkpoints. In single-level bin packing mode, we still assign all partitions of a skipped topic to the same workunit / task, so that a single empty task will be created for this topic, instead of one empty task per partition. So single-level bin packing is more like 1.5-level bin packing.

- Modified the way mapper load is calculated: added a new factor of `log(number of tasks assigned to this mapper)`. This is because, as said above, each task is associated with a non-trivial cost. So, running 1000 small tasks in a mapper takes much longer than running 3 big tasks with the same amount of data.

- Previously the existence of topic schemas was verified in the mapper, so if a topic has no schema, we would still create workunits for it as usual, e.g., create 100 huge workunits which occupy 100 mappers, and these mappers ended up doing nothing. Now this is verified in the driver and disqualified topics will be skipped.

- Some other small changes and refactoring, including changing the "Queue stats of fork" log message to debug level since it sometimes floods the mapper log.

Tested on Nertz and this has lead to a significant reduction in the time each run takes.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -13,11 +13,8 @@
 package gobblin.source.extractor.extract.kafka;
 
 import java.io.IOException;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.PriorityQueue;
 import java.util.Set;
 import java.util.regex.Pattern;
 
@@ -27,25 +24,19 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.collect.MinMaxPriorityQueue;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closer;
-import com.google.common.primitives.Longs;
 import com.google.gson.Gson;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.SourceState;
 import gobblin.configuration.State;
 import gobblin.configuration.WorkUnitState;
-import gobblin.metrics.GobblinMetrics;
-import gobblin.metrics.Tag;
-import gobblin.source.extractor.WatermarkInterval;
 import gobblin.source.extractor.extract.EventBasedSource;
+import gobblin.source.extractor.extract.kafka.workunit.packer.KafkaWorkUnitPacker;
 import gobblin.source.workunit.Extract;
-import gobblin.source.workunit.MultiWorkUnit;
 import gobblin.source.workunit.WorkUnit;
 import gobblin.util.DatasetFilterUtils;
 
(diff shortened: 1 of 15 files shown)


zliu41:
@chavdar could you review?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -154,76 +164,120 @@ private void createEmptyWorkUnitsForSkippedPartitions(Map<String, List<WorkUnit>
...
-    for (List<WorkUnit> workUnitsForTopic : workUnits.values()) {
+    double totalEstDataSize = 0;
+    for (List<WorkUnit> workUnitsForTopic : workUnitsByTopic.values()) {
       for (WorkUnit workUnit : workUnitsForTopic) {
         setWorkUnitEstSize(workUnit);
         totalEstDataSize += getWorkUnitEstSize(workUnit);
       }
     }
-    long avgGroupSize = (long) ((double) totalEstDataSize / (double) numOfMultiWorkunits / 3.0);
+    if (isWorkUnitPreGroupingEnabled(state)) {
```
chavdar: Can you abstract-away the bin-packing strategy with an interface and a couple of implementing classes. This will make adding new strategies easier in the future and the code cleaner?

chavdar:
Is it possible to emit some metric to measure the effectiveness of the bin-packing algorithm? Say, comparing the prediction with the observed time after the run is done.

zliu41:
@chavdar updated. Added the following interfaces and classes: 
- `KafkaWorkUnitPacker`
 - `KafkaSingleLevelWorkUnitPacker`
 - `KafkaBiLevelWorkUnitPacker`
- `KafkaWorkUnitSizeEstimator`
 - `KafkaAvgRecordTimeBasedWorkUnitSizeEstimator`
 - `KafkaAvgRecordSizeBasedWorkUnitSizeEstimator`

For metrics, we can't (and don't need to) estimate the exact time each mapper takes. We only estimate the relative load among different mappers, which hopefully is proportional to their running times. So I've added the min estimated load and max estimated load of all mappers to the state, so that MetricReport reporter can report it to MetricReport which can then be compared against the actual mapper running time.

chavdar:
Mostly minor comments. Overall, LGTM

PR #345, opened 2015-09-28
Enable cleaning entire job's staging/output data at once

zliu41:
@liyinan926 can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -319,7 +319,7 @@ public abstract class AbstractJobLauncher implements JobLauncher {
    * Subclasses can override this method to do whatever processing on the {@link TaskState}s,
    * e.g., aggregate task-level metrics into job-level metrics.
    */
-  protected void postProcessTaskStates(@SuppressWarnings("unused") List<TaskState> taskStates) {
+  protected void postProcessTaskStates(List<TaskState> taskStates) {
     // Do nothing
   }
 
@@ -453,8 +453,19 @@ public abstract class AbstractJobLauncher implements JobLauncher {
    * and data is successfully committed because the staging data has already been moved
    * to the job output directory. But in case the job fails and data is not committed,
    * we want the staging data to be cleaned up.
+   *
+   * Property {@link ConfigurationKeys#CLEANUP_STAGING_DATA_PER_TASK} controls whether to cleanup
+   * staging data per task, or to cleanup entire job's staging data at once.
    */
   private void cleanupStagingData(JobState jobState) {
+    if (this.jobContext.shouldCleanupStagingDataPerTask()) {
+      cleanupStagingDataPerTask(jobState);
+    } else {
+      cleanupStagingDataForEntireJob(jobState);
+    }
+  }
+
+  private static void cleanupStagingDataPerTask(JobState jobState) {
     Closer closer = Closer.create();
     Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();
     try {
@@ -471,6 +482,14 @@ public abstract class AbstractJobLauncher implements JobLauncher {
       } catch (IOException e) {
         LOG.error("Failed to clean staging data", e);
       }
+    }
+  }
+
+  private static void cleanupStagingDataForEntireJob(JobState jobState) {
+    try {
+      JobLauncherUtils.cleanJobStagingData(jobState, LOG);
+    } catch (IOException e) {
+      LOG.error("Failed to clean staging data for job " + jobState.getJobId(), e);
     }
   }
 }
--- a/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
@@ -114,6 +114,39 @@ public class JobLauncherUtils {
   }
 
   /**
+   * Cleanup staging data of all tasks of a job.
+   *
+   * @param state job state
+   */
+  public static void cleanJobStagingData(State state, Logger logger) throws IOException {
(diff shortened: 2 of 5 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -316,10 +332,40 @@ public void launchJob(JobListener jobListener) throws JobException {
...
+   *
+   * Otherwise, if {@link ConfigurationKeys#TASK_DATA_ROOT_DIR_KEY} is specified, use its value plus "/task-staging".
+   */
+  private static void setTaskStagingDir(State state) {
+    if (state.contains(ConfigurationKeys.WRITER_STAGING_DIR)) {
+      LOG.warn("Property " + ConfigurationKeys.WRITER_STAGING_DIR + " is deprecated. No need to use it if "
+          + ConfigurationKeys.TASK_DATA_ROOT_DIR_KEY + " is specified.");
+    } else {
+      String workingDir = state.getProp(ConfigurationKeys.TASK_DATA_ROOT_DIR_KEY);
+      state.setProp(ConfigurationKeys.WRITER_STAGING_DIR, new Path(workingDir, "task-staging").toString());
```
liyinan926: `task-staging` can be defined as a constant.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -316,10 +332,40 @@ public void launchJob(JobListener jobListener) throws JobException {
   }
 
   /**
+   * If {@link ConfigurationKeys#WRITER_STAGING_DIR} (which is deprecated) is specified, use its value.
+   *
+   * Otherwise, if {@link ConfigurationKeys#TASK_DATA_ROOT_DIR_KEY} is specified, use its value plus "/task-staging".
+   */
+  private static void setTaskStagingDir(State state) {
+    if (state.contains(ConfigurationKeys.WRITER_STAGING_DIR)) {
+      LOG.warn("Property " + ConfigurationKeys.WRITER_STAGING_DIR + " is deprecated. No need to use it if "
```
liyinan926: You probably can use `String.format` here.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -473,4 +535,12 @@ private void cleanupStagingData(JobState jobState) {
       }
     }
   }
+
+  private static void cleanupStagingDataForEntireJob(JobState jobState) {
+    try {
+      JobLauncherUtils.cleanJobStagingData(jobState, LOG);
+    } catch (IOException e) {
+      LOG.error("Failed to clean staging data", e);
```
liyinan926: Better to also include the job id here, similarly to line 525.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -453,8 +499,24 @@ private void unlockJob() {
...
-  private void cleanupStagingData(JobState jobState) {
+  private static void cleanupStagingData(JobState jobState) {
+    if (shouldCleanupStagingDataPerTask(jobState)) {
+      cleanupStagingDataPerTask(jobState);
+    } else {
+      cleanupStagingDataForEntireJob(jobState);
+    }
+  }
+
+  protected static boolean shouldCleanupStagingDataPerTask(JobState jobState) {
```
liyinan926: This method can also be moved to `JobContext`.

gobblin-utility/src/main/java/gobblin/util/JobLauncherUtils.java
```diff
@@ -114,6 +114,39 @@ public static void cleanStagingData(List<? extends State> states, Logger logger)
...
+    String writerFsUri = state.getProp(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, ConfigurationKeys.LOCAL_FS_URI);
+    FileSystem fs = getFsWithProxy(state, writerFsUri);
+
+    Path jobStagingPath = new Path(state.getProp(ConfigurationKeys.WRITER_STAGING_DIR));
+    logger.info("Cleaning up staging directory " + jobStagingPath);
+    HadoopUtils.deletePath(fs, jobStagingPath, true);
+
+    if (fs.listStatus(jobStagingPath.getParent()).length == 0) {
+      logger.info("Deleting directory " + jobStagingPath.getParent());
+      HadoopUtils.deletePath(fs, jobStagingPath.getParent(), true);
```
liyinan926: You can return after this.

zliu41: Each run will create a `task-data/[jobId]/task-staging` dir and `task-data/[jobId]/task-output` dir. I need to delete the parent of task staging and task output (i.e `task-data/[jobId]`) to avoid creating an empty dir in each run.

And since this structure is not enforced (one may set `writer.staging.dir` and `writer.output.dir` to any dir even though they are deprecated), I can't return here because I still need to check whether task output dir is deleted. I also can't just delete `task-data/[jobId]` for this reason.

zliu41:
@liyinan926 updated

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -243,6 +248,20 @@ public void launchJob(JobListener jobListener) throws JobException {
       jobState.setStartTime(startTime);
       jobState.setState(JobState.RunningState.RUNNING);
 
+      // Set up task staging dir and task output dir
```
liyinan926: Not sure why a comment was not showing up here. But I think this part is config-related and should better be put into `JobContext`.

zliu41: Fixed.

liyinan926:
LGTM.

PR #346, opened 2015-09-28
Implementation of async trash for parallel deletions.

ibuenros:
Implementation of async trash for parallel deletions.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/trash/AsyncTrash.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/trash/AsyncTrash.java
@@ -14,15 +14,14 @@ package gobblin.data.management.trash;
 
 import gobblin.util.Decorator;
 import gobblin.util.ExecutorsUtils;
+import gobblin.util.executors.ScalingThreadPoolExecutor;
 
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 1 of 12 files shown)


ibuenros:
@ydailinkedin @chavdar can you review? Thanks!

gobblin-data-management/src/main/java/gobblin/data/management/trash/AsyncTrash.java
```diff
@@ -0,0 +1,147 @@
...
+ *   implementation that will actually perform the deletes. This class acts as a {@link Decorator} of the
+ *   inner trash.
+ * </p>
+ *
+ * <p>
+ *   Trash methods will always return true, regardless of success of the actual trash operation. However, additional
+ *   methods are provided to get a future for the operation.
+ * </p>
+ */
+public class AsyncTrash implements GobblinProxiedTrash, Closeable, Decorator {
```
chavdar: Please use the approach from the parallel cleaner:
- Use a thread pool executor with 0 to DEFAULT_MAX_DELETING_THREADS threads
- Have a rate limiter to limit the number of operations per second

ibuenros: The rate limiter should be implemented through the filesystem, not in the trash itself. I will change the thread pool executor.

ibuenros: Actually, after trying this change and some investigation, the thread pool executor used for the parallel cleaner doesn't do what we thought it would, and it in fact will never create more than one thread. See https://github.com/kimchy/kimchy.github.com/blob/master/_posts/2008-11-23-juc-executorservice-gotcha.textile .
There is a work-around in that blog post, but I don't know what the licensing is.

chavdar: What about using a queue with fixed capacity - http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ArrayBlockingQueue.html ?

ibuenros:
@chavdar fixed the thread pool.

gobblin-data-management/src/main/java/gobblin/data/management/trash/GobblinProxiedTrash.java
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public boolean moveToTrashAsUser(Path path, final String user) throws IOException;
+
+  /**
+   * Move the path to trash as the owner of the path.
+   * @param path {@link org.apache.hadoop.fs.Path} to move.
+   * @return true if the move succeeded.
+   * @throws IOException
+   */
+  public boolean moveToTrashAsOwner(Path path) throws IOException;
```
ydailinkedin: I don't think this method is necessary for the interface. It can always be done by calling moveToTrashAsUser.

chavdar:
LGTM

ibuenros:
@ydailinkedin addressed comment.

ydailinkedin:
LGTM

PR #348, opened 2015-09-29
Fix incorrect FindBugs exclude rule

chavdar:
The previous commit inadvertently disabled another FindBugs exclude rule.

@ydailinkedin  can you review?


ydailinkedin:
LGTM

PR #349, opened 2015-09-29
One more FindBugs warning to ignore

chavdar:
@ydailinkedin  There is one more warning to ignore (for interfaces)


ydailinkedin:
LGTM

PR #350, opened 2015-09-29
Add env support to gobblin-standalone.sh

ahollenbach:
Changes to `gobblin-standalone.sh`:
- Allows values to be set via environment variables that can be
overridden by passing in the values as args
- Explicit setting of `FWDIR`, while still defaulting to the original value for
legacy reasons
- Explicit exits on status check


ahollenbach:
Most changes pulled in from @teasherm

ahollenbach:
@liyinan926

liyinan926:
LGTM. Thanks @ahollenbach!

PR #351, opened 2015-09-30
ETL-3519: Add mysql utils to gobblin-utility

ydailinkedin:
Tested using gobblin kafka jobs. @zliu41 @chavdar Can you review? Thanks!


liyinan926:
@ydailinkedin there's already a `DataSourceProvider` in `gobblin-metastore` that can be used to get a `DataSource` instance. See if you can use that instead.

ydailinkedin:
Close this one, since it somehow mixed with @zliu41 's commits

PR #352, opened 2015-09-30
Transparently handles rate controlled file systems when proxying as a…

ibuenros:
…nother user.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
@@ -18,10 +18,13 @@ import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.NonNull;
 
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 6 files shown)


ibuenros:
@ydailinkedin can you review?

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
```diff
@@ -134,16 +149,179 @@ public static FileSystem getProxiedFileSystemUsingKeytab(State state, URI fsURI,
...
+      return getProxiedFileSystemUsingToken(userNameToProxyAs, userNameToken, fsURI, conf, null);
+    } catch(IOException ioe) {
+      throw new ExecutionException(ioe);
+    }
+  }
+
+  /**
+   * Cached version of {@link ProxiedFileSystemUtils#createProxiedFileSystemUsingToken(String, Token, URI, Configuration)}.
+   */
+  @Builder(builderClassName = "ProxiedFileSystemFromToken", builderMethodName = "fromToken")
```
ydai1124: Is there any advantage to deprecate previous methods and using builder here? The previous methods can also call  `decorateFilesystemFromReferenceFS` to get a `ratecontrolledFs`.

ydai1124:
Talked to @ibuenros . LGTM!

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemCache.java
ibuenros: The argument list is already long, and it might get even longer in the future. Since some of the arguments are optional, it requires us to also make a bunch of other methods with different combinations of arguments, which becomes unwieldy as we add more arguments. The builder pattern is intended to fix that.
The methods above are marked as deprecated so that users use the builder pattern.

Issue #340, opened 2015-09-23
Job/Task Metrics Deadlock


akshayn1107:
We noticed a deadlock with `gobblin.runtime.util.TaskMetrics` and `gobblin.runtime.util.JobMetrics`. We are running gobblin in standalone mode. I believe the issue is with the `synchronized` methods `get` and `remove` in the two classes mentioned above. The `jstack` stacktrace is below:

```
2015-09-23 01:36:53
Full thread dump Java HotSpot(TM) 64-Bit Server VM (24.65-b04 mixed mode):

"Attach Listener" daemon prio=10 tid=0x00007fce34002000 nid=0x5ae9 waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"ForkExecutor-1" prio=10 tid=0x00007fcdf80b1000 nid=0x2e4a waiting on condition [0x00007fce213ec000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x000000078e6dd3a8> (a java.util.concurrent.SynchronousQueue$TransferStack)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
	at java.util.concurrent.SynchronousQueue$TransferStack.awaitFulfill(SynchronousQueue.java:458)
	at java.util.concurrent.SynchronousQueue$TransferStack.transfer(SynchronousQueue.java:359)
	at java.util.concurrent.SynchronousQueue.take(SynchronousQueue.java:925)
	at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1068)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)

"ForkExecutor-0" prio=10 tid=0x00007fcdf8089000 nid=0x2e49 waiting on condition [0x00007fce210e9000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x000000078e6dd3a8> (a java.util.concurrent.SynchronousQueue$TransferStack)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
[...]

PR #355, opened 2015-10-01
Fixed a bug that may cause a deadlock reported in issue #340

liyinan926:
The deadlock may happen because:

1. One thread calls `JobMetrics.remove`, which tries to remove itself as well as the `TaskMetrics` of all the tasks from the `GobblinMetricsRegistry` while holding a lock on `JobMetrics.class` since `JobMetrics.remove` is `synchronized`. When it tries to remove a `TaskMetrics` it also needs to acquire the lock on `TaskMetrics.class` since `TaskMetrics.remove` is `synchronized`.
2. Another thread instantiates a `TaskContext`, which calls `TaskMetrics.get` and needs to acquire the lock on `TaskMetrics.class`. In the constructor of `TaskMetrics`, however, the method `parentContextForTask` will call `JobMetrics.get`, which will need to acquire the lock on `JobMetrics.class`.
3. So the first thread needs to acquire both locks in the order of {`JobMetrics.remove`, `TaskMetrics.remove`}, whereas the second thread needs to acquire both locks in the order of {`TaskMetrics.remove`, `JobMetrics.remove`}, resulting in a deadlock.

The fix is to remove all `synchronized` on both `get` and `remove` from both `JobMetrics` and `TaskMetrics` and delegate all synchronization work to `GobblinMetricsRegistry`.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
+++ b/gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
@@ -57,6 +57,8 @@ public class GobblinMetrics {
 
   public static final String METRICS_STATE_CUSTOM_TAGS = "metrics.state.custom.tags";
 
+  protected static final GobblinMetricsRegistry GOBBLIN_METRICS_REGISTRY = GobblinMetricsRegistry.getInstance();
+
   /**
    * Enumeration of metric types.
    */
@@ -94,7 +96,7 @@ public class GobblinMetrics {
    * @param id the given {@link GobblinMetrics} ID
    * @return a {@link GobblinMetrics} instance
    */
-  public synchronized static GobblinMetrics get(String id) {
+  public static GobblinMetrics get(String id) {
     return get(id, null);
   }
 
@@ -105,7 +107,7 @@ public class GobblinMetrics {
    * @param parentContext the given parent {@link MetricContext}
    * @return a {@link GobblinMetrics} instance
    */
-  public synchronized static GobblinMetrics get(String id, MetricContext parentContext) {
+  public static GobblinMetrics get(String id, MetricContext parentContext) {
     return get(id, parentContext, Lists.<Tag<?>>newArrayList());
   }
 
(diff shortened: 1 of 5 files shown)


liyinan926:
@zliu41 can you review?

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -118,12 +120,17 @@ public synchronized static GobblinMetrics get(String id, MetricContext parentCon
...
    * @return a {@link GobblinMetrics} instance
    */
-  public synchronized static GobblinMetrics get(String id, MetricContext parentContext, List<Tag<?>> tags) {
-    GobblinMetricsRegistry registry = GobblinMetricsRegistry.getInstance();
-    if (!registry.containsKey(id)) {
-      registry.putIfAbsent(id, new GobblinMetrics(id, parentContext, tags));
-    }
-    return registry.get(id);
+  public static GobblinMetrics get(String id, MetricContext parentContext, List<Tag<?>> tags) {
+    return GOBBLIN_METRICS_REGISTRY.putIfAbsent(id, new GobblinMetrics(id, parentContext, tags));
```
zliu41: `putIfAbsent` returns "the previous value associated with the specified key, or null if there was no mapping for the key.". So I think you should return `GOBBLIN_METRICS_REGISTRY.get(id)`.

liyinan926: Ah, good catch. Actually I need something similar to `ConcurrentMap.getOrDefault`.

liyinan926:
@zliu41 Updated 3c814cb.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
```diff
@@ -27,33 +27,69 @@
...
+   * is found.
+   *
+   * @param id the given {@link GobblinMetrics} ID
+   * @param defaultValue the default {@link GobblinMetrics} instance
+   * @return the {@link GobblinMetrics} instance associated with a given ID or the given default
+   *         {@link GobblinMetrics} instance if no {@link GobblinMetrics} instance for the given ID
+   *         is found
+   */
+  public GobblinMetrics getOrDefault(String id, GobblinMetrics defaultValue) {
+    return this.metricsMap.asMap().getOrDefault(id, defaultValue);
```
zliu41: `getOrDefault` is only available in Java 8.

gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
```diff
@@ -72,13 +71,8 @@ public static JobMetrics get(String jobId) {
...
    */
-  public static synchronized JobMetrics get(JobState jobState) {
-    GobblinMetricsRegistry registry = GobblinMetricsRegistry.getInstance();
-    String name = name(jobState);
-    if (!registry.containsKey(name)) {
-      registry.putIfAbsent(name, new JobMetrics(jobState));
-    }
-    return (JobMetrics) registry.get(name);
+  public static JobMetrics get(JobState jobState) {
+    return (JobMetrics) GOBBLIN_METRICS_REGISTRY.getOrDefault(name(jobState), new JobMetrics(jobState));
```
zliu41: Do you not need to put `new JobMetrics(jobState)` into the map?

liyinan926: No, we don't want to replace the existing `JobMetrics` if it already exists.

zliu41: But if it doesn't exist in the map, will it ever be put into the map?

liyinan926: Yes.

liyinan926:
Updated 1aaa7b6.

zliu41:
Please take a look at failed unit tests.

liyinan926:
@zliu41 Updated to fix the unit tests a6139f6.

zliu41:
LGTM

PR #356, opened 2015-10-01
Report events before closing EventReporter

pcadabam:
- Report all the events before closing the reporter
- Removed call to close all reporters in GobblinMetrics as the reporters are already registered with closer.

@ibuenros can you take a look at this?


ibuenros:
LGTM

PR #359, opened 2015-10-02
ETL-3519: Add mysql utils to gobblin-utility

ydai1124:
Added jdbc module to gobblin-util.
Deprecated DataSourceProvider in gobblin-metastore.


gobblin-metastore/src/main/java/gobblin/metastore/DataSourceProvider.java
```diff
@@ -12,42 +12,21 @@
...
- *
- * @author ynli
+ * {@inheritDoc}
+ * @deprecated This class is replaced by {@link JobHistoryDataSourceProvider}
  */
-public class DataSourceProvider implements Provider<DataSource> {
-
-  private final BasicDataSource basicDataSource;
+@Deprecated
+public class DataSourceProvider extends JobHistoryDataSourceProvider {
```
liyinan926: Actually you don't need to keep this and deprecate as this class is not part of the public API.

gobblin-metastore/src/main/java/gobblin/metastore/JobHistoryDataSourceProvider.java
```diff
@@ -0,0 +1,40 @@
...
+
+
+/**
+ * This class extends {@link gobblin.util.jdbc.DataSourceProvider} with its own property keys.
+ */
+public class JobHistoryDataSourceProvider extends gobblin.util.jdbc.DataSourceProvider {
+
+  @Inject
+  public JobHistoryDataSourceProvider(@Named("dataSourceProperties") Properties properties) {
+    super();
```
liyinan926: This is not necessary as it's called implicitly.

gobblin-utility/src/main/java/gobblin/util/jdbc/DataSourceModule.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A Guice module defining the dependencies used by the jdbc data source.
+ */
+public class DataSourceModule extends AbstractModule {
+
+  private final Properties properties;
+
+  public DataSourceModule(Properties properties) {
+    this.properties = properties;
```
liyinan926: I suggest making a copy here so any potential updates to the `properties` won't affect the behavior of this module.

ydai1124:
@liyinan926 Updated.

chavdar:
LGTM

PR #360, opened 2015-10-02
Add SingleTaskDataPublisher and restore DataPublisher to the original API

zliu41:
@ibuenros could you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/Task.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/Task.java
@@ -36,6 +36,7 @@ import gobblin.fork.ForkOperator;
 import gobblin.instrumented.extractor.InstrumentedExtractorBase;
 import gobblin.instrumented.extractor.InstrumentedExtractorDecorator;
 import gobblin.publisher.DataPublisher;
+import gobblin.publisher.SingleTaskDataPublisher;
 import gobblin.qualitychecker.row.RowLevelPolicyCheckResults;
 import gobblin.qualitychecker.row.RowLevelPolicyChecker;
 import gobblin.runtime.util.RuntimeConstructs;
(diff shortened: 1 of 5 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -279,7 +280,8 @@ private void publishTaskData() throws IOException {
       @SuppressWarnings("unchecked")
       Class<? extends DataPublisher> dataPublisherClass = (Class<? extends DataPublisher>) Class.forName(
           this.taskState.getProp(ConfigurationKeys.DATA_PUBLISHER_TYPE, ConfigurationKeys.DEFAULT_DATA_PUBLISHER_TYPE));
-      DataPublisher publisher = closer.register(DataPublisher.getInstance(dataPublisherClass, this.taskState));
+      SingleTaskDataPublisher publisher =
```
ibuenros: This requires data publishers to be of type SingleTaskDataPublisher. If any user is already using a DataPublisher that is not of this type, Gobblin would fail.

zliu41: Publishing data in tasks is just recently added, unlikely anyone else is using it, even if someone is, chances are that he's extending basedatapublisher rather than directly extending datapublisher, which will be fine.

What's your suggestion?

ibuenros: Ok, I see this is only used if `publish.data.at.job.level` is false. Could you add a more descriptive error message to the catch block? Maybe mentions that publishing data at task level is only supported for publishers extending `SingleTaskDataPublisher`?

zliu41:
@ibuenros updated d2d4423

ibuenros:
LGTM

PR #361, opened 2015-10-02
Move clusterIdentifier logic from JobMetrics to ClusterNames

pcadabam:
@ibuenros please review

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ClustersNames.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ClustersNames.java
@@ -119,8 +119,20 @@ public class ClustersNames {
 
   /**
    *
-   * Returns the cluster name on which the application is running. Uses hadoop {@link Configuration} to get the url of
-   * the resourceManager or jobtracker. The URL is then translated into a human readable cluster name using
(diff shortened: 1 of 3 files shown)


gobblin-utility/src/main/java/gobblin/util/ClustersNames.java
```diff
@@ -112,6 +117,54 @@ private static String normalizeClusterUrl(String clusterIdentifier) {
...
+   * <p>
+   * <b>Standalone mode (outside of hadoop)</b> Uses the Hostname of {@link InetAddress#getLocalHost()}
+   * </p>
+   *
+   * <p>
+   * Use {@link #getClusterName(String)} if you already have the cluster URL
+   * </p>
+   *
+   */
+  public String getClusterName() {
```
ibuenros: We should make this take a Configuration as argument (in case not using default configuration).

pcadabam: Sure. Let's add an overload when we have such a use case.

pcadabam: Added support for passing configuration

ibuenros:
LGTM

PR #362, opened 2015-10-02
Made event submission synchronous to avoid missed events.

ibuenros:


ibuenros:
@pcadabam can you review?

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -663,9 +665,31 @@ public static Builder builder(String name) {
...
+    long attemptsLeft = 10;
+
+    while(attemptsLeft > 0) {
+      long key = rand.nextLong();
+      if(this.notificationTargets.putIfAbsent(key, target) == null) {
+        return key;
+      }
+      attemptsLeft--;
+    }
+
```
pcadabam: Use a UUID instead

pcadabam:
LGTM

gobblin-metrics/src/main/java/gobblin/metrics/MetricContext.java
```diff
@@ -663,9 +665,25 @@ public static Builder builder(String name) {
...
    * @param target A {@link com.google.common.base.Function} that will be run every time
    *               there is a new {@link gobblin.metrics.notification.Notification} in this context.
+   * @return a key for this notification target. Can be used to remove the notification target later.
    */
-  public void addNotificationTarget(Function<Notification, Void> target) {
-    this.notificationTargets.add(target);
+  public UUID addNotificationTarget(Function<Notification, Void> target) {
+
+    UUID uuid = UUID.randomUUID();
+    if(this.notificationTargets.putIfAbsent(uuid, target) == null) {
```
zliu41: @ibuenros `Map.putIfAbsent()` is not available in Java 6 and 7.

PR #363, opened 2015-10-02
MetricContext was using putIfAbsent incorrectly.

ibuenros:


ibuenros:
Merged because it was a quick fix.

PR #365, opened 2015-10-04
Missing properties in DatasetState

lbendig:
When a new ```DatasetState``` is created from a ```JobState``` object in ```JobState#newDatasetState```, the start time, end time and duration properties are not copied.
This will result 0 values in the job level properties in the state store files.


liyinan926:
LGTM.

lbendig:
@liyinan926 Thanks for committing it!

PR #366, opened 2015-10-04
Set Guava to version 15.0 in gobblin-mapreduce.sh

lbendig:
It seems to me that the Guava version was not adjusted in ```gobblin-mapreduce.sh``` when [Pull#280](https://github.com/linkedin/gobblin/pull/280) downgraded the version from 18 to 15.


lbendig:
@zliu41 Thanks for committing it!

PR #371, opened 2015-10-06
A couple of enhancement to Gobblin on Yarn

liyinan926:
1. Moved code that registers with a `EventBus` instance outside the constructors so to prevent objects to be registered from being exposed too early before exiting the constructors.
2. Use `TaskExecutor.execute` instead of `submit` whenever possible as `ExecutorService.submit` will not trigger uncaught exception handler if a task throws an unchecked exception.
3. Other minor code refactoring.


liyinan926:
@zliu41 can you review?

zliu41:
LGTM

PR #372, opened 2015-10-06
Kafka Adapter: Adding support for topic specific configs

sahilTakiar:
Adding a feature that allows users to specify configuration key, value pairs on a per topic basis. The approach is to use a config key called `kafka.topic.specific.state` that takes in a JSON array. More information is in the Java Docs.

Still working on finding a way to test the code.

@zliu41 can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -18,21 +18,34 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import lombok.Getter;
+import lombok.Setter;
+
 import org.apache.commons.lang.StringUtils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closer;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+
 import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.SourceState;
 import gobblin.configuration.State;
+import gobblin.configuration.StateUtils;
 import gobblin.configuration.WorkUnitState;
 import gobblin.source.extractor.extract.EventBasedSource;
 import gobblin.source.extractor.extract.kafka.workunit.packer.KafkaWorkUnitPacker;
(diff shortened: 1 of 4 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -83,31 +114,70 @@
...
-      KafkaWrapper kafkaWrapper = closer.register(KafkaWrapper.create(state));
-      List<KafkaTopic> topics = getFilteredTopics(kafkaWrapper, state);
-      for (KafkaTopic topic : topics) {
-        workUnits.put(topic.getName(), getWorkUnitsForTopic(kafkaWrapper, topic, state));
-      }
 
-      // Create empty WorkUnits for skipped partitions (i.e., partitions that have previous offsets,
-      // but aren't processed.
-      createEmptyWorkUnitsForSkippedPartitions(workUnits, state);
+    this.kafkaWrapper = KafkaWrapper.create(state);
```
zliu41: this line and the next can be combined.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -303,16 +376,21 @@ private boolean shouldMoveToLatestOffset(KafkaPartition partition, SourceState s
...
-    return getWorkUnitForTopicPartition(partition, offsets);
+    return getWorkUnitForTopicPartition(partition, offsets, topicSpecificState);
   }
 
-  private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, Offsets offsets) {
+  private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, Offsets offsets,
+      Optional<State> topicSpecificState) {
     Extract extract = this.createExtract(DEFAULT_TABLE_TYPE, DEFAULT_NAMESPACE_NAME, partition.getTopicName());
+    if (topicSpecificState.isPresent()) {
+      extract.addAll(topicSpecificState.get());
```
zliu41: Is there any benefit to add these states to `extract` instead of the workunit itself?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -389,10 +466,21 @@ private void startAtEarliestOffset() {
...
+   */
+  @AllArgsConstructor
+  private static class KafkaTopicNamePredicate implements Predicate<KafkaTopic> {
+
+    @NonNull
+    private final String topicName;
+
+    @Override
+    public boolean apply(KafkaTopic input) {
+      return input.getName().equals(this.topicName);
```
zliu41: please make it support regex so that we can specify `Voyager.*` instead of each individual topic.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
sahilTakiar: No, there is no functional benefit versus adding it on an `Extract` level vs. a `WorkUnit` level. Although I think it logically makes more sense to add it on an `Extract` level since each topic has a single `Extract`.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
sahilTakiar: Since I made `kafkaWrapper` an instance variable I have to assign it to `this.kafkaWrapper` first, and then register it. I don't think I can assign the variable and register it at the same time?

zliu41: I think you can do `this.kafkaWrapper = this.closer.register(KafkaWrapper.create(state));`

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
zliu41: It seems `workUnit.getProp()` cannot get properties in the extract of the workunit. Even if it can, the writer partition column property will be in both the workunit and the extract, causing conflict. So it's better to add them to the workunit.

sahilTakiar: Yes but the constructor `WorkUnit.create(extract)` will add all props in the `Extract` to the `WorkUnit`.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
sahilTakiar: Fixed

sahilTakiar:
@zliu41 updated and addressed comments. Added support for specifying regular expressions. Updated the unit tests accordingly.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
zliu41: It simply does `this.extract=extract`: https://github.com/linkedin/gobblin/blob/master/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java#L74

sahilTakiar: Fixed

sahilTakiar:
@zliu41 updated

zliu41:
LGTM, please add copyright block to `StateUtils` and `KafkaSourceTest`.

PR #373, opened 2015-10-07
Publish late kafka events to a separate dir if compaction is already done

ydai1124:
@zliu41 Can you review?


gobblin-compaction/src/main/java/gobblin/compaction/count/EventCountFromFileName.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Slf4j
+public class EventCountFromFileName implements HdfsEventCountProvider {
+  @Override
+  public long getEventCount(FileSystem fs, List<Path> paths) {
+    long eventCount = 0;
+    try {
+      for (Path path : paths) {
+          String[] parts = FilenameUtils.removeExtension(path.getName()).split("\\.");
+          eventCount += Long.parseLong(parts[parts.length - 1]);
```
zliu41: Instead of requiring the record count to be the last component (which isn't the case for compaction output files), one way to make it more flexible is to create an interface `FilenameFormat` with method `constructFilename(String... components)`, and create a subinterface `FilenameWithRecordCountFormat` with another method `getRecordCount(String fileName)`. Then pass an instance to this method.

Also please rename `event` to `record` in non-Kafka-specific classes since event is a concept specific to Kafka, and record is more general.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +410,48 @@ public Status status() {
...
+      paths.add(fileStatus.getPath());
+    }
+    return paths;
+  }
+
+  /**
+   * Submit an event reporting new and cumulative late event counts.
+   */
+  private void submitLateEvent(List<Path> lateFilePaths, Path lateDataPath) {
+    if (this.jobProps.contains(MRCompactor.COMPACTION_HDFS_EVENT_COUNT_PROVIDER)) {
```
zliu41: Whether to submit late event reporting should not be determined by whether `jobProps` contains `MRCompactor.COMPACTION_HDFS_EVENT_COUNT_PROVIDER` because (1) it's not intuitive; (2) if I don't want to submit late event but want to do some other things with row counts, then my states may contain this property.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -124,6 +132,9 @@
   public static final String COMPACTION_MR_JOB_TIMEOUT_MINUTES = COMPACTION_PREFIX + "mr.job.timeout.minutes";
   public static final long DEFAULT_COMPACTION_MR_JOB_TIMEOUT_MINUTES = Long.MAX_VALUE;
 
+  // The provider that provides event counts of given paths on HDFS.
+  public static final String COMPACTION_HDFS_EVENT_COUNT_PROVIDER = COMPACTION_PREFIX + "hdfs.count.provider";
```
zliu41: There should be a default value for this.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -105,6 +109,10 @@
   public static final String COMPACTION_DEST_SUBDIR = COMPACTION_PREFIX + "dest.subdir";
   public static final String DEFAULT_COMPACTION_DEST_SUBDIR = "daily";
 
+  // The subdir name of output dataset paths containing late events, e.g., "daily_late" in "/data/input/PasswordChangeEvent/daily_late/2015/09/06".
+  public static final String COMPACTION_DEST_LATE_SUBDIR = COMPACTION_PREFIX + "dest.late.subdir";
+  public static final String DEFAULT_COMPACTION_DEST_LATE_SUBDIR = "daily_late";
```
zliu41: You can just use `COMPACTION_DEST_SUBDIR` + `_late`, so that there's one less property the user needs to specify.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -427,14 +438,18 @@ private void createJobPropsForTopic(String topic, double priority) throws IOExce
...
     String builderClassName =
         this.state.getProp(COMPACTION_JOBPROPS_CREATOR_CLASS, DEFAULT_COMPACTION_JOBPROPS_CREATOR_CLASS) + "$Builder";
     boolean deduplicate = this.state.getPropAsBoolean(COMPACTION_DEDUPLICATE, DEFAULT_COMPACTION_DEDUPLICATE);
 
     try {
       return ((MRCompactorJobPropCreator.Builder<?>) Class.forName(builderClassName).newInstance()).withTopic(topic)
-          .withTopicInputDir(topicInputDir).withTopicOutputDir(topicOutputDir).withTopicTmpDir(topicTmpDir)
+          .withTopicInputDir(new Path(this.inputDir, new Path(topic, this.inputSubDir)))
+          .withTopicOutputDir(new Path(this.destDir, new Path(topic, this.destSubDir)))
+          .withTopicOutputBaseDir(new Path(this.destDir, topic))
```
zliu41: The style is not consistent here. Either change this line to `withTopicOutputLateDir(...)`, or if you want to use `withTopicOutputBaseDir()`, then you don't need `withTopicInputDir()`, `withTopicOutputDir()` and `withTopicTmpDir()` because they can all be constructed from base dir.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -154,8 +160,16 @@ public void run() {
...
-            this.deduplicate ? new Path(this.outputPath, MRCompactor.COMPACTION_LATE_FILES_DIRECTORY) : this.outputPath;
+        Path lateDataOutputPath = this.outputPath;
+        if (this.deduplicate) {
+          // Update the late data output path.
+          lateDataOutputPath = new Path(this.jobProps.getProp(MRCompactor.COMPACTION_JOB_DEST_LATE_DIR));
+          LOG.info(String.format("Will copy to late path %s since deduplication has been done. ", lateDataOutputPath));
+          if (!fs.exists(lateDataOutputPath)) {
+            fs.mkdirs(lateDataOutputPath);
+          }
+          this.submitLateEvent(lateFilePaths, lateDataOutputPath);
```
zliu41: `submitLateEvent` should also be called if `this.deduplicate=false`

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +410,48 @@ public Status status() {
...
+    })) {
+      paths.add(fileStatus.getPath());
+    }
+    return paths;
+  }
+
+  /**
+   * Submit an event reporting new and cumulative late event counts.
+   */
+  private void submitLateEvent(List<Path> lateFilePaths, Path lateDataPath) {
```
zliu41: Please rename parameters to `newLateFilePaths` and `lateDataDir`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -129,23 +135,23 @@ protected MRCompactorJobRunner(Dataset dataset, FileSystem fs, Double priority)
...
     Configuration conf = HadoopUtils.getConfFromState(this.jobProps);
-    DateTime jobStartTime = new DateTime(DateTimeZone
-        .forID(this.jobProps.getProp(MRCompactor.COMPACTION_TIMEZONE, MRCompactor.DEFAULT_COMPACTION_TIMEZONE)));
-
+    DateTime jobStartTime =
+        new DateTime(DateTimeZone.forID(this.jobProps.getProp(MRCompactor.COMPACTION_TIMEZONE,
+            MRCompactor.DEFAULT_COMPACTION_TIMEZONE)));
     try {
       if (this.jobProps.getPropAsBoolean(MRCompactor.COMPACTION_JOB_LATE_DATA_MOVEMENT_TASK, false)) {
         List<Path> lateFilePaths = Lists.newArrayList();
```
zliu41: Please rename to `newLateFilePaths` to avoid confusion with files in the `daily_late` folder.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +410,48 @@ public Status status() {
   public int compareTo(MRCompactorJobRunner o) {
     return Double.compare(o.priority, this.priority);
   }
+
+  private List<Path> getCurLateEventPath(Path lateDataPath) throws FileNotFoundException, IOException {
```
zliu41: Please rename to `getCumulatedLateFilePaths`.

ydai1124:
@zliu41 Updated.

ydai1124:
@zliu41 Updated.

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
```diff
@@ -0,0 +1,20 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+/**
+ * Interface for constructing a file name using given components as String.
+ */
+public interface FileNameFormat {
+  public abstract String constructFileName(String... components);
```
zliu41: no need to define method as abstract.

gobblin-compaction/src/main/java/gobblin/compaction/count/HdfsRecordCountProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.List;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * Get event count from HDFS, using given {@link FileSystem} and {@link List} of {@link Path}s.
+ */
+public interface HdfsRecordCountProvider {
```
zliu41: This can be used on other file systems than Hdfs. you can rename to `FsRecordCountProvider`

zliu41: Also not sure why you still need this interface. Can you just use a parameter that takes the name of a class that implements `FileNameWithRecordCountFormat`, then use the `getRecordCount` method in that class?

zliu41:
Currently when copy a file to late folder it always adds a component, if a file is first copied to hourly-late then to daily-late, it will add two components, making file names in the late folder inconsistent and hard to process. One option is to write a FilenameFormat for late data that does the following:
(1) It should enclose the original FilenameFormat class.
(2) If the file name doesn't exist in the late folder, do not rename.
(3) If it does exist, add a component "late[number]".
(4) `getRecordCount()` should remove all "late[number]" components, then call `getRecordCount()` of the enclose `FilenameFormat` instance.

gobblin-compaction/src/main/java/gobblin/compaction/count/HdfsRecordCountProvider.java
ydai1124: My assumption is that not all FsRecordCount is coming from file name.

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
```diff
@@ -0,0 +1,20 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+/**
+ * Interface for constructing a file name using given components as String.
+ */
+public interface FileNameFormat {
```
chavdar: I don't understand this abstraction. In how many ways can the components be concatenated? Shouldn't we fix what kind of components are concatenated?

gobblin-compaction/src/main/java/gobblin/compaction/count/RecordCountFromFileName.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public static final String FILE_NAME_WITH_RECORD_COUNT_FORMAT_CLASS = "filename.with.record.count.format.class";
+  public static final String DEFAULT_FILE_NAME_WITH_RECORD_COUNT_FORMAT_CLASS =
+      AvroHdfsTimePartitionedWithRecordCountsWriter.FileNameFormat.class.getName();
+
+  @Getter
+  @Setter
+  private FileNameWithRecordCountFormat fileNameFormat;
+
+  public RecordCountFromFileName(State properties) throws Exception {
```
chavdar: Please be more specific about the type of exceptions that can be thrown

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +418,97 @@ public Status status() {
...
+      }
+      return sb.toString();
+    }
+
+    @Override
+    public long getRecordCount(String fileName) {
+      if (!this.formatWithoutSuffix.isPresent()) {
+        throw new UnsupportedOperationException();
+      }
+      int last = fileName.lastIndexOf(SEPARATOR);
```
chavdar: What happens if the separator is missing or there is only one of it?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +418,97 @@ public Status status() {
...
+        if (i > 0) {
+          sb.append(SEPARATOR);
+        }
+        sb.append(components[i]);
+      }
+      return sb.toString();
+    }
+
+    @Override
+    public long getRecordCount(String fileName) {
```
chavdar: I don't understand what's going on in this method. It seems to have some implicit assumptions about what the components are but those are not documented.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +418,97 @@ public Status status() {
...
+      if (newLateRecordCountProvider instanceof RecordCountFromFileName) {
+        FileNameWithRecordCountFormat newLateRecordFileFormat =
+            ((RecordCountFromFileName) newLateRecordCountProvider).getFileNameFormat();
+        ((RecordCountFromFileName) lateDataDirRecordCountProvider).setFileNameFormat(new FileNameFormatWithSuffix(
+            Optional.of(newLateRecordFileFormat)));
+      }
+
+      Map<String, String> eventMetadataMap = Maps.newHashMap();
+      long newLateEventCount = newLateRecordCountProvider.getRecordCount(fs, newLateFilePaths);
+      eventMetadataMap.put("newLateEventCounts", Long.toString(newLateEventCount));
```
chavdar: please use named constants for the fixed string literals.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +418,97 @@ public Status status() {
...
+    })) {
+      paths.add(fileStatus.getPath());
+    }
+    return paths;
+  }
+
+  /**
+   * Submit an event reporting new and cumulative late event counts.
+   */
+  private void submitLateEvent(List<Path> newLateFilePaths, Path lateDataDir) {
```
chavdar: Can your rename to something like submitLateEventsTrackingEvent or similar? The current name is a bit misleading.

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
zliu41: Components are simply concatenated but different filenameformats have different requirements, eg if a filenameformat assumes 2nd component is row count, it should check whether there are at least 2 components and the 2nd component is number.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +418,97 @@ public Status status() {
...
+    try {
+      HdfsRecordCountProvider newLateRecordCountProvider =
```
chavdar: This and the next variables should be member variables which get initialized in the constructor.

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
ydai1124: I think it is hard to fix what kind components are concatenated, since we don't know the use cases, e.g., the fileNameFormat to use for AvroHdfsTimePartitionedWithRecordCountsWriter, and for late events.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ydai1124: Fixed.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ydai1124: Fixed.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ydai1124: Changed to `submitLateRecordCountsEvent`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ydai1124: Fixed.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
ydai1124: Fixed.

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
chavdar: In that case, I don't see the purpose of this interface. I don't see what is the abstraction represented by this interface. If the implementation does not now what are the components, the only thing it can change is the separator. Why is that important?

ydai1124:
@chavdar @zliu41 Updated.

gobblin-compaction/src/main/java/gobblin/compaction/count/RecordCountFromFileName.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class RecordCountFromFileName implements FsRecordCountProvider {
+
+  public static final String FILE_NAME_WITH_RECORD_COUNT_FORMAT_CLASS = "filename.with.record.count.format.class";
+  public static final String DEFAULT_FILE_NAME_WITH_RECORD_COUNT_FORMAT_CLASS =
+      AvroHdfsTimePartitionedWithRecordCountsWriter.FileNameFormat.class.getName();
+
+  @Getter
+  @Setter
+  private FileNameWithRecordCountFormat fileNameFormat;
```
chavdar: this should be final

gobblin-api/src/main/java/gobblin/util/FileNameFormat.java
chavdar: In the code, I never see this method accessed in a generic way, i.e. through the interface. It is always through a concrete implementation. It think you should drop this interface altogether.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -117,6 +134,8 @@
   protected final boolean deduplicate;
   protected final double priority;
   protected final EventSubmitter eventSubmitter;
+  private final FsRecordCountProvider newLateRecordCountProvider;
```
chavdar: I don't think we need to instances of the record provided.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -397,4 +441,103 @@ public Status status() {
...
+    }
+    return paths;
+  }
+
+  /**
+   * Submit an event reporting new and cumulative late record counts.
+   */
+  private void submitLateRecordCountsEvent(List<Path> newLateFilePaths, Path lateDataDir) {
+    try {
+      if (newLateRecordCountProvider instanceof RecordCountFromFileName) {
```
chavdar: There is some breakage of abstraction here if your have a generic interface like FsRecordCountProvider but then you have to check for specific implementations like RecordCountFromFileName).

ydai1124:
@chavdar @zliu41 Updated.

ydai1124:
@zliu41 Updated.

zliu41:
LGTM

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -129,34 +144,53 @@ protected MRCompactorJobRunner(Dataset dataset, FileSystem fs, Double priority)
...
+          }
+        }
+        Path lateDataOutputPath = this.outputPath;
+        Path lateDataDir = new Path(this.jobProps.getProp(MRCompactor.COMPACTION_JOB_DEST_LATE_DIR));
+        if (this.deduplicate) {
+          // Update the late data output path.
+          lateDataOutputPath = lateDataDir;
+          LOG.info(String.format("Will copy to late path %s since deduplication has been done. ", lateDataOutputPath));
+          if (!fs.exists(lateDataOutputPath)) {
+            fs.mkdirs(lateDataOutputPath);
```
chavdar: nit: Please check the return value of mkdirs.

chavdar:
One minor comment. Other than that, LGTM.

ydai1124:
@chavdar Updated.

PR #374, opened 2015-10-07
Set partition folder time as upstreamTime for time based compaction

pcadabam:
@zliu41 can you review this?
As discussed I have set the upstreamTime to midnight of partition date. This will used as date of arrival of data for compaction to compute SLA.


zliu41:
LGTM

PR #375, opened 2015-10-08
Added log aggregation to Gobblin on Yarn

liyinan926:
This PR adds log aggregation to Gobblin on Yarn. Implementation notes as follows:
1. There's a utility class `LogCopier` that is an `AbstractScheduledService` that periodically scans for new log files in a given source log directory and schedules a new copy task for each new source log file. Each source log file is copied to a destination directory on the destination file system.
2. The `LogCopier` remembers the current position in the input file stream of each log file and will seek to that position whenever the task starts to copy.
3. Both `GobblinApplicationMaster` and `GobblinWorkUnitRunner` now extend `GobblinYarnLogSource`, which handles to details of building a `LogCopier` to copy the container logs to HDFS.
4. `GobblinYarnAppLauncher` is a sink for logs that copies log files from HDFS to a local directory.

This PR also gets rid of the config key `gobblin.yarn.token.file.path` to favor a path under the app's working directory so we have one less config key to worry about.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
+++ b/gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
@@ -1,0 +1,446 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.logs;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.google.common.io.Closer;
+import com.google.common.io.Files;
+import com.google.common.util.concurrent.AbstractScheduledService;
+
+import gobblin.util.ExecutorsUtils;
+import gobblin.util.FileListUtils;
+
+
+/**
+ * A utility class that periodically reads log files in a source log file directory for changes
+ * since the last reads and appends the changes to destination log files with the same names as
+ * the source log files in a destination log directory. The source and destination log files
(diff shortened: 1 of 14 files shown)


liyinan926:
@zliu41 can you review?

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -0,0 +1,446 @@
...
+      this.srcLogFiles.addAll(newLogFiles);
+    }
+
+    for (final Path srcLogFile : newLogFiles) {
+      String destLogFileName = this.logFileNamePrefix.isPresent() ?
+          this.logFileNamePrefix.get() + "." + srcLogFile.getName() : srcLogFile.getName();
+      final Path destLogFile = new Path(this.destLogDir, destLogFileName);
+
+      // Submit one copy task per new source log file
+      this.logCopyExecutor.scheduleAtFixedRate(new Runnable() {
```
zliu41: if `srcLogFile` is deleted, or the job that generates `srcLogFile` has finished, is there a way to remove the Runnable from schedule, or is it scheduled forever?

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -0,0 +1,446 @@
...
+
+  private final int linesWrittenBeforeFlush;
+
+  private final ScheduledExecutorService logCopyExecutor;
+
+  // A set of current source log files being copied
+  private final Set<Path> srcLogFiles = Sets.newHashSet();
+
+  // A map from source log file paths to the most recent positions the copier have copied to
+  private final Map<Path, Long> currentPosMap = Maps.newHashMap();
```
zliu41: should this be a concurrent map?

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -0,0 +1,446 @@
...
+  private final Path srcLogDir;
+  private final Path destLogDir;
+
+  private final long sourceLogFileMonitorInterval;
+  private final long copyInterval;
+  private final TimeUnit timeUnit;
+
+  private final Set<String> logFileExtensions;
+
+  private final Optional<Pattern> filterRegexPattern;
```
zliu41: It is not clear whether it keeps this pattern or discards this pattern. Both options may be useful, so how about having a `whitelistPattern` and a `blacklistPattern`?
Also is it useful to support multiple patterns instead of a single pattern? If you have multiple patterns you may use `DatasetFilterUtils.stringInPatterns()`.

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
```diff
@@ -162,10 +161,9 @@ public static void addFileAsLocalResource(FileSystem fs, Path destFilePath, Loca
    * Get environment variables in a {@link java.util.Map} used when launching a Yarn container.
    *
    * @param yarnConfiguration a Hadoop {@link Configuration} object carrying Hadoop/Yarn configuration properties
-   * @param config a {@link Config} object carrying Gobblin Yarn application configuration properties
    * @return a {@link java.util.Map} storing environment variables used when launching a Yarn container
    */
-  public static Map<String, String> getEnvironmentVariables(Configuration yarnConfiguration, Config config) {
+  public static Map<String, String> getEnvironmentVariables(Configuration yarnConfiguration) {
```
zliu41: This change is causing build failure.

liyinan926: Fixed.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
liyinan926: Actually there's some `check-then-act` compound operations that cannot be made atomic even if using `ConcurrentMap`. Given that synchronization is needed anyway for such compound operations, using `ConcurrentMap` doesn't offer any benefit.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
liyinan926: Yes, good points. Will do.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
liyinan926: Good point. Will fix this.

liyinan926:
@zliu41 updated 5ca5211.

zliu41:
LGTM

PR #376, opened 2015-10-09
Add loop parallelization utility to ExecutorsUtils

pcadabam:
We find loop parallelizing logic in several parts of the code base hence trying to make a standard way of doing it.
Added a utility method to parallelize loops. Applies a {@link Function} to every element in a {@link List} in parallel by spawning threads. A list containing the result obtained by applying the function is returned. The method is a blocking call and will wait for all the elements in the list to be processed or timeoutInSecs which ever is earlier.

@chavdar can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
@@ -12,14 +12,23 @@
 
 package gobblin.util;
 
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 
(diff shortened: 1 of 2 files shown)


gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -146,4 +154,68 @@ public static void shutdownExecutorService(ExecutorService executorService, Opti
...
+   * @param threadCount to be used to process the list
+   * @param timeoutInSecs to wait for all the threads to complete
+   * @param logger an {@link Optional} wrapping a {@link Logger} to be used during shutdown
+   *
+   * @return a list containing the result obtained by applying the function on each element of the input list in the
+   *         same order
+   *
+   * @throws IllegalArgumentException if input list or function is null
+   */
+  public static <T, V> List<T> parallelize(List<V> list, final Function<V, T> function, int threadCount,
```
chavdar: 1) I think inputOutput is unnecessary. Your task logic can be along the lines of result.set(i, function(list.get(i)) if result is pre-allocated.

2) You need to handle errors in function() or timeouts. You need to use submit() instead of execute() and store all Futures. Once all tasks are done, you need to check if all tasks completed successfully. If any of them failed or timed out, parallelize() needs to through an exception.

gobblin-utility/src/test/java/gobblin/util/ExecutorsUtilsTest.java
```diff
@@ -47,4 +52,22 @@ public void run() {
 
     thread.run();
   }
+
+  @Test
+  public void testParallelize() {
+    List<Integer> nums = Lists.newArrayList();
```
chavdar: I'd add more elements than threads.

gobblin-utility/src/test/java/gobblin/util/ExecutorsUtilsTest.java
```diff
@@ -47,4 +52,22 @@ public void run() {
...
+      @Override
+      public String apply(Integer input) {
+        return Integer.toString(input * factor);
+      }
+    };
+
+    List<String> result = ExecutorsUtils.parallelize(nums, multiply , 2, 60, Optional.<Logger>absent());
+    Assert.assertEquals(Arrays.asList("15","50"), result);
+  }
 }
```
chavdar: We need tests for the following cases:
* A function call throws an Exception
* A function call times out

sahilTakiar:
@chavdar this may be a separate discussion, but are we ready to migrate to Java 8? Then we get this for free https://docs.oracle.com/javase/tutorial/collections/streams/parallelism.html#executing_streams_in_parallel

ibuenros:
I don't think we should migrate to java 8 yet. As of spring 2015 (latest
stats I could find in a quick search), java 8 adoption was at 20% according
to https://plumbr.eu/blog/java/java-version-statistics-2015-edition
Given that gobblin is open source, I think we should wait until there is
further adoption of java 8 before migrating.

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
pcadabam: Done

gobblin-utility/src/test/java/gobblin/util/ExecutorsUtilsTest.java
pcadabam: Done

gobblin-utility/src/test/java/gobblin/util/ExecutorsUtilsTest.java
pcadabam: Done

pcadabam:
@chavdar addressed all the comments. Please take a look.
@sahilTakiar Yes later versions of java have similar kind of utils. Some of the code used to wait for future tasks is made easy in java7 http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorCompletionService.html
But I think we will need this until we migrate to java8.

gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
```diff
@@ -146,4 +155,73 @@ public static void shutdownExecutorService(ExecutorService executorService, Opti
...
+   *           <li>if any thread was interrupted while waiting
+   *           <ul>
+   */
+  public static <T, F> List<T> parallelize(final List<F> list, final Function<F, T> function, int threadCount,
+      int timeoutInSecs, Optional<Logger> logger) throws ExecutionException {
+
+    Preconditions.checkArgument(list != null, "Input list can not be null");
+    Preconditions.checkArgument(function != null, "Function can not be null");
+
+    final List<T> results = Lists.newArrayList();
```
chavdar: nit: This is a small optimization. I'd probably use "new ArrayList<T>(list.size())". Thus we can avoid the dynamic growing of the list as we already now the final size.

chavdar:
One small comment. Overall, LGTM.

PR #380, opened 2015-10-12
Fix datasetbase when cleaning single version dataset

ydai1124:
@ibuenros  Can you review?


ibuenros:
Actually, I was working in a slightly safer fix for this. Let's talk before committing anything.

PR #383, opened 2015-10-13
Writer Partitioner interfaces implemented.

ibuenros:


ibuenros:
@chavdar @liyinan926 I haven't tested these changes yet, but putting it here so you can look at the design.

ibuenros:
@chavdar @liyinan926 changes are tested, can you review?

chavdar:
It's good to have a wiki describing the high-level design. It will be also useful for people wanting to provide their own implementations.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+/**
+ * {@link DataWriter} that partitions data using a partitioner, instantiates appropriate writers, and sends records to
+ * the chosen writer.
+ */
```
chavdar: Please document the S and D parameters.

ibuenros:
@chavdar absolutely, just want to have agreement on the general design before writing that.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+        this.builder = Optional.of(PartitionAwareDataWriterBuilder.class.cast(builder));
+        this.partitioner = Optional.of(
+            WriterPartitioner.class.cast(ConstructorUtils.invokeConstructor(
+                    Class.forName(state.getProp(ConfigurationKeys.WRITER_PARTITIONER_CLASS)), state)));
+        this.builder.get().validatePartitionSchema(this.partitioner.get().partitionSchema());
+      } catch(ReflectiveOperationException roe) {
+        throw new IOException(roe);
+      }
+    } else {
+      this.shouldPartition = false;
```
chavdar: Why do we need to support this case? If people don't want partitioning that shouldn't use PartitionedDataWriter.

ibuenros: In this design, Gobblin always uses `PartitionedDataWriter`. Users still specify their data writers as usual, and Gobblin transparently handles whether it should partition or not (depending on whether there is a partitioner class or not).

chavdar:
I think wiki is useful for the reviewers too (e.g. the decision to always use PartitionedDataWriter)

ibuenros:
@chavdar see wiki at https://github.com/linkedin/gobblin/wiki/Partitioned-Writers

gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedPartitionedDataWriterDecorator.java
```diff
@@ -0,0 +1,46 @@
...
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.instrumented.writer;
+
+import gobblin.configuration.State;
```
liyinan926: The order of imports is incorrect.

gobblin-core/src/main/java/gobblin/writer/AvroDataWriterBuilder.java
```diff
@@ -49,6 +53,10 @@
         String fileName = WriterUtils
             .getWriterFileName(properties, this.branches, this.branch, this.writerId, this.format.getExtension());
 
+        if(this.partition != null) {
+          fileName = new Path(AvroUtils.serializeAsPath(this.partition, true).toString(),fileName).toString();
```
liyinan926: Please add an empty space in `toString(),fileName`.

gobblin-core/src/main/java/gobblin/writer/AvroDataWriterBuilder.java
```diff
@@ -49,6 +53,10 @@
         String fileName = WriterUtils
             .getWriterFileName(properties, this.branches, this.branch, this.writerId, this.format.getExtension());
 
+        if(this.partition != null) {
```
liyinan926: Can we make `partition` an `Optional`?

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+      DataWriter<D> writer = this.partitionWriters.get(partition);
+      writer.write(record);
+    } catch(ExecutionException ee) {
+      throw new IOException(ee);
+    }
+  }
+
+  @Override public void commit() throws IOException {
+    for(Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+      entry.getValue().commit();
```
liyinan926: I suggest putting `commit` into a `try-catch` block so the commit failure of one partition will not cause other partitions to not be committed.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  @Override public void cleanup() throws IOException {
+    for(Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+      entry.getValue().cleanup();
```
liyinan926: Ditto, better to put `cleanup` in a `try-catch` block.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+  @Override public long bytesWritten() throws IOException {
+    long totalBytes = 0;
+    for(Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+      totalBytes += entry.getValue().bytesWritten();
+    }
+    return totalBytes;
+  }
+
+  @Override public void close() throws IOException {
+    for(Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
```
liyinan926: Please use a `Closer` here.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+      entry.getValue().close();
+    }
+  }
+
+  private DataWriter<D> createPartitionWriter(GenericRecord partition) throws IOException {
+    if(!this.builder.isPresent()) {
+      throw new IOException("Writer builder not found. This is an error in the code.");
```
liyinan926: Can you also include the `partition` in the error message?

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+ * Launcher for creating a Gobblin job from command line or IDE.
+ */
+public class CommandLineLauncher {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CommandLineLauncher.class);
+
+  public static void main(String[] args) {
+
+    if(args.length < 1) {
+      throw new RuntimeException("Missing argument. Expected path to configuration file.");
```
liyinan926: I think it's better to print a usage message followed by a `System.exit(1)`.

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+
+    try {
+      // Load framework configuration properties
+      PropertiesConfiguration configT = new PropertiesConfiguration();
+      configT.load(args[0]);
+      Configuration config = configT.interpolatedConfiguration();
+
+      Properties properties = ConfigurationConverter.getProperties(config);
+
+      LOG.info(properties.toString());
```
liyinan926: Should this be a `debug` level log?

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+
+      for (Properties jobProps : SchedulerUtils.loadJobConfigs(properties)) {
+        LOG.info("Running job.");
```
liyinan926: Better to also include the name of the job in the log.

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+        new LocalJobLauncher(jobProps).launchJob(null);
+      }
+    } catch (Exception ioe) {
+      throw new RuntimeException(ioe);
+    } finally {
```
liyinan926: Empty finally?

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -417,4 +418,30 @@ public static Schema switchName(Schema schema, String newName) {
...
+   *                          will be included.
+   * @return A relative path where each level is a field in the input record.
+   */
+  public static Path serializeAsPath(GenericRecord record, boolean includeFieldNames) {
+    if(record == null) {
+      return new Path("");
+    }
+    List<String> tokens = Lists.newArrayList();
+    for(Schema.Field field : record.getSchema().getFields()) {
+      String sanitizedName = field.name().replaceAll("[\\s/]","_");
```
liyinan926: You can define `"[\\s/]"` as a constant. Please also put an empty space before `"_"`.

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+
+      Properties properties = ConfigurationConverter.getProperties(config);
+
+      LOG.info(properties.toString());
+
+      for (Properties jobProps : SchedulerUtils.loadJobConfigs(properties)) {
+        LOG.info("Running job.");
+        new LocalJobLauncher(jobProps).launchJob(null);
+      }
+    } catch (Exception ioe) {
```
liyinan926: Why named `ioe`?

gobblin-core/src/main/java/gobblin/writer/PartitionAwareDataWriterBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package gobblin.writer;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+
+/**
+ * A {@link DataWriterBuilder} used with a partitioner. When provided with a partitioner, Gobblin will create a
```
liyinan926: `partitioner` to `{@link WriterPartitioner}`.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+          String.format("%s was specified but the writer %s does not support partitioning.",
+              ConfigurationKeys.WRITER_PARTITIONER_CLASS, builder.getClass().getCanonicalName()));
+
+      try {
+        this.shouldPartition = true;
+        this.builder = Optional.of(PartitionAwareDataWriterBuilder.class.cast(builder));
+        this.partitioner = Optional.of(
+            WriterPartitioner.class.cast(ConstructorUtils.invokeConstructor(
+                    Class.forName(state.getProp(ConfigurationKeys.WRITER_PARTITIONER_CLASS)), state)));
+        this.builder.get().validatePartitionSchema(this.partitioner.get().partitionSchema());
```
liyinan926: Should check the return value of `validatePartitionSchema`.

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.commons.configuration.ConfigurationConverter;
+import org.apache.commons.configuration.PropertiesConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Launcher for creating a Gobblin job from command line or IDE.
+ */
+public class CommandLineLauncher {
```
liyinan926: Following `CliMRJobLauncher`, can you rename this to `CliLocalJobLauncher` and move to it `gobblin.runtime.local`?

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.apache.commons.configuration.Configuration;
+import org.apache.commons.configuration.ConfigurationConverter;
+import org.apache.commons.configuration.PropertiesConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Launcher for creating a Gobblin job from command line or IDE.
```
liyinan926: Is this class supposed to run a single job or any jobs under a given job config dir? The code seems doing the latter. I would suggest doing the former and leaving the job of handling multiple jobs to the scheduler.

gobblin-core/src/main/java/gobblin/writer/WriterPartitioner.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package gobblin.writer;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+
+/**
+ * Partitions records in the writer phase.
```
liyinan926: Please make it clear in the Javadoc that implementation classes are required to have a constructor that takes in a `State` object.

gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedPartitionedDataWriterDecorator.java
```diff
@@ -0,0 +1,46 @@
...
+  public static final String PARTITION = "Partition";
+  private final GenericRecord partition;
+
+  public InstrumentedPartitionedDataWriterDecorator(DataWriter<D> writer, State state, GenericRecord partition) {
+    super(writer, state);
+    this.partition = partition;
+  }
+
+  @Override public List<Tag<?>> generateTags(State state) {
+    List<Tag<?>> tags = super.generateTags(state);
```
liyinan926: You probably want to make a copy here in case `super.generateTags` returns an `ImmutableList`. And you probably want to return an `ImmutableList` here too since semantically this list should be immutable once created.

gobblin-core/src/main/java/gobblin/writer/AvroDataWriterBuilder.java
```diff
@@ -32,7 +32,11 @@
  * @author ynli
  */
 @SuppressWarnings("unused")
-public class AvroDataWriterBuilder extends DataWriterBuilder<Schema, GenericRecord> {
+public class AvroDataWriterBuilder extends PartitionAwareDataWriterBuilder<Schema, GenericRecord> {
+
+  @Override public boolean validatePartitionSchema(Schema partitionSchema) {
```
liyinan926: You can make this a default implementation in `PartitionAwareDataWriterBuilder`?

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -0,0 +1,132 @@
...
+      this.partitionWriters.put(NON_PARTITIONED_WRITER_KEY,
+          new InstrumentedDataWriterDecorator<D>(builder.build(), state));
+      this.partitioner = Optional.absent();
+      this.builder = Optional.absent();
+    }
+  }
+
+  @Override public void write(D record) throws IOException {
+    try {
+      GenericRecord partition = this.shouldPartition ? this.partitioner.get().partitionForRecord(record) :
```
liyinan926: This is concerning me since a `GenericRecord` is (possibly) created and returned for each data record, which may not be trivial in terms of overhead. @chavdar what's your opinion on this?

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
ibuenros: The partition is irrelevant. The only way this condition can be reached is if the PartitionedWriter was constructed incorrectly, and in that case this will happen in the first partition, regardless of the specific value of the partition.

gobblin-core/src/main/java/gobblin/writer/AvroDataWriterBuilder.java
ibuenros: Is this a good idea? I don't think that most partition aware writers will be able to handle all schemas, so I prefer to require implementers to explicitly think about what their writer will support.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
ibuenros: Partitioners can always cache the partition to avoid the overhead. I understand the concern, but it is important for partitions to have a schema, and if we don't use Avro  objects, we would have to define a partition schema API.

ibuenros:
@liyinan926 addressed most comments, commented in a few.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
liyinan926: @chavdar what's your opinion on this?

ibuenros:
@liyinan926 Addressed comments

liyinan926:
LGTM.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -67,6 +68,7 @@
   private static final String FIELD_LOCATION_DELIMITER = ".";
 
   private static final String AVRO_SUFFIX = ".avro";
+  public static final String SPACE_SLASH_REGEX = "[\\s/]";
```
chavdar: HDFS does not allow ':' in file names. I'd also put '\' in the blacklisted characters.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaPartitioner.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+/**
+ * Partitioner that splits records by title.
+ */
+public class WikipediaPartitioner implements WriterPartitioner<GenericRecord> {
+
+  private static final String TITLE = "title";
+
+  private static final Schema SCHEMA = SchemaBuilder.record("ArticleTitle").fields().name(TITLE).
```
chavdar: It's always a good idea to add a namespace to names.

gobblin-core/src/test/java/gobblin/writer/test/TestPartitioner.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.avro.SchemaBuilder;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+
+
+public class TestPartitioner implements WriterPartitioner<String> {
+
+  public static final String PARTITION = "partition";
+
+  private static final Schema SCHEMA = SchemaBuilder.record("ArticleTitle").fields().name(PARTITION).
```
chavdar: nit: s/ArticleTitle/LetterPartition/

chavdar:
Overall, it looks good. I have some minor comments. Feel free to commit once those are addressed.

Some comments about the wiki:
* Please add a picture and/or example to demonstrate how all the classes interact together.
* Please add a section to include available partitioners (which I assume will come in subsequent commits).

PR #384, opened 2015-10-13
Make sure properties set in a pre-fork converter is not lost

liyinan926:
Currently in `TaskContext.getConverters(int index)`, a local copy of the `taskState` is created and used to initialize the converters. Because of this, any user-set properties in the converters are not available in the original task-level `TaskState` and therefore are lost. This PR fixed this. 

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
@@ -76,6 +76,8 @@ public class Fork implements Closeable, Runnable, FinalState {
 
   private final TaskContext taskContext;
   private final TaskState taskState;
+  // A TaskState instance specific to this Fork
+  private final TaskState forkTaskState;
   private final String taskId;
 
   private final int branches;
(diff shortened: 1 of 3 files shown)


liyinan926:
@sahilTakiar can you review?

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -407,31 +411,29 @@ private void processRecords() throws IOException, DataConversionException {
...
-    TaskState taskStateForFork = this.taskState;
-    if (this.branches > 1) {
-      // Make a copy if there are more than one fork
-      taskStateForFork = new TaskState(this.taskState);
-    }
+  private boolean checkDataQuality(Optional<Object> schema) throws Exception {
+    this.forkTaskState.setProp(ConfigurationKeys.EXTRACTOR_ROWS_EXPECTED,
+        this.taskState.getProp(ConfigurationKeys.EXTRACTOR_ROWS_EXPECTED));
+    this.forkTaskState.setProp(ConfigurationKeys.EXTRACTOR_ROWS_EXTRACTED,
+        this.taskState.getProp(ConfigurationKeys.EXTRACTOR_ROWS_EXTRACTED));
```
sahilTakiar: Shouldn't this only necessary if `branches > 1`

sahilTakiar:
LGTM besides the minor comment.

It seems as if a converter can only set properties in a task state if there is only 1 branch in the fork operator. Otherwise each branch gets its own copy of the task state, and properties set on the copies will not be persistent. Correct? We may need to rethink a way of getting metadata out of converters. It seems weird to say that properties set in a converter are persistent to the original task state only if there is 1 branch in the fork operator.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
liyinan926: Yes, fixed.

liyinan926:
@sahilTakiar Actually properties set by the `converters` in the pre-fork stream are persisted into the task-level `taskState`. Properties set by `converters` in any forked branches, however, are only persisted into the fork-specific `forkTaskState`s. This is still quite confusing and error-prone. I agree that we should rethink how we should organize metadata in a task and its forks.

PR #385, opened 2015-10-14
Late data handling for compaction

zliu41:
Main changes:
- Moved all folders (e.g., `inputPath`, `inputLatePath`, etc.) as fields of `Dataset` to avoid passing all of them around in method signatures.
- Moved the logic for finding late data files from `MRCompactorTimeBasedJobPropCreator` to `MRCompactorJobPropCreator` since they are also applicable to non time-based compaction.
- Added support for handling input late paths (since hourly compaction may create `hourly_late`, which will be the input late path for daily compaction) for both dedup and non-dedup cases. If input late dir has data and the dataset has not been compacted before, the dataset will always be deduped.
- For hourly compaction, the output file format in `hourly` folder is compaction format (start with `part-r`), while the file format in `hourly_late` folder is writer format. Both folders may contain late data which will be copied to `daily_late`, causing files in `daily_late` to have two different formats. The following has been done to resolve this inconsistency:
 - split `compaction.record.count.provider` into `compaction.input.record.count.provider` and `compaction.output.record.count.provider`.
 - change `RecordCountProvider` from interface to abstract class, and added new method `convertPath()`.
 - when a late data file is copied to output late dir, its file name is converted according to `compaction.output.record.count.provider`.
- Put all `RecordCountProvider` implementations in `gobblin-utility` to avoid circular dependency.
- Added option `compaction.recompact.from.dest.paths` for manual recompaction from output paths. When this is enabled, a dataset will be recompacted if there are late data in its `output_late` folder (e.g., `daily_late` for daily compaction). Its input folders (e.g., `hourly` and `hourly_late`) will be ignored. Also, in this case the compaction timestamp is not the compaction start time, but the latest modification time of all input files. The reason we can't use compaction job start time in this case is that between this two timestamps (i.e., latest modification time of all input files and compaction job start time), new data may be written into the input paths. These new data should be considered late data the next time we compact from the input paths.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
+++ b/gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
@@ -12,11 +12,34 @@
 
 package gobblin.util;
 
+import org.apache.commons.lang.NotImplementedException;
 import org.apache.hadoop.fs.Path;
 
+
 /**
- * Get record count from a given {@link Path}.
+ * A class for obtaining record counts from data.
  */
-public interface RecordCountProvider {
-  public long getRecordCount(Path path);
+public abstract class RecordCountProvider {
+
+  /**
+   * Get record count from a given {@link Path}.
+   */
+  public abstract long getRecordCount(Path path);
+
+  /**
+   * Convert a {@link Path} from another {@link RecordCountProvider} so that it can be used
+   * in {@link #getRecordCount(Path)} of this {@link RecordCountProvider}.
+   */
+  public Path convertPath(Path path, RecordCountProvider src) {
+    if (this.getClass().equals(src.getClass())) {
+      return path;
+    } else {
+      throw getNotImplementedException(src);
+    }
+  }
+
+  protected NotImplementedException getNotImplementedException(RecordCountProvider src) {
+    return new NotImplementedException(String.format("converting from %s to %s is not implemented",
+        src.getClass().getName(), this.getClass().getName()));
+  }
 }
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
@@ -14,17 +14,23 @@
 package gobblin.compaction.mapreduce;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
+import gobblin.compaction.Dataset;
 import gobblin.configuration.State;
+import gobblin.util.FileListUtils;
 
 
 /**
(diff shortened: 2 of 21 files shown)


zliu41:
@ydai1124 could you review?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -63,11 +64,17 @@ public static Configuration newConfiguration() {
     return results;
   }
 
+  public static List<FileStatus> listStatusRecursive(FileSystem fileSystem, Iterable<Path> paths) throws IOException {
```
ydai1124: `ListStatusRecursive` in `HadoopUtils` is overlapped with `listFilesRecursively` in `FileListUtils`. Can you take a look at `FileListUtils` and add this method there?

gobblin-utility/src/main/java/gobblin/util/recordcount/IngestionRecordCountProvider.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * This method currently supports converting the given {@link Path} to {@link CompactionRecordCountProvider}.
+   */
+  @Override
+  public Path convertPath(Path path, RecordCountProvider target) {
+    if (this.getClass().equals(target.getClass())) {
+      return path;
+    } else if (target.getClass().equals(CompactionRecordCountProvider.class)) {
+      String newFileName = new CompactionRecordCountProvider()
+          .constructFileName(CompactionRecordCountProvider.M_OUTPUT_FILE_PREFIX, this.getRecordCount(path));
```
ydai1124: Can you add some doc here explaining using `M_OUTPUT_FILE_PREFIX` as the default prefix?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
```diff
@@ -95,59 +109,169 @@ MRCompactorJobPropCreator build() {
...
-    List<State> emptyJobProps = Lists.newArrayList();
+  protected List<Dataset> createJobProps() throws IOException {
     if (!this.fs.exists(this.topicInputDir)) {
       LOG.warn("Input folder " + this.topicInputDir + " does not exist. Skipping topic " + this.topic);
-      return emptyJobProps;
+      return ImmutableList.<Dataset> of();
+    }
+
+    Dataset dataset = new Dataset.Builder().withTopic(this.topic).withPriority(this.priority)
+        .withInputPath(this.recompactFromOutputPaths ? this.topicOutputDir : this.topicInputDir)
```
ydai1124: With this setting, on-demand recompaction will only work for `topicOutputDir`, but not for `topicInputDir`. What about the late files in `topicInputDir`, i.e., hourly_late? When will they be recompacted?

gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
```diff
@@ -12,11 +12,34 @@
...
+  /**
+   * Get record count from a given {@link Path}.
+   */
+  public abstract long getRecordCount(Path path);
+
+  /**
+   * Convert a {@link Path} so that it can be used in {@link #getRecordCount(Path)} of another
+   * {@link RecordCountProvider}.
+   */
+  public Path convertPath(Path path, RecordCountProvider target) {
```
ydai1124: nit: Does it make more sense if changing this to convertPath(path, src)? So the output can be directly used in getRecordCount.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
zliu41: `ListStatusRecursive` has been in `HadoopUtils` for a long time. I can deprecate it.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
zliu41: There are two types of compactions: a regular compaction that runs twice a day and consumes data in inputdir, and an ad-hoc compaction that consumes data in outputdir. This is for ad-hoc, i.e. if there are enough data in `daily_late` we may recompact daily data.

zliu41:
@ydai1124 updated.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
ydai1124: Makes sense.

ydai1124:
LGTM.

PR #387, opened 2015-10-15
Fix record count in compaction sla event

pcadabam:
Fixed record count being 0 for a few topics. The compaction jobs for these topics were mapper only. The code previously only looked at reducer counters.
@zliu41 can you take a look at this?


zliu41:
LGTM

PR #391, opened 2015-10-16
Separate two types of non-deduping compaction

zliu41:
There are two types of non-deduping compaction: (1) the data doesn't need to be deduplicated, e.g., /data/service; (2) the data do need to be deduplicated, but the input data is already compacted, e.g., daily compaction for /data/tracking.

These two types need to be treated different when there's late data. For type (1), we simply copy late data to output folder; for type (2), we need to copy late data to _late folder. Also, for type (2), when we compact a folder for the first time, if the input folder contains late data (e.g., in `hourly_late`), then it will run a deduping compaction instead of non-deduping compaction.

Separated property `compaction.deduplicate` into `compaction.input.deduplicated` and `compaction.output.deduplicated`. In case (1), both are false. In case (2), both are true.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
@@ -44,7 +44,7 @@ public class CompactionSlaEventHelper {
   public static void populateState(Dataset dataset, Optional<Job> job, FileSystem fs) {
     setDatasetUrn(dataset);
     setPartition(dataset);
-    setDedupeStatus(dataset.jobProps());
+    setOutputDedupeStatus(dataset.jobProps());
     setPreviousPublishTime(dataset, fs);
     if (job.isPresent()) {
       setRecordCount(dataset.jobProps(), job.get());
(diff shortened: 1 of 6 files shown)


zliu41:
@ydai1124 can you review?

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -79,7 +79,8 @@ private static void setPreviousPublishTime(Dataset dataset, FileSystem fs) {
   }
 
   private static void setDedupeStatus(State state) {
```
ydai1124: rename to `setOutputDedupeStatus`?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
```diff
@@ -197,10 +202,19 @@ protected MRCompactorJobPropCreator(Builder<?> builder) {
     return Optional.of(dataset);
   }
 
+  private Object shouldDeduplicate() {
```
ydai1124: You may not need this method, since it is only one line and used only once. Also, it didn't cover the case in `addInputLateFilesForFirstTimeCompaction`.

zliu41:
Updated.

PR #392, opened 2015-10-19
Do not throw Exception in ParallelRunner.rename if dst exists

zliu41:
If a mapper publishes data for some tasks and then failed, and a new mapper is launched for the same tasks, the new mapper will still fail because it will fail to publish data for those tasks since they already exist in the dest folder.


zliu41:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #393, opened 2015-10-19
Replace postProcessTaskStates() with postProcessJobState()

zliu41:
Some metrics, e.g., `FAIL_TO_GET_OFFSET_COUNT`, are obtained in `Source` and added to `SourceState`, so they will be in `JobState` instead of `TaskState`.


zliu41:
@liyinan926 could you review?

liyinan926:
LGTM.

PR #396, opened 2015-10-21
Turn on mapreduce output compression by default

zliu41:
@chavdar please review


gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -133,6 +133,12 @@ public MRJobLauncher(Properties jobProps, Configuration conf) throws Exception {
     // Put job configuration properties into the Hadoop configuration so they are available in the mappers
     JobConfigurationUtils.putPropertiesIntoConfiguration(this.jobProps, this.conf);
 
+    // Turn on mapreduce output compression by default
```
liyinan926: The MR job does not use a `FileOutputFormat` to write data, so this has no effect. You don't need to do this here.

zliu41:
Updated.

liyinan926:
LGTM.

PR #398, opened 2015-10-21
Do not add JobState to WorkUnits until when the tasks start

liyinan926:
Currently in the MR mode, there's an optimization that avoids adding the JobState to the WorkUnits too early to reduce the memory footprint on the client side as well as reduce the cost of serializing the WorkUnits. The JobState is added to the WorkUnits, though, in the mappers after the WorkUnits are deserialized. Currently the Yarn mode is not doing this optimization so this PR fixes it.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/SerializationUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/SerializationUtils.java
@@ -14,13 +14,22 @@ package gobblin.util;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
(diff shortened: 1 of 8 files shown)


liyinan926:
@zliu41 can you review?

gobblin-utility/src/main/java/gobblin/util/SerializationUtils.java
```diff
@@ -102,4 +111,52 @@
...
+      state.write(dataOutputStream);
+    } catch (Throwable t) {
+      throw closer.rethrow(t);
+    } finally {
+      closer.close();
+    }
+  }
+
+  /**
+   * Deserialize/read a {@link State} instance from a file.
```
zliu41: Please clarify if the state passed to `deserializeState` contains properties, whether those properties will be removed.

liyinan926: Line 143 makes it explicit that it expects an `empty` `State` instance.

zliu41:
LGTM

PR #401, opened 2015-10-22
Dataset config

tuGithub:
@chavdar

### DIFF
--- a/gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
+++ b/gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
@@ -300,6 +300,25 @@ public class GobblinMetrics {
   }
 
   /**
+   * Starts metric reporting and appends the given metrics file suffix to the current value of
+   * {@link ConfigurationKeys#METRICS_FILE_SUFFIX}.
+   */
+  public void startMetricReportingWithFileSuffix(State state, String metricsFileSuffix) {
+    Properties metricsReportingProps = new Properties();
+    metricsReportingProps.putAll(state.getProperties());
+
+    String oldMetricsFileSuffix =
+        state.getProp(ConfigurationKeys.METRICS_FILE_SUFFIX, ConfigurationKeys.DEFAULT_METRICS_FILE_SUFFIX);
+    if (Strings.isNullOrEmpty(oldMetricsFileSuffix)) {
+      oldMetricsFileSuffix = metricsFileSuffix;
+    } else {
+      oldMetricsFileSuffix += "." + metricsFileSuffix;
+    }
+    metricsReportingProps.setProperty(ConfigurationKeys.METRICS_FILE_SUFFIX, oldMetricsFileSuffix);
+    startMetricReporting(metricsReportingProps);
+  }
+
+  /**
    * Start metric reporting.
    *
    * @param configuration configuration properties
@@ -328,6 +347,7 @@ public class GobblinMetrics {
 
     buildJmxMetricReporter(properties);
     if (this.jmxReporter.isPresent()) {
+      LOGGER.info("Will start reporting metrics to JMX");
       this.jmxReporter.get().start();
     }
 
@@ -346,7 +366,7 @@ public class GobblinMetrics {
    * Immediately trigger metric reporting.
    */
   public void triggerMetricReporting() {
-    for(ScheduledReporter reporter : this.scheduledReporters) {
+    for (ScheduledReporter reporter : this.scheduledReporters) {
       reporter.report();
     }
   }
@@ -359,6 +379,8 @@ public class GobblinMetrics {
       LOGGER.warn("Metric reporting has not started yet");
       return;
     }
+
+    LOGGER.info("Stopping metric reporting");
 
     if (this.jmxReporter.isPresent()) {
       this.jmxReporter.get().stop();
@@ -418,6 +440,8 @@ public class GobblinMetrics {
           .outputTo(output).build()));
       this.scheduledReporters.add(this.closer.register(OutputStreamEventReporter.forContext(this.metricContext)
           .outputTo(output).build()));
+
+      LOGGER.info("Will start reporting metrics to directory " + metricsLogDir);
     } catch (IOException ioe) {
       LOGGER.error("Failed to build file metric reporter for job " + this.id, ioe);
     }
@@ -470,23 +494,29 @@ public class GobblinMetrics {
       formatEnum = KafkaReportingFormats.JSON;
     }
 
-    if(metricsTopic.or(defaultTopic).isPresent()) {
+    if (metricsTopic.or(defaultTopic).isPresent()) {
       try {
         KafkaReporter.Builder<?> builder = formatEnum.metricReporterBuilder(this.metricContext, properties);
         this.scheduledReporters.add(this.closer.register(builder.build(brokers, metricsTopic.or(defaultTopic).get())));
       } catch (IOException exception) {
         LOGGER.error("Failed to create Kafka metrics reporter. Will not report metrics to Kafka.", exception);
       }
+    } else {
+      LOGGER.warn("Not reporting metrics to Kafka, no topic specified");
     }
 
-    if(eventsTopic.or(defaultTopic).isPresent()) {
+    if (eventsTopic.or(defaultTopic).isPresent()) {
       try {
         KafkaEventReporter.Builder<?> builder = formatEnum.eventReporterBuilder(this.metricContext, properties);
         this.scheduledReporters.add(this.closer.register(builder.build(brokers, eventsTopic.or(defaultTopic).get())));
       } catch (IOException exception) {
         LOGGER.error("Failed to create Kafka events reporter. Will not report events to Kafka.", exception);
       }
+    } else {
+      LOGGER.warn("Not reporting events to Kafka, no topic specified");
     }
+
+    LOGGER.info("Will start reporting metrics to Kafka");
   }
 
   /**
@@ -521,7 +551,5 @@ public class GobblinMetrics {
         LOGGER.warn("Could not create metric reporter from builder " + reporterClass + ".", exception);
       }
     }
-
   }
-
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
@@ -14,7 +14,6 @@ package gobblin.data.management.copy;
 
 import gobblin.data.management.partition.File;
 
-import java.io.IOException;
 import java.util.List;
 
 import lombok.AccessLevel;
@@ -41,15 +40,20 @@ import com.google.gson.reflect.TypeToken;
 @EqualsAndHashCode
 public class CopyableFile implements File {
 
-  private static Gson gson = new Gson();
+  private static final Gson GSON = new Gson();
+
   /** {@link FileStatus} of the existing origin file. */
   private FileStatus origin;
+
   /** Complete destination {@link Path} of the file. Dataset's final publish directory + {@link #relativeDestination} */
   private Path destination;
+
   /** {@link Path} to the file relative to the dataset's final publish directory */
   private Path relativeDestination;
+
   /** Desired {@link OwnerAndPermission} of the destination path. */
   private OwnerAndPermission destinationOwnerAndPermission;
+
   /**
    * Desired {@link OwnerAndPermission} of the ancestor directories of the destination path. The list is ordered from
    * deepest to highest directory.
@@ -65,6 +69,7 @@ public class CopyableFile implements File {
    * </p>
    */
   private List<OwnerAndPermission> ancestorsOwnerAndPermission;
+
   /** Checksum of the origin file. */
   private byte[] checksum;
 
@@ -79,18 +84,18 @@ public class CopyableFile implements File {
    * @param copyableFile to be serialized
    * @return serialized string
    */
-  public static String serialize(CopyableFile copyableFile) throws IOException {
-    return gson.toJson(copyableFile);
+  public static String serialize(CopyableFile copyableFile) {
+    return GSON.toJson(copyableFile);
   }
 
   /**
    * Serialize a {@link List} of {@link CopyableFile}s into a {@link String}.
    *
-   * @param copyableFile to be serialized
+   * @param copyableFiles to be serialized
    * @return serialized string
    */
-  public static String serializeList(List<CopyableFile> copyableFiles) throws IOException {
-    return gson.toJson(copyableFiles, new TypeToken<List<CopyableFile>>(){}.getType());
+  public static String serializeList(List<CopyableFile> copyableFiles) {
+    return GSON.toJson(copyableFiles, new TypeToken<List<CopyableFile>>(){}.getType());
   }
 
   /**
@@ -99,8 +104,8 @@ public class CopyableFile implements File {
    * @param serialized string
    * @return a new instance of {@link CopyableFile}
    */
-  public static CopyableFile deserialize(String serialized) throws IOException {
-    return gson.fromJson(serialized, CopyableFile.class);
+  public static CopyableFile deserialize(String serialized) {
+    return GSON.fromJson(serialized, CopyableFile.class);
   }
 
   /**
@@ -110,7 +115,12 @@ public class CopyableFile implements File {
    * @param serialized string
    * @return a new {@link List} of {@link CopyableFile}s
    */
-  public static List<CopyableFile> deserializeList(String serialized) throws IOException {
-    return gson.fromJson(serialized, new TypeToken<List<CopyableFile>>(){}.getType());
+  public static List<CopyableFile> deserializeList(String serialized) {
+    return GSON.fromJson(serialized, new TypeToken<List<CopyableFile>>(){}.getType());
+  }
+
+  @Override
+  public String toString() {
+    return serialize(this);
   }
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
@@ -29,4 +29,8 @@ public class FileAwareInputStream {
   private CopyableFile file;
   private FSDataInputStream inputStream;
 
+  @Override
+  public String toString() {
+    return this.file.toString();
+  }
 }
--- a/gobblin-metrics/src/main/java/gobblin/metrics/Tagged.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/Tagged.java
@@ -44,8 +44,8 @@ public class Tagged implements Taggable {
 
   @Override
   public void addTag(Tag<?> tag) {
-    Preconditions.checkNotNull(tag);
-    Preconditions.checkNotNull(tag.getValue());
+    Preconditions.checkNotNull(tag, "Cannot add a null Tag");
+    Preconditions.checkNotNull(tag.getValue(), "Cannot add a Tag with a null value. Tag: " + tag);
     this.tags.put(tag.getKey(), tag.getValue());
   }
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/TaskContext.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/TaskContext.java
@@ -17,7 +17,6 @@ import java.util.Collections;
 import java.util.List;
 
 import com.google.common.base.Enums;
-import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
 import com.google.common.collect.Lists;
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -47,7 +47,6 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
@@ -75,6 +74,7 @@ import gobblin.runtime.util.MetricGroup;
 import gobblin.runtime.util.TimingEventNames;
 import gobblin.source.workunit.MultiWorkUnit;
 import gobblin.source.workunit.WorkUnit;
+import gobblin.util.HadoopUtils;
 import gobblin.util.JobConfigurationUtils;
 import gobblin.util.JobLauncherUtils;
 import gobblin.util.ParallelRunner;
@@ -558,17 +558,8 @@ public class MRJobLauncher extends AbstractJobLauncher {
       if (Boolean.valueOf(
           configuration.get(ConfigurationKeys.METRICS_ENABLED_KEY, ConfigurationKeys.DEFAULT_METRICS_ENABLED))) {
         this.jobMetrics = Optional.of(JobMetrics.get(this.jobState));
-        String metricFileSuffix =
-            configuration.get(ConfigurationKeys.METRICS_FILE_SUFFIX, ConfigurationKeys.DEFAULT_METRICS_FILE_SUFFIX);
-        // If running in MR mode, all mappers will try to write metrics to the same file, which will fail.
-        // Instead, append the taskAttemptId to each file name.
-        if (Strings.isNullOrEmpty(metricFileSuffix)) {
-          metricFileSuffix = context.getTaskAttemptID().getTaskID().toString();
-        } else {
-          metricFileSuffix += "." + context.getTaskAttemptID().getTaskID().toString();
-        }
-        configuration.set(ConfigurationKeys.METRICS_FILE_SUFFIX, metricFileSuffix);
-        this.jobMetrics.get().startMetricReporting(configuration);
+        this.jobMetrics.get().startMetricReportingWithFileSuffix(HadoopUtils.getStateFromConf(configuration),
+            context.getTaskAttemptID().getTaskID().toString());
       }
     }
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
@@ -14,11 +14,10 @@ package gobblin.runtime.util;
 
 import java.util.List;
 
-import org.apache.hadoop.conf.Configuration;
-
 import com.google.common.collect.Lists;
 
 import gobblin.metrics.GobblinMetrics;
+import gobblin.metrics.MetricContext;
 import gobblin.metrics.Tag;
 import gobblin.runtime.JobState;
 import gobblin.runtime.TaskState;
@@ -35,9 +34,12 @@ public class JobMetrics extends GobblinMetrics {
   protected final String jobName;
 
   protected JobMetrics(JobState job) {
-    super(name(job), null, tagsForJob(job));
-    this.jobName = job.getJobName();
+    this(job, null);
+  }
 
+  protected JobMetrics(JobState job, MetricContext parentContext) {
+    super(name(job), parentContext, tagsForJob(job));
+    this.jobName = job.getJobName();
   }
 
   /**
@@ -59,6 +61,17 @@ public class JobMetrics extends GobblinMetrics {
    */
   public static JobMetrics get(String jobId) {
     return get(null, jobId);
+  }
+
+  /**
+   * Get a new {@link GobblinMetrics} instance for a given job.
+   *
+   * @param jobState the given {@link JobState} instance
+   * @param parentContext is the parent {@link MetricContext}
+   * @return a {@link JobMetrics} instance
+   */
+  public static JobMetrics get(JobState jobState, MetricContext parentContext) {
+    return (JobMetrics) GOBBLIN_METRICS_REGISTRY.getOrDefault(name(jobState), new JobMetrics(jobState, parentContext));
   }
 
   /**
@@ -93,19 +106,17 @@ public class JobMetrics extends GobblinMetrics {
 
   private static List<Tag<?>> tagsForJob(JobState jobState) {
     List<Tag<?>> tags = Lists.newArrayList();
-    tags.add(new Tag<String>("jobName", jobState.getJobName() == null ? "" : jobState.getJobName()));
-    tags.add(new Tag<String>("jobId", jobState.getJobId()));
+    tags.add(new Tag<>("jobName", jobState.getJobName() == null ? "" : jobState.getJobName()));
+    tags.add(new Tag<>("jobId", jobState.getJobId()));
 
     tags.addAll(getCustomTagsFromState(jobState));
     return tags;
   }
 
   /**
-   * @deprecated
-   * Use {@link ClustersNames#getInstance()#getClusterName()}
+   * @deprecated use {@link ClustersNames#getInstance()#getClusterName()}
    */
   public static String getClusterIdentifierTag() {
     return ClustersNames.getInstance().getClusterName();
   }
-
 }
--- a/gobblin-runtime/src/main/java/gobblin/runtime/util/TaskMetrics.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/util/TaskMetrics.java
@@ -62,7 +62,7 @@ public class TaskMetrics extends GobblinMetrics {
 
   protected static List<Tag<?>> tagsForTask(TaskState taskState) {
     List<Tag<?>> tags = Lists.newArrayList();
-    tags.add(new Tag<String>("taskId", taskState.getTaskId()));
+    tags.add(new Tag<>("taskId", taskState.getTaskId()));
     tags.addAll(getCustomTagsFromState(taskState));
     return tags;
   }
@@ -71,5 +71,4 @@ public class TaskMetrics extends GobblinMetrics {
     return JobMetrics.get(taskState.getProp(ConfigurationKeys.JOB_NAME_KEY), taskState.getJobId())
         .getMetricContext();
   }
-
 }
--- a/gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.yarn;
+
+import java.util.List;
+
+import org.apache.hadoop.yarn.api.records.ContainerId;
+
+import com.google.common.collect.ImmutableList;
+
+import gobblin.configuration.State;
+import gobblin.metrics.GobblinMetrics;
+import gobblin.metrics.Tag;
+
+
+/**
+ * Extension of {@link GobblinMetrics} specifically for YARN containers.
+ */
+public class ContainerMetrics extends GobblinMetrics {
+
+  private static final String APPLICATION_NAME = "applicationName";
+  private static final String APPLICATION_ID = "applicationId";
+  private static final String APPLICATION_ATTEMPT_ID = "applicationAttemptId";
+  private static final String CONTAINER_ID = "containerId";
+
+  protected ContainerMetrics(State containerState, String applicationName, ContainerId containerId) {
+    super(name(containerId), null, tagsForContainer(containerState, applicationName, containerId));
+  }
+
+  /**
+   * Get a {@link ContainerMetrics} instance given the {@link State} of a container, the name of the application the
+   * container belongs to, and the {@link ContainerId} of the container.
+   *
+   * @param containerState the {@link State} of the container
+   * @param applicationName a {@link String} representing the name of the application the container belongs to
+   * @param containerId the {@link ContainerId} of the container
+   * @return a {@link ContainerMetrics} instance
+   */
+  public static ContainerMetrics get(State containerState, String applicationName, ContainerId containerId) {
+    return (ContainerMetrics) GOBBLIN_METRICS_REGISTRY
+        .getOrDefault(name(containerId), new ContainerMetrics(containerState, applicationName, containerId));
+  }
+
+  private static String name(ContainerId containerId) {
+    return "gobblin.metrics." + containerId.toString();
+  }
+
+  private static List<Tag<?>> tagsForContainer(State containerState, String applicationName, ContainerId containerId) {
+    ImmutableList.Builder<Tag<?>> tags = new ImmutableList.Builder<>();
+    tags.add(new Tag<>(APPLICATION_NAME, applicationName));
+    tags.add(new Tag<>(APPLICATION_ID, containerId.getApplicationAttemptId().getApplicationId().toString()));
+    tags.add(new Tag<>(APPLICATION_ATTEMPT_ID, containerId.getApplicationAttemptId().getAttemptId()));
+    tags.add(new Tag<>(CONTAINER_ID, containerId.toString()));
+    tags.addAll(getCustomTagsFromState(containerState));
+    return tags.build();
+  }
+}
--- a/gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
@@ -17,13 +17,16 @@ import java.util.List;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+
 import org.apache.helix.task.Task;
 import org.apache.helix.task.TaskCallbackContext;
 import org.apache.helix.task.TaskConfig;
 import org.apache.helix.task.TaskResult;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 
@@ -35,6 +38,7 @@ import gobblin.runtime.JobState;
 import gobblin.runtime.TaskExecutor;
 import gobblin.runtime.TaskState;
 import gobblin.runtime.TaskStateTracker;
+import gobblin.runtime.util.JobMetrics;
 import gobblin.source.workunit.MultiWorkUnit;
 import gobblin.source.workunit.WorkUnit;
 import gobblin.util.JobLauncherUtils;
@@ -62,6 +66,8 @@ public class GobblinHelixTask implements Task {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(GobblinHelixTask.class);
 
+  private final Optional<ContainerMetrics> containerMetrics;
+  private final Optional<JobMetrics> jobMetrics;
   private final TaskExecutor taskExecutor;
   private final TaskStateTracker taskStateTracker;
 
@@ -74,8 +80,11 @@ public class GobblinHelixTask implements Task {
   private final FileSystem fs;
   private final StateStore<TaskState> taskStateStore;
 
-  public GobblinHelixTask(TaskCallbackContext taskCallbackContext, TaskExecutor taskExecutor,
-      TaskStateTracker taskStateTracker, FileSystem fs, Path appWorkDir) throws IOException {
+  public GobblinHelixTask(TaskCallbackContext taskCallbackContext, Optional<ContainerMetrics> containerMetrics, TaskExecutor taskExecutor,
+      TaskStateTracker taskStateTracker, FileSystem fs, Path appWorkDir)
+      throws IOException {
+    this.containerMetrics = containerMetrics;
+
     this.taskExecutor = taskExecutor;
     this.taskStateTracker = taskStateTracker;
 
@@ -85,10 +94,18 @@ public class GobblinHelixTask implements Task {
 
     this.fs = fs;
     Path taskStateOutputDir = new Path(appWorkDir, GobblinYarnConfigurationKeys.OUTPUT_TASK_STATE_DIR_NAME);
-    this.taskStateStore = new FsStateStore<TaskState>(this.fs, taskStateOutputDir.toString(), TaskState.class);
+    this.taskStateStore = new FsStateStore<>(this.fs, taskStateOutputDir.toString(), TaskState.class);
 
     Path jobStateFilePath = new Path(appWorkDir, this.jobId + "." + AbstractJobLauncher.JOB_STATE_FILE_NAME);
     SerializationUtils.deserializeState(this.fs, jobStateFilePath, this.jobState);
+
+    if (this.containerMetrics.isPresent()) {
+      // This must be done after the jobState is deserialized from the jobStateFilePath
+      // A reference to jobMetrics is required to ensure it is not evicted from the GobbinMetricsRegistry Cache
+      this.jobMetrics = Optional.of(JobMetrics.get(this.jobState, this.containerMetrics.get().getMetricContext()));
+    } else {
+      this.jobMetrics = Optional.absent();
+    }
   }
 
   @Override
@@ -97,8 +114,9 @@ public class GobblinHelixTask implements Task {
       Path workUnitFilePath =
           new Path(this.taskConfig.getConfigMap().get(GobblinYarnConfigurationKeys.WORK_UNIT_FILE_PATH));
 
-      WorkUnit workUnit = workUnitFilePath.getName().endsWith(AbstractJobLauncher.MULTI_WORK_UNIT_FILE_EXTENSION) ?
-          MultiWorkUnit.createEmpty() : WorkUnit.createEmpty();
+      WorkUnit workUnit =
+          workUnitFilePath.getName().endsWith(AbstractJobLauncher.MULTI_WORK_UNIT_FILE_EXTENSION) ? MultiWorkUnit
+              .createEmpty() : WorkUnit.createEmpty();
       SerializationUtils.deserializeState(this.fs, workUnitFilePath, workUnit);
 
       // The list of individual WorkUnits (flattened) to run
@@ -117,20 +135,21 @@ public class GobblinHelixTask implements Task {
         workUnits.add(workUnit);
       }
 
-      AbstractJobLauncher.runWorkUnits(this.jobId, this.participantId, workUnits, this.taskStateTracker,
-          this.taskExecutor, this.taskStateStore, LOGGER);
-
+      AbstractJobLauncher
+          .runWorkUnits(this.jobId, this.participantId, workUnits, this.taskStateTracker, this.taskExecutor,
+              this.taskStateStore, LOGGER);
       return new TaskResult(TaskResult.Status.COMPLETED, String.format("completed tasks: %d", workUnits.size()));
     } catch (InterruptedException ie) {
       Thread.currentThread().interrupt();
       return new TaskResult(TaskResult.Status.CANCELED, "");
     } catch (Throwable t) {
+      LOGGER.error("GobblinHelixTask failed due to " + t.getMessage(), t);
       return new TaskResult(TaskResult.Status.ERROR, Throwables.getStackTraceAsString(t));
     }
   }
 
   @Override
   public void cancel() {
-    // TODO: implementation cancellation.
+    // TODO: implement cancellation.
   }
 }
--- a/gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTaskFactory.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTaskFactory.java
@@ -16,12 +16,17 @@ import java.io.IOException;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+
 import org.apache.helix.task.Task;
 import org.apache.helix.task.TaskCallbackContext;
 import org.apache.helix.task.TaskFactory;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Counter;
+
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 
 import gobblin.runtime.TaskExecutor;
@@ -37,13 +42,27 @@ public class GobblinHelixTaskFactory implements TaskFactory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(GobblinHelixTaskFactory.class);
 
+  private static final String GOBBLIN_YARN_NEW_HELIX_TASK_COUNTER = "gobblin.yarn.new.helix.task";
+
+  private final Optional<ContainerMetrics> containerMetrics;
+
+  /**
+   * A {@link Counter} to count the number of new {@link GobblinHelixTask}s that are created.
+   */
+  private final Optional<Counter> newTasksCounter;
   private final TaskExecutor taskExecutor;
   private final TaskStateTracker taskStateTracker;
   private final FileSystem fs;
   private final Path appWorkDir;
 
-  public GobblinHelixTaskFactory(TaskExecutor taskExecutor, TaskStateTracker taskStateTracker,
-      FileSystem fs, Path appWorkDir) {
+  public GobblinHelixTaskFactory(Optional<ContainerMetrics> containerMetrics, TaskExecutor taskExecutor,
+      TaskStateTracker taskStateTracker, FileSystem fs, Path appWorkDir) {
+    this.containerMetrics = containerMetrics;
+    if (this.containerMetrics.isPresent()) {
+      this.newTasksCounter = Optional.of(this.containerMetrics.get().getCounter(GOBBLIN_YARN_NEW_HELIX_TASK_COUNTER));
+    } else {
+      this.newTasksCounter = Optional.absent();
+    }
     this.taskExecutor = taskExecutor;
     this.taskStateTracker = taskStateTracker;
     this.fs = fs;
@@ -53,7 +72,11 @@ public class GobblinHelixTaskFactory implements TaskFactory {
   @Override
   public Task createNewTask(TaskCallbackContext context) {
     try {
-      return new GobblinHelixTask(context, this.taskExecutor, this.taskStateTracker, this.fs, this.appWorkDir);
+      if (this.newTasksCounter.isPresent()) {
+        this.newTasksCounter.get().inc();
+      }
+      return new GobblinHelixTask(context, this.containerMetrics, this.taskExecutor, this.taskStateTracker, this.fs,
+          this.appWorkDir);
     } catch (IOException ioe) {
       LOGGER.error("Failed to create a new GobblinHelixTask", ioe);
       throw Throwables.propagate(ioe);
--- a/gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
@@ -33,6 +33,7 @@ import org.apache.hadoop.yarn.api.ApplicationConstants;
 import org.apache.hadoop.yarn.api.records.ApplicationAttemptId;
 import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.hadoop.yarn.util.ConverterUtils;
+
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
@@ -45,6 +46,7 @@ import org.apache.helix.messaging.handling.MessageHandlerFactory;
 import org.apache.helix.model.Message;
 import org.apache.helix.task.TaskFactory;
 import org.apache.helix.task.TaskStateModelFactory;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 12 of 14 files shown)


gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
```diff
@@ -0,0 +1,22 @@
+package gobblin.dataset.config;
+
+import java.util.*;
```
chavdar: Please avoid * imports.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
```diff
@@ -0,0 +1,22 @@
...
+import com.typesafe.config.Config;
+
+public class ConfigAccessor {
```
chavdar: All public APIs need to have javadocs.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/DatasetUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class DatasetUtils {
+  public static final String ID_DELEMETER = ".";
+  public static final String DATASETS_PREFIX = "config-ds";
+  public static final String TAGS_PREFIX = "config-tag";
+  public static final String CONSTS_PREFIX = "config-const";
+  public static final String ROOT = "root";
+  public static final String IMPORTED_TAGS = "imported-tags";
+
+  public static final String getParentId(String urn) {
```
chavdar: I thought we agreed on using URIs of the form: <config_store_scheme>://[<config_store_server>]:/config/path.

chavdar:
@ibuenros can you also review?

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,31 @@
+package gobblin.dataset.config;
+
+import java.util.*;
+
+import com.typesafe.config.Config;
+
+public interface ConfigStore {
```
chavdar: javadocs

chavdar:
Has this PR been fully updated? I still don't see javadocs for ConfigAccessor .

gobblin-dataset-config/src/main/java/gobblin/dataset/config/PlayGround.java
```diff
@@ -0,0 +1,63 @@
+
```
chavdar: This shouldn't be checked in.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public String getLatestVersion();
+
+  /**
+   * @return the configuration store scheme, example DAI-ETL, DALI or Espresso
+   */
+  public String getScheme();
+  
+  /**
+   * Loaded the latest configuration version.
```
chavdar: Where?

chavdar: Isn't the store stateless? I thought all getConfig() methods on the store API should have a version.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private String configVersion;
```
chavdar: Both fields should be final.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
```diff
@@ -0,0 +1,22 @@
...
+  private ConfigStore configStore;
+  
+  public Config getConfig(String urn){
```
chavdar: I assume there should be a constructor that takes a configVersion and configStore parameters.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigClient.java
```diff
@@ -0,0 +1,37 @@
+package gobblin.dataset.config;
+
+import java.util.*;
+
+/*
+ 
+ config-stores{
```
chavdar: We should be using the Java ServiceLoader API (http://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html). The Hadoop FileSystem implementations use a similar pattern.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+  public void loadConfigs();
+  
+  /**
+   * Loaded the specified configuration version.
+   * @param version ,the configuration version to be loaded
+   */
+  public void loadConfigs(String version);
+  
+  /**
+   * @param urn input urn
```
chavdar: Please be specific in the javadocs what "input urn" means, e.g. "a dataset urn", "a tag urn", or both.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
```diff
@@ -0,0 +1,22 @@
...
+    return configStore.getConfig(urn, configVersion);
+  }
+  
+  public Map<String, Config> getTaggedConfig(String urn){
```
chavdar: Please make sure variable/parameter naming is informative. What "urn" is that?

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+   * @return the Config of input urn
+   */
+  public Config getConfig(String urn);
+  
+  /**
+   * @param urn input urn
+   * @param version, specified configuration version
+   * @return the Config of input urn for the specified configuration version
+   */
+  public Config getConfig(String urn, String version);
```
chavdar: I suggest that in the config store API we use parsed URI objects. This will ensure they are properly formatted. The ConfigAccessor API can use String objects as parameters (for convenience) and parse them to URI objects.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+  
+  /**
+   * @param tag, input tag
+   * @return the Map whose key is the dataset which has input tag associated, not include descendant datasets,
+   *  value is resolved config for the corresponding key. If input is dataset, result is empty as dataset is not taggable
+   */
+  public Map<String, Config> getTaggedConfig(String tag);
```
chavdar: There should be only methods that require an explicit version.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+   *  value is resolved config ,against specifed configuration version, for the corresponding key. 
+   *  If input is dataset, result is empty as dataset is not taggable
+   */
+  public Map<String, Config> getTaggedConfig(String tag, String version);
+  
+  /**
+   * @param tag input tag
+   * @return all the tags associated with input tag directly or in-directly 
+   */
+  public List<String> getAssociatedTags(String tag);
```
chavdar: This should return Collection<URI>

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+  
+  
+  /**
+   * @param tag input tag
+   * @param version, specified configuration version
+   * @return all the tags ,against specifed configuration version, associated with input tag directly or in-directly 
+   */
+  public List<String> getAssociatedTags(String tag, String version);
```
chavdar: This should return Collection&lt;UR&gt;

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,91 @@
...
+  public Map<String, Config> getTaggedConfig(String tag);
+  
+  /**
+   * @param tag, input tag
+   * @param version, specified configuration version
+   * @return the Map whose key is the dataset, against specifed configuration version, which has input tag associated, not include descendant datasets,
+   *  value is resolved config ,against specifed configuration version, for the corresponding key. 
+   *  If input is dataset, result is empty as dataset is not taggable
+   */
+  public Map<String, Config> getTaggedConfig(String tag, String version);
```
chavdar: I think you should return only the dataset URN/URI similarly to getAssociatedTags().

gobblin-dataset-config/src/main/java/gobblin/dataset/config/FileBasedConfigStore.java
```diff
@@ -0,0 +1,142 @@
...
+    this.loadedConfigVersion = version;
+    initialized = true;
+  }
+
+  private void initialCheck(String version) {
+    if (!this.initialized) {
+      this.loadConfigs(version);
+    }
+
+    if (!this.loadedConfigVersion.equals(version)) {
```
chavdar: It is the ConfigAccessor responsibility, not the ConfigStore responsibility to ensure using a consistent version. The ConfigStore should expose only the low-level primitives to access the config.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/FileBasedConfigStore.java
```diff
@@ -0,0 +1,142 @@
...
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * This class is the implementation for the filed based config store. 
+ * @author mitu
+ *
+ */
+public class FileBasedConfigStore implements ConfigStore {
```
chavdar: Is this some prototype implementation or the "real deal"? What is the config layout? How are tags used? Has the config layout been discussed at the review and agreed on? Didn't we discuss that config is going to be deployed on HDFS? The current implementation seems to use regular files.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/PlayGround.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigAccessor.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
tuGithub: Fixed

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,61 @@
...
+  
+  /**
+   * @param tag - the tag which need to be checked for. If the input tag is dataset, empty list will be returned
+   * @param version - specified configuration version
+   * @return all the tags, for the input version, associated with input tag directly or in-directly 
+   */
+  public List<String> getAssociatedTags(String tag, String version);
```
chavdar: The return type should be Collection<URI> since order is not important. As mentioned in my previous comment, the type for tag should be URI too.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,61 @@
...
+  
+  
+  /**
+   * @param tag - the tag which need to be checked for
+   * @param version - specified configuration version
+   * @return the Map whose key is the dataset, against specifed configuration version, which has input tag associated, 
+   * not include descendant datasets, value is resolved config for the corresponding key. 
+   *  If input is dataset, result is empty as dataset is not taggable
+   */
+  public Map<String, Config> getTaggedConfig(String tag, String version);
```
chavdar: The return type should be Collection<URI> since order is not important. As mentioned in my previous comment, the type for tag should be URI too.

gobblin-dataset-config/src/main/java/gobblin/dataset/config/ConfigStore.java
```diff
@@ -0,0 +1,61 @@
...
+   * @return the configuration store scheme, example DAI-ETL, DALI or Espresso
+   */
+  public String getScheme();
+  
+  /**
+   * @param urn - the urn which need to get the configuration
+   * @param version - specified configuration version
+   * @return the resolved Config of input urn for the specified configuration version
+   */
+  public Config getConfig(String urn, String version);
```
chavdar: As mentioned in my previous comment, the type for tag should be URI too.

gobblin-dataset-config/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,17 @@
+package gobblin.config.client;
+
+import java.net.URI;
+
+import com.typesafe.config.Config;
+
+import gobblin.config.configstore.ConfigStore;
+
+public interface ConfigClient {
```
chavdar: What is this interface? What is the difference to gobblin.dataset.config.ConfigClient?

chavdar:
Please have a separate PR with just the APIs.

gobblin-dataset-config/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: gobblin.dataset.config.ConfigClient is the deprecated one and will not be used. 

This file is used for client to get configs

chavdar: I don't understand. What happened to the ConfigAccessor? What about versions? Why do clients need to know about ConfigStores?

gobblin-dataset-config/src/main/java/gobblin/config/configstore/ConfigStore.java
```diff
@@ -0,0 +1,45 @@
...
+import com.typesafe.config.Config;
+
+/**
+ * A ConfigStore is the configuration store per scheme.
+ * 
+ * Each ConfigStore will have it's current version bundled 
+ * @author mitu
+ *
+ */
+public interface ConfigStore extends Serializable{
```
chavdar: What happened to versions?

gobblin-dataset-config/src/test/resources/filebasedConfig/v1.0.1/application.conf
```diff
@@ -0,0 +1,40 @@
+// circular dependency for config-tag.t1.t2.t3 -> config-tag.tag1.tag2 -> config-tag.t1.t2.t3
```
chavdar: I assume these files are included in this PR by mistake. I don't see any class using them.

tuGithub: will remote those files after I made all the unit testing working for the new code as I may need to for test cases

gobblin-dataset-config/src/main/java/gobblin/config/configstore/ConfigStore.java
tuGithub: The current version now is bundled with ConfigStore, that is why the getCurrentVersion() function appeared.

ibuenros:
can we rename the library gobblin-config or gobblin-config-management?

gobblin-dataset-config/src/main/java/gobblin/config/client/ConfigClient.java
ibuenros: Is this intended to be the main entry point to the library for users?

gobblin-dataset-config/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * @param cs
+   * @param uri - the uri relative to the input ConfigStore
+   * @return
+   */
+  public Config getConfig(ConfigStore cs, URI uri);
```
ibuenros: I don't think ConfigStore should be a parameter, it should be inferred from the uri

ibuenros:
Most of the interfaces and methods need javadoc

gobblin-dataset-config/src/main/java/gobblin/config/configstore/ConfigStore.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * @return the configuration store scheme, example DAI-ETL, DALI or Espresso
+   */
+  public String getScheme();
+  
+  public String getCurrentVersion();
+  
+  public URI getParent(URI uri);
```
ibuenros: I thought this would be inferred from the path format of the uri?

gobblin-dataset-config/src/main/java/gobblin/config/configstore/ConfigStoreWithResolution.java
```diff
@@ -0,0 +1,14 @@
...
+import java.net.URI;
+import java.util.Collection;
+
+import com.typesafe.config.Config;
+
+public interface ConfigStoreWithResolution extends ConfigStore{
+  
+  public Config getResolvedConfig(URI uri);
+  
+  public Collection<URI> getResolvedImports(URI uri);
```
ibuenros: what does this return?

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,122 @@
...
+   */
+  public static ConfigClient createConfigClientWithStableVersion( VERSION_STABILITY_POLICY policy ){
+    return null;
+  }
+  
+  /**
+   * 
+   * @return the {@ConfigClient} with VERSION_STABILITY_POLICY.CACHE_CONFIG_IN_MEMORY
+   */
+  public static ConfigClient createDefaultConfigClient( ){
```
ibuenros: I would prefer that the default is RAISE_ERROR ( CACHE_CONFIG_IN_MEMORY can easily become a memory leak)

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,122 @@
...
+   * @return - the directly and indirectly specified configuration in {@com.typesafe.config.Config} format for input uri 
+   * 
+   * <p>
+   * detail logics:
+   * 0. Used the ConfigStoreAccessor in the cached map if previous ConfigStore been queried
+   * 1. Based the scheme name, using {@java.util.ServiceLoader} to find the first {@gobblin.config.configstore.ConfigStoreFactory}
+   * 2. Use ConfigStoreFactory to create ConfigStore 
+   *    using getDefaultConfigStore() if Authority is missing in uri
+   *    using createConfigStore() if Authority is present
+   * 3. Find ConfigStore's root URI by back tracing the input uri, the path which contains "_CONFIG_STORE" is the root
```
ibuenros: This step is ConfigStore dependent. ConfigClient should simply use `getStoreRoot` to get the root, it is up to the store / store factory how they determine the root.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,122 @@
...
+   */
+  public Collection<URI> getImported(URI uri, boolean recursive){
+    return null;
+  }
+  
+  /**
+   * 
+   * @param uri - URI which must start with scheme name
+   * @param recursive - indicate to get the imported by URI recursively or not
+   * @return The java.util.Collection which contains all the URI imported by with input uri
```
ibuenros: "imported by with input uri" -> "which import input uri"

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,122 @@
...
+   */
+  public Collection<URI> getImportedBy(URI uri, boolean recursive){
+    return null;
+  }
+  
+  /**
+   * 
+   * @param uri - clean the cache for the configuration store which specified by input URI
```
ibuenros: Specify that this will cause a new version of the URI to be retrieved next time it is called, which could be different from the last config received.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * 
+   * @param uri - clean the cache for the configuration store which specified by input URI 
+   */
+  public void clearCache(URI uri){
+    
+  }
+  
+  // key is the store ROOT, must use TreeMap
+  Map<URI, ConfigStoreAccessor> theMap;
```
ibuenros: A name other than `theMap` would be nice.

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithImportedByRecursively.java
```diff
@@ -0,0 +1,26 @@
+package gobblin.config.configstore;
+
+import java.net.URI;
+import java.util.Collection;
+import java.util.Map;
+
+import com.typesafe.config.Config;
+
+public interface ConfigStoreWithImportedByRecursively {
```
ibuenros: I think we can combine this with ConfigStoreWithImportedBy

ibuenros:
I looked at all the API classes, are other classes ready?

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithImportedByRecursively.java
tuGithub: Prefer to have another interface as we have two interfaces for getConfig and getResolvedConfig

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: Fixed

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+
+  public BaseHdfsConfigStore(URI root) {
+    this(root, new SimpleVersionComparator());
+  }
+
+  public BaseHdfsConfigStore(URI root, VersionComparator<String> vc) {
+    try {
+      this.storeURI = root;
+      this.location = new Path(root.getPath());
+      this.fs = this.location.getFileSystem(new Configuration());
```
ibuenros: I don't believe this will work as intended. If URI is hdfs://myserver:9000/path/to/config, then root.getPath = /path/to/config, so and getFileSystem will default to the one defined in Configuration regardless of the authority in the input uri.

tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,257 @@
...
+    
+    if(!(cs instanceof ConfigStoreWithStableVersion)){
+      if(this.policy == VERSION_STABILITY_POLICY.RAISE_ERROR){
+        throw new Exception(String.format("Try to connect to unstable config store ", cs.getStoreURI()));
+      }
+    }
+    
+    // ConfigStoreFactory scheme name could be different than configStore's StoreURI's scheme name
+    URI csRoot = cs.getStoreURI();
+    URI key = new URI(csFactory.getScheme(), csRoot.getAuthority(), csRoot.getPath(), csRoot.getQuery(), csRoot.getFragment());
```
ibuenros: We could just throw an error if the scheme is different. Is there a point to allowing this use case?

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,257 @@
...
+   * 
+   * @return the {@ConfigClient} with policy set to VERSION_STABILITY_POLICY.RAISE_ERROR
+   */
+  public static ConfigClient createDefaultConfigClient() {
+    // create with stable versions
+    return createConfigClientWithStableVersion(VERSION_STABILITY_POLICY.RAISE_ERROR);
+  }
+
+  private ConfigStoreAccessor getConfigStoreAccessor(URI uri) throws Exception {
+    if (this.configStoreMap.containsKey(uri)) {
```
ibuenros: You want to check that the map contains an ancestor of uri, not necessarily the uri itself.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,257 @@
...
+    
+    ConfigStoreAccessor csa = this.getConfigStoreAccessor(uri);
+    
+    URI rel_uri = getRelativeUriToStore(csa.store.getStoreURI(), uri);
+    
+    if(csa.store instanceof ConfigStoreWithResolution){
+      return ((ConfigStoreWithResolution) csa.store).getResolvedConfig(rel_uri, csa.version);
+    }
+    
+    SimpleConfigStoreResolver resolver = new SimpleConfigStoreResolver(csa.store);
```
ibuenros: Can we cache the resolvers? They can themselves do some optimizations.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * @param uris - Collection of URI, each one much start with scheme name
+   * @return - the java.util.Map. Key of the map is the URI, value of the Map is getConfig(URI key)
+   */
+  public Map<URI, Config> getConfigs(Collection<URI> uris) throws Exception{
+    Map<URI, Config> result = new HashMap<URI, Config>();
+    Iterator<URI> it = uris.iterator();
+    
+    URI tmp;
+    while(it.hasNext()){
```
ibuenros: can't you just use a foreach loop?

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+  protected final VersionComparator<String> vc;
+
+  public BaseHdfsConfigStore(URI root) {
+    this(root, new SimpleVersionComparator());
+  }
+
+  public BaseHdfsConfigStore(URI root, VersionComparator<String> vc) {
+    try {
+      this.storeURI = root;
+      this.location = new Path(root.getPath());
```
ibuenros: `Path` has a constructor from `URI`, is there a reason for this transformation?

gobblin-config-management/src/main/java/gobblin/config/configstore/VersionComparator.java
```diff
@@ -0,0 +1,8 @@
+package gobblin.config.configstore;
+
+import java.util.Collection;
+
+public interface VersionComparator<V> {
```
ibuenros: Could this just be an extension of Comparator<String>, and then we can just use java's max? No need for `getCurrentVersion`.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+  @Override
+  public URI getStoreURI(){
+    return storeURI;
+  }
+  
+  @SuppressWarnings("deprecation")
+  protected Path getVersionRoot(String version){
+    Path versionRoot = new Path(this.location, version);
+    try {
+      if(this.fs.isDirectory(versionRoot)){
```
ibuenros: This check will be run almost every time a config is requested. Doing it once should be enough.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+    } catch (IOException ioe) {
+      LOG.error("can not find the version " + version + " error:  " + ioe.getMessage(), ioe);
+      throw new VersionDoesNotExistException(ioe.getMessage());
+    } 
+  }
+
+  @Override
+  public Collection<URI> getChildren(URI uri, String version) {
+    if (uri == null)
+      return null;
```
ibuenros: return empty collection instead?

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+    
+    Path self = getPath(uri, version) ;
```
ibuenros: You also want to prepend the root

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+      }
+
+      FSDataInputStream configFileStream = closer.register(this.fs.open(configFile));
+      return ConfigFactory.parseReader(new InputStreamReader(configFileStream)).resolve();
+    }
+    catch (IOException ioe) {
+      LOG.error(String.format("Got error when get own config for %s, exception %s", self, ioe.getMessage()), ioe);
+      throw new RuntimeException(ioe);
+    }
+    finally {
```
ibuenros: Could use try with resource

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+  public Config getOwnConfig(URI uri, String version) {
+    if (uri == null)
+      return ConfigFactory.empty();
+
+    Closer closer = Closer.create();
+    Path self = getPath(uri, version);
+    Path configFile = new Path(self, CONFIG_FILE_NAME);
+    try {
+      if (!this.fs.isFile(configFile)) {
+        return ConfigFactory.empty();
```
ibuenros: Shouldn't this look for an existing ancestor? i.e. it may be a phantom node.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  protected String getRelativePath(Path p, String version) {
+    String root = PathUtils.getPathWithoutSchemeAndAuthority(this.getVersionRoot(version)).toString();
+    String input = PathUtils.getPathWithoutSchemeAndAuthority(p).toString();
+
+    if (input.equals(root)) {
+      return "";
+    }
+    return input.substring(root.length()+1);
```
ibuenros: `URI` has a relativize method. Also, what if path is not a descendant of root?

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+
+    return uri.toString().length()==0;
+  }
+  
+  protected final Path getPath(URI uri, String version){
+    Path self ;
+    if(isRootURI(uri)){
+      self = this.getVersionRoot(version);
+    }else{
+      self = new Path(this.getVersionRoot(version), uri.toString());
```
ibuenros: `URI.toString()` is a relatively expensive operation, and `Path` uses URI's internally either way. It is more efficient to avoid `toString` as much as possible. For example `new Path(this.getVersionRoot(version), new Path(uri))`

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,214 @@
...
+import com.google.common.io.Closer;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * The BaseHdfsConfigStore implements basic functions from ConfigStore which is related to HDFS
+ * @author mitu
+ *
+ */
+public abstract class BaseHdfsConfigStore implements ConfigStore {
```
ibuenros: This should implement store with version.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/ETLHdfsConfigStore.java
```diff
@@ -0,0 +1,165 @@
...
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+
+/**
+ * ETLHdfsConfigStore is used for ETL configuration dataset management
+ * @author mitu
+ *
+ */
+public class ETLHdfsConfigStore extends HdfsConfigStoreWithOwnInclude implements ConfigStoreWithResolution,
```
ibuenros: Is this useful as an open source class?

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/ETLHdfsConfigStore.java
```diff
@@ -0,0 +1,165 @@
...
+    ConfigStoreWithImportedByRecursively, ConfigStoreWithStableVersion {
```
ibuenros: So far, this seems to use the default implementations for resolution and recursion, is there a point for it to extend the interfaces (it might give the wrong idea that it is doing things efficiently).

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: URIComparator will take care of the comparation

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: will fix it

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: will fix it

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: will fix it

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: will store to cache

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: all the URI returned by config store is relative URIs, so can not prepend root

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: This is the direct config for that node, so should return empty.

For getResolvedConfig, it will search ancestors

gobblin-config-management/src/main/java/gobblin/config/configstore/VersionComparator.java
tuGithub: The SimpleVersionComparator/VersionComparator also check whether the String input is valid version or not. Hard to achieve that if only extends Comparator

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: make it private to guarantee the path is the descendant of root

Use URI build in function to get relative URI

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: This function is used to check for Circular dependency as well, not  just for getting Path

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
tuGithub: ETLHdfsConfigStore implements ConfigStoreWithStableVersion. Not sure BaseHdfs should implements ConfigStoreWithStableVersion

gobblin-config-management/src/main/java/gobblin/config/utils/URIComparator.java
```diff
@@ -0,0 +1,56 @@
+package gobblin.config.utils;
+
+import java.net.URI;
+import java.util.Comparator;
+
+
+public class URIComparator implements Comparator<URI> {
```
ibuenros: This comparator is not consistent with equals and it is also not anti-symmetric (e.g. a > b does not necessarily imply b < a). This can lead to strange bugs. I suggest using the natural ordering of URIs, and instead using the method TreeMap.floorKey to determine ascendance.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
ibuenros: See comment on URI comparator.

tuGithub: fixed it

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/ETLHdfsConfigStore.java
tuGithub: Made the ETLHdfsConfigStore NOT ConfigStoreWithResolution, relied on the Client to do the resolution

gobblin-config-management/src/main/java/gobblin/config/utils/URIComparator.java
tuGithub: fixed it

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+  private final TreeMap<URI, ConfigStoreAccessor> configStoreMap = new TreeMap<URI, ConfigStoreAccessor>();
+
+  // key is the configStore scheme name, value is the ConfigStoreFactory
+  private static final Map<String, ConfigStoreFactory> configStoreFactoryMap =
+      new HashMap<String, ConfigStoreFactory>();
+
+  static {
+    ServiceLoader<ConfigStoreFactory> loader = ServiceLoader.load(ConfigStoreFactory.class);
+    for (ConfigStoreFactory f : loader) {
+      configStoreFactoryMap.put(f.getScheme(), f);
```
ibuenros: Can we make this initialization lazy?

gobblin-config-management/src/main/java/gobblin/config/utils/PathUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return null;
+  }
+
+  /**
+   * @param parent - parent URI
+   * @param child - descendant URI
+   * @return true if o2 is the descendant of o1 
+   */
+  public static boolean checkDescendant(URI parent, URI child) {
```
ibuenros: We have at least one other implementation of this method in gobblin-data-management (it is also simpler). We should just use one in gobblin-util.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+    ConfigStore cs = csFactory.createConfigStore(uri);
+
+    if (!(cs instanceof ConfigStoreWithStableVersion)) {
+      if (this.policy == VERSION_STABILITY_POLICY.RAISE_ERROR) {
+        throw new Exception(String.format("Try to connect to unstable config store ", cs.getStoreURI()));
+      }
+    }
+
+    URI key = cs.getStoreURI();
+    ConfigStoreAccessor value = new ConfigStoreAccessor(cs, cs.getCurrentVersion());
```
ibuenros: variable name should be more descriptive

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+   * 
+   * <p>
+   * detail logics:
+   * 1. Used the ConfigStoreAccessor in the cached map if previous ConfigStore been queried
+   * 2. Based the scheme name, using {@java.util.ServiceLoader} to find the first {@gobblin.config.configstore.ConfigStoreFactory}
+   * 3. Use ConfigStoreFactory to create ConfigStore 
+   * 4. Build ConfigStoreAccessor by checking the current version of the ConfigStore. Added the entry to theMap
+   * 5. If the ConfigStore is NOT ConfigStoreWithResolution, need to do resolution in this client
+   */
+  public Config getConfig(URI uri) throws Exception {
```
chavdar: Please don't have "throws Exception" because it is meaningless. Either we know what exceptions can happen and we should have an explicit list of exception classes or don't specify anything and use RuntimeException.

chavdar: Let's also avoid generic names as "uri" as they convey very little information. Let's, for example, agree to call the config URIs, configKeyUri if it is a URI or configKey if it is a String.

chavdar: We need a convenience method getConfig(String configKey) which will parse the string into a URI.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+
+    SimpleConfigStoreResolver resolver = csa.resolver;
+    return resolver.getResolvedConfig(rel_uri, csa.version);
+  }
+
+  /**
+   * @param uris - Collection of URI, each one much start with scheme name
+   * @return - the java.util.Map. Key of the map is the URI, value of the Map is getConfig(URI key)
+   */
+  public Map<URI, Config> getConfigs(Collection<URI> uris) throws Exception {
```
chavdar: Ditto on "uris" name, "throws Exception", convenience method taking a Collection<String>.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithResolution.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface ConfigStoreWithResolution extends ConfigStore {
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - the directly and indirectly specified configuration in com.typesafe.config.Config format for input uri 
+   *  against input configuration version
+   */
+  public Config getResolvedConfig(URI uri, String version);
```
chavdar: We need another method Map<URI, Config> getResolvedConfigs(Collection<URI>, String version) for batched fetches.

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStore.java
```diff
@@ -0,0 +1,47 @@
...
+   */
+  public Collection<URI> getOwnImports(URI uri, String version);
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - the directly specified configuration in com.typesafe.config.Config format for input uri 
+   *  against input configuration version
+   */
+  public Config getOwnConfig(URI uri, String version);
```
chavdar: I think we need another method for batched fetches Map<URI, Config> getOwnConfigs(Collection<URI>, String version)

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithImportedBy.java
```diff
@@ -0,0 +1,33 @@
...
+  public Collection<URI> getImportedBy(URI uri, String version);
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - The {@java.util.Map}. The key of the Map is the URI directly import input uri 
+   *  against input configuration version. The value of the Map the directly specified configuration in 
+   *  com.typesafe.config.Config format for corresponding key
+   */
+  public Map<URI, Config> getConfigsImportedBy(URI uri, String version);
```
chavdar: Does this method return own configs or resolved configs? I don't think this method should be in this interface. Instead, I think we need a couple more methods for batch fetches in ConfigStore and ConfigStoreWithResolution.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+    Map<URI, Config> result = new HashMap<URI, Config>();
```
chavdar: You can add a TODO comment that we can have a logic if the number URIs is large, we can parallelize the calls using the ThreadPool.

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStore.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return - the direct children URIs for input uri against input configuration version
+   */
+  public Collection<URI> getChildren(URI uri, String version);
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - the directly imported URIs for input uri against input configuration version
+   */
+  public Collection<URI> getOwnImports(URI uri, String version);
```
chavdar: This has to return a List<URI> because the order is important!

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @Override
+  public Config getResolvedConfig(URI uri, String version) {
+    UUID traceID = UUID.randomUUID();
+    LOG.info(String.format("Getting resolved config for uri: %s for version %s, with traceID: %s", 
+        uri, version, traceID));
+    return getResolvedConfigWithTrace(uri, version, traceID.toString());
+  }
+
+  protected Config getResolvedConfigWithTrace(URI uri, String version, String traceID) {
```
chavdar: what is "traceID" ?

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+ * @author mitu
+ *
+ */
+@SuppressWarnings("rawtypes")
+public class ConfigClient {
+
+  private static final Logger LOG = Logger.getLogger(ConfigClient.class);
+
+  public static enum VERSION_STABILITY_POLICY {
+    RAISE_ERROR,
```
chavdar: I think we need a "NO_STABILITY" which means that no stability is required.

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithStableVersion.java
```diff
@@ -0,0 +1,12 @@
+package gobblin.config.configstore;
+
+/**
+ * ConfigStoreWithStableVersion is used to indicate that the configuration store support stable version. 
+ * 
+ * Once the version been published to the configuration store, the content will not be changed
+ * @author mitu
```
chavdar: Please crossreference ConfigClient.VERSION_STABILITY_POLICY

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+ * ConfigClient is the consumer facing class for retrieving configuration against all configuration stores.
+ * @author mitu
+ *
+ */
+@SuppressWarnings("rawtypes")
+public class ConfigClient {
+
+  private static final Logger LOG = Logger.getLogger(ConfigClient.class);
+
+  public static enum VERSION_STABILITY_POLICY {
```
chavdar: This should be named VersionStabilityPolicy. We generally reserve ALL_CAPS for constants.

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: added

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: Fixed

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: done

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: done

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithStableVersion.java
tuGithub: done

gobblin-config-management/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: is NO_STABILITY the same as CACHE_CONFIG_IN_MEMORY?

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/SimpleConfigStoreResolver.java
tuGithub: The client may query multiple URIs and even multiple times for the same URI. traceID is used to annotate the single session of the query. Easy to find the whole query in Log

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStore.java
tuGithub: Should we have the same function in ConfigStoreWithResolution?

public Collection<URI> getImportsRecursively(URI uri, String version)

gobblin-config-management/src/main/java/gobblin/config/configstore/ConfigStoreWithResolution.java
tuGithub: let us have a talk for the new functions in the interface

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/SimpleVersionFinder.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Collection;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import gobblin.config.configstore.VersionFinder;
+
+
+public class SimpleVersionFinder implements VersionFinder<String> {
```
chavdar: I don't think we should care about the specific version string format. I suggest that we say that we have a "versions" subdirectory and every non-hidden file under it is a version. 

I don't know what our versioning scheme will be.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/SimpleVersionFinder.java
```diff
@@ -0,0 +1,80 @@
...
+      if (P1.matcher(s).matches() || P2.matcher(s).matches() || P3.matcher(s).matches()) {
+        res.add(s);
+      }
+    }
+
+    return res;
+  }
+
+  @Override
+  public String getCurrentVersion(Collection<String> input) {
```
chavdar: I don't think this will work. For a store with stable versions, we need to be able to preserve the versions even if they are rolled back. This means that the latest (greatest) version may be rolled back and cannot be used as current. I think we have to use the _CONFIG_STORE file to store the latest version.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,213 @@
+package gobblin.config.configstore.impl;
```
chavdar: I think package should be something along the lines of gobblin.config.store.hadoopfs . Hadoopfs should certainly be somewhere in the package name.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
chavdar: I think we shouldn't use HDFS but HadoopFS . We very well may have to use the store deployed on Local FS for stand-alone deployments.

gobblin-config-management/src/main/java/gobblin/config/configstore/impl/BaseHdfsConfigStore.java
```diff
@@ -0,0 +1,213 @@
...
+  private final Map<String, Path> versionRootMap = new HashMap<String, Path>();
+  private final URI storeURI;
+  private final Path location;
+  private final String currentVersion;
+  protected final FileSystem fs;
+  protected final VersionFinder<String> vc;
+
+  /**
+   * 
+   * @param physical_root - actual hdfs URI, example: new URI("hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest")
```
chavdar: please remove references to linkedin-specific paths.

gobblin-config-management/src/main/java/gobblin/config/utils/PathUtils.java
```diff
@@ -0,0 +1,87 @@
+package gobblin.config.utils;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.apache.hadoop.fs.Path;
```
chavdar: I am not sure we should depend on this. This will make the client implementation dependent on Hadoop which it should not be.

gobblin-config-management/src/test/java/gobblin/config/configstore/impl/FilesUtil.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class FilesUtil {
+
+  /**
+   * Used to sync two directories recursively
+   * @param input - input directory, must be an existing directory
+   * @param output - output directory, must be Non existing 
+   * @throws IOException
+   */
+  public static void SyncDirs(File input, File output) throws IOException {
```
chavdar: standard naming convention is for functions to start with a small letter.

gobblin-config-management/build.gradle
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
```
chavdar: please remove this file

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStore.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return - the direct children URIs for input uri against input configuration version
+   */
+  public Collection<URI> getChildren(URI uri, String version) throws VersionDoesNotExistException;
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - the directly imported URIs for input uri against input configuration version
+   */
+  public Collection<URI> getOwnImports(URI uri, String version) throws VersionDoesNotExistException;
```
chavdar: This should be a List<URI> since the order of imports matters.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreCreationException.java
```diff
@@ -0,0 +1,17 @@
+package gobblin.config.store.api;
+
+public class ConfigStoreCreationException extends Exception {
```
chavdar: To make this exception class more useful, I'd add a storeURI attribute with a getter and also automatically include the store URI in the message.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithResolution.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+  public Config getResolvedConfig(URI uri, String version) throws VersionDoesNotExistException;
+
+  /**
+   * @param uri - the uri relative to this configuration store
+   * @param version - specify the configuration version in the configuration store.
+   * @return - the directly and indirectly imported URIs followed the imported path for input uri 
+   *  against input configuration version
+   */
+  public Collection<URI> getImportsRecursively(URI uri, String version) throws VersionDoesNotExistException;
```
chavdar: The return result needs to List<URI> as order is significant.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/VersionDoesNotExistException.java
```diff
@@ -0,0 +1,17 @@
+package gobblin.config.store.api;
+
+public class VersionDoesNotExistException extends Exception {
+
+  /**
+   * 
+   */
+  private static final long serialVersionUID = 2736458021800944664L;
+
+  public VersionDoesNotExistException(String message) {
```
chavdar: I think this exception should take a store URI and a version instead of a generic message. It should also define getters for those fields.

chavdar:
I also think we need an interface

interface ConfigStoreWithBatchFetches extends ConfigStore {
    Map<URI, Config> getOwnConfigs(Collection<URI> configUris, String version)
}

gobblin-config-management/build.gradle
tuGithub: removed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStore.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreCreationException.java
tuGithub: updated

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithResolution.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/VersionDoesNotExistException.java
tuGithub: updated

tuGithub:
added interface ConfigStoreWithBatchFetches

chavdar:
LGTM

PR #402, opened 2015-10-22
Fix 2 findbugs correctness warnings

zliu41:
The reason to change `topicSpecificStateMap` from `Map<KafkaTopic, State>` to  `Map<String, State>` is that some methods were calling `get` and passes a `String`. And `KafkaTopic` doesn't even implement `equals` and `hashCode`.


zliu41:
@sahilTakiar can you review?

sahilTakiar:
LGTM

PR #405, opened 2015-10-23
fix bug in AvroSchemaFieldRemover

zliu41:
@chavdar


chavdar:
Please add unit test(s).

PR #408, opened 2015-10-23
Add unit test for AvroSchemaFieldRemover

zliu41:


gobblin-core/src/test/java/gobblin/converter/filter/AvroSchemaFieldRemoverTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @BeforeClass
+  public void setUp() throws IOException {
+    this.fs = FileSystem.get(new Configuration());
+  }
+
+  @Test
+  public void testRemoveFields() throws IllegalArgumentException, IOException {
+    Schema convertedSchema1 =
+        convertSchema("gobblin-core/src/test/resources/converter/recursive_schema_1.avsc", "results.activityInfo");
```
chavdar: This relies on the test being run from a specific directory (the root). That may vary depending on if we use Eclipse, IDEA, or the command-line. Since the avsc files is already in src/test/resources, it will be automatically included as a resource in the test jar. We should use ClassLoader.getResourceAsStream("converter/recursive_schema_1.avsc") to read the file instead.

zliu41:
@chavdar updated.

gobblin-core/src/main/java/gobblin/converter/filter/AvroSchemaFieldRemover.java
```diff
@@ -99,6 +99,9 @@ private Schema removeFields(Schema schema, Map<String, Schema> schemaMap) {
   }
 
   private Schema removeFieldsFromRecords(Schema schema, Map<String, Schema> schemaMap) {
+    Schema newRecord = Schema.createRecord(schema.getName(), schema.getDoc(), schema.getNamespace(), schema.isError());
```
chavdar: nit: This may be safe now because the processing is single-threaded but I think, in general, it is dangerous to put the new record schema in the map before fully constructed. Previously, it was constructed at the end. Why did you have to move it at the beginning?

zliu41: This is the way Camus does it. The original code failed in one of the tests with infinite recursion. I'm not yet exactly sure why it is the case since that test case was a bit complicated.

chavdar: This is probably related to the recursion on line 110. Could try to understand why this is needed and if we can avoid it? For me, this is a dangerous and error-prone approach.

chavdar: I chatted with @zliu41 and the current approach should be fine since the schema is no longer a member field and it is used only in the context of removeFields() recursion.

LGTM with @zliu41  adding some code comments about the situation.

PR #409, opened 2015-10-24
Fixed an issue that may cause Kafka adapter to miss data if the Fork fails

liyinan926:
1. Introduced a new `JobCommitPolicy.COMMIT_SUCCESSFUL_TASKS` that will only commit successful tasks. This is recommended to be used with jobs with task-level data publishing so tasks can control independently if their output data should be published.
2. Updated `KafkaExtractor` to skip updating the actual high watermark if the `WorkUnit` fails.
3. Used a `ExecutorCompletionService` to run `Fork`s of a `Task` and for waiting for the `Fork`s to finish, instead of a `CountDownLatch`.
4. Other code factoring.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
@@ -274,12 +274,16 @@ public abstract class KafkaExtractor<S, D> extends EventBasedExtractor<S, D> {
     this.workUnitState.setProp(ConfigurationKeys.ERROR_MESSAGE_INVALID_SCHEMA_ID_COUNT, this.invalidSchemaIdCount);
     this.workUnitState.setProp(ConfigurationKeys.ERROR_MESSAGE_UNDECODABLE_COUNT, this.undecodableMessageCount);
 
-    // Commit actual high watermark for each partition
-    for (int i = 0; i < this.partitions.size(); i++) {
-      LOG.info(String.format("Actual high watermark for partition %s=%d, expected=%d", this.partitions.get(i),
-          this.nextWatermark.get(i), this.highWatermark.get(i)));
+    if (this.workUnitState.getWorkingState() == WorkUnitState.WorkingState.COMMITTED) {
+      // Commit actual high watermark for each partition
+      for (int i = 0; i < this.partitions.size(); i++) {
+        LOG.info(String.format("Actual high watermark for partition %s=%d, expected=%d", this.partitions.get(i),
+            this.nextWatermark.get(i), this.highWatermark.get(i)));
+      }
+      this.workUnitState.setActualHighWatermark(this.nextWatermark);
+    } else {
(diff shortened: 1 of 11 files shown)


liyinan926:
@zliu41 can you review?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -232,20 +230,19 @@ public void run() {
 
       try {
         if (shouldPublishDataInTask()) {
-
           // If data should be published by the task, publish the data and set the task state to COMMITTED.
           // Task data can only be published after all forks have been closed by closer.close().
           publishTaskData();
           this.taskState.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
```
zliu41: Seems the taskstate should be set to `COMMITTED` here only if the task succeeded. Otherwise the mapper will not fail and we won't notice the task failed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
```diff
@@ -274,12 +274,16 @@ public void close() throws IOException {
     this.workUnitState.setProp(ConfigurationKeys.ERROR_MESSAGE_INVALID_SCHEMA_ID_COUNT, this.invalidSchemaIdCount);
     this.workUnitState.setProp(ConfigurationKeys.ERROR_MESSAGE_UNDECODABLE_COUNT, this.undecodableMessageCount);
 
-    // Commit actual high watermark for each partition
-    for (int i = 0; i < this.partitions.size(); i++) {
-      LOG.info(String.format("Actual high watermark for partition %s=%d, expected=%d", this.partitions.get(i),
-          this.nextWatermark.get(i), this.highWatermark.get(i)));
+    if (this.workUnitState.getWorkingState() == WorkUnitState.WorkingState.COMMITTED) {
```
zliu41: in `JobContext.persistDatasetState()` it rolls back the actual high watermark for uncommitted workunitstates. I think this part can be done there so that extractor implementations don't have to worry about job commit policy.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
liyinan926: If the job commit policy is `COMMIT_ON_PARTIAL_SUCCESS`, we will still commit output data of failed tasks, so setting task state to `COMMITTED` does not make nonsense. Also the client/driver will only check individual task states in `JobContext` if and only if the commit policy is `COMMIT_ON_FULL_SUCCESS`. 

I think we need a meeting to go through the different commit policies and their relationship with the final task/job states and make the semantics clear and stick to such semantics. This is currently a quite error-prone and not intuitive at all.

liyinan926:
@zliu41 updated 036bebb.

liyinan926:
@zliu41 can you take a look again?

zliu41:
LGTM

PR #410, opened 2015-10-27
Generalized datasets. Added partitionable dataset, copyable dataset.

ibuenros:


ibuenros:
@chavdar @pcadabam this is PR 1 for the distcp - sftp implementation. It includes only changes to dataset stuff.

PR #411, opened 2015-10-27
Dataset and copyable dataset changes for gobblin distcp

pcadabam:
@ibuenros and @chavdar please take a look at this.

Split this PR https://github.com/linkedin/gobblin/pull/403 into 2 parts. It was difficult to split part 1 as they were closely related.

Part 1 - Dataset and CopyableDataset changes 
Part 2 - SFTP changes

This PR is for part 1. I will send out another PR for part 2 after merging this.


gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Implementation of {@link CopyableDataset} that creates a {@link CopyableFile} for every file that is a descendant if
+ * the root directory.
+ */
+public class RecursiveCopyableDataset extends SinglePartitionCopyableDataset {
+
+  public static final String TARGET_DIRECTORY = "gobblin.copyable.dataset.target.directory";
+
+  private final Path rootPath;
+  private final FileSystem _fs;
```
ibuenros: can we remove the underscore in the name?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+  private final Path targetDirectory;
+  private final Properties properties;
+  private LoadingCache<Path, OwnerAndPermission> ownerAndPermissionCache;
+  private final PathFilter pathFilter;
+
+  public RecursiveCopyableDataset(FileSystem fs, Path rootPath, Properties properties) {
+
+    Preconditions.checkArgument(
+        !Strings.isNullOrEmpty(properties.getProperty(ConfigurationKeys.DATA_PUBLISHER_FINAL_DIR)), "Missing property "
```
ibuenros: DATA_PUBLISHER_FINAL_DIR is a gobblin ingestion key, is it a good idea to use it for the copyable dataset?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+              PathUtils.getPathWithoutSchemeAndAuthority(datasetRoot()));
+
+      Path outputPath = new Path(this.targetDirectory, relativeOutputPath);
+
+      OwnerAndPermission ownerAndPermission =
+          new OwnerAndPermission(file.getOwner(), file.getGroup(), file.getPermission());
+      List<OwnerAndPermission> ancestorOwnerAndPermissions = Lists.newArrayList();
+      try {
+        Path currentPath = PathUtils.getPathWithoutSchemeAndAuthority(file.getPath());
+        while (currentPath != null && !currentPath.getParent().equals(this.rootPath)) {
```
ibuenros: This can actually throw an NPE if currentPath has no parent.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDataset.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface CopyableDataset extends PartitionableDataset<CopyableFile> {
+
+  /**
+   * Find all {@link CopyableFile}s in this dataset.
+   * @return List of {@link CopyableFile}s in this dataset.
+   * @throws IOException
+   * @param targetFileSystem
+   */
+  public List<CopyableFile> getCopyableFiles(FileSystem targetFileSystem) throws IOException;
```
chavdar: I don't think we need the targetFileSystem parameter. That should be populated by the finder.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -0,0 +1,166 @@
...
+ * Abstraction for a file to copy from {@link #origin} to {@link #destination}.
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode
+public class CopyableFile implements File, Writable {
+
+  public static final String SERIALIZED_COPYABLE_FILE = "gobblin.copy.serialized.copyable.file";
```
chavdar: I don't understand why we need this property. If someone needs to serialize the object in the state, they can explicitly set whatever property they want.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -0,0 +1,166 @@
...
+
+  /** {@link FileStatus} of the existing origin file. */
+  private FileStatus origin;
```
chavdar: I don't think this should be part of the object:

1) We don't want to force a FS operation to get the status for every single file (that can be expensive!)
2) File status may change over time. It is better to read it whenever needed. 

At the very best, the instance here should be just a cached copy which is populated on demand.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -0,0 +1,166 @@
...
+   * For example, if {@link #destination} is /a/b/c/file, then the first element of this list is the desired owner and
+   * permission for directory /a/b/c, the second is the desired owner and permission for directory /a/b, and so on.
+   * </p>
+   *
+   * <p>
+   * If there are fewer elements in the list than ancestor directories in {@link #destination}, it is understood that
+   * extra directories are allowed to have any owner and permission.
+   * </p>
+   */
+  private List<OwnerAndPermission> ancestorsOwnerAndPermission;
```
chavdar: Do we need such freedom? Can't we just have one OwnerAndPermission for any ancestor directory that needs to be created?

ibuenros: if we are copying a directory, we probably want to replicate the attributes of each subdirectory. Since we only copy files, for each file we need to store the attributes of its parent directories at least up to the directory that is being copied.

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
pcadabam: Some thoughts 
- I think we can remove this. This is currently used to set the destination path in CopyableFile which is not used. Only the relative path is used. 
I feel the CopyableFile should only care about "relative" destination path. The writers create this relative path in the task output directory and write the file in it. Publisher will decide where the directory structures in task output should be published. 
If we need to publish to multiple base directories, they should probably be a different job as the publish happens at job level.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -0,0 +1,166 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode
+public class CopyableFile implements File, Writable {
+
+  public static final String SERIALIZED_COPYABLE_FILE = "gobblin.copy.serialized.copyable.file";
+
+  /** {@link FileStatus} of the existing origin file. */
+  private FileStatus origin;
+  /** Destination {@link Path} of the file. */
+  private Path destination;
```
chavdar: Is this a FQ path including file system URI? If please, please document. I am not sure how this is different from relativeDestination?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Serialize an instance of {@link CopyableFile} into a {@link String}. Usually use to store the copyableFile in state
+   * at key {@link #SERIALIZED_COPYABLE_FILE}
+   *
+   * @param copyableFile to be serialized
+   * @return serialized string
+   */
+  public static String serializeCopyableFile(CopyableFile copyableFile) throws IOException {
```
chavdar: I think if we are serializing to a String, we better use something human-readable like json.

gobblin-data-management/src/main/java/gobblin/data/management/partition/File.java
```diff
@@ -0,0 +1,25 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.partition;
+
+import org.apache.hadoop.fs.FileStatus;
+
+
+/**
+ * Interface representing a File.
```
chavdar: Please document why we need this.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
pcadabam: destination - file path including final publish directory
relativeDestination - file path relative to dataset root.

Eg. /export/content/data/LogData/2015/10/11/f1.txt
dataset root - /export/content/data
destination - /users/cadabam/copied/LogData/2015/10/11/f1.txt
relativeDestination - /LogData/2015/10/11/f1.txt

In the current implementation I am using relativePath as the publisher final directory is same for all datasets. The publisher decides the final publish directory and creates /LogData/2015/10/11/f1.txt under it.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
pcadabam: @ibuenros did you have a specific reason to choose this over json?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
pcadabam: Moved to source

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDataset.java
pcadabam: Removed

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
pcadabam: Fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
pcadabam: In case we want to support multiple datasets with different publish final directories we can pass the publisher final directory through the constructor instead of a properties object

gobblin-data-management/src/main/java/gobblin/data/management/partition/File.java
pcadabam: Done

pcadabam:
@chavdar and @ibuenros please take another look. 
I have addressed most of the comments and replied to others

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
ibuenros: Mainly that is more compact, and most Hadoop objects are already Writeable, which only supports byte serialization.
I am not opposed to json, but it will take more work I'm guessing.

pcadabam: Changed it to json.

pcadabam:
Changed serialization to json

gobblin-data-management/src/main/java/gobblin/data/management/copy/IdentityCopyableDatasetFinder.java
```diff
@@ -0,0 +1,32 @@
+package gobblin.data.management.copy;
```
ibuenros: This class is unused at the moment, and it is replaceable by `CopyableGlobDatasetFinder`, can we remove it?

gobblin-data-management/src/main/java/gobblin/data/management/dataset/DatasetUtils.java
```diff
@@ -0,0 +1,92 @@
...
+      throw new IOException(exception);
+    } catch (IllegalAccessException exception) {
+      throw new IOException(exception);
+    } catch (InvocationTargetException exception) {
+      throw new IOException(exception);
+    }
+
+  }
+
+  public static PathFilter instantiatePathFilter(Properties props) {
```
ibuenros: javadoc

gobblin-data-management/src/main/java/gobblin/data/management/partition/File.java
ibuenros: This is also the abstraction that partitionable dataset uses.

gobblin-data-management/src/main/java/gobblin/data/management/copy/IdentityCopyableDatasetFinder.java
pcadabam: Removed

gobblin-data-management/src/main/java/gobblin/data/management/dataset/DatasetUtils.java
pcadabam: Added

gobblin-runtime/src/main/java/gobblin/runtime/CommandLineLauncher.java
```diff
@@ -0,0 +1,49 @@
+package gobblin.runtime;
```
ibuenros: Actually, can you remove this class?
A better version will be added in a different PR (with the partitioned writer).

pcadabam: Removed

ibuenros:
LGTM (make sure to squash the commits before merging)

pcadabam:
@chavdar can I merge this?

chavdar:
LGTM

PR #414, opened 2015-10-28
Add option to set max time ago for AvroHdfsTimePartitionedWriter

zliu41:
Records older than max allowed time ago will be dropped.

Currently there's no limit, and we had situations where some Kafka partitions contain records in thousands of hours, which caused the mapper to OOM.

There is a similar property in Camus.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
@@ -20,11 +20,13 @@ import java.util.Map.Entry;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.hadoop.fs.Path;
-
+import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.Period;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
-
+import org.joda.time.format.PeriodFormatter;
+import org.joda.time.format.PeriodFormatterBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -79,6 +81,8 @@ import gobblin.util.WriterUtils;
 public class AvroHdfsTimePartitionedWriter implements DataWriter<GenericRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(AvroHdfsTimePartitionedWriter.class);
 
+  public static final String TIME_PARTITIONED_WRITER_MAX_TIME_AGO = "time.partitioned.writer.max.time.ago";
+
   /**
    * This is the base file path that all data will be written to. By default, data will be written to
    * /datasetName/daily/[yyyy]/[MM]/[dd]/.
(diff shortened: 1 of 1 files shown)


zliu41:

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -150,14 +157,34 @@ public AvroHdfsTimePartitionedWriter(Destination destination, String writerId, S
...
-            ConfigurationKeys.DEFAULT_WRITER_PARTITION_TIMEZONE)));
+        .withZone(this.timeZone);
 
     this.partitionColumns = getWriterPartitionColumns();
+    this.earliestAllowedTimestamp = getEarliestAllowedTimestamp();
+  }
+
+  private Optional<Long> getEarliestAllowedTimestamp() {
+    if (!this.properties.contains(TIME_PARTITIONED_WRITER_MAX_TIME_AGO)) {
+      return Optional.<Long> absent();
```
liyinan926: Additional empty space before `absent`.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -150,14 +157,34 @@ public AvroHdfsTimePartitionedWriter(Destination destination, String writerId, S
...
+      DateTime currentTime = new DateTime(this.timeZone);
+      PeriodFormatter periodFormatter = getPeriodFormatter();
+      String maxTimeAgoStr = this.properties.getProp(TIME_PARTITIONED_WRITER_MAX_TIME_AGO);
+      Period maxTimeAgo = periodFormatter.parsePeriod(maxTimeAgoStr);
+      return Optional.of(currentTime.minus(maxTimeAgo).getMillis());
+    }
+  }
+
+  private PeriodFormatter getPeriodFormatter() {
+    return new PeriodFormatterBuilder().appendMonths().appendSuffix("m").appendDays().appendSuffix("d").appendHours()
```
liyinan926: Seems this can be defined as a constant.

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -79,6 +81,8 @@
 public class AvroHdfsTimePartitionedWriter implements DataWriter<GenericRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(AvroHdfsTimePartitionedWriter.class);
 
+  public static final String TIME_PARTITIONED_WRITER_MAX_TIME_AGO = "time.partitioned.writer.max.time.ago";
```
chavdar: Can you add javadocs about the format?

chavdar:
Please add a unit test for this?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsTimePartitionedWriter.java
```diff
@@ -195,6 +222,10 @@ protected long getRecordTimestamp(Optional<Object> writerPartitionColumnValue) {
   }
 
   protected void write(GenericRecord record, long recordTimestamp) throws IOException {
+    if (recordTooOld(recordTimestamp)) {
+      return;
```
chavdar: We need a metric for how many rows have been dropped.

chavdar:
Is that needed anymore now that we have AvroRecordTimestampLowerBoundPolicy ?

zliu41:
`AvroHdfsTimePartitionedWriter` is gone.

PR #415, opened 2015-10-28
Do not override `MultiWorkUnit` grouping in Gobblin Yarn

liyinan926:
Currently, the `GobblinHelixJobLauncher` flattens `MultiWorkUnit`s and hope for Helix to do a balanced assignments of tasks to containers. However, both Gobblin Kafka and IDPC HDFS Purger already have their own logic of grouping `WorkUnit`s into `MultiWorkUnit`s. It's good to keep their logic and not to override it.


liyinan926:
@zliu41 can you review?

sahilTakiar:
LGTM

PR #418, opened 2015-10-29
Deleted deprecated classes that are no longer being used

liyinan926:
1. Deleted classes for the very old deprecated standalone mode that are no longer used.
2. Added logging for task failures in `MRJobLauncher` on the client/driver side so task exceptions are available even without needing to check the individual mapper logs.
3. Some other code refactoring.
 

liyinan926:
@sahilTakiar can you review?

liyinan926:
@zliu41 can you review?

liyinan926:
@zliu41 can you take a look?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinOutputCommitter.java
```diff
@@ -61,10 +62,10 @@ public void abortJob(JobContext jobContext, JobStatus.State state) throws IOExce
     FileSystem fs = FileSystem.get(fsUri, conf);
 
     Path mrJobDir = new Path(conf.get(ConfigurationKeys.MR_JOB_ROOT_DIR_KEY), conf.get(ConfigurationKeys.JOB_NAME_KEY));
-    Path jobInputDir = new Path(mrJobDir, "input");
+    Path jobInputDir = new Path(mrJobDir, MRJobLauncher.INPUT_DIR_NAME);
 
-    if (!fs.exists(jobInputDir) || !fs.getFileStatus(jobInputDir).isDir()) {
-      LOG.warn("Folder " + jobInputDir + " containing serialized WorkUnits doesn't exist. No data will cleaned up.");
+    if (!fs.exists(jobInputDir) || !fs.isDirectory(jobInputDir)) {
```
zliu41: why change this line?

liyinan926: `fs.isDirectory` is a simpler way of doing the same thing and `isdir` in the original code has been deprecated.

gobblin-core/src/main/java/gobblin/converter/string/TextToStringConverter.java
```diff
@@ -0,0 +1,50 @@
...
+import gobblin.converter.SingleRecordIterable;
+
+
+/**
+ * An implementation of {@link Converter} that converts input records of type {@link Text} to strings.
+ *
+ * @author ynli
+ */
+@SuppressWarnings("unused")
+public class TextToStringConverter extends Converter<Schema, Schema, Text, String> {
```
zliu41: Why use Avro schema in this class?

liyinan926: Good catch. The schema can be any type here since it's not used. I will change it to `Object`.

liyinan926:
Updated.

zliu41:
LGTM

PR #423, opened 2015-10-30
Part2 sftp copy

pcadabam:
This is part 2 of https://github.com/linkedin/gobblin/pull/403

Tasks from meeting, 
1. Use password encryption for SFTP creds and decryption passphrase - Done
2. Open ticket to explore unarchival in converter - Done
3. Rename archive..writer to Tar..writer or support other archival types in archive writer -  Done
4. Implement permission -  Done
5. Add preserver flags (group, owner, permission) to job config -  Will be done in config package
6. split pull request to Dataset, Copyable Dataset, SFTP -  Done

I have added support to handle multiple datasets in the writer and the publisher. The task staging and task output now have the complete path of the final destination under them. 
The publisher goes over each dataset in the task out put and publishes them.

##### E.g. 

Source FileSystem
/src/a/b/c /t1.txt
/src/a/b/c /t2.txt

Let's say /src/a/b/c needs to be copied to /dest/data/a/b/c,

##### After InputStreamDataWriter.write
Gobblin's task staging directory looks like,
-/tmp/gobblin/task_staging/task_Id1/dest/data/a/b/c/t1.txt
-/tmp/gobblin/task_staging/task_Id2/dest/data/a/b/c/t2.txt

##### After InputStreamDataWriter.commit
Gobblin's task output directory looks like,
-/tmp/gobblin/task_output/dest/data/a/b/c/t1.txt
-/tmp/gobblin/task_output/dest/data/a/b/c/t2.txt

##### After CopyDataPublisher.publish

Destination FileSystem looks like,
/dest/data/a/b/c/t1.txt
/dest/data/a/b/c/t2.txt

Each dataset is renamed at a time by the publisher. In this case /dest/data.


pcadabam:
@chavdar and @ibuenros please take a look

pcadabam:
@liyinan926 can you also review?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -51,10 +55,52 @@ public SftpFsHelper(State state) {
     this.state = state;
   }
 
-  public ChannelSftp getSftpConnection() {
```
ibuenros: Is it ok to change this public API?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -43,7 +47,7 @@
  */
 public class SftpFsHelper implements SizeAwareFileBasedHelper {
```
ibuenros: Can you add some explanation on  how sftp helper works? I see channels, and channelexec, but without knowing more about Sftp it's hard to understand how all of this fits together.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -51,10 +55,52 @@ public SftpFsHelper(State state) {
...
+      channelSftp = (ChannelSftp) session.openChannel("sftp");
+      channelSftp.connect();
+      channels.add(channelSftp);
+      return channelSftp;
+    } catch (JSchException e) {
+      throw new SftpException(0, "Cannot open a channel to SFTP server", e);
+    }
+  }
+
+  public ChannelExec getExecChannel(String command) throws SftpException {
```
ibuenros: javadoc

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,268 @@
...
+    } catch (SftpException e) {
+      throw new IOException(e);
+    }
+    return true;
+  }
+
+  @Override
+  public boolean delete(Path path, boolean recursive) throws IOException {
+    if (!recursive) {
+      throw new UnsupportedOperationException("Non recursive delete is not supported on this file system");
```
ibuenros: Could we instead check if the directory is empty, and only throw exception if it is nonempty?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+    return delete(path);
+  }
+
+  @Override
+  public FileStatus getFileStatus(Path path) throws IOException {
+
+    try {
+      SftpATTRS sftpAttrs = fsHelper.getSftpChannel().stat(path.toString());
+      FsPermission permission = new FsPermission((short) sftpAttrs.getPermissions());
```
ibuenros: Is this tested? There is an issue where permissions are sometimes serialized to short in base 8 and sometimes in base 10, we need to make sure sftp and hadoop are compatible in this sense.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.hadoop.fs.FileSystem;
+
+import com.google.common.collect.Lists;
+
+
+/**
+ * {@link gobblin.source.Source} that generates work units from {@link gobblin.data.management.copy.CopyableDataset}s.
+ *
+ */
+public class CopySource extends AbstractSource<String, FileAwareInputStream> {
```
ibuenros: Any changes in this class?

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/InputStreamDataWriter.java
```diff
@@ -0,0 +1,140 @@
...
+    setFilePermissions(fileAwareInputStream.getFile());
+  }
+
+  /**
+   * Sets the owner/group and permission for the file in the task staging directory
+   *
+   * @param file
+   * @throws IOException
+   */
+  protected void setFilePermissions(CopyableFile file) throws IOException {
```
ibuenros: We should also set the metadata of the parent directories if available (under ancestorsOwnerAndPermission)

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/InputStreamDataWriter.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Override
+  public long recordsWritten() {
+    return filesWritten;
+  }
+
+  @Override
+  public long bytesWritten() throws IOException {
+    return bytesWritten;
```
ibuenros: This won't work atm, will it?

gobblin-data-management/src/main/java/gobblin/data/management/util/filters/GpgPathFilter.java
```diff
@@ -0,0 +1,13 @@
+package gobblin.data.management.util.filters;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+
+
+public class GpgPathFilter implements PathFilter {
```
ibuenros: javadoc, what is this used for? why do you check for `"Log"`?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -51,10 +55,52 @@ public SftpFsHelper(State state) {
...
+      throw new SftpException(0, "Cannot open a channel to SFTP server", e);
+    }
   }
 
+  private void cleanupSftpChannels(){
+    log.info("Exiting all open channels");
+    for (Channel channel : channels) {
+      if (channel != null) {
+        if (channel instanceof ChannelSftp) {
+          ((ChannelSftp)channel).exit();
```
chavdar: 1. The javadocs say "exit" is a synonym of disconnect() so you probably don't need the cast.
2. I'd put the exit()/disconnect() in a try-catch (RuntimeException) block so that errors in on channel do not affect the others. If there was an exception thrown, you can record it and re-throw it at the end.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -96,16 +144,12 @@ public void connect()
 
       session.connect();
 
-      channelSftp = (ChannelSftp) session.openChannel("sftp");
-      channelSftp.connect();
       log.info("Finished connecting to source");
     } catch (JSchException e) {
+      cleanupSftpChannels();
```
chavdar: Why do you need to close all open channels?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -121,7 +165,7 @@ public InputStream getFileStream(String file)
       throws FileBasedHelperException {
     SftpGetMonitor monitor = new SftpGetMonitor();
     try {
-      return this.channelSftp.get(file, monitor);
+      return getSftpChannel().get(file, monitor);
```
chavdar: Do you know if the channel gets auto-closed if someone closes the inputstream? My guess is probably not. In that case, I suggest you return a wrapper InputStream that will do it since you have a dedicated channel per file. An even better solution is to have a pool of InputStreams that can be reused. My guess is that creating a new channel is an expensive operation because you need to have PK authentication.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
chavdar: I think the name is misleading. Typically, I expect a help class to be a set of static methods. This class does session management so I think it should be named SftpSession.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -43,7 +47,7 @@
  */
 public class SftpFsHelper implements SizeAwareFileBasedHelper {
   private static Logger log = LoggerFactory.getLogger(SftpFsHelper.class);
-  private ChannelSftp channelSftp;
+  Set<Channel> channels = Sets.newConcurrentHashSet();
```
chavdar: See my comments below. I think we should be using a channel pool and try to reuse open channels. At the very least these should weak references so those get collected asap instead of staying in memory and using up resources.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,268 @@
...
+    } catch (FileBasedHelperException e) {
+      throw new IOException(e);
+    }
+  }
+
+  @Override
+  public boolean mkdirs(Path path, FsPermission permission) throws IOException {
+    try {
+      fsHelper.getSftpChannel().mkdir(path.toString());
+      fsHelper.getSftpChannel().chmod((int) permission.toShort(), path.toString());
```
chavdar: Can't you reuse the above channel?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -51,10 +55,52 @@ public SftpFsHelper(State state) {
...
+  /**
+   * Create new channel every time a command needs to be executed. This is required to support execution of multiple
+   * commands in parallel. All created channels are cleaned up when the session is closed.
+   * @see #cleanupSftpChannels()
+   * @see #close()
+   *
+   * @return a new {@link ChannelSftp}
+   * @throws SftpException
+   */
+  public ChannelSftp getSftpChannel() throws SftpException {
```
chavdar: This may be the time to start switching to Java 7. I suggest that we return a wrapper CloseableChannelSftp which implements the AutoCloseable interface. That will allow us to write something like:

try (CloseableChannelSftp channel = helper.getSfpChannel()) {
  channel.foo();
} 

That will take care of the channel management.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,268 @@
...
+      InputStream is = fsHelper.getFileStream(path.toString());
+      return new FSDataInputStream(new BufferedFSInputStream(new SftpFsFileInputStream(is), bufferSize));
+    } catch (FileBasedHelperException e) {
+      throw new IOException(e);
+    }
+  }
+
+  @Override
+  public FSDataInputStream open(Path path) throws IOException {
+    return open(path, 10000);
```
chavdar: Please use a named constant. Typically, buffer sizes are powers of 2.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
chavdar: I thought we were going to bundle a FileAwareInputStream and a CopyableFile into a class to represent the record type for processing.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public static final String DEFAULT_DATASET_PROFILE_CLASS_KEY = CopyableGlobDatasetFinder.class.getCanonicalName();
```
chavdar: DATASET_PROFILE or DATASET_FINDER?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -0,0 +1,132 @@
...
+          DatasetUtils.instantiateDatasetFinder(state.getProperties(), getSourceFileSystem(state),
+              DEFAULT_DATASET_PROFILE_CLASS_KEY);
+      List<CopyableDataset> copyableDatasets = datasetFinder.findDatasets();
+      for (CopyableDataset copyableDataset : copyableDatasets) {
+
+        Collection<Partition<CopyableFile>> partitions =
+            copyableDataset.partitionFiles(copyableDataset.getCopyableFiles());
+
+        for (Partition<CopyableFile> partition : partitions) {
+          Extract extract = new Extract(Extract.TableType.SNAPSHOT_ONLY, "gobblin.copy", partition.getName());
```
chavdar: "gobblin.copy" should be a named constant.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private FileSystem getSourceFileSystem(State state) throws IOException {
+
+    Configuration conf = HadoopUtils.getConfFromState(state);
+    String uri = state.getProp(ConfigurationKeys.SOURCE_FILEBASED_FS_URI, ConfigurationKeys.LOCAL_FS_URI);
+    return FileSystem.get(URI.create(uri), conf);
+  }
+
+  public static void serializeCopyableFiles(State state, List<CopyableFile> copyableFiles) throws IOException {
+    state.appendToListProp(SERIALIZED_COPYABLE_FILE, CopyableFile.serializeList(copyableFiles));
```
chavdar: s/SERIALIZED_COPYABLE_FILE/SERIALIZED_COPYABLE_FILES/

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -0,0 +1,132 @@
...
+
+
+/**
+ * {@link gobblin.source.Source} that generates work units from {@link gobblin.data.management.copy.CopyableDataset}s.
+ *
+ */
+public class CopySource extends AbstractSource<String, FileAwareInputStream> {
+
+  public static final String DEFAULT_DATASET_PROFILE_CLASS_KEY = CopyableGlobDatasetFinder.class.getCanonicalName();
+  public static final String SERIALIZED_COPYABLE_FILE = "gobblin.copy.serialized.copyable.file";
```
chavdar: s/gobblin.copy.serialized.copyable.file/gobblin.copy.serialized.copyable.files/

gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
```diff
@@ -0,0 +1,34 @@
...
+
+
+/**
+ * A wrapper to {@link InputStream} that represents an entity to be copied. The enclosed {@link CopyableFile} instance
+ * contains file Metadata like permission, destination path etc. required by the writers and converters.
+ */
+@AllArgsConstructor
+@Getter
+@Setter
+public class FileAwareInputStream {
```
chavdar: Object should be immutable.

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DecryptConverter.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public Iterable<FileAwareInputStream> convertRecord(String outputSchema, FileAwareInputStream fileAwareInputStream,
+      WorkUnitState workUnit) throws DataConversionException {
+
+    try {
+      fileAwareInputStream.setInputStream(GPGFileDecrypter.decryptFile(fileAwareInputStream.getInputStream(),
+          passphrase));
+      return new SingleRecordIterable<FileAwareInputStream>(fileAwareInputStream);
```
chavdar: See comment about about immutability of FileAwareInputStream. You should create a new instance of FileAwareInputStream.

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/InputStreamExtractor.java
```diff
@@ -0,0 +1,84 @@
...
+    this.fileIterator = filesIterator;
+  }
+
+  /**
+   * @return Constant string schema.
+   * @throws IOException
+   */
+  @Override
+  public String getSchema() throws IOException {
+    return "InputStream";
```
chavdar: Please use a named constant. I suggest using FileAwareInputStream.class.getName()

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/InputStreamExtractor.java
```diff
@@ -0,0 +1,84 @@
...
+ * An implementation of {@link Extractor} that extracts {@link InputStream}s. This extractor is suitable for copy jobs
+ * where files from any source to a sink. The extractor extracts a {@link FileAwareInputStream} which encompasses an
+ * {@link InputStream} and a {@link CopyableFile} for every file that needs to be copied.
+ *
+ * <p>
+ * In Gobblin {@link Extractor} terms, each {@link FileAwareInputStream} is a record. i.e one record per copyable file.
+ * The extractor is capable of extracting multiple files
+ * <p>
+ */
+public class InputStreamExtractor implements Extractor<String, FileAwareInputStream> {
```
chavdar: I think FileAwareInputStreamExtractor is a more specific name.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+
+
+/**
+ * A {@link DataPublisher} to {@link CopyableFile}s from task output to final destination
```
chavdar: Please document what state the publisher expects.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -0,0 +1,123 @@
...
+    for (Path datasetDestinationRoot : datasetRoots.keySet()) {
+      this.publishDataset(datasetDestinationRoot, datasetRoots.get(datasetDestinationRoot));
+    }
+
+  }
+
+  /**
+   * Create a multimap that maps a dataset destination path to all workunits that belong to this dataset.
+   */
+  private Multimap<Path, WorkUnitState> getDatasetRoots(Collection<? extends WorkUnitState> states) throws IOException {
```
chavdar: Why aren't the datasets explicitly serialized in the state? Then we don't have to have this logic to try to "infer" the roots. Further, the current API looks unusual. Why does the method return a multimap, when all that matters are the keys. Why not return a Dataset set?

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/UnArchiveConverter.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.InputStream;
+import java.util.zip.GZIPInputStream;
+
+
+/**
+ * A {@link Converter} that converts an archived {@link InputStream} to a tar {@link InputStream}. Wraps the given
+ * archived {@link InputStream} with {@link GZIPInputStream} Use this converter if the {@link InputStream} from source
+ * is compressed.
+ */
+public class UnArchiveConverter extends Converter<String, String, FileAwareInputStream, FileAwareInputStream> {
```
chavdar: This looks more like an UnGzipConverter or InflaterConverter, not an UnArchiveConverter. Can you also add support for both gzip and Zip to make it more generic? For example, you can add a config option to specify the type of inflater algorithm:
* GZIP
* ZIP
*AUTO_BY_NAME (default)

AUTO_BY_NAME can check the extension and use GZIP for .gz and .tgz and ZIP for .zip.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/TarArchiveInputStreamDataWriter.java
```diff
@@ -0,0 +1,120 @@
...
+          firstEntrySeen = true;
+        }
+
+        log.info("Unarchiving " + tarEntryStagingPath);
+
+        if (tarEntry.isDirectory() && !fs.exists(tarEntryStagingPath)) {
+          fs.mkdirs(tarEntryStagingPath);
+        } else if (!tarEntry.isDirectory()){
+
+          FSDataOutputStream out = fs.create(tarEntryStagingPath, true);
```
chavdar: You should use https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FSDataInputStream.html#read%28org.apache.hadoop.io.ByteBufferPool,%20int%29 and https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Channels.html#newChannel(java.io.OutputStream) to avoid unnecessary copy of data to the heap.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: added the old API and deprecated it.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Done

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Looked at jsch code base. The framework keeps track of all the open channels and manages closing all channels for a session when session.disconnect is called. We do not need to manage this.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Look at previous comment. We do not need this because jsch already does it.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: 1. Channel does NOT get closed when inputStream is closed. when inputStream#close is called, jsch only sends a packet with close bytes to mark this inputStream as closed.
2. It does not seem like channel creation is expensive (from my basic code scan or JSCH.) Session creating is expensive as auth happens at session level.
Channels get cleaned up when session is closed. In MR mode, each mapper creates a session and a channel for ever file it opens. I do not think we expect thousands of files per mapper.
In this case is number of channels still a concern?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: I think it is better to either completely implement is or throw UnsupportOperationException. Otherwise it confuses the caller as it works for "some" cases.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
pcadabam: Logic is mostly same.
- A workunit can handle multiple CopyableFiles
- Replace casting of dataset and copyableFile with generics

gobblin-data-management/src/main/java/gobblin/data/management/util/filters/GpgPathFilter.java
pcadabam: I used Log prefix for testing. Will remove it.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
ibuenros: I believe what I describe is the behavior of the HDFS filesystem (except it probably doesn't throw exception and instead just returns false)

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: I have added some javadoc for channelExec and channelSftp
@chavdar this class already exists in gobblin. I have added a couple of methods to the SftpFsHelper to get channels.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Channel's API does not provide a way to say if it is being used or not. 
Channel#isClosed(), Channel#isConnected() are the only two API's provided. 
Now if we want to reuse a channel we have to call Channel#disconnect() after every action so the next action can re-connect and use it. Since connection and disconnection is the 'heavier' operation I do not see value in reusing. 
If garbage collection the concern? If so I can use pool.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: Done. Also reused the channel where possible

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Right now I am explicitly calling channel.disconnect after I finish using a channel. Yes TryWithResource is the ideal way. I would want to do the java upgrade in a different PR.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: Done. Changed it 8192

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
pcadabam: Currently both are used. @ibuenros can you comment?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
ibuenros: Dataset profile was intended to include dataset finder + default configurations for specific types of dataset (e.g. DatabaseSnapshot, or TrackingData). With the whole configuration management stuff, it might be less relevant. As features evolved (e.g. around IDPC and Kafka) they became almost synonymous.
Since the cleaner uses "profile.class", I think we should stick with profile for configuration keys.

gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
pcadabam: Done. Removed setters

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DecryptConverter.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/InputStreamExtractor.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/InputStreamExtractor.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
pcadabam: Done. Added in the javadoc for the constructor

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
chavdar: I don't see a DatasetProfile class. What is the property supposed to contain? Currently, it is contains a DatasetFinder implementation.

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/UnArchiveConverter.java
pcadabam: I have made the UnArchiveConverter to support both gzip and zip streams. I think we can avoid another config key. I have implemented an UnzipConverter and UnGzipConverter. If the user wants to specify the inflater alogorithm he can do so by specifying the right converter.
The UnArchiveConverter will look at the extension and call UnZipConverter/UnGzipConverter by extension.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
pcadabam: Done. Implemented dataset serialization

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/TarArchiveInputStreamDataWriter.java
pcadabam: Done. I have used {@link ReadableByteChannel} and {@link WritableByteChannel} to copy the bytes. 
See StreamUtils#copy

pcadabam:
I have updated the PR with all the changes discussed. @chavdar  and @ibuenros  please take a look.

**Major changes** -

1. Implement a way to serialize CopyableDataset in the Source and deserialize them in the publisher
2. Implement setting ancestor permissions and children permission in the writer
3. Sftp Channel management
  * Disconnect channel after use
  * Reuse channel where multiple commands are executed
  * Add a hook to InputStream to disconnect channel when InputStream#close() is called
  * Remove storing all channels in a set locally as jsch already does this internally

4. Use java nio channles to copy bytes from InputStream to OutputStream in the Writers

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
chavdar: In that case, can you return CloseableChannelSftp which extends ChannelSftp and implements Closeable .

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,270 @@
...
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
+  /**
+   * A {@link SeekableFSInputStream} that holds a handle on the Sftp {@link Channel} used to open the
+   * {@link InputStream}. The {@link Channel} is disconnected when {@link InputStream#close()} is called.
+   */
+  private class SftpFsFileInputStream extends SeekableFSInputStream {
+
+    private Channel channel;
```
chavdar: final

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,270 @@
...
+
+    public SftpFsFileInputStream(InputStream in) {
```
chavdar: why is this needed? It will cause the close() method to throw an NPE for this.channel.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
```diff
@@ -0,0 +1,270 @@
...
+  public FSDataOutputStream create(Path arg0, FsPermission arg1, boolean arg2, int arg3, short arg4, long arg5,
+      Progressable arg6) throws IOException {
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
+  /**
+   * A {@link SeekableFSInputStream} that holds a handle on the Sftp {@link Channel} used to open the
+   * {@link InputStream}. The {@link Channel} is disconnected when {@link InputStream#close()} is called.
+   */
+  private class SftpFsFileInputStream extends SeekableFSInputStream {
```
chavdar: This class should probably be static. It doesn't seem to use anything from SftpLightWeightFileSystem .

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/UnArchiveConverter.java
chavdar: I still think UnArchiveConverter is too generic and misleading. It certainly does not handle tar (tape archive) files.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/InputStreamDataWriter.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.hadoop.io.IOUtils;
+
+import com.google.common.io.Closer;
+
+
+/**
+ * A {@link DataWriter} to write {@link FileAwareInputStream}
+ */
+@Slf4j
+public class InputStreamDataWriter implements DataWriter<FileAwareInputStream> {
```
chavdar: This should be named FileAwareInputStreamDataWriter

gobblin-data-management/src/main/java/gobblin/data/management/util/filters/GpgPathFilter.java
```diff
@@ -0,0 +1,27 @@
...
+
+package gobblin.data.management.util.filters;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+
+/**
+ * A {@link PathFilter} to accept encrypted tarball files.
+ */
+public class GpgPathFilter implements PathFilter {
```
chavdar: This is not a generic class and should not be in gobblin-data-management.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: Done

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: Done

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpLightWeightFileSystem.java
pcadabam: Removed this constructor

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/InputStreamDataWriter.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/util/filters/GpgPathFilter.java
pcadabam: Moved it to gobblin-utility

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/UnArchiveConverter.java
pcadabam: I think we should get rid of UnArchiveConverter and punt/implement it when we know how to handle other archive types. We should have UnGzipConverter that handles tar.gz and .tgz files.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: As discussed offline this is not possible without delegating all the methods of ChannelSftp

pcadabam:
@chavdar I addressed all comments from 11/05. Please take a look

PR #428, opened 2015-11-03
Remove duplicate closer registration

zliu41:
It was added in 7598696 since there was a warning in eclipse that writer is never closed. But the created writer is registered again in line 70 which causes the writer to be closed twice (which will fail).


zliu41:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #430, opened 2015-11-04
Findbugs fix on CliLocalJobLauncher

ibuenros:


ibuenros:

PR #432, opened 2015-11-05
Switched to use helix 0.6.x for Gobblin on Yarn

liyinan926:
Main changes:

1. Switched to Helix `0.6.6-SNAPSHOT`.
2. Added implementation of `GobblinHelixJobLauncher.executeCancellation`.
3. Other fixes and code refactorings.


liyinan926:
@zliu41 can you review?

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -487,8 +487,10 @@ private static void cleanupStagingDataPerTask(JobState jobState) {
     Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();
     try {
       for (TaskState taskState : jobState.getTaskStates()) {
+        TaskState fatTaskState = new TaskState(taskState);
+        fatTaskState.addAllIfNotExist(jobState);
```
zliu41: This could be slow if the number of tasks is large. Can you just add the properties needed?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -111,12 +114,35 @@ public GobblinHelixJobLauncher(Properties jobProps, HelixManager helixManager, P
...
+    } finally {
+      super.close();
+    }
+  }
+
+  @Override
   protected void runWorkUnits(List<WorkUnit> workUnits) throws Exception {
     try {
       submitJobToHelix(createJob(workUnits));
+      LOGGER.info(String.format("Submitted job %s to Helix", this.jobContext.getJobId()));
```
zliu41: Should this line and the next line be synchronized?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -111,12 +114,35 @@ public GobblinHelixJobLauncher(Properties jobProps, HelixManager helixManager, P
   }
 
   @Override
+  public void close() throws IOException {
+    try {
+      if (this.jobSubmitted && !this.jobComplete) {
```
zliu41: You can call `executeCancellation` here

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
```diff
@@ -121,28 +122,34 @@ public TaskResult run() {
...
   @Override
   public void cancel() {
-    LOGGER.info(String.format("Cancelling WorkUnit %s of job %s", this.taskId, this.jobId));
+    LOGGER.info(String.format("Cancelling task %s of job %s", this.taskId, this.jobId));
     this.futureOptional.transform(new Function<Future<?>, Object>() {
       @Nullable
       @Override
       public Object apply(Future<?> input) {
-        return input.cancel(true);
+        return input.cancel(false);
```
zliu41: what's the reason of this change?

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
liyinan926: We are doing the same in `MRJobLauncher`. I will try to fix both places.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
liyinan926: To allow already-started tasks to finish gracefully.

zliu41: If an `InterruptedException` is caught in line 133, shouldn't the task be cancelled asap? Also what about hanging tasks?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: No, this method will only be called from a single thread.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
liyinan926: Fixed.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
liyinan926: It all depends on if the task responds to interruption or not. But you are right, if an InterruptedException is caught in line 133, then cancellation should be immediate. I will change it back.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
liyinan926: Actually seems there's a bunch of config properties needed so it might be a good idea to keep the logic here instead of iterating through every possible config property. I think a better approach would be to put `fatTaskState.addAllIfNotExist(jobState);` into the `ParallelRunner`.

liyinan926: It seems it's pretty tricky to add this into `ParallelRunner`.  I think this should be fine because this is entirely in memory, so it should be much less significant than deleting files from HDFS.

liyinan926:
@zliu41 Updated  d0774f6.

zliu41:
LGTM

PR #434, opened 2015-11-05
Retrofit writers to work with partitioning

zliu41:
@ibuenros @liyinan926 can you review?


gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
```diff
@@ -25,11 +27,17 @@
  */
 public abstract class DataWriterBuilder<S, D> {
 
+  @Getter
```
ibuenros: Is there a reason for a public Getter?

gobblin-core/src/main/java/gobblin/serde/HiveSerDeWrapper.java
```diff
@@ -48,33 +47,27 @@
...
-    AVRO(AvroSerDe.class.getName(), AvroContainerInputFormat.class.getName(), AvroContainerOutputFormat.class.getName(),
-        WriterOutputFormat.AVRO.getExtension()),
-    ORC(OrcSerde.class.getName(), OrcInputFormat.class.getName(), OrcOutputFormat.class.getName(),
-        WriterOutputFormat.ORC.getExtension()),
+    AVRO(AvroSerDe.class.getName(), AvroContainerInputFormat.class.getName(),
+        AvroContainerOutputFormat.class.getName()),
+    ORC(OrcSerde.class.getName(), OrcInputFormat.class.getName(), OrcOutputFormat.class.getName()),
     PARQUET(ParquetHiveSerDe.class.getName(), MapredParquetInputFormat.class.getName(),
-        MapredParquetOutputFormat.class.getName(), WriterOutputFormat.PARQUET.getExtension());
+        MapredParquetOutputFormat.class.getName());
```
ibuenros: What are the changes in this file? Why was the extension removed?

gobblin-core/src/main/java/gobblin/writer/FsDataWriterBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+
+  protected String getPartitionedFileName(State properties, String originalFileName) {
+    boolean includePartitionerFieldNames = properties.getPropAsBoolean(
+        ForkOperatorUtils.getPropertyNameForBranch(WRITER_INCLUDE_PARTITION_IN_FILE_NAMES, this.branches, this.branch),
+        false);
+    boolean removeIllegalTokens = properties.getPropAsBoolean(ForkOperatorUtils
+        .getPropertyNameForBranch(WRITER_REPLACE_PATH_SEPARATORS_IN_PARTITIONS, this.branches, this.branch), false);
+
+    return new Path(
+        AvroUtils.serializeAsPath(this.partition.get(), includePartitionerFieldNames, removeIllegalTokens).toString(),
```
ibuenros: Can you rename removeIllegalTokens for removePathSeparators?

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
```diff
@@ -0,0 +1,109 @@
...
+  public static final String WRITER_PARTITION_PATTERN = ConfigurationKeys.WRITER_PREFIX + ".partition.pattern";
+  public static final String DEFAULT_WRITER_PARTITION_PATTERN = "yyyy/MM/dd/HH";
+  public static final String WRITER_PARTITION_TIMEZONE = ConfigurationKeys.WRITER_PREFIX + ".partition.timezone";
+  public static final String DEFAULT_WRITER_PARTITION_TIMEZONE = "America/Los_Angeles";
+
+  public static final String PARTITIONED_PATH = "partitionedPath";
+  public static final String PREFIX = "prefix";
+  public static final String SUFFIX = "suffix";
+
+  private static final Schema SCHEMA = SchemaBuilder.record("GenericRecordTimePartition").namespace("gobblin.writer")
```
ibuenros: The schema should be dynamic, e.g. if prefix is not set, then the schema should not contain a prefix field (so that no token gets generated), same for the suffix.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * A {@link WriterPartitioner} that partitions a record based on a timestamp and a {@link DateTimeFormat}.
+ *
+ * A prefix and a suffix can be added to the formatted timestamp, e.g., the partition path can be
+ * 'prefix/2015/11/05/suffix'.
+ *
+ * @author ziliu
+ */
+public abstract class TimeBasedWriterPartitioner<D> implements WriterPartitioner<D> {
```
ibuenros: Suggestion:
I think passing in the pattern is pointless, as it will always be of the form yyyy/MM/... until a certain granularity (e.g. no one reasonable would put month first). Instead, users can just set the granularity of their partitioner (i.e. day, hour, minute).
The schema should be created dynamically, so if  a user sets granularity=hour, then the schema would be `{'year':int, 'month':int, 'day':int, 'hour':int}` (possibly plus prefix and suffix). Then the `PartitionedWriter` will construct the path and be able to for example add field names correctly.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -71,41 +74,42 @@ public PartitionedDataWriter(DataWriterBuilder<S, D> builder, final State state)
       try {
         this.shouldPartition = true;
         this.builder = Optional.of(PartitionAwareDataWriterBuilder.class.cast(builder));
-        this.partitioner = Optional.of(
-            WriterPartitioner.class.cast(ConstructorUtils.invokeConstructor(
-                    Class.forName(state.getProp(ConfigurationKeys.WRITER_PARTITIONER_CLASS)), state)));
+        this.partitioner = Optional.of(WriterPartitioner.class.cast(
```
ibuenros: Can you update the wiki with the new signature for the constructor?

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
zliu41: The constructor signature of some writers are changed from `(State properties, String fileName, Schema schema, int numBranches, int branchId)` to `(FsDataWriterBuilder<?, D> builder, State properties)`. They now need to obtain `fileName, schema, numBranches, branchId` from the builder. It is currently fine since those writers and builders are in the same package, but it is possible that in the future a writer is added in a different package.

gobblin-core/src/main/java/gobblin/serde/HiveSerDeWrapper.java
zliu41: Extensions can be specified using `writer.output.format`. This is used by `FsDataWriter` to construct the file names (if not exists it will default to avro), so one should use this property rather than specifying an extension with the serde.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
zliu41: I think this way is more flexible, since someone may want `yyyy-mm/DD/hh`, `yyyy/mm-DD-hh`, etc.

gobblin-core/src/main/java/gobblin/writer/FsDataWriterBuilder.java
zliu41: yes, was a typo

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
zliu41: I've made `AvroUtils.serializeAsPath()` ignore empty field values if `includeFieldNames=false`, which achieves the same thing.

ibuenros: But if includeFieldNames=true, then the user would see `prefix=/partitionedPath=2015/10/01/suffix=` which is probably not what they want.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
ibuenros: My concern is what if the user wants the path to be `year=2015/month=10/day=1`? This is a common pattern, and java time formats are not very friendly with constant text.

zliu41: makes sense, I'll update to support both options

zliu41:
@ibuenros updated f771eb2

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
zliu41: will do

ibuenros:
LGTM other than a couple comments.

zliu41:
Updated. @liyinan926 please have a look.

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
sahilTakiar: You can add the Getter annotation to the class declaration and it will be applied to all instance variables in the class. Saves the time of adding the annotation for each individual instance variable.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
```diff
@@ -0,0 +1,219 @@
...
+  public static final String WRITER_PARTITION_TIMEZONE = ConfigurationKeys.WRITER_PREFIX + ".partition.timezone";
+  public static final String DEFAULT_WRITER_PARTITION_TIMEZONE = ConfigurationKeys.PST_TIMEZONE_NAME;
+  public static final String WRITER_PARTITION_GRANULARITY = ConfigurationKeys.WRITER_PREFIX + ".partition.granularity";
+  public static final String DEFAULT_WRITER_PARTITION_GRANULARITY = Granularity.HOUR.toString();
+
+  public static final String PARTITIONED_PATH = "partitionedPath";
+  public static final String PREFIX = "prefix";
+  public static final String SUFFIX = "suffix";
+
+  public enum Granularity {
```
liyinan926: Instead of defining a new enum, is it possible to just use `TimeUnit`?

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  private String getWriterPartitionSuffix(State state, int numBranches, int branchId) {
+    String propName = ForkOperatorUtils.getPropertyNameForBranch(WRITER_PARTITION_SUFFIX, numBranches, branchId);
+    return state.getProp(propName, StringUtils.EMPTY);
+  }
+
+  private Granularity getGranularity(State state, int numBranches, int branchId) {
+    String propName = ForkOperatorUtils.getPropertyNameForBranch(WRITER_PARTITION_GRANULARITY, numBranches, branchId);
+    String granularityValue = state.getProp(propName, DEFAULT_WRITER_PARTITION_GRANULARITY);
```
liyinan926: I suggest instead of using a string form `DEFAULT_WRITER_PARTITION_GRANULARITY`, use the corresponding enum value instead. This way, you don't need the precondition check below.

gobblin-core/src/main/java/gobblin/writer/FsDataWriterBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+import gobblin.util.WriterUtils;
+
+
+/**
+ * A abstract {@link DataWriterBuilder} for building {@link DataWriter}s that write to
+ * {@link org.apache.hadoop.fs.FileSystem}s.
+ *
+ * @author ziliu
+ */
+public abstract class FsDataWriterBuilder<S, D> extends PartitionAwareDataWriterBuilder<S, D> {
```
liyinan926: Missing `@param <S>` and `@param <d>`.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
zliu41: unfortunately `TimeUnit` does not have month or year. What we need is a calendar unit rather than time unit.

gobblin-core/src/main/java/gobblin/writer/partitioner/TimeBasedWriterPartitioner.java
zliu41: changed to enum value, but the precondition check is still needed, since the user may specify an invalid value, e.g., "week", which deserves an exception.

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
zliu41: fixed.

zliu41:
Updated.

PR #437, opened 2015-11-05
Trash handles deletes of files already existing in trash correctly.

ibuenros:


ibuenros:
@chavdar simple PR for fix of deleting files already existing in trash.

chavdar:
LGTM

PR #440, opened 2015-11-10
Fixed Gobblin Yarn to be able to run MultiWorkUnits

liyinan926:
Originally Gobblin on Yarn was written with the assumption that we would rely on Helix to group/pack `WorkUnit`s into groups and assign them properly to the available containers. So the code originally flattened the `MultiWorkUnit`s. Later it was realized this would not be too practical since Helix does not have the knowledge of loads of individual `WorkUnit`s. In addition, the Gobblin Kafka adapter is already doing bin packing so we decided to not override that logic and changed the code to not flatten `MultiWorkUnit`s. However, the parts that actually run the job/tasks are not updated accordingly to work with `MultiWorkUnit`s. This PR fixed it. 

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -19,6 +19,7 @@ import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,6 +34,7 @@ import com.google.common.io.Closer;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
+import gobblin.metastore.StateStore;
 import gobblin.metrics.GobblinMetrics;
 import gobblin.metrics.GobblinMetricsRegistry;
 import gobblin.metrics.MetricContext;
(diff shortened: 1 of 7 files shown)


liyinan926:
@zliu41 can you review?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
```diff
@@ -98,82 +93,97 @@ public GobblinHelixTask(TaskCallbackContext taskCallbackContext, TaskExecutor ta
...
-      switch (workingState) {
-        case SUCCESSFUL:
-          return new TaskResult(TaskResult.Status.COMPLETED, "task id: " + this.taskId);
-        case FAILED:
-          return new TaskResult(TaskResult.Status.ERROR, "task id: " + this.taskId);
-        case CANCELLED:
-          return new TaskResult(TaskResult.Status.CANCELED, "task id: " + this.taskId);
-        default:
-          throw new IllegalStateException("Unexpected result task state: " + workingState);
+      Path workUnitFilePath =
```
sahilTakiar: A lot of the code below seems similar to what is done in `MRJobLauncher.TaskRunner` anyway we can abstract the code to a new class so we can re-use it for both MR jobs and YARN jobs?

liyinan926: Sure. Will do.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTask.java
```diff
@@ -98,82 +93,97 @@ public GobblinHelixTask(TaskCallbackContext taskCallbackContext, TaskExecutor ta
...
+      runWorkUnits(workUnits);
+
+      return new TaskResult(TaskResult.Status.COMPLETED, String.format("completed tasks: %d", workUnits.size()));
     } catch (IOException ioe) {
-      LOGGER.error("Failed to build task " + this.taskId, ioe);
-      return new TaskResult(TaskResult.Status.ERROR, "task id: " + this.taskId);
-    } catch (ExecutionException ee) {
-      LOGGER.error("Failed to run task " + this.taskId, ee);
-      return new TaskResult(TaskResult.Status.ERROR, "task id: " + this.taskId);
+      return new TaskResult(TaskResult.Status.ERROR, Throwables.getStackTraceAsString(ioe));
```
sahilTakiar: Is it necessary to have a catch for `IOException` as it is handled in the same was as `Throwable`

liyinan926: Good catch. It used to need this but no longer the case now.

liyinan926:
@sahilTakiar Updated.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -520,6 +482,117 @@ private void cleanupStagingData(JobState jobState) {
...
+   * @param jobId the job ID
+   * @param workUnits the given list of {@link WorkUnit}s to submit to run
+   * @param taskStateTracker a {@link TaskStateTracker} for task state tracking
+   * @param taskExecutor a {@link TaskExecutor} for task execution
+   * @param taskStateStore a {@link StateStore} for storing {@link TaskState}s
+   * @param logger a {@link Logger} for logging
+   * @throws IOException if there's something wrong with any IO operations
+   * @throws InterruptedException if the task execution gets cancelled
+   */
+  public static void runWorkUnits(String jobId, String containerId, List<WorkUnit> workUnits,
```
sahilTakiar: Any reason to make this method `static`? It looks like any class that is using this method will extend `AbstractJobLauncher` anyway. So we can just make this `protected void runWorkUnits(...)`.

Same for the method below.

liyinan926: Actually no. `GobblinHelixTask` and `MRJobLauncher.TaskRunner` use it and both of them do not extend `AbstractJobLauncher`.

sahilTakiar:
LGTM

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -520,6 +482,117 @@ private void cleanupStagingData(JobState jobState) {
...
+      TaskStateTracker taskStateTracker,  TaskExecutor taskExecutor, StateStore<TaskState> taskStateStore,
+      Logger logger) throws IOException, InterruptedException {
+
+    if (workUnits.isEmpty()) {
+      logger.warn("No work units to run in container " + containerId);
```
zliu41: You are using both `LOG` and `logger` in this class.

liyinan926: This is intentional so the logger specific to the classes that call this method can be used.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -520,6 +482,117 @@ private void cleanupStagingData(JobState jobState) {
...
+   * </p>
+   *
+   * @param jobId the job ID
+   * @param workUnits the given list of {@link WorkUnit}s to submit to run
+   * @param stateTracker a {@link TaskStateTracker} for task state tracking
+   * @param taskExecutor a {@link TaskExecutor} for task execution
+   * @param countDownLatch a {@link java.util.concurrent.CountDownLatch} waited on for job completion
+   * @return a list of {@link Task}s from the {@link WorkUnit}s
+   */
+  public static List<Task> runWorkUnits(String jobId, List<WorkUnit> workUnits, TaskStateTracker stateTracker,
```
zliu41: This method can be private

liyinan926: This method is called directly in `LocalJobLauncher`.

PR #441, opened 2015-11-11
Improve topic isolation in MRCompactor

zliu41:
Previously if the MR compaction job for a topic fails it won't affect other topics, but if it fails when creating job properties for a topic or submitting the topic to the job executor, the whole job will fail immediately and no other topic will be processed.

This PR fixes it -- in the above cases the topic will be skipped without affecting other topics.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
@@ -13,7 +13,9 @@
 package gobblin.compaction;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.hadoop.fs.Path;
 
@@ -104,11 +106,11 @@ public class Dataset implements Comparable<Dataset> {
   private final Path outputLatePath;
   private final Path outputTmpPath;
   private final List<Path> additionalInputPaths;
-  private final List<Throwable> throwables;
+  private final Collection<Throwable> throwables;
 
   private State jobProps;
   private double priority;
-  private DatasetState state;
+  private AtomicReference<DatasetState> state;
 
   private Dataset(Builder builder) {
     this.topic = builder.topic;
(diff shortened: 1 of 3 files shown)


zliu41:
@liyinan926 can you review?

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
```diff
@@ -237,10 +237,22 @@ public void setJobProps(State jobProps) {
...
+   */
   public synchronized void addThrowable(Throwable t) {
     this.throwables.add(t);
   }
 
+  /**
+   * Skip the {@link Dataset} by setting its {@link DatasetState} to {@link DatasetState#COMPACTION_COMPLETE},
+   * and record the given {@link Throwable} in the {@link Dataset}.
+   */
+  public void skip(Throwable t) {
```
liyinan926: Seems you use `synchronized` in `addThrowable`, but not here. If you do need synchronization, I suggest you change `this.throwables` to a `ConcurrentHashSet` or if you really need a `List`, to something like `Collections.synchronizedList`.

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
```diff
@@ -237,10 +237,22 @@ public void setJobProps(State jobProps) {
...
+    this.state = DatasetState.COMPACTION_COMPLETE;
```
liyinan926: Does `this.state` also needs to be guarded? If so, you probably should use a `AtomicReference` instead.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -636,7 +646,9 @@ private int getNumDatasetsVerifiedTogether() {
 
   private void setAllDatasetStatesToVerified() {
     for (Dataset dataset : this.datasets) {
-      dataset.setState(VERIFIED);
+      if (dataset.state() == UNVERIFIED) {
```
liyinan926: This is another good indicator that you actually need to use a `AtomicReference` if you need to guard `state` since this check-then-act can be done in a single function call.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -714,12 +726,17 @@ private void submitCompactionJobsAndWaitForCompletion() throws IOException {
...
-      jobRunner.proceed();
+
+    try {
+      MRCompactorJobRunner jobRunner = getMRCompactorJobRunner(dataset, dataset.priority());
+      this.jobRunnables.put(dataset, jobRunner);
+      if (proceed) {
+        jobRunner.proceed();
+      }
+      this.jobExecutor.execute(jobRunner);
+    } catch (Throwable t) {
```
liyinan926: Do you also need to log the `Throwable`?

zliu41: No need to log here, all throwables will be printed together at the end of the whole job.

zliu41:
Addressed comments.

PR #443, opened 2015-11-12
Add option to use prime number of reducers in compaction

zliu41:
Add an option to use a prime number of reducers for compaction, which is less subject to hashcode skew.

Also changed the default data size a reducers processes from 256M to 512M, which brings the number of reducers a job uses more in-line with Camus.


zliu41:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #445, opened 2015-11-12
Change RowLevelPolicyChecker.executePolicies to return false if type=OPTIONAL and record failed

zliu41:
@liyinan926 can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
+++ b/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
@@ -18,6 +18,8 @@ import java.util.List;
 
 import org.apache.hadoop.fs.Path;
 
+import com.google.common.io.Closer;
+
 import gobblin.configuration.State;
 import gobblin.util.FinalState;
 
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
```diff
@@ -34,24 +34,23 @@ public RowLevelPolicyChecker(List<RowLevelPolicy> list) {
...
-          this.writer.open(new Path(p.getErrFileLocation(), p.toString().replaceAll("\\.", "-") + ".err"));
-          this.writer.write(record);
-        } else {
-          this.writer.write(record);
+      if (result.equals(RowLevelPolicy.Result.FAILED)) {
+        if (p.getType().equals(RowLevelPolicy.Type.FAIL)) {
+          throw new RuntimeException("RowLevelPolicy " + p + " failed on record " + record);
+        } else if (p.getType().equals(RowLevelPolicy.Type.ERR_FILE)) {
+          if (!errFileOpen) {
+            this.writer.open(new Path(p.getErrFileLocation(), p.toString().replaceAll("\\.", "-") + ".err"));
```
liyinan926: We should make `RowLevelErrFileWriter` a `Closeable` and use a `Closer` to close it here.

zliu41:
Updated.

PR #446, opened 2015-11-13
Permission fixes for distcp-ng

pcadabam:
Fixes -
1. Ancestor permissions were set on final destination path instead of staging path
2. Always grant execute permissions to directories created in staging directory to be able to publish them.

@sahilTakiar can you review this?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
@@ -26,6 +26,7 @@ import gobblin.writer.DataWriter;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.Collections;
 import java.util.List;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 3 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -155,9 +158,39 @@ private void setPathPermission(Path path, OwnerAndPermission ownerAndPermission)
...
+   * directory. The publisher needs it to publish it to the final directory and list files under this directory.
+   */
+  private OwnerAndPermission addExecutePermissionsIfRequired(FileStatus file, OwnerAndPermission ownerAndPermission) {
+
+    if (!file.isDir()) {
+      return ownerAndPermission;
+    }
+    FsAction newOwnerAction = ownerAndPermission.getFsPermission().getUserAction();
+
+    if (ownerAndPermission.getFsPermission().getUserAction() == FsAction.READ) {
```
sahilTakiar: use a `switch...case` statement instead, and throw `IllegalStateException` if it is anything but `READ`, `WRITE`, `READ_WRITE`.

sahilTakiar:
Minor comment otherwise LGTM.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
pcadabam: Changed it to enum. We dont need to throw exception as we retain the permission if it is other than these. Which means the permissions already have execute.

PR #447, opened 2015-11-13
Removed dependency on commons-lang3

liyinan926:
Seems `commons-lang3` was added just for one use in `gobblin-runtime`: to use the method `ExceptionUtils.getStackTrace()`. Replaced this usage with `Throwables.getStackTraceAsString()` so the dependency on `commons-lang3` can be removed.


sahilTakiar:
LGTM

PR #448, opened 2015-11-13
ETL-3797: Add option to modify property in state store file.

ydai1124:
Tested locally.
@sahilTakiar  Can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
@@ -28,9 +28,17 @@ import org.apache.commons.cli.Option;
 import org.apache.commons.cli.OptionBuilder;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.io.SequenceFile;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.compress.DefaultCodec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.common.io.Closer;
 import com.google.gson.stream.JsonWriter;
 
@@ -38,6 +46,7 @@ import gobblin.configuration.ConfigurationKeys;
 import gobblin.metastore.StateStore;
 import gobblin.runtime.FsDatasetStateStore;
 import gobblin.runtime.JobState;
+import gobblin.runtime.TaskState;
 
 
 /**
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
```diff
@@ -87,6 +98,64 @@ public void convert(String jobName, String jobId, Writer writer)
...
+   */
+  public void modify(String jobName, String key, String newValue) throws IOException {
+    this.modify(jobName, "current", key, newValue);
+  }
+  
+  /**
+   * Replace the original state file with a new state.
+   */
+  private void replaceStateStore(String jobName, String jobId, JobState newState) throws IOException {
+    Path storePath = new Path(storeUrl);
```
liyinan926: You should try to use the `StateStore` API instead. Even though the current implementation is based on SequenceFile, it may change in the future.

gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
```diff
@@ -87,6 +98,64 @@ public void convert(String jobName, String jobId, Writer writer)
...
+  /**
+   * Modify the configuration value of a given job instance.
+   */
+  public void modify(String jobName, String jobId, String key, String newValue) throws IOException {
+    List<? extends JobState> jobStates = this.jobStateStore.getAll(jobName, jobId + JOB_STATE_STORE_TABLE_SUFFIX);
+    if (jobStates.isEmpty()) {
+      LOGGER.warn(String.format("No job state found for job with name %s and id %s", jobName, jobId));
+      return;
+    }
+    JobState jobState = jobStates.get(0);
```
sahilTakiar: Why not set the property in all `JobState`s that match the specified `jobName` and `jobId`.

sahilTakiar:
@liyinan926 if we want to make a `State` object in the `StateStore` mutable, should we not just add some more methods to the `StateStore` interface for modifying specific `State` objects?

gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
ydai1124: Added `modifyAll` to convert all past 'JobState's`that have the same `jobName`. Isn't it the case that a pair of `jobName` and `jobId` maps to only one `JobState`?

gobblin-runtime/src/main/java/gobblin/runtime/util/JobStateToJsonConverter.java
ydai1124: Change this part to use `FsDatasetStateStore.put()`.

ydai1124:
Updated. Tested locally.

chavdar:
@sahilTakiar @liyinan926 @ydai1124 what is the state of this PR?

ydai1124:
@liyinan926 @sahilTakiar can you take another look?

sahilTakiar:
Is there a dire need to get this checked in soon? If not I think we should do a little more design work before making these changes. The PR adds modification functionality only for a `StateStore` that is backed by a `FsStateStore`. I think we should instead take the following approach.

* New interface called `ModifiableStateStore`, with methods like `modify(String storeName, String tableName, String key, Object newValue)`.
* Have `FsStateStore` implement this new interface and move all the code in this PR into this new method.
* Modify the the command line tool changes in this PR to work with `ModifiableStateStore`.
    * This has the advantage that the command line tool is not tied to any specific State Store implementation. Currently, this PR assumes we are interacting with an `FsStateStore`.

liyinan926:
Agreed with @sahilTakiar.

ydai1124:
Makes sense. Updated ETL-3793 and close this PR now.

PR #449, opened 2015-11-13
Make partitioned data writer class configurable

zliu41:
Need to extend `PartitionedDataWriter` in Kafka adapter for collecting audit counts for each record written.


zliu41:
@ibuenros can you review?

PR #450, opened 2015-11-13
Make writerId unique among writers of the same task

zliu41:
Which makes it easier for multiple writers to pass states to the same state object.


zliu41:
@ibuenros can you review?

ibuenros:
LGTM

PR #451, opened 2015-11-13
Fixed a test that wouldn't compile in Hudson.

ibuenros:


ibuenros:
Trivial test fix. Self-approving.

PR #452, opened 2015-11-13
Make known hosts optional for sftp connection

pcadabam:
@liyinan926 can you review this?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
@@ -33,6 +33,7 @@ import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -157,11 +157,16 @@ public void connect()
         }
       }
 
-      jsch.setKnownHosts(knownHosts);
-
       session = jsch.getSession(userName, hostName, port);
       session.setConfig("PreferredAuthentications","publickey");
 
+      if (knownHosts == null) {
```
liyinan926: Please be more defensive and use `Strings.isNullOrEmpty(knownHosts)`.

liyinan926:
LGTM other than the minor comment.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
pcadabam: Changed

PR #453, opened 2015-11-14
Set group after the staging output file gets created

liyinan926:
Due to some recent changes to `FsDataWriter` and its subclasses, `FsDataWriter` will throw an `FileNotFoundException` if a group is specified and it tries to set the group on the staging output file, which won't be created until later in the constructor of `AvroHdfsDataWriter` and `SimpleDataWriter`. Attempt to set group on a file that does not exist will throw an `FileNotFoundException`. 

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
@@ -12,6 +12,7 @@
 
 package gobblin.writer;
 
+import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


liyinan926:
@zliu41 can you review?

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -139,22 +139,33 @@ public FsDataWriter(FsDataWriterBuilder<?, D> builder, State properties) throws
...
     return this.closer.register(this.fs.create(this.stagingFile, this.filePermission, true, this.bufferSize,
         this.replicationFactor, this.blockSize, null));
   }
 
   /**
+   * Set the group name of the staging output file.
+   *
+   * @throws IOException if it fails to set the group name
+   */
+  protected void setGroup() throws IOException {
```
zliu41: You can rename it to `setStagingFileGroup`. Also add a precondition check for the existence of the staging file.

zliu41:
LGTM otherwise

PR #457, opened 2015-11-16
Submitting event metadata for each successful job, including task exceptions

sahilTakiar:
* Created a new class called `JobExecutionEventSubmitter` that submits metadata for a given `JobState`, the events are submitted using a `EventSubmitter`
* `JobExecutionEventSubmitter` is wrapped in a `JobListener` as is invoked as any other `JobListener`
* Added a new util class called `JobListeners` that provides a way of combining multiple `JobListener`s into one `JobListener` and running them in parallel
* The config key `ConfigurationKeys.TASK_FAILURE_EXCEPTION_KEY` is updated not just for Task failures, but also for task publishing failures

@liyinan926 can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -96,6 +96,9 @@ public abstract class AbstractJobLauncher implements JobLauncher {
   // An EventBuilder with basic metadata.
   protected final EventSubmitter eventSubmitter;
 
+  // A list of JobListeners that will be injected into the user provided JobListener
+  private final List<JobListener> mandatoryJobListeners = Lists.newArrayList();
+
   public AbstractJobLauncher(Properties jobProps) throws Exception {
     Preconditions.checkArgument(jobProps.containsKey(ConfigurationKeys.JOB_NAME_KEY),
         "A job must have a job name specified by job.name");
@@ -118,6 +121,9 @@ public abstract class AbstractJobLauncher implements JobLauncher {
         });
 
     this.eventSubmitter = new EventSubmitter.Builder(this.runtimeMetricContext, "gobblin.runtime").build();
+
+    JobExecutionEventSubmitter jobExecutionEventSubmitter = new JobExecutionEventSubmitter(this.eventSubmitter);
+    this.mandatoryJobListeners.add(new JobExecutionEventSubmitterListener(jobExecutionEventSubmitter));
   }
 
   /**
@@ -153,17 +159,23 @@ public abstract class AbstractJobLauncher implements JobLauncher {
     }
 
     synchronized (this.cancellationExecution) {
+      Closer closer = Closer.create();
       try {
         while (!this.cancellationExecuted) {
           // Wait for the cancellation to be executed
           this.cancellationExecution.wait();
         }
 
-        if (jobListener != null) {
-          jobListener.onJobCancellation(this.jobContext.getJobState());
-        }
+        JobListener parallelJobListener = closer.register(getParallelCombinedJobListener(jobListener));
+        parallelJobListener.onJobCancellation(this.jobContext.getJobState());
       } catch (InterruptedException ie) {
         Thread.currentThread().interrupt();
+      } finally {
+        try {
+          closer.close();
+        } catch (IOException e) {
+          throw new JobException("Failed to execute all JobListeners", e);
+        }
       }
     }
   }
@@ -263,12 +275,6 @@ public abstract class AbstractJobLauncher implements JobLauncher {
 
         // Write job execution info to the job history store upon job termination
         this.jobContext.storeJobExecutionInfo();
-
-        if (this.jobContext.getJobMetricsOptional().isPresent()) {
-          this.jobContext.getJobMetricsOptional().get().triggerMetricReporting();
-          this.jobContext.getJobMetricsOptional().get().stopMetricReporting();
-          JobMetrics.remove(jobState);
-        }
       } finally {
         launchJobTimer.stop();
         unlockJob();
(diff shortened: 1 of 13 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -283,8 +297,18 @@ public void launchJob(JobListener jobListener) throws JobException {
       }
     }
 
-    if (jobListener != null) {
-      jobListener.onJobCompletion(jobState);
+    Closer closer = Closer.create();
+    try {
+      int numThreads = this.jobContext.getJobState().getPropAsInt(JobListeners.PARALLEL_JOB_LISTENER_THREADS,
```
liyinan926: Seems this code can be abstracted into a method and called in both places.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -483,6 +507,17 @@ private void cleanupStagingData(JobState jobState) {
   }
 
   /**
+   * Combines the specified {@link JobListener} with the {@link #mandatoryJobListeners} for this job. Uses
+   * {@link JobListeners#parallelJobListener(List, int)} to create a {@link CloseableJobListener} that will execute all
+   * the {@link JobListener}s in parallel.
+   */
+  private CloseableJobListener getParallelCombinedJobListener(JobListener jobListener, int numThreads) {
+    List<JobListener> jobListeners = Lists.newArrayList(this.mandatoryJobListeners);
```
liyinan926: You can do `ImmutableList.copyOf()`.

gobblin-runtime/src/main/java/gobblin/runtime/CloseableJobListener.java
```diff
@@ -0,0 +1,13 @@
+package gobblin.runtime;
+
+import java.io.Closeable;
+
+
+/**
+ * Extension of {@link JobListener} that also extends {@link Closeable}.
+ *
+ * @see JobListener
```
liyinan926: `{@link JobListener}.`

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -427,4 +429,14 @@ private void persistDatasetState(String datasetUrn, JobState.DatasetState datase
     datasetState.setId(datasetUrn);
     this.datasetStateStore.persistDatasetState(datasetUrn, datasetState);
   }
+
+  /**
+   * Sets the {@link ConfigurationKeys#TASK_FAILURE_EXCEPTION_KEY} for each given {@link TaskState} to the given
+   * {@link Throwable}.
+   */
+  private void setTaskFailureException(List<TaskState> taskStates, Throwable t) {
```
liyinan926: I suggest using `Collection<TaskState>` instead of `List<TaskState>` here.

gobblin-runtime/src/main/java/gobblin/runtime/JobExecutionEventSubmitter.java
```diff
@@ -0,0 +1,116 @@
...
+
+    jobMetadataBuilder.put(JOB_ID, jobState.getJobId());
+    jobMetadataBuilder.put(JOB_NAME, jobState.getJobName());
+    jobMetadataBuilder.put(JOB_START_TIME, Long.toString(jobState.getStartTime()));
+    jobMetadataBuilder.put(JOB_END_TIME, Long.toString(jobState.getEndTime()));
+    jobMetadataBuilder.put(JOB_STATE, jobState.getState().toString());
+    jobMetadataBuilder.put(JOB_LAUNCHED_TASKS, Integer.toString(jobState.getTaskCount()));
+    jobMetadataBuilder.put(JOB_COMPLETED_TASKS, Integer.toString(jobState.getCompletedTasks()));
+    jobMetadataBuilder.put(JOB_LAUNCHER_TYPE, jobState.getLauncherType().toString());
+    jobMetadataBuilder.put(JOB_TRACKING_URL, getValueIfPresent(jobState.getTrackingURL()));
```
liyinan926: `getValueIfPresent` can be done using `Optional.or`.

gobblin-runtime/src/main/java/gobblin/runtime/JobListeners.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ParallelJobListener.class);
+
+    private final List<JobListener> jobListeners;
+    private final ExecutorService executor;
+    private final List<Future> futures;
+
+    public ParallelJobListener(List<JobListener> jobListeners, int numThreads) {
+      this.jobListeners = jobListeners;
+      this.executor = Executors.newFixedThreadPool(numThreads,
```
liyinan926: I think we can use a `CachedThreadPool` instead of a fixed-size thread pool.

gobblin-runtime/src/main/java/gobblin/runtime/JobListeners.java
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Implementation of {@link CloseableJobListener} that executes a given {@link List} of {@link JobListener}s in parallel.
+   */
+  private static final class ParallelJobListener implements CloseableJobListener {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ParallelJobListener.class);
+
+    private final List<JobListener> jobListeners;
+    private final ExecutorService executor;
+    private final List<Future> futures;
```
liyinan926: Instead of using a list of `Future`, you can use a `CompletionService`.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -304,11 +304,13 @@ private void publishTaskData() throws IOException {
...
       publisher.publish(this.taskState);
-    } catch (IOException e) {
-      throw closer.rethrow(e);
-    } catch (Throwable t) {
+    } catch (ClassCastException e) {
       LOG.error(String.format("To publish data in task, the publisher class (%s) must extend %s",
-          ConfigurationKeys.DATA_PUBLISHER_TYPE, SingleTaskDataPublisher.class.getSimpleName()), t);
+          ConfigurationKeys.DATA_PUBLISHER_TYPE, SingleTaskDataPublisher.class.getSimpleName()), e);
+      this.taskState.setTaskFailureException(e);
+      closer.rethrow(e);
```
liyinan926: `throw closer.rethrow(e);`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
sahilTakiar: I tried that but an exception gets thrown in `JobListeners.parallelJobListener` when the method tries to remove `null`s from the given `List` (line 41). I stuck with this approach because it allows the `JobListeners.parallelJobListener` to be defensive against `null` entries.

sahilTakiar:
@liyinan926 updated

liyinan926:
LGTM.

PR #459, opened 2015-11-17
ETL-3802: Trigger compaction if late data exceeds threshold

ydai1124:
@zliu41 Can you review?
Has not tested yet. Want to first make sure this is the appropriate way to trigger recompaction.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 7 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -857,6 +914,13 @@ protected void afterExecute(Runnable r, Throwable t) {
 
           // Compaction job of a dataset is successful.
           jobRunner.getDataset().setState(COMPACTION_COMPLETE);
+
+          // Add a new MPCompactor job if recompaction is necessary.
+          if (jobRunner.getDataset().needToRecompact()) {
```
zliu41: the dataset state should be set to `COMPACTION_COMPLETE` only if it doesn't need to be recompacted. Otherwise `submitCompactionJobsAndWaitForCompletion()` will terminate the job.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -857,6 +914,13 @@ protected void afterExecute(Runnable r, Throwable t) {
...
+            for (Dataset recompactDataset: MRCompactor.this.createDatasetForRecompact(jobRunner.getDataset())) {
+              MRCompactor.this.runCompactionForDataset(recompactDataset, true);
```
zliu41: You just need to set the dataset state to `VERIFIED` here. The recompaction job will then be submitted in `submitCompactionJobsAndWaitForCompletion()`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -449,13 +462,28 @@ private void createJobPropsForTopics(List<String> topics) {
     List<String> list = this.state.getPropAsList(COMPACTION_NORMAL_PRIORITY_TOPICS, StringUtils.EMPTY);
     return DatasetFilterUtils.getPatternsFromStrings(list);
   }
+  
+  private Map<String, Double> getTopicsAndRecompactThreshold() {
+    Map<String, Double> topicsAndRecompactThreshold = Maps.newHashMap();
+    for (String entry : this.state.getPropAsList(COMPACTION_LATEDATA_THRESHOLD_FOR_RECOMPACT_PER_TOPIC, StringUtils.EMPTY)) {
+      String[] topicAndRecompactThreshold = Iterables.toArray(Splitter.on(TOPIC_AND_RECOMPACT_THRESHOLD_SEPARATOR)
```
zliu41: This should support regex, and multiple topics with the same threshold. The format could look like this: `A.*,B.*:0.2 C.*,D.*:0.3`

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
ydai1124: The recompaction job is based on a new `dataset`: 'recompactDataset', and its initial state is `unverified`.

zliu41: `MRCompactor.compact()` terminates once all datasets in `this.datasets` are `COMPACTION_COMPLETE`. Since the new dataset is not added to `this.datasets`, `MRCompactor` will terminate and the recompaction job will be killed.

Even if the new dataset is added to `this.datasets`, there will be race conditions: `MRCompactor may determine all datasets are done before the new dataset is added. So the right approach is to reuse the original dataset and never set it to `COMPACTION_COMPLETE` until the recompaction is done.

ydai1124:
@zliu41 Updated.

zliu41:
LGTM

ydai1124:
@zliu41 Fixed several bugs during test. Can you review?

zliu41:
LGTM

PR #460, opened 2015-11-17
Setting job.launcher.type for YARN jobs

sahilTakiar:
Setting the `job.launcher.type` key to `LauncherTypeEnum.YARN` for YARN jobs. This allows us to differentiate between jobs that run on YARN vs. MR vs. Local when consuming events produced by `JobExecutionEventSubmitter`.


sahilTakiar:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #461, opened 2015-11-17
Generalize and abstract KafkaAvroExtractor

zliu41:
Currently `KafkaAvroExtractor` requires a specific implementation of the schema registry, as well as a specific format of the Kafka record payload (first byte = 0, next 16 bytes are schema ID, remaining bytes are the record). It is not applicable anywhere else outside LinkedIn.

This PR generalizes it and makes it an abstract class.

Also created an interface `KafkaSchemaRegistry` which can be implemented in different ways.


zliu41:
@liyinan926 can you review?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
```diff
@@ -43,7 +42,7 @@
  *
  * @author ziliu
  */
-public class KafkaAvroSchemaRegistry {
+public class KafkaAvroSchemaRegistry implements KafkaSchemaRegistry<String, Schema> {
```
chavdar: The presence of KafkaAvroSchemaRegistry and KafkaSchemaRegistry in gobblin-metrics is a bit strange, isn't it?

PR #467, opened 2015-11-18
Add support for reconnecting to running Yarn app after driver restart

liyinan926:
1. Added support to `GobblinYarnAppLauncher` to be able to reconnect after restart to the Yarn application it launched previously. The way it works is straightforward: upon starting, it checks if a file storing the previous Yarn application ID exists, and if so, it reads the previous application ID and checks if the ID matches any active Yarn application. If so, it simply starts the application state monitor. Otherwise, it submits a new application and persist the application ID. 
2. Switched to `MetricRegistry` for JVM metrics in `GobblinApplicationMaster` and `GobblinWorkUnitRunner` since `MetricContext` no longer supports registering Dropwizard metric instances.
3. Added some unit tests. More to come once I figure out how to test the parts that interact with Yarn and Helix.


liyinan926:
@zliu41 @ibuenros @sahilTakiar can you review?

sahilTakiar:
Can we move all processing of the applicationId (storing, reading, deleting the applicationId) behind an interface. That way we can change out where the applicationId is stored in the future. For example, we may decide that instead of storing the applicationId in a file on HDFS, we want to store the applicationid in Zookeeper.

liyinan926:
@sahilTakiar updated.

gobblin-yarn/src/main/java/gobblin/yarn/FsApplicationIdStore.java
```diff
@@ -0,0 +1,92 @@
...
+  public FsApplicationIdStore(FileSystem fs, Path applicationIdFilePath) {
+    this.fs = fs;
+    this.applicationIdFilePath = applicationIdFilePath;
+  }
+
+  @Override
+  public void put(String applicationId) throws IOException {
+    delete();
+
+    Closer closer = Closer.create();
```
sahilTakiar: Can we use `try with resources` now since we migrated to Java 7.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -147,18 +141,8 @@ public GobblinApplicationMaster(String applicationName, Config config) throws Ex
 
     this.serviceManager = new ServiceManager(services);
 
-    List<Tag<?>> tags = ImmutableList.<Tag<?>>builder()
```
sahilTakiar: Whats the reason for removing all of these.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -288,6 +331,38 @@ public void handleApplicationReportArrivalEvent(ApplicationReportArrivalEvent ap
...
+    return Optional.of(setupAndSubmitApplication());
+  }
+
+  private Optional<ApplicationId> getReconnectableApplicationId() throws YarnException, IOException {
+    Optional<String> existingApplicationId = this.applicationIdStore.get();
+    if (!existingApplicationId.isPresent()) {
+      return Optional.absent();
+    }
+
+    List<ApplicationReport> applicationReports =
```
sahilTakiar: Would it not be more efficient to first get the `ApplicationReport` if the specified `applicationId` and then check if its `state` is in re-connectable?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: We have to switch back to use `MetricRegistry` as `MetricContext` no longer allows registering standard Dropwizard metrics, like the ones registered in this class.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: The Yarn API does not support creating an `ApplicationId` object from a string ID, while `getApplicationReport` needs an `ApplicationId` object.

liyinan926:
@sahilTakiar updated 9f25a4c.

sahilTakiar:
LGTM

gobblin-yarn/src/main/java/gobblin/yarn/ApplicationIdStore.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>
+ *   This is useful in {@link GobblinYarnAppLauncher} to retrieve and persist
+ *   the application ID of the Yarn application it launches so it is able to
+ *   reconnect to the application upon restart.
+ * </p>
+ *
+ * @author ynli
+ */
+interface ApplicationIdStore extends Closeable {
```
zliu41: Why is it package private? nothing outside the package can see it.

liyinan926: Yes, this is intentional since right now only `GobblinYarnAppLauncher` will use it.

gobblin-yarn/src/main/java/gobblin/yarn/FsApplicationIdStore.java
```diff
@@ -0,0 +1,78 @@
...
+  private final Path applicationIdFilePath;
+
+  public FsApplicationIdStore(FileSystem fs, Path applicationIdFilePath) {
+    this.fs = fs;
+    this.applicationIdFilePath = applicationIdFilePath;
+  }
+
+  @Override
+  public void put(String applicationId) throws IOException {
+    delete();
```
zliu41: no need to delete, `fs.create()` will overwrite it.

gobblin-yarn/src/main/java/gobblin/yarn/FsApplicationIdStore.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    try (InputStreamReader isReader = new InputStreamReader(this.fs.open(this.applicationIdFilePath))) {
+      return Optional.of(CharStreams.toString(isReader));
+    }
+  }
+
+  @Override
+  public void delete() throws IOException {
+    if (this.fs.exists(this.applicationIdFilePath)) {
```
zliu41: You can use `HadoopUtils.deletePath()` which saves the `if` block.

gobblin-yarn/src/main/java/gobblin/yarn/FsApplicationIdStore.java
```diff
@@ -0,0 +1,78 @@
...
+      this.fs.delete(this.applicationIdFilePath, false);
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
+    delete();
```
zliu41: Is it possible that `close()` is called on unexpected shut down and the application ID is lost, making it impossible to reconnect?

liyinan926: Yes, it is possible. I will make `close` a no-op.

liyinan926:
@zliu41 updated 276e858.

gobblin-yarn/src/main/java/gobblin/yarn/FsApplicationIdStore.java
```diff
@@ -0,0 +1,76 @@
...
+      return Optional.absent();
+    }
+
+    try (InputStreamReader isReader = new InputStreamReader(this.fs.open(this.applicationIdFilePath))) {
+      return Optional.of(CharStreams.toString(isReader));
+    }
+  }
+
+  @Override
+  public void delete() throws IOException {
```
zliu41: Can you add some javadoc on how this should be used, in particular, is it mandatory or optional to call this method on normal shutdown? What is the consequence if the file is not deleted, will there be any trouble for the next run?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
ibuenros: I thought we were registering these metrics to the root metric context instead?

zliu41:
Seems java 8 build is consistently failing.

liyinan926:
Synced up with @ibuenros. We realized since the file storing the application ID is under a directory named after the application name, if it happens to have two instances from the same user with the same application name (maybe by mistake), the two instances will try updating the same file. So storing the application ID offers no real benefit over simply using the application name to match the application. So I'm gonna remove `ApplicationIdStore` and `FsApplicationIdStore` and rely on application name.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
liyinan926: Actually we could not simply use `RootMetricContext`. First of all, there's no way to register non-context-aware metrics unless we have a `MetricRegistry` instance in `RootMetricContext` since `InnerMetricContext` does not allow registering standard metrics. But once we add a  `MetricRegistry` instance, then we have to worry about how to report them. Apparently the reporters in `RootMetricContext` only work with `ContextAwareMetric`s.

liyinan926:
Updated 3d72580.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -102,6 +118,19 @@
 
   private static final Splitter SPLITTER = Splitter.on(',').omitEmptyStrings().trimResults();
 
+  // The set of Yarn application types this class is interested in. This is used to
+  // lookup the application this class has launched previously upon restarting.
+  private static final Set<String> APPLICATION_TYPES = ImmutableSet.of("YARN");
```
ibuenros: Can we use a custom application type instead? That gives us a more precise way of finding the correct running application. You can use https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/yarn/api/records/ApplicationSubmissionContext.html#setApplicationType(java.lang.String) to set the application type.

ibuenros:
LGTM other than one comment.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
liyinan926: Sure.

PR #468, opened 2015-11-19
Check path existance before calling fileSystem.listStatus

pcadabam:
@ibuenros can you review this? This is the fix for failing build on hudson. Hadoop2 freaks out if you call liststatus on a path that does not exist.


ibuenros:
LGTM

PR #470, opened 2015-11-19
Add schema registry to KafkaAvroExtractor

zliu41:
1. Added schema registry to `KafkaAvroExtractor`, since people are using Confluent's schema registry. It is not mandatory, and if one doesn't want to use schema registry, she needs to override `getExtractorSchema()` and get the schema in her own way.
2. Changed `KafkaSchemaRegistry` from interface to abstract class, and moved some code over from `KafkaAvroSchemaRegistry` (mainly the loading cache part), which can be reused by other implementations of schema registry.

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
@@ -13,10 +13,8 @@
 package gobblin.metrics.kafka;
 
 import java.io.IOException;
-import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
 
 import gobblin.util.AvroUtils;
 import org.apache.avro.Schema;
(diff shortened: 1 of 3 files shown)


zliu41:
@liyinan926 can you review?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
```diff
@@ -216,109 +195,44 @@ public synchronized String register(Schema schema) throws SchemaRegistryExceptio
...
 
-    private Schema fetchSchemaByKey(String key) throws SchemaRegistryException, HttpException, IOException {
-      String schemaUrl = KafkaAvroSchemaRegistry.this.url + GET_RESOURCE_BY_ID + key;
-
-      GetMethod get = new GetMethod(schemaUrl);
-
-      int statusCode;
-      String schemaString;
+    Schema schema;
+    if (schemaString.startsWith("{")) {
```
liyinan926: Is this necessary? Can we just parse the schema and catch a `ParseException`?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaSchemaRegistry.java
```diff
@@ -21,32 +39,162 @@
...
-  public K register(S schema, String name) throws SchemaRegistryException;
+  public abstract K register(S schema, String name) throws SchemaRegistryException;
+
+  private class KafkaSchemaCacheLoader extends CacheLoader<K, S> {
+
+    private final Map<K, FailedFetchHistory> failedFetchHistories;
+
+    private KafkaSchemaCacheLoader() {
+      super();
+      this.failedFetchHistories = Maps.newHashMap();
```
liyinan926: Should a `ConcurrentHashMap` be used?

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaAvroSchemaRegistry.java
zliu41: Don't remember the reason to do this, but I don't think it's necessary.

zliu41:
@liyinan926 updated

PR #472, opened 2015-11-19
Submit events from distcp-ng

pcadabam:
@ibuenros can you review this?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
@@ -34,6 +34,8 @@ import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
 import gobblin.instrumented.writer.InstrumentedDataWriterDecorator;
 import gobblin.instrumented.writer.InstrumentedPartitionedDataWriterDecorator;
+import gobblin.util.AvroUtils;
+import gobblin.util.FinalState;
 import gobblin.writer.partitioner.WriterPartitioner;
 
 
@@ -44,7 +46,7 @@ import gobblin.writer.partitioner.WriterPartitioner;
  * @param <D> record type.
  */
 @Slf4j
-public class PartitionedDataWriter<S, D> implements DataWriter<D> {
+public class PartitionedDataWriter<S, D> implements DataWriter<D>, FinalState {
 
   private static final GenericRecord NON_PARTITIONED_WRITER_KEY =
       new GenericData.Record(SchemaBuilder.record("Dummy").fields().endRecord());
@@ -170,5 +172,36 @@ public class PartitionedDataWriter<S, D> implements DataWriter<D> {
     }
     return this.builder.get().forPartition(partition).withWriterId(this.baseWriterId + "_" + this.writerIdSuffix++)
         .build();
+  }
+
+  @Override
+  public State getFinalState() {
+
+    State state = new State();
+    try {
+      for (Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+        if (entry.getValue() instanceof FinalState) {
+
+          State partitionFinalState = ((FinalState) entry.getValue()).getFinalState();
+
+          if (this.shouldPartition) {
+            for (String key : partitionFinalState.getPropertyNames()) {
+              // Prevent overwriting final state across writers
+              partitionFinalState.setProp(key + "_" + AvroUtils.serializeAsPath(entry.getKey(), false, true),
+                  partitionFinalState.getProp(key));
+            }
+          }
+
+          state.addAll(partitionFinalState);
+        }
+      }
+      state.setProp("RecordsWritten", recordsWritten());
+      state.setProp("BytesWritten", bytesWritten());
+    } catch (Exception exception) {
+      log.warn("Failed to get final state." + exception.getMessage());
+      // If Writer fails to return bytesWritten, it might not be implemented, or implemented incorrectly.
+      // Omit property instead of failing.
+    }
+    return state;
   }
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.data.management.copy.publisher;
+
+import com.google.common.collect.ImmutableMap;
+
+import gobblin.configuration.WorkUnitState;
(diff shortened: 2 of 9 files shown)


gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -171,4 +172,24 @@ public void close() throws IOException {
...
   }
+
+  @Override
+  public State getFinalState() {
+
+    State state = new State();
+    try {
+      for (Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+        if (entry.getValue() instanceof FinalState) {
+          state.addAll(((FinalState)entry.getValue()).getFinalState());
```
ibuenros: This will overwrite values that are repeated in multiple writers. Can we keep them separate somehow? Like adding partition info as suffix?

pcadabam: Done.
key names will be SomeKey_0 for first partition.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class CopyEventSubmitterHelper {
+
+  private static final String DATASET_PUBLISHED_EVENT_NAME = "DatasetPublished";
+  private static final String DATASET_PUBLISHED_FAILED_EVENT_NAME = "DatasetPublishFailed";
+  public static final String DATASET_ROOT_METADATA_NAME = "datasetRoot";
+  public static final String DATASET_TARGET_ROOT_METADATA_NAME = "datasetTargetRoot";
+
+  public static void submitSuccessfulDatasetPublish(EventSubmitter eventSubmitter, CopyableDataset copyableDataset) {
+    eventSubmitter.submit(DATASET_PUBLISHED_EVENT_NAME, getDatasetEventMetadata(copyableDataset));
```
ibuenros: Can we make these SLAEvents?

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
ibuenros: Unfortunately there is no consistency in order of partitions, so the suffix _0 does not mean anything. Can we just serialize the values of the partition (the partition is an avro record, so it shouldn't be hard)?

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
pcadabam: Done.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -171,4 +173,35 @@ public void close() throws IOException {
...
+    try {
+      for (Map.Entry<GenericRecord, DataWriter<D>> entry : this.partitionWriters.asMap().entrySet()) {
+        if (entry.getValue() instanceof FinalState) {
+
+          State partitionFinalState = ((FinalState) entry.getValue()).getFinalState();
+
+          if (this.shouldPartition) {
+            for (String key : partitionFinalState.getPropertyNames()) {
+              // Prevent overwriting final state across writers
+              partitionFinalState.setProp(key + "_" + AvroUtils.serializeAsPath(entry.getKey(), true, true),
```
ibuenros: can we set includeFieldNames to false to reduce the length of the string?

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
```diff
@@ -0,0 +1,60 @@
...
+   * Submit an sla event when a {@link CopyableFile} is published. The <code>workUnitState</code> passed should have the
+   * required {@link SlaEventKeys} set.
+   *
+   * @see {@link SlaEventSubmitter#submit()} for all the required {@link SlaEventKeys} to be set
+   *
+   * @param eventSubmitter
+   * @param workUnitState
+   */
+  static void submitSuccessfulFilePublish(EventSubmitter eventSubmitter, WorkUnitState workUnitState) {
+    new SlaEventSubmitter(eventSubmitter, FILE_PUBLISHED_EVENT_NAME, workUnitState.getProperties()).submit();
```
ibuenros: iirc slaeventsubmitter will error out if certain fields are not set. Is this call guaranteed to set all required fields?

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
```diff
@@ -0,0 +1,60 @@
...
+
+  static void submitSuccessfulDatasetPublish(EventSubmitter eventSubmitter, CopyableDataset copyableDataset) {
+    SlaEventSubmitter.builder().eventSubmitter(eventSubmitter).eventName(DATASET_PUBLISHED_EVENT_NAME)
+        .datasetUrn(copyableDataset.datasetRoot().toString()).partition(copyableDataset.datasetRoot().toString())
+        .additionalMetadata(DATASET_TARGET_ROOT_METADATA_NAME, copyableDataset.datasetTargetRoot().toString()).build()
+        .submit();
+  }
+
+  static void submitFailedDatasetPublish(EventSubmitter eventSubmitter, CopyableDataset copyableDataset) {
+    SlaEventSubmitter.builder().eventSubmitter(eventSubmitter).eventName(DATASET_PUBLISHED_FAILED_EVENT_NAME)
```
ibuenros: I'm not sure failure should be an SLA event, or at least we should decide what exactly counts as SLA event.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
pcadabam: Yes the CopySource sets all the required fields in the workunit.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
pcadabam: Makes sense. Changed it to eventSubmitter.submit

ibuenros:
LGTM

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -171,4 +173,35 @@ public void close() throws IOException {
...
+              partitionFinalState.setProp(key + "_" + AvroUtils.serializeAsPath(entry.getKey(), false, true),
+                  partitionFinalState.getProp(key));
+            }
+          }
+
+          state.addAll(partitionFinalState);
+        }
+      }
+      state.setProp("RecordsWritten", recordsWritten());
+      state.setProp("BytesWritten", bytesWritten());
```
zliu41: can you use `ConfigurationKeys.WRITER_BYTES_WRITTEN`?

pcadabam: I am not sure about that. There may already be consumers who use/expect a metric with "BytesWritten". 
The metric names BytesWritten and RecordsWritten were already used in other writers.

zliu41: Makes sense if it's been there before. I thought this is something you just added.

PR #473, opened 2015-11-20
Make JobState use local job launcher by default

zliu41:
The recent change to `JobState` is breaking Kafka adapter since Kafka adapter uses a subclass of `MRJobLauncher` that is not in github. (`com.linkedin.gobblinjobmetrics.KafkaMRJobLauncher`)

This PR fixes it by having `getLauncherType()` return `LOCAL` if the job launcher name is not found in `JobLauncherType`.


zliu41:
@liyinan926 @sahilTakiar can one of you review?

liyinan926:
LGTM.

PR #474, opened 2015-11-20
ETL-3803: remove "topic" from gobblin-compaction so that gobblin-compaction no longer requires a topic name, but just input/output path..

ydai1124:
Currently gobblin-compaction has a specific way of defining topics and datasets: the user specifies a "compaction.input.dir" property, e.g., /data/tracking, then each subfolder under "compaction.input.dir" is considered a "topic", and it is up to MRCompactorJobPropCreator to determine what are the "datasets" for each topic, e.g., /data/tracking/PVE/hourly/2015/11/15 is a dataset for topic PVE. A compaction job will run on each dataset.

Many use cases don't fit in the scenario where each subfolder under "compaction.input.dir" should be considered a topic (e.g., maybe the user just want to deduplicate /user/xyz/data). Moreover, "topic" is a Kafka-specific concept.

Given the existence of dataset, "topic" should be no longer be used. MRCompactorJobPropCreator should directly find datasets given "compaction.input.dir". This PR added `DatasetsFinder` to find compaction datasets with input/output path information. `SimpleDatasetsFinder` treats "compaction.input.dir" as one single dataset for compaction. `TimeBasedSubDirDatasetsFinder` will be used to find datasets under "compaction.input.dir" in the form of [compaction.input.dir]/[subdir]/[compaction.timebased.folder.pattern].

@zliu41 Can you review?


ydai1124:
Updated with some fix. Tested in Azkaban with regular compaction, late data movement, and recompaction.

zliu41:
Please resolve conflicts.

ydai1124:
@zliu41 Updated.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -316,6 +289,17 @@ private FileSystem getFileSystem() throws IOException {
     }
   }
 
+  private DatasetsFinder getDatasetsFinder() {
+    try {
+      return (DatasetsFinder) Class.forName(this.state.getProp(COMPACTION_DATASETS_FINDER, DEFAULT_COMPACTION_DATASETS_FINDER))
+        .getConstructor(State.class, FileSystem.class, String.class, String.class, String.class, List.class, List.class,
```
zliu41: There are too many constructor parameters. You can just pass `state` and get everything else from it in the dataset finder.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/SubDirDatasetsFinder.java
```diff
@@ -0,0 +1,113 @@
...
+ * A {@link DatasetsFinder} based on subdir of the inputDir.
+ * Each subfolder in {@link #inputDir} is considered a dataset, if it satisfies blacklist and whitelist.
+ * 
+ * For example, suppose {@link #inputDir} is '/inputdata', and a topic is
+ * '/inputdata/mytopic'. If this topic is considered a {@link Dataset}, all data under this folder will be compacted.
+ * If a particular day's data of this topic is considered a {@link Dataset}, e.g.,
+ * '/inputdata/mytopic/daily/2015/08/09', then data under this folder will be compacted.
+ */
+@Slf4j
+public class SubDirDatasetsFinder extends DatasetsFinder {
```
zliu41: There should also be a `TimeBasedSubDirDatasetsFinder`.

ydai1124:
@zliu41 Updated.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/TimeBasedSubDirDatasetsFinder.java
```diff
@@ -0,0 +1,113 @@
...
+ * A {@link DatasetsFinder} based on subdir of the inputDir.
+ * Each subfolder in {@link #inputDir} is considered a dataset, if it satisfies blacklist and whitelist.
+ *
+ * For example, suppose {@link #inputDir} is '/inputdata', and a topic is
+ * '/inputdata/mytopic'. If this topic is considered a {@link Dataset}, all data under this folder will be compacted.
+ * If a particular day's data of this topic is considered a {@link Dataset}, e.g.,
+ * '/inputdata/mytopic/daily/2015/08/09', then data under this folder will be compacted.
+ */
+@Slf4j
+public class TimeBasedSubDirDatasetsFinder extends DatasetsFinder {
```
zliu41: Since this class is called `TimeBasedSubDirDatasetsFinder`, Most of the logic in `MRCompactorTimeBasedJobPropCreator` should be moved here, such as parsing the folder path to get folder time, check whether folder time is within allowed period, etc.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/DatasetsFinder.java
```diff
@@ -0,0 +1,127 @@
...
+  private FileSystem getFileSystem() {
+    try {
+      if (this.state.contains(MRCompactor.COMPACTION_FILE_SYSTEM_URI)) {
+        URI uri = URI.create(this.state.getProp(MRCompactor.COMPACTION_FILE_SYSTEM_URI));
+        return FileSystem.get(uri, this.conf);
+      } else {
+        return FileSystem.get(this.conf);
+      }
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to get filesystem for datasetsFinder " + e, e);
```
zliu41: `+ e` is unnecessary

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -316,6 +272,15 @@ private FileSystem getFileSystem() throws IOException {
     }
   }
 
+  private DatasetsFinder getDatasetsFinder() {
+    try {
+      return (DatasetsFinder) Class.forName(this.state.getProp(COMPACTION_DATASETS_FINDER, DEFAULT_COMPACTION_DATASETS_FINDER))
+          .getConstructor(State.class).newInstance(state);
+      } catch (Exception e) {
+        throw new RuntimeException("Failed to initiailize DatasetsFinder " + e, e);
```
zliu41: ditto

ydai1124:
@zliu41 Updated.

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
```diff
@@ -118,13 +112,13 @@ public Dataset build() {
   private Path inputPath;
   private Path inputLatePath;
   private State jobProps;
+  private Optional<State> extraDatasetProps;
```
zliu41: Can remove `extraDatasetProps` and reuse  `jobProps`?
You can add a method `Dataset.setJobProp()` that sets a single property.

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -56,8 +56,7 @@ public static void setUpstreamTimeStamp(State state, long time) {
   }
 
   private static void setDatasetUrn(Dataset dataset) {
-    dataset.jobProps().setProp(SlaEventKeys.DATASET_URN_KEY,
-        new Path(dataset.jobProps().getProp(MRCompactor.COMPACTION_DEST_DIR), dataset.topic()).toString());
+    dataset.jobProps().setProp(SlaEventKeys.DATASET_URN_KEY, dataset.outputPath());
```
zliu41: Is this correct? Originally `DATASET_URN_KEY` is set to the topic folder, e.g., /data/tracking/PVE. But `dataset.outputPath()` includes the time bucket.

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
ydai1124: The reason to have `extraDatasetProps` is because `jobProps` is set in JobPropsCreater, but `extraDatasetProps` is set in `DatasetsFinder`. The problem of reusing `jobProps` is that it cannot differentiate replacing a `jobProps` and adding extra properties to a `jobProps`.

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
ydai1124: Fixed.

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
zliu41: Can you use `setJobProp()` to add a single property and `setJobProps()` to set multiple properties?

gobblin-compaction/src/main/java/gobblin/compaction/event/CompactionSlaEventHelper.java
```diff
@@ -56,8 +59,19 @@ public static void setUpstreamTimeStamp(State state, long time) {
   }
 
   private static void setDatasetUrn(Dataset dataset) {
-    dataset.jobProps().setProp(SlaEventKeys.DATASET_URN_KEY,
-        new Path(dataset.jobProps().getProp(MRCompactor.COMPACTION_DEST_DIR), dataset.topic()).toString());
+    dataset.jobProps().setProp(SlaEventKeys.DATASET_URN_KEY, getDatasetUrnFromDataset(dataset));
+  }
+
+  private static String getDatasetUrnFromDataset(Dataset dataset) {
```
zliu41: This implementation is too specific. You can move this method to `Dataset`, so if someone wants to define dataset urn differently, he can override it in a subclass of `Dataset`.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/DatasetsFinder.java
```diff
@@ -0,0 +1,183 @@
...
+    List<String> list = this.state.getPropAsList(MRCompactor.COMPACTION_NORMAL_PRIORITY_TOPICS, StringUtils.EMPTY);
+    return DatasetFilterUtils.getPatternsFromStrings(list);
+  }
+
+  private boolean getRecompactDatasets() {
+    return this.state.getPropAsBoolean(MRCompactor.COMPACTION_RECOMPACT_FROM_DEST_PATHS,
+        MRCompactor.DEFAULT_COMPACTION_RECOMPACT_FROM_DEST_PATHS);
+  }
+
+  public double getDatasetPriority(String datasetName) {
```
zliu41: This should be protected

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/DatasetsFinder.java
```diff
@@ -0,0 +1,183 @@
...
+            + MRCompactor.COMPACTION_LATEDATA_THRESHOLD_FOR_RECOMPACT_PER_DATASET + ".");
+      } else {
+        topicRegexAndRecompactThreshold.put(topicsAndRecompactThreshold.get(0),
+            Double.parseDouble(topicsAndRecompactThreshold.get(1)));
+      }
+    }
+    return topicRegexAndRecompactThreshold;
+  }
+
+  public double getDatasetRecompactThreshold(String datasetName) {
```
zliu41: This should be protected

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -125,8 +114,8 @@
   public static final String COMPACTION_HIGH_PRIORITY_TOPICS = COMPACTION_PREFIX + "high.priority.topics";
   public static final String COMPACTION_NORMAL_PRIORITY_TOPICS = COMPACTION_PREFIX + "normal.priority.topics";
   public static final String COMPACTION_JOBPROPS_CREATOR_CLASS = COMPACTION_PREFIX + "jobprops.creator.class";
-  public static final String DEFAULT_COMPACTION_JOBPROPS_CREATOR_CLASS =
-      "gobblin.compaction.mapreduce.MRCompactorTimeBasedJobPropCreator";
+  public static final String DEFAULT_COMPACTION_JOBPROPS_CREATOR_CLASS = MRCompactorJobPropCreator.class.getName();
```
zliu41: Since now there is only one job prop creator class, you can remove the reflection.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -248,15 +232,10 @@ public MRCompactor(Properties props) throws IOException {
...
-    this.inputSubDir = getInputSubDir();
-    this.inputLateSubDir = getInputLateSubDir();
-    this.destDir = getDestDir();
-    this.destSubDir = getDestSubDir();
-    this.destLateSubDir = getDestLateSubDir();
     this.tmpOutputDir = getTmpOutputDir();
     this.fs = getFileSystem();
-    this.datasets = Sets.newHashSet();
+    this.datasetsFinder = getDatasetsFinder();
+    this.datasets = this.datasetsFinder.findDatasets();
```
zliu41: You can merge this line with the previous.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -585,7 +457,7 @@ private void verifyDataCompleteness() {
   private boolean shouldVerifyCompletenessForDataset(Dataset dataset, List<Pattern> blacklist,
       List<Pattern> whitelist) {
     return !datasetAlreadyCompacted(this.fs, dataset)
-        && DatasetFilterUtils.survived(dataset.topic(), blacklist, whitelist);
+        && DatasetFilterUtils.survived(this.datasetsFinder.findDatasetName(dataset), blacklist, whitelist);
```
zliu41: You can put this method in `Dataset`.

ydai1124:
@zliu41 Updated.

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
```diff
@@ -17,12 +17,15 @@
...
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.fs.Path;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 
 import lombok.extern.slf4j.Slf4j;
 
+import gobblin.compaction.mapreduce.MRCompactor;
```
chavdar: The dataset class should implement the gobblin.data.management.dataset.Dataset interface.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/DatasetsFinder.java
```diff
@@ -0,0 +1,177 @@
...
+import gobblin.configuration.State;
+import gobblin.util.DatasetFilterUtils;
+import gobblin.util.HadoopUtils;
+
+
+/**
+ * {@link Dataset}s finder to identify datasets, using given properties.
+ */
+@Slf4j
+public abstract class DatasetsFinder {
```
chavdar: This should implement gobblin.data.management.retention.dataset.finder.DatasetFinder<T>

chavdar:
@ydai1124 can you please describe the high-level design in the PR description?

gobblin-compaction/src/main/java/gobblin/compaction/Dataset.java
ydai1124: Gobblin compaction dataset is different from the `gobblin.data.management.dataset.Dataset`. It contains the `inputPath` and `outputPath`, which could be different and not suitable for `gobblin.data.management.dataset.Dataset#datasetRoot()`.

ydai1124:
@chavdar Changed both title and description.

ydai1124:
@chavdar @zliu41 Moved `dataset` and `datasetFinder` to `gobblin-api`. @ibuenros Can you also take a look on the last commit?

chavdar:
LGTM

PR #478, opened 2015-11-24
Add row checker final state to task state

zliu41:
Currently fork's row checker final states are added, but task's row checker's are not. This PR fixes it.


zliu41:
@liyinan926 @ibuenros can either of you review?

ibuenros:
LGTM

PR #482, opened 2015-11-24
Refactor of distcp to allow for more complex publish targets and easi…

ibuenros:
…er DatasetFinder implementation


ibuenros:
@liyinan926 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetMetadata.java
```diff
@@ -0,0 +1,63 @@
...
+@EqualsAndHashCode(callSuper=false)
+public class CopyableDatasetMetadata {
+
+  public CopyableDatasetMetadata(CopyableDataset copyableDataset, Path datasetTargetRoot) {
+    this.datasetRoot = copyableDataset.datasetRoot();
+    this.datasetTargetRoot = datasetTargetRoot;
+  }
+
+  @Getter
+  private Path datasetRoot;
```
liyinan926: Seems they can be final.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetMetadata.java
```diff
@@ -0,0 +1,63 @@
...
+  @Getter
+  private Path datasetTargetRoot;
+  private static Gson gson = new Gson();
```
liyinan926: Why not make it a `static final Gson GSON = ...`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -44,20 +44,14 @@
...
   private final Path rootPath;
   private final FileSystem fs;
-  private final Path targetDirectory;
   private final Properties properties;
   private LoadingCache<Path, OwnerAndPermission> ownerAndPermissionCache;
   private final PathFilter pathFilter;
 
   public RecursiveCopyableDataset(final FileSystem fs, Path rootPath, Properties properties) {
 
-    Preconditions.checkArgument(
```
liyinan926: Why this is gone?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -129,6 +137,19 @@ protected FileSystem getSourceFileSystem(State state) throws IOException {
     return FileSystem.get(URI.create(uri), conf);
   }
 
+  private FileSystem getTargetFileSystem(State state) throws IOException {
+    return WriterUtils.getWriterFS(state, 1, 0);
+  }
+
+  private Path getTargetRoot(State state, DatasetFinder<?> datasetFinder, CopyableDataset dataset) {
+    Preconditions.checkArgument(state.contains(ConfigurationKeys.DATA_PUBLISHER_FINAL_DIR));
```
liyinan926: Better to use the second argument to include a message here.

gobblin-data-management/src/main/java/gobblin/data/management/util/PathUtils.java
```diff
@@ -53,4 +57,16 @@ public static Path getPathWithoutSchemeAndAuthority(Path path) {
...
     return new Path(StringUtils.removeStart(path.toString(), Path.SEPARATOR));
   }
+
+  /**
+   * Finds the deepest ancestor of input that is not a glob.
+   */
+  public static Path deepestNonGlobPath(Path input) {
+    Path commonRoot = input;
+
+    while(GLOB_TOKENS.matcher(commonRoot.toString()).find()) {
```
liyinan926: Seems you can also do `GLOB_TOKENS.matcher(commonRoot.getName()).matches()`?

liyinan926: Also the condition is better to be `commonRoot != null && ...`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
ibuenros: It is no longer the responsibility of the Dataset to figure out the target, so it doesn't need this option anymore.

gobblin-data-management/src/main/java/gobblin/data/management/util/PathUtils.java
ibuenros: For the first comment, no, we need to match against entire string. For example `/foo/*/bar` is a glob, even though the name doesn't include a glob. I will fix the second.

ibuenros:
@liyinan926 addressed comments.

liyinan926:
LGTM.

PR #483, opened 2015-11-25
Limit the number of minutes and bytes per log file in LogCopier

zliu41:
@liyinan926 can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
+++ b/gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
@@ -38,6 +38,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.base.Stopwatch;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
@@ -50,6 +51,7 @@ import gobblin.configuration.ConfigurationKeys;
 import gobblin.util.DatasetFilterUtils;
 import gobblin.util.ExecutorsUtils;
 import gobblin.util.FileListUtils;
+import gobblin.util.HadoopUtils;
 
 
 /**
@@ -77,6 +79,8 @@ import gobblin.util.FileListUtils;
  *           .writeTo(new Path(destLogDir))
  *           .useInitialDelay(60)
  *           .useCopyInterval(60)
+ *           .useMaxMinutesPerLogFile(60)
+ *           .useMaxBytesPerLogFile(1024 * 1024)
  *           .useTimeUnit(TimeUnit.SECONDS)
  *           .build();
  *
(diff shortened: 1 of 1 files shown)


gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -407,23 +443,49 @@ public LogCopier build() {
...
+      this.destLogDir = destLogFile.getParent();
+      this.destLogFileBaseName = FilenameUtils.removeExtension(destLogFile.getName());
+      this.destLogFileExtension = FilenameUtils.getExtension(destLogFile.getName());
+      this.currentDestLogFile = getCurrentDestLogFile();
+      this.watch.start();
+    }
+
+    private Path getCurrentDestLogFile() {
+      return new Path(this.destLogDir,
+          this.destLogFileBaseName + "." + System.currentTimeMillis() + "." + this.destLogFileExtension);
```
liyinan926: I suggest putting the timestamp at the end.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -407,23 +443,49 @@ public LogCopier build() {
...
+    private Path currentDestLogFile;
 
     // The task maintains the current source log file position itself
     private long currentPos = 0;
 
     LogCopyTask(Path srcLogFile, Path destLogFile) {
       this.srcLogFile = srcLogFile;
-      this.destLogFile = destLogFile;
+      this.destLogDir = destLogFile.getParent();
+      this.destLogFileBaseName = FilenameUtils.removeExtension(destLogFile.getName());
```
liyinan926: You can use Guava's `Files` utility class to replace `FilenameUtils`.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
```diff
@@ -407,23 +443,49 @@ public LogCopier build() {
...
+        LOGGER.debug(String.format("Copying changes from %s to %s", this.srcLogFile, this.currentDestLogFile));
+        copyChangesOfLogFile(srcFs.makeQualified(this.srcLogFile), destFs.makeQualified(this.currentDestLogFile));
       } catch (IOException ioe) {
-        LOGGER.error(String.format("Failed while copying logs from %s to %s", this.srcLogFile, this.destLogFile), ioe);
+        LOGGER.error(String.format("Failed while copying logs from %s to %s", this.srcLogFile, this.currentDestLogFile),
+            ioe);
+      }
+    }
+
+    private void getNextLogFileIfNeeded() throws IOException {
```
liyinan926: You will also need to reset `this.currentPos`.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
liyinan926: I think a better approach is to always name the current log file as normal like `GobblinApplicationMaster.stdout` and rename it to something like `GobblinApplicationMaster.stdout.<timestamp>` when necessary and re-create a new empty file named `GobblinApplicationMaster.stdout`.

gobblin-utility/src/main/java/gobblin/util/logs/LogCopier.java
zliu41: currentPos is for the source file, not dest file.

zliu41:
@liyinan926 updated 94026e1

zliu41:
@liyinan926 updated 6c87cc8

liyinan926:
LGTM.

PR #485, opened 2015-11-26
ETL-3765: Report record count after compaction is done

ydai1124:
@zliu41 Can you review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
@@ -16,7 +16,6 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 
 import org.apache.commons.io.FilenameUtils;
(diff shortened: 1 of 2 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -504,7 +503,7 @@ public int compareTo(MRCompactorJobRunner o) {
     return Double.compare(o.dataset.priority(), this.dataset.priority());
   }
 
-  private List<Path> getCumulativeLateFilePaths(Path lateDataDir) throws FileNotFoundException, IOException {
+  private List<Path> getApplicableFilePaths(Path lateDataDir) throws FileNotFoundException, IOException {
```
zliu41: Not sure what this method name means. Applicable to what? Please use a more meaningful name and/or add javadoc.
Also if it is not for late data, the parameter should not be called `lateDataDir`

zliu41: Also `FileNotFoundException` is redundant

ydai1124:
@zliu41 Updated.

zliu41:
LGTM

Issue #486, opened 2015-11-27
MRCompactorAvroKeyDedupJobRunner and schema changes


mwol:
Input avro schema is set at job level (MRCompactorAvroKeyDedupJobRunner#configureSchema), this should not be the case, as the avro file's own schema will be ignored. e.g. I ran into an error, where I could not even read a file any more, after changing one data type from ["null","string] to ["null","long"] and had mixed input files.
Changes like this are not common and should not occur frequently but anyway, the error should happen when trying to merge files together and not during the reading process, this would give the opportunity to implement some merging/conversion strategy.

```
2015-11-26 17:00:38,924 WARN [main] org.apache.hadoop.mapred.YarnChild: Exception running child : org.apache.avro.AvroTypeException: Found string, expecting union
	at org.apache.avro.io.ResolvingDecoder.doAction(ResolvingDecoder.java:292)
	at org.apache.avro.io.parsing.Parser.advance(Parser.java:88)
	at org.apache.avro.io.ResolvingDecoder.readIndex(ResolvingDecoder.java:267)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:155)
	at org.apache.avro.generic.GenericDatumReader.readField(GenericDatumReader.java:193)
	at org.apache.avro.reflect.ReflectDatumReader.readField(ReflectDatumReader.java:230)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:183)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:151)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:142)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:233)
	at org.apache.avro.mapreduce.AvroRecordReaderBase.nextKeyValue(AvroRecordReaderBase.java:118)
	at org.apache.avro.mapreduce.AvroKeyRecordReader.nextKeyValue(AvroKeyRecordReader.java:53)
	at gobblin.compaction.mapreduce.avro.AvroKeyCombineFileRecordReader.nextKeyValue(AvroKeyCombineFileRecordReader.java:63)
	at org.apache.hadoop.mapreduce.lib.input.CombineFileRecordReader.nextKeyValue(CombineFileRecordReader.java:69)
	at org.apache.hadoop.mapred.MapTask$NewTrackingRecordReader.nextKeyValue(MapTask.java:556)
[...]

PR #487, opened 2015-11-30
MRCompactorAvroKeyDedupJobRunner and schema changes #486

mwol:
A proposal for a solution to https://github.com/linkedin/gobblin/issues/486.
I implemented new class AvroCombineFileSplit which stores the avro schema for each split, determined by the corresponding input file.
I added property "avro.single.input.schema" to keep old behaviour.

What do you think?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
@@ -22,8 +22,8 @@ import java.util.Comparator;
 import java.util.List;
 
 import org.apache.avro.Schema;
-import org.apache.avro.SchemaCompatibility;
 import org.apache.avro.Schema.Field;
+import org.apache.avro.SchemaCompatibility;
 import org.apache.avro.SchemaCompatibility.SchemaCompatibilityType;
 import org.apache.avro.mapred.AvroKey;
 import org.apache.avro.mapred.AvroValue;
@@ -64,6 +64,12 @@ public class MRCompactorAvroKeyDedupJobRunner extends MRCompactorJobRunner {
   private static final String COMPACTION_JOB_PREFIX = "compaction.job.";
 
   /**
+   * If true, the latest schema, determined from the input files, will be used as single schema for all input files,
+   * otherwise, the avro each input file will be determined and splits will be created with respect to the input file's schema
+   */
+  private static final String COMPACTION_JOB_AVRO_SINGLE_INPUT_SCHEMA = COMPACTION_JOB_PREFIX + "avro.single.input.schema";
+
+  /**
    * Properties related to the avro dedup compaction job of a dataset.
    */
   private static final String COMPACTION_JOB_AVRO_KEY_SCHEMA_LOC = COMPACTION_JOB_PREFIX + "avro.key.schema.loc";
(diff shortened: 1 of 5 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
```diff
@@ -99,7 +108,9 @@ protected void configureJob(Job job) throws IOException {
 
   private void configureSchema(Job job) throws IOException {
     Schema newestSchema = getNewestSchemaFromSource(job);
-    AvroJob.setInputKeySchema(job, newestSchema);
+    if (useSingleInputSchema) {
+      AvroJob.setInputKeySchema(job, newestSchema);
+    }
     AvroJob.setMapOutputKeySchema(job, this.shouldDeduplicate ? getKeySchema(job, newestSchema) : newestSchema);
```
zliu41: If `MapOutputKeyValueSchema` and `MapOutputValueSchema` is still `newestSchema`, I guess the error will still happen in the mapper (instead of RecordReader)? I think you can use `context.getInputSplit().getSchema()` in the mapper as the mapper output key/value schema.

mwol: It is true, you still get an error in the Mapper, something like ` org.apache.avro.UnresolvedUnionException: Not in union ["null","long"]: <someStringData>` but anyway, the type of error is different and therefore solvable. Before, the error occurred while reading the data files. For my special case, I solved it by just adding some lines of code to the map() method to handle the problem (`final Object jp = key.datum().get("jp"); if (jp instanceof String) { key.datum().put("jp", Long.valueOf((String) jp));}`), for general solutions I would suggest some pluggable mechanism (e.g. sub classing `TaskInputOutputContext`) and maybe one could use version information also from AvroSchema registry, if present?

Hmm, I see you point, at first I thought it is still necessary to use `newestSchema` as  mapper output key/value schema, due to deduping, but it is true for map only jobs. 
For deduping, the error would have to be handled in the reducer, though (if the changed field is a key field). 
Maybe it is better to have one common schema for the compacted data?

mwol: I added the change in the `AvroKeyMapper (`keySchema = ((AvroCombineFileSplit) context.getInputSplit()).getSchema();`) to my commit

mwol: hmm, as far as I see, you can only easily set the outputkey schema in the Mapper, but for deduplication, this can differ from the inputschema due to `AvroJob.setMapOutputKeySchema(job, this.shouldDeduplicate ? getKeySchema(job, newestSchema) : newestSchema);` in `MRCompactorAvroKeyDedupJobRunner`.

zliu41: Makes sense

zliu41:
Since the mapper will fail if there's an incompatible schema, can you print the file paths that cause problems in the mapper, like `throw new IOException("Unable to process paths " + ((CombineFileSplit) context.getInputSplit()).getPaths(), e);`
Because the Avro exception message doesn't contain the file path, which can be annoying to debug.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/MRCompactorAvroKeyDedupJobRunner.java
```diff
@@ -87,8 +93,11 @@
 
   private static final DedupKeyOption DEFAULT_DEDUP_KEY_OPTION = DedupKeyOption.KEY;
 
+  private final boolean useSingleInputSchema;
+
   public MRCompactorAvroKeyDedupJobRunner(Dataset dataset, FileSystem fs, Double priority) {
     super(dataset, fs, priority);
+    this.useSingleInputSchema = this.dataset.jobProps().getPropAsBoolean(COMPACTION_JOB_AVRO_SINGLE_INPUT_SCHEMA);
```
zliu41: Please add a default value `false`, which improves readability.

mwol: done ...

mwol:
ok, I added the changes, I also added serialization for schema `AvroCombineFileSplit` to make it accessible in the Mapper, as you suggested.

PR #489, opened 2015-11-30
Fixing gobblin-yarn.sh conf file location

sahilTakiar:
@liyinan926 can you review?


liyinan926:
LGTM.

bin/gobblin-yarn.sh
```diff
@@ -50,9 +50,10 @@ function stop() {
 
 FWDIR="$(cd `dirname $0`/..; pwd)"
 FWDIR_LIB=${FWDIR}/lib
-FWDIR_CONF=${FWDIR}/conf
+FWDIR_CONF=${FWDIR}/yarn/conf
```
kadaan: Why is this yarn/conf not conf/yarn.  Especially when the yarn config files are at conf/yarn

PR #491, opened 2015-12-01
Moved logic in RecursiveCopyableDataset to a CopyableDataset builder.

ibuenros:
This PR creates a CopyableFile builder to reuse logic in different CopyableDatasets and fixes some bugs with copy operations.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.copy;
+
+import lombok.Data;
+
+import org.apache.hadoop.fs.Path;
+
+
+/**
+ * Configuration for Gobblin distcp jobs.
+ */
+@Data
+public class CopyConfiguration {
+
+  /**
+   * Directory where dataset should be replicated.
+   * This directory corresponds to the {@link CopyableDataset#datasetRoot} in the new location.
+   */
+  private final Path targetRoot;
+  /**
+   * Preserve options passed by the user.
+   */
+  private final PreserveAttributes preserve;
+  /**
+   * {@link CopyContext} for this job.
+   */
+  private final CopyContext copyContext;
+
+}
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
@@ -37,15 +37,21 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 
 import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
+import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 
 
 /**
@@ -59,6 +65,7 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
   private static final String COPY_PREFIX = "gobblin.copy";
   public static final String SERIALIZED_COPYABLE_FILES = COPY_PREFIX + ".serialized.copyable.files";
   public static final String SERIALIZED_COPYABLE_DATASET = COPY_PREFIX + ".serialized.copyable.datasets";
+  public static final String PRESERVE_ATTRIBUTES_KEY = COPY_PREFIX + ".preserved.attributes";
 
   /**
    * <ul>
@@ -81,6 +88,8 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
   public List<WorkUnit> getWorkunits(SourceState state) {
 
     List<WorkUnit> workUnits = Lists.newArrayList();
+    CopyContext copyContext = new CopyContext();
+
     try {
 
       DatasetFinder<CopyableDataset> datasetFinder =
@@ -93,8 +102,11 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
 
         Path targetRoot = getTargetRoot(state, datasetFinder, copyableDataset);
 
-        Collection<Partition<CopyableFile>> partitions =
-            copyableDataset.partitionFiles(copyableDataset.getCopyableFiles(targetFs, targetRoot));
+        CopyConfiguration copyConfiguration = new CopyConfiguration(targetRoot,
+            PreserveAttributes.fromMnemonicString(state.getProp(PRESERVE_ATTRIBUTES_KEY)), copyContext);
+
+        Collection<CopyableFile> files = copyableDataset.getCopyableFiles(targetFs, copyConfiguration);
+        Collection<Partition<CopyableFile>> partitions = partitionCopyableFiles(files);
 
         for (Partition<CopyableFile> partition : partitions) {
           Extract extract = new Extract(Extract.TableType.SNAPSHOT_ONLY, COPY_PREFIX, partition.getName());
@@ -106,7 +118,7 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
             GobblinMetrics.addCustomTagToState(workUnit, new Tag<String>(
                 CopyEventSubmitterHelper.DATASET_ROOT_METADATA_NAME, copyableDataset.datasetRoot().toString()));
             workUnit.setProp(SlaEventKeys.DATASET_URN_KEY, copyableDataset.datasetRoot().toString());
-            workUnit.setProp(SlaEventKeys.PARTITION_KEY, partition.getName());
+            workUnit.setProp(SlaEventKeys.PARTITION_KEY, copyableFile.getPartition());
             workUnit.setProp(SlaEventKeys.ORIGIN_TS_IN_MILLI_SECS_KEY, copyableFile.getFileStatus().getModificationTime());
             workUnits.add(workUnit);
           }
@@ -187,4 +199,21 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
   public static CopyableDatasetMetadata deserializeCopyableDataset(State state) throws IOException {
     return CopyableDatasetMetadata.deserialize(state.getProp(SERIALIZED_COPYABLE_DATASET));
   }
+
+  private Collection<Partition<CopyableFile>> partitionCopyableFiles(Collection<CopyableFile> files) {
+    Map<String, Partition.Builder<CopyableFile>> partitionBuildersMaps = Maps.newHashMap();
+    for (CopyableFile file : files) {
+      if (!partitionBuildersMaps.containsKey(file.getPartition())) {
+        partitionBuildersMaps.put(file.getPartition(), new Partition.Builder<CopyableFile>(file.getPartition()));
+      }
+      partitionBuildersMaps.get(file.getPartition()).add(file);
+    }
+    return Lists.newArrayList(Iterables.transform(partitionBuildersMaps.values(),
+        new Function<Partition.Builder<CopyableFile>, Partition<CopyableFile>>() {
+          @Nullable @Override public Partition<CopyableFile> apply(Partition.Builder<CopyableFile> input) {
+            return input.build();
+          }
+        }));
+  }
+
 }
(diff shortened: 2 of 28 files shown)


ibuenros:
@liyinan926 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
   /** Checksum of the origin file. */
   private byte[] checksum;
 
+  public static Builder builder(FileSystem originFs, FileStatus origin, Path targetRoot, Path datasetRoot) {
```
liyinan926: I think this defeats the purpose of having a builder class, which is to avoid having constructor or factory class that have too many arguments. Why not simply have a method that returns a plain `Builder` instance, like `public static Builder builder`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+
+    return _hiddenBuilder().originFS(originFs).rootPath(datasetRoot).
+        origin(origin).destination(targetPath).relativeDestination(relativePath);
+  }
+
+  public static class Builder {
+
+    private FileSystem originFs;
+    private Path rootPath;
+    private static Cache<Path, OwnerAndPermission> OWNER_AND_PERMISSION_CACHE;
```
liyinan926: No all capital unless this is a `static final`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+
+    private synchronized static Cache<Path, OwnerAndPermission> ownerAndPermissionCache() {
+      if(OWNER_AND_PERMISSION_CACHE == null) {
```
liyinan926: Empty space after `if`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+
+    public Builder rootPath(Path rootPath) {
+      this.rootPath = rootPath;
+      return this;
+    }
+
+    public CopyableFile build() throws IOException {
+
+      if(this.destinationOwnerAndPermission == null) {
+        this.destinationOwnerAndPermission =
```
liyinan926: It is possible that `origin` is `null`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+            new OwnerAndPermission(this.origin.getOwner(), this.origin.getGroup(), this.origin.getPermission());
+      }
+      if(this.ancestorsOwnerAndPermission == null) {
+        this.ancestorsOwnerAndPermission = replicateOwnerAndPermission(this.originFs, this.origin.getPath());
```
liyinan926: Ditto, it is possible that `originalFs` is null.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+      }
+      if(this.checksum == null) {
```
liyinan926: Empty space after `if`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -68,6 +78,83 @@
...
+          ancestorOwnerAndPermissions.add(ownerAndPermissionCache().get(originFs.makeQualified(currentPath),
+              new Callable<OwnerAndPermission>() {
+            @Override public OwnerAndPermission call() throws Exception {
+              FileStatus fs = originFs.getFileStatus(thisPath);
+              return new OwnerAndPermission(fs.getOwner(), fs.getGroup(), fs.getPermission());
+            }
+          }));
+        }
+      } catch (ExecutionException ee) {
+        // When cache loader failed.
```
liyinan926: Better to take some action instead of swallowing the exception.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -178,31 +178,43 @@ public static void renameRecursively(FileSystem fileSystem, Path from, Path to)
...
-         * will be created by inheriting it's parent permission (/data/gobblin/)
-         */
-        if (!fileSystem.exists(toFilePath.getParent())) {
-          fileSystem.mkdirs(toFilePath.getParent());
-        }
-
-        if (!fileSystem.rename(fromFile.getPath(), toFilePath)) {
-          throw new IOException(String.format("Failed to rename %s to %s.", fromFile.getPath(), toFilePath));
+      if (!safeRenameIfNotExists(fileSystem, fromFile.getPath(), toFilePath)) {
+        if(fromFile.isDir()) {
```
liyinan926: `Path` has a method `isDir()`?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -178,31 +178,43 @@ public static void renameRecursively(FileSystem fileSystem, Path from, Path to)
...
-         * When the directory Log is being renamed from source to destination, task-output directory may not exist. It
-         * will be created by inheriting it's parent permission (/data/gobblin/)
-         */
-        if (!fileSystem.exists(toFilePath.getParent())) {
-          fileSystem.mkdirs(toFilePath.getParent());
-        }
-
-        if (!fileSystem.rename(fromFile.getPath(), toFilePath)) {
-          throw new IOException(String.format("Failed to rename %s to %s.", fromFile.getPath(), toFilePath));
+      if (!safeRenameIfNotExists(fileSystem, fromFile.getPath(), toFilePath)) {
```
liyinan926: I saw you use `synchronized` for `safeRenameIfNotExists`. But this `if` block is a check-and-act compound operation that will not be thread safe even if `safeRenameIfNotExists` is.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
ibuenros: fromFile is a FileStatus

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
ibuenros: It's fine, I believe. `safeRenameIfNotExists` returns false if the directory already exists, in which case we will try with the child directory. Assuming that directories are only being created at the target, and not deleted, this should eliminate race conditions.

liyinan926:
@ibuenros seems like not all my comments have been addressed. For example, I'm still seeing `if(` and `for(`.

ibuenros:
@liyinan926 addressed previous comments

ibuenros:
addressed more comments (although the comments seem to have disappeared from the pull request)

gobblin-data-management/src/main/java/gobblin/data/management/copy/PreserveAttributes.java
```diff
@@ -0,0 +1,77 @@
...
+   * * r -> preserve replication
+   * * b -> preserve block size
+   * * u -> preserve owner
+   * * g -> preserve group
+   * * p -> preserve permissions
+   *
+   * @param s String of the form [rbugp]
+   * @return Parsed {@link PreserveAttributes}
+   */
+  public static PreserveAttributes fromString(String s) {
```
chavdar: nit: I suggest renaming to something like fromMnemonicString(String s) and also have the reverse: toMnemonicString()

gobblin-data-management/src/main/java/gobblin/data/management/copy/PreserveAttributes.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Parse {@link PreserveAttributes} from a string of the form [rbugp]:
+   * * r -> preserve replication
+   * * b -> preserve block size
+   * * u -> preserve owner
+   * * g -> preserve group
+   * * p -> preserve permissions
+   *
+   * @param s String of the form [rbugp]
```
chavdar: Are the brackets part of the string? Should we also have the shortcut "a" for all?

gobblin-data-management/src/main/java/gobblin/data/management/copy/PreserveAttributes.java
```diff
@@ -0,0 +1,77 @@
...
+   * @return Parsed {@link PreserveAttributes}
+   */
+  public static PreserveAttributes fromString(String s) {
+    if (Strings.isNullOrEmpty(s)) {
+      return new PreserveAttributes(0);
+    }
+    int value = 0;
+    for (Option option : Option.values()) {
+      if (s.indexOf(option.token) > 0) {
```
chavdar: >= 0

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.hadoop.fs.Path;
+
+
+/**
+ * Configuration for Gobblin distcp jobs.
+ */
+@Data
+public class CopyConfiguration {
+
+  private final Path target;
```
chavdar: Is the target a directory, a file, or both? Some javadocs might be worthwhile.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -93,8 +101,11 @@
 
         Path targetRoot = getTargetRoot(state, datasetFinder, copyableDataset);
 
-        Collection<Partition<CopyableFile>> partitions =
-            copyableDataset.partitionFiles(copyableDataset.getCopyableFiles(targetFs, targetRoot));
+        CopyConfiguration copyConfiguration = new CopyConfiguration(targetRoot,
```
chavdar: Is the PRESERVE_ATTRIBUTES_KEY mandatory? Can't we have a sensible default (""), i.e. don't preserve anything?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -187,4 +198,21 @@ public static void serializeCopyableDataset(State state, CopyableDatasetMetadata
...
+
+  private Collection<Partition<CopyableFile>> partitionCopyableFiles(Collection<CopyableFile> files) {
+    Map<String, Partition.Builder<CopyableFile>> partitionBuildersMaps = Maps.newHashMap();
+    for (CopyableFile file : files) {
+      if (!partitionBuildersMaps.containsKey(file.getPartition())) {
+        partitionBuildersMaps.put(file.getPartition(), new Partition.Builder<CopyableFile>(file.getPartition()));
+      }
+      partitionBuildersMaps.get(file.getPartition()).add(file);
+    }
+    return Lists.newArrayList(Iterables.transform(partitionBuildersMaps.values(),
```
chavdar: IMHO trying to use java as python is ugly. :-)

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDataset.java
```diff
@@ -25,7 +25,7 @@
 /**
  * {@link Dataset} that supports finding {@link CopyableFile}s.
  */
-public interface CopyableDataset extends PartitionableDataset<CopyableFile> {
+public interface CopyableDataset extends Dataset {
 
   /**
    * Find all {@link CopyableFile}s in this dataset.
```
chavdar: The javadoc needs to be made more precise? Can the result contain directories? Shouldn't the return result be a Collection ? I can't think of a case where the order is important.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -113,4 +221,30 @@ public static CopyableFile deserialize(String serialized) throws IOException {
...
+  @EqualsAndHashCode
+  public static class DatasetAndPartition {
+    private final CopyableDatasetMetadata dataset;
+    private final String partition;
+
+    /**
+     * @return a unique string identifier for this {@link DatasetAndPartition}.
+     */
+    public String identifier() {
+      return Integer.toString(Math.abs(this.dataset.hashCode() + this.partition.hashCode()));
```
chavdar: This does not look unique to me. We should use UUID instead.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -67,6 +80,101 @@
...
+    return _hiddenBuilder().originFS(originFs).rootPath(datasetRoot).
+        origin(origin).destination(targetPath).relativeDestination(relativePath).
+        preserve(copyConfiguration.getPreserve());
+  }
+
+  public static class Builder {
+
+    private FileSystem originFs;
+    private Path rootPath;
+    private static Cache<Path, OwnerAndPermission> ownerAndPermissionCache;
```
chavdar: No static objects please as these have unpredictable side effects especially if we start reusing containers in YARN. You may need to create a CopyContext object to hold such state.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
ibuenros: It is not mandatory. If it is null it will default to preserve nothing (this is done in PreserveAttributes itself).

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
ibuenros: What if I'm trying to use Scala? :P

gobblin-data-management/src/main/java/gobblin/data/management/copy/PreserveAttributes.java
ibuenros: is "all" common enough to warrant this? I would prefer users be aware what "all" means by having to include all 5 characters.

chavdar: distcp does not have so I guess we don't need it.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
chavdar: I think the result is that it looks like Perl -- a sequence of seemingly random characters. :-)

ibuenros:
@chavdar addressed comments

liyinan926:
LGTM.

PR #492, opened 2015-12-01
Allow uses to specify a dir for yarn .job files

sahilTakiar:
Changed `gobblin.yarn.job.conf.package.path` to `gobblin.yarn.job.conf.dir`. This parameter now takes in a directory rather than requiring a tarball.

Internally, Gobblin will take the directory and package it into a tarball and submit it as an `ARCHIVE` resource to YARN. The code to make the tarball is in `StreamUtils`.

@liyinan926 can you review?


gobblin-utility/src/main/java/gobblin/util/io/StreamUtils.java
```diff
@@ -96,4 +109,102 @@ public static long copy(ReadableByteChannel inputChannel, WritableByteChannel ou
...
+  /**
+   * Similiar to {@link #tar(FileSystem, Path, Path)} except the source and destination {@link FileSystem} can be different.
+   *
+   * @see {@link #tar(FileSystem, Path, Path)}
+   */
+  public static void tar(FileSystem sourceFs, FileSystem destFs, Path sourcePath, Path destPath)
+      throws IOException {
+    try (FSDataOutputStream fsDataOutputStream = destFs.create(destPath);
+        TarArchiveOutputStream tarArchiveOutputStream = new TarArchiveOutputStream(
+            new GzipCompressorOutputStream(fsDataOutputStream), Charsets.UTF_8.name())) {
```
liyinan926: You can use `ConfigurationKeys.DEFAULT_CHARSET_ENCODING`.

gobblin-utility/src/main/java/gobblin/util/io/StreamUtils.java
```diff
@@ -96,4 +109,102 @@ public static long copy(ReadableByteChannel inputChannel, WritableByteChannel ou
...
+
+      FileStatus fileStatus = sourceFs.getFileStatus(sourcePath);
+
+      if (fileStatus.isDir()) {
+        dirToTarArchiveOutputStreamRecursive(fileStatus, sourceFs, new File("."), tarArchiveOutputStream);
+      } else {
+        try (FSDataInputStream fsDataInputStream = sourceFs.open(sourcePath)) {
+          fileToTarArchiveOutputStream(fileStatus, fsDataInputStream, new File(sourcePath.getName()),
```
liyinan926: So you are assuming `sourcePath` is a local file?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -536,8 +537,8 @@ private void addAppRemoteFiles(String hdfsFileList, Map<String, LocalResource> r
   private void addJobConfPackage(String jobConfPackagePath, Path destDir, Map<String, LocalResource> resourceMap)
       throws IOException {
     Path srcFilePath = new Path(jobConfPackagePath);
-    Path destFilePath = new Path(destDir, srcFilePath.getName());
-    this.fs.copyFromLocalFile(srcFilePath, destFilePath);
+    Path destFilePath = new Path(destDir, srcFilePath.getName() + ".tar.gz");
```
liyinan926: It's worth defining `.tar.gz` as a constant in `GobblinYarnConfigurationKeys`.

gobblin-yarn/src/main/java/gobblin/yarn/JobConfigurationManager.java
```diff
@@ -62,6 +63,7 @@ public JobConfigurationManager(EventBus eventBus, Optional<String> jobConfPackag
   protected void startUp() throws Exception {
     if (this.jobConfPackagePath.isPresent()) {
       File path = new File(this.jobConfPackagePath.get());
+      path = new File(path.getParent(), path.getName() + ".tar.gz");
```
liyinan926: Why you need to append `.tar.gz`?

gobblin-utility/src/test/java/gobblin/util/io/StreamUtilsTest.java
```diff
@@ -0,0 +1,123 @@
+package gobblin.util.io;
```
liyinan926: Open source header missing.

gobblin-utility/src/main/java/gobblin/util/io/StreamUtils.java
sahilTakiar: `sourcePath` and `destPath` can be either HDFS or local files depending on the `FileSystem` implementation. The reason we have to make a `File` object for `sourcePath` is that `TarArchiveOutputStream` uses `File` objects to make new entries in the tarball.

e.g. `tarArchiveOutputStream.putArchiveEntry(new TarArchiveEntry(new File("."));`

gobblin-yarn/src/main/java/gobblin/yarn/JobConfigurationManager.java
sahilTakiar: The config `gobblin.yarn.job.conf.dir` will specify a directory that contains the `.job` files. Then `StreamUtils.tar` will create a tarball with the name specified by the config + `.tar.gz` at the end. So the `ARCHIVE` added in the `GobblinYarnAppLauncher` is `config.getString(GobblinYarnConfigurationKeys.JOB_CONF_DIR_KEY) + .tar.gz`.

So in the AppMaster we need a append `.tar.gz` to the end of the name specified by `JOB_CONF_DIR_KEY`.

sahilTakiar:
@liyinan926 comments addressed

sahilTakiar:
@liyinan926 comments addressed

liyinan926:
One more comment: since it `gobblin.yarn.job.conf.dir` may be used to specify a single job config file, it may make more sense renaming it to `gobblin.yarn.job.conf.path` and add Javadoc explaining what it means, otherwise LGTM.

sahilTakiar:
@liyinan926 updated. I also made a few changes to `GobblinYarnConfigurationKeys` so that the javadocs read better. Mainly I moved some of the block comments to inline comments so that they don't show up in the javadocs (the block comments were showing up in the javadocs but only for the first config key, so I don't think it makes much sense to include them there).

PR #493, opened 2015-12-01
Make it possible for RowLevelErrFileWriter to write to HDFS

zliu41:
This PR fixes two issues:
1. Originally `RowLevelErrFileWriter` can only write to local FS.
2. Originally multiple tasks write the error records into the same err file. Now task id is added to the err file name to avoid this.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
+++ b/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
@@ -14,10 +14,11 @@ package gobblin.qualitychecker.row;
 
 import java.io.BufferedWriter;
 import java.io.Closeable;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
 import com.google.common.io.Closer;
@@ -32,17 +33,24 @@ import gobblin.configuration.ConfigurationKeys;
  * @author stakiar
  */
 public class RowLevelErrFileWriter implements Closeable {
+  private final FileSystem fs;
   private final Closer closer = Closer.create();
   private BufferedWriter writer;
+
+  public RowLevelErrFileWriter(FileSystem fs) {
+    this.fs = fs;
+  }
 
   /**
    * Open a BufferedWriter
    * @param errFilePath path to write the file
    */
   public void open(Path errFilePath) throws IOException {
-    this.writer =
-        closer.register(new BufferedWriter(new OutputStreamWriter(new FileOutputStream(errFilePath.toString()),
-            ConfigurationKeys.DEFAULT_CHARSET_ENCODING)));
+    this.fs.mkdirs(errFilePath.getParent());
+    OutputStream os =
+        this.closer.register(this.fs.exists(errFilePath) ? this.fs.append(errFilePath) : this.fs.create(errFilePath));
+    this.writer = this.closer
+        .register(new BufferedWriter(new OutputStreamWriter(os, ConfigurationKeys.DEFAULT_CHARSET_ENCODING)));
   }
 
   /**
--- a/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
+++ b/gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
@@ -16,26 +16,33 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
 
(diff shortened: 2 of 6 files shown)


zliu41:
@sahilTakiar or @liyinan926 can you review?

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
```diff
@@ -61,6 +68,15 @@ public boolean executePolicies(Object record, RowLevelPolicyCheckResults results
     return true;
   }
 
+  private Path getErrFilePath(RowLevelPolicy policy) {
+    String errFileName = policy.toString().replaceAll("\\.", "-");
```
chavdar: I am not sure why "." is escaped. Regardless, we should also escape ":" and "/" since those are not allowed in HDFS file names. We should probably have a utility function to do that.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
```diff
@@ -16,26 +16,33 @@
...
+  private final State state;
+  private final FileSystem fs;
   private boolean errFileOpen;
   private final Closer closer;
   private RowLevelErrFileWriter writer;
 
-  public RowLevelPolicyChecker(List<RowLevelPolicy> list) {
+  public RowLevelPolicyChecker(List<RowLevelPolicy> list, State state) throws IOException {
     this.list = list;
+    this.state = state;
```
chavdar: Can you just save the Id instead of the entire state? id seems to get the only thing used.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
```diff
@@ -61,6 +68,15 @@ public boolean executePolicies(Object record, RowLevelPolicyCheckResults results
     return true;
   }
 
+  private Path getErrFilePath(RowLevelPolicy policy) {
```
chavdar: Please add a unit test for this.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
zliu41: `this.fs` is also obtained from the state

chavdar: Same comment still applies. :-)

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
```diff
@@ -32,17 +33,24 @@
...
+
   /**
    * Open a BufferedWriter
    * @param errFilePath path to write the file
    */
   public void open(Path errFilePath) throws IOException {
-    this.writer =
-        closer.register(new BufferedWriter(new OutputStreamWriter(new FileOutputStream(errFilePath.toString()),
-            ConfigurationKeys.DEFAULT_CHARSET_ENCODING)));
+    this.fs.mkdirs(errFilePath.getParent());
```
sahilTakiar: Shouldn't we check if this path exists before executing mkdirs.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
```diff
@@ -32,17 +33,24 @@
...
+    OutputStream os =
+        this.closer.register(this.fs.exists(errFilePath) ? this.fs.append(errFilePath) : this.fs.create(errFilePath));
```
sahilTakiar: Do appends work on our clusters? I was under the impression that at some point it was disabled.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
zliu41: mkdirs will ignore existing dir.

gobblin-core/src/main/java/gobblin/qualitychecker/row/RowLevelErrFileWriter.java
zliu41: Since this is open source we can keep it. Append is not supposed to happen, since now `errFilePath` contains task state ID, so each task should write to a unique err file, unless someone uses the same row level policy twice in the same task.

zliu41:
Updated. 9f16bcb

zliu41:
@chavdar @sahilTakiar any other comments?

chavdar:
LGTM

PR #494, opened 2015-12-01
Resolve race condition between writer and metric updater

zliu41:
There is a race condition between `FsDataWriter.addRecordCountToFileName()` and `AvroHdfsDataWriter.bytesWritten()`. Between `if (!this.fs.exists(this.outputFile))` and `return this.fs.getFileStatus(this.outputFile).getLen()`, the output file could be renamed (adding row count to file name). Adding synchronization to resolve this. 

Also, reset `this.outputFile` to new path after renaming.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
@@ -89,7 +89,7 @@ public class AvroHdfsDataWriter extends FsDataWriter<GenericRecord> {
   }
 
   @Override
-  public long bytesWritten() throws IOException {
+  public synchronized long bytesWritten() throws IOException {
     if (!this.fs.exists(this.outputFile)) {
       return 0;
     }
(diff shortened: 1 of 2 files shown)


zliu41:
@liyinan926 can you review?

gobblin-core/src/main/java/gobblin/writer/AvroHdfsDataWriter.java
```diff
@@ -89,7 +89,7 @@ public long recordsWritten() {
   }
 
   @Override
-  public long bytesWritten() throws IOException {
+  public synchronized long bytesWritten() throws IOException {
```
liyinan926: `synchronized` should also be applied to `FsDataWriter.bytesWritten`.

zliu41: `FsDataWrtiter` does not have `bytesWritten`. `SimpleDataWriter.bytesWritten()` doesn't need synchronization because it doesn't check the output file.

liyinan926:
LGTM.

PR #498, opened 2015-12-03
Fixed gobblin yarn job config path

kadaan:
Fixed config file to account for recent change in property name of gobblin.yarn.job.conf.package.path to gobblin.yarn.job.conf.path


sahilTakiar:
This bug was introduced in https://github.com/linkedin/gobblin/pull/492, I just fixed it in https://github.com/linkedin/gobblin/pull/499

PR #500, opened 2015-12-03
Fixing task reporting of bytes written

sahilTakiar:
* Fixing reporting for byte level metrics in `TaskState`.
* Fix applies to `BYTES` and `BYTES_PER_SECOND` metrics. 
* Changed the method `TaskState.updateByteMetrics` to follow the same logic as `TaskState.updateRecordMetrics`.

@liyinan926 and @ibuenros can you review

### DIFF
--- a/gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedDataWriterBase.java
+++ b/gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedDataWriterBase.java
@@ -16,6 +16,8 @@ package gobblin.instrumented.writer;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import com.codahale.metrics.Meter;
@@ -25,6 +27,8 @@ import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
+import lombok.extern.slf4j.Slf4j;
+
 import gobblin.configuration.State;
 import gobblin.instrumented.Instrumentable;
 import gobblin.instrumented.Instrumented;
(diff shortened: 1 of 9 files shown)


liyinan926:
LGTM.

sahilTakiar:
* Added `synchronized` to `TaskState.updateRecordMetrics` and to `TaskState.updateByteMetrics`.
** The reason is that each method performs get and set semantics (e.g. get the current value of the counter and use it to set the new value).
** These methods are executed in `AbstractTaskStateTracker.TaskMetricsUpdater` and in `Fork.commitData`
* Added call to `updateRecordMetrics` in `Fork.commitData` the reason is that the record-level and byte-level metrics should be updated one last time after `DataWriter.commit` is invoked to ensure that the metrics are up to date, since the `TaskMetricsUpdater` is only scheduled to run at a specific interval.
* Removed the restriction the `DataWriter.bytesWritten` can only be called after `DataWriter.commit` has been invoked.
** @liyinan926 do you know why this restriction was added in the first place? I don't see why it needs to be there, and it is useful to get this data from the `DataWriter` on a regular interval so it can be reported while the `Task` is running.
** In addition, this restriction has never been honored, as it is already being invoked on a scheduled basis in `TaskMetricsUpdater`.

liyinan926:
The reason for the restriction is file length won't be available until the file is either flushed or closed.

liyinan926:
LGTM.

sahilTakiar:
@ibuenros addressed the comments from our discussion, please review.

gobblin-core/src/main/java/gobblin/instrumented/writer/InstrumentedDataWriterBase.java
```diff
@@ -170,10 +198,84 @@ public State getFinalState() {
   public void close()
       throws IOException {
     this.closer.close();
+    if (this.writerMetricsUpdaterFuture.isPresent()) {
+      this.writerMetricsUpdaterFuture.get().cancel(false);
```
ibuenros: Will this also shutdown the thread? We had issues were non-shutdown threads would block Gobblin from exiting.

ibuenros:
One comment, LGTM otherwise.

sahilTakiar:
I think calling `ScheduledFuture.cancel` should be sufficient as I tested it locally and there were no hung threads, but just to be sure I changed the code to shutdown the Thread Pool directly using the `ExecutorUtils.shutdownExecutorService`.

PR #502, opened 2015-12-03
Adding instrumentation to publish to Maven Central

chavdar:
Rework of the Maven Publishing code:

* Switch to using the legacy "maven" plugin instead of the new "maven-publish" due to issues with signing artifacts
* Add support for signing the artifacts
* Use group com.linkedin.gobblin instead of gobblin as per Sonatype requirements
* Add POM metadata as per Sonatype requirements
* Rename gobblin-utility.tar.gz and gobblin-compaction.tar.gz to gobblin-utility.tgz and gobblin-compaction.tgz since the Gradle signing plugin does not work correctly if a file has two extensions
* Add a helper script maven-sonatype.sh with instructions on how to publish the artifacts

@liyinan926 @ibuenros Can you review?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ apply from: environmentScript
 ext.publishToMaven = project.hasProperty('publishToMaven')
 if (ext.publishToMaven) {
     plugins.apply('maven')
-    //workaround for a bug in gradle's "maven" plugin. See https://discuss.gradle.org/t/error-in-parallel-build/7215/3
+    // Workaround for a bug in gradle's "maven" plugin. See https://discuss.gradle.org/t/error-in-parallel-build/7215/3
     project.setProperty("org.gradle.parallel", "false")
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -23,8 +23,15 @@ File getEnvironmentScript()
 
 apply from: environmentScript
 
+ext.publishToMaven = project.hasProperty('publishToMaven')
+if (ext.publishToMaven) {
+    plugins.apply('maven')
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -23,8 +23,15 @@ File getEnvironmentScript()
 
 apply from: environmentScript
 
+ext.publishToMaven = project.hasProperty('publishToMaven')
+if (ext.publishToMaven) {
+    plugins.apply('maven')
+    //workaround for a bug in gradle's "maven" plugin. See https://discuss.gradle.org/t/error-in-parallel-build/7215/3
```
liyinan926: Better to use a consistent style for comment: one empty space after `//` and first letter capital case.

build.gradle
```diff
@@ -41,6 +48,9 @@ if (!project.hasProperty('version') || project.version == 'unspecified') {
     else {
         project.version = tagStr
     }
+    if (!project.hasProperty('useHadoop2')) {
+      project.version = project.version + "-hadoop1"
```
liyinan926: Should we also include the target Hadoop version specified by `hadoopVersion`?

build.gradle
chavdar: Will fix.

build.gradle
chavdar: Is that necessary? We need "-hadoop1" to differentiate between Hadoop-1 and Hadoop-2 artifacts. Beyond that the POM dependencies should take care of specifying the version of Hadoop1 used.

liyinan926: Makes sense.

chavdar:
PR updated.

liyinan926:
LGTM.

ibuenros:
LGTM

PR #504, opened 2015-12-04
Fixing findbugs warnings from PR 500

sahilTakiar:


sahilTakiar:
@ibuenros can you take a look?

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -405,7 +405,9 @@ public boolean setIdentity(String privateKey, JSch jsch) {
         log.warn("Failed to set identity using HDFS file. Will attempt next strategy. " + e.getMessage());
       } finally {
         try {
-          privateKeyStream.close();
+          if (privateKeyStream != null) {
```
liyinan926: Can we use the Java 7's try (resource) clause?

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -50,8 +51,8 @@
 @Slf4j
 public class FileAwareInputStreamDataWriter implements DataWriter<FileAwareInputStream> {
 
-  protected volatile long bytesWritten = 0;
-  protected volatile long filesWritten = 0;
+  protected AtomicLong bytesWritten = new AtomicLong();
```
liyinan926: Both can be final.

liyinan926:
LGTM.

ibuenros:
@sahilTakiar what issues are addressed here? I saw use of reentrant locks where normal synchronization should work, am I missing something?

sahilTakiar:
The FindBugs report complain that this is a partial synchronization of some of the `Meter`s. Only `recordsWrittenMeter` and `bytesWrittenMeterLock` need synchronized access as they may be modified in the `commit()`, `regenerateMetrics()` or by the `WriterMetricsUpdater`.

The original code had synchronization in the `commit()` method in the `WriterMetricsUpdater`, but not in the `regenerateMetrics()` method, which also modifies the record and byte level meters.

FindBugs complained about this so I worked to add synchronization to `regenerateMetrics()`, but I hit a few issues:
* You cannot simply add the `synchronization` keyword to the `regenerateMetrics()` method because then you have partial synchronization of all the other meters, which FindBugs will once again complain about
* You cannot do the following either:

```
synchronized (this.recordsWrittenMeter) {
    this.recordsWrittenMeter = Optional.absent()
}
````

    * FindBugs will once again complain about the above pattern.
    * http://findbugs.sourceforge.net/bugDescriptions.html#ML_SYNC_ON_FIELD_TO_GUARD_CHANGING_THAT_FIELD and http://stackoverflow.com/a/8787344 have explanations of why this is the case.

chavdar:
@sahilTakiar Shouldn't it be:

synchronized (this) {
    this.recordsWrittenMeter = Optional.absent()
}

sahilTakiar:
@chavdar you are right. Updated.

chavdar:
LGTM

ibuenros:
LGTM

PR #505, opened 2015-12-04
Separate publisher filesystem from writer filesystem

kadaan:
Support for publishing data to a different filesystem from where data is written.  The use case of this is writing the files to HDFS, but publishing them to S3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ out/
 **/mainGeneratedDataTemplate
 **/mainGeneratedRest
 gobblin-dist*
+gobblin.tar.gz
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -60,7 +60,8 @@ public class BaseDataPublisher extends SingleTaskDataPublisher {
   private static final Logger LOG = LoggerFactory.getLogger(BaseDataPublisher.class);
 
   protected final int numBranches;
-  protected final List<FileSystem> fileSystemByBranches;
+  protected final List<FileSystem> writerFileSystemByBranches;
+  protected final List<FileSystem> publisherFileSystemByBranches;
   protected final List<Optional<String>> publisherFinalDirOwnerGroupsByBranches;
   protected final List<FsPermission> permissions;
   protected final Closer closer;
@@ -79,16 +80,22 @@ public class BaseDataPublisher extends SingleTaskDataPublisher {
 
     this.numBranches = this.getState().getPropAsInt(ConfigurationKeys.FORK_BRANCHES_KEY, 1);
 
-    this.fileSystemByBranches = Lists.newArrayListWithCapacity(this.numBranches);
+    this.writerFileSystemByBranches = Lists.newArrayListWithCapacity(this.numBranches);
+    this.publisherFileSystemByBranches = Lists.newArrayListWithCapacity(this.numBranches);
     this.publisherFinalDirOwnerGroupsByBranches = Lists.newArrayListWithCapacity(this.numBranches);
     this.permissions = Lists.newArrayListWithCapacity(this.numBranches);
 
     // Get a FileSystem instance for each branch
     for (int i = 0; i < this.numBranches; i++) {
-      URI uri = URI.create(this.getState().getProp(
+      URI writerUri = URI.create(this.getState().getProp(
           ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, this.numBranches, i),
           ConfigurationKeys.LOCAL_FS_URI));
-      this.fileSystemByBranches.add(FileSystem.get(uri, conf));
+      this.writerFileSystemByBranches.add(FileSystem.get(writerUri, conf));
+
+      URI publisherUri = URI.create(this.getState().getProp(
+          ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISHER_FILE_SYSTEM_URI, this.numBranches, i),
+          writerUri.toString()));
+      this.publisherFileSystemByBranches.add(FileSystem.get(publisherUri, conf));
 
       // The group(s) will be applied to the final publisher output directory(ies)
       this.publisherFinalDirOwnerGroupsByBranches.add(Optional.fromNullable(this.getState().getProp(ForkOperatorUtils
@@ -162,12 +169,12 @@ public class BaseDataPublisher extends SingleTaskDataPublisher {
   protected void publishData(WorkUnitState state, int branchId, boolean publishSingleTaskData,
       Set<Path> writerOutputPathsMoved) throws IOException {
     // Get a ParallelRunner instance for moving files in parallel
-    ParallelRunner parallelRunner = this.getParallelRunner(this.fileSystemByBranches.get(branchId));
+    ParallelRunner parallelRunner = this.getParallelRunner(this.writerFileSystemByBranches.get(branchId));
 
     // The directory where the workUnitState wrote its output data.
     Path writerOutputDir = WriterUtils.getWriterOutputDir(state, this.numBranches, branchId);
 
-    if (!this.fileSystemByBranches.get(branchId).exists(writerOutputDir)) {
+    if (!this.writerFileSystemByBranches.get(branchId).exists(writerOutputDir)) {
       LOG.warn(String.format("Branch %d of WorkUnit %s produced no data", branchId, state.getId()));
       return;
     }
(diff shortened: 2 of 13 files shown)


kadaan:
Weird that this builds fine on my machine, but fails within travis.  I see the issue and am adding a new variable 'hadoopAwsVersion' to allow overriding (but it defaults to 2.6.0).

kadaan:
Seems like this is an intermittent failure.

sahilTakiar:
Seem the build failed due to PR https://github.com/linkedin/gobblin/pull/506 @liyinan926 do we know the root cause?

kadaan:
The weird thing is that have have seen the java7 and java8 builds each succeed in Travis, but not reliably.

liyinan926:
The build failed on Hudson because one unit test assumes some ordering that seems not guaranteed on Hudson. I noticed that build on Java 7 has been pretty stable and not so on Java 8.

build.gradle
```diff
@@ -108,7 +112,7 @@ ext.externalDependency = [
   "hadoopYarnClient": "org.apache.hadoop:hadoop-yarn-client:"+hadoopVersion,
   "hadoopYarnMiniCluster": "org.apache.hadoop:hadoop-minicluster:"+hadoopVersion,
   "hadoopAnnotations": "org.apache.hadoop:hadoop-annotations:"+hadoopVersion,
-  "hadoopAws": "org.apache.hadoop:hadoop-aws:2.6.0",
+  "hadoopAws": "org.apache.hadoop:hadoop-aws:"+hadoopAwsVersion,
```
liyinan926: `hadoopAws` is not available prior to 2.6.0.

.gitignore
```diff
@@ -50,3 +50,5 @@ out/
 **/mainGeneratedDataTemplate
 **/mainGeneratedRest
 gobblin-dist*
+gobblin.tar.gz
+masterPwd
```
liyinan926: You don't need this since a test has been fixed to delete `masterPwd` after it is done.

gobblin-core/build.gradle
```diff
@@ -51,7 +51,7 @@ dependencies {
 
   if (project.hasProperty('useHadoop2')) {
     compile externalDependency.avroMapredH2
-    testRuntime externalDependency.hadoopAws
+    compile externalDependency.hadoopAws
```
liyinan926: I don't think we need compile dependency on `hadoopAws`.

build.gradle
kadaan: I know it isn't but there is the possibility to use other versions. For instance, we are running in Qubole and they have their own version.

gobblin-core/build.gradle
kadaan: I changed it to a compile dependency because it is needed to use the s3a filesystem.  Support for the s3a filesystem was the point of this PR.

.gitignore
kadaan: I'll remove the masterPwd.  It should be fine to leave gobblin.tar.gz, right?

gobblin-core/build.gradle
liyinan926: I don't see any explicit reference to the `S3FileSystem` in any of the changes.

build.gradle
liyinan926: This will break build as the default version is 2.3.0.

gobblin-core/build.gradle
kadaan: There isn't an explicit reference, but as soon as you use a uri like `s3a://...` for one of the fs.uri properties (in my case `data.publisher.fs.uri`) there will be a runtime error because the hadoop-aws library is missing.

liyinan926: Yes, that's why this is a `runtime` dependency.`

build.gradle
kadaan: It shouldn't because it defaults:
`
if (!project.hasProperty('hadoopAwsVersion')) {
  hadoopAwsVersion = '2.6.0'
}
`

gobblin-core/build.gradle
kadaan: Good point

kadaan:
@liyinan926 I believe I have address all of your issues.  Anything you still feel is outstanding?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -79,16 +80,23 @@ public BaseDataPublisher(State state) throws IOException {
...
     // Get a FileSystem instance for each branch
     for (int i = 0; i < this.numBranches; i++) {
-      URI uri = URI.create(this.getState().getProp(
-          ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, this.numBranches, i),
-          ConfigurationKeys.LOCAL_FS_URI));
-      this.fileSystemByBranches.add(FileSystem.get(uri, conf));
+      String writerUriStr = this.getState().getProp(
+              ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_FILE_SYSTEM_URI, this.numBranches, i),
+              ConfigurationKeys.LOCAL_FS_URI);
+      URI writerUri = URI.create(writerUriStr);
```
liyinan926: You can combine this and the next line.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -79,16 +80,23 @@ public BaseDataPublisher(State state) throws IOException {
...
+      this.writerFileSystemByBranches.add(FileSystem.get(writerUri, conf));
+
+      URI publisherUri = URI.create(this.getState().getProp(
```
liyinan926: It's more readable to create a instance for the URI string like `writerUriStr`.

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -236,6 +236,34 @@ public Void call() throws Exception {
     }));
   }
 
+  public void movePath(final Path src, final FileSystem dstFs, final Path dst, final Optional<String> group) {
```
liyinan926: Javadoc missing. Can you add a unit test for this?

kadaan:
@liyinan926 I believe I have addressed all of the outstanding issues.  Anything else you would like improved?

.gitignore
```diff
@@ -50,3 +50,4 @@ out/
 **/mainGeneratedDataTemplate
 **/mainGeneratedRest
 gobblin-dist*
+gobblin.tar.gz
```
liyinan926: You don't need this since it is covered by `gobblin-dist*`.

liyinan926:
Seems the build has been failing constantly on both Java 7 and 8. Can you check and make sure the build is fine?

.gitignore
kadaan: Ahhhh.  Maybe the name of it changed.

liyinan926: I think `gobblin-compaction` has a task to generate `gobblin.tar.gz`. OK, so ignore my comment on this.

kadaan: Okay

kadaan: Would you be opposed to adding these to the ignore list as well:
 * ParallelRunnerTest
 * GobblinHelixTaskTest
 * FsDatasetStateStoreTest/TestJob

kadaan:
@liyinan926 The build is failing due to gobblin-yarn tests.  These tests work when I run them locally.

.gitignore
liyinan926: No, under normal circumstances, those dirs should be removed.

liyinan926:
@kadaan the exception log you posted is expected, which is intentionally introduced by the `FaultyExtractor`. I suggest you pull the latest master and merge in the changes.

.gitignore
kadaan: Fair enough

kadaan:
@liyinan926 As far as I can tell, all the tests work when I run it locally.

kadaan:
@liyinan926 This is super annoying.  The test I created intentionally mocked the filesystem and file operations.  But in the logs I see this:
```
Gradle test > gobblin.util.ParallelRunnerTest.testMovePath FAILED
    java.io.IOException: java.util.concurrent.ExecutionException: java.io.IOException: /src/file.txt: No such file or directory
        at gobblin.util.ParallelRunner.close(ParallelRunner.java:286)
        at com.google.common.io.Closer.close(Closer.java:214)
        at gobblin.util.ParallelRunnerTest.testMovePath(ParallelRunnerTest.java:195)

        Caused by:
        java.util.concurrent.ExecutionException: java.io.IOException: /src/file.txt: No such file or directory
            at java.util.concurrent.FutureTask.report(FutureTask.java:122)
            at java.util.concurrent.FutureTask.get(FutureTask.java:192)
            at gobblin.util.ParallelRunner.close(ParallelRunner.java:281)
            ... 2 more

            Caused by:
            java.io.IOException: /src/file.txt: No such file or directory
```
Unfortunately, the stack trace doesn't really help to locate the problem.  All calls on the filesystem class are mocked.

kadaan:
To try to narrow this down I put the new test in it's own project and run the build with travis.  Unfortunately, no repro.

https://github.com/kadaan/hadoop_filesystem_test

kadaan:
I think I have a repro.  If you run the yarn tests when the useHadoop2 property is set, but the version is left at the default (2.3.0), then GobblinApplicationMasterTest.setUp,GobblinWorkUnitRunnerTest.setUp, and GobblinYarnAppLauncherTest.setUp fail due to:

kadaan:
@liyinan926 Fixed.  This should be good to merge now.

liyinan926:
LGTM. @sahilTakiar do you want to take another look?

sahilTakiar:
LGTM

liyinan926:
Thanks for contributing!

kadaan:
@liyinan926 @sahilTakiar Thanks for the comprehensive review!!!

PR #501, opened 2015-12-03
Unit tests for Gobblin Yarn part I

liyinan926:
This is part I of unit tests for Gobblin on Yarn. There are still a few classes that has dependency on Yarn that I'm trying to figure out the best way to test. Will create a PR for part II for those classes later.

PR #507, opened 2015-12-05
Unit tests for Gobblin Yarn part II

liyinan926:
This is the part II of unit tests for Gobblin Yarn. Part II additionally covers `YarnAppSecurityManager` and `YarnContainerSecurityManager`. Currently the only Yarn-dependent class that has no coverage is `YarnService`. Since `YarnService` runs in the `GobblinApplicationMaster` that is supposed to be launched by Yarn without user interference, I'm still trying to figure out a way of running it manually. This PR also addresses the left-over review comments from PR #501.  

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
@@ -246,7 +246,7 @@ public class ParallelRunner implements Closeable {
     } catch (InterruptedException ie) {
       throw new IOException(ie);
     } catch (ExecutionException ee) {
-      throw new IOException(ee);
+      throw new IOException(ee.getCause());
     } finally {
       ExecutorsUtils.shutdownExecutorService(this.executor, Optional.of(LOGGER));
     }
--- a/gobblin-yarn/src/test/java/gobblin/yarn/YarnSecurityManagerTest.java
+++ b/gobblin-yarn/src/test/java/gobblin/yarn/YarnSecurityManagerTest.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.yarn;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Collection;
+
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.RetryOneTime;
+import org.apache.curator.test.TestingServer;
+import org.apache.hadoop.conf.Configuration;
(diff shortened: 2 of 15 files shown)


liyinan926:
@sahilTakiar can you review?

gobblin-yarn/src/test/java/gobblin/yarn/YarnSecurityManagerTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    this.helixManager = HelixManagerFactory.getZKHelixManager(
+        helixClusterName, TestHelper.TEST_HELIX_INSTANCE_NAME, InstanceType.SPECTATOR, zkConnectingString);
+    this.helixManager.connect();
+
+    this.miniHDFSBaseDir = new Path(YarnSecurityManagerTest.class.getSimpleName());
+    this.dfsConfig = getTestConfiguration(this.miniHDFSBaseDir);
+    dfsConfig.set(MiniDFSCluster.HDFS_MINIDFS_BASEDIR, this.miniHDFSBaseDir.toUri().toString());
+    MiniDFSCluster.Builder builder = new MiniDFSCluster.Builder(dfsConfig);
+    this.miniDFSCluster = builder.numDataNodes(1).build();
```
sahilTakiar: You can combine the above two lines

sahilTakiar:
Do we have any documentation on the testing strategy?

liyinan926:
What you mean by testing strategy?

sahilTakiar:
More like what the tests do, how we are using the curator framework, how we are using the MiniDfsCluster, etc. 

* My understanding is that we are using the Curator framework to setup a local Zookeeper instance that Helix connects to, is that correct?
* Why is it necessary to use the Curator framework vs. just mocking HelixManager
* What is MiniDfsCluster needed for? Why is it necessary to use a MiniDfsCluster over a LocalFileSystem.
* Are these all unit tests? Is it possible to have an end-to-end integration tests.

liyinan926:
1. The Curator framework is a ZooKeeper client that is easier to use than the official ZooKeeper client library. 
2. The `MiniDFSCluster` was meant to provide a way to get a delegation token. I just realized it's not really necessary since it's essentially the same as `LocalFileSystem` for this, both return `null`, and that's why I need to use mocking. 
3. Yes, these are unit tests. 
4. Yes, I agree we need end-to-end integration tests.

liyinan926:
@sahilTakiar updated 29f4e46.

liyinan926:
Hey, @sahilTakiar, can you take another look d11cd1f?

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -246,7 +246,7 @@ public void close() throws IOException {
     } catch (InterruptedException ie) {
       throw new IOException(ie);
     } catch (ExecutionException ee) {
-      throw new IOException(ee);
+      throw new IOException(ee.getCause());
```
sahilTakiar: Why make this change? I believe this won't preserve the stack trace of the original exception.

liyinan926: `ExecutionException` is just a wrapper of the actual exception returned by `getCause`.

sahilTakiar:
LGTM

Only comment is why do we need the Curator Framework when we could just mock the `HelixManager` using Mockito? Would making a mock `HelixManager` be hard? The advantage of just using a mock is that we don't have to do things like `Thread.sleep(5000)` in the code to wait for the Helix + ZK interaction to complete.

liyinan926:
A `HelixManager` does not allow you to check arbitrary ZooKeeper path. In those tests, I really want to make sure the appropriate ZooKeeper paths get created in real. A mocked `HelixManager` simply won't do the job.

PR #509, opened 2015-12-08
Fix for StreamUtils.tar and Yarn LogCopier

sahilTakiar:
* Fix for `StreamUtils.tar` there were some problems with un-tarring the resulting object on the GNU tar distribution vs. the BSD tar distribution
* Fix for `LogCopier.Stopwatch`

@liyinan926 can you review?


liyinan926:
LGTM.

PR #513, opened 2015-12-09
Implemented performance tests for gobblin metrics.

ibuenros:


ibuenros:
@sahilTakiar can you review? Thanks!

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsPerformanceTest.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import org.testng.annotations.Test;
+
+import com.beust.jcommander.internal.Lists;
```
sahilTakiar: I think you meant to import the Guava `Lists` object, right?

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsPerformanceTest.java
```diff
@@ -0,0 +1,233 @@
...
+import com.google.common.collect.Sets;
+
+
+/**
+ * Class for running Gobblin metrics performance tests.
+ */
+@Slf4j
+public class MetricsPerformanceTest {
+
+  @Test(groups = { "performance" })
```
sahilTakiar: Can't you add the `groups = { "performance" }` on a class level so that you don't have to do it for each method?

sahilTakiar:
Should we create a wiki page on Github showing the results?

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsPerformanceTest.java
```diff
@@ -0,0 +1,233 @@
...
+   * @param timers Number of timers to generate per thread.
+   * @param queries Number of increments to do, divided among all threads.
+   * @return total QPS achieved (e.g. total increments per second in the {@link Incrementer}s)
+   * @throws Exception
+   */
+  @Builder(buildMethodName = "run", builderMethodName = "singleTest")
+  public static double _singleTest(int threads, int depth, int forkAtDepth, int counters, int meters, int histograms,
+      int timers, long queries) throws Exception {
+
+    System.gc();
```
sahilTakiar: Why is this needed?

gobblin-metrics/build.gradle
```diff
@@ -85,7 +85,7 @@ artifacts {
 
 test {
     useTestNG () {
-      excludeGroups 'ignore'
+      excludeGroups 'ignore', 'performance'
```
sahilTakiar: Why put this code into TestNg tests when we are going to exclude them from the build anyway? I think it would only make sense to make these tests if we make them fail if the performance numbers exceed a certain value, although even though approach isn't always reliable especially on shared build systems such as Travis.

It may make more sense to move these into a separate package called `gobblin.metrics.performance.eval` (or something like that), and have a shell script to build and generate a report on demand.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsUpdater.java
```diff
@@ -0,0 +1,124 @@
...
+      int timers, MetricContext baseContext) {
+    this.depth = depth;
+    this.random = new Random();
+
+    MetricContext tmpContext = baseContext == null ? RootMetricContext.get() : baseContext;
+    while(depth > 0) {
+      tmpContext = tmpContext.childBuilder(UUID.randomUUID().toString()).build();
+      depth--;
+    }
+    this.context = tmpContext;
```
sahilTakiar: Do we have a scenario where we have a single parent context, with multiple child contexts, and updates to the child context happen concurrently?

That setup may be more reflective of the actual setup in Gobblin right? e.g. in a container I have a single MetricContext for a Dataset, and multiple children MetricContext for each WorkUnit.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsPerformanceTest.java
ibuenros: I put it there to clean `MetricContext`s from other runs to ideally reduce overhead. It shouldn't affect anything negatively.

gobblin-metrics/build.gradle
ibuenros: Do you mean a different project? I wanted to avoid having a different project just for performance of metrics. Also, the tests are fairly slow, so I didn't want them to run with unit tests.
I can add a method to gradle (say `gradle performance`) to run the performance tests, would this work?

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsUpdater.java
ibuenros: `MetricsPerformanceTest.parallelizationTest` does this.

ibuenros:
I am also writing a wiki page on the results.

ibuenros:
@sahilTakiar addressed comments
introduced a gradle task `./gradlew performance` that will run performance tests.

gobblin-metrics/build.gradle
sahilTakiar: No we can leave it in the same project, but just in a different package. My main concern is this the code doesn't contain any actual tests, you don't have any Assertions or things like that; so it doesn't really needed to tagged as TestNg tests (e.g. with the `@Test` annotation). You could just make this regular code under `src/main/java/`, and have a shell script that invokes the main method of some class. However, it doesn't really matter, the current setup works too.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsUpdater.java
```diff
@@ -0,0 +1,124 @@
...
+
+import lombok.Builder;
+import lombok.Getter;
+
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import com.beust.jcommander.internal.Lists;
```
sahilTakiar: Same here, I think you meant Guava.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
```diff
@@ -0,0 +1,204 @@
...
+    TreeMap<List<Integer>, Double> results = Maps.newTreeMap(comparator);
+
+    for(List<Integer> p : parameters) {
+      results.put(p, _singleTest(p.get(0), p.get(1), p.get(2), p.get(3), p.get(4), p.get(5), p.get(6), queries));
+    }
+
+    System.out.println("===========================");
+    System.out.println(name);
+    System.out.println("===========================");
+    System.out.println("Threads\tDepth\tForkAtDepth\tCounters\tMeters\tHistograms\tTimers\tQPS");
```
sahilTakiar: Why not just use the Slf4j logger since the class has the `@Slf4j` annotation anyway.

ibuenros: Because print statements are automatically captured by TestNG and put into the testing report, while log statements aren't.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
```diff
@@ -0,0 +1,204 @@
...
+      queries = 50000000l;
+    }
+    if(Strings.isNullOrEmpty(name)) {
+      name = "Test";
+    }
+
+    Set<List<Integer>> parameters = Sets.cartesianProduct(threads, depth, forkAtDepth, counters, meters,
+        histograms, timers);
+
+    Comparator<List<Integer>> comparator = new Comparator<List<Integer>>() {
```
sahilTakiar: Can you place this in its own class with a more descriptive name.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
```diff
@@ -0,0 +1,204 @@
...
+        } else {
+          return 0;
+        }
+      }
+    };
+
+    TreeMap<List<Integer>, Double> results = Maps.newTreeMap(comparator);
+
+    for(List<Integer> p : parameters) {
+      results.put(p, _singleTest(p.get(0), p.get(1), p.get(2), p.get(3), p.get(4), p.get(5), p.get(6), queries));
```
sahilTakiar: Add Precondition to check that each `List` is of size 7.

sahilTakiar: Shouldn't we use the singleTest builder instead of the constructor.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
sahilTakiar: Also whats the reason for having this `Comparator` and the `TreeMap` as far as I can tell it only affects the order which the results are printed out.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsUpdater.java
```diff
@@ -0,0 +1,124 @@
...
+ *   An instance is created through a builder with the following optional parameters:
+ *   * depth: number of levels in generated {@link MetricContext} tree. Default: 0
+ *   * context: base {@link MetricContext} used as the root of the generated tree. Default: {@link RootMetricContext}.
+ *   * counters: number of counters to generate. Named counter0, counter1, ...
+ *   * meters: number of meters to generate. Named meter0, meter1, ...
+ *   * histograms: number of histograms to generate. Named histogram0, histogram1, ...
+ *   * timers: number of timers to generate. Named timer0, timer1, ...
+ * </p>
+ */
+public class MetricsUpdater implements Runnable {
```
sahilTakiar: Would recommend using `Callable` instead of `Runnable` as `Callable` has better handing of exceptions thrown in the `run()` method.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    List<Future<Long>> incrementerFutures = Lists.newArrayList();
+    for(Incrementer incrementer : incrementerList) {
+      incrementerFutures.add(executorService.submit(incrementer));
+    }
+
+    long totalTime = 0;
+    for(Future<Long> future : incrementerFutures) {
+      log.info("Waiting for future.");
```
sahilTakiar: Suggest removing this or making it more descriptive so it logs which some more metadata about the Future it is waiting on.

gobblin-metrics/src/test/java/gobblin/metrics/performance/PerformanceUtils.java
ibuenros: That is the only use, to make sure that when printing, the tests are displayed in a sensible order.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsUpdater.java
ibuenros: The idea is the `Incrementer` could take other `Runnables`, we need an interface for something that can be run, which is precisely `Runnable`.

ibuenros:
@sahilTakiar addressed or responded to all comments.

gobblin-metrics/src/test/java/gobblin/metrics/performance/MetricsPerformanceTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.testng.annotations.Test;
+
+import com.google.common.collect.Sets;
+
+
+/**
+ * Class for running Gobblin metrics performance tests.
+ */
+@Slf4j
```
sahilTakiar: Not used

sahilTakiar:
LGTM

PR #514, opened 2015-12-09
Add recompact information to compactionRecordCountEvent

ydai1124:
CompactionRecordCountEvent now reports whether re-compaction is needed/completed. 

@zliu41 Can you review?


zliu41:
LGTMr

PR #516, opened 2015-12-10
Improved Yarn container fault tolerance handling

liyinan926:
1. Added an option to use the same host as the preferred host when requesting a replacement container.
2. The Helix instance names of containers are now controlled and assigned by the ApplicationMaster, which maintains a fixed set of Helix instance names that get reused for replacement containers.
3. Added support to stop attempts on requesting a replacement container for a Helix instance if the retry count of the instance has reached the maximum.
4. Note that we don't measure how soon a Helix instance fails and stop retries if it's too soon, as Samza does. The reason is container failures in our case will mostly not be due to job/task issues and may be very likely transient. Even if a container fails consecutively really fast due to persistent issues, e.g., node failures, the system will stop retries after the maximum of retries gets reached anyway.


liyinan926:
@sahilTakiar can you review?

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -340,6 +369,60 @@ private String buildContainerCommand(Container container) {
...
 
   /**
+   * Check the exit status of a completed container and see if the replacement container
+   * should try to be started on the same node. Some exit status indicates a disk or
+   * node failure and in such cases the replacement container should try to be started on
+   * a different node.
+   */
+  private boolean shouldStickToTheSameNode(int containerExitStatus) {
+    switch (containerExitStatus) {
+      case ContainerExitStatus.DISKS_FAILED:
```
sahilTakiar: return false?

liyinan926: It will fall through.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -425,12 +506,9 @@ public void onContainerStarted(ContainerId containerId, Map<String, ByteBuffer>
...
     public void onContainerStatusReceived(ContainerId containerId, ContainerStatus containerStatus) {
       LOGGER.info(String.format("Received container status for container %s: %s", containerId, containerStatus));
-      if (containerStatus.getState() == ContainerState.COMPLETE) {
-        LOGGER.info(String.format("Container %s completed with status %s", containerId, containerStatus));
-        containerMap.remove(containerId);
 
-        LOGGER.info("Requesting a new container to replace the one that has completed");
-        eventBus.post(new NewContainerRequest(1));
+      if (containerStatus.getState() == ContainerState.COMPLETE) {
+        handleContainerCompletion(containerStatus);
```
sahilTakiar: Why do we invoke `handleContainerCompletion` here and in `onContainersCompleted` seems they would be duplicate invocations.

liyinan926: Actually I'm still not sure which one or both callback methods will be called in case a container completes. We can keep the one in `onContainersCompleted`.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -340,6 +369,60 @@ private String buildContainerCommand(Container container) {
...
+        return this.containerHostAffinityEnabled;
+    }
+  }
+
+  /**
+   * Handle the completion of a container. A new container will be requested to replace the one
+   * that just exited. Depending on the exit status and if container host affinity is enabled,
+   * the new container may or may not try to be started on the same node.
+   */
+  private void handleContainerCompletion(ContainerStatus containerStatus) {
```
sahilTakiar: My understanding is that since we have long running containers, the only time a container should complete is if there was either (1) on error in the container that caused it to crash, or (2) we are trying to shutdown the instance. If my understanding is correct, can you add that to the javadocs of this method. It's not terribly clear to somewhat not super familiar with the design.

Does `AMRMClientCallbackHandler.onContainersCompleted` get invoked when we try to shutdown the whole instance? Do we need any additional handling to ensure we have a clean shutdown.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -121,53 +142,52 @@ public YarnService(Config config, String applicationName, String applicationId,
...
+          this.requestedContainerCores));
       return;
     }
 
-    try {
-      requestContainers(newContainersRequested);
-    } catch (IOException | YarnException e) {
-      LOGGER.error("Failed to handle new container request", e);
-    }
+    requestContainer(newContainerRequest.getReplacedContainer().transform(new Function<Container, String>() {
```
sahilTakiar: Doesn't `getReplacedContainer` return `Optional<Container>`?

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
sahilTakiar: Don't you want to always use a different node if the disk failed? Regardless of the value of `containerHostAffinityEnabled`

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
liyinan926: `requestContainer` takes `Optional<String>`.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
liyinan926: Yes, it falls through to `ContainerExitStatus.ABORTED` in which case it returns `false`.

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
```diff
@@ -367,16 +441,23 @@ public void onContainersAllocated(List<Container> containers) {
...
-        containerMap.put(container.getId(), containerParticipantPair);
+        String instanceName = unusedHelixInstanceNames.poll();
+        if (Strings.isNullOrEmpty(instanceName)) {
+          // No unused instance name, so generating a new one.
+          instanceName = YarnHelixUtils.getHelixInstanceName(GobblinWorkUnitRunner.class.getSimpleName(),
+              helixInstanceIdGenerator.incrementAndGet());
+        }
+
+        final String finalInstanceName = instanceName;
+        containerMap.put(container.getId(), new AbstractMap.SimpleImmutableEntry<>(container, finalInstanceName));
```
sahilTakiar: Can you prefix `containerMap` with `this`. This way when looking at a PR or RB it is clear this is an instance variable, not a local variable.

liyinan926: No, this is not possible since it's in an inner class.

liyinan926:
@sahilTakiar updated e18cde6.

liyinan926:
@sahilTakiar can you take another look e18cde6?

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
sahilTakiar: This will trigger a FindBugs warning: http://findbugs.sourceforge.net/bugDescriptions.html#SF_SWITCH_FALLTHROUGH

sahilTakiar:
One minor comment. Otherwise LGTM

gobblin-yarn/src/main/java/gobblin/yarn/YarnService.java
liyinan926: OK, changed to return `false`.

PR #517, opened 2015-12-11
Create structure for gobblin-config-management

chavdar:
@tuGithub @ibuenros Can you review? This is prep work for tomorrow.


chavdar:
Merged gobblin-config-store-api and gobblin-config-common-impl into gobblin-config-core as discussed offline.

PR #518, opened 2015-12-11
Adding Container Metrics for gobblin-yarn

sahilTakiar:
* YARN containers were not reporting any Task Level Metrics (e.g. records extracted, records written, bytes written, etc.) because Metrics Reporting was not enabled in the containers
* Added `ContainerMetrics.java` to handle metrics reporting for each YARN container
* The metrics hierarchy is as follows `ContainerMetrics` -> `JobMetrics` -> `TaskMetrics`
* A `JobMetrics` instance is obtained in each `GobblinHelixTask`
* Unlike `MRJobLauncher` where the `JobMetrics` instance is manually removed from the `GobblinMetricsRegistry`, `GobblinHelixTask` does not remove the `JobMetrics` instance at the end of each task because there may be more tasks in the container that need the `JobMetrics` instance
* Reporting is done on the `ContainerMetrics` instance
* Testing is still in progress

@liyinan926 and @ibuenros can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
@@ -14,11 +14,10 @@ package gobblin.runtime.util;
 
 import java.util.List;
 
-import org.apache.hadoop.conf.Configuration;
-
 import com.google.common.collect.Lists;
 
 import gobblin.metrics.GobblinMetrics;
+import gobblin.metrics.MetricContext;
 import gobblin.metrics.Tag;
 import gobblin.runtime.JobState;
 import gobblin.runtime.TaskState;
@@ -35,9 +34,12 @@ public class JobMetrics extends GobblinMetrics {
   protected final String jobName;
 
   protected JobMetrics(JobState job) {
-    super(name(job), null, tagsForJob(job));
-    this.jobName = job.getJobName();
+    this(job, null);
+  }
 
+  protected JobMetrics(JobState job, MetricContext parentContext) {
+    super(name(job), parentContext, tagsForJob(job));
+    this.jobName = job.getJobName();
   }
 
   /**
@@ -59,6 +61,17 @@ public class JobMetrics extends GobblinMetrics {
    */
   public static JobMetrics get(String jobId) {
     return get(null, jobId);
+  }
+
+  /**
+   * Get a new {@link GobblinMetrics} instance for a given job.
+   *
+   * @param jobState the given {@link JobState} instance
+   * @param parentContext is the parent {@link MetricContext}
+   * @return a {@link JobMetrics} instance
+   */
+  public static JobMetrics get(JobState jobState, MetricContext parentContext) {
+    return (JobMetrics) GOBBLIN_METRICS_REGISTRY.getOrDefault(name(jobState), new JobMetrics(jobState, parentContext));
   }
 
   /**
@@ -93,19 +106,17 @@ public class JobMetrics extends GobblinMetrics {
 
   private static List<Tag<?>> tagsForJob(JobState jobState) {
     List<Tag<?>> tags = Lists.newArrayList();
-    tags.add(new Tag<String>("jobName", jobState.getJobName() == null ? "" : jobState.getJobName()));
-    tags.add(new Tag<String>("jobId", jobState.getJobId()));
+    tags.add(new Tag<>("jobName", jobState.getJobName() == null ? "" : jobState.getJobName()));
+    tags.add(new Tag<>("jobId", jobState.getJobId()));
 
     tags.addAll(getCustomTagsFromState(jobState));
     return tags;
   }
 
   /**
-   * @deprecated
-   * Use {@link ClustersNames#getInstance()#getClusterName()}
+   * @deprecated use {@link ClustersNames#getInstance()#getClusterName()}
    */
   public static String getClusterIdentifierTag() {
     return ClustersNames.getInstance().getClusterName();
   }
-
 }
--- a/gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.yarn;
+
+import java.util.List;
+
+import org.apache.hadoop.yarn.api.records.ContainerId;
+
+import com.google.common.collect.ImmutableList;
+
+import gobblin.configuration.State;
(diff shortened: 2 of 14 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
```diff
@@ -35,11 +34,15 @@
   protected final String jobName;
 
   protected JobMetrics(JobState job) {
-    super(name(job), null, tagsForJob(job));
-    this.jobName = job.getJobName();
+    this(job, null);
```
liyinan926: Additional empty line.

gobblin-runtime/src/main/java/gobblin/runtime/util/JobMetrics.java
```diff
@@ -62,6 +65,17 @@ public static JobMetrics get(String jobId) {
   }
 
   /**
+   * Get a new {@link GobblinMetrics} instance for a given job.
+   *
+   * @param jobState he given {@link JobState} instance
```
liyinan926: Typo `he` => `the`.

gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
```diff
@@ -0,0 +1,55 @@
+package gobblin.yarn;
```
liyinan926: Missing open source header.

gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+
+import com.google.common.collect.Lists;
```
liyinan926: Wrong order of imports.

gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+    return (ContainerMetrics) GOBBLIN_METRICS_REGISTRY
+        .getOrDefault(name(containerId), new ContainerMetrics(containerState, applicationName, containerId));
+  }
+
+  private static String name(ContainerId containerId) {
+    return containerId.toString();
+  }
+
+  private static List<Tag<?>> tagsForContainer(State containerState, String applicationName, ContainerId containerId) {
+    List<Tag<?>> tags = Lists.newArrayList();
```
liyinan926: You can use a `ImmutableList`.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixTaskFactory.java
```diff
@@ -14,16 +14,21 @@
 
 import java.io.IOException;
 
+import com.codahale.metrics.Counter;
+
```
liyinan926: Wrong order of imports.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
```diff
@@ -305,11 +332,13 @@ public ParticipantShutdownMessageHandler(Message message, NotificationContext co
...
       @Override
-      public HelixTaskResult handleMessage() throws InterruptedException {
+      public HelixTaskResult handleMessage()
+          throws InterruptedException {
         String messageSubType = this._message.getMsgSubType();
-        Preconditions.checkArgument(
-            messageSubType.equalsIgnoreCase(HelixMessageSubTypes.WORK_UNIT_RUNNER_SHUTDOWN.toString()),
-            String.format("Unknown %s message subtype: %s", Message.MessageType.SHUTDOWN.toString(), messageSubType));
+        Preconditions
+            .checkArgument(messageSubType.equalsIgnoreCase(HelixMessageSubTypes.WORK_UNIT_RUNNER_SHUTDOWN.toString()),
```
liyinan926: The way the lines are broken looks weird.

liyinan926:
Mostly formatting comments, otherwise LGTM.

liyinan926:
Seems the build was hanging.

sahilTakiar:
Yes, it hangs locally too. Looking into it still.

sahilTakiar:
I'm still addressing the comments and fixing the build. However, I tested this on Nertz and it does not seem to be working. No metrics are being emitted and I am not sure why. It only works if I invoke `jobMetrics.startMetricsReporting`, but not when I do `containerMetrics.startMetricsReporting`.

@liyinan926, @ibuenros do either of you see anything wrong with this implementation?

My understanding is that since `ContainerMetrics` is the parent of `JobMetrics` and, `JobMetrics` is the parent of `TaskMetrics`, that if I invoke `containerMetrics.startMetricsReporting` everything should be reported properly.

sahilTakiar:
@liyinan926 and @ibuenros I found the issue.

It seems that in Typesafe's Config library the `.conf` files actually allow you to express the config in JSON format (according to the documentation on their Github page it actually uses a JSON superset called HOCON https://github.com/typesafehub/config#using-hocon-the-json-superset).

HOCON provides multiple ways of writing valid JSON (check the documentation link for examples). One such way is using the dot-notation. So a config such as `my.test.config=30` will actually be converted to

```
{
  "my" : {
    "test" : {
      "config" : "30"
    }
  }
}
```

The reason that this is problematic is that if you have the following two lines in your config file:

```
metrics.reporting.kafka.topic=MetricReport
metrics.reporting.kafka.topic.events=GobblinTrackingEvent
```

Then the resulting config is:

```
{
  "metrics": {
    "reporting": {
      "kafka": {
        "topic": {
          "events": "GobblinTrackingEvent"
        }
      }
    }
  }
}
```

My understanding is that you the first config sets a node in the JSON tree to `MetricReport`, but then the second config over-writes that node to `{"events" : "GobblinTrackingEvent"}`.

This is exactly the bug that I was hitting when testing this on Nertz. No metrics were being reported because the code could not find the name of the Kafka topic to report to. What happens is that `GobblinMetrics.startMetricsReporting` takes in a `Properties` object. So we have to use `YarnHelixUtils.configToProperties` to convert a `Config` object to a `Properties` object. The method uses `Config.entrySet` to get all the configs in the `Config` object. It seems as though `entrySet` will only return the leaf values of the JSON tree, so the resulting `Properties` object has nothing set for `metrics.reporting.kafka.topic`.

Is this behavior expected? Do we know of a way around this? This seems very different from how `gobblin.configuration.State` works, so it seems we could easily hit this issue with other config keys?

For now I can unblock this PR, because it seems the `gobblin-metrics` uses 3 config keys for reporting data to Kafka `metrics.reporting.kafka.topic.metrics`, `metrics.reporting.kafka.topic.events`, and `metrics.reporting.kafka.topic`. The last value is used as the default if either of the other two are not set. So we can still get things to work, but I still think we should discuss this on Monday.

gobblin-yarn/src/main/java/gobblin/yarn/ContainerMetrics.java
sahilTakiar: What's the right order?

liyinan926: `org` should be above `com`. https://github.com/linkedin/gobblin/wiki/CodingStyle.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
sahilTakiar: I agree but according to Intellij it follows the Gobblin coding style.

sahilTakiar:
The issue above has been discussed offline.

PR updated to address all comments. Confirmed this is working as expected after testing on Nertz.

@liyinan926, @ibuenros any more comments?

liyinan926:
LGTM.

PR #520, opened 2015-12-12
Improvements for Javadoc generation

chavdar:
Change:
1. Added a new task javadocTarball that will create a full tarbal with all javadocs under build/gobblin-javadocs-all
2. Added logic to add links to online Gobblin javadocs so @link references to other projects can work

@liyinan926 @sahilTakiar can you review?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -231,7 +231,7 @@ permalink: /javadoc/${version}/
 //Javadoc initialization for subprojects
 ext.javadocVersion = null != project.version ? project.version : "latest"
 if (ext.javadocVersion.indexOf('-') > 0) {
-  //remove any "-" addons from the version
+  // Remove any "-" addons from the version
   ext.javadocVersion = javadocVersion.substring(0, javadocVersion.indexOf('-'))
 }
 
@@ -252,7 +252,7 @@ subprojects {
     
     rootProject.tasks.javadocTarball.dependsOn project.tasks.javadoc
     if ( rootProject.ext.javadocPackages.contains(project.name)) {
-      def cs = rootProject.tasks.javadocTarball.into(project.name){from(fileTree(dir: "${project.buildDir}/docs/javadoc/"))}
+      rootProject.tasks.javadocTarball.into(project.name){from(fileTree(dir: "${project.buildDir}/docs/javadoc/"))}
     } 
   }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ idea.project {
 }
 
 ext.build_script_dir = "${projectDir.path}/build_script"
-
+ext.javadocPackages = new ArrayList<String>()
 ext.isDefaultEnvironment = !project.hasProperty('overrideBuildEnvironment')
 
 File getEnvironmentScript()
@@ -203,6 +203,29 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
 
 import javax.tools.ToolProvider
 
+task javadocTarball(type: Tar) {
+  baseName = "gobblin-javadoc-all"
+  destinationDir = new File(project.buildDir, baseName)
+  compression = Compression.GZIP
+  extension = 'tgz'
+  description = "Generates a tar-ball with all javadocs to ${destinationDir}/${archiveName}"
+}
+
+javadocTarball << {
+  def indexFile = new File(destinationDir, "index.md")
+  def version = null != project.version ? project.version : "latest"
+  indexFile << """----
+layout: page
+title: Gobblin Javadoc packages ${version}
+permalink: /javadoc/${version}/
+----
+
+"""
+  rootProject.ext.javadocPackages.each {
+    indexFile << "* [${it}](${it})\n"
+  }
+}
+
 subprojects {
 plugins.withType(JavaPlugin) {
   plugins.apply('idea')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -203,6 +202,61 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
...
+  destinationDir = new File(project.buildDir, baseName)
+  compression = Compression.GZIP
+  extension = 'tgz'
+  description = "Generates a tar-ball with all javadocs to ${destinationDir}/${archiveName}"
+}
+
+javadocTarball << {
+  def indexFile = new File(destinationDir, "index.md")
+  def version = rootProject.ext.javadocVersion
+  indexFile << """----
```
liyinan926: Line 216-220 seem not well formatted in terms of indention.

build.gradle
```diff
@@ -203,6 +202,61 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
...
+}
+
+ext.javadocPackages = new HashSet<String>()
+subprojects.each{Project pr -> 
+  if (file(pr.projectDir.absolutePath + "/src/main/java").exists()) {
+    rootProject.ext.javadocPackages += pr.name
+  }
+}
+ext.javadocLinks = ext.javadocPackages.collect{"http://linkedin.github.io/gobblin/javadoc/${ext.javadocVersion}/${it}/"}
+//println "GOBBLIN_LINKS=" + ext.javadocLinks
```
liyinan926: Should this be deleted?

liyinan926:
LGTM other than the formatting comments.

build.gradle
```diff
@@ -380,6 +430,7 @@ plugins.withType(JavaPlugin) {
   tasks.javadoc.options.links "http://typesafehub.github.io/config/latest/api/",
                               "https://docs.oracle.com/javase/7/docs/api/",
                               "http://docs.guava-libraries.googlecode.com/git-history/v15.0/javadoc/"
+  rootProject.ext.javadocLinks.each{tasks.javadoc.options.links it}
```
sahilTakiar: Is there no programmatic way to fetch these for each library we use? Or do we have to manually add them?

If its manual it would probably be good to add the Hadoop javadocs (some other nice ones would be Avro,

build.gradle
```diff
@@ -203,6 +202,61 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
...
+ext.javadocPackages = new HashSet<String>()
+subprojects.each{Project pr -> 
+  if (file(pr.projectDir.absolutePath + "/src/main/java").exists()) {
+    rootProject.ext.javadocPackages += pr.name
+  }
+}
+ext.javadocLinks = ext.javadocPackages.collect{"http://linkedin.github.io/gobblin/javadoc/${ext.javadocVersion}/${it}/"}
+//println "GOBBLIN_LINKS=" + ext.javadocLinks
+
+subprojects {
```
sahilTakiar: Can this subprojects block be combined with the one below?

build.gradle
chavdar: That's intentional. We are storing a multiline string to a file and indentation will become part of the value which is incorrect.

build.gradle
chavdar: For external libraries, there is no way we can add those programatically. We have to add those by hand.

For internal, it should be possible to figure out the links from projects in dependencies but it is a bit trickier. I'll leave that for V.next;

build.gradle
chavdar: Yes, I find it harder to read when we have one ginormous subprojects {} block. Further, "subproject {} " (configuration) is different from "subproject.each {}" (iteration)  and those get evaluated at different times.

chavdar:
@liyinan926 @sahilTakiar updated with a few more links

build.gradle
```diff
@@ -203,6 +205,59 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
...
+  plugins.withType(JavaPlugin) {
+ 
+    //Sometimes generating javadocs can lead to OOM. This may needs to be increased.
+    javadoc {
+      options.JFlags = ["-Xmx256m"]
+    }
+    
+    rootProject.tasks.javadocTarball.dependsOn project.tasks.javadoc
+    if ( rootProject.ext.javadocPackages.contains(project.name)) {
+      def cs = rootProject.tasks.javadocTarball.into(project.name){from(fileTree(dir: "${project.buildDir}/docs/javadoc/"))}
```
sahilTakiar: is `def cs` used anywhere?

build.gradle
```diff
@@ -203,6 +205,59 @@ task wrapper(type: Wrapper) { gradleVersion = '1.12' }
...
+subprojects.each{Project pr -> 
+  if (file(pr.projectDir.absolutePath + "/src/main/java").exists()) {
+    rootProject.ext.javadocPackages += pr.name
+  }
+}
+
+subprojects {
+  plugins.withType(JavaPlugin) {
+ 
+    //Sometimes generating javadocs can lead to OOM. This may needs to be increased.
```
sahilTakiar: space after the `//`, same comment above.

sahilTakiar:
Minor comments, otherwise LGTM.

build.gradle
chavdar: The cs variable not, but the rest of the line adds the javadoc files to the common tarball.

chavdar:
Superseded by the squashed PR: https://github.com/linkedin/gobblin/pull/527

PR #524, opened 2015-12-14
Add RawDataset and RawDatasetRetentionPolicy

zliu41:
A `RawDataset` is a dataset with a corresponding refined dataset (e.g., raw dataset = tracking hourly, refined dataset = tracking daily). A version of a raw dataset is deletable only if it satisfies certain conditions with the corresponding version of the refined dataset. For example, the modification time of the raw dataset should be earlier than the modification time of the corresponding refined dataset.

A `RawDatasetRetentionPolicy` embeds another `RetentionPolicy`. In `listDeletableVersions(List)` it applies the embedded `RetentionPolicy`'s predicate, as well as `listQualifiedRawDatasetVersions(Collection)`.

### DIFF
--- a/gobblin-data-management/build.gradle
+++ b/gobblin-data-management/build.gradle
@@ -17,6 +17,7 @@ dependencies {
   compile project(":gobblin-utility")
 
   compile externalDependency.commonsLang
+  compile externalDependency.commonsLang3
   compile externalDependency.guava
   compile externalDependency.slf4j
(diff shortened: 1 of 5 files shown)


zliu41:
@ibuenros @chavdar can you review?

zliu41:
This is a temporary solution and is expected to be deprecated in the near future.

PR #530, opened 2015-12-16
Update job execution info in the execution history store upon every task completion

liyinan926:
1. Introduced a new class `DistributedJobLauncher` from which the `MRJobLauncher` and `GobblinHelixJobLauncher` inherit. `DistributedJobLauncher` handles tasks common to both subclasses necessary for launching jobs on a distributed execution framework such as Hadoop MR and Yarn.
2. `DistributedJobLauncher` uses a single-thread scheduled collector to collect output `TaskState`s periodically, and post an `NewOutputTaskStateEvent` through an `EventBus` whenever a new output `TaskState` is collected. The `JobContext` now subscribes to the `NewOutputTaskStateEvent`s and update the job execution info stored in the execution history store. Output `TaskState` files get deleted once the `TaskState`s they store are successfully collected so the next run of the collector only sees new files.
3. `LocalJobLauncher` has also been refactored to update the job execution info upon every task completion.


liyinan926:

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -331,6 +321,7 @@ protected void postProcessTaskStates(List<TaskState> taskStates) {
    * Subclasses can override this method to do whatever processing on the {@link JobState} and its
    * associated {@link TaskState}s, e.g., aggregate task-level metrics into job-level metrics.
    */
+  @SuppressWarnings("deprecation")
```
zliu41: Can you explain in Javadoc why this is deprecated and what should the caller use instead.

liyinan926: No, this is not deprecated. `postProcessTaskStates` is. This is just to surpress deprecation warning because of the call to `postProcessTaskStates`.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -267,6 +272,13 @@ void storeJobExecutionInfo() {
     }
   }
 
+  @Subscribe
+  public void handleNewOutputTaskStateEvent(NewOutputTaskStateEvent newOutputTaskStateEvent) {
+    this.jobState.addTaskState(newOutputTaskStateEvent.getTaskState());
+    // Update the job execution history store upon every task completion
+    storeJobExecutionInfo();
```
zliu41: Calling this for every `newOutputTaskStateEvent`, what's the overhead? Is it possible to call this method after a certain number of `newOutputTaskStateEvent` have been received?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -415,9 +417,20 @@ public void markTaskCompletion() {
...
     this.taskState.setProp(ConfigurationKeys.TASK_RETRIES_KEY, this.retryCount.get());
   }
 
+  /**
+   * Return whether this {@link Task} is completed or not.
+   *
+   * @return {@code true} if this {@link Task} is completed or {@code false} otherwise
+   */
+  public boolean isCompleted() {
+    return this.completed;
```
zliu41: Can you use TaskState.COMMITTED/SUCCESSFUL instead? Also, you may want to clarify whether "completed" means succeeded, or either succeeded or failed.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
liyinan926: The overhead is pretty much negligible compared with the cost of collecting the `TaskState`. But yes we can definitely change this to carry a list of `TaskState`s.

liyinan926:
@zliu41 updated b04f803.

gobblin-runtime/src/main/java/gobblin/runtime/local/LocalTaskStateTracker.java
```diff
@@ -93,9 +103,16 @@ public void onTaskCompletion(Task task) {
...
 
+    // Add the TaskState of the completed task to the JobState so when the control
+    // returns to the launcher, it sees the TaskStates of all completed tasks.
+    this.jobState.addTaskState(task.getTaskState());
+
     // Mark the completion of this task
     task.markTaskCompletion();
 
+    // Notify the listeners for the completion of the task
+    this.eventBus.post(new NewTaskCompletionEvent(ImmutableList.of(task.getTaskState())));
```
zliu41: There doesn't seem to be any benefit doing it here. You can just post this event once in LocalJobLauncher after all tasks finished.

liyinan926: The goal is to trigger an update of the job execution store upon every task completion.

zliu41:
LGTM

gobblin-runtime/src/main/java/gobblin/runtime/DistributedJobLauncher.java
```diff
@@ -0,0 +1,179 @@
...
+ *   output {@link TaskState}s stored in files output by the mappers or containers.
+ * </p>
+ *
+ * @author ynli
+ */
+public abstract class DistributedJobLauncher extends AbstractJobLauncher {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DistributedJobLauncher.class);
+
+
```
sahilTakiar: Delete extra line

gobblin-runtime/src/main/java/gobblin/runtime/DistributedJobLauncher.java
```diff
@@ -0,0 +1,179 @@
...
+      @Override
+      public void run() {
+        try {
+          collectOutputTaskStates(outputTaskStateDir);
+        } catch (IOException ioe) {
+          LOGGER.error("Failed to collect output task states", ioe);
+          jobContext.getJobState().setState(JobState.RunningState.FAILED);
+        }
+      }
+    }, 0, this.outputTaskStatesCollectorIntervalSeconds, TimeUnit.SECONDS);
```
sahilTakiar: Set initial delay to `this.outputTaskStatesCollectorIntervalSeconds`

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -123,7 +118,9 @@ public MRJobLauncher(Properties jobProps) throws Exception {
   }
 
   public MRJobLauncher(Properties jobProps, Configuration conf) throws Exception {
-    super(jobProps, ImmutableMap.<String, String> of());
+    super(jobProps, FileSystem
```
sahilTakiar: Can you move the creation of the `FileSystem` into its own method (something like `buildFileSystem`), so that the `super` declaration isn't so long.

gobblin-runtime/src/main/java/gobblin/runtime/DistributedJobLauncher.java
```diff
@@ -0,0 +1,179 @@
...
+  // Number of ParallelRunner threads to be used for state serialization/deserialization
+  protected final int stateSerDeRunnerThreads;
+
+  // Interval in seconds between two runs of the collector of output TaskStates
+  private final int outputTaskStatesCollectorIntervalSeconds;
+
+  // A scheduled executor for the collector of output TaskStates
+  private final ScheduledExecutorService outputTaskStatesCollectorExecutor;
+
+  public DistributedJobLauncher(Properties jobProps, FileSystem fs, Map<String, String> eventMetadata)
```
sahilTakiar: Instead of making this a parent class I would suggest either:

1: Making a `TaskStateCollectorService implements AbstractScheduledService` and discard `DistributedJobLauncher`. I don't see much benefit in putting all the task collection logic in a parent class.

or

2: Make `DistributedJobLauncher implement AbstractIdleService` and make `MRJobLauncher implement AbstractIdleService`. The `start()` method of `DistributedJobLauncher` invoke `scheduleOutputTaskStatesCollector` and have the `start()` method of `MRJobLauncher()` invoke the `super` method, as well as starting any other thread pools it needs.

liyinan926: Can you explain what's the benefit of putting this logic in an `AbstractScheduledService` and having to use a `ServiceManager` to properly use it? My thought is this logic is specific to the launching and execution of a job, instead of a generic one that can be used in other places. Additionally we also need to call `collectOutputTaskStates` one last time manually to avoid missing any output `TaskState`s.

sahilTakiar: The collection of Task States from HDFS and their de-serialization into memory can be thought of as a Scheduled Service. It just reads from HDFS on a periodic basis and then posts it's result on an `EventBus`. Its a service that needs to be run by a job launcher during a job's execution. We can use `AbstractedScheduledService.runOneIteration` to do the final collection.

IMO it fits into a Service pattern more than into an Abstract class pattern, but it works either way, so I will leave it up to you.

liyinan926: Sure. We cannot have `MRJobLauncher extends AbstractedScheduledService` since it already `extends DistributedJobLauncher`. So option 1 would be the only choice. Will update the PR.

liyinan926:
@sahilTakiar updated 8ed6cfc.

sahilTakiar:
LGTM

PR #531, opened 2015-12-16
config management common impl

tuGithub:
@chavdar @ibuenros 

Please review this PR


gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
```diff
@@ -0,0 +1,103 @@
...
+   * @param uri - current URI to check with
+   * @param previous - all previously imports
+   */
+  private static void checkAncestorCircularDependency(ConfigStore cs, String version, URI initialURI, URI uri,
+      List<URI> previous) throws VersionDoesNotExistException{
+    String pStr = uri.getPath();
+    if (pStr.length() == 0)
+      return;
+
+    Path p = (new Path(uri.getPath())).getParent();
```
ibuenros: We shouldn't depend on Hadoop to do these operations. See http://stackoverflow.com/questions/10159186/how-to-get-parent-url-in-java on how to get parent uri.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * 
+   * @param cs - ConfigStore to check with
+   * @param version - ConfigStore version to check with
+   * @param initialURI - the initial URI passed from public function checkCircularDependency(...)
+   * @param uri - current URI to check with
+   * @param previous - all previously imports
+   */
+  private static void checkAncestorCircularDependency(ConfigStore cs, String version, URI initialURI, URI uri,
```
ibuenros: Can you include an explanation on how the algorithm works?

ibuenros: I would suggest simplifying the algorithm a bit: have a private method `getAllOwnImports` that concatenates `ownImports` and the implicit parent import. Then treat them as a single class of imports, ie. you don't need both `checkAncestorCircularDepdendency` and `checkImportCircularDependency`, a single one will be enough.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
```diff
@@ -0,0 +1,103 @@
...
+import gobblin.config.store.api.VersionDoesNotExistException;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.hadoop.fs.Path;
+
+
+public class CircularDependencyChecker {
```
ibuenros: The methods in this class just check whether a particular URI is a seed for a cycle. I am assuming the intention of this is to first check for cycles, then do resolution? It should be possible to check for cycles at the same time as resolution is being computed.
I think it would be useful to have a method to check for cycles in the entire store, without requiring a seed. You can use Strongly connected component finding algorithms to do this ( for example https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm ).

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
```diff
@@ -0,0 +1,103 @@
...
+    // should check parent imports chain
+    checkImportCircularDependency(cs, version, parent, parent, current);
+
+    if (parent != null) {
+      checkAncestorCircularDependency(cs, version, initialURI, parent, current);
+    }
+  }
+
+  private static void checkImportCircularDependency(ConfigStore cs, String version, URI initialURI, URI uri,
+      List<URI> previous) throws VersionDoesNotExistException{
```
ibuenros: We should declare that `CircularDependencyException` is thrown. I'm fine with it being a RuntimeException, but users should be aware what to catch if they care about it.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+    String auth2 = child.getAuthority();
+    result = stringEquals(auth1, auth2);
+    if (!result)
+      return result;
+
+    String path1 = parent.getPath();
+    String path2 = child.getPath();
+
+    // path2 starts with path1
+    if (path2.equals(path1) || path2.indexOf(path1 + "/") == 0) {
```
ibuenros: Couldn't you just use `path2.startsWith(path1)`?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+  public static boolean checkDescendant(URI parent, URI child) {
+    if (parent == null && child == null) {
+      return true;
+    }
+
+    if (parent == null || child == null) {
+      return false;
+    }
+
+    boolean result = false;
```
ibuenros: you don't need this variable

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+    String path2 = child.getPath();
+
+    // path2 starts with path1
+    if (path2.equals(path1) || path2.indexOf(path1 + "/") == 0) {
+      return true;
+    }
+    return false;
+  }
+
+  public static boolean stringEquals(String s1, String s2) {
```
ibuenros: Could use either apache commons or guava to do this.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.config.common.impl;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.apache.hadoop.fs.Path;
```
ibuenros: We should not be using Hadoop in this package.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/ImportMappings.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package gobblin.config.common.impl;
+
+import java.net.URI;
+import java.util.Collection;
+import java.util.Map;
+
+
+public interface ImportMappings {
```
ibuenros: Can you add a javadoc to the interface?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Utility class to provide {@ConfigStoreWithResolution} functionality based on given {@ConfigStore}
+ * @author mitu
+ *
+ */
+public class SimpleConfigStoreResolver implements ConfigStoreWithResolution {
+
+  private static final Logger LOG = Logger.getLogger(SimpleConfigStoreResolver.class);
+  private final ConfigStore store;
```
ibuenros: Consider using Lombok's @delegate ( https://projectlombok.org/features/Delegate.html )

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+    LOG.info(String.format("Getting resolved config for uri: %s for version %s, with traceID: %s", 
+        uri, version, traceID));
+    return getResolvedConfigWithTrace(uri, version, traceID.toString());
+  }
+
+  protected Config getResolvedConfigWithTrace(URI uri, String version, String traceID) throws VersionDoesNotExistException{
+    CircularDependencyChecker.checkCircularDependency(store, version, uri);
+    Config self = this.getOwnConfig(uri, version);
+    for (Entry<String, ConfigValue> entry : self.entrySet()) {
+      LOG.info(String.format("key %s, value %s from uri %s, with traceID %s",
```
ibuenros: this should be a debug (or even trace) log

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  @Override
+  public List<URI> getImportsRecursively(URI uri, String version) throws VersionDoesNotExistException{
+    CircularDependencyChecker.checkCircularDependency(this, version, uri);
+
+    List<URI> result = getOwnImports(uri, version);
+    // root can not include anything, otherwise will have circular dependency 
+    if (isRootURI(uri)) {
+      return result;
```
ibuenros: if uri is root, you should return empty list. Not necessary to get own imports.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    // The order is already decided in imported List : higher priority appear in the head
+    List<URI> imported = this.getOwnImports(uri, version);
```
ibuenros: you already have the own imports stored in `result`.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
```diff
@@ -0,0 +1,119 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  private void addToImportMapping(URI u, Collection<URI> imported, Map<URI, Collection<URI>> targetImportMap,
+      Map<URI, Collection<URI>> targetImportedByMap) {
+    if (!targetImportMap.containsKey(u)) {
+      targetImportMap.put(u, new HashSet<URI>());
+    }
+    targetImportMap.get(u).addAll(imported);
```
ibuenros: You can use Guava's multimap to simplify this a bit.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
```diff
@@ -0,0 +1,119 @@
...
+      Collection<URI> currentLevel = this.store.getChildren(root, version);
+
+      while (currentLevel.size() != 0) {
+        Collection<URI> nextLevel = new ArrayList<URI>();
+        for (URI u : currentLevel) {
+          List<URI> ownImported = this.store.getOwnImports(u, version);
+          this.addToImportMapping(u, ownImported, this.importMap, this.importedByMap);
+
+          if (this.store instanceof ConfigStoreWithResolution) {
+            Collection<URI> resolved = ((ConfigStoreWithResolution) this.store).getImportsRecursively(u, version);
```
ibuenros: If the store is not `ConfigStoreWithResolution`, the recursive mappings will not be filled. You should specify this somewhere.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
```diff
@@ -0,0 +1,119 @@
...
+  public SimpleImportMappings(ConfigStore cs, String version) throws VersionDoesNotExistException{
+    this.store = cs;
+    this.version = version;
+    buildMaps();
+  }
+
+  private void buildMaps() throws VersionDoesNotExistException{
+    try {
+      URI root = new URI("");
+      Collection<URI> currentLevel = this.store.getChildren(root, version);
```
ibuenros: nit: you don't need a `currentLevel` and `nextLevel`, you can do the usual BFS using a single queue.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
tuGithub: nice catch, fixed it

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
tuGithub: if path2 is /user/mitu/foobar and path1 is /user/mitu/foo, then this will not work

Added more checks into this class

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/CircularDependencyChecker.java
chavdar: I agree. There should be two circular dependency checkers. One should be used during import resolution. It's a matter of simply keeping state of which URL have already been imported. When we see a new import we check if it is in the "seen" set. If yes, there is a circular dependency, otherwise add it to the set. The is obviously P1.

The second implementation is  a store validation that can be run when we build a new config artifact for deployment. In this case, we need to check for any cycles. This is P2.

Further, I don't know why all the methods are static. There is obviously some state, say the ConfigStore and versio that is carried around. This should be put in object state.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+  public static URI getParentURI(URI uri) {
+    String pStr = uri.getPath();
+    if (pStr.length() == 0)
+      return null;
+
+    Path p = (new Path(uri.getPath())).getParent();
+    try {
+      return new URI(uri.getScheme(), uri.getAuthority(), p.toString(), uri.getQuery(), uri.getFragment());
+    } catch (URISyntaxException e) {
+      // Should not come here
```
chavdar: Throw an AssertionError. Obviously, something wrong happened.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/ImportMappings.java
tuGithub: done

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
tuGithub: import lombok.Delegate; is deprecated

the replaced one is still in experimental, will use that later

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
tuGithub: changed to debug level

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    boolean result = false;
+    String scheme1 = parent.getScheme();
+    String scheme2 = child.getScheme();
+    result = stringEquals(scheme1, scheme2);
+    if (!result)
+      return result;
+
+    String auth1 = parent.getAuthority();
```
chavdar: We need some concept of config key URI normalization. It is possible that one of the config key URIs is missing authority, i.e. we need to use the default authority . If the method assumes normalized config key URIs then, it should be stated in the javadocs.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
chavdar: You can still use path2.startsWith(path + "/") which is more efficient. In the case of startsWith, we'll stop at the first mismatch. In the case of indexOf, the whole path2 needs to be scanned.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
ibuenros: Instead of using `String.format` you can do `LOG.info("key %s, value %s from uri %s, with traceID %s", entry.getKey(), ...)`. It is faster, as it only assembles the string if it is needed for logging.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/PathUtils.java
```diff
@@ -0,0 +1,99 @@
...
+      return new URI(uri.getScheme(), uri.getAuthority(), p.toString(), uri.getQuery(), uri.getFragment());
+    } catch (URISyntaxException e) {
+      // Should not come here
+      e.printStackTrace();
+    }
+    return null;
+  }
+
+  /**
+   * @param parent - parent URI
```
chavdar: I'd like to stop using URI in comments and variable names as it is very non-descriptive. We should use "config key", "config key URI", etc.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
chavdar: I'd say, don't even implement the ConfigStoreWithResolution altogether. Stores that need the functionality can instantiate the resolver rather than extend it.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
tuGithub: added the job doc in the constructor

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+  @Override
+  public Config getResolvedConfig(URI uri, String version) throws VersionDoesNotExistException{
+    UUID traceID = UUID.randomUUID();
+    LOG.info(String.format("Getting resolved config for uri: %s for version %s, with traceID: %s", 
+        uri, version, traceID));
+    return getResolvedConfigWithTrace(uri, version, traceID.toString());
+  }
+
+  protected Config getResolvedConfigWithTrace(URI uri, String version, String traceID) throws VersionDoesNotExistException{
+    CircularDependencyChecker.checkCircularDependency(store, version, uri);
```
chavdar: Coming back to my original comment about the checker. It should be a stateful object that keeps track of the "seen" imports and gets pass around when we recurs into getResolvedConfigWithTrace()

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+    URI parent = PathUtils.getParentURI(uri);
+    Config res = getResolvedConfigWithTrace(parent, version, traceID);
+    return res;
+  }
+
+  protected static final boolean isRootURI(URI uri) {
+    if (uri == null)
+      return false;
+
+    return uri.toString().length() == 0;
```
chavdar: Isn't this uri.getPath() == "/" ?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleConfigStoreResolver.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  @Override
+  public List<URI> getImportsRecursively(URI uri, String version) throws VersionDoesNotExistException{
```
chavdar: Should this method also return implicit imports coming from the ancestors including the ancestors themselves? Can you remind me what the intended purpose of this method is?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
```diff
@@ -0,0 +1,119 @@
...
+import gobblin.config.store.api.VersionDoesNotExistException;
+
+
+/**
+ * Utility class to build the import mapping per {@ConfigStore} per version.
+ * 
+ * @author mitu
+ *
+ */
+public class SimpleImportMappings implements ImportMappings {
```
chavdar: I think the global cyclic checker should use an existing import mapping since it has the graph constructed already.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
tuGithub: fixed

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/SimpleImportMappings.java
tuGithub: fixed

chavdar:
LGTM

PR #534, opened 2015-12-16
Fixing bug with GobblinMetricsRegistry.getOrDefault

sahilTakiar:
While doing some testing for ETL-3891 I found that a bunch of duplicate `MetricContext`s were being created. The duplicate `MetricContext`s had name conflicts and thus had a `UUID` appended at the end of the name. The reason the duplicates were being created is due to the behavior of the `GobblinMetricsRegistry.getOrDefault()` method as well as the `get()` methods in `GobblinMetrics`, `JobMetrics`, `ContainerMetrics`, and `TaskMetrics`.

The `getOrDefault` behavior would construct the default value regardless of if the `id` already existed in the `GobblinMetricsRegistry` Cache. So basically, each invocation of `GobblinMetrics.get()` caused a new `GobblinMetrics` instance to be created, which in turn created a new `MetricContext`. These duplicate `MetricContext`s have no references so they are GC'ed pretty fast, but since it seems that every `MetricContext` has a `gobblin.metrics.notifications.timer` it still reports a bunch of garbage info.

@liyinan926 and @ibuenros can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
+++ b/gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
@@ -15,26 +15,30 @@ package gobblin.metrics;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 
 
 /**
- * Registry that stores instances of {@link GobblinMetrics} identified by an arbitrary string id.
- * The static method getInstance() provides a static instance of this this class that should be considered
- * the global registry of metrics.
- * An application could also instantiate one or more registries to for example separate instances of
- * {@link GobblinMetrics} into different scopes.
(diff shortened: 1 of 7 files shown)


gobblin-core/src/main/java/gobblin/metrics/GobblinMetricsRegistry.java
```diff
@@ -34,7 +38,7 @@
   private final Cache<String, GobblinMetrics> metricsMap = CacheBuilder.newBuilder().softValues().build();
```
liyinan926: This should be renamed to `metricsCache`.

liyinan926:
LGTM. The behavior you mentioned above was introduced as part of the efforts to fix a deadlock issue related to the use of `synchronized` around the `get` method in those classes. See #340 and #355. I didn't realize there's a `gobblin.metrics.notifications.timer` in each `MetricContext`, but a `Callable` should be used to load a value in case of a cache miss at the time to avoid creating temporary `MetricContext` instances.

ibuenros:
Good find! By the way, you can use the method `buildStrict()` when building a `MetricContext` to throw an error on conflict instead of silently modifying the name.

ibuenros:
LGTM

sahilTakiar:
I see, yes good to know.

PR #535, opened 2015-12-16
Change hive-exec dependency to core.jar

zliu41:
The original jar is a fat jar that contains dependencies such as commons-lang and Guava, which shadow the intended versions.

This also means that gobblin-core no longer has commons-lang, so changing to commons-lang3.


liyinan926:
LGTM.

PR #536, opened 2015-12-17
Fixing dangling MetricContext creation in ForkOperator

sahilTakiar:
The class `InstrumentedForkOperatorBase` would create a new `MetricContext` in its default constructor, but it would make it with no parent (the correct parent should get the `Task` level `MetricContext`). Then during the `init()` method it would create a new `MetricContext` (this time with the correct parent). The first `MetricContext` would be GC'ed quite quickly as it has no references, but it causes some duplicate reporting so it's best to remove it.

@ibuenros can you review

### DIFF
--- a/gobblin-core/src/main/java/gobblin/instrumented/fork/InstrumentedForkOperatorBase.java
+++ b/gobblin-core/src/main/java/gobblin/instrumented/fork/InstrumentedForkOperatorBase.java
@@ -35,36 +35,47 @@ import gobblin.metrics.Tag;
 
 
 /**
- * package-private implementation of instrumentation for {@link gobblin.fork.ForkOperator}.
- * See {@link gobblin.instrumented.fork.InstrumentedForkOperator} for extensible class.
+ * Package-private implementation of instrumentation for {@link gobblin.fork.ForkOperator}.
+ *
+ * @see {@link gobblin.instrumented.fork.InstrumentedForkOperator} for extensible class.
  */
 abstract class InstrumentedForkOperatorBase<S, D> implements Instrumentable, ForkOperator<S, D> {
 
-  private boolean instrumentationEnabled = false;
-  protected final Closer closer = Closer.create();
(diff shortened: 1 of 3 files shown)


liyinan926:
LGTM. Seems the build failed on Java 7 due to the following test failure. Maybe you can give a longer timeout to it in this PR:

```
Gradle test > gobblin.scheduler.JobConfigFileMonitorTest.testAddNewJobConfigFile FAILED
    java.util.concurrent.TimeoutException: Timeout waiting for condition '4 scheduled jobs'.
        at gobblin.testing.AssertWithBackoff.assertTrue(AssertWithBackoff.java:185)
        at gobblin.testing.AssertWithBackoff.assertTrue(AssertWithBackoff.java:134)
        at gobblin.testing.AssertWithBackoff.assertEquals(AssertWithBackoff.java:147)
        at gobblin.scheduler.JobConfigFileMonitorTest.testAddNewJobConfigFile(JobConfigFileMonitorTest.java:91)
```

sahilTakiar:
Increased the timeout from 5000 ms to 7500 ms. The test also failed on Hudson, so this change should also prevent that from happening again.

liyinan926:
LGTM.

ibuenros:
LGTM

bjvanov:
@sahilTakiar It looks like I just experienced something similar which caused my Java 8 default Travis build to fail. It happened during my latest commit to [PR1074](https://github.com/linkedin/gobblin/pull/1074)

From the raw log:
image

Sorry for the picture but I'm on my cell. Do you think this is related to this previous bug?

sahilTakiar:
The test `JobConfigFileMonitorTest` has notoriously been very shaky on Travis. IIRC it fails semi often. Generally, just re-running the Travis build fixes it. It seems the Travis build for https://github.com/linkedin/gobblin/pull/1074 is working now, so it shouldn't be an issue.

PR #539, opened 2015-12-17
Fixed an incorrect import issue.

ibuenros:


ibuenros:

PR #540, opened 2015-12-17
Fixed an oddness check.

ibuenros:


ibuenros:

PR #541, opened 2015-12-17
Add support for tunneling JDBC traffic over an HTTP proxy

kunalkandekar:
Frequently data stores to be accessed by Gobblin reside outside data centers. In these cases, outbound access from a data center typically needs to go through a gateway proxy for security purposes. In some cases this is an HTTP proxy. However, some protocols like JDBC don't support the concept of "proxies", let alone HTTP proxies, and hence a solution is needed to enable this.

This Pull Request provides a method of tunneling JDBC connections over an HTTP proxy. Note that it is currently only implemented for JDBC, but can be extended to work with any other TCP-based protocol.

The way this works for JDBC is:
1. When a JdbcProvider is invoked, it checks if the WorkUnit has a proxy host and port defined.
2. If so, it extracts the remote host and port from the connectionUrl.
3. It then creates a Tunnel instance configured with the remote host and port. The Tunnel starts a thread that listens on an arbitrary port on localhost.
4. The JdbcProvider then changes the connectionUrl to replace the remote host and port with the localhost and port and passes it on to the driver.
5. Hence when the driver creates a JDBC connection, it connects to the Tunnel socket. The Tunnel then connects to the remote host through the proxy via a HTTP CONNECT request.
6. If established successfully, the Tunnel then just relays bytes back and forth.
7. When the JDBC data source is closed down, the Tunnel is shut down as well.

The Tunnel can accept as many connections as the JdbcExtractor opens. The Tunnel uses NIO to minimize resource usage.


chavdar:
It looks like the patch uses Java 8 features. Currently, Gobblin is Java 7-based.

kunalkandekar:
Ah, right. Will fix.

kunalkandekar:
The test that fails has a 137 return code from its Gradle Test Executor. It could be that a specific test is failing because it's using up too much memory, in which case TravisCI SIGKILLS it (see: https://discuss.gradle.org/t/gradle-travis-ci/11928/3). I'll try to fix the test to use less memory.

kunalkandekar:
@chavdar It seems TravisCI is killing certain test processes (possibly because they are consuming too many resources.) I just significantly reduced the memory usage of some tests, but they still continue to get killed. Can you think of anything we can do to debug or sidestep this issue?

kunalkandekar:
While I'm trying to debug the TravisCI/gradle issue, here are instructions for performing end-to-end testing for these changes using gobblin standalone and a publicly accessible MySQL database: https://gist.github.com/kunalkandekar/b4f8c5fff6df62084e93

chavdar:
@sahilTakiar can you please help reviewing this?

sahilTakiar:
Will start reviewing today.

sahilTakiar:
Few general comments:

1: Can you squash all your commits into a single commit (http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git)
2: Can you remove all mentions of "GaaP" so that this code is not tied to a specific proxy - you should just be able to make it a configuration option (like `user.agent.name` or something along those lines)
3: Is there a more detailed design doc on how this all works?
5: Adding more javadocs to the `Tunnel.java` class would be very helpful, and consider breaking the `Tunnel.java` class down into multiple classes as it is pretty long.

kunalkandekar:
Thanks for the comments, will incorporate the feedback.

liyinan926:
@kunalkandekar just one more comment: Gobblin is not using the standard LinkedIn coding style so none of our classes use variable names starting with `_`. Check it out here https://github.com/linkedin/gobblin/wiki/CodingStyle.

kunalkandekar:
Thanks @sahilTakiar, am incorporating your feedback. Out of curiosity, could you explain the rationale between squashing commits into a single one? There is no detailed design doc, only the comment on the original commit, will include details in the JavaDocs.

sahilTakiar:
We generally try to have 1 commit per new feature in Gobblin. This way we can have a single, descriptive commit message describing the change.

kunalkandekar:
@sahilTakiar I see. Would it make sense to squash the commits once I push my current changes or should I do so beforehand?

chavdar:
@pcadabam Can you also help @sahilTakiar with the review?

pcadabam:
Looking into it.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,501 @@
...
+
+import static java.nio.channels.SelectionKey.OP_READ;
+
+
+/**
+ * @author navteniev@linkedin.com
+ *
+ * Implements a tunnel through a proxy to resource on the internet
+ */
+public class Tunnel {
```
pcadabam: This class seems too long. Can you pull out the private classes into a different file? Also Can you add more javadoc for the public methods and the classes?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,501 @@
...
+    try {
+      _server = ServerSocketChannel.open().bind(null);
+      startTunnelThread();
+
+      return this;
+    } catch (IOException ioe) {
+      LOG.error("Failed to open the tunnel", ioe);
+    }
+
+    return null;
```
pcadabam: Can you throw the IOException instead of returning null?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,501 @@
...
+        localAddress = _server.getLocalAddress();
+      }
+      if (localAddress instanceof InetSocketAddress) {
+        return ((InetSocketAddress) localAddress).getPort();
+      }
+    } catch (IOException e) {
+      LOG.error("Failed to get tunnel port", e);
+    }
+
+    return -1;
```
pcadabam: What does -1 mean? If you are using this as a sentinel can you create a named static constant?

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+@Test(singleThreaded = true)
+public class TestTunnelWithArbitraryTCPTraffic {
```
pcadabam: Can you add a test group?

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+    try {
+      Thread.sleep(millis);
+    } catch (InterruptedException e) {
+    }
+  }
+
+  @BeforeClass
+  void setUp() throws IOException {
+    _doubleEchoServer = startDoubleEchoServer();
+    System.out.println("Double Echo Server on " + _doubleEchoServer.getServerSocketPort());
```
pcadabam: Replace this with LOG

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+    _delayedDoubleEchoServer = startDoubleEchoServer(1000);
+    System.out.println("Delayed DoubleEchoServer on " + _delayedDoubleEchoServer.getServerSocketPort());
```
pcadabam: same as above

sahilTakiar:
@kunalkandekar you can go ahead to push all your local changes, and squash the commits once the PR is ready to be merged

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+    _talkFirstEchoServer = startTalkFirstEchoServer();
+    System.out.println("TalkFirstEchoServer on " + _delayedDoubleEchoServer.getServerSocketPort());
```
pcadabam: same as above

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+      ALL_THREADS.add(this);
+      return this;
+    }
+
+    @Override
+    public void run() {
+      try {
+        runQuietly();
+      } catch (Exception e) {
+        e.printStackTrace();
```
pcadabam: Remove

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
pcadabam: Please move the mockServer implementations out of this class and have have tests in this class.

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+  @Test(timeOut = 5000)
+  public void testDirectConnectionToEchoServer() throws IOException {
+    try {
+      SocketChannel client = SocketChannel.open();
+      client.connect(new InetSocketAddress("localhost", _doubleEchoServer.getServerSocketPort()));
+      writeToSocket(client, "Knock\n".getBytes());
+      String response = readFromSocket(client);
+      client.close();
+      assertEquals(response, "Knock Knock\n");
+    } catch (Exception e) {
```
pcadabam: The test should fail if there was an exception. Throw all exceptions

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+
+  // Baseline test to ensure clients work without tunnel
+  @Test(timeOut = 5000)
+  public void testDirectConnectionToEchoServer() throws IOException {
+    try {
+      SocketChannel client = SocketChannel.open();
+      client.connect(new InetSocketAddress("localhost", _doubleEchoServer.getServerSocketPort()));
+      writeToSocket(client, "Knock\n".getBytes());
+      String response = readFromSocket(client);
+      client.close();
```
pcadabam: Move to finally?

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
```diff
@@ -0,0 +1,859 @@
...
+      }
+    }.start();
+  }
+
+  private static String generateMsgFromServer(int i) {
+    return i + " " + StringUtils.repeat("Babble babble ", 10000) + "\n";
+  }
+
+  // Baseline tests to ensure simultaneous data transfer protocol is fine (disabled for now)
+  @Test(enabled = false, timeOut = 30000)
```
pcadabam: What is the unit for timeout? Is 30000 too high?

gobblin-tunnel/src/test/java/gobblin/tunnel/TunnelTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+  @AfterMethod
+  public void reset(){
+    _mockServer.reset();
+  }
+
+  @Test
+  public void mustBuildTunnelAndStartAcceptingConnections()
+      throws Exception {
+    Tunnel tunnel = Tunnel.build("example.org", 80, "localhost", PORT);
```
pcadabam: May be you should have a range of ports instead of just 10928. If that port is not available the build will fail.

gobblin-tunnel/src/test/java/gobblin/tunnel/TunnelTest.java
```diff
@@ -0,0 +1,238 @@
...
+    _mockServer.stop();
+  }
+
+  @AfterMethod
+  public void reset(){
+    _mockServer.reset();
+  }
+
+  @Test
+  public void mustBuildTunnelAndStartAcceptingConnections()
```
pcadabam: Can the tests run in parallel? They all seem to use the same port

pcadabam:
Overall looks good. 
Main comments around documentation and some refactor. Like @sahilTakiar said please squash the commits and update the pr.
Here are the steps

git rebase -i HEAD~<num_commits_to_squash>
An editor will show up where you can squash/pick for commits to squash.
git push upstream master --force

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: Sure, this also what @sahilTakiar suggested. I'm refactoring accordingly.

gobblin-tunnel/src/test/java/gobblin/tunnel/TestTunnelWithArbitraryTCPTraffic.java
kunalkandekar: It's milliseconds, so this is about 30 seconds. This test is usually done within a second on my box, but I wasn't sure how long it would take on more resource-constrained environments like Travis. Can reduce it to 5000 like other tests if this isn't what is causing the processes to be killed.

gobblin-tunnel/src/test/java/gobblin/tunnel/TunnelTest.java
kunalkandekar: Right, will try to make this more robust, probably by binding to any available port.

gobblin-tunnel/src/test/java/gobblin/tunnel/TunnelTest.java
kunalkandekar: I'll see if they can. My current concern is that if this spawns threads and we run too many tests in parallel, we may cause Travis to kill the process.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
```diff
@@ -21,18 +27,50 @@
...
+
+      try {
+        tunnel = Tunnel.build(remoteHost, remotePort, proxyHost, proxyPort);
+        int tunnelPort = tunnel.getPort();
+        //mangle connectionUrl, replace hostname with localhost -- hopefully the hostname is not needed!!!
+        String newConnectionUrl = connectionUrl.replaceFirst(remoteHost, "127.0.0.1")
+            .replaceFirst(":" + remotePort, ":" + tunnelPort);
+        connectionUrl = newConnectionUrl;
+      } catch (IOException ioe) {
+        LOG.warn("Failed to open tunnel via proxy " + proxyHost + ":" + proxyPort
```
sahilTakiar: This method should just `throw IOException` rather than catch the `IOException` and simply print a warning. This way the job will fail if the connection to the tunnel fails, which is better than simply printing a warning and allowing the job to continue.

gobblin-tunnel/src/main/java/gobblin/tunnel/AcceptHandler.java
```diff
@@ -0,0 +1,46 @@
+package gobblin.tunnel;
```
sahilTakiar: This class along with a few others are missing the Apache license header.

gobblin-tunnel/src/main/java/gobblin/tunnel/ProxySetupHandler.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handler for setting up the connection from the Tunnel to the intermediate proxy via HTTP CONNECT.
+ */
+class ProxySetupHandler implements Callable<HandlerState> {
+  private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(Tunnel.class);
```
sahilTakiar: Added `org.slf4j.Logger` to the imports list.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
```diff
@@ -21,18 +27,50 @@
...
+
+    if (proxyHost != null && proxyPort > 0) {
+      String remoteHost = "";
+      int remotePort = 0;
+      // TODO make connection Url parsing much more robust -- some connections URLs can have colons and slashes in the
+      // weirdest places
+      int hostStart = connectionUrl.indexOf("://") + 3;
+      int portStart =  connectionUrl.indexOf(":", hostStart);
+      remoteHost = connectionUrl.substring(hostStart, portStart);
+      remotePort = Integer.decode(connectionUrl.substring(portStart + 1, connectionUrl.indexOf("/", portStart)));
```
sahilTakiar: I think you can simply the `connectionUrl` parsing by creating a (URI)[https://docs.oracle.com/javase/7/docs/api/java/net/URI.html] via `URI.create(connectionUrl)` and then using methods such as `getScheme()` or `getPort()` to retrieve the relevant information.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
```diff
@@ -21,18 +27,50 @@
...
+
+      try {
+        tunnel = Tunnel.build(remoteHost, remotePort, proxyHost, proxyPort);
+        int tunnelPort = tunnel.getPort();
+        //mangle connectionUrl, replace hostname with localhost -- hopefully the hostname is not needed!!!
+        String newConnectionUrl = connectionUrl.replaceFirst(remoteHost, "127.0.0.1")
```
sahilTakiar: Can we make `127.0.0.1` configurable so that in case this port is needed it can be changed. The default can still be `127.0.0.1`

sahilTakiar: Ignore this forgot `127.0.0.1` is the same as `localhost`.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+  private Thread thread;
+  private final Config config;
+
+  private Tunnel(String remoteHost, int remotePort, String proxyHost, int proxyPort) {
+    config = new Config(remoteHost, remotePort, proxyHost, proxyPort);
+  }
+
+  private Tunnel open() throws IOException {
+    try {
+      server = ServerSocketChannel.open().bind(null);
```
sahilTakiar: Why bind this to a null `SocketAddress`?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @author navteniev@linkedin.com
+ * @author kkandekar@linkedin.com
+ */
+public class Tunnel {
+  public static final int NON_EXISTENT_PORT = -1;
+
+  private static final Logger LOG = LoggerFactory.getLogger(Tunnel.class);
+
+  private ServerSocketChannel server;
```
sahilTakiar: Why do we need a `ServerSocketChannel` over just a regular `SocketChannel`?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+        LOG.info("Closed tunnel.");
+      } catch (IOException ioe) {
+        LOG.warn("Failed to shutdown tunnel", ioe);
+      }
+    }
+  }
+
+  public static Tunnel build(String remoteHost, int remotePort, String proxyHost, int proxyPort)
+      throws IOException {
+    return new Tunnel(remoteHost, remotePort, proxyHost, proxyPort).open();
```
sahilTakiar: What happens when there are multiple invocations of this method within the same JVM? If I want to open multiple JDBC connections, each to a different URL will a new `ServerSocketChannel` be created for each URL?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
kunalkandekar: Actually, I tried that initially and it didn't work. I was surprised too, but for instance URI.create("jdbc:mysql://useastdb.ensembl.org:5306/homo_sapiens_core_82_38?user=anonymous").getScheme() returns "jdbc", getHost() returns null and getPort() returns -1. I think it's because JDBC connectionUrls look like "jdbc:mysql://somehost:12345", and the colon in the scheme throws the Java URI parsing logic off.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: That just means that it will get bound to whichever interface and port is available. Since we don't want to (or need to) configure specific ports, this suffices.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: This is the socket that listens for connections from the client (the JDBC driver), hence it's a server socket. Java distinguishes between server and client sockets, whereas this distinction does not really exist at the C socket layer, where a server socket is identified by the bind and listen calls.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+  private void startTunnelThread() {
+    thread = new Thread(new Dispatcher(), "Tunnel Listener");
+    thread.start();
+  }
+
+  public boolean isTunnelThreadAlive() {
+    return (thread != null && thread.isAlive());
+  }
+
+  private class Dispatcher implements Runnable {
```
sahilTakiar: I would suggest having this implement `Callable` instead of `Runnable`. The advantage is that `Callable` can throw checked exceptions (e.g. an `IOException`), and can have a return value.

In order to launch a `Callable`, you have to create invoke the [`ExecutorService.submit`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#submit(java.util.concurrent.Callable)) method. This method will return a [`Future`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html) object, which allows one to manage the lifecycle of the launched thread.

You make an `ExecutorService` using the following code:

`ExecutorService es = Executors.newSingleThreadExecutor(ExecutorUtils.newThreadFactory(Optional.of(LOG), Optional.of("Tunnel-Dispatcher-Thread-Pool"))`.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: It's not really URLs, but different hostnames. A Tunnel is currently pre-configured to always connect to the same hostname. But the Tunnel does support multiple concurrent connections to the same target host. So you could potentially use the same Tunnel instance for different URLs on the same hostname, depending on what the client requires and protocol supports.

If you want to connect to different hostnames, yes, you would need multiple instances of the Tunnel (and correspondingly the internal ServerSocketChannel). This is not currently a problem, because a) that would imply multiple JdbcProvider instances would be created, each with its own Tunnel instance and b) currently for JdbcExtractors, Gobblin does not appear to support multiple source.conn.* config settings per extractor anyway.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+        selector = Selector.open();
+      } catch (IOException e) {
+        LOG.error("Could not open selector",e);
+      }
+    }
+
+    @Override
+    public void run() {
+      try {
+        new AcceptHandler(server, selector, config);
```
sahilTakiar: Having this method in the `Dispatcher` seems a little counter-intuitive. It would be cleaner to remove this line and add the following code to the `open()` method (specifically after line 75).

`server.register(new AcceptHandler(selector, config))`

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
```diff
@@ -0,0 +1,178 @@
...
+    return (thread != null && thread.isAlive());
+  }
+
+  private class Dispatcher implements Runnable {
+
+    private Selector selector;
+
+    public Dispatcher() {
+      try {
+        selector = Selector.open();
```
sahilTakiar: Does this need to be initialized in the `Dispatcher`? You can probably create this in the `open()` method and pass it to the `Dispatcher` as a parameter.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: I see makes sense

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
sahilTakiar: hmm thats annoying, yes there doesn't seem to be a standard way of doing this in Java, so the substring operations should be sufficient.

gobblin-tunnel/src/main/java/gobblin/tunnel/AcceptHandler.java
```diff
@@ -0,0 +1,46 @@
...
+  private final Config config;
+
+  AcceptHandler(ServerSocketChannel server, Selector selector, Config config)
+      throws IOException {
+    this.config = config;
+    this.server = server;
+    this.selector = selector;
+
+    this.server.configureBlocking(false);
+    this.server.register(selector, SelectionKey.OP_ACCEPT, this);
```
sahilTakiar: Given the comments in `Tunnel.java` I think this two lines can be moved to the `Tunnel.open()` method.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: So will the current implementation re-use the same `Tunnel` instance for the same target hostname? Since the `JdbcProvider.connect` method is called in each `JdbcExtractor` that means for each new `JdbcExtractor` the method `Tunnel.build` will be invoked. That will in turn cause a new `ServerSocketChannel` and `Dispatcher` thread to be created. This seems inefficient if we can just re-use the same `Tunnel` instance right?

I ask because we currently run a single Gobblin job (a single JVM) that ingests from about 6-8 different sources, pulling in over 100 different tables. They don't all run at the same time, but most are scheduled to run pretty frequently. And each one will launch maybe 1-2 JdbcExtractors (on average). So I just want to make sure we don't start using too many resources.

kunalkandekar: True, right now, it's one Tunnel instance per JdbcProvider i.e. per JdbcExtractor. So it's a single server socket and a single thread per extractor instance, regardless of the number of connections, which off-hand does not seem too extravagant. I would say the biggest inefficiency is requiring two sockets per host (client -> Tunnel, Tunnel -> proxy) where one (client -> server) would do, but that just consumes file descriptors so it still seems affordable.

Agreed that it would be better to re-use the same Tunnel instance where possible, but that would lead to a large number of complications that may not be justified given the potential resource savings. We would have to implement some sort of coordination across multiple extractors so that they could share Tunnel instances where possible.

In our current job that ingests from 6 - 8 sources and a 100 different tables, if we have only up to 2 extractors, we would only have 2 Tunnels (with a server socket a thread each) for that job. My instinctive guess is that it should be OK with respect to resource usage, but maybe I need more information about the typical operating environment.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcProvider.java
kunalkandekar: I wrapped the IOException inside an IllegalStateException and re-threw it. That's an un-checked exception, so the JdbcProvider (and hence JdbcExtractor) interface does would get polluted with (hopefully rare) exceptions. Would that work?

sahilTakiar: Yes, that should work. Thanks.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: I see. Would it be possible to at least re-use the same `ServerSocketChannel`? That should already be do-able right? The `ServerSocketChannel` will create a new `SocketChannel` for each incoming connection from the `JDBCExtractor`, so there shouldn't be much coordination required as the actual connections are still separated.

Yes, I don't really know for sure that it would require a lot of resources, I just wanted to surface this issue in case it was a concern; but if you think it should be ok, then I think we can continue with the current approach and make any optimizations later on.

gobblin-tunnel/src/main/java/gobblin/tunnel/ProxySetupHandler.java
```diff
@@ -0,0 +1,178 @@
...
+  private int totalBytesRead = 0;
+  private final Config config;
+
+  ProxySetupHandler(SocketChannel client, Selector selector, Config config)
+      throws IOException {
+    this.config = config;
+    this.client = client;
+    this.selector = selector;
+    buffer = ByteBuffer.wrap(
+        String.format("CONNECT %s:%d HTTP/1.1\r\nUser-Agent: GobblinTunnel\r\nservice-name: gobblin\r\n" +
```
sahilTakiar: Can we not use a standard Apache `HttpClient` to do the HTTP CONNECT + to handle the response?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: Hmm, it may be possible to re-use the server socket, maybe by using a static member variable effectively making it a singleton, but that would make things quite complicated. The issue would be cleaning up of resources -- we need to know when to close the socket. We could do reference-counting of Tunnel instances, but even then, if the number of instances drops to 0, we'll close the socket. Then if another Tunnel instance is created, we'd have to open a new socket. All this would have to be thread-safe. While not rocket science, it seems like a lot of complexity to reduce the number of sockets we use. We could always take this approach if it turns out we need to.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: I guess that's possible, but the Dispatcher does not return anything, and there is nothing external waiting on the thread to complete, so the Future is unneeded. The Thread lifecycle is pretty simple: it's created in Tunnel#open and killed in Tunnel#close. A Callable could throw a checked Exception, but then we'd need to keep the Future and get() it, which would be in Tunnel#close. However, it would already been caught and logged in call() and we can't do anything more in close(). Seems to me both approaches are equivalent, except this uses an ExecutorService and Future instead of a Thread. Are there any particular advantages of one over the other?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: Makes sense, will try that out.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: Good idea. This will also allow the open() method to fail early and cleanly if Selector#open fails.

gobblin-tunnel/src/main/java/gobblin/tunnel/ProxySetupHandler.java
kunalkandekar: The latest version of Apache HttpClient (4.x) does not expose an HttpConnect method. It performs a CONNECT request automatically when configured with a proxy. It's possible to intercept the CONNECT request but not possible to send one explicitly.

Furthermore, since we need access to and control over the socket to do subsequent relaying, we would need a way to inject this socket into a HttpClient instance, and it gets quite involved. (See HttpClientFactory and HttpClientConnectionManager in Apache HttpClient.) This is probably the simplest way to do a CONNECT.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: I think there a few advantages:

* I think the `Dispatcher.run()` method should be throwing an `IOException` instead of catching it and logging it - mainly because I think it is better for these exceptions to be propagated up to the caller - there are a few places in the code where `IOExceptions` are caught and logged, but I think it is better practice to just throw them (unless they are caused my closing of resources) - mainly because we have seen lots of problems where this behavior causes a major feature to fail, but allows the job to continue - without a failed job our monitoring infra has no way of knowing a job is misbehaving / failing - so when in doubt we try to propagate exceptions as much as possible

* `ExecutoreService` allows you to use `ExecutorUtils.newThreadFactory(Optional.of(LOG), Optional.of("Tunnel-Dispatcher-Thread-Pool")` which logs any unchecked exceptions; if this isn't use and a thread throws an unchecked exception (e.g. `RuntimeException`), the thread will simply hang and probably cause the JVM to hang to and the logs won't tell us why

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: Yes, I was thinking using some type of singleton structure too. For now we can leave it be and if we see any issues we can re-visit this.

gobblin-tunnel/src/main/java/gobblin/tunnel/ProxySetupHandler.java
sahilTakiar: Ok I see.

gobblin-tunnel/src/main/java/gobblin/tunnel/ProxySetupHandler.java
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Handler for setting up the connection from the Tunnel to the intermediate proxy via HTTP CONNECT.
+ */
+class ProxySetupHandler implements Callable<HandlerState> {
+  private static final Logger LOG = LoggerFactory.getLogger(Tunnel.class);
+
+  private static final ByteBuffer OK_REPLY = ByteBuffer.wrap("HTTP/1.1 200".getBytes());
+  private static final Set<ByteBuffer> OK_REPLIES = new HashSet<ByteBuffer>(
+      Arrays.asList(OK_REPLY, ByteBuffer.wrap("HTTP/1.0 200".getBytes())));
```
sahilTakiar: Should this be `HTTP 1.1` instead of `HTTP 1.0` - might be good to make the HTTP version a string constant so that version changes are easy.

kunalkandekar: The HTTP/1.0 string is actually in addition to the HTTP/1.1 response. This ProxySetupHandler compares the response from the proxy against both strings. This is to handle the case where the intermediate proxy is based on HTTP 1.0 rather than 1.1. May help to move them to string constants, however.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: True, it's better for exceptions to be propagated to the caller, but in this case, the exception may occur in the middle of a long-ish running thread (e.g. remote server or proxy closes socket), which means it is actually occurring somewhere other the caller. That is, the exception may occur long after JdbcProvider calls Tunnel.open() or long before it calls Tunnel.close()  -- e.g. when fetching the query results. Hence it's not possible for the Tunnel to log the exception at an accurate place within the context of the rest of the Extractor.

However, when this happens, the JdbcExtractor would also inevitably experience an exception as the JDBC driver fails to communicate with the Tunnel, so that would be a more accurate place for the exception to be logged.

As for resource clean-up, when a thread throws an un-checked exception it simply exits. It won't be logged, which is a shortcoming, but it won't cause a hung thread. It's possible to handle that using  UncaughtExceptionHandler, but at that point I guess it may be better to use ExecutorService. I'll try that out.

kunalkandekar: Hmm, it seems ExecutorUtils is not a standard Java class, so we'd have to re-implement it. Also, using ExecutorService seems overkill if we only ever use it for one, un-reusable thread (see e.g. http://stackoverflow.com/questions/19000324/advantage-of-using-executors-newsinglethreadexecutor). In this case it seems better to use plain Threads and UncaughtExceptionHandler.

gobblin-tunnel/src/test/java/gobblin/tunnel/EasyThread.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A convenient, trackable, easy-to-cleanup wrapper around threads.
+ *
+ * @author kkandekar@linkedin.com
+ */
+abstract class EasyThread extends Thread {
+  static Set<EasyThread> ALL_THREADS = Collections.synchronizedSet(new HashSet<EasyThread>());
```
sahilTakiar: make this `final` so it can't be mutated by any other class

gobblin-tunnel/src/test/java/gobblin/tunnel/EasyThread.java
```diff
@@ -0,0 +1,43 @@
...
+    ALL_THREADS.add(this);
+    return this;
+  }
+
+  @Override
+  public void run() {
+    try {
+      runQuietly();
+    } catch (Exception ignored) {
+    }
```
sahilTakiar: Are we sure about this behavior? Do we not want to at least print the exception out to standard err?

gobblin-tunnel/src/test/java/gobblin/tunnel/MockServer.java
```diff
@@ -0,0 +1,124 @@
...
+              addSocket(clientSocket);
+              handleClientSocket(clientSocket);
+            } catch (IOException e) {
+              e.printStackTrace();
+              stopServer();
+            }
+          }
+        }.startThread());
+      } catch (IOException ignored) {
+      }
```
sahilTakiar: `ignored.printStackTrace`?

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
sahilTakiar: Is there a guarantee the `JdbcExtractor` will also throw an exception? Just to play devils advocate, is it possible that the `ReadWriterHandler` fails to reads a few bytes, drops them and logs an exception, but then succeeds to read the next set of bytes? In this scenario the `JdbcExtractor` wouldn't know that data has been dropped and wouldn't throw an exception.

gobblin-tunnel/src/main/java/gobblin/tunnel/ReadWriteHandler.java
```diff
@@ -0,0 +1,188 @@
...
+    this.client = client;
+    this.selector = selector;
+
+    // drain response that is not part of proxy's 200 OK and is part of data pushed from server, and push to client
+    if (mixedServerResponseBuffer.limit() > mixedServerResponseBuffer.position()) {
+      this.client.configureBlocking(true);
+      OutputStream clientOut = this.client.socket().getOutputStream();
+      clientOut.write(mixedServerResponseBuffer.array(), mixedServerResponseBuffer.position(),
+          mixedServerResponseBuffer.limit() - mixedServerResponseBuffer.position());
+      clientOut.flush();
```
sahilTakiar: add `clientOut.close()`

gobblin-tunnel/src/test/java/gobblin/tunnel/EasyThread.java
kunalkandekar: This is meant to catch a number of exceptions that are normally thrown as part of these tests, such as IOExceptions when sockets are closed or InterruptedExceptions when threads are interrupted. We could log these, but it would simply pollute the logs. No harm in that I guess.

gobblin-tunnel/src/test/java/gobblin/tunnel/MockServer.java
kunalkandekar: Same as above.

gobblin-tunnel/src/main/java/gobblin/tunnel/ReadWriteHandler.java
kunalkandekar: We don't want to close the client yet, since it will be used from this point forward to relay bytes back-and-forth. At this point we are just draining any part of the server-response that might have been intermingled with the proxy's 200 OK response before we start the main Tunnel operation.

gobblin-tunnel/src/main/java/gobblin/tunnel/Tunnel.java
kunalkandekar: There is no guarantee because it depends on the implementation of the internal JDBC driver being used, but it is highly likely assuming a sane implementation. It is not possible for the ReadWriteHandler to drop a few bytes and continue executing, because if there's an IOException, the socket is closed and the Tunnel thread exits. This should cause the JDBC driver on the other end to also throw some exception as it fails to read/write to the Tunnel socket.

Now if there is a non-IOException, that could be a problem, since the Tunnel thread will then exit without closing the sockets. In my testing, the MySql JDBC driver times out pretty quickly if the other end of the socket is closed, but to be more careful, I'll add a block to catch all exceptions and close the sockets to force an exception on the other end.

sahilTakiar:
@kunalkandekar LGTM!

Are there any other changes you want to commit before I merge this?

kunalkandekar:
Hey @sahilTakiar, I'm just running some end-to-end tests once more with some adverse situations (connections being cut unexpectedly, the proxy dying halfway through, etc.) to ensure these are handled cleanly. Will ping as soon as I'm done.

sahilTakiar:
Can you resolve the merge conflicts?

kunalkandekar:
Whoops, hadn't noticed the conflicts, sorry. Merged and pushed.

sahilTakiar:
Do you know what's going on with the Java 8 build on Travis? Let's just disable any tests that currently don't run on Travis using the `disabledOnTravis` tag.

kunalkandekar:
@sahilTakiar I've tried disabling all Tunnel tests by adding them to the "disabledOnTravis" group, but they are apparently still running and the build is still failing. Could you take a look and let me know if I'm doing it wrong?

sahilTakiar:
@kunalkandekar thanks for doing all of this and for addressing all the comments! This feature has now been merged!

kunalkandekar:
@sahilTakiar nice! Thanks!

PR #543, opened 2015-12-18
Fix for HADOOP-12169

kadaan:
Added a fix for a hadoop issue (https://issues.apache.org/jira/browse/HADOOP-12169) which affects the s3a filesystem and results in duplicate files appearing in the results of ListStatus.  In the process, extracted a base class for all FsHelper classes based on the hadoop filesystem.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
@@ -14,10 +14,6 @@ package gobblin.source.extractor.hadoop;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.List;
 
 import org.apache.avro.Schema;
 import org.apache.avro.file.DataFileReader;
@@ -25,8 +21,6 @@ import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.mapred.FsInput;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FileStatus;
-import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.compress.CompressionCodec;
 import org.apache.hadoop.io.compress.CompressionCodecFactory;
@@ -39,57 +33,17 @@ import gobblin.source.extractor.filebased.FileBasedHelperException;
 import gobblin.source.extractor.filebased.SizeAwareFileBasedHelper;
 import gobblin.source.extractor.utils.ProxyFsInput;
 import gobblin.util.HadoopUtils;
-import gobblin.util.ProxiedFileSystemWrapper;
 
-
-public class AvroFsHelper implements SizeAwareFileBasedHelper {
+public class AvroFsHelper extends HadoopFsHelper implements SizeAwareFileBasedHelper {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(AvroFsHelper.class);
-
-  private State state;
-  private final Configuration configuration;
-  private FileSystem fs;
 
   public AvroFsHelper(State state) {
     this(state, HadoopUtils.newConfiguration());
   }
 
   public AvroFsHelper(State state, Configuration configuration) {
-    this.state = state;
-    this.configuration = configuration;
-  }
-
-  public FileSystem getFileSystem() {
-    return this.fs;
-  }
-
-  private void createFileSystem(String uri) throws IOException, InterruptedException, URISyntaxException {
-
-    if (state.getPropAsBoolean(ConfigurationKeys.SHOULD_FS_PROXY_AS_USER,
-        ConfigurationKeys.DEFAULT_SHOULD_FS_PROXY_AS_USER)) {
-      // Initialize file system as a proxy user.
-      this.fs =
-          new ProxiedFileSystemWrapper().getProxiedFileSystem(state, ProxiedFileSystemWrapper.AuthType.TOKEN,
-              state.getProp(ConfigurationKeys.FS_PROXY_AS_USER_TOKEN_FILE), uri);
-
-    } else {
-      // Initialize file system as the current user.
-      this.fs = FileSystem.get(URI.create(uri), this.configuration);
-    }
-  }
-
-  @Override
-  public void connect() throws FileBasedHelperException {
-    URI uri = null;
-    try {
-      this.createFileSystem(state.getProp(ConfigurationKeys.SOURCE_FILEBASED_FS_URI));
-    } catch (IOException e) {
-      throw new FileBasedHelperException("Cannot connect to given URI " + uri + " due to " + e.getMessage(), e);
-    } catch (URISyntaxException e) {
-      throw new FileBasedHelperException("Malformed uri " + uri + " due to " + e.getMessage(), e);
-    } catch (InterruptedException e) {
-      throw new FileBasedHelperException("Interruptted exception is caught when getting the proxy file system", e);
-    }
+    super(state, configuration);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
+package gobblin.source.extractor.hadoop;
```
liyinan926: Missing open source header.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
...
+
+import gobblin.configuration.ConfigurationKeys;
```
liyinan926: Wrong order of imports, see https://github.com/linkedin/gobblin/wiki/CodingStyle.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class HadoopFsHelper implements FileBasedHelper {
```
liyinan926: No Javadoc.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public FileSystem getFileSystem() {
+    return this.fs;
+  }
+
+  @Override
+  public void connect() throws FileBasedHelperException {
+    String uri = state.getProp(ConfigurationKeys.SOURCE_FILEBASED_FS_URI);
+    try {
+      if (uri == null) {
```
liyinan926: Please use `Strings.isNullOrEmpty(uri)` instead.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
...
+  protected HadoopFsHelper(State state, Configuration configuration) {
+    this.state = state;
+    this.configuration = configuration;
+  }
+
+  protected State getState() {
+    return this.state;
+  }
+
+  public FileSystem getFileSystem() {
```
liyinan926: Does this need to be `public`?

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
```diff
@@ -0,0 +1,95 @@
...
+    } catch (URISyntaxException e) {
+      throw new FileBasedHelperException("Malformed uri " + uri + " due to " + e.getMessage(), e);
+    } catch (InterruptedException e) {
+      throw new FileBasedHelperException("Interrupted exception is caught when getting the proxy file system", e);
+    }
+  }
+
+  @Override
+  public List<String> ls(String path) throws FileBasedHelperException {
+      List<String> results = new ArrayList<>();
```
liyinan926: You can use an `ImmutableList.Builder` and return an `ImmutableList`.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -98,37 +52,13 @@ public void close() throws FileBasedHelperException {
...
-      }
-    }
-  }
-
-  @Override
   public InputStream getFileStream(String path) throws FileBasedHelperException {
     try {
       Path p = new Path(path);
-      InputStream in = this.fs.open(p);
+      InputStream in = this.getFileSystem().open(p);
```
liyinan926: Please put this in `try (InputStream in = this.getFileSystem().open(new Path(path)))`.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -139,14 +69,14 @@ public InputStream getFileStream(String path) throws FileBasedHelperException {
   public Schema getAvroSchema(String file) throws FileBasedHelperException {
     DataFileReader<GenericRecord> dfr = null;
     try {
-      if (state.getPropAsBoolean(ConfigurationKeys.SHOULD_FS_PROXY_AS_USER,
+      if (this.getState().getPropAsBoolean(ConfigurationKeys.SHOULD_FS_PROXY_AS_USER,
           ConfigurationKeys.DEFAULT_SHOULD_FS_PROXY_AS_USER)) {
         dfr =
-            new DataFileReader<GenericRecord>(new ProxyFsInput(new Path(file), this.fs),
```
liyinan926: Please use a `Closer` to register `dfr` and `FsInput`/`ProxyFsInput`. And please rename `dfs` to some more meaningful name.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
```diff
@@ -165,16 +95,16 @@ public Schema getAvroSchema(String file) throws FileBasedHelperException {
 
   public DataFileReader<GenericRecord> getAvroFile(String file) throws FileBasedHelperException {
```
liyinan926: Please add Javadoc for this method and make it clear that the caller is responsible for closing the returned `DataFileReader`.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
kadaan: `HadoopFsHelper` needs to be public because people implementing other FsHelper outside of the gobblin project will need access.  `getFileSystem` needs to be public because it is used by `SizeAwareFileBasedHelperDecorator`.

liyinan926: Got it. I meant `getFileFileSystem`.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
kadaan: I would prefer to not make this change at the moment.  This code was pulled directly from `AvroFsHelper.java`.  This brings up the broader question as to when changes like this should be made.  My preference would be to minimize the changes, only fixing obvious issues.  If another PR is needed, to address these issues, then so be it.

liyinan926: Fine.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
kadaan: I don't think this can be put in a try-with-resources statement because the `InputStream` is potentially returned on line 63.

liyinan926: Ah. Didn't realize that. Then please make it clear in the Javadoc that it's the caller's responsibility to close it.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
kadaan: I'm going to pass on this as well.  This is legacy code and a refactoring that I'm not prepared to test right now.

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/HadoopFsHelper.java
kadaan: Added

gobblin-core/src/main/java/gobblin/source/extractor/hadoop/AvroFsHelper.java
liyinan926: Yes, this is legacy code but the way it uses the `Closeable`s is not done right and risks leaking resources, so this must be fixed.

kadaan: This will be done in a different PR.

liyinan926: OK.

liyinan926:
LGTM.

Issue #542, opened 2015-12-18
Cannot publish libraries to mavenLocal


kadaan:
I'm trying to modify gobblin and then test out the changes with an external project that includes additional sources, extractors, etc.  I don't want to publish the libraries to artifactory unless the changes have been accepted back in gobblin trunk.  To deal with this I want to be able to publish the libraries to mavenLocal.  Unfortunately, give then current build.gradle this doesn't seem possible.  I added the following to the build.gradle and was able to publish the libraries to mavenLocal:
```
    install {
      repositories {
        mavenInstaller {
          mavenLocal()
          pom.project {
            name "${project.name}"
            packaging 'jar'
            description 'Gobblin Ingestion Framework'
            url 'https://github.com/linkedin/gobblin/'
          }
        }
      }
    }
```

While this worked fine, there is an issue with the gobblin-rest-api project.  The published library gobblin-rest-api-0.5.0-383-g93635fd contains no code.  It turns out that the code is within a different library `mainDataTemplateJar` which resolves to gobblin-rest-api-data-template-0.5.0-383-g93635fd.jar.  For some reason, when I publish to artifactory the gobblin-rest-api-0.5.0-383-g93635fd.jar contains code.  

I've come up with a work around by adding the following to the gobblin-rest-api build.gradle:
```
artifacts {
    runtime mainDataTemplateJar
    archives mainDataTemplateJar
}

def runtimeJarArtifact
configurations.runtime.artifacts.each {
    if (it.name == "gobblin-rest-api-data-template") {
        it.name = "gobblin-rest-api"
    } else if (it.classifier == "") {
        runtimeJarArtifact = it
    }
}
configurations.runtime.artifacts.remove(runtimeJarArtifact)

def archiveJarArtifact
configurations.archives.artifacts.each {
    if (it.name == "gobblin-rest-api-data-template") {
        it.name = "gobblin-rest-api"
    } else if (it.classifier == "") {
        archiveJarArtifact = it
    }
}
configurations.archives.artifacts.remove(archiveJarArtifact)
```

It feels like this hack should not be needed and I must be doing something wrong.  My gradle-fu is admittedly weak, so any advice on how to get gobblin to properly publish to mavenLocal would be much appreciated.

PR #544, opened 2015-12-18
Support for publishing libraries to mavenLocal

kadaan:
Fix for #542 and #545


chavdar:
Hey, thanks for catching this. It looks like we've missed the following section from the Pegasus doc: https://github.com/linkedin/rest.li/wiki/Gradle-build-integration#publishing-maven-artifacts . It looks similar to yours. What do you think about adding the "official" documentation?

kadaan:
I tried adding that last night and while it meant the data-template jar was uploaded to artifactory, the empty jar was still there.  The layout would then be:

 gobblin-rest-api-VERSION.jar 				261 bytes
 gobblin-rest-api-VERSION-sources.jar 			261 bytes
 gobblin-rest-api-VERSION-javadoc.jar 			261 bytes
 gobblin-rest-api-VERSION-data-template.jar 	41 kilobytes

I’m not confident that this would actually solve the problem.  In addition, it seems misleading that the jar, sources.jar and javadoc.jar are empty.

kadaan:
The build failure should be fixed by #548

chavdar:
@kadaan can you have a look at https://github.com/linkedin/gobblin/pull/550 ? This generates two more artifacts gobblin-rest-api-data-template and gobblin-rest-api-rest-client . gobblin-rest-api.jar is empty because there is no code in src/main/java . The concern that I have about your PR is that if someone adds code under src/main/java , it will be dropped since gobblin-rest-api.jar gets overwritten.

PR #548, opened 2015-12-18
Increase timeout in BoundedBlockingRecordQueueTest

chavdar:
There seems to be an intermittent failure on Travis for this test.

@liyinan926 could you review?


liyinan926:
LGTM.

PR #549, opened 2015-12-19
Migrating KafkaReporter and OutputStreamReporter to MetricReportReporter

sahilTakiar:
A majority of these changes are just refactoring / reformatting related. Very little logic has changed, with a number of exceptions:

* Added a shutdown hook to `RootMetricContext` that closes the `RootMetricContext` and stops all reporters
* Modified the `ScheduledReporter.stopImpl` method to invoke `report` one last time before shutting down
* Remove all uses of TypeSafe's `Config` and replaced it with `Properties`
* Added a new default tag called `metricContextName` which prints the name of the `MetricContext` as one of the `Tag`s
* Modified `GobblinMetrics.buildCustomMetricReporters`
    * There are now two factories for custom reporters: `CustomCodahaleReporterFactory` and `CustomReporterFactory`
    * The reason two different factories are necessary is that only metric reporters were migrated to `gobblin.metrics.report.ScheduledReporter`, while `EventSubmitters` still extend `com.codahale.metrics.ScheduledReporter`
* A number of other changes to `GobblinMetrics` to support using both `gobblin.metrics.report.ScheduledReporter` and `com.codahale.metrics.ScheduledReporter`
* Migrated `KafkaReporter` and `OutputStreamReporter` to extend `MetricReportReporter`
* Still don't know why `KafkaReporterTest` and `KafkaAvroReporterTest` are failing, there seems to be some race condition I haven't been able to pinpoint yet
* A number of changes to how `KafkaAvroReporter`s and `KafkaReporter`s are made - mainly due to the fact that they no longer require a `MetricContext` or a `MetricRegistry` to be created
* `GobblinMetrics`, `RootMetricContext` and `ScheduledReporter` are good classes to start with when reviewing this code

@ibuenros can you review?

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/RootMetricContext.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/RootMetricContext.java
@@ -15,6 +15,7 @@ package gobblin.metrics;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.IOException;
 import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.util.List;
@@ -50,7 +51,6 @@ public class RootMetricContext extends MetricContext {
   private final ReferenceQueue<MetricContext> referenceQueue;
   private final Set<InnerMetricContext> innerMetricContexts;
   private final ScheduledExecutorService referenceQueueExecutorService;
-
   private final Set<ContextAwareReporter> reporters;
 
   private volatile boolean reportingStarted;
@@ -58,13 +58,15 @@ public class RootMetricContext extends MetricContext {
   private RootMetricContext(List<Tag<?>> tags) throws NameConflictException {
     super(ROOT_METRIC_CONTEXT, null, tags, true);
     this.innerMetricContexts = Sets.newConcurrentHashSet();
-    this.referenceQueue = new ReferenceQueue<MetricContext>();
+    this.referenceQueue = new ReferenceQueue<>();
     this.referenceQueueExecutorService = MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(1,
         ExecutorsUtils.newThreadFactory(Optional.of(log), Optional.of("GobblinMetrics-ReferenceQueue"))));
     this.referenceQueueExecutorService.scheduleWithFixedDelay(new CheckReferenceQueue(), 0, 2, TimeUnit.SECONDS);
 
     this.reporters = Sets.newConcurrentHashSet();
     this.reportingStarted = false;
+
+    addShutdownHook();
   }
 
   private static void initialize() {
@@ -72,7 +74,7 @@ public class RootMetricContext extends MetricContext {
       INSTANCE = new RootMetricContext(Lists.<Tag<?>>newArrayList());
     } catch (NameConflictException nce) {
       // Should never happen, as there is no parent, so no conflict.
-      throw new RuntimeException("Failed to generate root metric context. This is an error in the code.");
+      throw new IllegalStateException("Failed to generate root metric context. This is an error in the code.", nce);
     }
   }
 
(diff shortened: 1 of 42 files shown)


gobblin-metrics/src/main/java/gobblin/metrics/context/filter/ContextFilterFactory.java
```diff
@@ -30,30 +29,30 @@
...
    * Modify the configuration to set the {@link ContextFilter} class.
-   * @param config Input {@link Config}.
+   *
+   * @param props the {@link Properties} too add the {@link ContextFilter} specification to
    * @param klazz Class of desired {@link ContextFilter}.
-   * @return Modified {@link Config}.
    */
-  public static Config setContextFilterClass(Config config, Class<? extends ContextFilter> klazz) {
-    return config.withValue(CONTEXT_FILTER_CLASS, ConfigValueFactory.fromAnyRef(klazz.getCanonicalName()));
+  public static void setContextFilterClass(Properties props, Class<? extends ContextFilter> klazz) {
```
ibuenros: I would prefer to leave this class with Config instead of Properties. This class was created for use with metrics reporting config files which need features from Config.

gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
```diff
@@ -69,27 +68,43 @@ public static int parsePeriodToSeconds(String periodStr) {
     }
   }
 
-  public static Config setReportingInterval(Config config, long reportingInterval, TimeUnit reportingIntervalUnit) {
+  public static void setReportingInterval(Properties props, long reportingInterval, TimeUnit reportingIntervalUnit) {
```
ibuenros: This should stay as `Config`.

gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
```diff
@@ -69,27 +68,43 @@ public static int parsePeriodToSeconds(String periodStr) {
...
+  private final double rateFactor;
+  private final double durationFactor;
+
   private final ScheduledExecutorService executor;
   private Optional<ScheduledFuture> scheduledTask;
   private int reportingPeriodSeconds;
 
-  public ScheduledReporter(String name, Config config) {
-    super(name, config);
+  public ScheduledReporter(String name, TimeUnit rateUnit, TimeUnit durationUnit, Properties props) {
```
ibuenros: Why the change in signature?

ibuenros:
@sahilTakiar in general it looks good. I noticed a few changes in API that may not work with the future plans for the library. Let's talk about these plans to make sure any changes made right now are compatible with them.

sahilTakiar:
@ibuenros updated based on comments and offline discussion

* Created a new utility class called `ConfigUtils` with a method `propertiesToConfig` which transforms a `Properties` object to a `Config` object and checks that no two keys are prefixes of each other (turns out the typesafe library already has an easy way of doing the check)
* Restored all usages of the `Config` object
* Moved usage of `rateUnit` and `durationUnit` to `MetricReportReporter` which is where it is specified in the builder
* The constructors for `KafkaReporter` and `OutputStreamReporter` now take in a `Builder` and `Config` object - I decided to leave the `Builder` parameter as removing it would require too much re-factoring and this PR is already long

gobblin-metrics/src/main/java/gobblin/metrics/RootMetricContext.java
```diff
@@ -163,4 +166,22 @@ protected void addMetricContext(MetricContext context) {
...
+   * Add a shutwon hook that first closes the {@link RootMetricContext} and then invokes {@link #stopReporting()}. This
+   * ensures any resources obtained by the {@link RootMetricContext} are released, and that all reporting started on the
+   * {@link RootMetricContext} stops properly.
+   */
+  private void addShutdownHook() {
+    Runtime.getRuntime().addShutdownHook(new Thread() {
+      @Override
+      public void run() {
+        try {
+          close();
```
ibuenros: Closing before calling a method is a weird pattern. Does it need to be done in this order?

sahilTakiar: No, switched.

sahilTakiar:
@ibuenros comments addressed and build is now passing.

ibuenros:
LGTM

PR #551, opened 2015-12-21
Marker apis

chavdar:
@ibuenros @sahilTakiar Can you review?

### DIFF
--- a/gobblin-api/src/main/java/gobblin/annotation/Stable.java
+++ b/gobblin-api/src/main/java/gobblin/annotation/Stable.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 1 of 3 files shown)


gobblin-api/src/main/java/gobblin/annotation/Stable.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Denotes an API that is still stable and should not change.
+ */
+@Documented @Retention(value=RetentionPolicy.CLASS) @Target(value=ElementType.TYPE)
+public @interface Stable {
```
pcadabam: Do we need an explicit @Stable annotation? If some API is not marked alpha or beta is should be stable.

chavdar: Yes, I think we need Stable because none of our APIs are currently annotated and still undergo changes. I think marking something as Stable is a deliberate decision which we have to make on a case-by-case basis.

pcadabam: I see. So we go over existing APIs and mark them as stable where necessary? What is the status of an API that does not have any annotation, stable or changeable?

ibuenros:
LGTM

gobblin-api/src/main/java/gobblin/annotation/Stable.java
chavdar: It should be assumed changeable (Beta) for now unless it is a new feature like gobblin-yarn or gobblin-config-management (Alpha).

pcadabam:
LGTM

PR #552, opened 2015-12-21
Change hive version to 0.14.0

zliu41:
Hive 1.2.1 shadows some of the Hadoop 2.3.0 jars. Change to 0.14.0.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ if (!project.hasProperty('hadoopVersion')) {
 }
 
 if (!project.hasProperty('hiveVersion')) {
-  ext.hiveVersion = '1.2.1'
+  ext.hiveVersion = '0.14.0'
 }
 
 if (!project.hasProperty('pegasusVersion')) {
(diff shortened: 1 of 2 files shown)


chavdar:
The build fails because of a dependency that cannot be found (https://s3.amazonaws.com/archive.travis-ci.org/jobs/98180109/log.txt)


    * What went wrong:
    Could not resolve all dependencies for configuration ':gobblin-yarn:testCompile'.
    > Could not find org.apache.calcite:calcite-core:0.9.2-incubating-SNAPSHOT.
      Required by:
          gobblin:gobblin-yarn:unspecified > gobblin:gobblin-core:unspecified > org.apache.hive:hive-exec:0.14.0
    > Could not find org.apache.calcite:calcite-avatica:0.9.2-incubating-SNAPSHOT.
      Required by:
          gobblin:gobblin-yarn:unspecified > gobblin:gobblin-core:unspecified > org.apache.hive:hive-exec:0.14.0

lbendig:
Hive 0.14 uses datanucleus-core and datanucleus-rdbms version `3.2.10`.
When Gobblin used `4.0.5` I had to downgrade the version to avoid runtime errors in the Hive compactor, like:
`java.lang.NoSuchMethodError: org.datanucleus.util.Localiser.getInstance(Ljava/lang/String;Ljava/lang/ClassLoader;)Lorg/datanucleus/util/Localiser;`
Now `4.1.2` is set, but I guess the downgrading might still be needed.

PR #553, opened 2015-12-21
Config api changes

chavdar:
Changes to the ConfigStore API
* Cleanup javadocs
* Replace references to config key URIs with ConfigKeyPaths
* Add  VersionStabilityPolicy
* Add Alpha annotations

@ibuenros @zliu41 @ydai1124 can you review?


gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/client/api/VersionStabilityPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.config.client.api;
+
+import gobblin.annotation.Alpha;
+import gobblin.config.store.api.ConfigStoreWithStableVersioning;
+
+/**
+ * This version specifies the behavior expected by the client application when making repeated
```
ibuenros: change `version` to `policy`

chavdar: fixed.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigKeyPath.java
```diff
@@ -12,28 +12,45 @@
...
+
  * @author mitu
  *
  */
+@Alpha
 public interface ConfigKeyPath {
 
+  /**
+   * The path to the parent.
+   * @throws IllegalArgumentException if the current node is the root.
```
ibuenros: Hadoop would just throw null for this. I'm ok if we want to avoid NPE, but I don't think it should be an `IllegalArgumentException` (no arguments). What about `UnsupportedOperationException`?

chavdar: Ok, changed to UnsupportedOperationException . I find that NPE most of the time means that someone is not checking for boundary conditions. :-)

ydai1124:
@chavdar Is it necessary to put everything in the `gobblin.config.store.api`, for example, the custom exception classes `VersionDoesNotExistException`?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithStableVersioning.java
```diff
@@ -0,0 +1,46 @@
...
+
+import gobblin.annotation.Alpha;
+import gobblin.config.client.api.VersionStabilityPolicy;
+
+/**
+ * The ConfigStoreWithStableVersion annotation is used to indicate that the configuration store
+ * supports stable versioning. This means that:
+ *
+ * <ul>
+ *   <li>Once published the version will remain available for at a day even if it gets rolled back.
```
ibuenros: `at a day` -> `at least a day` ?

ibuenros:
LGTM other than minor comments.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithStableVersioning.java
ydai1124: Why the default length is "at least a day"? Is it defined somewhere?

chavdar:
@ydai1124 The exception is part of the API so I tihnk it should be there. Any alternatives?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithStableVersioning.java
chavdar: Fixed.

ydai1124:
@chavdar No, if that is part of the API :).

chavdar:
Updated the PR with addressing the code review comments and adding three more factory methods to ConfigKeyPath .

chavdar:
Never mind, the factory methods need to be static and in the actual implementation.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/client/api/VersionStabilityPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+  CROSS_JVM_STABILITY,
+  /** Reading the same config key and version from the same JVMs must return the same result. */
+  STRONG_LOCAL_STABILITY,
+  /**
+   * The application does not depend on getting the same config  for the same key and version but
+   * the client library may use caching to improve performance. This means that the application
+   * may read a stale config if the underlying store does not support stable versioning. */
+  WEAK_LOCAL_STABILITY,
+  /**
+   * The application needs to read the most recent config if the underlying store does not support
```
zliu41: Should there be a policy for applications that always need to read the freshest, regardless of whether the underlying store supports stable versioning or not?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithStableVersioning.java
```diff
@@ -0,0 +1,47 @@
...
+
+import gobblin.annotation.Alpha;
+import gobblin.config.client.api.VersionStabilityPolicy;
+
+/**
+ * The ConfigStoreWithStableVersion annotation is used to indicate that the configuration store
+ * supports stable versioning. This means that:
+ *
+ * <ul>
+ *   <li>Once published the version will remain available for at least a day even if it gets rolled
```
zliu41: Why should it be a day, not 3 days or a week?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/client/api/VersionStabilityPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.config.client.api;
+
+import gobblin.annotation.Alpha;
+import gobblin.config.store.api.ConfigStoreWithStableVersioning;
+
+/**
+ * This policy specifies the behavior expected by the client application when making repeated
+ * calls to the fetch the configuration object for the same config key and version. This interface
```
zliu41: extra "the"

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStore.java
```diff
@@ -32,30 +48,54 @@
...
+   * @param  configKey      the config key path whose tags are needed
+   * @param  version        the configuration version in the configuration store.
+   * @return the paths of the directly imported config keys for the specified config key and
+   * version. Note that order is significant the earlier URI in the List will have higher priority
+   * when resolving configuration conflicts.
+   * @throws VersionDoesNotExistException if the requested config version does not exist (any longer)
+   *
    */
-  public List<URI> getOwnImports(URI uri, String version) throws VersionDoesNotExistException;
+  public List<ConfigKeyPath> getOwnImports(ConfigKeyPath configKey, String version)
```
zliu41: This doesn't need to be a `List`, seems `Collection` is a better fit.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/client/api/VersionStabilityPolicy.java
chavdar: That part will come in a subsequent commit. The config client libraries "remembers" the current version of a given config store on first use. Subsequent calls to the same store will use the same version. The assumption is that applications typically want some predictability/consistency for the options that they read. This is what a given instance of the client library guarantees. If the applications really, really always care about the latest value, then they should instantiate a new copy of the client library every time they need to access a config.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStoreWithStableVersioning.java
chavdar: Good question. Some reasonable bound for what a single job may take to run.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/client/api/VersionStabilityPolicy.java
chavdar: fixed.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigStore.java
chavdar: As noted in the javadoc, the order is actually significant since it governs to which configuration to fallback if it is not in the own/immediate configuration object.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/api/ConfigKeyPath.java
```diff
@@ -12,28 +12,45 @@
...
-   * for example 
-   * configuration store with root URI: etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest
-   * getAbsolutePathString will return /datasets/a1/a2 where datasets is a subdirectory under /user/mitu/HdfsBasedConfigTest
+   * The absolute configuration key path. This is joining all path components from the root using
+   * "/" as a separator.
    */
   public String getAbsolutePathString();
-  
+
+  /** Check if the current path is the root path ("/"). */
```
ydai1124: Should this check for root store path of configStore#getStoreURI(), instead of the absolute root path "/"?

chavdar: Paths passed to config stores have the root URI already removed from them.

ydai1124:
LGTM.

chavdar:
Superseded by the squash commit in https://github.com/linkedin/gobblin/pull/560

PR #555, opened 2015-12-22
Add mySql to gobblin-metastore runtime dependency

ydai1124:
Found the missing dependency when testing `JobExecutionInfoServer`. 
@liyinan926 Can you review?


liyinan926:
LGTM.

PR #557, opened 2015-12-22
Added email notification on Yarn application shutdown

liyinan926:
Also added an option to control the maximum allowed number consecutive number of failures to get the ApplicationReport in case the Yarn cluster is down, e.g., fir maintenance.


liyinan926:

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnConfigurationKeys.java
```diff
@@ -26,6 +26,9 @@
   public static final String APPLICATION_NAME_OPTION_NAME = "app_name";
   public static final String APP_QUEUE_KEY = GOBBLIN_YARN_PREFIX + "app.queue";
   public static final String APP_REPORT_INTERVAL_MINUTES_KEY = GOBBLIN_YARN_PREFIX + "app.report.interval.minutes";
+  public static final String MAX_GET_APP_REPORT_FAILURES_KEY = GOBBLIN_YARN_PREFIX + "max.get.app.report.failures";
+  public static final String EMAIL_NOTIFICATION_ON_SHUTDOWN_KEY =
```
sahilTakiar: Can we have defaults for both these parameters. I say default number of retries is 4, and emails are turned off by default.

liyinan926: Actually the default is specified in `reference.conf` as required by Typesafe config.

sahilTakiar:
LGTM

gobblin-yarn/src/main/java/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -314,6 +344,33 @@ public void handleApplicationReportArrivalEvent(ApplicationReportArrivalEvent ap
...
+      } finally {
+        if (this.emailNotificationOnShutdown) {
+          sendEmailOnShutdown(Optional.of(applicationReport));
+        }
+      }
+    }
+  }
+
+  @Subscribe
+  public void handleGetApplicationReportFailureEvent(
```
zliu41: If this method could be called concurrently, it should probably be synchronized to avoid sending multiple email notifications. If not, it isn't necessary to use AtomicInteger.

liyinan926: No, this method should not be called concurrently. I just feel an `AtomicInteger` is a good choice for a counter although it's not really used for concurrency.

liyinan926:
@zliu41 updated 958b45f.

zliu41:
LGTM

liyinan926:
@zliu41 actually I just realized using a plain integer has one problem: the `getApplicationReportFailureCount` is shared and updated in both event handlers `handleApplicationReportArrivalEvent` and `handleGetApplicationReportFailureEvent`, which may be called by different threads. So for any change to `getApplicationReportFailureCount` in any of the two event handlers to be visible to the other, it at least needs to be `volatile`. So I still think `AtomicInteger` is a good choice here.

PR #559, opened 2015-12-22
Retry failed workunits in FileBasedSource

kadaan:
Fixed FileBasedSource so that it will include previously failed WorkUnits even if there are no new files in the source


liyinan926:
LGTM.

liyinan926:
@sahilTakiar do you want to take a look?

sahilTakiar:
LGTM

PR #565, opened 2015-12-22
Preparation for migration to Gradle 2: add ext. to custom gradle props

chavdar:


kadaan:
Yay!!!!

chavdar:
Merging since this is a mechanical change and build passes.

PR #566, opened 2015-12-22
Adding final tag to the last metric report reported

sahilTakiar:
* Adding a tag named `finalMetricReport` that will be set to true on the final metric report reported, this tag will not be present for all other metric reports
* The tag is enabled when a `ScheduledReporter` is stopped or a `MetricContext` is removed
* Implementation is done by introducing a few new methods in `ScheduledReporter`
* Looking at `ScheduledReporter` is the best place to start for this PR

@ibuenros can you review?

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/reporter/MetricReportReporter.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/reporter/MetricReportReporter.java
@@ -59,7 +59,9 @@ import gobblin.metrics.Tag;
  */
 public abstract class MetricReportReporter extends ScheduledReporter {
 
-  private final static Logger LOGGER = LoggerFactory.getLogger(MetricReportReporter.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(MetricReportReporter.class);
+
+  private static final String FINAL_TAG_KEY = "finalMetricReport";
 
   @Getter
   private final TimeUnit rateUnit;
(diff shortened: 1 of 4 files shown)


ibuenros:
Why not inject the final tag in the `ScheduledReporter`?

sahilTakiar:
There may be other implementations of `ScheduledReporter` that want to do something else on the final report. Plus, the fact that we need to tag the final report is specific to `MetricReportReporter` because it is reporting metrics.

Updated to fix some build errors.

gobblin-metrics/src/main/java/gobblin/metrics/reporter/MetricReportReporter.java
```diff
@@ -172,6 +174,18 @@ public T withTag(String key, String value) {
     }
   }
 
+  @Override
+  protected void report(SortedMap<String, Gauge> gauges, SortedMap<String, Counter> counters,
+      SortedMap<String, Histogram> histograms, SortedMap<String, Meter> meters, SortedMap<String, Timer> timers,
+      Map<String, Object> tags, boolean isFinal) {
+    if (isFinal) {
+      report(gauges, counters, histograms, meters, timers,
+          ImmutableMap.<String, Object>builder().putAll(tags).put(FINAL_TAG_KEY, true).build());
```
liyinan926: You can use `Boolean.TRUE` to avoid the boxing.

liyinan926:
LGTM other than one minor comment.

Issue #561, opened 2015-12-22
BaseDataPublisher marks WorkUnitState as COMMITTED even if the publish fails


kadaan:
**This issue can result in data loss**

The BaseDataPublisher uploads the output files to the publish location and in the process marks the WorkUnitStates as COMMITTED.  Currently the code uploads each output directory using `ParallelRunner.movePath()`.  This fires off a future and returns.  After returning, the WorkUnitState is marked as COMMITTED.  This means that WorkUnitStates are marked as COMMITTED before the file is even uploaded.  Additionally, because the entire folder is being uploaded as a set via the underlying Hadoop call `FileUtil.copy()`, there is no way of tracking which files were uploaded successfully and which were not.

```
public class BaseDataPublisher extends SingleTaskDataPublisher {
  public void publishData(Collection<? extends WorkUnitState> states) throws IOException {
    ...
    for (WorkUnitState workUnitState : states) {
      for (int branchId = 0; branchId < this.numBranches; branchId++) {
        publishMultiTaskData(workUnitState, branchId, writerOutputPathsMoved);
      }
      workUnitState.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
    }
  }

  private void publishMultiTaskData(WorkUnitState state, int branchId, Set<Path> writerOutputPathsMoved) throws IOException {
    publishData(state, branchId, false, writerOutputPathsMoved);
  }

  protected void publishData(WorkUnitState state, int branchId, boolean publishSingleTaskData, Set<Path> writerOutputPathsMoved) throws IOException {
    ...
    parallelRunner.movePath(writerOutputDir, this.publisherFileSystemByBranches.get(branchId), publisherOutputDir, this.publisherFinalDirOwnerGroupsByBranches.get(branchId));
    ...
  }
}

public class ParallelRunner implements Closeable {
  public void movePath(final Path src, final FileSystem dstFs, final Path dst, final Optional<String> group) {
    this.futures.add(this.executor.submit(new Callable<Void>() {
      @Override
      public Void call() throws Exception {
        ...
        HadoopUtils.movePath(fs, src, dstFs, dst);
        ...
      }
    }));
  }
}
```

@zliu41, @sahilTakiar, @liyinan926 Feedback?

zliu41:
@kadaan yes this is definitely a bug. Thanks for pointing out. The fix doesn't seem trivial and I'll seek opinions from other contributors. If you have any suggestions please let us know.

kadaan:
[...]

PR #569, opened 2015-12-23
Exactly-once POC

zliu41:
Design: https://github.com/linkedin/gobblin/wiki/Exactly-Once-Support

Summary of changes:

- Exacty once semantics will be used if `delivery.semantics=EXACTLY_ONCE`, and data are published in the job (and not in the tasks).
- Created `CommitStep`, `CommitSequence`, `CommitSequenceStore` and their implementations.
- When exactly-once is used:
 - `JobContext.commit()` will first construct a `CommitSequence`, persist the `CommitSequence` into the `CommitSequenceStore`, then execute the `CommitSequence`. If the `CommitSequence` is executed successfully, it will be deleted from the `CommitSequenceStore`.
 - `DataPublisher` will not actually publish the data, instead, it will create a `CommitSequence`. This is because the data can only be published after the `CommitSequence` has been fully constructed and persisted.
 - Staging data of a job will only be cleaned if all `CommitSequence`s of the job are successfully completed.
 - When launching a job, `AbstractJobLauncher` will check the `CommitSequenceStore` to see whether there are any unfinished `CommitSequences` for the current job. If there are, it will execute those `CommitSequences`. If successful, it will delete them from the `CommitSequenceStore`, cleanup the staging data, and run the new WorkUnits.

The features below are not included in this PR and will be in subsequent PRs:
- Enable exactly-once when publishing data from tasks.
- Generate and execute `CommitSequence`s in containers rather than the driver.
- Isolation between differnet datasets of a job: if the `CommitSequence` of a dataset cannot be executed, this dataset should be blacklisted and it should not affect other datasets.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
+++ b/gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import java.util.List;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+import gobblin.annotation.Alpha;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * A sequence of {@link CommitStep}s that should be executed atomically.
+ *
+ * <p>
+ *    Typical pattern for building a {@link CommitSequence} with two {@link CommitStep}s:
+ *
+ *    <pre>
+ *      {@code
+ *        CommitSequence sequence = CommitSequence.newBuilder()
+ *                                  .withJobName(jobName)
+ *                                  .withDatasetUrn(datasetUrn)
+ *
+ *                                  .beginStep(FsRenameCommitStep.Builder.class)
+ *                                  .withProps(props)
+ *                                  .from(srcPath)
+ *                                  .to(dstPath)
+ *                                  .endStep()
+ *
+ *                                  .beginStep(DatasetStateCommitStep.Builder.class)
+ *                                  .withProps(props)
+ *                                  .withDatasetUrn(datasetUrn)
+ *                                  .withDatasetState(datasetState)
+ *                                  .endStep()
+ *
+ *                                  .build();
+ *      }
+ *    </pre>
+ * </p>
+ *
+ * @author ziliu
+ */
+@Alpha
+@Slf4j
+public class CommitSequence {
+
+  @Getter
+  private final String jobName;
+
+  @Getter
+  private final String datasetUrn;
+
+  private final List<CommitStep> steps;
+
+  private CommitSequence(Builder builder) {
+    this.jobName = builder.jobName;
+    this.datasetUrn = builder.datasetUrn;
+    this.steps = ImmutableList.copyOf(builder.steps);
+  }
+
+  public static class Builder {
+
+    private String jobName;
+    private String datasetUrn;
+    private final List<CommitStep> steps = Lists.newArrayList();
+
+    /**
+     * Set the job name for the commit sequence.
+     */
+    public Builder withJobName(String jobName) {
+      this.jobName = jobName;
+      return this;
+    }
+
+    /**
+     * Set the dataset URN for the commit sequence.
+     */
+    public Builder withDatasetUrn(String datasetUrn) {
+      this.datasetUrn = datasetUrn;
+      return this;
+    }
+
+    /**
+     * Build a {@link CommitStep}.
+     *
+     * @param builderClass The builder class for the {@link CommitStep}, which should extend
+     * {@link CommitStepBase.Builder}.
+     * @return An instance of the builder class for the {@link CommitStep}.
+     */
+    public <T extends CommitStepBase.Builder<?>> T beginStep(Class<T> builderClass) {
+      try {
+        //return (T) ConstructorUtils.invokeConstructor(builderClass, this);
+        return (T) builderClass.getDeclaredConstructor(this.getClass()).newInstance(this);
+      } catch (ReflectiveOperationException e) {
+        throw new RuntimeException("Failed to instantiate " + builderClass, e);
+      }
+    }
+
+    /**
+     * Add a {@link CommitStep} to the commit sequence.
+     */
+    public Builder addStep(CommitStep step) {
+      this.steps.add(step);
+      return this;
+    }
+
+    public CommitSequence build() {
+      Preconditions.checkState(!Strings.isNullOrEmpty(this.jobName), "Job name not specified for commit sequence");
+      Preconditions.checkState(!Strings.isNullOrEmpty(this.datasetUrn),
+          "Dataset URN not specified for commit sequence");
+      Preconditions.checkState(!this.steps.isEmpty(), "No commit steps specified for the commit sequence");
+
+      return new CommitSequence(this);
+    }
+
+  }
+
+  /**
+   * Execute the {@link CommitStep}s in the order they are added to the commit sequence.
+   */
+  public void execute() {
+    try {
+      for (CommitStep step : this.steps) {
+        if (!step.isCompleted()) {
+          step.execute();
+        }
+      }
+    } catch (Throwable t) {
+      log.error("Commit failed for dataset " + this.datasetUrn, t);
+      throw Throwables.propagate(t);
+    }
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+}
--- a/gobblin-api/src/main/java/gobblin/commit/CommitSequenceStore.java
+++ b/gobblin-api/src/main/java/gobblin/commit/CommitSequenceStore.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import java.io.IOException;
+import java.util.Collection;
+
+import com.google.common.base.Optional;
+
+import gobblin.annotation.Alpha;
+
+
+/**
+ * A store for {@link CommitSequence}s. A {@link CommitSequence} is identified by job name and dataset URN.
+ *
+ * @author ziliu
+ */
+@Alpha
+public interface CommitSequenceStore {
+
+  /**
+   * Whether a {@link CommitSequence} with the given job name exists in the store.
+   */
+  boolean exists(String jobName) throws IOException;
+
+  /**
+   * Whether a {@link CommitSequence} with the given job name and dataset URN exists in a store.
+   */
+  public boolean exists(String jobName, String datasetUrn) throws IOException;
+
+  /**
+   * delete a given job name from the store along with all {@link CommitSequence}s associated with this job.
+   */
+  public void delete(String jobName) throws IOException;
+
+  /**
+   * delete the {@link CommitSequence} for the given job name and dataset URN.
+   */
+  public void delete(String jobName, String datasetUrn) throws IOException;
+
+  /**
+   * Put a {@link CommitSequence} with the given job name and dataset URN.
+   *
+   * @throws IOException if a {@link CommitSequence} for the given job name and dataset URN exists in the store.
+   */
+  public void put(String jobName, String datasetUrn, CommitSequence commitSequence) throws IOException;
+
+  /**
+   * Get a {@link Collection} of dataset URNs with the given job name.
+   */
+  public Collection<String> get(String jobName) throws IOException;
+
+  /**
+   * Get the {@link CommitSequence} associated with the given job name and dataset URN.
+   */
+  public Optional<CommitSequence> get(String jobName, String datasetUrn) throws IOException;
+
+}
--- a/gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
+++ b/gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import java.io.IOException;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+
+import gobblin.annotation.Alpha;
+import gobblin.configuration.State;
+
+
+/**
+ * A base implementation of {@link CommitStep}.
+ *
+ * @author ziliu
+ */
+@Alpha
+public abstract class CommitStepBase implements CommitStep {
+
+  protected final State props;
+
+  protected CommitStepBase(Builder<? extends Builder<?>> builder) {
+    Preconditions.checkNotNull(builder.props);
+    this.props = builder.props;
+  }
+
+  public abstract static class Builder<T extends Builder<?>> {
+    private final Optional<CommitSequence.Builder> commitSequenceBuilder;
+
+    protected State props;
+
+    protected Builder() {
+      this.commitSequenceBuilder = Optional.<CommitSequence.Builder> absent();
+    }
+
+    protected Builder(CommitSequence.Builder commitSequenceBuilder) {
+      Preconditions.checkNotNull(commitSequenceBuilder);
+      this.commitSequenceBuilder = Optional.of(commitSequenceBuilder);
+    }
+
+    @SuppressWarnings("unchecked")
+    public T withProps(State props) {
+      this.props = new State(props.getProperties());
+      return (T) this;
+    }
+
+    public CommitSequence.Builder endStep() throws IOException {
+      Preconditions.checkState(this.commitSequenceBuilder.isPresent());
+      this.commitSequenceBuilder.get().addStep(build());
+      return commitSequenceBuilder.get();
+    }
+
+    public abstract CommitStep build() throws IOException;
+  }
+
+}
--- a/gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
+++ b/gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import com.google.common.base.Enums;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+
+
+/**
+ * The semantics for data delivery.
+ *
+ * @author ziliu
+ */
+public enum DeliverySemantics {
+
+  /**
+   * Each data record from the source is guaranteed to be delivered at least once.
+   */
+  AT_LEAST_ONCE,
+
+  /**
+   * Each data record from the source is guaranteed to be delievered exactly once.
+   */
+  EXACTLY_ONCE;
+
+  /**
+   * Get the devliery semantics type from {@link ConfigurationKeys#DELIVERY_SEMANTICS}.
+   * The default value is {@link Type#AT_LEAST_ONCE}.
+   */
+  public static DeliverySemantics parse(State state) {
+    String value =
+        state.getProp(ConfigurationKeys.GOBBLIN_RUNTIME_DELIVERY_SEMANTICS, AT_LEAST_ONCE.toString()).toUpperCase();
+    Optional<DeliverySemantics> semantics = Enums.getIfPresent(DeliverySemantics.class, value);
+    Preconditions.checkState(semantics.isPresent(), value + " is not a valid delivery semantics");
+    return semantics.get();
+  }
+}
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -403,7 +403,6 @@ public class ConfigurationKeys {
   public static final String TASK_STATE_COLLECTOR_INTERVAL_SECONDS = "task.state.collector.interval.secs";
   public static final int DEFAULT_TASK_STATE_COLLECTOR_INTERVAL_SECONDS = 60;
 
-
   /**
    * Configuration properties for email settings.
    */
@@ -507,6 +506,7 @@ public class ConfigurationKeys {
   /**
    * Other configuration properties.
    */
+  public static final String GOBBLIN_RUNTIME_DELIVERY_SEMANTICS = "gobblin.runtime.delivery.semantics";
   public static final Charset DEFAULT_CHARSET_ENCODING = Charsets.UTF_8;
   public static final String TEST_HARNESS_LAUNCHER_IMPL = "gobblin.testharness.launcher.impl";
   public static final int PERMISSION_PARSING_RADIX = 8;
--- a/gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
+++ b/gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import java.io.IOException;
+import java.net.URI;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.base.Preconditions;
+
+import gobblin.annotation.Alpha;
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+import gobblin.util.HadoopUtils;
+
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * A {@link CommitStep} for renaming files within a {@link FileSystem} or between {@link FileSystem}s.
+ *
+ * @author ziliu
+ */
+@Alpha
+@Slf4j
+public class FsRenameCommitStep extends CommitStepBase {
+
+  private final Path srcPath;
+  private final Path dstPath;
+  private final String srcFsUri;
+  private final String dstFsUri;
+  private transient FileSystem srcFs;
+  private transient FileSystem dstFs;
+
+  private FsRenameCommitStep(Builder<? extends Builder<?>> builder) throws IOException {
+    super(builder);
+
+    this.srcPath = builder.srcPath;
+    this.dstPath = builder.dstPath;
+    this.srcFs = builder.srcFs != null ? builder.srcFs
+        : getFileSystem(this.props.getProp(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
+    this.srcFsUri = this.srcFs.getUri().toString();
+    this.dstFs = builder.dstFs != null ? builder.dstFs
+        : getFileSystem(this.props.getProp(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
+    this.dstFsUri = this.dstFs.getUri().toString();
+  }
+
+  public static class Builder<T extends Builder<?>> extends CommitStepBase.Builder<T> {
+
+    public Builder() {
+      super();
+    }
+
+    public Builder(CommitSequence.Builder commitSequenceBuilder) {
+      super(commitSequenceBuilder);
+    }
+
+    private Path srcPath;
+    private Path dstPath;
+    private FileSystem srcFs;
+    private FileSystem dstFs;
+
+    @Override
+    public T withProps(State props) {
+      return (T) super.withProps(props);
+    }
+
+    @SuppressWarnings("unchecked")
+    public T from(Path srcPath) {
+      this.srcPath = srcPath;
+      return (T) this;
+    }
+
+    @SuppressWarnings("unchecked")
+    public T to(Path dstPath) {
+      this.dstPath = dstPath;
+      return (T) this;
+    }
+
+    @SuppressWarnings("unchecked")
+    public T withSrcFs(FileSystem srcFs) {
+      this.srcFs = srcFs;
+      return (T) this;
+    }
+
+    @SuppressWarnings("unchecked")
+    public T withDstFs(FileSystem dstFs) {
+      this.dstFs = dstFs;
+      return (T) this;
+    }
+
+    @Override
+    public CommitStep build() throws IOException {
+      Preconditions.checkNotNull(this.srcPath);
+      Preconditions.checkNotNull(this.dstPath);
+
+      return new FsRenameCommitStep(this);
+    }
+  }
+
+  private FileSystem getFileSystem(String fsUri) throws IOException {
+    return FileSystem.get(URI.create(fsUri), HadoopUtils.getConfFromState(this.props));
+  }
+
+  @Override
+  public boolean isCompleted() throws IOException {
+    if (this.dstFs == null) {
+      this.dstFs = getFileSystem(this.dstFsUri);
+    }
+    return this.dstFs.exists(this.dstPath);
+  }
+
+  @Override
+  public void execute() throws IOException {
+    if (this.srcFs == null) {
+      this.srcFs = getFileSystem(this.srcFsUri);
+    }
+    if (this.dstFs == null) {
+      this.dstFs = getFileSystem(this.dstFsUri);
+    }
+    log.info(String.format("Moving %s to %s", this.srcPath, this.dstPath));
+    HadoopUtils.movePath(this.srcFs, this.srcPath, this.dstFs, this.dstPath);
+  }
+}
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -34,6 +34,9 @@ import com.google.common.collect.Maps;
 import com.google.common.eventbus.EventBus;
 import com.google.common.io.Closer;
 
+import gobblin.commit.CommitSequence;
+import gobblin.commit.CommitSequenceStore;
+import gobblin.commit.DeliverySemantics;
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
 import gobblin.metastore.StateStore;
@@ -204,6 +207,13 @@ public abstract class AbstractJobLauncher implements JobLauncher {
             "Previous instance of job %s is still running, skipping this scheduled run", this.jobContext.getJobName()));
       }
 
+      if (this.jobContext.getSemantics() == DeliverySemantics.EXACTLY_ONCE) {
+
+        // If exactly-once is used, commit sequences of the previous run must be successfully compelted
+        // before this run can make progress.
+        executeUnfinishedCommitSequences(jobState.getJobName());
+      }
+
       TimingEvent workUnitsCreationTimer =
           this.eventSubmitter.getTimingEvent(TimingEventNames.LauncherTimings.WORK_UNITS_CREATION);
       // Generate work units of the job from the source
@@ -308,6 +318,19 @@ public abstract class AbstractJobLauncher implements JobLauncher {
 
     if (jobState.getState() == JobState.RunningState.FAILED) {
       throw new JobException(String.format("Job %s failed", jobId));
+    }
+  }
+
+  private void executeUnfinishedCommitSequences(String jobName) throws IOException {
+    Preconditions.checkState(this.jobContext.getCommitSequenceStore().isPresent());
+    CommitSequenceStore commitSequenceStore = this.jobContext.getCommitSequenceStore().get();
+
+    for (String datasetUrn : commitSequenceStore.get(jobName)) {
+      Optional<CommitSequence> commitSequence = commitSequenceStore.get(jobName, datasetUrn);
+      if (commitSequence.isPresent()) {
+        commitSequence.get().execute();
+      }
+      commitSequenceStore.delete(jobName, datasetUrn);
     }
   }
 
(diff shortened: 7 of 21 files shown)


zliu41:
@liyinan926 @chavdar @ydai1124 can you review?

kadaan:
@zliu41 Can we discuss #561 before moving forward with this?  I believe that the result of fixing #561 will change how this needs to be implemented.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+ * <p>
+ *    Typical pattern for building a {@link CommitSequence} with two {@link CommitStep}s:
+ *
+ *    <pre>
+ *      {@code
+ *        CommitSequence sequence = CommitSequence.newBuilder()
+ *                                  .withJobName(jobName)
+ *                                  .withDatasetUrn(datasetUrn)
+ *
+ *                                  .buildStep(FsRenameCommitStep.Builder.class)
```
liyinan926: I would suggest rename this to `beginStep` and `stepDone` to `endStep`.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+ *                                  .build();
+ *      }
+ *    </pre>
+ * </p>
+ *
+ * @author ziliu
+ */
+public class CommitSequence {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CommitSequence.class);
```
liyinan926: Seems you are using `@Slf4j` for some classes. Please make sure the use is consistent.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @Getter
+  private final String datasetUrn;
+
+  private final List<CommitStep> steps;
+
+  private CommitSequence(Builder builder) {
+    this.jobName = builder.jobName;
+    this.datasetUrn = builder.datasetUrn;
+    this.steps = builder.steps;
```
liyinan926: It's better to make a copy using a `ImmutableList` here.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public static class Builder {
+
+    private String jobName;
+    private String datasetUrn;
+    private final List<CommitStep> steps = Lists.newArrayList();
+
+    public Builder withJobName(String jobName) {
```
liyinan926: Missing Javadoc.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+   * Execute the {@link CommitStep}s in the order they are added to the commit sequence.
+   */
+  public void execute() {
+    try {
+      for (CommitStep step : this.steps) {
+        if (!step.isCompleted()) {
+          step.execute();
+        }
+      }
+    } catch (Throwable t) {
```
liyinan926: This means if any `CommitStep` fails, the `execute` fails. Should the try-catch block surround `step.execute` instead?

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+      LOG.error("Commit failed for dataset " + this.datasetUrn, t);
+    }
+  }
+
+
+  /**
+   * Append a {@link CommitStep}.
+   */
+  public void append(CommitStep step) {
```
liyinan926: Why it allows `append`? I think `CommitSequence` should be immutable once built and the `Builder` class is the only class that can modify the state.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,153 @@
...
+    this.steps.add(step);
+  }
+
+  /**
+   * Append all {@link CommitStep}s in another commit sequence.
+   */
+  public void append(CommitSequence other) {
```
liyinan926: Ditto.

gobblin-api/src/main/java/gobblin/commit/CommitSequenceStore.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public interface CommitSequenceStore {
+
+  /**
+   * Whether a job name exists in the store.
+   */
+  boolean exists(String jobName) throws IOException;
+
+  /**
+   * Whether a job name and dataset URN exists in a store.
```
liyinan926: Incomplete Javadocs.

gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
```diff
@@ -0,0 +1,64 @@
...
+    private final Optional<CommitSequence.Builder> commitSequenceBuilder;
+
+    protected State props;
+
+    protected Builder() {
+      this.commitSequenceBuilder = Optional.<CommitSequence.Builder> absent();
+    }
+
+    protected Builder(CommitSequence.Builder commitSequenceBuilder) {
+      this.commitSequenceBuilder = Optional.of(commitSequenceBuilder);
```
liyinan926: Unless you add a `Preconditions.checkNotNull`, then please use `fromNullable`.

gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public T withProps(State props) {
+      this.props = props;
```
liyinan926: Better to make a copy here.

gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
```diff
@@ -0,0 +1,55 @@
...
+import gobblin.configuration.State;
+
+
+/**
+ * The semantics for data delivery.
+ *
+ * @author ziliu
+ */
+public final class DeliverySemantics {
+  public enum Type {
```
liyinan926: Please add an empty line before `public enum Type`.

gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
```diff
@@ -0,0 +1,55 @@
...
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+
+
+/**
+ * The semantics for data delivery.
+ *
+ * @author ziliu
+ */
+public final class DeliverySemantics {
```
liyinan926: Ah, you can actually make this class an `enum` and have a static method named `parse(State state)`.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -18,19 +18,39 @@
...
+ * If exactly-once semantics is used, it will construct a {@link CommitSequence} instead of actually publishing
+ * the data. The data will be published when the {@link CommitSequence} is executed, which will be after the
+ * {@link CommitSequence} has been fully constructed and persisted into the {@link gobblin.commit.CommitSequenceStore}.
  */
 public abstract class DataPublisher implements Closeable {
 
   protected final State state;
 
+  protected final DeliverySemantics.Type deliverySemantics;
+
```
liyinan926: This can be final.

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -18,19 +18,39 @@
...
+  protected CommitSequence.Builder commitSequenceBuilder;
+
+  @Getter
+  protected CommitSequence commitSequence = null;
```
liyinan926: Can this be an `Optional`?

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,123 @@
...
+  private final String dstFsUri;
+  private transient FileSystem srcFs;
+  private transient FileSystem dstFs;
+
+  private FsRenameCommitStep(Builder builder) throws IOException {
+    super(builder);
+    this.srcPath = builder.srcPath;
+    this.dstPath = builder.dstPath;
+    this.srcFs = builder.srcFs != null ? builder.srcFs
+        : getFileSystem(Optional.fromNullable(this.props.getProp(ConfigurationKeys.FS_URI_KEY)));
```
liyinan926: We use `ConfigurationKeys.LOCAL_FS_URI` as the default in other places.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,123 @@
...
+    this.srcFsUri = this.srcFs.getUri().toString();
+    this.dstFs = builder.dstFs != null ? builder.dstFs
+        : getFileSystem(Optional.fromNullable(this.props.getProp(ConfigurationKeys.FS_URI_KEY)));
+    this.dstFsUri = this.dstFs.getUri().toString();
+  }
+
+  public static class Builder extends CommitStepBase.Builder<Builder> {
+
+    public Builder() {
```
liyinan926: You don't need to make this constructor explicit.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,123 @@
...
+    public Builder(CommitSequence.Builder commitSequenceBuilder) {
+      super(commitSequenceBuilder);
+    }
+
+    private Path srcPath;
+    private Path dstPath;
+    private FileSystem srcFs;
+    private FileSystem dstFs;
+
+    public Builder from(Path srcPath) {
```
liyinan926: Please add `Preconditions.checkNotNull`.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,123 @@
...
+      this.srcPath = srcPath;
+      return this;
+    }
+
+    public Builder to(Path dstPath) {
+      this.dstPath = dstPath;
+      return this;
+    }
+
+    public Builder withSrcFs(FileSystem srcFs) {
```
liyinan926: Ditto. Please add a precondition check.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,123 @@
...
+    if (fsUri.isPresent()) {
+      return FileSystem.get(URI.create(fsUri.get()), HadoopUtils.getConfFromState(this.props));
+    } else {
+      return FileSystem.get(HadoopUtils.getConfFromState(this.props));
+    }
+  }
+
+  @Override
+  public boolean isCompleted() throws IOException {
+    if (this.dstFs == null) {
```
liyinan926: Why you need this `if` block?

zliu41: `FileSystem` cannot be serialized and deserialized by Gson. So when a `FsRenameCommitStep` is obtained from a Json String, the `FileSystem` will be null.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
zliu41: Different parts of a `CommitSequence` are built in different places, e.g., the part for renaming files and the part for persisting states are built in different classes, and they need to be concatenated.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
zliu41: Since the `CommitStep`s need to be executed in a particular order, once a step fails, no further steps should be executed.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -332,7 +373,14 @@ void commit() throws IOException {
         allDatasetsCommit = false;
       } finally {
         try {
-          persistDatasetState(datasetUrn, datasetState);
+          if (commitSequence.isPresent()) {
+            commitSequence.get().append(buildDatasetStateCommitStep(datasetUrn, datasetState).get());
```
liyinan926: Seems this is the only use case of `append`, and it seems `execute` only happens here after. So I think you can keep using the builder here instead of appending to a `CommitSequence`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -198,6 +200,10 @@ public void launchJob(JobListener jobListener) throws JobException {
             "Previous instance of job %s is still running, skipping this scheduled run", this.jobContext.getJobName()));
       }
 
+      if (this.jobContext.getSemanticsType() == DeliverySemantics.Type.EXACTLY_ONCE) {
+        executeUnfinishedCommitSequences(jobState.getJobName());
```
liyinan926: If `executeUnfinishedCommitSequences` throws an exception, the rest of the job flow will be aborted and if the exception is persistent, the job will never make progress.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -619,6 +591,81 @@ public static void runWorkUnits(String jobId, String containerId, List<WorkUnit>
...
+  private void cleanLeftoverStagingData(List<WorkUnit> workUnits, JobState jobState) throws JobException {
+
+    try {
+      if (this.jobContext.getSemanticsType() == DeliverySemantics.Type.EXACTLY_ONCE
+          && this.jobContext.getCommitSequenceStore().get().exists(jobState.getJobName())) {
+        LOG.error("Job " + jobState.getJobName() + " has unfinished commit sequences. Will not clean up staging data.");
+        return;
+      }
+    } catch (IOException e) {
+      throw new JobException("Failed to clean up staging data", e);
```
liyinan926: The log message should be something like `Failed to check unfinished commit sequences`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -619,6 +591,81 @@ public static void runWorkUnits(String jobId, String containerId, List<WorkUnit>
...
+  private void cleanupStagingData(JobState jobState) throws JobException {
+
+    try {
+      if (this.jobContext.getSemanticsType() == DeliverySemantics.Type.EXACTLY_ONCE
+          && this.jobContext.getCommitSequenceStore().get().exists(jobState.getJobName())) {
+        LOG.error("Job " + jobState.getJobName() + " has unfinished commit sequences. Will not clean up staging data.");
+        return;
+      }
+    } catch (IOException e) {
+      throw new JobException("Failed to clean up staging data", e);
```
liyinan926: Ditto.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -256,7 +294,7 @@ boolean shouldCleanupStagingDataPerTask() {
    * @return a {@link Map} from dataset URNs to {@link JobState.DatasetState}s representing the dataset states
    */
   Map<String, JobState.DatasetState> getDatasetStatesByUrns() {
-    return ImmutableMap.copyOf(this.datasetStatesByUrns.or(Maps.<String, JobState.DatasetState>newHashMap()));
+    return ImmutableMap.copyOf(this.datasetStatesByUrns.or(Maps.<String, JobState.DatasetState> newHashMap()));
```
liyinan926: There should be an empty after `>`. I see this in many places.

gobblin-runtime/src/main/java/gobblin/runtime/commit/DatasetStateCommitStep.java
```diff
@@ -0,0 +1,110 @@
...
+    public Builder() {
+      super();
+    }
+
+    public Builder(CommitSequence.Builder commitSequenceBuilder) {
+      super(commitSequenceBuilder);
+    }
+
+    public Builder withDatasetUrn(String datasetUrn) {
+      this.datasetUrn = datasetUrn;
```
liyinan926: Ditto. Please add some precondition check.

gobblin-runtime/src/main/java/gobblin/runtime/commit/DatasetStateCommitStep.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void execute() throws IOException {
+    log.info("Persisting dataset state for dataset " + this.datasetUrn);
+    getDatasetStateStore().persistDatasetState(this.datasetUrn, this.datasetState);
+  }
+
+  private FsDatasetStateStore getDatasetStateStore() throws IOException {
+    if (this.stateStore == null) {
+      Preconditions.checkArgument(this.props.contains(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY),
```
liyinan926: Again, `ConfigurationKeys.LOCAL_FS_URI` is used as default for `STATE_STORE_ROOT_DIR_KEY` in other places. So please make it consistent.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
zliu41: To meet the requirement of exactly-once, if the previous run's commit sequence failed to complete, the next run should not make further progress.

gobblin-utility/src/main/java/gobblin/util/filters/HiddenFilter.java
```diff
@@ -0,0 +1,29 @@
...
+
+
+/**
+ * A {@link PathFilter} that filters out hidden files (those starting with '_' or '.').
+ */
+public class HiddenFilter implements PathFilter {
+
+  @Override
+  public boolean accept(Path path) {
+    return !(path.getName().startsWith("_") || path.getName().startsWith("."));
```
liyinan926: I would suggest putting the offending characters into an `ImmutableSet`, like https://github.com/dnuttle/hadoop-lab/blob/master/src/main/java/net/nuttle/hadoop/HiddenPathFilter.java.

gobblin-utility/src/main/java/gobblin/util/io/GsonInterfaceAdapter.java
```diff
@@ -0,0 +1,89 @@
...
+ * @author ziliu
+ *
+ * @param <T> The interface or abstract type to be serialized and deserialized with {@link Gson}.
+ */
+public class GsonInterfaceAdapter<T> implements JsonSerializer<T>, JsonDeserializer<T> {
+
+  private static final String OBJECT_TYPE = "object-type";
+  private static final String OBJECT_DATA = "object-data";
+
+  public JsonElement serialize(T object, Type interfaceType, JsonSerializationContext context) {
```
liyinan926: Missing `@Override` annotation for this and `deserialize`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -306,6 +312,18 @@ public void launchJob(JobListener jobListener) throws JobException {
...
 
+  private void executeUnfinishedCommitSequences(String jobName) throws IOException {
+    Preconditions.checkState(this.jobContext.getCommitSequenceStore().isPresent());
+
+    for (String datasetUrn : this.jobContext.getCommitSequenceStore().get().get(jobName)) {
+      Optional<CommitSequence> commitSequence = this.jobContext.getCommitSequenceStore().get().get(jobName, datasetUrn);
+      if (commitSequence.isPresent()) {
+        commitSequence.get().execute();
+      }
+      this.jobContext.getCommitSequenceStore().get().delete(jobName, datasetUrn);
```
liyinan926: Seems you can have a local variable for `this.jobContext.getCommitSequenceStore()`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
liyinan926: Then please add some comment here documenting the behavior.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -619,6 +591,81 @@ public static void runWorkUnits(String jobId, String containerId, List<WorkUnit>
...
+   *
+   * Property {@link ConfigurationKeys#CLEANUP_STAGING_DATA_PER_TASK} controls whether to cleanup
+   * staging data per task, or to cleanup entire job's staging data at once.
+   *
+   * Staging data will not be cleaned if the job has unfinished {@link CommitSequence}s.
+   */
+  private void cleanLeftoverStagingData(List<WorkUnit> workUnits, JobState jobState) throws JobException {
+
+    try {
+      if (this.jobContext.getSemanticsType() == DeliverySemantics.Type.EXACTLY_ONCE
```
liyinan926: Seems the if condition deserves a private method since it appears twice.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -397,24 +445,28 @@ private void finalizeDatasetStateBeforeCommit(JobState.DatasetState datasetState
...
 
   /**
    * Commit the output data of a dataset.
    */
   @SuppressWarnings("unchecked")
-  private void commitDataset(JobState.DatasetState datasetState) throws IOException {
+  private Optional<CommitSequence> commitDatasetOrGenerateDataCommitSequence(JobState.DatasetState datasetState)
+      throws IOException {
     Closer closer = Closer.create();
+    Optional<CommitSequence> commitSequence = Optional.<CommitSequence> absent();
```
liyinan926: Ditto. No empty space after `>` please.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -423,12 +475,28 @@ private void commitDataset(JobState.DatasetState datasetState) throws IOExceptio
...
   private void persistDatasetState(String datasetUrn, JobState.DatasetState datasetState) throws IOException {
+    finalizeDatasetState(datasetState, datasetUrn);
+    LOG.info("Persisting dataset state for dataset " + datasetUrn);
+    this.datasetStateStore.persistDatasetState(datasetUrn, datasetState);
+  }
+
+  private Optional<CommitStep> buildDatasetStateCommitStep(String datasetUrn, JobState.DatasetState datasetState)
+      throws IOException {
+
+    finalizeDatasetState(datasetState, datasetUrn);
```
liyinan926: Seems you can move `finalizeDatasetState(datasetState, datasetUrn)` out from both `persistDatasetState` and this method into `commit`.

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
```diff
@@ -0,0 +1,141 @@
...
+
+    Path datasetPath = new Path(jobPath, datasetUrn);
+    try (DataOutputStream dos = this.fs.create(datasetPath)) {
+      dos.writeBytes(GSON.toJson(commitSequence));
+    }
+  }
+
+  @Override
+  public Collection<String> get(String jobName) throws IOException {
+    List<String> datasetUrns = Lists.newArrayList();
```
liyinan926: You can return an `ImmutableList` instead.

gobblin-runtime/src/test/java/gobblin/runtime/commit/FsCommitSequenceStoreTest.java
```diff
@@ -0,0 +1,99 @@
...
+import gobblin.runtime.commit.FsCommitSequenceStore;
+import gobblin.util.io.GsonInterfaceAdapter;
+
+
+/**
+ * Tests for {@link FsCommitSequenceStore}.
+ *
+ * @author ziliu
+ */
+public class FsCommitSequenceStoreTest {
```
liyinan926: Seems you don't have unit tests for `CommitSequence`?

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
zliu41: Added preconditions check in constructor.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
zliu41: This is automatically added by Eclipse, I don't know a way to disable it.

gobblin-runtime/src/main/java/gobblin/runtime/commit/DatasetStateCommitStep.java
zliu41: Added in constructor

gobblin-runtime/src/test/java/gobblin/runtime/commit/FsCommitSequenceStoreTest.java
zliu41: Added `CommitSequenceTest`

zliu41:
Updated

liyinan926:

kadaan:
@liyinan926 Can you look at #572 as well?  It is related and I believe it needs to be addressed before this POC is completed.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,140 @@
...
+     * Build a {@link CommitStep}.
+     *
+     * @param builderClass The builder class for the {@link CommitStep}, which should extend
+     * {@link CommitStepBase.Builder}.
+     * @return An instance of the builder class for the {@link CommitStep}.
+     */
+    public <T extends CommitStepBase.Builder<?>> T beginStep(Class<T> builderClass) {
+      try {
+        return ConstructorUtils.invokeConstructor(builderClass, this);
+      } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
```
ibuenros: nit: You can catch `ReflectiveOperationException` instead.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,124 @@
...
+    private Path dstPath;
+    private FileSystem srcFs;
+    private FileSystem dstFs;
+
+    public Builder from(Path srcPath) {
+      this.srcPath = srcPath;
+      return this;
+    }
+
+    public Builder to(Path dstPath) {
```
ibuenros: FYI: this kind of pattern will produce compile errors depending on the order on which builder methods are called. For example `builder.to(path).withprops(props)` will compile, but `builder.withProps(props).to(path)` will not compile. This is very frustrating to debug.

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
```diff
@@ -0,0 +1,142 @@
...
+    if (exists(jobName, datasetUrn)) {
+      throw new IOException(String.format("CommitSequence already exists for job %s, dataset %s", jobName, datasetUrn));
+    }
+
+    Path jobPath = new Path(this.rootPath, jobName);
+    this.fs.mkdirs(jobPath);
+
+    Path datasetPath = new Path(jobPath, datasetUrn);
+    try (DataOutputStream dos = this.fs.create(datasetPath)) {
+      dos.writeBytes(GSON.toJson(commitSequence));
```
ibuenros: Has this been tested extensively? There are certain classes that are not easily serializable (e.g. FileSystem). The serialization will probably work fine, but deserialization will fail, as some fields in FileSystem are transient. Gson has something called an instance creator, but they might be needed for any transient fields.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
zliu41: Why won't `builder.withProps(props).to(path)` compile? Seems to me it should.

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
zliu41: Yes, both existing `CommitStep` implementations are serializable. Non-serializable fields are marked `transient` and are instantiated from the state during deserialization.

zliu41:
Updated

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
ibuenros: Sorry, you're right, this will not produce that compile error because `withProps` is returning the correct type. However, this Builder is not easily extensible (for example, for a subclass of `FsRenameCommitStep`), will that be a problem?

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
ibuenros: Is there a unit test for the `get` method?

zliu41: yes, `FsCommitSequenceStoreTest.testGet()`

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
zliu41: Do you mean the private fields of the Builder should be changed to protected? I think that's reasonable.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,143 @@
...
+   */
+  public void execute() {
+    try {
+      for (CommitStep step : this.steps) {
+        if (!step.isCompleted()) {
+          step.execute();
+        }
+      }
+    } catch (Throwable t) {
+      log.error("Commit failed for dataset " + this.datasetUrn, t);
```
chavdar: Shouldn't we rethow t ?

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
ibuenros: No, the return types are wrong. Say you have
```java
class OtherBuilder extends FsRenameCommitStep.Builder {
   public OtherBuilder withFoo(Foo) {
      ...
   }
}
```
 The line `new OtherBuilder().to("path").withFoo(foo)` will not compile because `new OtherBuilder().to("path")`, according to the compiler, is of type `FsRenameCommitStep.Builder`, which does not have a `withFoo(Foo)` method. However, `new OtherBuilder().withFoo(foo).to("path")` will compile fine.

gobblin-api/src/main/java/gobblin/commit/CommitSequenceStore.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * A store for {@link CommitSequence}s. A {@link CommitSequence} is identified by job name and dataset URN.
+ *
+ * @author ziliu
+ */
+public interface CommitSequenceStore {
+
+  /**
+   * Whether a job name exists in the store.
```
chavdar: Can you clarify what it means for a job name to exist in the store? There is a CommitSequence for some dataset?

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -332,7 +370,17 @@ void commit() throws IOException {
         allDatasetsCommit = false;
       } finally {
         try {
-          persistDatasetState(datasetUrn, datasetState);
+          finalizeDatasetState(datasetState, datasetUrn);
+
+          if (commitSequenceBuilder.isPresent()) {
+            CommitSequence commitSequence = commitSequenceBuilder.get()
```
ydai1124: Can you create a method for this block. e.g. `persistCommitSequenceStore`?

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -397,24 +445,28 @@ private void finalizeDatasetStateBeforeCommit(JobState.DatasetState datasetState
...
+        || (this.jobCommitPolicy == JobCommitPolicy.COMMIT_ON_FULL_SUCCESS
+            && datasetState.getState() == JobState.RunningState.SUCCESSFUL);
   }
 
   /**
    * Commit the output data of a dataset.
    */
   @SuppressWarnings("unchecked")
-  private void commitDataset(JobState.DatasetState datasetState) throws IOException {
+  private Optional<CommitSequence.Builder> commitDatasetOrGenerateCommitSequenceBuilder(
```
ydai1124: Why not making `Optional<CommitSequence.Builder>` a class variable, and keep the return type `void`?

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
```diff
@@ -0,0 +1,142 @@
...
+ *    JSON string using {@link Gson}. Thus it requires that all {@link CommitStep}s can be serialized
+ *    and deserialized with {@link Gson}.
+ * </p>
+ *
+ * @author ziliu
+ */
+public class FsCommitSequenceStore implements CommitSequenceStore {
+
+  public static final String COMMIT_SEQUENCE_STORE_FS_URI = "commit.sequence.store.fs.uri";
+  public static final String COMMIT_SEQUENCE_STORE_DIR = "commit.sequence.store.dir";
```
ydai1124: Can we have a default value for `COMMIT_SEQUENCE_STORE_DIR`? Similar as in the `gobblin-standalone.properties`.

gobblin-runtime/src/test/java/gobblin/runtime/commit/CommitSequenceTest.java
```diff
@@ -0,0 +1,95 @@
...
+public class CommitSequenceTest {
+
+  private static final String ROOT_DIR = "commit-sequence-test";
+
+  private FileSystem fs;
+  private CommitSequence sequence;
+
+  @BeforeClass
+  public void setUp() throws IOException {
+    this.fs = FileSystem.get(URI.create("file:///"), new Configuration());
```
ydai1124: You can just use `FileSystem.getLocal(newConfiguration())`.

gobblin-runtime/src/test/java/gobblin/runtime/commit/CommitSequenceTest.java
```diff
@@ -0,0 +1,95 @@
...
+        .withProps(state).endStep().build();
+  }
+
+  @AfterClass
+  public void tearDown() throws IOException {
+    this.fs.delete(new Path(ROOT_DIR), true);
+  }
+
+  @Test
+  public void testExecute() throws IOException {
```
ydai1124: Do we have test for committing steps fail in the middle and verifying the state of commit sequence store?

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
zliu41: got it

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
zliu41: The default is in gobblin-mapreduce.properties and gobblin-standalone.properties.

gobblin-runtime/src/test/java/gobblin/runtime/commit/CommitSequenceTest.java
zliu41: This is not the responsibility of `CommitSequence` or `CommitSequenceStore`. It is the responsibility of whoever uses them to ensure that they only delete `CommitSequence` from `CommitSequenceStore` when appropriate.

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Add a {@link CommitStep} to the commit sequence.
+     */
+    public Builder addStep(CommitStep step) {
+      this.steps.add(step);
+      return this;
+    }
+
+    public CommitSequence build() {
+      return new CommitSequence(this);
```
chavdar: The build() method should do validation of the builder variables, e.g. name and datasetUrn should not be null. Is it acceptable to have no commit steps?

gobblin-api/src/main/java/gobblin/commit/CommitSequence.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ *                                  .build();
+ *      }
+ *    </pre>
+ * </p>
+ *
+ * @author ziliu
+ */
+@Slf4j
+public class CommitSequence {
```
chavdar: Please add @Alpha annotations.

gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @SuppressWarnings("unchecked")
+    public T withProps(State props) {
+      this.props = new State(props.getProperties());
+      return (T) this;
+    }
+
+    public CommitSequence.Builder endStep() throws IOException {
+      Preconditions.checkState(this.commitSequenceBuilder.isPresent());
+      this.commitSequenceBuilder.get().addStep(build());
```
chavdar: You should verify that the builder is present or throw a RuntimeException.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -507,6 +506,7 @@
   /**
    * Other configuration properties.
    */
+  public static final String DELIVERY_SEMANTICS = "delivery.semantics";
```
chavdar: Please make sure the property is namespaced, say, with "gobblin.runtime." Thus, we can avoid clashes with user-defined properties.

gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
```diff
@@ -0,0 +1,50 @@
...
+   * Each data record from the source is guaranteed to be delievered exactly once.
+   */
+  EXACTLY_ONCE;
+
+  /**
+   * Get the devliery semantics type from {@link ConfigurationKeys#DELIVERY_SEMANTICS}.
+   * The default value is {@link Type#AT_LEAST_ONCE}.
+   */
+  public static DeliverySemantics parse(State state) {
+    String value = state.getProp(ConfigurationKeys.DELIVERY_SEMANTICS, AT_LEAST_ONCE.toString()).toUpperCase();
```
chavdar: We should have a default, AT_LEAST_ONCE .

gobblin-api/src/main/java/gobblin/publisher/DataPublisher.java
```diff
@@ -18,19 +18,38 @@
...
 
 
 /**
  * Defines how to publish data and its corresponding metadata. Can be used for either task level or job level publishing.
+ *
+ * If exactly-once semantics is used, it will construct a {@link CommitSequence} instead of actually publishing
+ * the data. The data will be published when the {@link CommitSequence} is executed, which will be after the
+ * {@link CommitSequence} has been fully constructed and persisted into the {@link gobblin.commit.CommitSequenceStore}.
  */
 public abstract class DataPublisher implements Closeable {
```
chavdar: I think for now we should use a different publisher class. I don't think the current API for DataPublisher fits into the new model. For example, there shouldn't be publishData() and publishMetadata() methods. All of these should be commit steps.

We'll have to figure out later how to do the migration. We may have to introduce some common interface.

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,134 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * A {@link CommitStep} for renaming files within a {@link FileSystem} or between {@link FileSystem}s.
+ *
+ * @author ziliu
+ */
+@Slf4j
+public class FsRenameCommitStep extends CommitStepBase {
```
chavdar: Could add a unit test for this?

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @SuppressWarnings("unchecked")
+    public T withDstFs(FileSystem dstFs) {
+      this.dstFs = dstFs;
+      return (T) this;
+    }
+
+    @Override
+    public CommitStep build() throws IOException {
+      return new FsRenameCommitStep(this);
```
chavdar: Please add validation if srcPath and dstPath have been properly set.

gobblin-runtime/src/main/java/gobblin/runtime/commit/FsCommitSequenceStore.java
```diff
@@ -0,0 +1,140 @@
...
+ *    This implementation serializes a {@link CommitSequence} along with all its {@link CommitStep}s into a
+ *    JSON string using {@link Gson}. Thus it requires that all {@link CommitStep}s can be serialized
+ *    and deserialized with {@link Gson}.
+ * </p>
+ *
+ * @author ziliu
+ */
+public class FsCommitSequenceStore implements CommitSequenceStore {
+
+  public static final String COMMIT_SEQUENCE_STORE_FS_URI = "commit.sequence.store.fs.uri";
```
chavdar: Please add a "gobblin.runtime" namespace.

gobblin-api/src/main/java/gobblin/commit/CommitStepBase.java
zliu41: this is verified in the previous line.

gobblin-api/src/main/java/gobblin/commit/DeliverySemantics.java
zliu41: Yes, default is AT_LEAST_ONCE

gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
zliu41: Added.

zliu41:
@chavdar @ydai1124 addressed comments (637cf20)

kadaan:
@zliu41 Am I correct in believing that this implementation requires that the pending upload files be stored on a shared filesystem such as S3 or HDFS?  If so, this should be called out.  Additionally, we should detect use of LocalFileSystem and fail to start.

PR #572, opened 2015-12-23
Add file level uploading and state tracking to BaseDataPublisher.

kadaan:


gobblin-api/src/main/java/gobblin/configuration/SourceState.java
```diff
@@ -51,7 +52,7 @@
  */
 public class SourceState extends State {
 
-  private static final Set<Extract> EXTRACT_SET = Sets.newConcurrentHashSet();
+  private static final Set<Extract> EXTRACT_SET = Sets.newSetFromMap(new ConcurrentHashMap<Extract, Boolean>());
```
zliu41: What's the benefit of this change?

kadaan: The prior causes me so many headaches because it is only in guava 15+, but the hadoop dependencies use an older version of guava.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -299,14 +270,75 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+      if (!this.writerFileSystemByBranches.get(branchId).exists(writerOutputDir)) {
+        LOG.warn(String.format("Branch %d of WorkUnit %s produced no data", branchId, state.getId()));
+        continue;
+      }
+
+      // The directory where the final output directory for this job will be placed.
+      // It is a combination of DATA_PUBLISHER_FINAL_DIR and WRITER_FILE_PATH.
+      Path publisherOutputDir = getPublisherOutputDir(state, branchId);
+
+      return new WorkUnitPublishGroup(branchId, writerOutputDir, publisherOutputDir);
```
zliu41: Does this only return group with `branchId=0`? What about other branches?

kadaan: No, it should head down all branches.  From my understanding, each writerOutputDir is in one and only one branch.  If that is true, this code will return when the first branch that contains the writerOutputDir is found.

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -266,6 +268,9 @@ public Void call() throws Exception {
             if (group.isPresent()) {
               HadoopUtils.setGroup(dstFs, dst, group.get());
             }
+            if (commitAction != null) {
+                commitAction.apply();
```
zliu41: Why do you call `apply()` after moving every single file? Shouldn't it only be called once, after all files of a task have been moved?

kadaan: Good call

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -299,14 +270,75 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+      this.state = state;
+    }
+
+    @Override
+    public void apply() {
+       state.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
+    }
   }
 
+  private class WorkUnitPublishGroup {
```
zliu41: Could you summarize what is the benefit for grouping `WorkUnitStates` by `branchId`, `writerOutputDir`, `publisherOutputDir` like this, rather than handling all `WorkUnitStates` individually? It seems for `WorkUnitStates` in the same group, you are still handling them individually rather than doing something common to them.

kadaan: There is common setup code that needs to run for each branch/writerOutputDir/publisherOutputDir, such as deleting the publisherOutputDir.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
kadaan: Looking at this a little more, it seems as if the each workunit can be part of multiple branches and it should return a set of groups, one per branch the workunit is part of.

kadaan:
@zliu41 What do you guys do to run `gobblin.runtime.mapreduce.MRJobLauncherTest` locally.  Whenever I try it fails.

kadaan:
@zliu41 I changed how this works such that the state of a WorkUnit isn't set to COMMITTED until all file for each branch of the WorkUnit have been published. Please let me know if this behavior seems correct.

kadaan:
@zliu41 Anything else needed for this to be merged?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -15,13 +15,14 @@
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
+import com.google.common.collect.ImmutableList;
```
liyinan926: Please update the import order following the coding style guide.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -240,73 +224,76 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
+    }
 
+    public Path getPublisherOutputDir() {
+      return publisherOutputDir;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
```
liyinan926: This can be replaced by `if (!(o instanceof WorkUnitPublishGroup))`.

gobblin-utility/src/main/java/gobblin/util/Action.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+public interface Action {
+  public void apply();
```
liyinan926: We probably should allow it to thrown an exception.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -152,8 +153,27 @@ public static void movePath(FileSystem srcFs, Path src, FileSystem dstFs, Path d
     if (srcFs.getUri().equals(dstFs.getUri())) {
       renamePath(srcFs, src, dst);
     } else {
-      if (!FileUtil.copy(srcFs, src, dstFs, dst, true, false, dstFs.getConf())) {
-        throw new IOException(String.format("Failed to move %s to %s", src, dst));
+      boolean isSourceFileSystemLocal = srcFs instanceof LocalFileSystem;
```
liyinan926: This won't cover `RawLocalFileSystem`. What's the benefit of differentiating here?

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -23,6 +23,8 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.locks.Lock;
 
+import com.google.common.collect.ImmutableList;
```
liyinan926: Order of imports.

gobblin-utility/src/main/java/gobblin/util/ParallelRunner.java
```diff
@@ -249,31 +260,56 @@ public Void call() throws Exception {
...
+            lock.unlock();
           }
-          return null;
-        } catch (FileAlreadyExistsException e) {
-          LOGGER.warn(String.format("Failed to move %s to %s: dst already exists", src, dst), e);
-          return null;
-        } finally {
-          lock.unlock();
         }
+        if (commitAction != null) {
```
liyinan926: Can the `commitAction` be an `Optional`?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -240,73 +224,76 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
+       state.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
     }
-    return this.parallelRunners.get(uri);
   }
 
-  @Override
-  public void publishMetadata(Collection<? extends WorkUnitState> states) throws IOException {
-    // Nothing to do
-  }
+  private class WorkUnitPublishGroup {
```
liyinan926: Seems this can be a `static class`.

liyinan926:
@kadaan I just briefly went through the code and commented mostly on the syntax/style aspects. I will need a bit more time to go over this and PR #569 again to understand the overlap. Stay tuned.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
kadaan: The benefit of determining that the source filesystem is local is that we can optimize how we move the files to the destination filesystem.  For instance, when copying to s3a filesystem the file is read via an inputstream into a buffer file at fs.s3a.buffer.dir.  Once the data has been copied to the buffer file, it is then uploaded to s3.  If we know that the source filesystem is local we can enable an optimization which skips copying to the buffer file.  This happens because s3a implements moveFromLocalFile to not copy to the buffer file first.

I'm not certain when a RawLocalFileSystem is used and therefore didn't check for that.  Additionally, the moveFromLocalFile expects that the file is in a LocalFileSystem.

liyinan926: OK, got it.

kadaan:
@liyinan926 AFAIK, #569 doesn't really address these issues.  The underlying problem here is that currently publishing can "succeed" and the files don't actually make it to the publish location.  With this issue gobblin doesn't even ensure "at least once" publishing of the data.  #569 adds the concept of "exactly once" publishing, but I feel that "at least once" should be working correctly before adding more complexity.

liyinan926:
@kadaan got it. Yes, that's a bug that needs to be fixed.

kadaan:
@liyinan926 I detailed this issue in #561

kadaan:
@liyinan926 All outstanding comments addressed.

kadaan:
@zliu41 @liyinan926: Anything else you would like me to change?  Any further discussions around this needed?

zliu41:
@kadaan previously `publishMultiTaskData` has the option of moving the entire task output folder at once (if `data.publisher.replace.final.dir=true`). Now this option seems to be gone, and it always moves each file individually. Even with `ParallelRunner`, this could be a lot slower.

I'm thinking of adding that option back. When using that option, since there is only a single `rename` operation for each branch, we don't use `ParallelRunner` (hence no inconsistency issue) but move the folder synchronously in the publisher. What do you think?

kadaan:
@zliu41 It really is filesystem specific as to how the move can be done.  If you are moving from HDFS to the _same_ HDFS, then renaming the folder is fine.  As soon as you are doing any other kind of move (local -> HDFS, local -> S3, etc.) then the rename is an issue.  The problem really is that the rename is a hadoop filesystem call and what it does underneath is hidden from gobblin.  For instance, when the filesystem is HDFS then it can just rename the folder.  However, when the filesystem is S3, it needs to walk the directory tree and and upload each file individually.  

Really, what is needed is different code paths for the different filesystems.  That way we could do an atomic folder copy for HDFS and update all WorkUnitStates when done.  For the S3 filesystem we could move the folder like I have done in this PR and update the WorkUnitStates while moving.  The problem with this approach is that it doesn't extend to new filesystems.  

_The method in this PR will work for all hadoop filesystems._

kadaan:
@zliu41 As this seems to be a pretty high priority issue, I'd like to see if we could hash out a solution in a more interactive method.  What do you think about discussing this in IRC. I've setup a gobblin channel on Freenode and can add all of you guys as admins if this seems acceptable.

zliu41:
@kadaan sounds good. I haven't used freenode before, is it like a chatroom? Let me know how it should be used.

Actually it could be slower even if `data.publisher.replace.final.dir=false`, because previously, `addWriterOutputToExistingDir` doesn't walk the directory tree, but only does a `listStatus` and moves each file and subfolder directly under the task output folder to the publisher output folder. Now if we always moves files individually, I guess all jobs using `publishMultiTaskData` will suffer.

kadaan:
@zliu41 I added a badge for the IRC channel to the readme in this PR: #609

kadaan:
I'd really love to be able to quantify what the performance loss would be for HDFS.  For S3, there should be zero performance loss as internally the S3AFileSystem was walking the directory tree and uploading each file individually.

zliu41:
@cbotev what do you think?

kadaan:
@chavdar zliu41 was asking you a question, but I think he mis-typed your alias.

kadaan:
@chavdar @zliu41 @liyinan926 Would you guys have some time to get in IRC today and discuss this?

zliu41:
@sahilTakiar can you take look. Thanks.

sahilTakiar:
Will take a look today

kadaan:
One thought I had to solve your possible HDFS performance issues is to detect specific filesystems and handle them differently. For example, if we detect that the filesystem is HDFS we can optimize the publish by performing a rename of the folder and then mark the workunitstates as complete. On the other hand, for any other filesystem we will continue to use the method proposed in this PR.  

What do you think?

sahilTakiar:
To confirm we are talking about `HadoopUtils.movePath`? Doesn't this method already handle this? If the `URI`s are the same then it will just do a `renamePath` (so if its HDFS -> HDFS or local -> local then it just invokes a rename operation which is what we want); and this rename operation can be at a folder. We basically want to avoid renamining each file individually when the data is moving from HDFS -> HDFS or local -> local.

If the file systems are different then it uses `FileUtil.copy` which just walks the tree and uploads each file individually. This is fine, the only optimization we may want to make is to walk the tree ourselves and spawn a new thread for each file upload. I don't think it is possible to atomically bulk upload a bunch of files to HDFS, you would have to bulk upload them to a staging folder and then perform a rename operation.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -152,8 +153,31 @@ public static void movePath(FileSystem srcFs, Path src, FileSystem dstFs, Path d
     if (srcFs.getUri().equals(dstFs.getUri())) {
       renamePath(srcFs, src, dst);
     } else {
-      if (!FileUtil.copy(srcFs, src, dstFs, dst, true, false, dstFs.getConf())) {
-        throw new IOException(String.format("Failed to move %s to %s", src, dst));
+      // Determining if the source filesystem in local will allow us to call moveFromLocalFile and give the
+      // destination filesystem the ability optimize the transfer. For example, when the source file is
+      // local the S3AFileSystem will skip buffering it into the fs.s3a.buffer.dir and immediately transfer the
+      // file to S3.
+      boolean isSourceFileSystemLocal = srcFs instanceof LocalFileSystem;
```
sahilTakiar: Why do you need this optimization exactly? The method `FileSystem.moveFromLocalFile` just invokes `FileUtil.copy` anyway, to its really the same thing.

kadaan: The benefit of determining that the source filesystem is local is that we can optimize how we move the files to the destination filesystem. For instance, when copying to s3a filesystem the file is read via an inputstream into a buffer file at fs.s3a.buffer.dir. Once the data has been copied to the buffer file, it is then uploaded to s3. If we know that the source filesystem is local we can enable an optimization which skips copying to the buffer file. This happens because s3a implements moveFromLocalFile to not copy to the buffer file first.

kadaan:
You are correct that `movePath` correctly handles doing a rename when moving between the same filesystem and a copy/delete when moving between different filesystems.  Also, like you mentioned above, this PR already walks "the tree ourselves and spawn a new thread for each file upload".  It does this by calling `movePath` for each file within the tree.  When the call succeeds it updates the WorkUnitState to complete.  @zliu41 concern is that this will reduce performance for moves within the same filesystem.  My suggestion above was that the publisher detect that the publish is between the same filesystem and that the filesystem is HDFS (not sure what the FileSystem implementation for this is), then optimize the publish by calling `movePath` on each root folder (not walking the directory tree) and when successful update all WorkUnitStates for that path to complete.

Does this make sense?  I'll be away for a bit , but will be back on the computer in about 1.5 hours.  If you would like to discuss this more then, can we do it over IRC: https://webchat.freenode.net/?channels=gobblin

sahilTakiar:
Yes, that sounds good; but if `HadoopUtils.movePath` already does the differentiation between moves on the same `FileSystem` vs. moves on different `FileSystem`s why do we also need to add that distinction in the Publisher.

Can we not simply just say `HadoopUtils.movePath(rootDirectoryPath)` and let the method handle everything.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: Ok makes sense

kadaan:
In this PR, the BaseDataPublisher determines the "moves" needed and uses ParallelRunner to do the work.  `ParallelRunner.movePaths` in turn calls `HadoopUtils.movePath` do the actual move. Due to this, we need to figure out ahead of calling `ParallelRunner.movePaths` the set of moves that can be atomically accomplished.  

One potential issue is that in BaseDataPublisher, there is nothing that ensures that all files are being moved between the same source and destination filesystem.  Additionally, it seems as if a given WorkUnitState can have multiple files associated with it. Before marking the WorkUnitState as complete we need to ensure that all the associated files are moved to their destinations (even if they are different destinations).  All of this seems to really complicate the process.

sahilTakiar:
Could we instead have `ParallelRunner.movePaths` return a `List` of `MoveAction`s where a `MoveAction` is a wrapper about the source and destination `Path, and the `Future` object. Internally the method will create a `MoveAction` for an entire directory if it can commit that entire directory atomically, if it needs to upload each file individually then it creates a `MoveAction` for each file.

This should give the `BaseDataPublisher` enough information to determine when each `WorkUnitState` has been published, right?

kadaan:
@sahilTakiar Publishing now happens in two stages.  The first is to group all publish commands (source and destination path) by the source filesystem, source path root, destination filesystem, and destination path root.  Then we walk over the groups.  If the source and destination filesystems are equivalent and the filesystem supports atomically renaming a folder, we rename the source path root to the destination path root.  Otherwise, we run a move for each publish command.

NOTE: each publish command has an associated CommitAction which indicates the number of publishes which must succeed before the WorkUnitState is set to COMPLETE.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -240,73 +178,234 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
+        commitAction.apply();
+      }
     }
-    return this.parallelRunners.get(uri);
   }
 
-  @Override
-  public void publishMetadata(Collection<? extends WorkUnitState> states) throws IOException {
-    // Nothing to do
+  private static class CommitAction implements Action {
```
sahilTakiar: Can you rename this class to something like `MarkWorkUnitCommittedAction`, and also add some javadocs for this class.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -240,73 +178,234 @@ protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId)
...
-  public void publishMetadata(WorkUnitState state) throws IOException {
-    // Nothing to do
+  private static class PublishGroup {
+    private final int branchId;
+    private final FileSystem srcFs;
+    private final Path srcPath;
+    private final FileSystem dstFs;
+    private final Path dstPath;
+
+    public PublishGroup(int branchId, FileSystem srcFs, Path srcPath, FileSystem dstFs, Path dstPath) {
```
sahilTakiar: For a given `List` of `WorkUnitState`s, when will this code ever create duplicate `PublishGroup`s? Won't the (branch, srcFs, srcPath, dstFs, dstPath) uniquely identify every `WorkUnitState`?

kadaan: There will be many duplicates because srcPath and dstPath are just the root paths and not the full path to the file.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataPublisher.java
```diff
@@ -43,30 +36,4 @@
...
   }
-
-  /**
-   * This method needs to be overridden for TimePartitionedDataPublisher, since the output folder structure
-   * contains timestamp, we have to move the files recursively.
-   *
-   * For example, move {writerOutput}/2015/04/08/15/output.avro to {publisherOutput}/2015/04/08/15/output.avro
-   */
-  @Override
-  protected void addWriterOutputToExistingDir(Path writerOutput, Path publisherOutput, WorkUnitState workUnitState,
```
sahilTakiar: Can you explain a little more why this method is no longer necessary?

sahilTakiar:
@kadaan this approach looks good for the time being (pending the 1 question I had). A few final thoughts before I merge this:

* I still think that instead of passing in a `CommitAction` to `ParallelRunner.renamePath`, the method should just return a `List<(Future, Method Params)>`, but your approach works to so we can stick to this for now; I think the `ParallelRunner` doesn't have the right level of abstraction so that is part of the reason too
* Down the line, the `BaseDataPublisher` needs to be re-designed - the class has become too obtuse and should be broken down into simple components - but that is a question for another day

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataPublisher.java
kadaan: The first reason is that the `addWriterOutputToExistingDir` method no longer exists on `BaseDataPublisher`.

sahilTakiar: Ok, but is there any loss in functionality my removing this method? Will the `TimePartitionedDataPublisher` still work?

kadaan: As far as I could tell the main difference with the TimePartitionedDataPublisher and the BaseDataPublisher is:
  1. TimePartitionedDataPublisher _recursively_ listed all files under the writerOutputDir and moved them to the publisherOutputDir
  2. BaseDataPublisher listed all files under the writerOutputDir and moved them to the publisherOutputDir

BaseDataPublisher now copies the files as specified in `ConfigurationKeys.WRITER_FINAL_OUTPUT_FILE_PATHS`.  Because of this, there is no need to list the files (recursively or otherwise) within the writerOutputDir.

kadaan:
@sahilTakiar One concern I have is that the original BaseDataPublisher did a _ton_ of stuff and there aren't really a many tests around it.  For instance, there are different behaviors when publishData(state) is called vs. publishData(state[]).  I've tried to keep this behavior the same as before while also removing a bunch of different code branching and duplication.  I'm slightly concerned that I may have inadvertently changed some behavior.  My biggest concern is that Task.publishTaskData behavior has changed, but as I don't really know that code path it is hard for me to say what the change could be.

gobblin-core/src/main/java/gobblin/publisher/TimePartitionedDataPublisher.java
sahilTakiar: Ok that should work then.

sahilTakiar:
@kadaan yes I have the same concern about the lack of unit tests for this. If I have some time in the near future I may work on fixing that.

For the `Task.shouldPublishDataInTask()` method what was the specific concern you had? There isn't anything that jumps out to me that could cause problems. Also, as long as the publisher doesn't throw an error for the job it will be marked as committed (line 241 of `Task.java`).

Let me know if you all change have been committed, then I will do the merge.

kadaan:
My main concern is that the work unit could get marked as committed by the publisher before it really should be. After looking at this more today, I wonder if this may be more complex then needed. It seems like gobblin has a couple of different ways to handle publishing the data. 

The first is having the job perform the publish. This seems to be like a reduce stage in a Map-Reduce program as all tasks much complete and write to staging before any files can be published. This enables jobs (with one restriction) to be created that do not publish any files unless all tasks have completed successfully. The afore mentioned restriction is that the staging filesystem must support atomic folder renames (HDFS) so that the entire folder will (or will not) be moved from staging to publish. An additional restriction is that when gobblin is run under yarn, the filesystem must be shared among all tasks (HDFS). I belive the current BaseDataPublisher functions correctly with these assumptions. This is the default mode for gobblin. Trying to operate in this mode with the staging folder in S3 causes the problems that were the impetus for this PR.

The second is to allow the tasks to publish the files. The original code for this is similar to the generic publish code in this PR. This method can result in a subset of the files for the job being successfully published. Therefore it is impossible to perform an all or nothing publish. Additionally, it is not possible to have the publish overwrite the existing data.  That said, this method should be significantly faster as the publish can be distributed and can start sooner.  The staging filesystem _does not_ need to be shared among tasks. The filesystem _does not_ need to support atomic folder renames.  This mode is must be explicitly configured for each job.

Does this seem right?

It feels like job level publish was the original system and that task level publish was added afterwards.   Is this correct?

Thinking about the use cases for both of these methods leads me to:

    Job level publish with folder replace is a great fit for loading dimension tables.  If dimension tables were only partially replaced, queries which join against these tables would be "missing" data (due to being dropped out of the join).

    Task level commit seems to be a great fit for fact tables. Partial data commit will be fixed on subsequent runs. No records will be "missing". 

Does this match how you use this within LinkedIn?

If this is the case, I'm not sure that this PR is providing much benefit. I agree that the BaseDataPublisher is doing too much. It would be nice to split up the task and job publish codebases.

@sahilTakiar, let me know what you think.

sahilTakiar:
Hey @kadaan,

* Job-Level Publish
    * You are correct that we use this for publishing dimension data
    * Yes the idea is that we want this Publish to be as atomic as possible, as you said, partial reading of dimension data is bad
    * The Job-Level Publish has the advantage that all `Task`s have been completed - this means we know all the (say for a dimension table), is on HDFS, so we can safely publish it using one atomic operation

* Task-Level Publish
    * Yes, this is more useful for fact style data (e.g. the original use case for this was ingesting fact-style data from Kafka)
    * The Task Level Publish basically uses a `SingleTaskDataPublisher extends DataPublisher` - which allows you to publish data one `WorkUnitState` at a time.
    * I don't think the Task-Level Publish has any semantics or guarantees about overwriting data, or publishing data atomically, that is really up to the implementation.

However, as you said there are certain scenarios where you want Job-Level Publishing over Task-Level Publishing. When you want to publish data from multiple Tasks at the same time (usually to the same folder), you want to use the Job-Level Publisher.

Another relevant piece of information, is that Gobblin has a concept of `WorkUnit`s and `Extract`s. An `Extract` is a logical grouping of `WorkUnits` - such as a table in a database, or a partition in Kafka. The goal of the `BaseDataPublisher` was to publish all data from an `Extract` at the same time, atomically. By default, `WorkUnit`s from the same `Extract` end up in the same writer output folder.

I see a few options at this point:
* I still think this PR provides a concrete benefit, because rename operations (atomic or not) can fail, and we shouldn't be marking `WorkUnitState`s as `COMMITTED` until we know for sure the publish was successful.
* If this PR is getting too hectic, it may just be better to implement a new `DataPublisher` that solves the S3 problems.
* A lot of the problems seem to arise from the `ParallelRunner` - we could have a simple version of the `BaseDataPublisher` which does not use the `ParallelRunner`.

As for your last comment, yes there are way too many config options in Gobblin. We have discussed this problem quite a few times within our team and are still working on ways of simplifying the flow. We also want to work on general ease of use for Gobblin. Any suggestions or additional comments you have on this are definitely welcome. I also encourage you to fill out the following Google Form with any suggestions you have: http://goo.gl/forms/LtRRxlxIfp - we are trying to collect more feedback from our users to see what problems they are facing the most.

sahilTakiar:
@kadaan any thoughts on this? If you don't have resources to work on this, just let me know. It's still a high priority bug for us, so we want to get it fixed too.

kadaan:
@sahilTakair:  I’m still thinking about how best to use this PR.  We switched to task publishing which addressed most of our issues.  In this case, renames/moves that fail will mark the task as failed. At the moment we are just trying to get our first use case of gobblin into prod and with task publishing we can do that.  Additionally we need to fix an issue with state and duplicate data which is causing state sizes > 300mb.  Once we get this push done, I’ll come back to this PR.

kadaan:
@sahilTakiar Because #569 has been merged, it will take a little longer to get this cleaned up and working again.  I still think it is worth doing because I believe that it is a better overall solution than #658.

kadaan:
@sahilTakiar Have you guys looked at using parseq instead of ParallelRunner?

chavdar:
Hi!

That library requires Java 8. Gobblin runs on Java 7.

kadaan:
:(

sahilTakiar:
Haven't really look at ParSeq before, it seems version 1.4.2 is Java 7 compatible though. Or we can make a placeholder JIRA for ParSeq integration when Gobblin Java 8 migration is done.

PR #576, opened 2015-12-28
bug fix for addWriterOutputToExistingDir in recursive dir.

JiHyunSong:
when I using timebasedpartitioner, FileAlreadyExistsException occurs.

ref : https://groups.google.com/forum/#!topic/gobblin-users/UaBRPjJ7NfE

Thanks to @lihui

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -21,6 +21,8 @@ import java.util.Map;
 import java.util.Set;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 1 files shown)


JiHyunSong:
TODO : working at hadoop 1.2.1 ... I only used hadoop 2.x

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -258,37 +260,53 @@ protected void addSingleTaskWriterOutputToExistingDir(Path writerOutputDir, Path
...
-
-      // Preserve the file name if configured, use specified name otherwise
-      Path finalOutputPath =
-          preserveFileName
-              ? new Path(publisherOutputDir,
-                  workUnitState.getProp(ForkOperatorUtils.getPropertyNameForBranch(
-                      ConfigurationKeys.DATA_PUBLISHER_FINAL_NAME, this.numBranches, branchId)))
-          : new Path(publisherOutputDir, status.getPath().getName());
+    FileSystem fs = this.writerFileSystemByBranches.get(branchId);
+    RemoteIterator<LocatedFileStatus> iterator = fs.listFiles(writerOutputDir, true);
```
chavdar: Why do you need to use this versus listStatus() ? LocatedFileStatus is Hadoop 2-only and we try to maintain Hadoop 1 compatibility.

chavdar: Is it because you want to do the recursive list? The main issue with your approach is that it breaks atomicity of publishing. The current implementation tries to publish at the directory level -- a single rename publishes all files in a single directory. This is very important to avoid race conditions where downstream consumers may see partial data.

I think I understand the underlying issue but the fix may be slightly trickier. Before adding the partitioned writer the publishing happened at the level of directories which are direct children of the writer output directory. With the partitioned writer, that may not be the case any more and the atomic publish needs to happen at a deeper level but I don't think doing it at file level is the right thing to do in most cases.

kadaan: @chavdar The idea of atomic publishing here really only holds true for HDFS.  Once you publish to S3, this doesn't work.

sahilTakiar:
Closing this right now due to discussion from https://groups.google.com/forum/#!topic/gobblin-users/UaBRPjJ7NfE

PR #578, opened 2016-01-04
Bug fix for publishing to S3

sahilTakiar:
Bug fix in `WriterUtils.mkdirsWithRecursivePermission` for publishing to file systems where the root directory does not already exist.

@zliu41 can you review?


liyinan926:
LGTM.

PR #582, opened 2016-01-05
Adding configurable MetricFilter

sahilTakiar:
* Added a configuration key for specifying custom `MetricFilter`s
* Added `NotificationTimerFilter` to filter out the reporting of `MetricContext.notificationTimer`
* Fixed a dangling `MetricContext` in `InstrumentedConverterBase`

My original approach was to make the creation of the `notificationTimer` configurable but that turned out to be difficult given the current handling of config keys, so I added the `MetricFilter` for now since it will probably be a useful feature in other scenarios too.

@ibuenros can you review?

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
@@ -42,6 +42,9 @@ import lombok.extern.slf4j.Slf4j;
 
 import gobblin.metrics.InnerMetricContext;
 import gobblin.metrics.context.ReportableContext;
+import gobblin.metrics.metric.filter.MetricFilters;
+import gobblin.metrics.metric.filter.MetricNameRegexFilter;
+import gobblin.metrics.metric.filter.MetricTypeFilter;
 import gobblin.util.ExecutorsUtils;
 
 
@@ -61,6 +64,9 @@ public abstract class ScheduledReporter extends ContextAwareReporter {
       appendHours().appendSuffix("H").
       appendMinutes().appendSuffix("M").
       appendSeconds().appendSuffix("S").toFormatter();
+
+  private static final String METRIC_FILTER_NAME_REGEX = "metric.filter.name.regex";
+  private static final String METRIC_FILTER_TYPE_LIST = "metric.filter.type.list";
 
   @VisibleForTesting
   static int parsePeriodToSeconds(String periodStr) {
@@ -88,6 +94,8 @@ public abstract class ScheduledReporter extends ContextAwareReporter {
   }
 
   private final ScheduledExecutorService executor;
+  private final MetricFilter metricFilter;
+
   private Optional<ScheduledFuture> scheduledTask;
   private int reportingPeriodSeconds;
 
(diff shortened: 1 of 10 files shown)


gobblin-metrics/src/main/java/gobblin/metrics/metric/filter/NotificationTimerFilter.java
```diff
@@ -0,0 +1,24 @@
...
+ * Implementation of {@link MetricFilter} that ignores metrics with name
+ * {@link MetricContext#GOBBLIN_METRICS_NOTIFICATIONS_TIMER_NAME}.
+ *
+ * <p>
+ *   The {@link MetricContext#GOBBLIN_METRICS_NOTIFICATIONS_TIMER_NAME} metric is used to profile how long it takes to
+ *   send {@link gobblin.metrics.notification.Notification}s; this is useful for profiling the gobblin-metrics library.
+ *   So most of the time this {@link com.codahale.metrics.Timer} can be filtered out.
+ * </p>
+ */
+public class NotificationTimerFilter implements MetricFilter {
```
ibuenros: Can we make this more generic? We might add more profiling metrics in the future, and the same filter should work for all of them. Maybe have all profiling metrics share a prefix, and this filter should check for that prefix?

gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
```diff
@@ -97,6 +101,26 @@ public ScheduledReporter(String name, Config config) {
         ExecutorsUtils.newThreadFactory(Optional.of(log), Optional.of("metrics-" + name + "-scheduler")));
     this.reportingPeriodSeconds = parsePeriodToSeconds(
         config.hasPath(REPORTING_INTERVAL) ? config.getString(REPORTING_INTERVAL) : DEFAULT_REPORTING_INTERVAL_PERIOD);
+    this.metricFilter = createMetricFilter(this.config);
+  }
+
+  /**
+   * Create a {@link MetricFilter} from the config key {@link #METRIC_FILTER_CLASS}.
+   */
+  private MetricFilter createMetricFilter(Config config) {
```
ibuenros: See https://docs.google.com/document/d/1fX4Xg4LW0svfkSq3OsnjByQ3vxaILwAxRIaaDDzM78I/edit#heading=h.frw91twb0jx0 (LinkedIn private) for the plan on metric filters. The idea is for the declaration of metric filters to be expressive, instead of the usual classes. 
Metric filters will be based on regular expressions on metric names and metric types. If we have a prefix for profiling metrics, then we could use `^(?!internal\.profiling\.).*`.
If doing the regex is too much work, we can leave it for later, but let's make sure whatever is implemented now does not make it hard or ugly to implement the plan later on.

sahilTakiar:
@ibuenros read through your comments and the design doc. The approach you suggest didn't sound too difficult so I went ahead and implemented it (PR is updated with the suggested approach). I haven't added any javadocs or unit tests yet because I want to make sure the approach is right first. If this approach looks good, I will update the java docs and unit tests

gobblin-metrics/src/main/java/gobblin/metrics/metric/filter/MetricTypeRegexFilter.java
```diff
@@ -0,0 +1,24 @@
...
+  private final Matcher matcher;
+
+  public MetricTypeRegexFilter(String regex) {
+    this.regex = Pattern.compile(regex);
+    this.matcher = this.regex.matcher("");
+  }
+
+  @Override
+  public boolean matches(String name, Metric metric) {
+    return this.matcher.reset(metric.getClass().getSimpleName()).matches();
```
ibuenros: Is there a reason to use regex for the metric types? There are only five types of metrics, so it is easy to make it just a list. Using regex, on the other hand, requires for example matching capitalization of the class names.

sahilTakiar:
@ibuenros updated. `MetricTypeFilter` now takes in a `List` of `Metrics` rather than regex.

ibuenros:
Actually, for metric type, it might be easier to do the filtering in the `ScheduledReporter` itself. The `report` method calls `report(counters, meters, ...)`, so for each argument you could have a switch statement 
`report(metricTypesToReport.contains(Counter) ? context.getCounters(this.filter) : new ArrayList<>(), ...)`.

sahilTakiar:
Is there a benefit to doing the filtering in `ScheduledReporter` itself vs. using `MetricFilter`? The advantage of the current approach is that all metric filtering is done through a single interface, `MetricFilter`.

ibuenros:
It is more efficient (check only once per metric type, instead of checking every single metric), and imo easier to understand, but that's debatable.
The other approach is fine. Just a suggestion, instead of doing string comparison, consider using class comparison (e.g. `metric instanceof Counter`).

sahilTakiar:
Updated, added unit tests and java docs.

ibuenros:
LGTM

PR #584, opened 2016-01-05
Limit typesafe config conversion to metrics keys

pcadabam:
@ibuenros and @sahilTakiar can you please review?
The change limits the conversion of properties to typesafe config only for keys that start with metrics.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
@@ -12,16 +12,18 @@
 
 package gobblin.util;
 
+import gobblin.configuration.State;
+
 import java.util.Map;
 import java.util.Properties;
 
-import com.google.common.collect.ImmutableMap;
+import org.apache.commons.lang3.StringUtils;
 
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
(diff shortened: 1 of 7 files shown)


gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -67,9 +69,32 @@ public static State configToState(Config config) {
...
+   *
+   * @param properties the given {@link Properties} instance
+   * @param prefix of keys to be converted
+   * @return a {@link Config} instance
+   */
+  public static Config propertiesToConfig(Properties properties, Optional<String> prefix) {
     ImmutableMap.Builder<String, Object> immutableMapBuilder = ImmutableMap.builder();
     for (Map.Entry<Object, Object> entry : properties.entrySet()) {
-      immutableMapBuilder.put(entry.getKey().toString(), entry.getValue());
+      if (prefix.isPresent()) {
```
ibuenros: nit: you can replace this by
```java
if (StringUtils.startsWith(entry.getKey().toString(), prefix.or("")) {
  immutableMapBuilder.put(entry.getKey().toString(), entry.getValue());
}
```
to get rid of the if (slightly simpler imo).

ibuenros:
LGTM other than small comment.

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
pcadabam: Done

PR #586, opened 2016-01-06
Making ScheduledReporter ThreadPool daemon threads, fixed bugs in CliLocalJobLauncher

sahilTakiar:
* Created a new method `ExecutorUtils.newDaemonThreadFactory` that produces daemon threads
* Fixed bug in `CliLocalJobLauncher` where `LocalJobLauncher` wasn't being closed, causing a few other thread pools to not be shut down
* Fix to `ScheduledReporter.stopImpl` so that when the shutdown hook is invoked, the thread pool is first shutdown, and then one last invocation of `report()` is done
* Test locally, `RootMetricContext`'s `shutdownHook` is invoked, and the process terminates gracefully

@ibuenros and @pcadabam can you review?


pcadabam:
LGTM

ibuenros:
LGTM

PR #588, opened 2016-01-06
Hack to fix POM dependencies on gobblin-rest-api

chavdar:
I suspect that there is a bug in Gradle where dependencies on non-default configurations are not handled correctly. 

This hack rewrites the POM files for libraries which depend on the dataTemplate and restClient configurations of gobblin-rest-api to use the correct artifacts.


chavdar:
@liyinan926 can you review?

liyinan926:
LGTM.

PR #596, opened 2016-01-07
A small code fix to add a missing type parameter

liyinan926:


liyinan926:
Really simple fix, self approved.

PR #599, opened 2016-01-08
Added cloudera repository.

ibuenros:


chavdar:
Why is that needed?

ibuenros:
Users that try to compile with cdh releases get compilation errors, and a fix has been going around the user list involving adding the cloudera repository to all gradle projects.
This PR fixes the build for cdh releases.

chavdar:
OK.

PR #602, opened 2016-01-09
Gobblin Dashboard for jobs

eogren:
- Add a UI that queries the JobExecutionInfoServer for various job metrics and displays it in an easy to read format. If `admin.server.enabled` is set to true, this will be served on port 8000
 - A Cli tool is also provided, which can easily be run with the bin/gobblin-admin.sh script
 - Few changes to the JobExecutionInfoServer to retrieve the data we want

Thanks to @ahollenbach for doing most of the work!


ahollenbach:
Good changes 😊👍

chavdar:
@abti and @ibuenros Can you please review?

ibuenros:
@eogren The PR is quite large. Any suggestion on the order in which we should read the new classes? A start point? Thanks!

eogren:
Hey @ibuenros - sure, and sorry for the gigantic commit. It was tough to squash it into a bunch of smaller ones.

I would recommend starting with the gobblin-rest-service/ and gobblin-metastore/ changes which are the API changes necessary to power the dashboard.

Next, optionally, is probably gobblin-admin/.../Cli.java to get a sense of some of the queries we may be exposing.

The web files in gobblin-admin/ next - although the web UI is really just making calls to the Gobblin REST server and turning them into tables and fancy graphs.

Last is probably the changes to SchedulerDaemon and the AdminWebServer -- these just start an embedded Jetty server that serves all the web stuff to make it easier for people to get started with the dashboard.

Hope that helps

gobblin-admin/src/main/java/gobblin/cli/Cli.java
```diff
@@ -0,0 +1,567 @@
...
+
+/**
+ * A command line interface for interacting with Gobblin.
+ * From this tool, you should be able to:
+ *  * Check the status of Gobblin jobs
+ *  * View ...
+ *
+ * @author ahollenbach@nerdwallet.com
+ */
+public class Cli {
```
ibuenros: This class is too long atm. Can you split it? Suggestion: Have classes `JobCommand`, and a class `TaskCommand`.

gobblin-admin/src/main/java/gobblin/cli/Cli.java
```diff
@@ -0,0 +1,567 @@
...
+  private static enum Command {
```
ibuenros: Suggestion for alternative way of doing this:

```java
  private static enum Command {
    JOBS(JobCommand.getClass()), 
    TASKS(TaskCommand.getClass());

    private Class<? extends Command> commandClass;

    Command(Class<? extends Command> commandClass) {
      this.commandClass = commandClass;
    }

    public Command getCommandObject(CommandLine parsedOpts) {
      return ConstructorUtils.invokeConstructor(this.commandClass, parsedOpts);
    }
  }
```

then in `execute()` you can do:
```
Command.valueOf(this.command.toUpperCase()).getCommandObject(parsedOpts).execute();
```

IMO it makes it easier to add more commands in the future.

I would also move `Command` enum to a separate top level class.

gobblin-admin/src/main/java/gobblin/cli/Cli.java
```diff
@@ -0,0 +1,567 @@
...
+
+  private void executeTaskCommand() {
+    throw new NotImplementedException("Tasks lookup not yet implemented!");
+  }
+
+  private boolean shouldExecuteCommand(Command cmd) {
+    return this.command.equals(cmd.name().toLowerCase());
+  }
+
+  private JobExecutionInfo queryByJobId(String id, boolean includeProperties) {
```
ibuenros: Is it intentional that you are not using `includeJobProperties`?

gobblin-admin/src/main/java/gobblin/cli/Cli.java
```diff
@@ -0,0 +1,567 @@
...
+    new CliTablePrinter.Builder()
+        .labels(labels)
+        .data(data)
+        .flags(flags)
+        .delimiterWidth(2)
+        .build()
+        .printTable();
+  }
+
+  private JobExecutionInfo queryJobProperties(String id) {
```
ibuenros: This method is never used, but may be related to the above comment.

gobblin-admin/src/main/java/gobblin/cli/CliTablePrinter.java
```diff
@@ -0,0 +1,194 @@
...
+public class CliTablePrinter {
+  /**
+   * Labels for each columns
+   */
+  private List<String> labels;
+
+  /**
+   * A list of flag strings (corresponding to each column)
+   */
+  private List<String> flags;
```
ibuenros: Can you explain a bit more what flags mean? I see now they are the flags used for String formatting, but it wasn't clear at the beginning.

eogren:
Sorry for the delay - I split the Cli stuff into a bunch of smaller classes using a similar (although not exactly the same) method as proposed. Unfortunately it meant basically rewriting the Cli class so a lot of the original context may be lost.

ibuenros:
LGTM

ibuenros:
Can you make a wiki entry on the admin server?

ibuenros:
@abti can you take a final look at the PR before merging? Thanks.

abti:
@ibuenros sure I will go through it later today

PR #604, opened 2016-01-13
Fix the distribution tarball creation

kadaan:
Ensure that a gobblin-dist.tar.gz is created at the end of the build.

### DIFF
--- a/gobblin-distribution/build.gradle
+++ b/gobblin-distribution/build.gradle
@@ -44,9 +44,9 @@ task build(type: Tar, overwrite: true) {
 
   doLast {
     copy {
-      from buildDir.path + '/distributions/gobblin-distribution.tar.gz'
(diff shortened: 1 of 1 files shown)


chavdar:
LGTM

sahilTakiar:
@kadaan can you explain the nature of this PR and why it is necessary? I noticed today that the creation of the `gobblin-dist.tar.gz` is broken. I manually deleted the `gobblin-dist.tar.gz` and when I built the project again a new `gobblin-dist.tar.gz` was not created.

kadaan:
@sahilTakiar When building locally I noticed that the gobblin-dist.tar.gz was not created.  I checked the gobblin/build/gobblin-distribution/distributions folder and noticed that the tarball was there, except it was named: gobblin-distribution-0.6.0-213-gc7ab443.tar.gz.  I changed the code as shown in the PR and the tarball started showing up in the correct location.  I'm not sure what might have changed to cause this to break.

PR #606, opened 2016-01-13
Hive Registration initial commit

zliu41:
This is the initial commit the outlines the approach.

`HiveRegister` is the main class. It registers a `HiveRegistrable` using a `HiveRegistrationPolicy` which specifies how the registration should be done, and a `HiveStorageDescriptorManager` to manage the storage. `HiveStorageDescriptorManager` uses a `HiveSchemaManager` to manage the schema.


zliu41:
@sahilTakiar @ydai1124 @vasanthrajamani can you review?

sahilTakiar:
Can you add markers to all the interfaces (https://github.com/linkedin/gobblin/pull/551)

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+        } catch (NoSuchObjectException e) {
+          // Partition does not exist. Nothing to do
+        }
+      }
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private void addPartition(Table table, HivePartition partition, StorageDescriptor sd) throws TException {
```
sahilTakiar: I believe in the old code base when you add a partition you also need to check if the partition contains a new schema, and if so you need to alter the table to reflect the new schema, will that still be possible with this approach?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationPolicy.java
```diff
@@ -0,0 +1,82 @@
...
+  public String getTableName(HiveRegistrable registrable);
+
+  /**
+   * Location of the table corresponding to the given {@link HiveRegistrable}.
+   *
+   * If the given {@link HiveRegistrable} represents a table, this method should return
+   * {@link HiveRegistrable#getPath()}. If it represents a partition, this method should return
+   * the location of the table where this partition belongs to.
+   */
+  public String getTableLocataion(HiveRegistrable registrable);
```
chavdar: s/getTableLocataion/getTableLocation/

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationPolicy.java
```diff
@@ -0,0 +1,82 @@
...
+
+import com.google.common.base.Optional;
+
+
+/**
+ * An interface for specifying how to register a {@link HiveRegistrable}.
+ *
+ * @author ziliu
+ */
+public interface HiveRegistrationPolicy {
```
chavdar: Please mark all interfaces as @Alpha

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+        Executors.newFixedThreadPool(this.state.getPropAsInt(HIVE_REGISTER_THREADS, DEFAULT_HIVE_REGISTER_THREADS),
+            ExecutorsUtils.newThreadFactory(Optional.of(log), Optional.of("HiveRegister")));
+  }
+
+  /**
+   * Register a {@link HiveRegistrable}, which can be a table or a partition.
+   *
+   * This method is asynchronous and returns immediately.
+   */
+  public void register(final HiveRegistrable registrable) throws IOException {
```
chavdar: This should return a ListenableFuture. Otherwise, there is no way to know if the registration succeeded or not.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    Preconditions.checkArgument(state.contains(HIVE_OWNER), "Missing required property " + HIVE_OWNER);
+
+    this.state = state;
+    this.hiveDbRootDir = this.state.getProp(HIVE_DB_ROOT_DIR);
+    this.hiveOwner = this.state.getProp(HIVE_OWNER);
+    this.partitionManager = partitionManager;
+    this.sdManager = sdManager;
+    this.client = new HiveMetaStoreClient(new HiveConf());
+    this.executor =
+        Executors.newFixedThreadPool(this.state.getPropAsInt(HIVE_REGISTER_THREADS, DEFAULT_HIVE_REGISTER_THREADS),
```
chavdar: We should be using a ScalingThreadPoolExecutor instead. There is no need to have all threads sitting idly.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: You're right, will add this.
btw the old code doesn't update the schema if "the schema size was large and still is large", do you know why?
(https://gitli.corp.linkedin.com/ds-platform/hive-registration/source/032742da3c602d66cfde7670765658d18d1ab4b2:hive-registration-impl/src/main/java/com/linkedin/etl/hive/KafkaHiveHourlyRegister.java#L169)

sahilTakiar: By default, the old code based would register the Avro schema using `avro.schema.literal` (the value would be the actual schema), but only if the schema did not exceed a certain size. If it exceeded a size it would register the schema using `avro.schema.url` and the value would be a URI pointing to a file (typically on HDFS), that contains the schema.

The reason this was put in is that Hive would choke of the `avro.schema.literal` was too long, and the table would be un-querable (I believe Identity.Profile hit this issue); which is why this feature was added.

However, that was a while ago, so I'm not sure if that bug is still in Hive.

zliu41: Got it. Actually the Avro schema will be managed by `HiveAvroSchemaManager` which is not part of this PR.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+        if (existsAny(dbName, HiveRegister.this.partitionManager.getDominatingTableNames(registrable))
+            || existsAny(dbName, tableName, HiveRegister.this.partitionManager.getDominatingPartitions(registrable))) {
+          return null;
+        }
+
+        createDbIfNotExists(dbName);
+
+        Optional<HivePartition> partition = HiveRegister.this.partitionManager.getPartition(registrable);
+
+        Table table = getOrCreateTable(dbName, tableName, partition, HiveRegister.this.sdManager
```
ydai1124: Do you need to call `this.client.createTable` for the returned `Table`?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+
+  private Table getOrCreateTable(String dbName, String tableName, Optional<HivePartition> partition,
+      StorageDescriptor sd) {
+    Lock lock = locks.get(dbName + tableName);
+    lock.lock();
+
+    try {
+      return this.client.getTable(dbName, tableName);
+    } catch (NoSuchObjectException e) {
+      return createTable(dbName, tableName, partition, sd);
```
ydai1124: Do you need to call `this.client.createTable` for the returned `Table`?

gobblin-hive-registration/src/main/java/gobblin/hive/HivePartition.java
```diff
@@ -0,0 +1,54 @@
...
+    this.values = ImmutableList.copyOf(values);
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < keys.size(); i++) {
+      sb.append(keys.get(i).getName());
+      sb.append("=");
+      sb.append(values.get(i));
```
ydai1124: Better to have a separator, e.g., ";" after each value is appended.

alpha:
@chavdar  Sorry for the offtopic, but I have acquired the username @alpha as an inactive username, and I am most likely not the person you're looking for. Really sorry again for the inconvenience

PR #616, opened 2016-01-19
Distcp persists and reuses failed work units.

ibuenros:
This PR adds a feature to Gobblin distcp to persist and recover late-failing work units (e.g. can't set permissions).

Change summary:
* Added a guid to `CopyableFile`.
* `CopySource` adds a work unit guid to each work unit.
* `FileAwareInputStream` catches late failures and persists work units. To do this, it moves copied files to a persistent location, identified by guid and a shortened version of their original path. Furthermore, for each work unit, the writer will search the persistent location for matching files, if one is found, it is recovered instead of doing the copy.
* Created `DistcpConverter` as a superclass for all distcp converters.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.copy.converter;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.hadoop.fs.FSDataInputStream;
+
+import com.google.common.base.Function;
+
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.Converter;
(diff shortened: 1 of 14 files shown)


ibuenros:
@pcadabam can you review?

pcadabam:
Overall the approach looks good to me. I have a few comments.
1. Can you extract the persistence and recovery logic into its own class? The writer class seems heavy.
2. How is recovery managed for subclasses of FileAwareInputStreamDataWriter? For example TarArchiveInputStreamDataWriter overrides the write method to extract and write a tar ball file.
3. The Sha computation logic can be moved to static util class/methods

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
```diff
@@ -0,0 +1,98 @@
...
+import gobblin.data.management.copy.CopyableFile;
+import gobblin.data.management.copy.FileAwareInputStream;
+import gobblin.util.PathUtils;
+
+
+/**
+ * Abstract class for distcp {@link Converter}. Simply transforms the {@link InputStream} in the
+ * {@link FileAwareInputStream}, and possibly modifies extensions of the output file.
+ */
+public abstract class DistcpConverter extends Converter<String, String, FileAwareInputStream, FileAwareInputStream> {
```
pcadabam: Why is this refactor required?

ibuenros: Not required. Reduces code repetition (it's there because at some point in the development of the feature it was required, and I saw no reason to remove it).

ibuenros:
Thanks @pcadabam 
1. Will do.
2. The subclass would need to do a similar implementation. Since I'm refactoring methods to a different class, it can use those same methods. I will do the changes to the untar writer as well, but in a different PR. (untar writer should persist the entire generated directory).
3. will do.

ibuenros:
@pcadabam made requested changes:
* Created a class `RecoveryHelper` containing all the logic for persisting and recovery.
* Created a class `Guid` containing methods for creating and handling guids.

pcadabam:
LGTM other than the small comment.
- Add some unit tests for RecoveryHelper

PR #617, opened 2016-01-19
Fix distribution creation for both install and upload

kadaan:
Make the distribution build script detect the `uploadArchives` target so that it can know the correct name for the distribution tarball.

As discovered by @sahilTakiar after #604 was merged, the `install` and `uploadArchives` targets result in differently named tarballs.  Prior to #604, when the `uploadArchives` target was specified the tarball would not be copied to gobblin root.  Post #604, when the `install` target is specified the tarball will not be copied to gobblin root.  This PR makes to work either way.


sahilTakiar:
LGTM

PR #624, opened 2016-01-21
Close JDBC connections after executing sqls

narasimhareddyv:
JDBC connections are not getting closed as per the existing code. So a lot of connections are opened at the source end(RightNow) and the mysql database is down.

So I made a small change in JDBCExtractor to close connections in the finally block. This is the only change. But you may see small alignment differences in the diff.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
@@ -12,27 +12,6 @@
 
 package gobblin.source.extractor.extract.jdbc;
 
-import gobblin.configuration.ConfigurationKeys;
-import gobblin.configuration.WorkUnitState;
-import gobblin.password.PasswordManager;
-import gobblin.source.extractor.DataRecordException;
-import gobblin.source.extractor.exception.HighWatermarkException;
-import gobblin.source.extractor.exception.RecordCountException;
-import gobblin.source.extractor.exception.SchemaException;
-import gobblin.source.extractor.extract.Command;
-import gobblin.source.extractor.extract.CommandOutput;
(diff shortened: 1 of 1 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -12,6 +12,27 @@
 
 package gobblin.source.extractor.extract.jdbc;
 
+import gobblin.configuration.ConfigurationKeys;
```
ydai1124: Can you revert back the import order (https://github.com/linkedin/gobblin/wiki/CodingStyle)?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -627,6 +630,10 @@ private void parseInputQuery(String query) {
       resultSet = statement.getResultSet();
     } catch (Exception e) {
       log.error("Failed to execute sql:" + query + " ;error-" + e.getMessage(), e);
+    } finally {
+      if (connection != null) {
```
ydai1124: Can you catch the SQLException and add a log during closing, instead of throw it?

narasimhareddyv: Updated the pull request. Can you review it.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
narasimhareddyv: Updated the pull request. Can you review it.

ydai1124:
LGTM.

PR #626, opened 2016-01-22
Distcp can persist work units on failed writer commit.

ibuenros:


ibuenros:

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -66,6 +68,7 @@
   protected final Closer closer = Closer.create();
   protected CopyableDatasetMetadata copyableDatasetMetadata;
   protected final RecoveryHelper recoveryHelper;
+  protected Optional<CopyableFile> actualProcessedCopyableFile;
```
pcadabam: add some comment as to what this means

pcadabam:
LGTM with that minor comment

PR #627, opened 2016-01-22
client API/impl check in

tuGithub:


tuGithub:
@chavdar , @ibuenros 
please review it when you have time.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+   * key1: "etl-hdfs:/"
+   * key2: "etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest/"
+   * 
+   */
+  private final TreeMap<URI, ConfigStoreAccessor> configStoreAccessorMap = new TreeMap<>();
+  
+  private final ConfigStoreFactoryRegister configStoreFactoryRegister;
+
+  private ConfigClient(VersionStabilityPolicy policy) {
+    this.policy = policy;
```
ibuenros: Can you make this call `this(policy, new ConfiStoreFactoryRegister())` instead? It makes for easier maintanability in the future.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+   * @throws ConfigStoreCreationException
+   * @throws VersionDoesNotExistException
+   */
+  public Map<URI, Config> getConfigs(Collection<URI> configKeyUris)throws ConfigStoreFactoryDoesNotExistsException,
+  ConfigStoreCreationException, VersionDoesNotExistException {
+    if(configKeyUris == null || configKeyUris.size()==0 )
+      return Collections.emptyMap();
+    
+    Map<URI, Config> result = new HashMap<>();
+    Map<ConfigStoreAccessor, Collection<ConfigKeyPath>> partitionedAccessor = new HashMap<>();
```
ibuenros: Suggestion: use multimap instead

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+        return floorKey;
+      }
+      
+      if(floorKey.getAuthority()==null || configKeyURI.getAuthority()==null){
+        return null;
+      }
+      
+      // both are absolute URI
+      if(floorKey.getAuthority().equals(configKeyURI.getAuthority())){
+        if(configKeyURI.getPath().startsWith(floorKey.getPath())) {
```
ibuenros: `startsWith` is not enough. `"/data/databases".startsWith("/dat")` is `true`, but that is not a valid match. You need to check whether `floorKey` is an ancestor of `configKeyURI`.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+  ConfigStoreCreationException, VersionDoesNotExistException{
+    
+    ConfigStoreAccessor result;
+    ConfigStoreFactory<ConfigStore> csFactory = this.getConfigStoreFactory(configKeyURI);
+    ConfigStore cs = csFactory.createConfigStore(configKeyURI);
+
+    if (!(cs instanceof ConfigStoreWithStableVersioning)) {
+      if (this.policy == VersionStabilityPolicy.CROSS_JVM_STABILITY) {
+        throw new RuntimeException(String.format(
+            "with policy set to %s, connect not connect to unstable config store %s",
```
ibuenros: The message is weird: "connect not connect"

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+      inMemoryValueInspector = new InMemoryValueInspector(rawValueInspector, false);
+      result = new ConfigStoreAccessor(cs, inMemoryValueInspector, inMemoryTopology);
+    }
+    // Non ConfigStoreWithStableVersioning but require STRONG_LOCAL_STABILITY, use Strong reference cache
+    else if (this.policy == VersionStabilityPolicy.STRONG_LOCAL_STABILITY) {
+      inMemoryValueInspector = new InMemoryValueInspector(rawValueInspector, true);
+      result = new ConfigStoreAccessor(cs, inMemoryValueInspector, inMemoryTopology);
+    }
+    // Require No cache
+    else {
```
ibuenros: What about cs that is not `ConfigStoreWithStableVersioning` but policy is `WEAK_LOCAL_STABILITY`?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+    if(matchedFloorKey!=null){
+      result = this.configStoreAccessorMap.get(matchedFloorKey);
+      return result;
+    }
+    
+    result = createNewConfigStoreAccessor(configKeyURI);
+    ConfigStore cs = result.configStore;
+    
+    // put to cache
+    this.configStoreAccessorMap.put(cs.getStoreURI(), result);
```
ibuenros: What happens if `cs.getStoreURI()` is not an ancestor of `configKeyURI`? Then repeatedly calling `getConfig(configKeyURI)` would instantiate a new store every time.
Our contract should probably by that `createNewConfigStoreAccessor(configKeyURI).configStore.getStoreURI()` must be an ancestor of `configKeyURI`, and explicitly check for this condition on the code.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,352 @@
...
+    
+    // put default root URI in cache as well for the URI which missing authority 
+    if(configKeyURI.getAuthority() == null){
+      this.configStoreAccessorMap.put(ConfigClientUtils.getDefaultRootURI(configKeyURI, cs), result);
```
ibuenros: It seems the contract here is that for empty authority, the factory should always return the same `ConfigStore`.
If that is the case, then it is clearer to put a method in the factory `getConfigStoreWithMissingAuthority` and use that method in this case.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+
+import com.google.common.base.Preconditions;
+
+import gobblin.config.common.impl.SingleLinkedListConfigKeyPath;
+import gobblin.config.store.api.ConfigKeyPath;
+import gobblin.config.store.api.ConfigStore;
+
+public class ConfigClientUtils {
```
ibuenros: Can you put javadoc?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,137 @@
...
+   * 
+   * @param configKeyURI
+   * @param cs
+   * @return
+   */
+  public static URI getDefaultRootURI(URI configKeyURI, ConfigStore cs){
+    preconditionCheck(configKeyURI, cs);
+    
+    if(configKeyURI.getAuthority()!=null){
+      return cs.getStoreURI();
```
ibuenros: Isn't the assumption that the authority is missing? Shouldn't this error out instead?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,137 @@
...
+  public static Collection<URI> getAbsoluteURI(Collection<ConfigKeyPath> configKeyPaths, ConfigStore cs){
+    Collection<URI> result = new ArrayList<>();
+    for(ConfigKeyPath p: configKeyPaths){
+      result.add(getAbsoluteURI(p, cs));
+    }
+    
+    return result;
+  }
+  
+  private static void preconditionCheck(URI configKeyURI, ConfigStore cs){
```
ibuenros: Can you have a clearer name on what this is checking? Like `checkMatchingSchemeAndAuthority`?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
```diff
@@ -0,0 +1,44 @@
...
+
+import gobblin.config.store.api.ConfigStoreFactory;
+
+
+public class ConfigStoreFactoryRegister {
+  private static final Logger LOG = Logger.getLogger(ConfigStoreFactoryRegister.class);
+  
+  //key is the configStore scheme name, value is the ConfigStoreFactory
+  @SuppressWarnings("rawtypes")
+  private final Map<String, ConfigStoreFactory> configStoreFactoryMap = new HashMap<>();
```
ibuenros: Make this an immutable map. (Helps with thread-safety)

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: Wait for the API changes of ConfigStore

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: Wait for the API changes of ConfigStore

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: done

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: Wait for the API changes of ConfigStore

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: done

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
tuGithub: done

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: removed this function

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,350 @@
...
+      ConfigStoreAccessor accessor = this.getConfigStoreAccessor(u);
+      ConfigKeyPath configKeypath = ConfigClientUtils.buildConfigKeyPath(u, accessor.configStore);
+      partitionedAccessor.put(accessor, configKeypath);
+    }
+    
+    for(Map.Entry<ConfigStoreAccessor, Collection<ConfigKeyPath>> entry: partitionedAccessor.asMap().entrySet()){
+      Map<ConfigKeyPath, Config> batchResult= entry.getKey().valueInspector.getResolvedConfigs(entry.getValue());
+      // translate the ConfigKeyPath to URI
+      for(Map.Entry<ConfigKeyPath, Config> resultEntry: batchResult.entrySet()){
+        URI absURI = ConfigClientUtils.getAbsoluteURI(resultEntry.getKey(), entry.getKey().configStore);
```
chavdar: This conversion to an absolute URI in the result may confuse the users. It may be hard to map for them to map from a relative URI that they have passed to an absolute one. We should preserve the key URIs.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,174 @@
...
+    
+    if(!stringSame(descendant.getScheme(), ancestor.getScheme())){
+      return false;
+    }
+    
+    if(!stringSame(descendant.getAuthority(), ancestor.getAuthority())){
+      return false;
+    }
+    
+    return isAncestorOrSame(new Path(descendant.getPath()), new Path(ancestor.getPath()));
```
chavdar: We should be using ConfigKeyPath here.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
```diff
@@ -0,0 +1,49 @@
...
+  //key is the configStore scheme name, value is the ConfigStoreFactory
+  @SuppressWarnings("rawtypes")
+  private final ImmutableMap<String, ConfigStoreFactory> configStoreFactoryMap ;
+
+  @SuppressWarnings("rawtypes")
+  public ConfigStoreFactoryRegister() {
+    ServiceLoader<ConfigStoreFactory> loader = ServiceLoader.load(ConfigStoreFactory.class);
+    ImmutableMap.Builder<String, ConfigStoreFactory> builder = new ImmutableMap.Builder<>();
+    for (ConfigStoreFactory f : loader) {
+      builder.put(f.getScheme(), f);
```
chavdar: I suggest that you abstract this away as a public register() method so users can also programatically register new factories.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,174 @@
...
+    for(ConfigKeyPath p: configKeyPaths){
+      result.add(getAbsoluteURI(p, cs));
+    }
+    
+    return result;
+  }
+  
+  private static void checkMatchingSchemeAndAuthority(URI configKeyURI, ConfigStore cs){
+    if(configKeyURI == null || cs == null){
+      throw new IllegalArgumentException("input can not be null");
```
chavdar: nit: I suggest that all checks are either Preconditions or all throw IllegalArgumentException.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,350 @@
...
+    return ConfigClientUtils.getAbsoluteURI(result, accessor.configStore);
+  }
+  
+  private URI getMatchedFloorKeyFromCache(URI configKeyURI){
+    URI floorKey = this.configStoreAccessorMap.floorKey(configKeyURI);
+    if(floorKey==null) {
+      return null;
+    }
+    
+    // both scheme name and authority name should match
```
chavdar: This is not correct. If the authority of configKeyURI is null, then only the scheme should match.

gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
```diff
@@ -0,0 +1,209 @@
...
+import gobblin.config.client.api.VersionStabilityPolicy;
+import gobblin.config.common.impl.ConfigStoreValueInspector;
+import gobblin.config.common.impl.SingleLinkedListConfigKeyPath;
+import gobblin.config.store.api.ConfigKeyPath;
+import gobblin.config.store.api.ConfigStore;
+import gobblin.config.store.api.ConfigStoreFactory;
+
+@Test(groups = { "gobblin.config.common.impl" })
+
+public class TestConfigClient {
```
chavdar: It is very hard to figure out what's going on here. For example, it is hard to figure out what are the config keys and values in the mock store. Based on looking at the implementation, I am not sure the handling of relative config keys is correct but I can't figure out how the test works.

I think it would have been useful to have a simple store implementation which is just a map from a config key to a Properties/Config.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: changed the comments

gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
tuGithub: mmm. I can change the unit testing, but I am not sure why you say the relative config key handling is not correct. Let us talk

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
```diff
@@ -0,0 +1,176 @@
...
+   * 2. etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest/datasets/a1/a2 which has the authority
+   * @param cs            : the config store of the input configKeyURI
+   * @return              : return the URI of the same format with the input configKeyURI
+   * 
+   * for example, configKeyPath is /tags/retention, 
+   * with configKeyURI as etl-hdfs:///datasets/a1/a2, return "etl-hdfs:///tags/retention
+   * with configKeyURI as etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest , then return
+   * etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest/tags/retention
+   */
+  public static URI buildURI(ConfigKeyPath configKeyPath, URI configKeyURI, ConfigStore cs){
```
chavdar: I don't understand what this method is supposed to do. Looking at the two examples: in the first case, the path replaces the path in the URI; in the second case, the path is appended. Why do those two cases have a completely different treatment?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
```diff
@@ -0,0 +1,360 @@
...
+      if(!inputURI.containsKey(accessor)){
+        inputURI.put(accessor, u);
+      }
+    }
+    
+    for(Map.Entry<ConfigStoreAccessor, Collection<ConfigKeyPath>> entry: partitionedAccessor.asMap().entrySet()){
+      Map<ConfigKeyPath, Config> batchResult= entry.getKey().valueInspector.getResolvedConfigs(entry.getValue());
+      // translate the ConfigKeyPath to URI
+      for(Map.Entry<ConfigKeyPath, Config> resultEntry: batchResult.entrySet()){
+        URI formattedUri = ConfigClientUtils.buildURI(resultEntry.getKey(), inputURI.get(entry.getKey()),
```
chavdar: See my comment below about. I don't understand at a high level what buildURI() is supposed to do. This method seems to be overly complicated. I think it is OK to say that getConfigs() works only for configs coming from the same store/accessor. This will simplify the implementation significantly. You can just keep a reverse map from a ConfigKeyPath to a ConfigKeyURI that was used to generate it. Then populating the result is done using this map.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @SuppressWarnings("rawtypes")
+  public ConfigStoreFactory getConfigStoreFactory(String scheme){
+    return configStoreFactoryMap.get(scheme);
+  }
+  
+  @SuppressWarnings("rawtypes")
+  public void register(String scheme, ConfigStoreFactory factory){
+    this.configStoreFactoryMap.put(scheme, factory);
```
chavdar: nit: it may be useful for debugging to log the schemes/factories that are registered.

gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
chavdar: Nevermind. I think I figured out my concern about handling relative config keys.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientUtils.java
tuGithub: changed to NOT use that method in getConfigs()

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClient.java
tuGithub: fixed

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigStoreFactoryRegister.java
tuGithub: fixed

chavdar:
LGTM

chavdar:
@tuGithub This is complete, right?

abti:
@vasanthrajamani Please assign and priortize

abti:
https://issues.apache.org/jira/browse/GOBBLIN-146

Please update your PR title with following prefix: [GOBBLIN-146]

PR #629, opened 2016-01-24
Gobblin_password_encryptor should be included in the tarball

lbendig:
As all the jars needed by the password encryptor cli tool are in the ```gobblin-dist/lib``` directory,
it's not necessary to build ```gobblin-utility.tar.gz``` manually as suggested in the [documentation]
(https://github.com/linkedin/gobblin/wiki/Working-with-Job-Configuration-Files#password-encryption) .

This patch copies the shell script to gobblin-dist/bin. Also added a null check to the password
decryptor to fix the NPE in case of unset passwords.


chavdar:
LGTM

lbendig:
Thank you, Chavdar!

PR #630, opened 2016-01-25
Gobblin Oozie config files

sahilTakiar:
Hackday project: Added example configuration files for running Gobblin on Oozie
* Added a few changes to `CliOptions.java` so that it can read configuration files from hdfs, not just the local filesystem - this is necessary for Oozie, as it reads its configuration files from HDFS
* Added configuration files under `gobblin-oozie/src/test/resources`
* The config files run Gobblin in local mode in a Java process
* `gobblin-oozie-example-system.properties` is an example sysconfig for running Gobblin on Oozie
* `gobblin-oozie-example-workflow.properties` is the job config file that Oozie accepts
* `gobblin-oozie-example-workflow.xml` is the Oozie workflow that launches Gobblin as a Java process

Documentation for these configs and a guide on how to run Gobblin on Oozie is here: https://github.com/linkedin/gobblin/wiki/Gobblin-Schedulers

Tested using the Hortonworks Sandbox on VirtualBox.

### DIFF
--- a/gobblin-oozie/src/test/resources/local/gobblin-oozie-example-workflow.properties
+++ b/gobblin-oozie/src/test/resources/local/gobblin-oozie-example-workflow.properties
@@ -1,0 +1,12 @@
+# An example job.properties file for launching Gobblin jobs via Oozie
+
+# Set the Name Node URI e.g. hdfs://sandbox.hortonworks.com:8020
+name.node=hdfs://localhost:62705
+
+# Set the Resource Manager URI e.g. sandbox.hortonworks.com:8050
+resource.manager=localhost:54580
+
+nameNode=${name.node}
+jobTracker=${resource.manager}
+
+oozie.wf.application.path=${name.node}/path/to/oozie/workflows
--- a/gobblin-oozie/src/test/resources/local/gobblin-oozie-example-system.properties
+++ b/gobblin-oozie/src/test/resources/local/gobblin-oozie-example-system.properties
@@ -1,0 +1,40 @@
+# An example system.properties file for Gobblin, useful for when launching Gobblin on Oozie
+
+# Gobblin work dir
+gobblin.work.dir=gobblin-out
+
+# Thread pool settings for the task executor
+taskexecutor.threadpool.size=2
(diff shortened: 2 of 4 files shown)


sahilTakiar:
Has anyone used Oozie before? If so, please review.

ydai1124:
@sahilTakiar I will review. I did a POC for Gobblin on Oozie before.

gobblin-oozie/src/test/resources/local/gobblin-oozie-example-workflow.properties
```diff
@@ -0,0 +1,12 @@
+# An example job.properties file for launching Gobblin jobs via Oozie
+
+# Set the Name Node URI e.g. hdfs://sandbox.hortonworks.com:8020
+name.node=hdfs://localhost/62705
```
ydai1124: Should this be the port number `hdfs://localhost:62705`?

ydai1124:
Just a minor comment. Overall LGTM. Thank you for doing this!
I also added a comment in your wiki.

gobblin-oozie/src/test/resources/local/gobblin-oozie-example-workflow.properties
sahilTakiar: Yup, good catch will fix.

sahilTakiar:
@ydai1124 what comment did you add to the wiki? Is there something that needs to be addressed, or did you just make an edit.

lbendig:
@sahilTakiar Thanks for adding the Oozie scheduling.
I also implemented an Oozie integration in a way that
- it uses sharelib to store the gobblin-dist jars
- the workflow and coordinator definitions are generated and copied to HDFS on the fly and the job is submitted  to the Oozie server from code.
So essentially one only needs to add the scheduling properties to the job configuration and submit it as a gobblin-oozie job.

Would it make sense to contribute these add-ons to the project?

sahilTakiar:
Hey @lbendig 

Yes that would be great! If you could create a PR for your changes I can take a look sometime this week.

Thanks

--Sahil

PR #631, opened 2016-01-25
Client side logging in MR mode

lbendig:
Client side logging in MR mode

Currently when a job is submitted through ```gobblin-mapreduce.sh``` the client side logging is not enabled. This patch sets up logging using similar command line options as in the standalone mode.

Remark: [PR#383](https://github.com/linkedin/gobblin/pull/383) added log4j.properties to the gobblin-runtime jar. If it is not needed, I think it would be probably better to remove it, otherwise if one needs this logging configuration then this jar should be loaded on the classpath first.


chavdar:
@ydai1124 @pcadabam Could you please review?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
```diff
@@ -41,33 +40,33 @@
...
-    final Properties jobProps = JobConfigurationUtils.combineSysAndJobProperties(this.sysConfig, this.jobConfig);
+  public int run(String[] args) throws Exception {
 
     Closer closer = Closer.create();
     try {
-      closer.register(new MRJobLauncher(jobProps, getConf())).launchJob(null);
-    } catch (JobException je) {
-      System.err.println("Failed to launch the job due to the following exception:");
-      System.err.println(je.toString());
+      try {
```
ydai1124: What is the point of using nested try catch here?

bin/gobblin-mapreduce.sh
```diff
@@ -102,7 +116,7 @@ set_user_jars "$JARS"
 LIBJARS=$USER_JARS$separator$FWDIR_LIB/gobblin-metastore.jar,$FWDIR_LIB/gobblin-metrics.jar,\
 $FWDIR_LIB/gobblin-core.jar,$FWDIR_LIB/gobblin-api.jar,$FWDIR_LIB/gobblin-utility.jar,\
 $FWDIR_LIB/guava-15.0.jar,$FWDIR_LIB/avro-1.7.7.jar,$FWDIR_LIB/metrics-core-3.1.0.jar,\
-$FWDIR_LIB/gson-2.3.1.jar,$FWDIR_LIB/joda-time-2.9.jar,$FWDIR_LIB/data-1.15.9.jar
+$FWDIR_LIB/gson-2.3.1.jar,$FWDIR_LIB/joda-time-2.9.jar,$FWDIR_LIB/data-1.15.9.jar,$FWDIR_LIB/commons-lang3-3.4.jar
```
ydai1124: Is commons-lang3 here necessary?

ydai1124:
@lbendig Thank you for contributing! Talked to @ibuenros , I've created https://github.com/linkedin/gobblin/pull/640 to move `log4j.properties`.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
sahilTakiar: Since Gobblin is on Java 7 now you can use `try-with-resources` instead of Guava's `Closer`. This should also allow you to avoid the nested `try-catch`.

sahilTakiar:
Does `-Dlog4j.configuration=file:/path/to/user-specific/log4j.xml` not work for this?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
lbendig: @ydai1124 , @sahilTakiar Thanks for your comments. Changed code to use try-with.

bin/gobblin-mapreduce.sh
lbendig: I've remarked that several classes now use commons.lang3. E.g: JdbcExtractor.
I assume that --jars is used for passing source specific jars like jdbc drivers but commons.lang3 seems to me too common to enumerate each time.

lbendig:
@sahilTakiar Indeed. The idea with the custom logging property (gobblin.log4j.conf) was to not change the Hadoop logging (MR) properties by accident, but this is not the case. I Moved the logging configuration to ```HADOOP_CLIENT_OPTS```.

ydai1124:
LGTM . @lbendig Can you resolve the conflicts of this PR? Thanks!

lbendig:
@ydai1124 Fixed.

ydai1124:
LGTM. @sahilTakiar Any more comments?

bin/gobblin-mapreduce.sh
```diff
@@ -41,12 +42,16 @@ do
...
+      shift
+      ;;
     --workdir)
       WORK_DIR="$2"
       shift
       ;;
-    --jars)
-      JARS="$2"
+    --logdir)
+      GOBBLIN_LOG_DIR="$2"
```
sahilTakiar: `GOBBLIN_LOG_DIR` -> `LOG_DIR`

bin/gobblin-mapreduce.sh
```diff
@@ -20,6 +20,7 @@ function print_usage(){
   echo "  --jars <comma-separated list of job jars>      Job jar(s): if not set, \"$FWDIR_LIB\" is examined"
   echo "  --workdir <job work dir>                       Gobblin's base work directory: if not set, taken from \${GOBBLIN_WORK_DIR}"
   echo "  --projectversion <version>                     Gobblin version to be used. If set, overrides the distribution build version"
+  echo "  --logdir <log dir>                             Gobblin's log directory: if not set, taken from \${GOBBLIN_LOG_DIR}"
```
sahilTakiar: Can you update the description, it seems the behavior is:

1. `--log-dir` is first checked, if set this the directory used
2. `GOBBLIN_LOG_DIR` is checked second, if set this directory is used, logs are written to `currentDir/logs`
3. If neither of the above two options are specified

sahilTakiar:
Minor comments, everything else LGTM.

bin/gobblin-mapreduce.sh
lbendig: Oops. Fixed.

bin/gobblin-mapreduce.sh
lbendig: Fixed

lbendig:
@sahilTakiar I think I'd definitely need to document the recent changes in the MR launcher script
(logdir, project version) under the [Hadoop MapReduce Deployment] (https://github.com/linkedin/gobblin/wiki/Gobblin%20Deployment#hadoop-mapreduce-deployment-) section.

sahilTakiar:
I went ahead and updated the wiki page you mentioned.

lbendig:
Thank you for the review!

PR #633, opened 2016-01-25
Fixing tarball build

sahilTakiar:
@ydai can you review?

* Fixes the tarball build
* Added the version to the name of the tarball

### DIFF
--- a/gobblin-distribution/build.gradle
+++ b/gobblin-distribution/build.gradle
@@ -41,11 +41,6 @@ dependencies {
   }
 }
 
-distributionTarballName = 'gobblin-distribution'
-if (tasks.findIndexOf { task -> task.name.equals('uploadArchives') } >= 0) {
-  distributionTarballName = distributionTarballName + '-' + project.version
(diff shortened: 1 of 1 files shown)


gobblin-distribution/build.gradle
```diff
@@ -63,9 +58,8 @@ task build(type: Tar, overwrite: true) {
 
   doLast {
     copy {
-      from buildDir.path + '/distributions/' + distributionTarballName + '.tar.gz'
+      from buildDir.path + '/distributions/' + baseName + '-' + project.version + '.tar.gz'
```
ydai1124: Do we need to change the wiki page as well?

ydai1124:
One comment. LGTM.

gobblin-distribution/build.gradle
sahilTakiar: Updated the wiki pages

ydai1124: LGTM.

PR #634, opened 2016-01-26
Reusing connection and closing at the end rather than closing it imme…

narasimhareddyv:
Reusing database connection instead of closing it immediately after query execution


ydai1124:
LGTM

PR #635, opened 2016-01-26
Version of Gobblin dependencies are missing from gobblin-mapreduce.sh

lbendig:
Version of Gobblin dependencies are missing from gobblin-mapreduce.sh

During the creation of the Gobblin tarball the current project version is applied to the jars
however, the dependencies in the ```gobblin-mapreduce``` launcher script have no version.

This patch introduces a placeholer ```GOBBLIN_VERSION``` in the script which is substituted with the actual version during ```:gobblin-distribution:build```.

### DIFF
--- a/bin/gobblin-mapreduce.sh
+++ b/bin/gobblin-mapreduce.sh
@@ -4,6 +4,9 @@
 ############### Run Gobblin Jobs on Hadoop MR ################
 ##############################################################
 
+# Set during the distribution build
+GOBBLIN_VERSION=@project.version@
+
 FWDIR="$(cd `dirname $0`/..; pwd)"
 FWDIR_LIB=$FWDIR/lib
 FWDIR_CONF=$FWDIR/conf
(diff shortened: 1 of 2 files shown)


bin/gobblin-mapreduce.sh
```diff
@@ -4,6 +4,9 @@
 ############### Run Gobblin Jobs on Hadoop MR ################
 ##############################################################
 
+# Set during the distribution build
+GOBBLIN_VERSION=@project.version@
```
sahilTakiar: Does this work for you? I added this line, and `echo $GOBBLIN_VERSION` right after, and it just printed out `@project.version@`.

Could it be simper to just include `gobblin-*.jar` in the `libjars`.

chavdar: @sahilTakiar  that will be dangerous as we can't control which version will get loaded. Did you apply the build.gradle changes below ? That seems to be a standard way to replace tokens in a file: https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:filter(groovy.lang.Closure)

lbendig: Thanks for your comments.
@sahilTakiar the idea is that _during the build when the tarball is created_ the 
```gobblin-mapreduce.sh``` is not just taken from bin/ but the project version is substituted as well. So the gobblin-mapreduce.sh in the tarball should contain something like this:
GOBBLIN_VERSION="0.6.0-233-g9f7dd61".

sahilTakiar: Ah I see. That makes much more sense. In that case, is it possible to make a few more additions to `bin/gobblin-mapreduce.sh` to ensure backwards compatibility:

* `bin/gobblin-mapreduce.sh` accepts a `projectVersion` parameter
    * If specified, it uses this as the `GOBBLIN_VERSION`
    * If not specified, it uses whatever has already been set for `GOBBLIN_VERSION`
        * If a valid `GOBBLIN_VERSION` has been set the script proceeds as usual
        * If a non-valid `GOBBLIN_VERSION` has been set - e.g. if it is still `@project.version@` then it should fail saying no `projectVersion` has been set.

lbendig: @sahilTakiar thanks for your inputs, this definitely makes it more robust. Fixed.

sahilTakiar:
LGTM

sahilTakiar:
Thanks for contributing! Merging.

lbendig:
@sahilTakiar @chavdar Thanks for the review!

PR #637, opened 2016-01-27
Ensure that the output file list does not contain duplicates

kadaan:
NOTE: duplicates were being added to the list due to task retries


sahilTakiar:
LGTM

PR #638, opened 2016-01-27
Adding Simple HDFS ConfigStore

sahilTakiar:
* `SimpleHDFSConfigStore implements ConfigStore`
* `SimpleHDFSConfigStoreFactory implements ConfigStoreFactory`

A detailed explanation of both classes are in the JavaDocs.

`SimpleHDFSConfigStoreFactory` is a good starting point.

@tuGithub and @ibuenros can you review?


sahilTakiar:
Detailed documentation for the expected directory layout for `SimpleHDFSConfigStore` is documented in the Java Docs.

This implementation is heavily based on Min's original implementation: 

https://github.com/tuGithub/gobblin-1/tree/playground/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/hadoopfs/store

and his design doc:

https://docs.google.com/document/d/1CEHXVM0H65v3PhmcdKAkjZFN0fUORKFBqk6yFwhbK3s/edit#heading=h.baom2n5pep5l (internal LinkedIn doc)

gobblin-config-management/gobblin-config-core/build.gradle
```diff
@@ -12,10 +12,13 @@ apply plugin: 'java'
 
 dependencies {
     compile project(':gobblin-api')
```
ibuenros: Unrelated, but why does config-core require gobblin-api?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+      if (!this.fs.exists(configStoreDir)) {
+        throw new VersionDoesNotExistException(getStoreURI(), "CURRENT", "");
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(
+          String.format("Error while checking if the configStoreDir: \"%s\" exists", configStoreDir), e);
+    }
+
+    try {
+      for (FileStatus status : this.fs.listStatus(configStoreDir)) {
```
ibuenros: I think we should explicitly sort the paths lexicographically. The order of listStatus is not part of the contract, and different file systems might have different orderings.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+    this.fs = fs;
+
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(physicalStoreRoot.getScheme()),
+        "The physicalStoreRoot must have a valid scheme!");
+    Preconditions.checkArgument(physicalStoreRoot.getScheme().equals(fs.getUri().getScheme()),
+        "The scheme of the physicalStoreRoot and the filesystem must match!");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(physicalStoreRoot.getPath()),
+        "The path of the physicalStoreRoot must be valid as it is the root of the store!");
+
+    this.physicalStoreRoot = physicalStoreRoot;
```
ibuenros: Say the user does `getConfig("simple-hdfs://mycluster/mystore/myconfig")`, what will the value of `physicalStoreRoot` be? (It seems that it will be `hdfs://mycluster/mystore`?)
What will the return value of `getStoreRoot()` be? Per contract with the config library, it should be `simple-hdfs://mycluster/mystore`, but I don't believe that's what it will return.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+   * @throws VersionDoesNotExistException if the version specified cannot be found in the {@link ConfigStore}.
+   */
+  @Override
+  public Config getOwnConfig(ConfigKeyPath configKey, String version)
+      throws VersionDoesNotExistException {
+    Preconditions.checkNotNull(configKey, "configKey cannot be null!");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(version), "version cannot be null or empty!");
+
+    Path configDir = getDatasetDirForKey(configKey, version);
+    checkDatasetDirExists(configKey, configDir);
```
ibuenros: I'm not sure this check should be done. Querying for `/foo/bar` is valid even if `/foo/bar` doesn't exist, in that case you should just return empty?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+   */
+  @Override
+  public List<ConfigKeyPath> getOwnImports(ConfigKeyPath configKey, String version)
+      throws VersionDoesNotExistException {
+    Preconditions.checkNotNull(configKey, "configKey cannot be null!");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(version), "version cannot be null or empty!");
+
+    List<ConfigKeyPath> configKeyPaths = new ArrayList<>();
+    Path configDir = getDatasetDirForKey(configKey, version);
+    checkDatasetDirExists(configKey, configDir);
```
ibuenros: This might have a similar issue as my comment in `getOwnConfig`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+
+    try {
+      for (FileStatus fileStatus : this.fs.listStatus(configDir, new IncludesConfFilterFilter())) {
```
ibuenros: Why not just do `fs.getFileStatus(new Path(configDir, "includes.conf"))`?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(version), "version cannot be null or empty!");
+
+    Path configDir = getDatasetDirForKey(configKey, version);
+    checkDatasetDirExists(configKey, configDir);
+
+    try {
+      for (FileStatus fileStatus : this.fs.listStatus(configDir, new MainConfFileFilter())) {
+        if (!fileStatus.isDir()) {
+          try (InputStream mainConfInputStream = this.fs.open(fileStatus.getPath())) {
+            return ConfigFactory.parseReader(new InputStreamReader(mainConfInputStream)).resolve();
```
ibuenros: You should not be resolving here, resolution is always lazy.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+
+  /**
+   * Retrieves the name of a dataset from a given {@link ConfigKeyPath}.
+   */
+  private Path getDatasetFromConfigKey(ConfigKeyPath configKey) {
+    Preconditions.checkArgument(configKey.getAbsolutePathString().startsWith(this.physicalStoreRoot.getPath()),
+        String.format("The configKey \"%s\" must begin with \"%s\"!", configKey, this.physicalStoreRoot.getPath()));
+
+    return PathUtils.withoutLeadingSeparator(
+        new Path(StringUtils.removeStart(configKey.getAbsolutePathString(), this.physicalStoreRoot.getPath())));
```
ibuenros: What does this line do?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+  /**
+   * Constructs a {@link Path} that points to the location of the given version of the {@link ConfigStore} on HDFS. If
+   * this {@link Path} does not exist, a {@link VersionDoesNotExistException} is thrown.
+   */
+  private Path getVersionRoot(String version)
+      throws VersionDoesNotExistException {
+
+    Path versionRootPath = PathUtils.combinePaths(this.physicalStoreRoot.toString(), CONFIG_STORE_NAME, version);
+    try {
+      if (this.fs.isDirectory(versionRootPath)) {
```
ibuenros: Maybe we should cache known versions? This would have to check for existence of version every call to `getConfig`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+ *
+ * <p>
+ *   This class is not responsible for deploying configurations from an external source to HDFS, only for reading them.
+ * </p>
+ *
+ * @see {@link SimpleHDFSConfigStoreFactory}
+ */
+public class SimpleHDFSConfigStore implements ConfigStore {
+
+  public static final String SIMPLE_HDFS_SCHEME_PREFIX = "simple-";
```
ibuenros: what is this used for?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -0,0 +1,111 @@
...
+   */
+  private URI getStoreRoot(FileSystem fs, URI configKey) throws ConfigStoreCreationException {
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(configKey.getPath()),
+        String.format("Root Path must be specified for configKey: %s", configKey));
+    Path path = new Path(configKey.getPath());
+
+    while (path != null) {
+      try {
+        if (fs.exists(path)) {
+          for (FileStatus fileStatus : fs.listStatus(path)) {
```
ibuenros: you can do fs.getFileStatus instead

ibuenros: Could you do `fs.getFileStatus(new Path(path, SimpleHDFSConfigStore.CONFIG_STORE_NAME))` instead?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+  public static final String CONFIG_STORE_NAME = "_CONFIG_STORE";
```
tuGithub: The _CONFIG_STORE should be an identity file , not a directory, the structure of the hdfs may look like:
.
└── HdfsBasedConfigTest
    ├── _CONFIG_STORE
    └── v3.0
        ├── datasets
        │   ├── a1
        │   │   └── a2
        │   │       ├── a3
        │   │       │   ├── includes
        │   │       │   └── main.conf
        │   │       └── main.conf
        │   └── data
        │       └── tracking
        │           └── main.conf

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+  /**
+   * Returns a {@link String} representation of the highest version stored in the {@link ConfigStore}. This method
+   * simply determines the highest version by doing an {@code ls} on the store root, and returning the first entry
+   * found. Since this method performs an {@code ls}, the highest version should also correspond to the version that
+   * comes first in lexicographical order.
+   *
+   * @return a {@link String} representing the current (highest) version of the {@link ConfigStore}.
+   */
+  @Override
+  public String getCurrentVersion() {
```
tuGithub: should loop through the versions in the root and find the latest one

tuGithub: previously I used SimpleVersionFinder.java to find the version,

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: The JavaDocs for `ConfigStore.getStorURI` says `Obtains the config store root URI. This typically represents the physical location of the store...` so I thought it should return `hdfs` rather than `simple-hdfs`. Right?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: Yes, I will fix this in this method as well as the others.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: Yes, deleted `IncludesConfFilterFilter` and `MainConfFileFilter`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  /**
+   * Returns a {@link URI} representing the physical store {@link URI} where the {@link URI#getPath()} is the path to
+   * the root of the {@link ConfigStore}.
+   *
+   * @return a {@link URI} representing the physical store {@link URI} (e.g. hdfs://[authority]:[port][path-to-root]).
+   */
+  @Override
+  public URI getStoreURI() {
```
tuGithub: We may need to pass in two store root, one is physical  , one is logical, for example, client app pass in URI 

etl-hdfs://nertznamenode:9000/$configStoreRoot/data/database

the physical location may be hdfs://nertznamenode:9000/$configStoreRoot/data/database
but the store root URI should be etl-hdfs://nertznamenode:9000/$configStoreRoot/data/database

as the store root uri should have the same scheme/authority as the client URI

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: Basically it ensures the dataset doesn't being with `/` so that it can be combined with the result of `getVersionRoot`. I have updated the code a bit to make it clearer.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: It's basically the prefix of the logical store. For `SimpleHDFSConfigStore` the logical scheme is `simple-hdfs` for `SimpleLocalHDFSConfigStore` the logical scheme is `simple-local`. The name `simple-` is arbitrary. I thought it useful to have a prefix since there may be multiple `ConfigStore`s that right to HDFS. We may want to create additional HDFS backed `ConfigStore`s that are more efficient.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
sahilTakiar: But that would throw a `FileNotFoundException` if `new Path(path, SimpleHDFSConfigStore.CONFIG_STORE_NAME)` does not exist, which can happen as `SimpleHDFSConfigStore.CONFIG_STORE_NAME` is only suppose to exist in one of the parent directories.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: Is there advantage of having this directory layout over the current one?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: Yes, I saw that. For now I just stuck to a simple lexicographical sort, but if we need something more complex I can look into `SimpleVersionFinder.java`.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: I believe @ibuenros brought up a similar comment. Let's discuss offline. I can change it to the logical URI, but the JavaDocs in `ConfigStore` say it should return the physical URI.

gobblin-config-management/gobblin-config-core/build.gradle
sahilTakiar: I believe its for the `Alpha` annotations.

sahilTakiar:
@ibuenros, @tuGithub PR updated, comments addressed.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
ibuenros: The contract was unclear before. For correct working of the library, the store uri should be `simple-hdfs...`. Min is adding an explicit check for this, and returning `hdfs` will throw an error.

sahilTakiar: Ok, fixed and updated.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -0,0 +1,108 @@
...
+
+import gobblin.config.store.api.ConfigStoreCreationException;
+import gobblin.config.store.api.ConfigStoreFactory;
+
+
+/**
+ * An implementation of {@link ConfigStoreFactory} for creating {@link SimpleHDFSConfigStore}s. This class only works
+ * the physical scheme {@link #HDFS_SCHEME_NAME}.
+ */
+public class SimpleHDFSConfigStoreFactory implements ConfigStoreFactory<SimpleHDFSConfigStore> {
```
tuGithub: ConfigStoreFactory need to handle the URI without authority / config store root, for example 
etl-hdfs:///datasets/a1/a2

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,374 @@
...
+
+  /**
+   * Constructs a {@link SimpleHDFSConfigStore} using a given {@link FileSystem} and a {@link URI} that points to the
+   * physical location of the store root.
+   *
+   * @param fs the {@link FileSystem} the {@link ConfigStore} is stored on.
+   * @param physicalStoreRoot the fully qualified {@link URI} of the store root, the {@link URI#getScheme()} of the
+   *                          {@link URI} should match the {@link FileSystem#getScheme()} of the given {@link FileSystem}.
+   */
+  protected SimpleHDFSConfigStore(FileSystem fs, URI physicalStoreRoot) {
```
tuGithub: Prefer to not have the FileSystem as parameter as through physicalStoreRoot can get the FileSystem.

Also, prefer to have "URI physical_root, URI logic_root" as parameter  , then getStoreURI will return logic_root

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,374 @@
...
+  private Path getDatasetDirForKey(ConfigKeyPath configKey, String version) throws VersionDoesNotExistException {
+    System.out.println("getVersionRoot(version) " + getVersionRoot(version));
+    System.out.println("getDatasetFromConfigKey(configKey) " + getDatasetFromConfigKey(configKey));
+    return new Path(getVersionRoot(version), getDatasetFromConfigKey(configKey));
+  }
+
+  /**
+   * Retrieves the name of a dataset from a given {@link ConfigKeyPath}.
+   */
+  private String getDatasetFromConfigKey(ConfigKeyPath configKey) {
```
tuGithub: configKey is relative path, not absolute, for example; configKeyPath is /data/databases
but store root will be etl-hdfs://eat1-nertznn01.grid.linkedin.com:9000/user/mitu/HdfsBasedConfigTest/

sahilTakiar:
@tuGithub comments addressed, PR updated

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,370 @@
...
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(
+          String.format("Error while checking if the configStoreDir: \"%s\" exists", configStoreDir), e);
+    }
+
+    try {
+      FileStatus[] fileStatuses = this.fs.listStatus(configStoreDir);
+      Arrays.sort(fileStatuses, new FileStatusPathNameComparator());
+      return fileStatuses[fileStatuses.length - 1].getPath().getName();
```
tuGithub: should only return the version which is a directory

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,370 @@
...
+   * @throws VersionDoesNotExistException if the version specified cannot be found in the {@link ConfigStore}.
+   */
+  @Override
+  public Config getOwnConfig(ConfigKeyPath configKey, String version)
+      throws VersionDoesNotExistException {
+    Preconditions.checkNotNull(configKey, "configKey cannot be null!");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(version), "version cannot be null or empty!");
+
+    Path datasetDir = getDatasetDirForKey(configKey, version);
+    Path mainConfFile = new Path(datasetDir, "main.conf");
```
tuGithub: please define main.conf in static variable

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -0,0 +1,370 @@
...
+   */
+  @Override
+  public List<ConfigKeyPath> getOwnImports(ConfigKeyPath configKey, String version)
+      throws VersionDoesNotExistException {
+    Preconditions.checkNotNull(configKey, "configKey cannot be null!");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(version), "version cannot be null or empty!");
+
+    List<ConfigKeyPath> configKeyPaths = new ArrayList<>();
+    Path datasetDir = getDatasetDirForKey(configKey, version);
+    Path includesFile = new Path(datasetDir, "includes.conf");
```
tuGithub: please define "includes.conf" in static variable

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -0,0 +1,160 @@
...
+   * {@link FileSystem} implementations, subclasses should override this method.
+   */
+  protected String getPhysicalScheme() {
+    return HDFS_SCHEME_NAME;
+  }
+
+  /**
+   * Gets a default root directory if one is not specified. The default root dir is {@code /user/[current-user]/}.
+   */
+  protected Path getDefaultRootDir() throws IOException {
```
tuGithub: I would think the default store root is user independent , maybe from a configuration file.

Otherwise, different user will get different config store root, which will be wrong

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -0,0 +1,160 @@
...
+    return new Path("user", UserGroupInformation.getCurrentUser().getUserName());
+  }
+
+  /**
+   * Gets a default authority if one is not specified. The default authority is the authority of the {@link FileSystem}
+   * the current process is running on. For example, when running on a HDFS node, the authority will taken from the
+   * NameNode {@link URI}.
+   */
+  private String getDefaultAuthority() throws IOException {
```
tuGithub: This has the assumption that in each cluster, there will be a config store , not sure whether that is valid

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -0,0 +1,160 @@
...
+   *   not contain the dataset root. In which case it uses the {@link #getDefaultRootDir()} as the root directory. If
+   *   the default root dir does not contain the {@link SimpleHDFSConfigStore#CONFIG_STORE_NAME} then a
+   *   {@link ConfigStoreCreationException} is thrown.
+   * </p>
+   */
+  private URI getStoreRoot(FileSystem fs, URI configKey) throws ConfigStoreCreationException {
+    if (Strings.isNullOrEmpty(configKey.getAuthority())) {
+      try {
+        Path defaultRootDir = getDefaultRootDir();
+        if (!fs.exists(new Path(defaultRootDir, SimpleHDFSConfigStore.CONFIG_STORE_NAME))) {
```
tuGithub: The ConfigStore assume the CONFIG_STORE_NAME is a directory, we need to add the check in the ConfigStoreFactory ?

That is why I prefer to have CONFIG_STORE_NAME is NOT the parent directory of all the versions

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
sahilTakiar: What do you mean exactly? The method currently returns the version name, e.g. `v1.0` or `v1.1`. Do you want the full path to the version?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
sahilTakiar: This check is only for when an authority is not specified. In which case a default root is provided, `/jobs/[current-user]`. We have to check the root contains a dir called `_CONFIG_STORE`. So the check simply `fs.exists(/jobs/[current-user]/_CONFIG_STORE)`.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
sahilTakiar: Discussed offline, sticking to the current approach.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
sahilTakiar: Discussed offline, sticking to the current approach.

sahilTakiar:
@tuGithub comments addressed, PR updated

sahilTakiar:
@tuGithub PR updated

tuGithub:
LGTM

sahilTakiar:
@ibuenros any more comments?

gobblin-config-management/gobblin-config-core/build.gradle
```diff
@@ -12,10 +12,14 @@ apply plugin: 'java'
...
+    compile project(':gobblin-utility')
 
     compile externalDependency.slf4j
     compile externalDependency.log4j
+    compile externalDependency.lombok
     compile externalDependency.typesafeConfig
+    compile externalDependency.commonsLang3
```
ibuenros: can we use `commonsLang` instead of `commonsLang3`? we should stop mixing them...

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
ibuenros: Just for the record, why was the decision to stick with this?

ibuenros:
Couple comments, LGTM otherwise.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
sahilTakiar: We talked to Chavdar, and he suggested that the `ConfigStoreFactory` needs some type of `configure(Config config)` method so that it can configure itself. This `Config` object would also contain reasonable defaults. However, since that method currently does not exist, we decided just to stick with this for now.

gobblin-config-management/gobblin-config-core/build.gradle
sahilTakiar: Fixed

PR #641, opened 2016-01-28
POC embedded sql for config-driven retention management

pcadabam:
This is a POC for exploring SQL for retention management.
Dataset structure on HDFS will be loaded into an in-memory database (in the case Derby) by the DatasetFinders. The retention policies will be represented as SQL queries.

The POC has two tables 
- Database snapshots(dataset_path VARCHAR(255), name VARCHAR(255), path VARCHAR(255), ts TIMESTAMP, row_count bigint)
- Kafka daily partitioned data Daily_Partitions (dataset_path VARCHAR(255), path VARCHAR(255), ts TIMESTAMP)

1) Keep the last 2
SELECT * FROM Snapshots ORDER by ts desc LIMIT 2;

2) Keep partitions for last 2 years
SELECT * FROM Daily_Partitions WHERE Timestamp_diff(now, ts, "Days") <= 730 ;

UDFs are defined in SqlUdfs.java

@chavdar can you review?

### DIFF
--- a/gobblin-data-management/src/test/java/gobblin/data/management/retention/sql/SqlBasedRetentionPoc.java
+++ b/gobblin-data-management/src/test/java/gobblin/data/management/retention/sql/SqlBasedRetentionPoc.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2014-2015 LinkedIn Corp. All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


pcadabam:
Merging. Got a verbal LGTM from @chavdar

PR #642, opened 2016-01-28
Fixing NPE in CliOptions.java

sahilTakiar:


sahilTakiar:
Trivial fix, self-approving.

PR #645, opened 2016-01-29
Fix parent directory creation in distcp-ng

pcadabam:
@ibuenros can you review?
We rely on fs.mkdirs to create the parent directories of a CopyableFile if not exists. If the directories exist (created by other mappers), fs.mkdirs will return a false which we ignore.


ibuenros:
LGTM

PR #646, opened 2016-01-29
Downgraded jetty version to be java 7 compatible.

ibuenros:


ibuenros:
Trivial change. Self-approved.
@eogren I downgraded the version of jetty because the previous versions was compiled in java 8, Gobblin must be java 7 compatible atm.

eogren:
Oops! Sorry about that, should have tested with the 7 JDK.

PR #647, opened 2016-01-29
Add option to set group for distcp-ng

pcadabam:
@ibuenros can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
@@ -12,29 +12,68 @@
 
 package gobblin.data.management.copy;
 
+import java.util.Properties;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
 
 import org.apache.hadoop.fs.Path;
+
+import com.google.common.base.Optional;
 
 
 /**
  * Configuration for Gobblin distcp jobs.
  */
 @Data
+@AllArgsConstructor
+@Builder
 public class CopyConfiguration {
 
+  public static final String COPY_PREFIX = "gobblin.copy";
+  public static final String PRESERVE_ATTRIBUTES_KEY = COPY_PREFIX + ".preserved.attributes";
+  public static final String DESTINATION_GROUP_KEY = COPY_PREFIX + ".dataset.destination.group";
+
   /**
-   * Directory where dataset should be replicated.
-   * This directory corresponds to the {@link CopyableDataset#datasetRoot} in the new location.
+   * Directory where dataset should be replicated. This directory corresponds to the {@link CopyableDataset#datasetRoot}
+   * in the new location.
    */
   private final Path targetRoot;
+
   /**
    * Preserve options passed by the user.
    */
   private final PreserveAttributes preserve;
+
   /**
    * {@link CopyContext} for this job.
    */
   private final CopyContext copyContext;
+
+  private final Optional<String> targetGroup;
+
+  public static class CopyConfigurationBuilder {
+
+    private PreserveAttributes preserve;
+    private Optional<String> targetGroup;
+    private CopyContext copyContext;
+
+    public CopyConfigurationBuilder(Properties properties) {
+      this.targetGroup =
+          properties.containsKey(DESTINATION_GROUP_KEY) ? Optional.of(properties.getProperty(DESTINATION_GROUP_KEY))
+              : Optional.<String> absent();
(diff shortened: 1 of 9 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
```diff
@@ -12,29 +12,44 @@
...
 
+import com.google.common.base.Optional;
+
 
 /**
  * Configuration for Gobblin distcp jobs.
  */
 @Data
+@AllArgsConstructor
 public class CopyConfiguration {
```
ibuenros: Can we refactor this to use a builder, and be able to build from options? `DESTINATION_GROUP_KEY` and `PRESERVE_ATTRIBUTES_KEY` should be members of this class instead of `CopySource`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
```diff
@@ -12,29 +12,44 @@
...
   /**
    * Directory where dataset should be replicated.
    * This directory corresponds to the {@link CopyableDataset#datasetRoot} in the new location.
    */
   private final Path targetRoot;
+
   /**
    * Preserve options passed by the user.
    */
+
```
ibuenros: is there a reason for these line breaks?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -177,7 +177,14 @@ public CopyableFile build() throws IOException {
 
       if (this.destinationOwnerAndPermission == null) {
         String owner = this.preserve.preserve(Option.OWNER) ? this.origin.getOwner() : null;
-        String group = this.preserve.preserve(Option.GROUP) ? this.origin.getGroup() : null;
+
+        String group = null;
+        if (this.preserve.preserve(Option.GROUP)) {
+          group = this.origin.getGroup();
+        } else if (this.configuration.getTargetGroup().isPresent()) {
+          group = this.configuration.getTargetGroup().get();
```
ibuenros: Do we also have to assign the group to the ancestors?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
pcadabam: Removed.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
pcadabam: Yes we need to set. Fixed that.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
pcadabam: What does building from options mean?

ibuenros: Something like `CopyConfiguration.builder(Properties props).otherMethods.build()`, the builder will parse defaults from the Properties passed in (like preserve flags, target group).

pcadabam:
Addressed pr comments. @ibuenros take another look.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
```diff
@@ -12,29 +12,68 @@
...
+    private PreserveAttributes preserve;
+    private Optional<String> targetGroup;
+    private CopyContext copyContext;
+
+    public CopyConfigurationBuilder(Properties properties) {
+      this.targetGroup =
+          properties.containsKey(DESTINATION_GROUP_KEY) ? Optional.of(properties.getProperty(DESTINATION_GROUP_KEY))
+              : Optional.<String> absent();
+      this.preserve = PreserveAttributes.fromMnemonicString(properties.getProperty(PRESERVE_ATTRIBUTES_KEY));
```
ibuenros: Should this have a fallback if the property is not set?

ibuenros:
Just one comment, otherwise LGTM.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyConfiguration.java
pcadabam: PreserveAttributes.fromMnemonicString handles null already.

PR #650, opened 2016-01-30
Javadoc task should pick up system proxy settings

lbendig:
When Gobblin is built behind a corporate firewall build times might be extremely long since
the gradle javadoc task doesn't pick up the proxy settings by default (see [GRADLE-1228](https://issues.gradle.org/browse/GRADLE-1228)) and timeouts
when fetching external javadocs.
By setting the ```java.net.useSystemProxies``` flag the system-wide proxy settings will be applied which seems to solve this issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -266,8 +266,9 @@ subprojects {
   plugins.withType(JavaPlugin) {
  
     // Sometimes generating javadocs can lead to OOM. This may needs to be increased.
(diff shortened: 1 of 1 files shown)


sahilTakiar:
LGTM

lbendig:
@sahilTakiar Thanks for the review!

PR #651, opened 2016-02-01
Hive registration initial commit

zliu41:
This is the initial commit for Hive registration. It mainly contains interfaces, abstract classes, base implementations, as well as the main class `HiveRegister`.

- `HiveRegister` performs Hive registration given a `HiveSpec`, which specifies which path, which db, which table, etc., to register.  There are several extensions to `HiveSpec`, including `HiveSpecWithPredicates`, `HiveSpecWithPreActivities` and `HiveSpecWithPostActivities`. A base implementation `SimpleHiveSpec` is included.

- A `HiveSpec` is generated by a `HiveRegistrationPolicy` given a `Path`. A base implementation `HiveRegistrationPolicyBase` is included.

- To do Hive registration, a `StorageDescriptor` and some schema properties are often required. These are managed by `HiveStorageDescriptorManager` and `HiveSchemaManager`.


zliu41:

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+   * @return a {@link ListenableFuture} for the process of registering the given {@link HiveSpec}.
+   */
+  public ListenableFuture<Void> register(final HiveSpec spec) throws IOException {
+    ListenableFuture<Void> future = this.executor.submit(new Callable<Void>() {
+
+      @Override
+      public Void call() throws Exception {
+
+        if (spec instanceof HiveSpecWithPredicates && !evaluatePredicates((HiveSpecWithPredicates) spec)) {
+          log.info("Skipping " + spec + " since predicates return false");
```
sahilTakiar: I think you need a `return` after this line.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        if (spec instanceof HiveSpecWithPreActivities) {
+          for (Activity<HiveMetaStoreClient> activity : ((HiveSpecWithPreActivities) spec).getActivities()) {
+            activity.execute(HiveRegister.this.client);
```
sahilTakiar: Do we expect a requirement to execute these activities in parallel?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+        "Missing required property " + HiveRegProps.HIVE_DB_ROOT_DIR);
+    db.setLocationUri(this.hiveDbRootDir.get() + Path.SEPARATOR + dbName + HIVE_DB_EXTENSION);
+
+    Lock lock = locks.get(dbName);
+    lock.lock();
+
+    try {
+      client.createDatabase(db);
+    } catch (AlreadyExistsException e) {
+      log.warn("Hive database " + dbName + " already exists");
```
sahilTakiar: We probably don't need these since (besides the initial run), this will always be printed out.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    } catch (TException e) {
+      log.error("Unable to create Hive database " + dbName);
+      throw Throwables.propagate(e);
```
sahilTakiar: Why do we need to log the error and re-throw it? I would either, wrap it in a checked exception like `CannotCreateHiveDBError` and re-throw it, or just `throw Throwables.propagate(e)`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private Table createTable(String dbName, String tableName, Optional<HivePartition> partition, StorageDescriptor sd) {
+    Table table = new Table();
+    table.setParameters(Maps.<String, String> newHashMap());
+    table.getParameters().put("EXTERNAL", Boolean.TRUE.toString());
```
sahilTakiar: Why don't you built the parameters in a `Map` first and then call `table.setParameters(map)`.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    table.setDbName(dbName);
+    table.setTableName(tableName);
+    if (partition.isPresent()) {
+      table.setPartitionKeys(partition.get().getKeys());
+    }
+
+    Preconditions.checkState(this.hiveOwner.isPresent(), "Missing required property " + HiveRegProps.HIVE_OWNER);
```
sahilTakiar: `checkState` to `checkArgument`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    try {
+      this.client.createTable(table);
+    } catch (TException e) {
+      log.error(String.format("Unable to create table %s from db %s", tableName, dbName));
+      throw Throwables.propagate(e);
+    }
+    return table;
+  }
+
+  public void deregister(String dbName, String tableName) {
```
sahilTakiar: Method name sounds a little value, deregistering what exactly? Recommend changing to `deregisterTable` or `dropTable`. Similar comment for the overloaded methods.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+    } catch (TException e) {
+      log.error(String.format("Unable to check existence of Hive partition %s in table %s in db %s", partitionValues,
+          tableName, dbName));
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private void addPartition(Table table, HivePartition partition) throws TException {
+    Preconditions.checkArgument(table.getPartitionKeysSize() == partition.getValues().size());
```
sahilTakiar: Probably need to check if the `dbName` `tableName` and `serDe` of the given `table` and `partition` match.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+  private final ListeningExecutorService executor;
+  private final List<Future<Void>> futures = Lists.newArrayList();
+  private final Striped<Lock> locks = Striped.lazyWeakLock(Integer.MAX_VALUE);
+
+  public HiveRegister(State state) throws MetaException {
+    this.props = new HiveRegProps(state);
+    this.hiveDbRootDir = this.props.getDbRootDir();
+    this.hiveOwner = this.props.getHiveOwner();
+    this.client = new HiveMetaStoreClient(new HiveConf());
+    this.executor = MoreExecutors.listeningDecorator(
```
sahilTakiar: What's the point of using `MoreExecutors.listeningDecorator`?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSchemaManager.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Get an instance of {@link HiveSchemaManager}.
+   *
+   * @param type The {@link HiveSchemaManager} type. It could be AVRO, NOP or the name of a class that implements
+   * {@link HiveSchemaManager}.
+   * @param props A {@link State} object used to instantiate {@link HiveSchemaManager}.
+   * @return
+   */
+  public static HiveSchemaManager getInstance(String type, State props) {
```
sahilTakiar: Consider using a `[ServiceLoader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)` instead of this pattern. PR https://github.com/linkedin/gobblin/pull/627 has an example of how to use it. This way each implementation of `HiveSchemaManager` does not have to explicitly register itself with this class. It should also allow you to make this class an interface instead of an abstract class.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveNopSchemaManager.java
```diff
@@ -0,0 +1,40 @@
...
+import gobblin.configuration.State;
+
+
+/**
+ * A {@link HiveSchemaManager} that does nothing. This should be used when registering a table
+ * that does not require specifying schema properties.
+ *
+ * @author ziliu
+ */
+@Alpha
```
sahilTakiar: I think the `Alpha` annotations are typically meant just for public APIs (e.g. interfaces or abstract classes), although there is no harm in tagging concrete class too.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,302 @@
...
+
+      @Override
+      public Void call() throws Exception {
+
+        if (spec instanceof HiveSpecWithPredicates && !evaluatePredicates((HiveSpecWithPredicates) spec)) {
+          log.info("Skipping " + spec + " since predicates return false");
+        }
+
+        if (spec instanceof HiveSpecWithPreActivities) {
+          for (Activity<HiveMetaStoreClient> activity : ((HiveSpecWithPreActivities) spec).getActivities()) {
```
sahilTakiar: Should the `Activity` take in a `HiveMetaStoreClient`? Maybe it should take in a `HiveRegister` instead? I'm guessing a common use case for pre and post activities would be to drop partitions? Since the `HiveRegister` class already has methods to drop partitions it may be best to use those, especially since those both are protected by locks. Providing unrestricted access to `HiveMetaStoreClient` may be dangerous since the registration is multi-threaded.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Not really. This is for the use case of dropping the hourly partitions after registering the daily partition. I'm not even sure whether dropping multiple partitions from the same table can be parallelized or not, but even if it can it likely won't bring any significant performance gain.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: This is just in case the exception message is not clear enough. Adding a log here to ensure printing a message that we can control.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: This method takes a `Table` and partition names and values. The actual `Partition` object is constructed in this method using the dbname, tablename and serde of the given `Table`.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: `register()` returns a `ListenableFuture` so that the caller of `register()` can add whatever callbacks they want.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: great point

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSchemaManager.java
zliu41: Took a look at `ServiceLoader`, it seems to require that the concrete classes have zero-argument constructor. But `HiveSchemaManager` needs to be instantiated with a `State`.

zliu41:
@sahilTakiar updated 45d36e8

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: fixed

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  /**
+   * Register a table or partition given a {@link HiveSpec}.
+   *
+   * This method is asynchronous and returns immediately.
+   *
+   * @return a {@link ListenableFuture} for the process of registering the given {@link HiveSpec}.
+   */
+  public ListenableFuture<Void> register(final HiveSpec spec) throws IOException {
```
ydai1124: Is it possible to have multiple `HiveSpec`, e.g., both `HiveSpecWithPredicates` and `HiveSpecWithPreActivities`?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+    return true;
+  }
+
+  private void createDbIfNotExists(String dbName) {
+    Database db = new Database();
+    db.setName(dbName);
+
+    Preconditions.checkState(this.hiveDbRootDir.isPresent(),
+        "Missing required property " + HiveRegProps.HIVE_DB_ROOT_DIR);
+    db.setLocationUri(this.hiveDbRootDir.get() + Path.SEPARATOR + dbName + HIVE_DB_EXTENSION);
```
ydai1124: Do you need to deal with multiple `Path.SEPARATOR` if one put one more separator to the end of `this.hiveDbRootDir.get()`? If that is the case, you can use `new Path(String, String).toString()` to deal with that.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+
+    Lock lock = locks.get(dbName);
+    lock.lock();
+
+    try {
+      client.createDatabase(db);
+      log.info("Created database " + dbName);
+    } catch (AlreadyExistsException e) {
+      // Database already exists. Nothing to do.
```
ydai1124: Can you add one more log if the db exists?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+    } catch (TException e) {
+      log.error(String.format("Unable to check existence of table %s in db %s", tableName, dbName));
+      throw Throwables.propagate(e);
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  public boolean existsPartition(String dbName, String tableName, List<String> partitionValues) {
+    Lock lock = locks.get(dbName + tableName);
```
ydai1124: Do you need to lock the whole table when checking the existence of some partitions?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+    } catch (TException e) {
+      log.error(String.format("Unable to check existence of Hive partition %s in table %s in db %s", partitionValues,
+          tableName, dbName));
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private void addPartition(Table table, HivePartition partition, Path partitionLocation) throws TException {
+    Preconditions.checkArgument(table.getPartitionKeysSize() == partition.getValues().size());
```
ydai1124: When could this mismatch happen?

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSchemaManager.java
```diff
@@ -0,0 +1,84 @@
...
+    log.info("Schema is " + schema);
+    si.getParameters().put(SCHEMA_LITERAL, schema.toString());
+  }
+
+  @SuppressWarnings("deprecation")
+  private Schema getSchema(Path path) {
+    try {
+      for (FileStatus status : this.fs.listStatus(path)) {
+        if (status.isDir()) {
+          Schema schema = getSchema(status.getPath());
```
ydai1124: If there are multiple subdirs, are we assuming files under all subdirs have the same schema?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
sahilTakiar: I generally think throwing a checked exception liked `CannotCreateHiveDBError` is a better way of adding a custom message to any exception thrown.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSchemaManager.java
sahilTakiar: `HiveSchemaManager` can just have a `configure(State state)` method, and probably a `String getType()` method that returns either `nop` or `avro`. Functionality its the same, so its up to you.

gobblin-hive-registration/build.gradle
```diff
@@ -0,0 +1,82 @@
...
+      excludeGroups 'ignore'
+      if (project.hasProperty('useHadoop2')) {
+        excludeGroups 'Hadoop1Only'
+      }
+    }
+    workingDir rootProject.rootDir
+    maxParallelForks = 1
+}
+
+clean {
```
sahilTakiar: Is this necessary? Same applies to the `artifacts` and `testJar` closures and the `tests` configuration.

gobblin-hive-registration/src/main/java/gobblin/hive/HivePartition.java
```diff
@@ -0,0 +1,56 @@
...
+    this.values = ImmutableList.copyOf(values);
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < keys.size(); i++) {
+      sb.append(keys.get(i).getName());
+      sb.append("=");
+      sb.append(values.get(i));
```
sahilTakiar: Probably want a `, ` in between each key value pair

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
```diff
@@ -0,0 +1,97 @@
...
+  public static final String HIVE_REGISTER_THREADS = "hive.register.threads";
+  public static final int DEFAULT_HIVE_REGISTER_THREADS = 20;
+  public static final String HIVE_TABLE_PROPS = "hive.table.props";
+
+  private static final Splitter SPLITTER = Splitter.on('=').trimResults().omitEmptyStrings();
+
+  @Getter
+  private final State tableProps;
+
+  public HiveRegProps(State props) {
```
sahilTakiar: Why not make this constructor take in two `State` objects, one `props` and the other `tableProps`. Seems cleaner the forcing the caller to put all table props into `hive.table.props`.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -0,0 +1,112 @@
...
+@Alpha
+public abstract class HiveRegistrationPolicyBase implements HiveRegistrationPolicy {
+
+  public static final String HIVE_DATABASE_NAME = "hive.database.name";
+  public static final String HIVE_DATABASE_REGEX = "hive.database.regex";
+  public static final String HIVE_TABLE_NAME = "hive.table.name";
+  public static final String HIVE_TABLE_REGEX = "hive.table.regex";
+
+  private static final Pattern NAME_PATTERN_1 = Pattern.compile("[a-z0-9][a-z0-9_]*");
+  private static final Pattern NAME_PATTERN_2 = Pattern.compile(".*[a-z_].*");
```
sahilTakiar: Make names of these variables more descriptive.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * A base implementation of {@link HiveRegistrationPolicy}. It obtains database name from
+ * property {@link #HIVE_DATABASE_NAME} or {@link #HIVE_DATABASE_REGEX} (group 1), obtains
+ * table name from property {@link #HIVE_TABLE_NAME} and {@link #HIVE_TABLE_REGEX} (group 1),
+ * and builds a {@link SimpleHiveSpec}.
+ *
+ * @author ziliu
+ */
+@Alpha
+public abstract class HiveRegistrationPolicyBase implements HiveRegistrationPolicy {
```
sahilTakiar: I would suggest splitting this into two classes one for `HiveRegisrationPolicyFromState` and `HiveRegistrationPolicyFromRegex`. The former checks the given `State` to see if the db and table are specified, the later uses a regex to find the db and table.

Is there a concrete use case for having a policy where the state is first checked, and then a regex is used?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+    Map<String, String> parameters = Maps.newHashMap();
+    parameters.put("EXTERNAL", Boolean.TRUE.toString());
+    Table table = new Table();
+    table.setParameters(parameters);
+    table.setDbName(dbName);
+    table.setTableName(tableName);
+
+    Preconditions.checkArgument(this.hiveOwner.isPresent(), "Missing required property " + HiveRegProps.HIVE_OWNER);
+    table.setOwner(this.hiveOwner.get());
+    table.setTableType(TableType.EXTERNAL_TABLE.toString());
```
cwsteinbach: Why are you assuming that this is an external table?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+      log.info("Created table " + tableName + " in db " + dbName);
+    } catch (TException e) {
+      log.error(String.format("Unable to create table %s from db %s", tableName, dbName));
+      throw Throwables.propagate(e);
+    }
+    return table;
+  }
+
+  public boolean existsTable(String dbName, String tableName) {
+    Lock lock = locks.get(dbName + tableName);
```
cwsteinbach: "dbName + tableName" is not a unique lock ID. For example the tables a.bc and ab.c both map to the same lock ID.

cwsteinbach: I recommend creating a TableLock subclass which specifically takes dbName and tableName as input parameters in order to abstract out the creation of the lock ID.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+
+    Lock lock = locks.get(table.getDbName() + table.getTableName());
+    lock.lock();
+
+    try {
+      this.client.add_partition(p);
+      log.info(String.format("Added partition %s to table %s with location %s", partition, table.getTableName(),
+          p.getSd().getLocation()));
+    } catch (AlreadyExistsException e) {
+      // Partition already exists. Nothing to do.
```
cwsteinbach: I think it's worth emitting a log message about this. Also, what if the partition that already exists has different values for the StorageDescriptor, location, etc?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,351 @@
...
+
+  public void dropPartition(String dbName, String tableName, List<String> partitionValues) {
+    Lock lock = locks.get(dbName + tableName);
+    lock.lock();
+
+    try {
+      client.dropPartition(dbName, tableName, partitionValues, false);
+      log.info("Dropped partition " + partitionValues + " in table " + tableName + " in db " + dbName);
+    } catch (NoSuchObjectException e) {
+      // Partition does not exist. Nothing to do
```
cwsteinbach: Worth logging a WARN message here.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
zliu41: Yes, for example for tracking, the db name is just "tracking". The table name is a component in the path, which can be retrieved using a regex.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
zliu41: yes I can add another constructor that takes two arguments.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: I guess this class is for the purpose of registering a specific path ingested by Gobblin in Hive. So it should be external table.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: good point, will do alter_table and alter_partition if the new table or partition is different.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSchemaManager.java
zliu41: That's true. But if the user directly instantiates a `HiveSchemaManager` implementation using the constructor, he may forget to call `configure(state)`. So for me the current approach is preferred

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: A concrete `HiveSpec` implementation can implement both `HiveSpecWithPredicates` and `HiveSpecWithPreActivities`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: This may flood the log. The db will only be created once, and after that every single registration of a partition will print this log msg.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: If someone creates a table that partitions on column A, and then try to add a partition whose partition key is A and B.

That being said I should probably compare the actual partition columns rather than the size.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSchemaManager.java
zliu41: Yes, it is not `HiveRegister`'s responsibility to verify that, because it is too expensive. If some user tries to register a path that contains different schemas, the registration will simply fail.

gobblin-hive-registration/build.gradle
zliu41: probably not. Don't remember where I copied this from.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Instead of locking on a single `HiveMetaStoreClient` object, I probably need to use a pool of `HiveMetaStoreClient` objects, since `HiveMetaStoreClient` doesn't seem to support concurrency.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Instead of locking on a single `HiveMetaStoreClient` object, I probably need to use a pool of `HiveMetaStoreClient` objects, since `HiveMetaStoreClient` doesn't seem to support concurrency.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: I'll change the name of this method to `dropPartitionIfExists`. Since it is normal that a partition doesn't exist, there shouldn't be a log warn.

zliu41:
@sahilTakiar @ydai1124 @cwsteinbach updated 99035f7

Main changes:
- Use a pool of `HiveMetaStoreClient`s instead of a single `HiveMetaStoreClient`, because `HiveMetaStoreClient` doesn't seem to support concurrent requests.
- Added a striped lock class (`HiveStripedLocks`) as suggested by @cwsteinbach 
- When creating a table/partition, if the table/partition already exists, it checks whether the table or partition needs to be altered (`needToUpdateTable` and `needToUpdatePartition`). For example, update is needed if there's a schema change. If so it alters the table or partition.
- Addressed other comments.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -0,0 +1,143 @@
...
+    return getDatabaseOrTableName(path, HIVE_TABLE_NAME, HIVE_TABLE_REGEX);
+  }
+
+  protected String getDatabaseOrTableName(Path path, String nameKey, String regexKey) {
+    String name = null;
+    if (this.props.contains(nameKey)) {
+      name = this.props.getProp(nameKey);
+    } else if (this.props.contains(regexKey)) {
+      String regex = this.props.getProp(regexKey);
+      name = Pattern.compile(regex).matcher(path.toString()).group();
```
sahilTakiar: Make a instance variable for the result of `Pattern.compile(regex)` since its will be expensive to compile a regex for every give `Path`

gobblin-hive-registration/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+  testCompile externalDependency.calciteCore
+  testCompile externalDependency.calciteAvatica
+  testCompile externalDependency.jhyde
+  testCompile externalDependency.testng
+  testCompile externalDependency.byteman
+  testCompile externalDependency.bytemanBmunit
+  testRuntime externalDependency.derby
+}
+
+task testJar(type: Jar, dependsOn: testClasses) {
```
sahilTakiar: Don't need this

gobblin-hive-registration/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+    baseName = "test-${project.archivesBaseName}"
+    from sourceSets.test.output
+}
+
+configurations {
```
sahilTakiar: Don't need this either.

gobblin-hive-registration/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+  compile externalDependency.commonsPool
+  compile externalDependency.findBugs
+  compile externalDependency.lombok
+  if (project.hasProperty('useHadoop2')) {
+    compile externalDependency.avroMapredH2
+  } else {
+    compile externalDependency.avroMapredH1
+  }
+
+  testCompile externalDependency.calciteCore
```
sahilTakiar: Don't think you need any `testCompile` dependencies as there are no tests.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSchemaManager.java
```diff
@@ -0,0 +1,84 @@
...
+    super(props);
+    this.fs = FileSystem.get(HadoopUtils.getConfFromState(props));
+  }
+
+  @Override
+  public void addSchemaProperties(SerDeInfo si, Path path) {
+    Schema schema = getSchema(path);
+
+    Preconditions.checkNotNull(schema);
+    log.info("Schema is " + schema);
```
ydai1124: Can you add the `path` info to this log message, e.g., `Schema retrieved from <path> is ...`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,413 @@
...
+    if (partition.isPresent()) {
+      table.setPartitionKeys(partition.get().getKeys());
+    }
+
+    table.setSd(sd);
+    return table;
+  }
+
+  private boolean needToUpdateTable(Table existingTable, Table newTable) {
+    return !existingTable.equals(newTable);
```
sahilTakiar: How is this going to work? Looking at http://grepcode.com/file/repo1.maven.org/maven2/org.apache.hive/hive-metastore/1.0.1/org/apache/hadoop/hive/metastore/api/Table.java#Table.equals%28org.apache.hadoop.hive.metastore.api.Table%29 a `Table` is compared based on its create time, so I would guess the result of `createTableWithoutRegistering` will never be equal to `existingTable`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * An implementation of {@link BasePoolableObjectFactory} for {@link HiveMetaStoreClient}.
+ *
+ * @author ziliu
+ */
+public class HiveMetaStoreClientFactory extends BasePoolableObjectFactory<HiveMetaStoreClient> {
+
+  @Override
+  public HiveMetaStoreClient makeObject() throws Exception {
+    return new HiveMetaStoreClient(new HiveConf());
```
sahilTakiar: When will all the `HiveMetaStoreClient`s be closed?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * An implementation of {@link BasePoolableObjectFactory} for {@link HiveMetaStoreClient}.
+ *
+ * @author ziliu
+ */
+public class HiveMetaStoreClientFactory extends BasePoolableObjectFactory<HiveMetaStoreClient> {
+
+  @Override
+  public HiveMetaStoreClient makeObject() throws Exception {
```
sahilTakiar: This should probably return `IMetaStoreClient` instead as according to the java docs that seems to be the public API:

https://hive.apache.org/javadocs/r1.0.1/api/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.html
https://hive.apache.org/javadocs/r1.0.1/api/org/apache/hadoop/hive/metastore/IMetaStoreClient.html

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -0,0 +1,413 @@
...
+      throw Throwables.propagate(e);
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private Table createTableWithoutRegistering(String dbName, String tableName, Optional<HivePartition> partition,
+      StorageDescriptor sd) {
+    Map<String, String> parameters = Maps.newHashMap();
+    parameters.put("EXTERNAL", Boolean.TRUE.toString());
```
sahilTakiar: Is this needed considering the `table.setTableType(TableType.EXTERNAL_TABLE.toString())` is also invoked.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
ydai1124: The `regex` here is only used for matching the string representation of `path`. But the `HIVE_DATABASE_REGEX`/ `HIVE_DATABASE_REGEX` is very generic naming. Can we rename it?

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    if (isNameValid(name)) {
```
ydai1124: Can we have one more log here?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
```diff
@@ -0,0 +1,116 @@
...
+    this.tableProps = createTableProps();
+  }
+
+  /**
+   * @param props Properties required by {@link HiveRegister} to do Hive registration
+   * @param tableProps Hive properties that will be added to the Hive table when creating the table,
+   * e.g., orc.compress=SNAPPY
+   */
+  public HiveRegProps(State props, State tableProps) {
+    this.tableProps = tableProps;
```
sahilTakiar: Missing `super(props)`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
sahilTakiar: Why have the `HiveRegProps(State props)` constructor at all? It seems weird to force all values of `tablesProps` into a single props value. It's likely the invoker of this constructor will have to manually do this, and they won't realize they have to do this until they hit a bug.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveMetaStoreClientFactory.java
zliu41: good point

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: My understanding is a Table's create time is not set by default, and will only be set if one explicitly does so. Otherwise it is 0.
Even if it is always automatically set, it shouldn't cause any problem if we always alter the table just to update the create time?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveMetaStoreClientFactory.java
zliu41: agreed

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: @cwsteinbach can chime in here but this is what the current code does.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSchemaManager.java
zliu41: I'll actually remove this line since printing every table's schema is too much for driver log

gobblin-hive-registration/src/main/java/gobblin/hive/HiveMetaStoreClientFactory.java
zliu41: Added a destroyObject method in `HiveMetaStoreClientFactory` which will take care of this.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
zliu41: If the name is not valid, the exception message will be clear enough.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
zliu41: Because if we run it on Azkaban, then the program starts with a single property object from Azkaban. Someone has to separate tableProps from the property object somewhere, if not here then somewhere else. I don't see a reason why it should be somewhere else.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
sahilTakiar: Then `needToUpdateTable` will basically return true every time, because `existingTable` will have a valid created date (like 2015-02-08 00:00:01), but `newTable` will always have a created date of 0. So `alterTable` will be called every time, which won't cause any problems, but will be slightly less efficient.

What are the actual parameters we need to look for when deciding if the table needs to be altered? Can't there be a custom `Comparator` for comparing `Table` objects.

zliu41: I can set newTable's create time to be the same as existingTable's create time, then call equals().

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
zliu41: changed to `dbNamePattern` and `tableNamePattern`.

zliu41:
@sahilTakiar @ydai1124 updated 7855bf7

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
sahilTakiar: Yeah, that should work, although you will probably have to do this same for `lastAccessTime` and maybe even some other ones, there are some other fields in the `Table.equals` method, but I'm not sure what they all mean.

sahilTakiar:
A few more minor comments, overall it looks good.

ydai1124:
LGTM

PR #657, opened 2016-02-03
Issue #561 - fix for BaseDataPublisher to mark WorkingState correctly

sahilTakiar:
Fix for https://github.com/linkedin/gobblin/issues/561

* This should ensure that for a given `WorkUnitState` its data has been committed before the `WorkUnitState.WorkingState` is set to `COMMITTED`.
* The solution is a bit of hack because the code will now commit data for all `WorkUnitState`s first and then it will set the `WorkingState` of each `WorkUnitState` to `COMMITTED`
    * This is better than the previous behavior as it may cause duplicate data, but it will never drop data
    * If users need exactly-once guarantees they should use the `CommitSequencePublisher` from the exactly-once POC

@zliu41 can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -65,6 +65,7 @@ public class BaseDataPublisher extends SingleTaskDataPublisher {
   protected final List<Optional<String>> publisherFinalDirOwnerGroupsByBranches;
   protected final List<FsPermission> permissions;
   protected final Closer closer;
+  protected final Closer parallelRunnerCloser;
   protected final int parallelRunnerThreads;
   protected final Map<String, ParallelRunner> parallelRunners = Maps.newHashMap();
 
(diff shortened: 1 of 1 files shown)


kadaan:
@sahilTakiar: I'm wondering how this will work correctly with S3 given that job based publishing move/renames a folder at a time and that is not an atomic action.  #572 handles this differently by switching to renaming/moving the individual files within the folder. 

#569 allows for exactly-once publishing, but it too won't work correctly with S3.  For instance, while running a FsRenameCommitStep it could fail to copy the entire folder.  The root of the folder will exist and when it is retried it will be skipped because `this.dstFs.exists(this.dstPath)` return `true`.

As I said in my last comment on #572, I still want to get this in. It is a high priority for us (as well as you), but making sure this works with a wide range of possible filesystems is also important. We made good progress today on getting gobblin running in our stage environment and so I should be able to get back to #572 soon.  Please let me know how you want to proceed.

sahilTakiar:
@kadaan this is meant as more of a temp fix for https://github.com/linkedin/gobblin/issues/561 I agree it is not going to address all the issues for S3 publishing, but I just want to get this in so that it doesn't cause any data loss for any other users.

zliu41:
LGTM.
@kadaan I agree this won't work for exactly-once on S3. Something like #572 is still needed.

kadaan:
@zliu41 I'd still like to get #572 in.  It will take me a bit longer due to the changes from #569.

sahilTakiar:
@kadaan yes, lets still work on #572 although since this is such a big change, we may want to have a design doc outlining the changes. @chavdar what do you think?

kadaan:
@sahilTakiar I think this might have broken task based publishing.  The issue is that the `publishData(WorkUnitState state)` method never calls `parallelRunnerCloser.close()`

Issue #658, opened 2016-02-03
ParallelRunner can close without all futures finishing


kadaan:
The way `ParallelUtils.close()` works, the close method can return and futures can still complete in the background.  This can cause problems with code that depends on ParallelRunner and expect that the tasks are done when the ParallelRunner is closed (e.g. `TaskStateCollectorService`).  Additionally, it makes ParallelRunner more confusing.  If all tasks complete successfully, then ParallelRunner guarantees that they will be done when `ParallelUtils.close()` exits.  On the other hand, if a single task fails, then `ParallelUtils.close()` can exit before the rest of the tasks are done.

The following test shows the issue:
```java
@Test
  public void testWaitsForFuturesWhenClosing() throws IOException, InterruptedException {
    final AtomicBoolean flag = new AtomicBoolean();
    final CountDownLatch latch1 = new CountDownLatch(1);
    final CountDownLatch latch2 = new CountDownLatch(1);
    Path src1 = new Path("/src/file1.txt");
    Path src2 = new Path("/src/file2.txt");
    FileSystem fs = Mockito.mock(FileSystem.class);
    Mockito.when(fs.exists(src1)).thenReturn(true);
    Mockito.when(fs.delete(src1, true)).thenAnswer(new Answer<Boolean>() {
      @Override
      public Boolean answer(InvocationOnMock invocation) throws Throwable {
        latch1.countDown();
        return false;
      }
    });
    Mockito.when(fs.exists(src2)).thenReturn(true);
    Mockito.when(fs.delete(src2, true)).thenAnswer(new Answer<Boolean>() {
      @Override
      public Boolean answer(InvocationOnMock invocation) throws Throwable {
        latch1.await();
        long end = System.currentTimeMillis() + 70000;
        while (System.currentTimeMillis() < end) {
          try {
            Thread.sleep(Math.max(1, end - System.currentTimeMillis()));
          } catch (Exception ignored) {
          }
        }
        flag.set(true);
        latch2.countDown();
        return true;
      }
    });

    boolean caughtException = false;
    ParallelRunner parallelRunner = new ParallelRunner(2, fs);
    try {
      parallelRunner.deletePath(src1, true);
      parallelRunner.deletePath(src2, true);
      System.out.println(System.currentTimeMillis() + ": START - ParallelRunner.close()");
      parallelRunner.close();
    } catch (IOException e) {
      caughtException = true;
    }
    Assert.assertTrue(caughtException);
[...]

PR #661, opened 2016-02-03
Change ParallelRunner.close to wait for all futures to finish

kadaan:
Fix for #658


chavdar:
I agree that catching ExecutionException should be moved in the loop. I think that InterruptedException should be handled differently. Typically, that's triggered if the service needs to shutdown. In that case, ParallelRunner.close() should call cancel(true) on the remaining futures..

kadaan:
@chavdar How about this?

chavdar:
LGTM.

PR #664, opened 2016-02-03
Fix broken Gobblin version resolution ( fixes #662 )

### LINKED ISSUE #662: "Gobblin project version is not set!" or jars not found/do not exist
When running the `gobblin-mapreduce.sh` job I always receiving a `Gobblin project version is not set!` response.  If I attempt to pass a version with`--projectversion`, then I get a \<whatever>-\<projectversion>.jar does not exist or can't be found in /gobblin-dist/lib/...  
  
In `gobblin-mapreduce.sh` the variable is set:
`# Set during the distribution build`
`GOBBLIN_VERSION="0.6.2-4-g1f0109a"`  
  
There is a check for a `GOBBLIN_VERSION` further down that seems that it will always return the "Gobblin project version is not set!" response anytime an attempt in made to run the `gobblin-mapreduce.sh` job with the initial build.  
`if ( [ -z "$GOBBLIN_VERSION" ] || [ "$GOBBLIN_VERSION" == ""0.6.2-4-g1f0109a"" ] ); then`
`  die "Gobblin project version is not set!"`
`fi` 
  
Changing `[ "$GOBBLIN_VERSION" == ""0.6.2-4-g1f0109a"" ]` to `[ "$GOBBLIN_VERSION" != ""0.6.2-4-g1f0109a"" ]` resulted in the job running successfully.

lbendig:
ReplaceTokens replaces all occurences of a given token, but this is undesirable
when substituting `@project.version@` as it also replaces the value in the if condition
that checks for this placeholder, thus breaking the script.

### DIFF
--- a/gobblin-distribution/build.gradle
+++ b/gobblin-distribution/build.gradle
@@ -42,8 +42,6 @@ dependencies {
   }
 }
 
-import org.apache.tools.ant.filters.ReplaceTokens
-
 task build(type: Tar, overwrite: true) {
   extension = 'tar.gz'
   baseName = project.name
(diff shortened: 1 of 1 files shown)


sahilTakiar:
LGTM!

lbendig:
@sahilTakiar thank you for the review!

PR #665, opened 2016-02-03
Gobblin-compaction tarball doesn't contain gobblin-compaction.jar

lbendig:
The `:gobblin-compaction:createCompactionTar` task looks for `gobblin-compaction.jar` instead of the 
versioned jar name.


sahilTakiar:
LGTM!

lbendig:
@sahilTakiar thank you for the review!

PR #669, opened 2016-02-04
Parallelized copy listing generation in distcp.

ibuenros:
This PR changes work unit generation in `CopySource`:
* Enable parallelization: each dataset will be handled by a separate thread.
* Sets `CopySource` to allow for work unit limit and dataset / partition priorities in the future.

`ConcurrentBoundedWorkUnitList` is the core of these features: it is a concurrent, bounded, prioritized list of work units. Attempting to insert a file set when it doesn't fit will evict strictly lower priority file sets if and only if doing so would allow the higher priority file set to fit in the list. All file sets (groups of `WorkUnit`) are handled as a unit, so each file set is either completely evicted or retained.


ibuenros:
@pcadabam can you review when you get a chance?

gobblin-data-management/src/main/java/gobblin/data/management/copy/ConcurrentBoundedWorkUnitList.java
```diff
@@ -0,0 +1,153 @@
...
+    this.currentSize = 0;
+    this.maxSize = maxSize;
+    this.comparator = comparator;
+    this.workUnitsMap = new TreeMap<>(new AugmentedComparator(comparator));
+    this.rejectedFileSet = false;
+  }
+
+  /**
+   * Add a file set to the container.
+   * @param fileSet File set, expressed as a {@link Partition} of {@link CopyableFile}s.
```
pcadabam: Is there a reason we call fileset instead of partition? How is a fileset different from a partition?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -88,57 +101,128 @@
   @Override
   public List<WorkUnit> getWorkunits(SourceState state) {
 
-    List<WorkUnit> workUnits = Lists.newArrayList();
     CopyContext copyContext = new CopyContext();
 
     try {
 
+      FileSystem sourceFs = getSourceFileSystem(state);
+      FileSystem throttledSourceFs = getOptionallyThrottledFileSystem(sourceFs, state);
```
pcadabam: I think the logic for creating an (optional) thorttled fs can be in getSourceFileSystem and getTargetFileSystem

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -88,57 +101,128 @@
...
-          DatasetUtils.instantiateDatasetFinder(state.getProperties(), getSourceFileSystem(state),
-              DEFAULT_DATASET_PROFILE_CLASS_KEY);
+          DatasetUtils.instantiateDatasetFinder(state.getProperties(), sourceFs, DEFAULT_DATASET_PROFILE_CLASS_KEY);
       List<CopyableDataset> copyableDatasets = datasetFinder.findDatasets();
-      FileSystem targetFs = getTargetFileSystem(state);
 
-      for (CopyableDataset copyableDataset : copyableDatasets) {
+      // TODO: The comparator sets the priority of file sets. Currently, all file sets have the same priority, this needs to
+      // be pluggable.
+      ConcurrentBoundedWorkUnitList workUnitList = new ConcurrentBoundedWorkUnitList(100000,
```
pcadabam: Should we keep 100000 as the default size and make it configurable? If not we can have a no args constructors and have the CouncurrentBoundedWorkUnitList assume size and comparator

gobblin-data-management/src/main/java/gobblin/data/management/copy/ConcurrentBoundedWorkUnitList.java
ibuenros: The reason is a previous discussion on the definition of "partition", and "file set" was a more agreeable term. I don't want to change the name file set, but we can change the name of the class `Partition` to `FileSet`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
ibuenros: The plan is for it to be configurable. Future PR.

ibuenros:
@pcadabam addressed comments.

pcadabam:
LGTM

PR #677, opened 2016-02-08
Fix race condition in FsStateStore

zliu41:
@ibuenros can you review?

### DIFF
--- a/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
@@ -27,10 +27,13 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.compress.DefaultCodec;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
 import gobblin.configuration.State;
+import gobblin.util.HadoopUtils;
 
 
 /**
@@ -52,8 +55,13 @@ import gobblin.configuration.State;
  */
 public class FsStateStore<T extends State> implements StateStore<T> {
 
+  public static final String TMP_FILE_PREFIX = "_tmp_";
+  private static final ImmutableCollection<String> FS_SCHEMES_NO_TMP_FILE_FOR_PUT =
+      ImmutableSortedSet.<String> orderedBy(String.CASE_INSENSITIVE_ORDER).add("s3").add("s3a").add("s3n").build();
+
   protected final Configuration conf;
   protected final FileSystem fs;
+  protected boolean useTmpFileForPut;
 
   // Root directory for the task state store
   protected final String storeRootDir;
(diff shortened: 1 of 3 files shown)


chavdar:
Can you please explain what the race is?

zliu41:
@chavdar see https://jira01.corp.linkedin.com:8443/browse/ETL-4046

kadaan:
@zliu41 @chavdar We have seen a similar race which was causing our yarn gobblin jobs to fail.  What it looked like was the TaskStateCollectorService was trying to read a sequence file that had not been completely written.  We added retries to the deserialize and that seemed to fix it.

@zliu41 Can you share the pertinent info for those of us who can't access jira?

zliu41:
@kadaan In `FsStateStore.put()`, when it calls `SequenceFile.createWriter`,
the TaskState file will be created immediately. If this file is picked up
by `TaskStateCollectorService` before the `put` method returns, it will collect an empty file and subsequently causes EOFException.

kadaan:
That is exactly the problem we were seeing.  We added a retry with backoffs.

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -174,11 +172,13 @@ public void putAll(String storeName, String tableName, Collection<T> states)
     } finally {
       closer.close();
     }
+
+    Path tablePath = new Path(new Path(this.storeRootDir, storeName), tableName);
+    HadoopUtils.renamePath(this.fs, tmpTablePath, tablePath);
```
kadaan: Is this going to have the desired effect on filesystems other than HDFS (such as S3)?

kadaan: As an alternative, why not write to LocalFileSystem and then run a `moveFromLocalFile`?

ibuenros: @kadaan These files are temporary and only contain metadata. They are deleted after the Gobblin run. These files should not be placed in S3, instead they should go in the HDFS file system (even if it is ephemeral).

zliu41: As long as it is atomic and doesn't create a partial file at `tablePath`. Since this is renaming a single file, it should be fine.

kadaan: @ibuenros That is true for TaskState, but what about FsDatasetStateStore?

zliu41: There's no `put` method in `FsDatasetStateStore`.

kadaan: But it extends FsStateStore. I was wondering if @ibuenros statement about the transient nature of the data holds true for FsDatasetStateStore as well.

ibuenros: I have the same concerns as @kadaan . @zliu41 what are all of the uses of this method? and for `putAll`?

zliu41:
@ibuenros updated d33d7a3

zliu41:
Any other comments @ibuenros ?

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -52,8 +55,12 @@
  */
 public class FsStateStore<T extends State> implements StateStore<T> {
 
+  public static final String TMP_FILE_PREFIX = "_tmp_";
+  private static final ImmutableCollection<String> FS_SCHEMES_NO_TMP_FILE_FOR_PUT = ImmutableSet.<String> of("S3");
```
ibuenros: it should include all of "s3", "s3a", "s3n"

ibuenros:
just one comment, LGTM otherwise @kadaan what do you think of this approach?

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
kadaan: Why not just put all three in a treeset with a case insensitive comparitor?

kadaan:
Overall, looks good.  I worry about the proliferation of code that handles incompatibilities with hadoop filesystem implementations.  I think going forward it might be nice to have our own filesystem interface with which we can enforce stricter guarantees.  For instance we could require that an implementor if the interface indicate whether specific operations are atomic and take the appropriate action.

ibuenros:
I understand your point, but I don't think having our own interface is the correct solution: we would have to manually extend all implementations of hadoop's `FileSystem`, and it would be harder for users to benefit from already existing implementations.
Maybe we could have a `FileSystem` decorator instead with a few of the methods you mention. We could then detect certain known schemes with incompatibilities and apply fixes across the entire code, but any `FileSystem` we haven't considered yet would still work appropriately.

kadaan:
That would be great, but I think users of gobblin should be able to provide new decorators without having to extend gobblin itself.

PR #680, opened 2016-02-09
Fix a ConcurrentModificationException in MRCompactor

jeanrichard:
I am trying to use the MRCompactor. I have added a launcher class similar to ```AzkabanCompactionJobLauncher``` (not included in this PR). While testing it I ran into a ```ConcurrentModificationException``` in MRCompactor. Looking at the code, this seems to happen because we are mutating a ```HashSet``` in ```createJobPropsForDataset``` while iterating on it in ```createJobPropsForDatasets```. This PR attempts to fix this.


zliu41:
LGTM

PR #682, opened 2016-02-09
various bug fixes when integrate test with hdfs store

tuGithub:


tuGithub:
@chavdar , @sahilTakiar 

please review it when you have time

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -301,7 +301,11 @@ public Config getOwnConfig(ConfigKeyPath configKey, String version)
    * datasets.
    */
   private Path getDatasetDirForKey(ConfigKeyPath configKey, String version) throws VersionDoesNotExistException {
-    return new Path(getVersionRoot(version), getDatasetFromConfigKey(configKey));
+    String datasetFromConfigKey = getDatasetFromConfigKey(configKey);
+    if(datasetFromConfigKey==null || datasetFromConfigKey.trim().equals(""))
```
pcadabam: StringUtils.isNotBlank

gobblin-config-management/gobblin-config-client/src/main/resources/META-INF/services/gobblin.config.store.api.ConfigStoreFactory
```diff
@@ -0,0 +1 @@
+gobblin.config.store.hdfs.SimpleHDFSConfigStoreFactory
```
sahilTakiar: What's this file for?

sahilTakiar:
Pending Pradhan's comment, the changes to the `SimpleHDFSConfigStore` LGTM.

gobblin-config-management/gobblin-config-client/src/main/resources/META-INF/services/gobblin.config.store.api.ConfigStoreFactory
tuGithub: This is used for ServiceLoader to load service automatically

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
tuGithub: fixed

tuGithub:
@sahilTakiar 

Please review with the updated version

sahilTakiar:
Everything LGTM, but I don't understand what `getImportsRecursivelyIncludePhantomFromCache` is for can you add a more detailed explanation?

tuGithub:
@sahilTakiar 

Phantom node is the directory which does NOT exist on HDFS. For example, we have /data/databases in the ConfigStore, the phantom node may be /data/databases/foobar

The ConfigClient.getConfig(phantomNode) will return the resolved config of it's parent ( or ancestor )
So, just to make the logic in sync, the ConfigClient.getImports(phantomNode, true) , which return the imports recursively, need to return the imports recursively result of it's parent ( or ancestor )

sahilTakiar:
Why is that only applicable for `getImportsRecursively`?

sahilTakiar:
Discussed offline, LGTM

PR #674, opened 2016-02-08
Lifecycle Events

kadaan:
Add ability to define additional JobListeners in the job properties, added new JobListener lifecycle events, and moved the listener code into it's own package

PR #684, opened 2016-02-09
Add inline Hive registration to Gobblin job

zliu41:
Added a `BaseDataPublisherWithHiveRegistration` which records the paths that need to be registered as a state property. After `JobContext` commits the data, it will retrieve the paths from the state and register them in Hive.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -274,7 +274,14 @@ public class ConfigurationKeys {
    * Configuration properties for the data publisher.
    */
   public static final String DATA_PUBLISHER_PREFIX = "data.publisher";
+
+  /**
+   * @deprecated Use {@link #TASK_DATA_PUBLISHER_TYPE} and {@link #JOB_DATA_PUBLISHER_TYPE}.
+   */
+  @Deprecated
   public static final String DATA_PUBLISHER_TYPE = DATA_PUBLISHER_PREFIX + ".type";
+  public static final String JOB_DATA_PUBLISHER_TYPE = DATA_PUBLISHER_PREFIX + ".job.type";
+  public static final String TASK_DATA_PUBLISHER_TYPE = DATA_PUBLISHER_PREFIX + ".task.type";
   public static final String DEFAULT_DATA_PUBLISHER_TYPE = "gobblin.publisher.BaseDataPublisher";
   public static final String DATA_PUBLISHER_FILE_SYSTEM_URI = DATA_PUBLISHER_PREFIX + ".fs.uri";
   public static final String DATA_PUBLISHER_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".final.dir";
@@ -285,6 +292,7 @@ public class ConfigurationKeys {
   public static final String DATA_PUBLISHER_PERMISSIONS = DATA_PUBLISHER_PREFIX + ".permissions";
   public static final String PUBLISH_DATA_AT_JOB_LEVEL = "publish.data.at.job.level";
   public static final boolean DEFAULT_PUBLISH_DATA_AT_JOB_LEVEL = true;
+  public static final String PUBLISHER_DIRS = DATA_PUBLISHER_PREFIX + ".output.dirs";
 
   /**
    * Configuration properties used by the extractor.
@@ -513,10 +521,16 @@ public class ConfigurationKeys {
   public static final String AZKABAN_EXECUTION_DAYS_LIST = "azkaban.execution.days.list";
 
   /**
+   * Hive registration properties
+   */
+  public static final String HIVE_REGISTRATION_POLICY = "hive.registration.policy";
+
+  /**
    * Other configuration properties.
    */
   public static final String GOBBLIN_RUNTIME_DELIVERY_SEMANTICS = "gobblin.runtime.delivery.semantics";
   public static final Charset DEFAULT_CHARSET_ENCODING = Charsets.UTF_8;
   public static final String TEST_HARNESS_LAUNCHER_IMPL = "gobblin.testharness.launcher.impl";
   public static final int PERMISSION_PARSING_RADIX = 8;
+
 }
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -29,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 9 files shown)


zliu41:

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisherWithHiveRegistration.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  protected void movePath(ParallelRunner parallelRunner, Path src, Path dst, int branchId) throws IOException {
+    pathsToRegister.addAll(recordPathsToRegister(src, dst, branchId));
+    super.movePath(parallelRunner, src, dst, branchId);
+  }
+
+  @SuppressWarnings("deprecation")
+  protected Collection<Path> recordPathsToRegister(Path src, Path dst, int branchId) throws IOException {
```
ydai1124: I think the logic to determine which paths to be registered should be in source, when creating work units. Then this method can just fetch this information from each `workunitstate`.

zliu41: `Source` doesn't have the information of which paths the data is published to. For example, the data may be bucketed by the timestamp of each extracted record, which is not available to `Source`.

ydai1124:
Talked to Ziyang. LGTM.

kadaan:
One thing I would like to point out is that it would be nice to come up with a rich set of lifecycle events which can be used to perform functionality such as this.  That way every feature doesn't need to involve changes to AbstractJobLauncher, JobContext, etc.  Additionally, it would mean that users of gobblin could create functionality like this without having to modify gobblin.  As a simple example look at: #674.  The events added here have allowed us to get exactly once semantics for task based publishing (before you guys release your real solution for this).

zliu41:
Yes the observer pattern could be a good fit here if there are multiple events that need to be triggered by something. However Hive registration is not merely some independent observer here; it is a critical step (similar as publishing data) that needs to succeed before the dataset state can be persisted (otherwise the registration will be lost).

sahilTakiar:
Why does `JobContext` need to change? Why can't `BaseDataPublisherWithHiveRegistration` simply do the hive registration.

zliu41:
When publishing data in tasks, each mapper uses a publisher to publish its own data. I want to avoid creating 200+ connections to HiveMetaStore. Also, different mappers may publish to the same path so these paths will be registered multiple times.

kadaan:
How does work with exactly once semantics?

zliu41:
@kadaan another PR will deal with that. We probably need to make some changes to the `CommitStep` interface such as adding an argument (HiveMetaStoreClient) to the `execute()` method (otherwise everytime `execute()` is called it needs to create a new connection to Hive metastore). We are considering a few options.

kadaan:
Also, while I agree that for certain circumstances that HiveRegistration is critical, it is not a needed step for every gobblin use case.  I'm concerned about adding more and more logic to AbstractJobLauncher and JobContext that is not truly universal.

sahilTakiar:
@zliu41 I think @kadaan brings up a good point. Perhaps it's possible to let Gobblin specify both a Task-Level Publisher and a Job-Level Publisher.

That should (1) avoid any changes to `JobContext`, and (2) let the hive-registration code be run in the driver.

zliu41:
@sahilTakiar makes sense but the only problem is, the Job-level publisher will need to check whether it should publish data AND do hive registration, or only do Hive registration. Something like "if (data is already published in task level publisher) then (skip data publishing and only do Hive registration)". This is a little awkward. What if Hive registration is also done in the task level publisher?

sahilTakiar:
You brought up the point that too many connections to the Hive Metastore may be dangerous. Its a valid concern, although maybe we can ask hadoop-dev and see what they think. If we can make all the connections at once, then we can just do the hive-registration in the task-level publisher.

If not, then we can have a `HiveRegPublisher` that only does hive-registration, not any publishing of data. 

For Task-Level publishing in Kafka the data is published at a task level, and then registered at the job level using `HiveRegPublisher`.

For Job-Level publishing, we can modify the `BaseDataPublisher` to use some composition technique with the `HiveRegPublisher`. A config option can control if `BaseDataPublisher` invokes the `HiveRegPublisher.publish()` method.

zliu41:
Even if we can make that many Hive connections, if two mappers publish to the same path, they'll be trying to add the same partition simultaneously which will likely be problematic.

Your alternative approach sounds good

zliu41:
@sahilTakiar updated. My feeling is that this new approach is more cumbersome than just making the change to `JobContext`. Let me know what you think.

kadaan:
@zliu41 Which approach are you talking about #701?

zliu41:
No, #701 is irrelevant. I was referring to the previous approach in this PR (I squashed that commit) which added a `doHiveRegistration` method in `JobContext` after the data is published.

sahilTakiar:
Is `BaseDataPublisherDocumentingPathsPublished` necessary? Shouldn't the final output paths already be recorded in the `WorkUnitState`s? If they aren't we should just add the feature directly to `BaseDataPublisher`.

kadaan:
I don't think that the publishoutputpaths are recorded in the workunitstates, but I think it would be a good idea to store them there. It would be trivial to add them in BaseDataPublisher.

zliu41:
@sahilTakiar  removed `BaseDataPublisherDocumentingPathsPublished` and added that into `BaseDataPublisher`.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -274,7 +274,14 @@
...
    */
   public static final String DATA_PUBLISHER_PREFIX = "data.publisher";
+
+  /**
+   * @deprecated Use {@link #TASK_DATA_PUBLISHER_TYPE} and {@link #JOB_DATA_PUBLISHER_TYPE}.
+   */
+  @Deprecated
   public static final String DATA_PUBLISHER_TYPE = DATA_PUBLISHER_PREFIX + ".type";
+  public static final String JOB_DATA_PUBLISHER_TYPE = "job." + DATA_PUBLISHER_PREFIX + ".type";
+  public static final String TASK_DATA_PUBLISHER_TYPE = "task." + DATA_PUBLISHER_PREFIX + ".type";
```
sahilTakiar: Can you switch the order of `task.` and `DATA_PUBLISHER_PREFIX`, otherwise its not really a prefix. Same for the the `job` param above.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -285,6 +292,7 @@
   public static final String DATA_PUBLISHER_PERMISSIONS = DATA_PUBLISHER_PREFIX + ".permissions";
   public static final String PUBLISH_DATA_AT_JOB_LEVEL = "publish.data.at.job.level";
   public static final boolean DEFAULT_PUBLISH_DATA_AT_JOB_LEVEL = true;
+  public static final String PUBLISHER_DIRS = "publisher.output.dirs";
```
sahilTakiar: `publisher` -> `DATA_PUBLISHER_PREFIX`

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -286,10 +301,25 @@ protected void addWriterOutputToExistingDir(Path writerOutputDir, Path publisher
 
   protected void movePath(ParallelRunner parallelRunner, Path src, Path dst, int branchId) throws IOException {
     LOG.info(String.format("Moving %s to %s", src, dst));
+    this.publisherOutputDirs.addAll(recordPublisherOutptuDirs(src, dst, branchId));
     parallelRunner.movePath(src, this.publisherFileSystemByBranches.get(branchId), dst,
         this.publisherFinalDirOwnerGroupsByBranches.get(branchId));
   }
 
+  @SuppressWarnings("deprecation")
+  protected Collection<Path> recordPublisherOutptuDirs(Path src, Path dst, int branchId) throws IOException {
```
kadaan: recordPublisherOutptuDirs -> recordPublisherOutputDirs

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -286,10 +301,25 @@ protected void addWriterOutputToExistingDir(Path writerOutputDir, Path publisher
...
+
+    // Getting file status from src rather than dst, because at this time dst doesn't yet exist.
+    // If src is a dir, add dst to the set of paths. Otherwise, add dst's parent.
+    if (this.publisherFileSystemByBranches.get(branchId).getFileStatus(src).isDir()) {
+      LOG.info("Path to be registered: " + dst);
```
sahilTakiar: Probably don't need this log line or the one on line 318

sahilTakiar:
Mostly minor comments, approach LGTM

PR #685, opened 2016-02-09
Add wiki docs to Gobblin repo

zliu41:
I guess this is the way to make it possible to create PR on wiki docs, although I'm not 100% sure.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "gobblin-docs"]
+	path = gobblin-docs
+	url = https://github.com/linkedin/gobblin.wiki.git


abti:
LGTM, will wait for travis to finish before merge

kadaan:
I really have no experience here, but from what I read there are some downsides to submodule linking.  That said, either way submodule/subtree have specific steps that need to be done to keep things in sync.  Should those steps be detailed somewhere?

zliu41:
I've also heard bad things regarding submodules. I'll wait for someone more familiar with this stuff to confirm if this is the right approach.

kadaan:
http://stackoverflow.com/questions/6941688/how-to-integrate-a-github-wiki-into-the-main-project <http://stackoverflow.com/questions/6941688/how-to-integrate-a-github-wiki-into-the-main-project>

kadaan:
If we were to store the documentation within the gobblin repo, it would be nice to have the official documentation created with sphinx (we could use pandoc http://pandoc.org to convert from the current markdown files to reStructuredText files).  Hosting the documentation could be done on Read the Docs (ex. https://readthedocs.org/projects/airflow/).  Read the Docs can be triggered by git commits, supports multiple versions based on tags or branches, etc.

zliu41:
Thanks @kadaan . @sahilTakiar will have a look at these options and decide on the best approach.

PR #689, opened 2016-02-09
Increase the size of the property_value column in the history store from TEXT to MEDIUMTEXT

kadaan:
Certain properties such as `source.filebased.fs.snapshot` can be too large to fit in a TEXT column.

### DIFF
--- a/gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
+++ b/gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
@@ -21,7 +21,7 @@ CREATE TABLE IF NOT EXISTS gobblin_job_executions (
 	launched_tasks INT,
 	completed_tasks INT,
 	launcher_type ENUM('LOCAL', 'MAPREDUCE', 'YARN'),
-  tracking_url VARCHAR(512),
+	tracking_url VARCHAR(512),
 	created_ts TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
 	last_modified_ts TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
 	PRIMARY KEY (job_id),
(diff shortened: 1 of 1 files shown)


gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
```diff
@@ -91,29 +91,29 @@ CREATE TABLE IF NOT EXISTS gobblin_task_metrics (
...
 
 CREATE TABLE IF NOT EXISTS gobblin_job_properties (
-  property_id BIGINT(21) NOT NULL AUTO_INCREMENT,
-  job_id VARCHAR(128) NOT NULL,
-  property_key VARCHAR(128) NOT NULL,
-  property_value TEXT NOT NULL,
+	property_id BIGINT(21) NOT NULL AUTO_INCREMENT,
+	job_id VARCHAR(128) NOT NULL,
+	property_key VARCHAR(128) NOT NULL,
+	property_value MEDIUMTEXT NOT NULL,
```
abti: Please correct me if I am wrong, this (MEDIUMTEXT column type) will add about ~16 MB per record in the table irrespective of property value size. 

However, most of the property values will be much smaller < 100 bytes, so the table will grow quite rapidly for regularly running Gobblin instance. 

And for property: source.filebased.fs.snapshot, it seems to refer to all the files listed in current snapshot (before distributing them across workunits), which can shoot beyond ~16k files limit for a big dataset (assuming 100 bytes per file path) 

However, since this fix patches an issue which would otherwise break the history store quite easily, I think we can merge this in and create another issue to track the concerns listed above. Thoughts?

kadaan: AFAIK, this is not how the text column works.  In general, all flavors of the TEXT column (TINYTEXT, TEXT, MEDIUMTEXT, and LONGTEXT) store the data outside the row.  Additionally, they operate similar to VARCHAR columns in that they only use the amount of space needed to store the data plus some additional metadata.  Since the TEXT data is stored off row, there is some additional cost when reading the data.

On another note, because of how the current state system works, all job properties are duplicated on the tasks.  Task may then contain additional properties (or overrides).  I think it would be worthwhile to de-dup the task properties against the job properties when storing the data.  The down side is that any consumer of the history would need to know that the tasks have all the job properties.  The advantage is that it would greatly decrease the amount of data stored in the DB and returned by the REST service.  What do you think?

kadaan:
@liyinan926 You mind chiming in?

liyinan926:
I think it's ok to merge and I agree we should de-dup task properties that are already in job properties, which can be in a separate PR.

kadaan:
@liyinan926 I will provide a PR for the de-dupe.  

Also, I find that saving history is extremely slow.  I was looking into replacing all the EXISTS->INSERT/UPDATE code with an UPSERT.  To do this, I believe the indexes on the metric and property tables would need to change.  In my local tests, it seems much faster.  Additionally, I was looking at making the insert of metrics/property be done in batches of up to 1000.  What do you think?

kadaan:
BTW, do you guys use the history store in prod?

sahilTakiar:
Using batches and upserts sounds like a good idea to me.

gobblin-metastore/src/main/resources/gobblin_job_history_store.sql
sahilTakiar: de-dups will help. It will also help if we stop adding all global state props to each `WorkUnitState`. I opened a ticket for addressing this: https://github.com/linkedin/gobblin/issues/696

sahilTakiar:
@abti any other comments? I think this should be good to merge.

kadaan:
@abti @sahilTakiar I'd like to close this one in favor of #699

PR #690, opened 2016-02-10
Add missing jar to MR runner script

muratoda:
The script was failing with an `Error: java.lang.ClassNotFoundException: com.typesafe.config.ConfigFactory`. Adding the missing dependency to the list of included libraries solved the issue.


sahilTakiar:
LGTM

sahilTakiar:
Created a follow up ticket to avoid having to do this in the future https://github.com/linkedin/gobblin/issues/693

PR #691, opened 2016-02-10
Fix permissions for directories in distcp.

ibuenros:
Issue:
Permissions were not being set correctly for newly generated directories in distcp. Root cause:
1. fs.mkdirs(Path, FsPermission) is affected by umask.
2. we were only setting permissions up to the dataset root. However, for patterns like `/foo/*/*`, where the dataset would be /foo/bar/dataset, we should also set permissions for /foo/bar and /foo/bar/dataset if these directories are newly created.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
@@ -219,7 +219,7 @@ public class CopyableFile implements File, HasGuid {
       List<OwnerAndPermission> ancestorOwnerAndPermissions = Lists.newArrayList();
       try {
         Path currentPath = PathUtils.getPathWithoutSchemeAndAuthority(path);
-        while (currentPath != null && currentPath.getParent() != null && !currentPath.getParent().equals(rootPathTmp)) {
+        while (currentPath != null && currentPath.getParent() != null) {
           currentPath = currentPath.getParent();
           final Path thisPath = currentPath;
           OwnerAndPermission ownerAndPermission = this.configuration.getCopyContext().getOwnerAndPermissionCache()
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
@@ -38,14 +38,12 @@ import gobblin.data.management.copy.writer.FileAwareInputStreamDataWriter;
 import gobblin.data.management.copy.CopyableFile;
 import gobblin.data.management.copy.writer.FileAwareInputStreamDataWriterBuilder;
 import gobblin.publisher.UnpublishedHandling;
-import gobblin.util.PathUtils;
 import gobblin.instrumented.Instrumented;
 import gobblin.metrics.GobblinMetrics;
 import gobblin.metrics.MetricContext;
 import gobblin.metrics.event.EventSubmitter;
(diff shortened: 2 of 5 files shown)


ibuenros:

ibuenros:
or @pcadabam ?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -219,7 +219,7 @@ public CopyableFile build() throws IOException {
       List<OwnerAndPermission> ancestorOwnerAndPermissions = Lists.newArrayList();
       try {
         Path currentPath = PathUtils.getPathWithoutSchemeAndAuthority(path);
-        while (currentPath != null && currentPath.getParent() != null && !currentPath.getParent().equals(rootPathTmp)) {
+        while (currentPath != null && currentPath.getParent() != null) {
```
ydai1124: Why removing the rootPath check?

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -146,13 +144,12 @@ private void publishFileSet(CopyableFile.DatasetAndPartition datasetAndPartition
...
         datasetWorkUnitStates.iterator().next().getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
-    Path datasetWriterOutputPath = new Path(new Path(writerOutputDir, datasetAndPartition.identifier()),
-        PathUtils.withoutLeadingSeparator(metadata.getDatasetTargetRoot()));
+    Path datasetWriterOutputPath = new Path(this.writerOutputDir, datasetAndPartition.identifier());
 
     log.info(String
         .format("Publishing fileSet from %s to %s", datasetWriterOutputPath, metadata.getDatasetTargetRoot()));
 
-    HadoopUtils.renameRecursively(fs, datasetWriterOutputPath, metadata.getDatasetTargetRoot());
+    HadoopUtils.renameRecursively(fs, datasetWriterOutputPath, findPathRoot(metadata.getDatasetTargetRoot()));
```
ydai1124: Does this change the publish behavior? Shouldn't we publish to the DatasetTargeRoot path defined from source?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
ydai1124: Nvmd. Please ignore this comment.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
ibuenros: It doesn't, line 147 was changed as well so the publish behavior is similar. The directory structure in dataset output mimics the directory structure in the file system. Previously we would move at the dataset level, so if dataset is /foo/bar/dataset, but /foo/bar didn't exist, it would be created automatically with default permissions (because we attempt to move .../staging/foo/bar/dataset to /foo/bar/dataset). 
Now we are moving at the highest level possible, i.e. highest level directory that doesn't exist in the target (we will move /staging/foo/bar to /foo/bar). This has the effect of creating any missing directories with the appropriate permissions.

ydai1124: I see.

ydai1124:
LGTM.

PR #692, opened 2016-02-10
Removing compile dependency on JCE

sahilTakiar:
Will address users who are hitting the error in https://github.com/linkedin/gobblin/issues/615


sahilTakiar:

PR #694, opened 2016-02-10
Adding throttled upload of sonatype packages

chavdar:
This is a hack to address the issue with the split repositories when uploading packages to Sonatype/Maven . This should be considered as a temp solution as we need to maintain the list of packages to upload manually.

@ibuenros @pcadabam can you review ?


upload-to-sonatype-hadoop1.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo "CLEANING"
+./gradlew clean
+
+for P in :gobblin-admin :gobblin-api :gobblin-azkaban :gobblin-compaction :gobblin-config-management:gobblin-config-core :gobblin-config-management:gobblin-config-client :gobblin-core :gobblin-data-management :gobblin-distribution :gobblin-example :gobblin-metastore :gobblin-metrics :gobblin-oozie :gobblin-rest-service:gobblin-rest-api :gobblin-rest-service:gobblin-rest-client :gobblin-rest-service:gobblin-rest-server :gobblin-runtime :gobblin-salesforce :gobblin-scheduler :gobblin-test :gobblin-test-harness :gobblin-utility :gobblin-yarn ; do 
+    echo "----------------------------" 
+    read -p "UPLOAD $P (y/n)" ans
```
ibuenros: Is this intended to be manual throttling? e.g. user waits to reply yes?

chavdar: No, this was meant:
(1) Give opportunity to check that there are no split sessions
(2) Skip packages that have already been uploaded 
(3) Allow to release individual packages.

(3) is not strictly necessary. It is possible to let packages accumulate in the same session and do a single release at the end.

ibuenros: Where exactly is the throttling then?

ibuenros: Maybe there should be an option "yes to all".

ibuenros:
LGTM

pcadabam:
LGTM

PR #699, opened 2016-02-11
Improve performance of history store

kadaan:
Improve performance of writing to the history store by using upserts, batching, and removing duplicate data.


kadaan:
@sahilTakiar @abti @liyinan926 Can you guys review this please?

PR #700, opened 2016-02-11
Add missing jars to gobblin mapreduce runner, sort.

muratoda:


bin/gobblin-mapreduce.sh
```diff
@@ -136,13 +137,16 @@ LIBJARS=(
...
-  $FWDIR_LIB/joda-time-2.9.jar
-  $FWDIR_LIB/data-1.15.9.jar
   $FWDIR_LIB/commons-lang3-3.4.jar
   $FWDIR_LIB/config-1.2.1.jar
+  $FWDIR_LIB/data-1.15.9.jar
+  $FWDIR_LIB/gson-2.3.1.jar
+  $FWDIR_LIB/joda-time-2.9.jar
+  $FWDIR_LIB/kafka-clients-0.8.2.1.jar
+  $FWDIR_LIB/metrics-core-2.2.0.jar
+  $FWDIR_LIB/metrics-core-3.1.0.jar
```
ibuenros: Why are both versions needed?

muratoda: The runner uses the `com.yammer.metrics.Metrics` class in the 2.x version. That has been removed in the 3.x version, with the functionality distributed to other classes in a new `com.codahale.metrics` package. Shall we modify the gobblin runtime code to only use the newer Metrics version instead?

ibuenros: That would be ideal, yes, but I don't know the level of effort. I'll just merge this for now, and leave the removal of 2.2.0 as a TODO.

PR #701, opened 2016-02-11
Reg compacted datasets in Hive

sahilTakiar:
* Initial PR to discuss the approach for adding Hive-Reg for datasets produced by `MRCompactor`
* Registration is done in a `AfterCompactionAction` once compaction has completed
* The `AfterCompactionAction`s are configurable and are invoked in `JobRunnerExecutor.afterExecute`

@zliu41 can you review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
@@ -58,6 +58,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 
 import gobblin.compaction.Compactor;
+import gobblin.compaction.listeners.CompactorListener;
 import gobblin.compaction.dataset.Dataset;
 import gobblin.compaction.dataset.DatasetsFinder;
 import gobblin.compaction.dataset.TimeBasedSubDirDatasetsFinder;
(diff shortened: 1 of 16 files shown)


zliu41:
This creates a HiveMetaStoreClient connection for each dataset. Can you create a HiveRegister object in MRCompactor and use it to register all datasets?

sahilTakiar:
@zliu41 addressed comments and modified approach slightly

* Instead of making a listener specific to `MRCompactor` I made a generic `CompactorListener`
* Added factory interfaces and implementations for creating `Compactor`s and `CompactorListener`s
* Re-factored `AzkabanCompactionJobLauncher` a bit to account for changes
* Main logic for this feature is in the class `HiveRegistrationCompactorListener`

zliu41:
We need a solution for retrying failed Hive registration, to avoid a dataset being published but not registered with Hive. Ideally this should leverage exactly-once, but since it likely won't be ready by Q1, a temp solution is needed, e.g., write failed dataset paths to an HDFS location which can be picked up by the next run to retry.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
```diff
@@ -788,6 +792,13 @@ protected void afterExecute(Runnable r, Throwable t) {
             // Set the dataset status to COMPACTION_COMPLETE if compaction is successful.
             jobRunner.getDataset().setState(COMPACTION_COMPLETE);
           }
+          if (MRCompactor.this.compactorListener.isPresent()) {
```
zliu41: This seems to allow only one listener. Can you make it possible to have multiple listeners?

sahilTakiar: Updated, it can now run multiple listeners serially.

sahilTakiar:
Yes, I will focus on the retry logic in the next PR as it may require some additional design work.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactor.java
zliu41: What is the change? seems there is still only one listener here. I thought `MRCompactor` should contain a collection of listeners.

sahilTakiar: There is a new class `SerialCompactorListener` that takes in a `List` of `CompactorListener`s and executes them one by one. This class is used by `ReflectionCompactorListenerFactory`.

zliu41:
LGTM, please add copyright block to all new files

PR #703, opened 2016-02-12
Remove scheme and authority from path before relativizing in version …

ibuenros:
…finder.


ibuenros:
Minor fix, self-approving.

PR #704, opened 2016-02-12
Remove @link tags from @see javadoc tags

kadaan:
Remove @link tags from @see javadoc as they are not needed and result in a bunch of warnings like:
```
/home/travis/build/kadaan/gobblin/gobblin-utility/src/main/java/gobblin/util/PathUtils.java:52: warning - Tag @see:illegal character: "123" in "{@link Path}, {@link URI}"
```


sahilTakiar:
LGTM thanks for making the fix

PR #706, opened 2016-02-12
Fixing CliOptions config file fs

sahilTakiar:
Fixing bug reported in https://github.com/linkedin/gobblin/issues/702

* The `JobLauncherUtils.fileToProperties` method takes a path to a props file and reads it using a `FileSystem` object
* The `FileSystem` object is created by invoking `path.getFileSystem(Configuration)`
* It seems that if the `path` has no scheme + authority the default `FileSystem` is used
* The default fs when running locally is the local fs, the default fs when running on a cluster is the HDFS fs
* This can be confusing as the script expects the config file to be in different filesystems depending on where things are run
* Behavior is changed so if no scheme + authority is specified the local fs is used by default, if users expect the files to be on HDFS then they should set the scheme to `hdfs`
** This should be much more intuitive as most of the time props files will be on the local fs, not on HDFS

@zliu41 can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/JobConfigurationUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/JobConfigurationUtils.java
@@ -13,15 +13,19 @@
 package gobblin.util;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Properties;
 
 import org.apache.commons.configuration.ConfigurationConverter;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.PropertiesConfiguration;
+
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 
 import gobblin.configuration.State;
-import org.apache.hadoop.fs.Path;
 
 
 /**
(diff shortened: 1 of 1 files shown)


sahilTakiar:
@zliu41 can you take a look at this PR? We have another user waiting for this to be fixed: https://groups.google.com/forum/#!topic/gobblin-users/297nRNlhuUs

gobblin-utility/src/main/java/gobblin/util/JobConfigurationUtils.java
```diff
@@ -78,10 +82,19 @@ public static void putStateIntoConfiguration(State state, Configuration configur
...
+  public static Properties fileToProperties(String fileName)
+      throws IOException, ConfigurationException, URISyntaxException {
+
     PropertiesConfiguration propsConfig = new PropertiesConfiguration();
-    propsConfig.load(filePath.getFileSystem(new Configuration()).open(filePath));
+    Configuration conf = new Configuration();
+    Path filePath = new Path(fileName);
+    URI fileURI = filePath.toUri();
+
+    if (fileURI.getScheme() == null && fileURI.getAuthority() == null) {
```
zliu41: Should this be `||` ? if either scheme or authority is null, local should be used

sahilTakiar: It's the same logic the Hadoop `FileSystem` uses to check whether or not it should use the default scheme: http://grepcode.com/file/repo1.maven.org/maven2/org.apache.hadoop/hadoop-common/2.3.0/org/apache/hadoop/fs/FileSystem.java#356

bin/gobblin-mapreduce.sh
```diff
@@ -18,6 +18,7 @@ function print_usage(){
   echo "  --jt <job tracker / resource manager URL>      Job submission URL: if not set, taken from \${HADOOP_HOME}/conf"
   echo "  --fs <file system URL>                         Target file system: if not set, taken from \${HADOOP_HOME}/conf"
   echo "  --jars <comma-separated list of job jars>      Job jar(s): if not set, \"$FWDIR_LIB\" is examined"
+  echo "  --conf <job conf file>                         Job configuration file: if not set, taken from \${JOB_CONFIG_FILE}"
```
zliu41: `--conf` doesn't seem to be optional, according to line 81.
If it is, then line 16 should be revised.

sahilTakiar: Updated

sahilTakiar:
@zliu41 updated

zliu41:
LGTM

PR #711, opened 2016-02-16
Add kafka simple json source

muratoda:
A json kafka source extending from KafkaSimpleSource to output kafka messages as json with offset, key and payload.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
@@ -1,0 +1,87 @@
+package gobblin.source.extractor.extract.kafka;
+
+public class KafkaRecord {
+
+    private long offset;
+    private String key;
+    private String payload;
+
+    public KafkaRecord(long offset, String key, String payload) {
+        super();
+        this.offset = offset;
+        this.key = key;
+        this.payload = payload;
+    }
+
+    public long getOffset() {
+        return offset;
+    }
+
+    public void setOffset(long offset) {
+        this.offset = offset;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public void setKey(String key) {
+        this.key = key;
+    }
+
+    public String getPayload() {
+        return payload;
+    }
+
+    public void setPayload(String payload) {
+        this.payload = payload;
+    }
+
+    @Override
+    public String toString() {
+        return "KafkaRecord [offset=" + offset + ", key=" + key + ", payload=" + payload + "]";
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((key == null) ? 0 : key.hashCode());
+        result = prime * result + (int) (offset ^ (offset >>> 32));
+        result = prime * result + ((payload == null) ? 0 : payload.hashCode());
+        return result;
+    }
+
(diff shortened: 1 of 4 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleJsonExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+
+import gobblin.configuration.WorkUnitState;
+import gobblin.source.extractor.Extractor;
+import gobblin.source.extractor.extract.kafka.KafkaSimpleExtractor;
+import kafka.message.MessageAndOffset;
+
+public class KafkaSimpleJsonExtractor extends KafkaSimpleExtractor implements Extractor<String, byte[]> {
+
+    private static final Gson gson = new Gson();
+    private static final Charset CHARSET = StandardCharsets.UTF_8;
```
muratoda: Or should I just use DefaultCharset?

sahilTakiar: Using `UTF_8` is fine, we follow a similar approach in the rest of the code.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
```diff
@@ -0,0 +1,87 @@
+package gobblin.source.extractor.extract.kafka;
+
+public class KafkaRecord {
+
+    private long offset;
+    private String key;
+    private String payload;
+
+    public KafkaRecord(long offset, String key, String payload) {
+        super();
```
sahilTakiar: Is this needed?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
```diff
@@ -0,0 +1,87 @@
+package gobblin.source.extractor.extract.kafka;
+
+public class KafkaRecord {
```
zliu41: Please make this class an inner class of `KafkaSimpleJsonExtractor`

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private long offset;
```
zliu41: Please make these fields final, and remove the setter methods since they are not needed.
Also you can replace the getter methods with a @Getter annotation above line 3.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaRecord.java
muratoda: I am actually using this domain class in another application that reads these generated json files to write them back to Kafka. It would be very convenient if I (or any other application that would read them) didn't have to define it again.

I think it could also be utilised by a future `KafkaDataWriter` as well?

sahilTakiar:
LGTM @zliu41 any other comments?

chavdar:
The PR had to be reverted because it broke the build.

muratoda:
Apologies for that. There was an error with package naming when I merged the files from my local project. I'll raise it again.

PR #714, opened 2016-02-16
Modify HiveSpec interface

zliu41:
Previously the `HiveSpec` interface had `getDbName()`, `getTableName()` and `getSd()`, so one can specify sd-related properties in the sd, but it was not possible to specify table properties that are not in the sd, such as retention, privilege, table type, etc.

The new `HiveSpec` interface has a `getTable()` method that returns a `Table` object containing all these properties. 

Also, `HiveRegistrationPolicy` interface is modified to return a collection of `HiveSpec`s, rather than a single `HiveSpec`, because a path may correspond to zero or multiple `HiveSpec`s.


zliu41:

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -362,41 +311,33 @@ private void addOrAlterPartition(IMetaStoreClient client, Table table, HiveParti
       try {
         Partition existingPartition =
             client.getPartition(table.getDbName(), table.getTableName(), partition.getValues());
-        if (needToUpdatePartition(existingPartition, newPartition)) {
-          client.alter_partition(table.getDbName(), table.getTableName(), newPartition);
+        partition.setCreateTime(existingPartition.getCreateTime());
```
ydai1124: What if no existing partition?

gobblin-hive-registration/src/main/java/gobblin/hive/spec/HiveSpec.java
```diff
@@ -33,25 +33,15 @@
   public Path getPath();
 
   /**
-   * Get the Hive database name.
+   * Get the Hive {@link Table}.
    */
-  public String getDbName();
+  public Table getTable();
```
ydai1124: Can you add more doc here indicating that `Table` returned here is not registered yet?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: `client.getPartition` will throw `NoSuchObjectException`

build.gradle
```diff
@@ -104,7 +104,7 @@ if (!project.hasProperty('bytemanVersion')) {
   ext.bytemanVersion = '2.2.1'
 }
 
-ext.avroVersion = '1.7.7'
+ext.avroVersion = '1.8.0'
```
sahilTakiar: Has version `1.8.0` been tested? Release notes say it may contain incompatible changes: https://issues.apache.org/jira/browse/AVRO?selectedTab=com.atlassian.jira.jira-projects-plugin:changelog-panel

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicy.java
```diff
@@ -30,9 +31,9 @@
 public interface HiveRegistrationPolicy {
 
   /**
-   * Get a {@link HiveSpec} for a {@link Path}, which can be used by {@link HiveRegister}
+   * Get a collection of {@link HiveSpec} for a {@link Path}, which can be used by {@link HiveRegister}
    * to register the given {@link Path}.
    */
-  public HiveSpec getHiveSpec(Path path) throws IOException;
+  public Collection<HiveSpec> getHiveSpec(Path path) throws IOException;
```
sahilTakiar: `getHiveSpec` to `getHiveSpecs`

sahilTakiar:
Minor comments, otherwise LGTM.

build.gradle
zliu41: Ok, since there are backward incompatible changes I will change it back.

ydai1124:
LGTM

PR #717, opened 2016-02-17
More flexible Admin configuration

kadaan:
1. Allow specifying the host for the AdminWebServer in the configuration file.
1. Allow the JobExecutionInfoServer to advertise a different URI from what it is listening on 
1. Fix one of the cdn dependencies to use the https scheme so that it will work correctly when using SSL/TLS.

_Rational:_ This enables us to put the Admin UI and Rest service behind an nginx proxy


kadaan:
@ibuenros Can you review this?

ibuenros:
LGTM

PR #721, opened 2016-02-17
Fixing build in CliOptions

sahilTakiar:


sahilTakiar:
Trivial change self approving

PR #723, opened 2016-02-18
Remove distcp reliance on dataset root.

ibuenros:
A lot of small refactoring to remove reliance of distcp on the concept of a dataset root.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
@@ -13,6 +13,7 @@
 package gobblin.data.management.copy;
 
 import gobblin.data.management.dataset.DatasetUtils;
+import gobblin.dataset.FileSystemDataset;
 import gobblin.util.PathUtils;
 import gobblin.util.FileListUtils;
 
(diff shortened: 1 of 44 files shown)


ibuenros:

ibuenros:
@ydai1124 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/dataset/DatasetUtils.java
```diff
@@ -63,7 +63,7 @@ public boolean accept(Path path) {
    * @throws IOException
    */
   @SuppressWarnings("unchecked")
-  public static <T extends Dataset> DatasetFinder<T> instantiateDatasetFinder(Properties props, FileSystem fs,
+  public static <T extends gobblin.dataset.Dataset> DatasetFinder<T> instantiateDatasetFinder(Properties props, FileSystem fs,
```
ydai1124: Since it has been changed to generic `dataset`, the constructor still expects for `FileSystem`, which implies it to be a `FileBasedDataset`. Are you planning to change `DatasetFinder` to be more generic as well?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -33,31 +34,41 @@
...
 
+    Path nonGlobSearchPath = PathUtils.deepestNonGlobPath(this.glob);
+
     List<FileStatus> files = FileListUtils.listFilesRecursively(this.fs, this.rootPath, this.pathFilter);
 
     List<CopyableFile> copyableFiles = Lists.newArrayList();
 
     for (FileStatus file : files) {
-      copyableFiles.add(CopyableFile.builder(this.fs, file, this.rootPath, configuration).fileSet(file.getPath().getParent().toString()).build());
+      Path filePathRelativeToSearchPath = PathUtils.relativizePath(PathUtils.getPathWithoutSchemeAndAuthority(file.getPath()),
```
ydai1124: This line has been used at several places and it is quite long. Can we create `PathUtils.relativizePathWithoutSchemeAndAuthority(Path, Path)` for it?

gobblin-data-management/src/main/java/gobblin/data/management/dataset/DatasetUtils.java
ibuenros: That is an interesting point, and we should probably do that in the future. However, I would want to push that to a different PR.

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
ibuenros: I changed the semantics of `PathUtils.relativizePath` instead (in reality, we've always assumed the new semantics).

ydai1124:
LGTM.

PR #727, opened 2016-02-19
Modified to add a configuration to skip previous run during FileBasedExtraction for full load

pldash:


sahilTakiar:
It would be better if you can make this modification `Source` agnostic so we don't have to add similar features to each `Source`.

Basically, a config called `state.store.ignore.previous.state`. If set to try to returns in an empty `Collection` of `WorkUnitState`s when `getPreviousWorkUnitStates` is invoked.

It should be a simple change to the `JobState` class. Specifically, on the following line: https://github.com/linkedin/gobblin/blob/master/gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java#L139

The line would be changed to something like:

```
if (jobPropsState.contains(ConfigurationKeys.SOURCE_STATE_IGNORE_PREVIOUS_STATE, false)) {
  this.jobState = new JobState(jobPropsState, new HashMap<>(), this.jobName, this.jobId);
} else {
  this.jobState = new JobState(jobPropsState, this.datasetStateStore.getLatestDatasetStatesByUrns(this.jobName), this.jobName, this.jobId);
```

@chavdar thoughts?

pcadabam:
@sahilTakiar can we use workunit.retry.policy = WorkUnitRetryPolicy.NEVER?

chavdar:
@pldash Can you remind me again what the usecase is?

pldash:
@chavdar This is to skip reading the previous workunits when we are trying to do a full load regularly and the source file is always the same. Every day the same file is updated with the latest snapshot.

chavdar:
@sahilTakiar I don't like the idea of such an option. This looks a big gun for people to shoot themselves in the foot. I don't see a reason to ignore the previous state. 

In this particular case, actually I think it is a short-coming of how state is stored currently for the FileBasedSource . The file name is not sufficient. We also need the file modification timestamp. Then the source should compare both names and timestamps of previously processed files and currently existing files.

One possible approach is to
* Define a class TimestampedFile contains of a file path string and a long last modification timestamp
* Define equals(), hashValue() in the natural way using both the path and the timestamp.
* A TimestampedFile is serialized to a string as a "<path>:::<timestamp>". 
* Similarly, a TimestampedFile can be created for the above string. For backwards compatibility, if :::<timestamp> is missing, we can assume 0 timestamp.
* ReUse the existing SOURCE_FILEBASED_FS_SNAPSHOT property but make it a list the string serializations of TimestampedFile
* getcurrentFsSnapshot() should return a Collection<TimestampedFile>
* When computing filesToPull we should use a HashSet instead of the current List to make removeAll() efficient.
* If we define TimestampedFile.equals() as above, the removeAll() will do the right thing and keep a file for pulling if the name has remained the same but the modification timestamp has changed.

sahilTakiar:
Agreed, didn't realize that was the use case.

Was this closed intentionally?

pldash:
@sahilTakiar Sorry i had closed this by mistake.

pldash:
@chavdar Is it really required to create a new class TimestampedFile. We can just get the list of files along with modified time in the currentFsSnapshot. Finally after removing the prevFsSnapshot we can split the name to pick out only the filename.

chavdar:
@pldash While it is not required, it makes the code a lot more readable. Anyway, if we have PR without that, I can review it.

pldash:
@chavdar Is it really required to create a new class TimestampedFile. We can just get the list of files along with modified time in the currentFsSnapshot. Finally after removing the prevFsSnapshot we can split the name to pick out only the filename.

chavdar:
LGTM. @sahilTakiar  ?

gobblin-core/src/main/java/gobblin/source/extractor/filebased/TimestampAwareFileBasedHelper.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.source.extractor.filebased;
+
+/**
+ * A File helper that extends {@link SizeAwareFileBasedHelper} and has functionality to get file size.
```
sahilTakiar: `to get file size` to `get file modified timestamp`

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
```diff
@@ -172,11 +180,13 @@ protected void initLogger(SourceState state) {
     try {
       log.info("Running ls command with input " + path);
       results = this.fsHelper.ls(path);
+      for (int i = 0; i < results.size(); i++) {
+        String filePath = state.getProp(ConfigurationKeys.SOURCE_FILEBASED_DATA_DIRECTORY) + "/" + results.get(i);
+        results.set(i, filePath + ":::" + this.fsHelper.getFileMTime(filePath));
```
sahilTakiar: Can you make `:::` a constant, and use that constant here and in the changes to `getWorkunits`

sahilTakiar:
Minor comments, in general LGTM

sahilTakiar:
@pldash merged

Issue #731, opened 2016-02-19
PasswordManager doesn't work when used under gobblin-yarn


kadaan:
The current implementation is PassowrdManager doesn't work correctly when used under gobblin-yarn. Unless the encryption file is stored on a distributed filesystem, it will not be available to the driver, app master, or container.  The encryption file can be copied to the appmaster and containers by using the gobblin.yarn.app.master.files.local  and gobblin.yarn.container.files.local properties.  After doing that the encryption file is available on every yarn node,but the PasswordManager doesn't have the correct path to the file.  

@sahilTakiar @zliu41 what is the best way to ensure that all machines in the cluster can decrypt encrypted settings?

kadaan:
@sahilTakiar @zliu41 Figured out why it wasn't working and have submitted a PR to fix it.

PR #733, opened 2016-02-21
Add ability to configure the encryption_key_loc filesystem

kadaan:
Fix for #731

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -501,6 +501,7 @@ public class ConfigurationKeys {
   /**
    * Password encryption and decryption properties.
    */
+  public static final String ENCRYPT_KEY_FS_URI = "encrypt.key.fs.uri";
   public static final String ENCRYPT_KEY_LOC = "encrypt.key.loc";
   public static final String ENCRYPT_USE_STRONG_ENCRYPTOR = "encrypt.use.strong.encryptor";
   public static final boolean DEFAULT_ENCRYPT_USE_STRONG_ENCRYPTOR = false;
--- a/gobblin-api/src/main/java/gobblin/password/PasswordManager.java
+++ b/gobblin-api/src/main/java/gobblin/password/PasswordManager.java
@@ -15,6 +15,7 @@ package gobblin.password;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.URI;
 import java.util.AbstractMap;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


kadaan:
# New Configuration
`encrypt.key.fs.uri` is optional.  If it is not provided, then the code will perform as it did before.

# Example Configuration
To allow encrypted password to be used when gobblin in run under yarn, you need to make sure that the master password file is copied to the app_master and containers.  This is done by adding `/opt/apps/gobblin/current/master.pass` to `gobblin.yarn.app.master.files.local` and `gobblin.yarn.container.files.local`.  Additionally, `encrypt.key.fs.uri` needs to specify the local filesystem `"file:///"` and `encrypt.key.loc` need specify the master password file at the root of the current directory `./master.pass`.

See the example below:
```
gobblin.yarn.app.master.files.local=${gobblin.yarn.conf.dir}"/log4j-yarn.properties,"${gobblin.yarn.conf.dir}"/application.conf,"${gobblin.yarn.conf.dir}"/reference.conf,{{ app_current_dir }}/master.pass"
gobblin.yarn.container.files.local=${gobblin.yarn.app.master.files.local}
...
encrypt.key.fs.uri="file:///"
encrypt.key.loc=./master.pass
```

kadaan:
@sahilTakiar @zliu41 Can you please review this?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -501,6 +501,7 @@
   /**
    * Password encryption and decryption properties.
    */
+  public static final String ENCRYPT_KEY_FS_URI = "encrypt.key.fs.uri";
```
chavdar: I am not sure we need a new configuration property. ENCRYPT_KEY_LOC can just contain a qualified path.

kadaan:
### The master.pass file exists in the current directory
screen shot 2016-02-25 at 1 01 43 pm

### But I can't get it like this
screen shot 2016-02-25 at 1 02 00 pm

### Or like this
screen shot 2016-02-25 at 1 02 31 pm

### But this works
screen shot 2016-02-25 at 1 03 10 pm

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
kadaan: @chavdar See the comment below.

chavdar:
OK. Digging through the Hadoop source code I found:

    /**
     * Pathnames with scheme and relative path are illegal.
     * @param path to be checked
     */
    void checkNotSchemeWithRelative() {

LGTM.

PR #736, opened 2016-02-22
Automatically enable JMX and pick an open port.

kadaan:
# New Configuration

- `gobblin.yarn.app.master.jmx.enabled` _(optional)_  If set to true, JMX will be enabled on the app_master on a randomly picked port.
- `gobblin.yarn.container.jmx.enabled` _(optional)_ If set to true, JMX will be enabled on the containers each with a randomly picked port.
- `gobblin.yarn.jmx.configuration` _(optional)_ If set, the value will be used when specifying the jmx configuration for containers or app_master which have jmx enabled.  The token `$JMX_PORT` will be replaced with a randomly selected port. _Default_: `-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.port=$JMX_PORT -Dcom.sun.management.jmxremote.rmi.port=$JMX_PORT`

# New Logging
JVM input arguments are logged on app_master and container start.  These can be used to determine the JMX port for the 
```
2016-02-22 09:31:13 UTC INFO  [main] gobblin.yarn.GobblinWorkUnitRunner  - JVM Input Arguments: -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.port=56298 -Dcom.sun.management.jmxremote.rmi.port=56298 -Xmx1216M
```

# Usage
With the default JMX configuration, to connect an app_master or container for which JMX is enabled:

1. Check the logs to find the JXM port by specified in the line starting with: `JVM Input Arguments:`
2. Open an ssh tunnel: `ssh -nNT -L <JMX_PORT>:localhost:<JMX_PORT> <USER>@<HOST>`
3. Connect to `localhost:<JMX_PORT>` with VisualVM or a similar tool.

### DIFF
--- a/gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
@@ -545,6 +545,8 @@ public class GobblinApplicationMaster extends GobblinYarnLogSource {
       Log4jConfigurationHelper.updateLog4jConfiguration(
           GobblinApplicationMaster.class, Log4jConfigurationHelper.LOG4J_CONFIGURATION_FILE_NAME);
 
+      LOGGER.info(YarnHelixUtils.getJvmInputArguments());
+
       ContainerId containerId =
           ConverterUtils.toContainerId(System.getenv().get(ApplicationConstants.Environment.CONTAINER_ID.key()));
       GobblinApplicationMaster applicationMaster =
--- a/gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
+++ b/gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
@@ -481,6 +481,8 @@ public class GobblinWorkUnitRunner extends GobblinYarnLogSource {
       Log4jConfigurationHelper.updateLog4jConfiguration(GobblinWorkUnitRunner.class,
           Log4jConfigurationHelper.LOG4J_CONFIGURATION_FILE_NAME);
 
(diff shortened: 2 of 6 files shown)


kadaan:
I was thinking about maybe doing this a different way.  What if we recognized certain "tokens" in the app_master and container jvm arguments and replaced them when creating the java command.  The first token would be "$PORT_X" where X is a number > 0.  We would identify all such tokens and replace them with a random port.  All instances of the token which share the same X would have the same port.  This would be much more generic as I could then use YourKit, JMX, etc with the app_master and containers.

ibuenros:
@kadaan In my experience, using random ports is not a good idea. Applications connecting to Gobblin with JMX will have a hard time figuring out the port.
I would suggest that user provides the port, and a port selection policy: either fail if the port is not available (default) or look for an available port in case of failure.

kadaan:
For this reason, I have added logging of the JVM input arguments when the app_master and container start. That way it is easy to figure out what the port will be.

That said, maybe the “token” should be `$PORT_<REQUESTED_PORT>` like `$PORT_10001`.  If port 10001 is open we take it.  If not we try 10002.  And so on.  It will still be “random” in Yarn because there are multiple container on the same node, so the JVM input argument logging will still be needed.  

As a longer term item, it would be really nice to be able to “classify” the ports and expose them via the admin UI.  That way I could see a list of nodes, see what containers exist, see the “open” ports, etc.

PR #737, opened 2016-02-22
Better travis scripts which support test error reporting

kadaan:


kadaan:
@sahilTakiar Can you speed this one through.  I pulled it out of #699 so that we could gain the benefit before @eogren finishes reviewing.

sahilTakiar:
LGTM

kadaan:
Thanks!

Issue #718, opened 2016-02-17
Publishing fails if same named file exists


kadaan:
If a file already exists, in the publish location, with the same name as the file being published, then the publish step will fail.  This will result in the workunit being retried on the next run of the job.  This pattern will keep repeating.

I propose that the following methods be changed to overwrite files with the same names:

1. `HadoopUtils.copy()`
1. `HadoopUtils.move()`
1. `HadoopUtils.rename()`

What do you think?

sahilTakiar:
I think adding a parameter `boolean overwrite` would be good, so that users of this method explicitly know what behavior they are signed up for when invoking the method.

kadaan:
If I add this, I want to add overwrite=true to the calls from BaseDataPublisher.

sahilTakiar:
Sure, but maybe we can make it a config option for `BaseDataPublisher` if it overwrites data or not. I would prefer it to be set to false by default so that users don't accidentally overwrite data that already exists.

kadaan:
Sure

PR #741, opened 2016-02-23
Fix #740 for FsStateStore.createAlias and removing usage of FileUtil.copy

### LINKED ISSUE #740: Fix Usage of FileUtil.copy() and FsStateStore.createAlias() function
`FsStateStore.createAlias` can leave a partial `current.jst` file if the function is interrupted during the execution of `FileUtil.copy`. Looking at the code in `FileUtil.copy` the code doesn't handle what happens if the copy is interrupted:

```
try {
  in = srcFS.open(src);
  out = dstFS.create(dst, overwrite);
  IOUtils.copyBytes(in, out, conf, true);
} catch (IOException e) {
  IOUtils.closeStream(out);
  IOUtils.closeStream(in);
  throw e;
}
```

Here are all the places `FileUtil.copy()` is being used:

```
./gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java: if (!FileUtil.copy(fs, src, fs, dst, false, fs.getConf())) {
./gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java: if (!FileUtil.copy(srcFs, src, dstFs, dst, true, false, dstFs.getConf())) {
./gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java: FileUtil.copy(this.fs, originalTablePath, this.fs, aliasTablePath, false, true, this.conf);
./gobblin-compaction/src/main/java/gobblin/compaction/hive/HdfsWriter.java: FileUtil.copy(fs, path, fs, new Path(destination), false, true, getConfiguration());
```

sahilTakiar:
* Fix for #740 
* Adds a overloaded `copyPath` method that allows writing data first to a temp file, and then committing the data using `renamePath`
    * This should fix the problems with `FsStateStore.createAlias`
* Removed all direct usages of `FileUtil.copy` with `HadoopUtils.copyPath`
* Unit tests are coming

@zliu41 can you review?

### DIFF
--- a/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
@@ -12,6 +12,8 @@
 
 package gobblin.metastore;
 
+import static gobblin.util.HadoopUtils.FS_SCHEMES_NON_ATOMIC;
+
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
@@ -20,15 +22,12 @@ import java.util.List;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.FileUtil;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.SequenceFile;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.compress.DefaultCodec;
 
 import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableCollection;
-import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
@@ -56,8 +55,6 @@ import gobblin.util.HadoopUtils;
 public class FsStateStore<T extends State> implements StateStore<T> {
 
   public static final String TMP_FILE_PREFIX = "_tmp_";
-  private static final ImmutableCollection<String> FS_SCHEMES_NO_TMP_FILE_FOR_PUT =
-      ImmutableSortedSet.<String> orderedBy(String.CASE_INSENSITIVE_ORDER).add("s3").add("s3a").add("s3n").build();
 
   protected final Configuration conf;
   protected final FileSystem fs;
@@ -72,14 +69,14 @@ public class FsStateStore<T extends State> implements StateStore<T> {
   public FsStateStore(String fsUri, String storeRootDir, Class<T> stateClass) throws IOException {
     this.conf = new Configuration();
     this.fs = FileSystem.get(URI.create(fsUri), this.conf);
-    this.useTmpFileForPut = !FS_SCHEMES_NO_TMP_FILE_FOR_PUT.contains(this.fs.getUri().getScheme());
+    this.useTmpFileForPut = !FS_SCHEMES_NON_ATOMIC.contains(this.fs.getUri().getScheme());
     this.storeRootDir = storeRootDir;
     this.stateClass = stateClass;
   }
 
   public FsStateStore(FileSystem fs, String storeRootDir, Class<T> stateClass) throws IOException {
     this.fs = fs;
-    this.useTmpFileForPut = !FS_SCHEMES_NO_TMP_FILE_FOR_PUT.contains(this.fs.getUri().getScheme());
+    this.useTmpFileForPut = !FS_SCHEMES_NON_ATOMIC.contains(this.fs.getUri().getScheme());
     this.conf = this.fs.getConf();
     this.storeRootDir = storeRootDir;
     this.stateClass = stateClass;
@@ -89,7 +86,7 @@ public class FsStateStore<T extends State> implements StateStore<T> {
     this.conf = new Configuration();
     Path storePath = new Path(storeUrl);
     this.fs = storePath.getFileSystem(this.conf);
-    this.useTmpFileForPut = !FS_SCHEMES_NO_TMP_FILE_FOR_PUT.contains(this.fs.getUri().getScheme());
+    this.useTmpFileForPut = !FS_SCHEMES_NON_ATOMIC.contains(this.fs.getUri().getScheme());
     this.storeRootDir = storePath.toUri().getPath();
     this.stateClass = stateClass;
   }
@@ -279,9 +276,10 @@ public class FsStateStore<T extends State> implements StateStore<T> {
     }
 
     Path aliasTablePath = new Path(new Path(this.storeRootDir, storeName), alias);
+    Path tmpAliasTablePath = new Path(aliasTablePath.getParent(), new Path(TMP_FILE_PREFIX, aliasTablePath.getName()));
     // Make a copy of the original table as a work-around because
     // Hadoop version 1.2.1 has no support for symlink yet.
-    FileUtil.copy(this.fs, originalTablePath, this.fs, aliasTablePath, false, true, this.conf);
+    HadoopUtils.copyFile(this.fs, originalTablePath, this.fs, aliasTablePath, tmpAliasTablePath, true, this.conf);
   }
 
   @Override
--- a/gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
@@ -12,18 +12,23 @@
 
 package gobblin.util;
 
-import gobblin.configuration.ConfigurationKeys;
-import gobblin.configuration.State;
-import gobblin.writer.DataWriter;
-
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map.Entry;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Lists;
+import com.google.common.io.BaseEncoding;
+import com.google.common.io.Closer;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -35,13 +40,13 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.FileUtil;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.io.IOUtils;
 import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.util.ReflectionUtils;
 
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
-import com.google.common.io.BaseEncoding;
-import com.google.common.io.Closer;
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+import gobblin.writer.DataWriter;
 
 
 /**
@@ -51,6 +56,21 @@ import com.google.common.io.Closer;
 public class HadoopUtils {
 
   public static final String HDFS_ILLEGAL_TOKEN_REGEX = "[\\s:\\\\]";
+
+  /**
+   * A {@link Collection} of all known {@link FileSystem} schemes that do not support atomic renames or copies.
+   *
+   * <p>
+   *   The following important properties are useful to remember when writing code that is compatible with S3:
+   *   <ul>
+   *     <li>Renames are not atomic, and require copying the entire source file to the destination file</li>
+   *     <li>Writes to S3 using {@link FileSystem#create(Path)} will first go to the local filesystem, when the stream
+   *     is closed the local file will be uploaded to S3</li>
+   *   </ul>
+   * </p>
+   */
+  public static final Collection<String> FS_SCHEMES_NON_ATOMIC = ImmutableSortedSet.orderedBy(
+      String.CASE_INSENSITIVE_ORDER).add("s3").add("s3a").add("s3n").build();
 
   public static Configuration newConfiguration() {
     Configuration conf = new Configuration();
@@ -101,6 +121,16 @@ public class HadoopUtils {
     }
   }
 
+  /**
+   * A wrapper around {@link FileSystem#delete(Path, boolean)} that only deletes a given {@link Path} if it is present
+   * on the given {@link FileSystem}.
+   */
+  public static void deleteIfExists(FileSystem fs, Path path, boolean recursive) throws IOException {
+    if (fs.exists(path)) {
+      deletePath(fs, path, recursive);
+    }
+  }
+
   public static void deletePathAndEmptyAncestors(FileSystem fs, Path f, boolean recursive) throws IOException {
     deletePath(fs, f, recursive);
     Path parent = f.getParent();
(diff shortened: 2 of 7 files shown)


gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+
+    if (!srcFs.exists(src)) {
+      throw new IOException(String.format("Cannot copy from %s to %s because src does not exist", src, dst));
+    }
+    try {
+      if (!FileUtil.copy(srcFs, src, dstFs, dst, deleteSource, overwriteDst, conf)) {
+        throw new IOException(String.format("Failed to copy %s to %s", src, dst));
+      }
+    } catch (IOException e1) {
+      FileSystem deleteFs;
```
zliu41: `FileUtil.copy` will throw `IOException` if `overwriteDst=false` and `dst` already exists. In this case you shouldn't attempt to delete `dst`

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+        throw new IOException(String.format("Cannot copy %s to %s because the dst exists", src, dst));
+      }
+      copyPath(srcFs, src, dstFs, tmp, false, overwriteTmp, conf);
+      try {
+        if (overwriteDst && dstFs.exists(dst)) {
+          deletePath(dstFs, dst, true);
+        }
+        renamePath(dstFs, tmp, dst);
+      } finally {
+        deletePath(dstFs, tmp, false);
```
zliu41: why should `recursive=false` here? If `src` is a directory it will throw IOException

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+      }
+      if (deleteSource) {
+        srcFs.delete(src, false);
```
zliu41: same question here

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
   }
 
   /**
-   * A wrapper around {@link HadoopUtils#renamePath(FileSystem, Path, Path)} and
-   * {@link FileUtil#copy(FileSystem, Path, FileSystem, Path, boolean, Configuration)} which will rename the path
-   * if the src and dst filesystems are the same; otherwise, the src file will be moved the the dst filesystem.
-   * An {@link IOException} if {@link FileUtil#copy(FileSystem, Path, FileSystem, Path, boolean, Configuration)} returns false.
+   * Copies a src {@link Path} from a srcFs {@link FileSystem} to a dst {@link Path} on a dstFs {@link FileSystem}. If
+   * either the srcFs or dstFs are S3 {@link FileSystem}s (as dictated by {@link #S3_SCHEMES}) then data is directly
+   * copied from the src to the dst. Otherwise data is first copied to a tmp {@link Path}, which is then renamed to the
```
zliu41: It's better to explain here why S3 doesn't need tmp files.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -52,6 +54,21 @@
...
   public static final String HDFS_ILLEGAL_TOKEN_REGEX = "[\\s:\\\\]";
 
+  /**
+   * A {@link Collection} of all S3 schemes.
+   *
+   * <p>
+   *   The following important properties are useful to remember when writing code that is compatible with S3:
+   *   <ul>
+   *     <li>Renames are not atomic, and require copying the entire source file to the destination file</li>
+   *     <li>Writes to S3 will first go to the local filesystem, when the stream is closed the local file will be
```
kadaan: This is technically not true if you use s3a and call moveFromLocalFile.  In that case, the file is not written to a temp file locally and then copied up.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+   * dst.
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the {@link Path} to copy from the source {@link FileSystem}
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to copy data to
+   * @param tmp the temporary {@link Path} to use when copying data
+   * @param deleteSource true if the source should be deleted after the copy is complete, false otherwise
+   * @param overwriteDst true if the destination should be overwritten, false otherwise
+   * @param overwriteTmp true if the temp file should be overwritten, false otherwise
```
kadaan: Why would you ever choose overwriteTmp=true, but overwriteDst=false?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+   * either the srcFs or dstFs are S3 {@link FileSystem}s (as dictated by {@link #S3_SCHEMES}) then data is directly
+   * copied from the src to the dst. Otherwise data is first copied to a tmp {@link Path}, which is then renamed to the
+   * dst.
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the {@link Path} to copy from the source {@link FileSystem}
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to copy data to
+   * @param tmp the temporary {@link Path} to use when copying data
+   * @param deleteSource true if the source should be deleted after the copy is complete, false otherwise
```
kadaan: I know that hadoop does it this way, but I find it a little disgusting that copyPath has a bool that changes it to movePath.  To reduce duplication, I would prefer a common private method and clean versions of movePath and copyPath without the cruft of the deleteSource parameter.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -52,6 +54,21 @@
...
+   * <p>
+   *   The following important properties are useful to remember when writing code that is compatible with S3:
+   *   <ul>
+   *     <li>Renames are not atomic, and require copying the entire source file to the destination file</li>
+   *     <li>Writes to S3 will first go to the local filesystem, when the stream is closed the local file will be
+   *     uploaded to S3</li>
+   *   </ul>
+   * </p>
+   */
+  public static final Collection<String> S3_SCHEMES = ImmutableSortedSet.orderedBy(
```
kadaan: One issue here is that I think what we are really trying to identify are schemes which support atomic move/copy.  

This reminds me of a conversation with @ibuenros:
-- ibuenros
...
Maybe we could have a FileSystem decorator instead with a few of the methods you mention. We could then detect certain known schemes with incompatibilities and apply fixes across the entire code, but any FileSystem we haven't considered yet would still work appropriately.
  
-- kadaan
That would be great, but I think users of gobblin should be able to provide new decorators without having to extend gobblin itself.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+   * Copies a src {@link Path} from a srcFs {@link FileSystem} to a dst {@link Path} on a dstFs {@link FileSystem}. If
+   * either the srcFs or dstFs are S3 {@link FileSystem}s (as dictated by {@link #S3_SCHEMES}) then data is directly
+   * copied from the src to the dst. Otherwise data is first copied to a tmp {@link Path}, which is then renamed to the
+   * dst.
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the {@link Path} to copy from the source {@link FileSystem}
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to copy data to
+   * @param tmp the temporary {@link Path} to use when copying data
```
kadaan: It would be really nice if the caller didn't need to pass in a temp path.  For instance, if I'm developing and only work against s3, then the temp path is not something I ever need.  The opposite is true for you guys because you use HDFS.  Could we allow temp paths by filesystem to be specified in the global configuration?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +149,113 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+   * </p>
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the {@link Path} to copy from the source {@link FileSystem}
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to copy data to
+   * @param deleteSource true if the source should be deleted after the copy is complete, false otherwise
+   * @param overwriteDst true if the destination should be overwritten, false otherwise
+   */
+  public static void copyPath(FileSystem srcFs, Path src, FileSystem dstFs, Path dst, boolean deleteSource,
```
kadaan: As stated above, I really think that this should be transparent for users/developers. We should default to the safest, but possibly less performant, methods of performing actions.  In the case of filesystems, we should be able to decorate specific filesystems with more performant versions.  For instance, the base copy method could be the local->temp->dest version.  But, the s3* filesystems could be decorated to perform local->dest.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: Good point.

I am going to replace `FileUtil.copy` completely because the method does not differentiate between errors occurring during copying of data vs. illegal arguments (both throw `IOException`)

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: Agreed, will make the changes.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: I agree this is the approach we probably want to go with. I will create a follow up issue for this, but its going to require a little more effort, so lets do it in a different PR. I think the work in this PR will help do some of the initial work for this also.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
sahilTakiar: Yes, it seems that the HDFS FileSystem originally tried to do something similar and had to make it configurable: https://issues.apache.org/jira/browse/HADOOP-11281

A `hdfs dfs -cp [src] [dst]` also uses a tmp file. It first copies data to `[dst]._COPYING_` and then does a rename. Seems there is also an option to disable this so that data is directly copied from the `src` to the `dst` (useful for S3). Unfortunately, the code doesn't seem to be publicly available.

sahilTakiar:
I deleted the `copyPath` using tmp file method because its going to more effort than I originally anticipated, I still thinks it may be useful in the future, but is not strictly necessary for fixing #740

sahilTakiar:
Updated. @zliu41 can you take another look?

zliu41:
Please replace `FS_SCHEMES_NO_TMP_FILE_FOR_PUT` in `FsStateStore` with the one defined in `HadoopUtils`

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -281,7 +280,7 @@ public void createAlias(String storeName, String original, String alias) throws
     Path aliasTablePath = new Path(new Path(this.storeRootDir, storeName), alias);
     // Make a copy of the original table as a work-around because
     // Hadoop version 1.2.1 has no support for symlink yet.
-    FileUtil.copy(this.fs, originalTablePath, this.fs, aliasTablePath, false, true, this.conf);
+    HadoopUtils.copyPath(this.fs, originalTablePath, this.fs, aliasTablePath, this.conf);
```
zliu41: Since `HadoopUtils.copyPath` can no longer overwrite, you need to first delete `current.jst`. However, if `current.jst` is deleted and then `copyPath` failed, there will be no `current.jst` and the next run will have no state.

kadaan: I concur.  Both copypath and movepath should have an overwrite option. It would also take care of the groundwork for #718.

sahilTakiar: I agree that these methods should have an `overwrite` flag, but adding it is a more work than I though it would be. I think we would have to replace `FileUtil.copy`. The problem is that `FileUtil.copy` has a few pre-condition checks that it does and if any fail it throws an `IOException`. It also throws on `IOException` if there is a problem copying the actual data. The problem is that there is no way to differentiate the two scenarios.

One of the requirements of `HadoopUtils.copyPath` should be that it doesn't leave partial data - it doesn't copy part of a file, fail and then quit. This can cause `EOFException`s (which is the original motivation for this PR), which is why in the `catch` block the the `dst` is deleted. This is harder when `overwrite=true` as the method should not delete the `dst` if a pre-condition fails.

sahilTakiar: For `FsStateStore` I changed it to write to a tmp path and then do a rename operation.

sahilTakiar:
@zliu41 updated

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -279,9 +276,13 @@ public void createAlias(String storeName, String original, String alias) throws
     }
 
     Path aliasTablePath = new Path(new Path(this.storeRootDir, storeName), alias);
+    Path tmpAliasTablePath = new Path(aliasTablePath.getParent(), new Path(TMP_FILE_PREFIX, aliasTablePath.getName()));
     // Make a copy of the original table as a work-around because
     // Hadoop version 1.2.1 has no support for symlink yet.
-    FileUtil.copy(this.fs, originalTablePath, this.fs, aliasTablePath, false, true, this.conf);
+    HadoopUtils.deleteIfExists(this.fs, tmpAliasTablePath, true);
+    HadoopUtils.copyPath(this.fs, originalTablePath, this.fs, tmpAliasTablePath, this.conf);
+    HadoopUtils.deleteIfExists(this.fs, aliasTablePath, true);
```
zliu41: If this line succeeds and the next line fails the state will be lost. One solution is to modify `FsDatasetStateStore.getLatestDatasetStatesByUrns` such that if `current.jst` doesn't exist, it will look for the most recent jst file.

sahilTakiar: True. I would rather not have the `StateStore` default to the most recently created file as that violates the contract of the `get(storeName, tableName, stateId)` method which is suppose to either return the `State` associated with `tableName` or `null` if it does not exist.

Will try to think of a safer approach.

sahilTakiar:
@zliu41 updated the `FsStateStore.createAlias` function

sahilTakiar:
Actually, I'm not sure how this is going to solve the scenario where Azkaban executes a SLA kill on the job. Looking at `MRJobLauncher.cancelJob` only tries to kill the Hadoop job, but doesn't try to cancel anything else like the `Publisher` or `StateStore` serialization. So if `cancelJob` is invoked while the alias is created, the alias creation will probably continue to run until Azkaban just invokes a `kill -9` on the process. And we will just hit the same `EOFException`.

Any ideas? If we could use a RDMS instead I feel like DB transactions would make this a lot easier.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -132,29 +162,93 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+  public static void copyFile(FileSystem srcFs, Path src, FileSystem dstFs, Path dst, boolean overwrite,
+      Configuration conf) throws IOException {
+
+    Preconditions.checkArgument(srcFs.isFile(src),
+        String.format("Cannot copy from %s to %s because src is not a file", src, dst));
+    Preconditions.checkArgument(overwrite || !dstFs.exists(dst),
+        String.format("Cannot copy from %s to %s because dst exists", src, dst));
+
+    try (InputStream in = srcFs.open(src); OutputStream out = dstFs.create(dst, overwrite)) {
+      IOUtils.copyBytes(in, out, conf, false);
```
zliu41: This seems to have the same problems as before. If this line fails, `current.jst` will either be a partially copied file (if line 247 is not executed, e.g., SLA kill or OOM) or gone.

There doesn't seem to be an easy way to guarantee that if the new `current.jst` isn't successfully generated, the previous `current.jst` must be restored. So I still think modifying `FsDatasetStateStore.getLatestDatasetState` is a better approach. I don't think it violates any contract, because `current.jst` is just a convenient way to find the latest `jst` file (so that we don't need to list all `jst` files and sort them). If somehow `current.jst` failed to be generated, I don't see a problem of using the latest `jst` file.

sahilTakiar: The problem is that the javadoc of `StateStore` doesn't make any mention of using `current`. So if I want to implement a new `StateStore` called `MySqlStateStore`, I have to know that I also need to handle the scenario that if `current` is passed in as the `tableName` that the application is expecting the `MySqlStateStore` to return the latest `State`.

A user who just looks at the `StateStore` interface wouldn't know that. I think adding `StateStore.getLatest` would be a better way of doing this.

I'm not sure how much this matters considering `StateStore` is used in a very specific way, but lets just sync up tomorrow after the stand up.

sahilTakiar:
@zliu41 updated, the only real change is that `copyPath` now catches `Throwable` instead of `Exception`, and I added a shutdown hook in `AbstractJobLauncher` to interrupt the main thread.

Adding a feature to `FsDatasetStateStore` to treat a `tableName` as `current` equivalent to the latest `.jst` files is slightly more difficult that originally thought, and the solution is a little hacky. The methods `getLatestDatasetState` and `getLatestDatasetStatesByUrns` would both have to be modified. The problem is that each dataset will have its own `-current.jst` file, I can explain more in person.

zliu41:
If we keep `current.jst`, then if copy fails, the previous `current.jst` needs to be restored. This doesn't seem to be the case in the current approach.

sahilTakiar:
@zliu41 `HadoopUtils.copyFile` has been updated

zliu41:
LGTM

PR #743, opened 2016-02-23
Fixing skipTestGroup option

sahilTakiar:
* Fixing option to skip specific test groups
* Currently, `travis/test.sh` skips tests with group `disabledOnTravis`, but this feature is not currently working, this PR should fix that

@zliu41 can you review?


zliu41:
LGTM

Issue #739, opened 2016-02-23
Gobblin Metrics race condition results in metric reporter failing to start


kadaan:
@ibuenros There is a race condition in the Gobblin Metrics initialization of `ScheduledReporter`.  The problem is the order of the following lines:
```java
public ScheduledReporter(String name, Config config) {
    super(name, config);
    this.executor = Executors.newSingleThreadScheduledExecutor(
        ExecutorsUtils.newDaemonThreadFactory(Optional.of(log), Optional.of("metrics-" + name + "-scheduler")));
    ...
```

The following stack show the problem:

1. `ScheduledReporter.init<>()`
1. `ContextAwareReporter.init<>()`
1. `RootMetricContext.addNewReporter()`
1. `ContextAwareReporter.start()`
1. `ScheduledReporter.startImpl()`

At this point, `ScheduledReporter.startImpl()` attempts to use `this.executor` and finds that it has not been initialized.

```java
this.scheduledTask = Optional.<ScheduledFuture>of(this.executor.scheduleAtFixedRate(new Runnable() {
  @Override public void run() {
    report();
  }
}, 0, this.reportingPeriodSeconds, TimeUnit.SECONDS));
```

sahilTakiar:
Good catch. Looks like this will only come up if `RootMetricContext.addNewReporter` is invoked after `startReporting` is invoked (because of the flag `RootMetricContext.reportingStarted`), which is probably why we have never seen this before. Should definitely be fixed because we should be able to add reporters even after reporting has started.

sahilTakiar:
Fixed in https://github.com/linkedin/gobblin/pull/747

sahilTakiar:
@kadaan re-opening this because #747 is causing a FindBugs warning to be triggered:

```
File: ScheduledReporter.java, Line: 109, Type: DC_DOUBLECHECK, Priority: Normal, Category: CORRECTNESS
Possible doublecheck on gobblin.metrics.reporter.ScheduledReporter.metricFilter in gobblin.metrics.reporter.ScheduledReporter.ensureMetricFilterIsInitialized(Config)

This method may contain an instance of double-checked locking.  This idiom is not correct according to the semantics of the Java memory model.  For more information, see the web page http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html.
```

I think the solution is to add `synchronized` to the method `ensureMetricFilterIsInitialized`

sahilTakiar:
[...]

PR #747, opened 2016-02-23
Fix for #739

kadaan:
@ydai1124 Fix for #739

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
@@ -59,6 +59,7 @@ public abstract class ScheduledReporter extends ContextAwareReporter {
    */
   public static final String REPORTING_INTERVAL = "reporting.interval";
   public static final String DEFAULT_REPORTING_INTERVAL_PERIOD = "1M";
+  private static final Object syncObj = new Object();
 
   public static final PeriodFormatter PERIOD_FORMATTER = new PeriodFormatterBuilder().
       appendHours().appendSuffix("H").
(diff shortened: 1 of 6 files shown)


gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
```diff
@@ -93,18 +93,14 @@ public static Config setReportingInterval(Config config, long reportingInterval,
...
   private Optional<ScheduledFuture> scheduledTask;
   private int reportingPeriodSeconds;
 
   public ScheduledReporter(String name, Config config) {
     super(name, config);
-    this.executor = Executors.newSingleThreadScheduledExecutor(
-        ExecutorsUtils.newDaemonThreadFactory(Optional.of(log), Optional.of("metrics-" + name + "-scheduler")));
-    this.reportingPeriodSeconds = parsePeriodToSeconds(
-        config.hasPath(REPORTING_INTERVAL) ? config.getString(REPORTING_INTERVAL) : DEFAULT_REPORTING_INTERVAL_PERIOD);
     this.metricFilter = createMetricFilter(this.config);
```
sahilTakiar: Can we move the initialization of this to `startImpl` I believe it will be used by the `report` method which is invoked by the thread pool.

kadaan: I had `createMetricFilter` in startImpl first, but `report` can be called without being kicked off from the threadpool.  This ensures that `report` will work without start being called first.  One scenario which was breaking was the tests for kafka reporter.

sahilTakiar: I see, but won't `metricFilter` hit the same issue that `executor` and `reportingPeriodSeconds` were facing? It won't be initialized in time and will probably hit a NPE.

While not the most elegant solution, we could set `this.metricFilter = createMetricFilter(this.config);` in both places and modifying `createMetricFilter` to include the below code:

```
synchronized (this.metricFilterLock) {
  if (this.metricFilter == null) {
    this.metricFilter = createMetricFilter(this.config);
  }
}
```

kadaan: Are you worried that the scheduled task will fire before the constructor is completed?

sahilTakiar: Yeah, that was my main concern. Its unlikely to happen, but it will be safer.

kadaan:
@sahilTakiar Updated

sahilTakiar:
LGTM

PR #752, opened 2016-02-24
Allow ports in the jvm arguments to be automatically picked.

kadaan:
@ibuenros @sahilTakiar Can you review?
### Usage

This allows the user of gobblin to specify JVM arguments to the app_master and containers and have gobblin automatically pick ports where needed.  This is done by detecting a specific token within the jvm argument string.  This token is in any of the following forms:
1. `${PORT_1234}` : Try to get port 1234 and fail if it is not available
2. `${PORT_1234+}`: Try to get any port from the specified port inclusive to the maximum port (65535) inclusive
3. `${PORT_?}`: Try to get a random port between the minimum port (1025) inclusive and the maximum port (65535) inclusive

Gobblin will attempt to pick the specified port, but if it is not available Gobblin will successively try each high port until an open one is found. The actual JVM argument script will be logged as the first log line of the app_master and containers.
### Example

```
gobblin.yarn.app.master.jvm.args="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=${PORT_5005+}"
```

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/PortUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/PortUtils.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import com.google.common.collect.Maps;
+
+import java.net.ServerSocket;
+import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class PortUtils {
+  public static final int MINIMUM_PORT = 1025;
+  public static final int MAXIMUM_PORT = 65535;
+  private static final Pattern PORT_REGEX =
+          Pattern.compile("\\$\\{PORT_(?>(?>\\?(\\d+))|(?>(\\d+)\\?)|(\\d+|\\?))\\}");
+  private final PortLocator portLocator;
+  private final ConcurrentMap<Integer, Boolean> assignedPorts;
+
+  public PortUtils() {
+    this(new ServerSocketPortLocator());
+  }
+
+  @VisibleForTesting
+  PortUtils(PortLocator locator) {
+    this.portLocator = locator;
+    this.assignedPorts = Maps.newConcurrentMap();
+  }
+
+  /**
+   * Replaces any port tokens in the specified string.
+   *
+   * NOTE: Tokens can be in the following forms:
+   *   1. ${PORT_123}
+   *   2. ${PORT_?123}
+   *   3. ${PORT_123?}
+   *   4. ${PORT_?}
+   *
+   * @param value The string in which to replace port tokens.
+   * @return The replaced string.
+   */
+  public String replacePortTokens(String value) {
+    BiMap<String, Optional<Integer>> portMappings = HashBiMap.create();
+    Matcher regexMatcher = PORT_REGEX.matcher(value);
+    while (regexMatcher.find()) {
+      String token = regexMatcher.group(0);
+      if (!portMappings.containsKey(token)) {
+        Optional<Integer> portStart = Optional.absent();
+        Optional<Integer> portEnd = Optional.absent();
+        String unboundedStart = regexMatcher.group(1);
+        if (unboundedStart != null) {
+          int requestedEndPort = Integer.parseInt(unboundedStart);
+          Preconditions.checkArgument(requestedEndPort <= PortUtils.MAXIMUM_PORT);
+          portEnd = Optional.of(requestedEndPort);
+        } else {
+          String unboundedEnd = regexMatcher.group(2);
+          if (unboundedEnd != null) {
+            int requestedStartPort = Integer.parseInt(unboundedEnd);
+            Preconditions.checkArgument(requestedStartPort >= PortUtils.MINIMUM_PORT);
+            portStart = Optional.of(requestedStartPort);
+          } else {
+            String absolute = regexMatcher.group(3);
+            if (!"?".equals(absolute)) {
+              int requestedPort = Integer.parseInt(absolute);
+              Preconditions.checkArgument(requestedPort >= PortUtils.MINIMUM_PORT &&
+                      requestedPort <= PortUtils.MAXIMUM_PORT);
+              portStart = Optional.of(requestedPort);
+              portEnd = Optional.of(requestedPort);
+            }
+          }
+        }
+        Optional<Integer> port = takePort(portStart, portEnd);
+        portMappings.put(token, port);
+      }
+    }
+    for (Map.Entry<String, Optional<Integer>> port : portMappings.entrySet()) {
+      if (port.getValue().isPresent()) {
+        value = value.replace(port.getKey(), port.getValue().get().toString());
+      }
+    }
+    return value;
+  }
+
+  /**
+   * Finds an open port. {@param portStart} and {@param portEnd} can be absent
+   *
+   * ______________________________________________________
+   * | portStart | portEnd  | takenPort                   |
+   * |-----------|----------|-----------------------------|
+   * |  absent   | absent   | random                      |
+   * |  absent   | provided | 1024 < port <= portEnd      |
+   * |  provided | absent   | portStart <= port <= 65535  |
+   * |  provided | provided | portStart = port = portEnd  |
+   * ------------------------------------------------------
+   *
+   * @param portStart the inclusive starting port
+   * @param portEnd the inclusive ending port
+   * @return The selected open port.
+   */
+  private synchronized Optional<Integer> takePort(Optional<Integer> portStart, Optional<Integer> portEnd) {
+    if (!portStart.isPresent() && !portEnd.isPresent()) {
+      for (int i = 0; i < 65535; i++) {
+        try {
+          int port = this.portLocator.random();
+          Boolean wasAssigned = assignedPorts.putIfAbsent(port, true);
+          if (wasAssigned == null || !wasAssigned) {
+            return Optional.of(port);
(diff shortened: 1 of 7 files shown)


gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -545,6 +545,8 @@ public static void main(String[] args) throws Exception {
       Log4jConfigurationHelper.updateLog4jConfiguration(
           GobblinApplicationMaster.class, Log4jConfigurationHelper.LOG4J_CONFIGURATION_FILE_NAME);
 
+      LOGGER.info(YarnHelixUtils.getJvmInputArguments());
```
ibuenros: Is this necessary?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
```diff
@@ -481,6 +481,8 @@ public static void main(String[] args) throws Exception {
       Log4jConfigurationHelper.updateLog4jConfiguration(GobblinWorkUnitRunner.class,
           Log4jConfigurationHelper.LOG4J_CONFIGURATION_FILE_NAME);
 
+      LOGGER.info(YarnHelixUtils.getJvmInputArguments());
```
ibuenros: Is this necessary?

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
```diff
@@ -199,4 +206,48 @@ public static void addFileAsLocalResource(FileSystem fs, Path destFilePath, Loca
...
+        try (ServerSocket ignored = new ServerSocket(port)) {
+          return Optional.of(port);
+        } catch (IOException ignored) {
+        }
+      }
+    }
+    return Optional.absent();
+  }
+
+  public static String getJvmInputArguments() {
```
ibuenros: Can you add javadoc? Especially to clarify what `getJvmInputArguments` does vs `getJvmArguments`

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
```diff
@@ -199,4 +206,48 @@ public static void addFileAsLocalResource(FileSystem fs, Path destFilePath, Loca
...
+    if (jvmArguments.isPresent()) {
+      return replacePortTokens(jvmArguments.get());
+    }
+    return StringUtils.EMPTY;
+  }
+
+  private static String replacePortTokens(String value) {
+    Map<String, Optional<Integer>> portMappings = Maps.newHashMap();
+    Matcher regexMatcher = PORT_REGEX.matcher(value);
+    while (regexMatcher.find()) {
```
ibuenros: Please document this algorithm.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
kadaan: Yes.  Otherwise you will not be able to see the ports.

ibuenros:
Overall it looks fine. I would still want to see a "strict" mode where we fail if a port is not available, at least for the AM. It can really mess up operability if some monitoring application is expecting a port and Gobblin silently runs on a different port.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
kadaan: Same as above

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
ibuenros: I see. So the client app will need to parse the logs (can you also include this somewhere?). Could we abstract this into a broadcast method which by default just logs, but can be overriden for other broadcast implementations (I could see written to a pid-like file, or using Zookeeper).

kadaan: At the moment the purpose is just so I can debug, profile, look at jxm, etc.  In the future I would like this information to be available somewhere via the admin api & ui.  It feels a little out of scope for me to do this now.

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
kadaan: Will do

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
kadaan: Sure

kadaan:
I will also add a strict mode.

gobblin-yarn/src/main/java/gobblin/yarn/YarnHelixUtils.java
```diff
@@ -199,4 +206,48 @@ public static void addFileAsLocalResource(FileSystem fs, Path destFilePath, Loca
 
     return environmentVariableMap;
   }
-}
+
+  public static Optional<Integer> takePort(int desiredPort) {
```
sahilTakiar: If the `desiredPort` is not available, is there a reason to sequentially increase the port value and check if each one is available. If the desiredPort is not available, I believe `new ServerSocket(0).getLocalPort()` will find an available port and return it.

kadaan: At first I was just getting a random port, but it seemed nicer to get a port that you requested.  The method here is how YourKit works and it feels pretty nice to me because when there are multiple containers on the same node they will all group closely together.

That said, if you are adamant I can change it.

ibuenros: I support @kadaan 's method. It is nice to have the containers with similar ports.

sahilTakiar: Not adamant on it, just curious. Makes sense for the the ports to be grouped together.

sahilTakiar: However, if someone accidentally puts `${PORT_1}` then this method could potentially take a long time. Correct me if I am wrong, but ports 1-49151 are used by other applications. If testing if a port is available takes 100 ms, then testing 1-49151 would take over 1 hour. Perhaps it good to require the value in to be above 49151.

Thoughts?

kadaan: I can add a minimum bounds check.  I also considered adding a optional maximum to the range.  Lastly, I might be wrong, but I thought that only the first ~1000 ports where “reserved”.

sahilTakiar: Yes I think you are right: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Registered_ports

Its still probably better to make sure the specified port is above 1024

kadaan: Will do.

kadaan:
@ibuenros @sahilTakiar I've updated this with the ability to specify four different port ranges.  They all are "strict" in that if a port cannot be found the application will crash.

The tokens are:

${PORT_1234} : Try to get port 1234 and fail if it is not available
${PORT_?1235}: Try to get any port from the minimum port (1025) inclusive to the specified port inclusive.
${PORT_1235?}: Try to get any port from the specified port inclusive to the maximum port (65535) inclusive
${PORT_?}: Try to get a random port between the minimum port (1025) inclusive and the maximum port (65535) inclusive

kadaan:
@ibuenros @sahilTakiar Can you review this?

sahilTakiar:
Is option 2 `${PORT_?1234}` needed? I just don't want us to create too many options for specifying the port. For option 3 `${PORT_1234+}` may be more intuitive than `${PORT_1234?}`

kadaan:
I can change that.

kadaan:
@sahilTakiar Changed the tokens to the format you suggested.

gobblin-utility/src/main/java/gobblin/util/PortUtils.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+
+import java.net.ServerSocket;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class PortUtils {
```
ibuenros: Can you add javadoc?

ibuenros:
Other than javadoc, this looks good to me. @sahilTakiar any other comments?

kadaan:
@ibuenros There is one issue.  The problem is that the app_master determines the ports, but I think it needs to keep track "globally" of what it has assigned and these ports need to be released later.  That way if the app_master spins up more containers it won't try to reuse them.  This problem is particularly problematic because if the port is already used it can keep a container from starting.

gobblin-utility/src/main/java/gobblin/util/PortUtils.java
```diff
@@ -0,0 +1,142 @@
...
+    this.portLocator = locator;
+  }
+
+  /**
+   * Replaces any port tokens in the specified string.
+   *
+   * NOTE: Tokens can be in the following forms:
+   *   1. ${PORT_123}
+   *   2. ${PORT_123+}
+   *   3. ${PORT_?}
```
sahilTakiar: Can you explain what each pattern means, and what the code will do for each pattern.

sahilTakiar:
Are we going to use this feature when launching the App Master? It seems to be only added for launching containers?

One more question, the code currently checks for available ports in the app master, but just because a port is available on the app master doesn't mean it will be available in a new container, right?

kadaan:
@sahilTakiar Don't merge this yet.  I have an issue I need to fix that turned up in testing.  The problem is that the ports need to be "taken" and "released" to by the app_master to ensyre that there

kadaan:
@sahilTakiar @ibuenros In theory there are some issues with this PR, but in practice it has been working well in our pre-prod environment for a while.  I use it all the time to connect to different containers via VisualVM, JavaMissionControl, JavaFlightRecorder, and YourKit.  That said, I'm open to any suggestions on how to make this safer so that we don't risk containers failing to start.

abti:
@htran1 Can you please review and add some thoughts since you recently played with Yarn mode?

kadaan:
@htran1 I'd love your comments.  I don't want to spend more time on something that won't be merged, so it would be nice to figure out a plan.

kadaan:
@htran1 Ping

abti:
@kadaan I went through the PR and the previous discussions, and I think we can merge it in and continue to evolve. 
So, can you please fix the git conflicts so that I can merge it in?

kadaan:
@abti I've rebased the changes on top of master.  Assuming the build works, you are free to do your final review and merge.

PR #755, opened 2016-02-24
Fixing skipTestGroup option (phase 2)

kadaan:
@sahilTakiar Seems like a few projects were missed in your last PR.  Specifically, gobblin-tunnel which was causing travis runs to timeout.  Also saw a number of other excludes and moved them up to the root build as they seemed generic enough.


sahilTakiar:
LGTM thanks for the fix

PR #758, opened 2016-02-24
Config management regression test

tuGithub:


tuGithub:
@chavdar , @sahilTakiar , please review it when you have time

sahilTakiar:
Can you add a description of what this PR entails?

sahilTakiar:
High Level Comment:
* Anyway we can get this to run via `TestNG` and not via Azkaban?
    * The current implementation seems to require running a job via Azkaban, which means these tests won't run unless someone manually uploads the job to Azkaban and runs them.
    * However, while these are integration tests I don't see anything that requires them to be run on Azkaban or in a real HDFS cluster, it could simply be run on top of the local file system (e.g. `FileSystem.getLocal()`
* The advantage would be that these tests get run during each build, which will help us catch any unexpected bugs earlier

tuGithub:
Added the design doc in the ticket: https://jira01.corp.linkedin.com:8443/browse/ETL-4050

If we only test based on testNG unit testing, that is not actual testing for HDFS store, especially through config client.

@chavdar , please advise.

Min

chavdar:
In general, unit tests should be used for testing a single component in isolation. Dependencies should be mocked or use simplified test versions.

Integration tests should use multiple realistic components potentially running on different Hadoop clusters. Eventually, we may have the tooling to setup one or more local Hadoop clusters for testing but we don't have that right now. Loading a test on Azkaban and running it there should be a temp solution.

What this PR is missing is a high-level description of how the integration test framework is working. It seems like RegressionTest is the input point but I am still trying to figure this out.

chavdar:
Do these tests rely on manual deployment of the config files on HDFS? If yes, we should integrate it with the config deployment code.

gobblin-config-management/gobblin-config-client/build.gradle
```diff
@@ -18,6 +18,8 @@ dependencies {
     compile externalDependency.slf4j
     compile externalDependency.log4j
     compile externalDependency.typesafeConfig
+    compile externalDependency.azkaban
```
chavdar: If the integration tests bring a dependency on azkaban, they should be in a separate directory with its own build.gradle. The config client should not depend on azkaban.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+
+import gobblin.config.store.api.ConfigKeyPath;
+
+
+/**
+ * Utility class to compare results
+ * @author mitu
+ *
+ */
+public class ValidationUtils {
```
chavdar: Please use a singleton pattern rather than static methods.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+      displayConfig(config1, "Config1 ");
+      displayConfig(config2, "Config2 ");
+      throw new RuntimeException("Config entries size not match");
+    }
+
+    for (Map.Entry<String, ConfigValue> entry : config1.entrySet()) {
+      if (!entry.getValue().equals(config2.getValue(entry.getKey()))) {
+        displayConfig(config1, "Config1 ");
+        displayConfig(config2, "Config2 ");
+        throw new RuntimeException("Config entry not match for " + entry.getKey());
```
chavdar: nit: s/not/don't/

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+  public static void validateConfigKeyPathsWithOrder(List<ConfigKeyPath> list1, List<ConfigKeyPath> list2) {
+    if (!continueCheck(list1, list2)) {
+      return;
+    }
+
+    for (int i = 0; i < list1.size(); i++) {
+      if (!list1.get(i).equals(list2.get(i))) {
+        display(list1, "List1 ");
+        display(list2, "List2 ");
+        throw new RuntimeException("ConfigKeyPath not match for " + list1.get(i));
```
chavdar: nit: s/not/don't/

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+        display(collection1, "Collection1 ");
+        display(collection2, "Collection2 ");
+        throw new RuntimeException("ConfigKeyPath not match for " + path1);
+      }
+    }
+  }
+
+  // treat empty collection the same as null
+  private static boolean continueCheck(Collection<ConfigKeyPath> collection1, Collection<ConfigKeyPath> collection2) {
+    if (collection1 == null) {
```
chavdar: I think it is shorter/easier to write:
int collection1Size = 0 ? null == collection1 : collection1.size();
int collection2Size = 0 ? null == collection2 : collection2.size();
if (collection1Size != collection2Size) {
  ...
}

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+    if (collection2 == null) {
+      if (collection1 == null || collection1.size() == 0) {
+        return false;
+      }
+
+      throw new RuntimeException("List size not match");
+    }
+
+    if (collection1.size() != collection2.size()) {
+      throw new RuntimeException("List size not match");
```
chavdar: It is useful to add the mismatched sizes.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+        throw new RuntimeException("ConfigKeyPath not match for " + path1);
+      }
+    }
+  }
+
+  // treat empty collection the same as null
+  private static boolean continueCheck(Collection<ConfigKeyPath> collection1, Collection<ConfigKeyPath> collection2) {
+    if (collection1 == null) {
+      if (collection2 == null || collection2.size() == 0) {
+        return false;
```
chavdar: Why are not collections considered equal if both are null? That seems counter-intuitive.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/ValidationUtils.java
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Compare two {@link Collection} , ignoring the order
+   * @param list1
+   * @param list2
+   */
+  public static void validateConfigKeyPathsWithOutOrder(Collection<ConfigKeyPath> collection1,
+      Collection<ConfigKeyPath> collection2) {
+
+    if (!continueCheck(collection1, collection2)) {
+      return;
```
chavdar: What is contract here? Are null input expected or not? Current logic seems strange where if one or both are null. They are silently ignored.

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/regressiontest/RegressionTest.java
```diff
@@ -0,0 +1,229 @@
...
+    super(jobId, LOG);
+    this.physicalConfigStorePath = props.getProperty(PHYSICAL_CONFIG_STORE_URI_KEY);
+    this.logicalConfigStorePath = props.getProperty(LOGICAL_CONFIG_STORE_URI_KEY);
+  }
+
+  @Override
+  public void run() throws Exception {
+    URI physicalURI = new URI(this.physicalConfigStorePath);
+    URI logicalURI = new URI(this.logicalConfigStorePath);
+    SimpleHDFSConfigStoreFactory storeFactory = new SimpleHDFSConfigStoreFactory();
```
sahilTakiar: Is the test framework suppose to be tied to `SimpleHDFSConfigStore`?

sahilTakiar:
@chavdar why are tests added to TestNG restricted to unit tests?

* The current store implementations only require working with a `FileSystem` I'm not sure why we can't use the `SimpleLocalHDFSConfigStoreFactory` for this
* We do have the ability to setup HDFS clusters using the [MiniClusters](https://wiki.apache.org/hadoop/HowToDevelopUnitTests), `GobblinYarnAppLauncherTest` is already doing this
    * This should also allow us to spawn multiple clusters, I'm not sure why these integration tests need that though
    * I've used `MiniDFSCluster` before and it works well
* Getting this to run via TestNG also ensures the tests get run on each build
    * Correct me if I am wrong but we currently have to run the integration tests on Azkaban manually; and even if we did automate runs on Azkaban that would only happen internally

tuGithub:
High level structure:

1. RegressionTest is the entry point of the test
2. Based on the configuration, ExpectedResultBuilder will build all the expected results by parsing the json file "expected.conf" for each node
3. There are multiple levels of validation
   A. validate through SimpleHDFSConfigStore
   B. validate through the InMemoryTopology and InMemoryValueInspector
   C. validate through the ConfigClient

abti:
@vasanthrajamani Can you please prioritize?

abti:
https://issues.apache.org/jira/browse/GOBBLIN-134

Please update your PR title with following prefix: [GOBBLIN-134]

PR #759, opened 2016-02-25
Allow downloading other filetypes in FileBasedExtractor

pcadabam:
Decouple downloading logic from extractor to allow users to download other filetypes like gzip, zip etc.
@sahilTakiar can you review?


sahilTakiar:
`FileBasedExtractor` should probably not be an `abstract` class anymore. Can you remove it?

Otherwise LGTM

PR #760, opened 2016-02-25
Per dataset retention blacklist

pcadabam:
With dataset management we allow per dataset blacklist as opposed to comma separated list of blacklists in the job props file. The pr sets dataset backlist using Config

@ibuenros or @ydai1124 can you review?


ibuenros:
LGTM

PR #764, opened 2016-02-26
Ensure that jobs cleanup correctly

### LINKED ISSUE #754: Gobblin Lock deleted after second scheduled run
When the gobblin lock is enabled, and the job is scheduled to run every 15 min., this is an example of a scenario I have observed:

00- 1st Gobblin Job starts. Lock file gets created in lock-dir
15- 1st Gobblin Job is still running. 2nd scheduled job does not run post which the lock file gets deleted even if the 1st Job is running
30- 1st Job is still running. 3rd Job starts since no lock file exists. Both 1st and 3rd Job fails

Is this expected behavior?

kadaan:
### Issues Fixed
#754
#785

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -13,8 +13,6 @@
 package gobblin.runtime;
 
 import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


kadaan:
@sahilTakiar Can you review this?

kadaan:
@liyinan926 @zliu41 Are you two interested in reviewing this?

sahilTakiar:
Is there a concrete use case for providing a `SKIP` option? I'm ok with just always having the job fail if the lock is already in use. This way its not possible for a user to shoot themselves in the foot by setting `job.lock.denied.behavior=SKIP` and later realize that even though the job ran it didn't do anything.

kadaan:
Yes.  If we schedule a job to run every 15 minutes and it takes longer than 15 minutes, a failure will be logged to the history DB.  Additionally, if we want to run the job "constantly" we schedule it to run once a minute.  This ensures that we process "constantly", but can cause a huge number of job failures to be recorded in the history DB.

I believe there are several ways to properly deal with the problems that can arise from the lock.  These are important whether skipping is enabled or not.  First, the job lock should require a heartbeat so that jobs that have crashed, frozen, or had machine die don't prevent the system from making progress.  Second, there should be metrics which indicate the number of job which are skipped or failed within a given unit of time.  This can be used to raise an alert to indicate that a problem exists with the job.  Third, the job run time can also be used to determine if there is a fundamental problem with the job in regards to the schedule and the amount of work to be done.  Lastly, SLAs can be use to determine if the job is healthy and processing correctly.

kadaan:
@sahilTakiar I looked at this again and I think I can accomplish my goals without the new setting.  I'll update the PR in a few.

kadaan:
I updated the PR and changed it a little to fix #785 as well.

sahilTakiar:
What scheduler are you using? If it's Quartz then it seems like there is an annotation to disallow concurrent executions: http://quartz-scheduler.org/api/2.2.0/org/quartz/DisallowConcurrentExecution.html

http://stackoverflow.com/questions/2676295/quartz-preventing-concurrent-instances-of-a-job-in-jobs-xml

kadaan:
@sahilTakiar I'm using Quartz, but I don't think the disallow concurrent execution annotation will work.  It is stated in at least one comment that the jobs get queued.

Either way, I removed the code for the job.lock.denied.behavior..

sahilTakiar:
LGTM

Issue #729, opened 2016-02-19
LogCopier uses too much CPU


kadaan:
We are running Gobblin on yarn with 5 nodes and 49 containers, resulting in 98 files to copy.  The LogCopier on the driver ends up running constantly and making the machine run at 50%+ cpu usage.

kadaan:
<img width="500" alt="screen shot 2016-02-19 at 10 59 37 am" src="https://cloud.githubusercontent.com/assets/1904898/13186442/148d47c6-d6fb-11e5-9900-aa1907ce83d9.png">
<img width="500" alt="screen shot 2016-02-19 at 11 03 59 am" src="https://cloud.githubusercontent.com/assets/1904898/13186445/16903952-d6fb-11e5-944f-3b4e553c8ce1.png">
<img width="500" alt="screen shot 2016-02-19 at 11 05 22 am" src="https://cloud.githubusercontent.com/assets/1904898/13186446/1a9173cc-d6fb-11e5-9999-3eb3c5eccac7.png">

sahilTakiar:
* What tool are you using to generate these images?
* There doesn't seem to be a config parameter to control the sync interval, which is 2 minutes by default. 2 minutes seems pretty long so not sure this is the issue
* The `LogCopier` reads each line in the log file into a `String` and then writes it to the destination, it may be more efficient to just do a raw byte level copy
    * It does this so it can apply an optional regex to each line to exclude certain patterns, however, this feature does not seem to be used so I'm not sure its needed
* The copier uses `FileSystem.append` don't have much experience with this function but I wonder how efficient it is

Nothing else looks out of the ordinary in the code base. Will need to dig deeper.

kadaan:
@sahilTakiar This was captured using YourKit. The majority of the time is taken in the ScheduledThreadPoolExecutor. I replaced it with HashedWheelTimer and the performance is way better. I will try to reproduce again and add more details. What else would you like to know?

kadaan:
I'm going to submit a PR with the ability to switch between two different scheduler: ScheduledThreadPoolExecutor and HashedWheelTimer.

sahilTakiar:
Thanks for the info. So based on the YourKit analysis, the issue is with `ScheduledThreadPoolExecutor.getTask()` taking up a lot of CPU, is that correct? If so I agree replacing it something more efficient makes sense. Is the `HashedWheelTimer` you are referring to the one from Netty?

[...]

PR #768, opened 2016-02-26
Switch LogCopier from ScheduledExecutorService to HashedWheelTimer

kadaan:
@sahilTakiar @ibuenros Can you please review this fix for #729. Also, add anyone else on your side who has had experience with this.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/concurrent/ScheduledTask.java
+++ b/gobblin-utility/src/main/java/gobblin/util/concurrent/ScheduledTask.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.concurrent;
+
+/**
+ * A task which can be scheduled to run recurrently by an instance of {@link TaskScheduler}.
+ *
+ * @author joelbaranick
+ */
+public interface ScheduledTask<K> {
+    /**
+     * The unique key used to identify this task.
+     *
+     * @return the unique key
+     */
+    K getKey();
+
+    /**
+     * Run one iteration of the scheduled task.
+     */
+    void runOneIteration();
+}
--- a/gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
+++ b/gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.concurrent;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.base.Function;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.Iterables;
+
+import lombok.Synchronized;
+
+
+/**
+ * Base class for a task scheduler that can run {@link ScheduledTask}s periodically. Subclasses can implement
+ * {@link #startImpl(Optional)} and {@link #closeImpl()}.
+ *
+ * <p>Implementations of this interface are expected to be thread-safe, and can be safely accessed
+ * by multiple concurrent threads.</p>
+ *
+ * @author joelbaranick
+ */
+public abstract class TaskScheduler<K, T extends ScheduledTask<K>> implements Closeable {
+  private final Cache<K, CancellableTask<K, T>> cancellableTaskMap = CacheBuilder.newBuilder().build();
+  private volatile boolean isStarted = false;
+
+  protected TaskScheduler() {
+  }
+
+  /**
+   * Start the {@link TaskScheduler}.
+   *
+   * @param name the name of the {@link TaskScheduler}
+   */
+  @Synchronized
+  final void start(Optional<String> name) {
+    if (!this.isStarted) {
+      this.startImpl(name);
+      this.isStarted = true;
+    }
+  }
+
+  /**
+   * Start the {@link TaskScheduler}.
+   *
+   * @param name the name of the {@link TaskScheduler}
+   */
+  abstract void startImpl(Optional<String> name);
+
+  /**
+   * Schedules a subclass of {@link ScheduledTask} to run periodically.
+   *
+   * @param task the subclass of {@link ScheduledTask} to run every period
+   * @param period the period between successive executions of the task
+   * @param unit the time unit of the initialDelay and period parameters
+   */
+  public final void schedule(final T task, final long period, final TimeUnit unit) throws IOException {
+    Preconditions.checkArgument(isStarted, "TaskScheduler is not started");
+    try {
+      CancellableTask<K, T> cancellableTask = this.cancellableTaskMap.get(task.getKey(),
+              new Callable<CancellableTask<K, T>>() {
+                @Override
+                public CancellableTask<K, T> call() {
+                  return scheduleImpl(task, period, unit);
+                }
+              });
+      if (cancellableTask.getScheduledTask() != task) {
+        throw new IOException("Failed to schedule task with duplicate key");
+      }
+    } catch (ExecutionException e) {
+      throw new IOException("Failed to schedule task", e);
+    }
+  }
+
+  /**
+   * Schedules a subclass of {@link ScheduledTask} to run periodically.
+   *
+   * @param task the subclass of {@link ScheduledTask} to run every period
+   * @param period the period between successive executions of the task
+   * @param unit the time unit of the period parameter
+   */
+  abstract CancellableTask<K, T> scheduleImpl(T task, long period, TimeUnit unit);
+
+  /**
+   * Gets all {@link ScheduledTask}s.
+   *
+   * @return the {@link ScheduledTask}s
+   */
+  public final Iterable<T> getScheduledTasks() {
+    return Iterables.transform(this.cancellableTaskMap.asMap().values(),
+          new Function<CancellableTask<K, T>, T>() {
+            @Override
+            public T apply(CancellableTask<K, T> cancellableTask) {
+              return cancellableTask.getScheduledTask();
+            }
+          });
+  }
+
+  /**
+   * Gets the {@link ScheduledTask} with the specified key.
+   *
+   * @return the {@link ScheduledTask}
+   */
+  public final Optional<T> getScheduledTask(K key) {
+    CancellableTask<K, T> cancellableTask = this.cancellableTaskMap.getIfPresent(key);
+    if (cancellableTask != null) {
+      return Optional.of(cancellableTask.getScheduledTask());
+    }
+    return Optional.absent();
+  }
+
+  /**
+   * Attempts to cancel the specified {@link ScheduledTask}.
+   *
+   * @param task the {@link ScheduledTask} to cancel
+   * @return true if the {@link ScheduledTask} was canceled; otherwise, false
+   */
+  public boolean cancel(T task) {
+    CancellableTask<K, T> cancellableTask = this.cancellableTaskMap.getIfPresent(task.getKey());
+    if (cancellableTask != null && cancellableTask.getScheduledTask() == task && cancellableTask.cancel()) {
+      this.cancellableTaskMap.invalidate(task.getKey());
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Closes this {@link TaskScheduler}, ensuring that new tasks cannot be created
+   * and cancelling existing tasks.
+   *
+   * @throws IOException if an I/O error occurs
+   */
+  @Override
+  @Synchronized
+  public final void close() throws IOException {
+    if (isStarted) {
+      isStarted = false;
+      this.closeImpl();
+    }
+  }
+
+  /**
+   * Closes this {@link TaskScheduler}, ensuring that new tasks cannot be created
+   * and cancelling existing tasks.
+   *
+   * @throws IOException if an I/O error occurs
+   */
+  abstract void closeImpl() throws IOException;
+}
--- a/gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
+++ b/gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.concurrent;
+
+import com.google.common.base.Optional;
+
+
+/**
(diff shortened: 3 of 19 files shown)


sahilTakiar:
@chavdar might be interested, more details are in this issue: #729

kadaan:
@chavdar @sahilTakiar @ibuenros Can you review this?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,81 @@
...
+  final void start(Optional<String> name) {
+    if (!this.isStarted) {
+      this.startImpl(name);
+      this.isStarted = true;
+    }
+  }
+
+  abstract void startImpl(Optional<String> name);
+
+  public final void schedule(T task, long interval, TimeUnit timeUnit) {
```
ibuenros: Should this return a key? or the cancellable task itself?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,81 @@
...
+    return ImmutableList.copyOf(this.cancellableScheduledTaskMap.values());
+  }
+
+  public final Iterable<CancellableScheduledTask<K>> getScheduledTasks(K key) {
+    return ImmutableList.copyOf(this.cancellableScheduledTaskMap.get(key));
+  }
+
+  public void cancel(CancellableScheduledTask<K> task) {
+    if (this.cancellableScheduledTaskMap.remove(task.getKey(), task)) {
+      task.cancel();
```
ibuenros: What happens if `task.cancel()` fails but the task has already been removed from the map?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  @Synchronized
+  public final void close() throws IOException {
+    if (isStarted) {
+      try {
+        this.closeImpl();
+      } finally {
+        isStarted = false;
```
ibuenros: I think `isStarted` should be set to false before calling `closeImpl()`. The call might be slow, but we should block new tasks as soon as the close starts.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,54 @@
+package gobblin.util.concurrent;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Maps;
+import org.apache.commons.lang.StringUtils;
+
+import java.util.Map;
+
+public class TaskSchedulerFactory {
+  private static final Map<String, Class> SCHEDULER_MAP = buildSchedulerMap();
```
ibuenros: We try to avoid static maps, as they may generate memory leaks. An alternative is to use a Cache.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: I agree it is a little weird.  The passed in `T` is of type `ScheduledTask` and has a key.  Therefore the ScheduledTask can be found in the current task list by it's key.  The `CancelableScheduledTask` is only there because the method of cancelation is different for each type of Scheduler.  I'm open to any ideas.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: How about adding a isClosing and setting that?  I don't want it to be able to be started while closing.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        try {
+          TaskScheduler<K, T> taskScheduler = (TaskScheduler<K, T>) SCHEDULER_MAP.get(schedulerType).newInstance();
+          taskScheduler.start(name);
+          return taskScheduler;
+        } catch (InstantiationException | IllegalAccessException e) {
+          throw new RuntimeException("Unable to instantiate task scheduler '" + name + "'.", e);
+        }
+      }
+    }
+    return get();
```
ibuenros: Can this be handled in this method instead of recursing?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
kadaan: You are referring to the Guava cache, which can be size bounded, right?  

In this case, how can it leak?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,54 @@
+package gobblin.util.concurrent;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Maps;
+import org.apache.commons.lang.StringUtils;
+
+import java.util.Map;
+
+public class TaskSchedulerFactory {
```
ibuenros: Can you explain what the rationale for this class is? It seems to be caching something, but I don't understand what.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
kadaan: Sure.

gobblin-utility/src/main/java/gobblin/util/concurrent/ScheduledTask.java
```diff
@@ -0,0 +1,19 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.concurrent;
+
+public interface ScheduledTask<K> {
+    K getKey();
+
+    void run();
```
ibuenros: Should we allow exceptions here?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
kadaan: This is so the TaskSchedulers can be loaded by name.  I want to be able to allow the TaskScheduler to be configured in the gobblin config so that we can switch between them before we are certain that the new Scheduler is performing correctly.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
ibuenros: Is there a reason to support restarting? IMO once it's closed, it should never be restarted, so you could just have a three state version instead: `unstarted`, `started`, `closed`.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
ibuenros: Then I would rather use an enum instead of a map (also solves the issue with the static map).

kadaan: I can do that.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
ibuenros: I am a bit confused, which task list are you referrring? `cancellableScheduledTaskMap` or a different list?

ibuenros:
Can you add javadoc and copyright block to all classes?

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: I'm referring to `cancellableScheduledTaskMap`.

ibuenros: My concern is, say `task.cancel()` throws a runtime exception, which is caught by the caller. If the caller attempts to cancel the task once again, the task will no longer be in the map, and nothing will be cancelled. Or am I understanding this wrong?

gobblin-utility/src/main/java/gobblin/util/concurrent/ScheduledTask.java
kadaan: I want to ensure that now exceptions can leak out other than RuntimeExceptions because depending on the implementation of the TaskScheduler, it cause break things.  For example, `ScheduledTaskExecutor` indicates: "...if any execution of the task encounters an exception, subsequent executions are suppressed..."

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: How about `task.cancel()` can return a boolean and if it returns `true` we remove it from the collection and return `true`; otherwise, leave it in the collection and return `false`.

kadaan:
@ibuenros Updated based on your comments.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,113 @@
...
+  @AllArgsConstructor
+  private enum TaskSchedulerType {
+    SCHEDULEDEXECUTORSERVICE(ScheduledExecutorServiceTaskScheduler.class),
+    HASHEDWHEELTIMER(HashedWheelTimerTaskScheduler.class);
+
+    @Getter
+    private final Class<? extends TaskScheduler> taskSchedulerClass;
+
+    public static Optional<TaskSchedulerType> parse(Optional<String> name) {
```
ibuenros: What is the point of returning an optional here? Couldn't you just return the default? (that seems to be the only way you are using `parse`)

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Gets an instance of the {@link TaskScheduler} with the specified type and ensures that it is started. If
+   * the type is unknown an instance of the default {@link TaskScheduler} will be returned.
+   *
+   * @param type the type of the {@link TaskScheduler}
+   * @param <K> the type of the key for the {@link ScheduledTask}
+   * @param <T> the type of the {@link ScheduledTask}
+   * @return an instance of {@link TaskScheduler}
+   */
+  public static <K, T extends ScheduledTask<K>> TaskScheduler<K, T> get(String type) {
```
ibuenros: This method's signature is too close to `get(Optional<String>)`, it will probably cause confusion in the future.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskSchedulerFactory.java
```diff
@@ -0,0 +1,113 @@
...
+   * @param name the name associated threads created by the {@link TaskScheduler}
+   * @param <K> the type of the key for the {@link ScheduledTask}
+   * @param <T> the type of the {@link ScheduledTask}
+   * @return an instance of {@link TaskScheduler}
+   */
+  public static <K, T extends ScheduledTask<K>> TaskScheduler<K, T> get(String type, Optional<String> name) {
+    return getImpl(Optional.fromNullable(type), name);
+  }
+
+  private static <K, T extends ScheduledTask<K>> TaskScheduler<K, T> getImpl(Optional<String> type, Optional<String> name) {
```
ibuenros: Should this signature be in fact `getImpl(String, Optional<String>)`? Given that type is not actually optional? The methods with no type can get the default type themselves.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,134 @@
...
+   * Schedules a subclass of {@link ScheduledTask} to run periodically.
+   *
+   * @param task the subclass of {@link ScheduledTask} to run every period
+   * @param period the period between successive executions of the task
+   * @param unit the time unit of the initialDelay and period parameters
+   */
+  public final void schedule(final T task, final long period, final TimeUnit unit) throws IOException {
+    Preconditions.checkArgument(isStarted, "TaskScheduler is not started");
+    try {
+      this.cancellableTaskMap.get(task.getKey(),
```
ibuenros: If a task with that key is already present, this will silently not schedule the new task. I think it should throw an exception instead.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.Iterables;
+
+import lombok.Synchronized;
+
+/**
+ * Base class for a task scheduler that can run {@link ScheduledTask}s periodically. Subclasses can implement
+ * {@link #startImpl(Optional)} and {@link #closeImpl()}.
+ */
+public abstract class TaskScheduler<K, T extends ScheduledTask<K>> implements Closeable {
```
ibuenros: Is thread safety a concern? If so, I believe most methods in this class are not thread safe. If not, please specify that in the javadoc.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,134 @@
...
+  private final Cache<K, CancellableTask<K, T>> cancellableTaskMap = CacheBuilder.newBuilder().build();
```
ibuenros: Is there a reason to use a `Cache` instead of a `ConcurrentMap`?

ibuenros:
Please add javadoc to all classes.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: Because cache has a thread-safe computeIfAbsent method that is Java7 compatible: `get(K, Callable<? extends V>)`

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: `start` and `close` are thread-safe because they are synchronized.  `schedule` is thread-safe because the underlying `get` call is thread-safe.

kadaan:
@ibuenros Updated based on your comments.  

1. Removed a bunch of overrides for `TaskSchedulerFactory`, and added one that takes the enum rather than the string.  
1. Made the `TaskSchedulerType` enum public.  
1. Indicated that `TaskScheduler` implementations should be thread-safe. 
1. Added javadocs for all non-test classes and method (even private and package visibility). 
1. Fail the `schedule` call if the task cannot be scheduler

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
```diff
@@ -0,0 +1,165 @@
...
+    return Optional.absent();
+  }
+
+  /**
+   * Attempts to cancel the specified {@link ScheduledTask}.
+   *
+   * @param task the {@link ScheduledTask} to cancel
+   * @return true if the {@link ScheduledTask} was canceled; otherwise, false
+   */
+  public boolean cancel(T task) {
```
ibuenros: This call is not thread-safe. While the task is canceling, a different thread might cancel the same task.

ibuenros:
Just one more comment.

gobblin-utility/src/main/java/gobblin/util/concurrent/TaskScheduler.java
kadaan: I don't think it matters for the following reasons:

1. The underlying `cancellableTaskMap` is thread safe
1. The `ScheduledTask.cancel` is safe to call multiple times
1. Until the `this.cancellableTaskMap.invalidate(task.getKey())` call, you cannot create another task with the same key.

The only improvement I can see is to add a reference check against the `T task` and the `cancellableTask.getScheduledTask()` to ensure that the passed in task is the same one that will be cancelled.  If not, just return false (because the passed in task has not been scheduled).

ibuenros: Why is it necessary safe to call `cancel()` multiple times? (The interface does not specify that, and even if it did, it is very easy for users to ignore that requirement).
The check that the task to cancel is the expected one is another good thing to add.

kadaan: Well, all implementations of the cancel method are currently safe.  For instance the `HashedWheelTimer` `Timeout` states: "If the task has been executed or cancelled already, it will return with no side effect."  Additionally, the `Future` states the `cancel` method returns: "false if the task could not be cancelled, typically because it has already completed normally".  I don't think it is unreasonable for interfaces to require that the implementations are thread-safe.  This is done in many other places.  My preference is to reduce high level locking as much as possible.

ibuenros: Ok, at least specify in the interface that multiple calls to cancel must be safe.

kadaan: FYI, the implementation of `CancellableTask` for `ScheduledExecutorServiceTaskScheduler` does not synchronize the `cancel` method because it isn't needed.  The implementation of `CancellableTask` for `HashedWheelTimerTaskScheduler` does synchronize the `cancel` method to ensure that the future isn't being set by the `runOneIteration` method and unset by `cancel` at the same time.

kadaan:
@ibuenros Updated with comment

ibuenros:
LGTM

kadaan:
@ibuenros Thanks!

PR #771, opened 2016-02-26
Fix for #739

kadaan:
@sahilTakiar Added @Synchronized to the method to fix a FindBugs issue in #747.  See: http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

### DIFF
--- a/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
+++ b/gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
@@ -39,6 +39,7 @@ import com.typesafe.config.Config;
 import com.typesafe.config.ConfigValueFactory;
 
 import lombok.extern.slf4j.Slf4j;
+import lombok.Synchronized;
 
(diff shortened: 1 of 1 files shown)


gobblin-metrics/src/main/java/gobblin/metrics/reporter/ScheduledReporter.java
```diff
@@ -105,13 +106,10 @@ public ScheduledReporter(String name, Config config) {
     ensureMetricFilterIsInitialized(config);
   }
 
+  @Synchronized
   private void ensureMetricFilterIsInitialized(Config config) {
     if (this.metricFilter == null) {
-      synchronized (this.syncObj) {
```
sahilTakiar: Can you delete the declaration and creation of `this.syncObj`

kadaan: Good call.. Too many open PRs.

sahilTakiar:
LGTM

Issue #770, opened 2016-02-26
Enabled FindBugs and CheckStyle on Travis builds


kadaan:
To ensure PR as correct as possible, both FindBugs and Checkstyle should be enabled on Travis builds.  Additionally, they should be either enabled by default on local builds, or easily enabled.

sahilTakiar:
FindBugs has been enabled on Travis: https://github.com/linkedin/gobblin/pull/984

PR #772, opened 2016-02-26
Upgrading and re-enabling Findbugs

sahilTakiar:
* Upgraded FindBugs to version 3.0.1 which is compatible with Java 8, this new version only requires Java 7+ to run
* I can't find an option to set a threshold for the number of allowable findbugs bugs before a failure is trigger
    * Unless we find a way to do so, we would have to resolve all findbugs bugs before adding this to the Travis build

@chavdar can you review?


sahilTakiar:
This should at least allow us and any other users to run findbugs using the following command: `./gradlew findbugsMain`

kadaan:
Can we enable it in travis by default?

kadaan:
Fix for part of #770

chavdar:
@sahilTakiar The threshold is actually part of the hudson plugin. We can check if there is some similar gradle plugin or we can define a gradle task to do that.

sahilTakiar:
@kadaan yes, we should aim to get FindBugs running on Travis, thanks for opening: https://github.com/linkedin/gobblin/issues/770

@chavdar yes I tried to find a similar functionality in Gradle, but couldn't find much. The default config options for the findbugs plugin doesn't seem to offer anything like that: https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.FindBugs.html

The FindBugs Gradle plugin does seem to track the info, I can't seem to determine if it exposes it or not: https://github.com/gradle/gradle/blob/master/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/internal/findbugs/FindBugsResult.java (`getBugCount()`)

sahilTakiar:
@ibuenros can you review this? Pretty minor change.

ibuenros:
LGTM

PR #773, opened 2016-02-27
Refactoring Hive registration to allow query-based approach

zliu41:
This PR enables Hive registration to support both MetaStoreClient-based registration and query-based registration (originally it only supports the former). 

Main changes:
- `HiveRegister` is now an abstract class. It can be implemented by `HiveMetaStoreBasedRegister` (which is similar as the original `HiveRegister` class), and `HiveQueryBasedRegister` (not in this PR).
- To remove dependency on Hive's `Table` and `Partition` classes, I created a `HiveTable` class and a `HivePartition` class in Gobblin, both of which extends `HiveRegistrationUnit`. Hive's `Table`/`Partition` can be converted to Gobblin's `HiveTable`/`HivePartition` (and vice versa) in `HiveMetaStoreUtils`.
- Removed `HiveStorageDescriptorManager`, since storage descriptors are now created in `HiveMetaStoreUtils` in the process of converting `HiveTable`/`HivePartition` to `Table`/`Partition`. `HiveSchemaManager` is renamed to `HiveSerDeManager`.

Now the only classes that depend on `hive-metastore` are `HiveMetaStoreBasedRegister` and `HiveMetaStoreUtils`.


zliu41:

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -0,0 +1,222 @@
...
+
+
+/**
+ * A class that represents a Hive table or partition.
+ *
+ * @author ziliu
+ */
+@Getter
+@Alpha
+public class HiveRegistrationUnit {
```
ibuenros: What is the rationale for not using Hive's `Table` and `Partition`?

zliu41: 1. I'd rather not depend on classes in hive-metastore if we are not using the metastore based approach, since there is a possibility that they may have backward incompatible changes.
2. There's a lot of commonality of Table and Partition, so it is convenient to make them have a common parent (`HiveRegistrationUnit`)

sahilTakiar:
What's the difference between `gobblin.data.management.hive.HiveRegistrableDataset` and `gobblin.hive.policy.HiveRegistrationPolicy`?

zliu41:
`HiveRegistrableDataset` is not currently used, and I'm not sure if it's needed. Conceptually it is a `Dataset` that contains a set of `HiveSpec`s. `HiveRegistrationPolicy` contains the logic to generate `HiveSpec`s for a specific path.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -0,0 +1,222 @@
...
+  protected final List<Column> columns = Lists.newArrayList();
+  protected final State props = new State();
+  protected final State storageProps = new State();
+  protected final State serDeProps = new State();
+
+  /**
+   * Table or Partition properties
+   */
+  protected Optional<Long> createTime;
+  protected Optional<Long> lastAccessTime;
```
ydai1124: What about properties like `Priviledges`, `Parameters`?

zliu41: `props` is `Parameters`. 
There are a few properties not included here, such as `Priviledges` and `SortedCols`, since their return types are not String and require adding more classes. If there's a use case that needs them we can add them later

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -0,0 +1,222 @@
...
+  public void setColumns(List<Column> columns) {
+    this.columns.clear();
+    this.columns.addAll(columns);
+  }
+
+  public void setProp(String key, Object value) {
+    this.props.setProp(key, value);
+  }
+
+  public void setStorageProp(String key, Object value) {
```
ydai1124: If user sets `storageProps` like `LOCATION`, does it need to be propagated to `this.location`?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveConstants.java
```diff
@@ -0,0 +1,56 @@
...
+import gobblin.annotation.Alpha;
+
+
+/**
+ * A class containing constants used in {@link HiveTable} and {@link HivePartition}.
+ */
+@Alpha
+public class HiveConstants {
+
+  private HiveConstants() {
```
sahilTakiar: Is this needed

gobblin-hive-registration/src/main/java/gobblin/hive/HivePartition.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A class that represents a Hive partition.
+ *
+ * @author ziliu
+ */
+@Getter
+@Alpha
+public class HivePartition extends HiveRegistrationUnit {
+
+  HivePartition(Builder builder) {
```
sahilTakiar: can this be `private`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegisterUtils.java
```diff
@@ -27,6 +27,9 @@
  */
 public class HiveRegisterUtils {
 
+  private HiveRegisterUtils() {
```
sahilTakiar: is this necessary?

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -217,34 +216,45 @@ public static Schema getSchemaFromDataFile(Path dataFile, FileSystem fs) throws
...
-      closer.close();
+    }
+  }
+
+  public static void writeSchemaToFile(Schema schema, Path filePath, FileSystem fs, boolean overwrite)
+      throws IOException {
+    if (!overwrite) {
+      Preconditions.checkState(!fs.exists(filePath), filePath + " already exists");
+    } else {
+      HadoopUtils.deletePath(fs, filePath, false);
```
sahilTakiar: this should be a recursive delete

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -0,0 +1,343 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * An implementation of {@link HiveRegister} that uses {@link IMetaStoreClient} for Hive registration.
+ *
+ * <p>
+ *   An instance of this class is constructed with a {@link State} object or obtained via
+ *   {@link HiveRegister#get(State)}. Property {@link HiveRegProps#HIVE_DB_ROOT_DIR} is required for registering
```
sahilTakiar: Update javadoc. Couple other references to `#register` and `HiveRegister` should be changed.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -49,57 +40,45 @@
...
-  private final List<Future<Void>> futures = Lists.newArrayList();
-  private final HiveStripedLocks locks = new HiveStripedLocks();
+  @Getter
+  protected final HiveRegProps props;
 
-  public HiveRegister(State state) throws IOException {
+  protected final Optional<String> hiveDbRootDir;
+  protected final ListeningExecutorService executor;
+  protected final List<Future<Void>> futures = Lists.newArrayList();
+  protected final HiveStripedLocks locks = new HiveStripedLocks();
```
sahilTakiar: are the locks needed when using the query based approach? https://cwiki.apache.org/confluence/display/Hive/Locking

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -0,0 +1,343 @@
...
+      Optional<HivePartition> partition = spec.getPartition();
+      createOrAlterTable(client, table);
+
+      if (partition.isPresent()) {
+
+        // Register a partition
+        addOrAlterPartition(client, table, HiveMetaStoreUtils.getPartition(partition.get()), spec.getPath());
+      }
+
+      if (spec instanceof HiveSpecWithPostActivities) {
```
sahilTakiar: Isn't this already happening in the `HiveRegister.register` method?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -162,185 +124,57 @@ private boolean evaluatePredicates(HiveSpecWithPredicates spec) {
...
+  public abstract Optional<HivePartition> getPartition(String dbName, String tableName, List<Column> partitionKeys,
+      List<String> partitionValues) throws IOException;
 
-  private boolean needToUpdatePartition(Partition existingPartition, Partition newPartition) {
-    return !existingPartition.equals(newPartition);
-  }
+  /**
+   * Alter the given {@link HiveTable}. An Exception should be thrown if the table does not exist.
+   */
+  public abstract void alterTable(HiveTable table) throws IOException;
```
sahilTakiar: Where is this method called?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -109,46 +88,29 @@ public HiveRegister(State state) throws IOException {
...
-          Table table = spec.getTable();
-          createDbIfNotExists(client, table.getDbName());
-
-          Optional<Partition> partition = spec.getPartition();
-          createOrAlterTable(client, table);
-
-          if (partition.isPresent()) {
-
-            // Register a partition
-            addOrAlterPartition(client, table, partition.get(), spec.getPath());
```
sahilTakiar: can't there be `abstract` methods for `createDbIfNotExists` `createOrAlterTable` and `addOrAlterPartition`?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -162,185 +124,57 @@ private boolean evaluatePredicates(HiveSpecWithPredicates spec) {
...
-    try {
-      return this.clientPool.borrowObject();
-    } catch (Exception e) {
-      throw new IOException("Unable to borrow " + IMetaStoreClient.class.getSimpleName());
-    }
-  }
+  /**
+   * Whether a table exists.
+   */
+  public abstract boolean existsTable(String dbName, String tableName) throws IOException;
```
sahilTakiar: Can we add full javadocs for all the `public abstract` methods

gobblin-hive-registration/src/main/java/gobblin/hive/HiveTable.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public void setTableType(String tableType) {
+    this.tableType = Optional.of(tableType);
+  }
+
+  public void setRetention(long retention) {
+    this.retention = Optional.of(retention);
+  }
+
+  public static class Builder extends HiveRegistrationUnit.Builder<Builder> {
```
sahilTakiar: Can we make builder package protected. Same goes for `HiveRegistrationUnit` and `HivePartition` - there shouldn't be any reason any one needs to create these objects outside this package correct?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -0,0 +1,222 @@
...
+
+    this.serDeType = Optional.fromNullable(this.serDeProps.getProp(HiveConstants.SERDE_TYPE));
+  }
+
+  public void setColumns(List<Column> columns) {
+    this.columns.clear();
+    this.columns.addAll(columns);
+  }
+
+  public void setProp(String key, Object value) {
```
sahilTakiar: Can you add documentation as to the difference between `setProp` `setStorageProp` and `setSerDeProp`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSerDeManager.java
```diff
@@ -0,0 +1,106 @@
...
+  protected final State props;
+
+  protected HiveSerDeManager(State props) {
+    this.props = props;
+  }
+
+  /**
+   * Add the appropriate SerDe properties (including schema properties) to the given {@link HiveRegistrationUnit}.
+   */
+  public abstract void addSerDeProperties(Path path, HiveRegistrationUnit hiveUnit) throws IOException;
```
sahilTakiar: What is `Path path` suppose to be?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSerDeManager.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Update the schema in the existing {@link HiveRegistrationUnit} into the schema in the new
+   * {@link HiveRegistrationUnit}.
+   */
+  public abstract void updateSchema(HiveRegistrationUnit existingUnit, HiveRegistrationUnit newUnit) throws IOException;
```
sahilTakiar: it may make more sense to have a `HiveSchema` class

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
```diff
@@ -0,0 +1,142 @@
...
+   * Add an Avro {@link Schema} to the given {@link HiveRegistrationUnit}. 
+   *
+   *  <p>
+   *    If {@link #USE_SCHEMA_FILE} is true, the schema will be added via {@link #SCHEMA_URL} pointing to
+   *    the schema file named {@link #SCHEMA_FILE_NAME}.
+   *  </p>
+   *
+   *  <p>
+   *    If {@link #USE_SCHEMA_FILE} is false, the schema will be obtained by {@link #getDirectorySchema(Path)}. 
+   *    If the length of the schema is less than {@link #SCHEMA_LITERAL_LENGTH_LIMIT}, it will be added via
```
sahilTakiar: Is there any upstream issue or documentation we can reference for why this behavior is necessary?

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
```diff
@@ -0,0 +1,142 @@
...
+   *
+   *  <p>
+   *    If {@link #USE_SCHEMA_FILE} is false, the schema will be obtained by {@link #getDirectorySchema(Path)}. 
+   *    If the length of the schema is less than {@link #SCHEMA_LITERAL_LENGTH_LIMIT}, it will be added via 
+   *    {@link #SCHEMA_LITERAL}. Otherwise, the schema will be written to {@link #SCHEMA_FILE_NAME} and added
+   *    via {@link #SCHEMA_URL}.
+   *  </p>
+   */
+  @Override
+  public void addSerDeProperties(Path path, HiveRegistrationUnit hiveUnit) throws IOException {
```
sahilTakiar: Can this all be set in the `HiveRegistrationUnit.Builder`? Couldn't there just be a `HiveSerDe` class that encapsulates all this logic, and each `HiveRegistrationUnit` has a `HiveSerDe` object.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveConstants.java
zliu41: Just to make this class non instantiable

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
zliu41: This java doc is actually up to date.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Not in this PR but some use cases will need to alter table e.g. update schema.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Yes for `createDbIfNotExists`. For the other two, if the table/partition already exists, there are different ways to determine whether the table/partition should be updated. So they are a little ambiguous to be part of the HiveRegister API.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveTable.java
zliu41: No, all implementations of `HiveRegistrationPolicy` need to create `HiveTable` or `HivePartition`.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
zliu41: Not necessary. `setProp` is supposed to set parameters, not properties like location.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
zliu41: Using schema literal has slightly better performance, but the length is limited by Hive metastore column width. http://www.cloudera.com/documentation/archive/impala/1-x/1-0-1/Installing-and-Using-Impala/ciiu_avro.html

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
zliu41: Yes we can make it a field in `HiveRegistrationUnit`. But I'd like to keep the `HiveSerDeManager` name since it is really a manager class, e.g., it contains logic to fetch and keep track of the schemas; it is also responsible for determining whether the table schema should be updated when adding a new partition.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSerDeManager.java
zliu41: Not necessarily agree, since serde and schema are tightly related (if you use Avro serde then you need to use Avro schema), so it makes sense to have a class that manages both.

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
sahilTakiar: I see my mistake

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
sahilTakiar: I see. Is it possible to have something like this for say `createOrAlterTable`:

```
public void createOrAlterTable(...) {
  if (tableExists(table)) {
    createTable(table);
  } else {
    alterTable(table);
}

public abstract tableExists(...);

public abstract createTable(...);

public abstract alterTable(...);
```

gobblin-hive-registration/src/main/java/gobblin/hive/HiveTable.java
sahilTakiar: I see, I ask because I don't want users to get confused between Gobblin's `HiveTable` and Hive's `Table`. Can you add a little more detail into the javadoc about the differences between the two, why Hive's APIs are not used, what the class is intended for, etc.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
sahilTakiar: Ok agreed. So is the plan to add a `HiveSeDeManager` field to `HiveRegistrationUnit`? The `HiveSerdeManager` would also encapsulate everything related schemas, maybe it even has an inner class called `HiveSchema` (just an idea).

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
zliu41: Actually I think it's good to propagate, so will make the change

gobblin-hive-registration/src/main/java/gobblin/hive/HiveTable.java
zliu41: Added to `HiveTable` and `HivePartition`

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
zliu41: Yes I added `HiveSerDeManager` as a field in `HiveRegistrationUnit`. I feel that if there's an inner class `HiveSchema` then subclasses of `HiveSerDeManager` will also need to extend `HiveSchema` which is a little cumbersome so I left it as it is for now.

zliu41:
@sahilTakiar @ydai1124 updated 1631d0b

sahilTakiar:
Any comments on: https://github.com/linkedin/gobblin/pull/773#discussion-diff-54774146L136

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -0,0 +1,399 @@
...
+        this.isCompressed = Optional.of((boolean) value);
+        break;
+      case HiveConstants.NUM_BUCKETS:
+        this.numBuckets = Optional.of((int) value);
+        break;
+      case HiveConstants.BUCKET_COLUMNS:
+        this.bucketColumns = Optional.of(Splitter.on(',').omitEmptyStrings().trimResults().splitToList((String) value));
+        break;
+      case HiveConstants.STORED_AS_SUB_DIRS:
+        this.isSotredAsSubDirs = Optional.of((boolean) value);
```
sahilTakiar: `this. isSotredAsSubDirs` to `this. isSortedAsSubDirs`

gobblin-hive-registration/src/main/java/gobblin/hive/HiveSerDeManager.java
sahilTakiar: Is `existingUnit` still necessary given that a `HiveSerDeManager` is now tied to a `HiveRegistrationUnit`. The `existingUnit` would be containing `HiveRegistrationUnit`.

zliu41: good point

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
zliu41: Overlooked it. Yes this makes sense

zliu41:
@sahilTakiar updated 1e97b16

sahilTakiar:
LGTM

PR #774, opened 2016-02-27
Add HDFS config deployment tool

pcadabam:
## Changes

- Added `StoreDeployer` which copies config files in classpath to a `ConfigStore`
- Added a new interface called `Deployable` which has a `deploy()` and a `rollback()` method. A `ConfigStore` can implement a `Deployable`. The `StoreDeployer` initializes the right `deployable` `ConfigStore` using the `storeUri`. 
- The actual configs should be in its own package/project and in the classpath. This project will add gobblin-confg-management as a dependency. On calling `StoreDeployer.deploy()`, the configs in classpath are deployed
- Rollback is not yet implemented

## Example

If "test-root" is a resource in classpath and all resources (config files) under it needs to be deployed
**Structure in Classpath:**
           test-root
           &emsp;/data
           &emsp;&emsp;/set1
           &emsp;&emsp;&emsp;/main.conf
           &emsp;/tag
           &emsp;&emsp;/tag1
           &emsp;&emsp;&emsp;/main.conf

A new version say 2.0.0 is created on HDFS under \_CONFIG\_STORE
**Structure on HDFS after deploy:**
           /_CONFIG\_STORE
           &emsp;/2.0.0
           &emsp;&emsp;/data
           &emsp;&emsp;&emsp;/set1
           &emsp;&emsp;&emsp;&emsp;/main.conf
           &emsp;&emsp;/tag
           &emsp;&emsp;&emsp;/tag1
           &emsp;&emsp;&emsp;&emsp;/main.conf

@chavdar and @sahilTakiar can you review?

### DIFF
--- a/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
+++ b/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2015-16 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.config.store.deploy;
+
+import java.io.IOException;
+
+import gobblin.config.store.api.ConfigStore;
+
+
+/**
+ * An interface to deploy and rollback {@link ConfigStore}s. {@link ConfigStore}s that implement {@link Deployable} can
+ * be deployed using {@link StoreDeployer}
+ *
+ * @param <D> {@link DeploymentConfig} or its subclasses that has configs to deploy the store
+ * @param <R> {@link RollbackConfig} or its subclasses that has configs to rollback
+ */
+public interface Deployable<D extends DeploymentConfig, R extends RollbackConfig> {
+
+  /**
+   * Deploy a version {@link DeploymentConfig#getNewVersion()} of configs provided by
+   * {@link DeploymentConfig#getDeployableConfigSource()} on the {@link ConfigStore}
+   *
+   * @param deploymentConfig to use for this deployment
+   */
+  public void deploy(D deploymentConfig) throws IOException;
+
+  /**
+   * Rollback to an older version of configs on the {@link ConfigStore} using {@link RollbackConfig}
+   *
+   * @param rollbackConfig to use for this rollback
+   */
+  public void rollback(R rollbackConfig) throws IOException;
+}
--- a/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/DeploymentConfig.java
+++ b/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/DeploymentConfig.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2015-16 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.config.store.deploy;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.ToString;
+
+/**
+ * Holds deployment configuration to be passed to {@link Deployable#deploy(DeploymentConfig)}
+ */
+@AllArgsConstructor
+@Getter
+@ToString
+public class DeploymentConfig {
+  /**
+   * The source to use for this deployment.
+   */
+  private final DeployableConfigSource deployableConfigSource;
+  /**
+   * Version number to be used for this deployment
+   */
+  private final String newVersion;
+}
--- a/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/StoreDeployer.java
+++ b/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/StoreDeployer.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2015-16 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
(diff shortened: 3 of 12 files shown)


pcadabam:
@sahilTakiar and @chavdar please review this when you get a chance.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+/**
+ * An interface to deploy and rollback {@link ConfigStore}s. {@link ConfigStore}s that implement {@link Deployable} can
+ * be deployed using {@link StoreDeployer}
+ *
+ * @param <D> {@link DeploymentConfig} or its subclasses that has configs to deploy the store
+ * @param <R> {@link RollbackConfig} or its subclasses that has configs to rollback
+ */
+public interface Deployable<D extends DeploymentConfig, R extends RollbackConfig> {
```
sahilTakiar: Is it necessary to have a separate `DeploymentConfig` and `RollbackConfig`. For the `deploy()` method you just need the `version`, and for the `rollback` method you also just need the version.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/DeploymentConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ * Holds deployment configuration to be passed to {@link Deployable#deploy(DeploymentConfig)}
+ */
+@AllArgsConstructor
+@Getter
+@ToString
+public class DeploymentConfig {
+  /**
+   * Top level resource name in classpath that contains configs to be deployed.
+   */
+  private final String storeRootNameInClasspath;
```
sahilTakiar: Does this need to be part of the config? Presumably, this would never change for single `ConfigStore`. Also, `storeRootNameInClasspath` seems implementation specific. What if down the line we want to deploy from some other location instead. I think there can be a `DeployableFromClasspath` class that contains logic for deploying from the classpath.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/StoreDeployer.java
```diff
@@ -0,0 +1,67 @@
...
+
+      if (storeUri.getScheme().equals(storeFactory.getScheme())) {
+
+        log.info("Using ConfigStore with scheme : " + storeFactory.getScheme());
+
+        ConfigStore configStore = storeFactory.createConfigStore(storeUri);
+
+        if (configStore instanceof Deployable<?, ?>) {
+
+          ((Deployable) configStore).deploy(new FsDeploymentConfig(classpathStoreRoot, version));
```
sahilTakiar: This makes this class tied to `FsDeploymentConfig` what if a different config needs to be used?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/DeploymentConfig.java
pcadabam: I think that makes sense. How about we make an abstraction for `ConfigSource` with a `Set<String> getConfigPaths` method?
In the `DeploymentConfig` class instead of having `storeRootNameInClasspath`, we can have a `getConfigSource` method that returns a `ClasspathConfgSource` which implements a `ConfigSource`.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -389,4 +417,95 @@ public boolean apply(FileStatus input) {
...
+
+    Path hdfsconfigStoreRoot = new Path(this.physicalStoreRoot.getPath(), CONFIG_STORE_NAME);
+
+    if (!fs.exists(hdfsconfigStoreRoot)) {
+      throw new IOException("Config store root not present at " + this.physicalStoreRoot.getPath());
+    }
+
+    Path hdfsNewVersionPath = new Path(hdfsconfigStoreRoot, deploymentConfig.getNewVersion());
+
+    ConfigurationBuilder cb =
```
sahilTakiar: All the logic for deploying a config store from the classpath is probably re-usable for other `ConfigStore`s I suggest creating an abstract class called `DeployableFromClasspath` that encapsulates all this logic.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
pcadabam: I think having different config classes is more extensible. We do not yet know the behavior of other ConfigStores. We don't yet know the metadata they may need for deployment and rollback. The rollback may have rollback policies like Rollback if xyz is true etc.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/StoreDeployer.java
pcadabam: For now yes. I think we can add that code when new stores are implemented.

sahilTakiar:
Just throwing out ideas, but it may be better to separate the source of the configs from the deployment / rollback of the configs. An interface `ConfigSource` can have implementations like `ClasspathConfigSource`, etc. The `Deployable.deploy` method just takes in streams of data and is solely responsible for writing the data out. `StoreDeployer` ties the two interfaces together.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
sahilTakiar: True, but any additional config they need can just be passed in via the constructor or an `init` method.

pcadabam:
@sahilTakiar I think your suggestion for decoupling `ConfigSource` and Deployment makes sense. I think having an instance of `ConifgSource` inside of `DeploymentConfig` is better. The `Deployable.deploy(DeploymentConfig)` will get call `DeploymentConfg.ConfigSource.getConfigsPath()` which will return a  `Set<String>` of source path either from classpath or somewhere else.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/DeploymentConfig.java
sahilTakiar: Makes sense to me. Only suggestion is that instead of returning `Path`s it may be better to return streams. If `ConfigSource` only returns location, then `ClasspathConfigSource` can only return locations too, which means it can't handle reading the files from the classpath.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/deploy/Deployable.java
pcadabam: Discussed offline. We will keep separate configs for Deployment and Rollback so that we do not have to provide default rollback configs when doing a deployment and vice versa

pcadabam:
Hi @sahilTakiar I have update the PR per comments and offline discussion. Please take a look

sahilTakiar:
Sorry for commenting on the forked version instead of the PR version, comments are still the same.

pcadabam:
@sahilTakiar  Addressed all PR comments.

sahilTakiar:
LGTM

PR #775, opened 2016-02-29
Fix javadoc warnings by only adding linksOffline to projects that the current project depends on.

kadaan:
@sahilTakiar Can you review this?  Also, it there a way to verify that the resultant javadocs are correct?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -350,6 +350,15 @@ ext.pomAttributes = {
   }
 }
 
+def getAllDependentProjects(project) {
+  def projectDependencies = project.configurations.runtime.getAllDependencies().withType(ProjectDependency)
+  def dependentProjects = projectDependencies*.dependencyProject
+  if (dependentProjects.size() > 0) {
+      dependentProjects.each { dependentProjects += getAllDependentProjects(it) }
+  }
+  return dependentProjects.unique()
+}
+
 subprojects {
   plugins.withType(JavaPlugin) {
     plugins.apply('idea')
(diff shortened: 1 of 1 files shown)


sahilTakiar:
What warnings are you referring to? I think there is a way to configure the Gradle javadoc task to fail if there are errors: https://docs.gradle.org/current/dsl/org.gradle.api.tasks.javadoc.Javadoc.html#org.gradle.api.tasks.javadoc.Javadoc:failOnError

kadaan:
There are a ton of warnings about package-list files not being available. I believe this is because the build added every project to every other project as linksOffline and build ordering would cause projects java doc task to run before the non-droendency linksOverride.

sahilTakiar:
Yes, I believe you are referring to these:

```
javadoc: warning - Error reading file: /.../gobblin-github/build/gobblin-data-management/docs/javadoc/package-list
javadoc: warning - Error reading file: /.../gobblin-github/build/gobblin-data-management/docs/javadoc/package-list
javadoc: warning - Error reading file: /.../gobblin-github/build/gobblin-hive-registration/docs/javadoc/package-list
```

I believe @chavdar has more experience with this so he may be able to comment too.

kadaan:
@sahilTakiar Those are the ones.

kadaan:
@chavdar Can you review this?  It removes a ton of annoying warnings which clog the build logs.

sahilTakiar:
I'm going to attempt to review this as Chavdar may be busy with other things.

build.gradle
```diff
@@ -513,6 +520,17 @@ subprojects {
   }
 }
 
+gradle.projectsEvaluated {
+  subprojects {
+    plugins.withType(JavaPlugin) {
+      getAllDependentProjects(project).each {
+        tasks.javadoc.options.linksOffline "http://linkedin.github.io/gobblin/javadoc/${javadocVersion}/${it.name}/",
```
sahilTakiar: slightly orthogonal to this PR, but any idea why we would need an offline line link to a previously published version of the Gobblin JavaDocs?

kadaan: Isn’t this to the current version of documentation?  I’m not sure why.

build.gradle
```diff
@@ -350,6 +350,17 @@ ext.pomAttributes = {
   }
 }
 
+def getAllDependentProjects(project) {
+  if(project.configurations.hasProperty("runtime")) {
```
sahilTakiar: is this necessary? if the subproject uses the Java plugin, then it should always have the `runtime` configuration

build.gradle
sahilTakiar: yeah, but we only do releases every once in a while, so this could be linking to javadocs that are weeks old

for now we can leave it as is

build.gradle
kadaan: I think that it can be removed.

kadaan:
@sahilTakiar Updated

sahilTakiar:
LGTM

PR #777, opened 2016-03-01
Adding Parallelization to WorkUnit Creation in KafkaSource

tuGithub:


tuGithub:
@zliu41 @chavdar , please review it when you have time

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -104,34 +108,50 @@
...
 
     List<KafkaTopic> topics = getFilteredTopics(state);
     Map<String, State> topicSpecificStateMap = getTopicSpecificState(topics, state);
+    
+    int numOfThreads =
+        state.getPropAsInt(KAFKA_SOURCE_WORK_UNITS_CREATION_THREADS, DEFAULT_THREAD_COUNT);
+    ExecutorService threadPool = Executors.newFixedThreadPool(numOfThreads,
+        ExecutorsUtils.newThreadFactory(Optional.of(LOG)));
+    
+    if(topics != null){
```
zliu41: `getFilteredTopics` doesn't return null.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -493,4 +531,31 @@ public boolean apply(KafkaTopic input) {
...
+      this.topic = topic;
+      this.state = state;
+      this.topicSpecificState = topicSpecificState;
+      this.allTopicWorkUnits = workUnits;
+    }
+    
+    @Override
+    public void run() {
+      List<WorkUnit> oneTopicWorkUnits = 
+          this.kafkaSource.getWorkUnitsForTopic(this.topic, this.state, topicSpecificState);
```
zliu41: You don't need to pass `kafkaSource` to this class. Here you can do `KafkaSource.this.getWorkUnitsForTopic`

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -493,4 +531,31 @@ public boolean apply(KafkaTopic input) {
...
+        KafkaSource<S,D> kafkaSource, Map<String, List<WorkUnit>> workUnits){
+      this.kafkaSource = kafkaSource;
+      this.topic = topic;
+      this.state = state;
+      this.topicSpecificState = topicSpecificState;
+      this.allTopicWorkUnits = workUnits;
+    }
+    
+    @Override
+    public void run() {
```
zliu41: Please catch all Throwables occurred in this method and log them. Otherwise if this method failed the error will be swallowed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -342,7 +370,12 @@ private long getPreviousOffsetForPartition(KafkaPartition partition, SourceState
         partition.getTopicName(), partition.getId()));
   }
 
-  private void getAllPreviousOffsets(SourceState state) {
+  // need to be synchronized as this.previousOffsets need to be initialized once
+  private synchronized void getAllPreviousOffsets(SourceState state) {
+    if(this.doneGettingAllPreviousOffsets.get()){
```
zliu41: Since this method is synchronized and is the only place that changes the value of `doneGettingAllPreviousOffsets`, you don't need to use AtomicBoolean. Just need to use a `volatile` boolean.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -183,8 +203,12 @@
   }
 
   private void createEmptyWorkUnitsForSkippedPartitions(Map<String, List<WorkUnit>> workUnits,
-      Map<String, State> topicSpecificStateMap) {
+      Map<String, State> topicSpecificStateMap, SourceState state) {
 
+    if(!this.doneGettingAllPreviousOffsets.get()){
```
pcadabam: Why do you need this check? Isn't this check already in `getAllPreviousOffserts(state)`?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -104,34 +108,50 @@
...
+      LOG.info("Begin using thread pool to create work units for topic size " + topics.size());
+    }
+    
     for (KafkaTopic topic : topics) {
-      workUnits.put(topic.getName(),
-          getWorkUnitsForTopic(topic, state, Optional.fromNullable(topicSpecificStateMap.get(topic.getName()))));
+      LOG.info("Using thread pool to create work units for topic " + topic.getName());
+      threadPool.submit(new WorkUnitCreator(topic, state, Optional.fromNullable(topicSpecificStateMap.get(topic.getName())), this, workUnits));
     }
+
```
pcadabam: Maybe useful to log a message here. We will know how long we are waiting for all the threads to complete.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -104,34 +108,50 @@
...
+    }
+    
     for (KafkaTopic topic : topics) {
-      workUnits.put(topic.getName(),
-          getWorkUnitsForTopic(topic, state, Optional.fromNullable(topicSpecificStateMap.get(topic.getName()))));
+      LOG.info("Using thread pool to create work units for topic " + topic.getName());
+      threadPool.submit(new WorkUnitCreator(topic, state, Optional.fromNullable(topicSpecificStateMap.get(topic.getName())), this, workUnits));
     }
+    
+    ExecutorsUtils.shutdownExecutorService(threadPool, Optional.of(LOG), 3600L, TimeUnit.SECONDS);
```
pcadabam: 1 `TimeUnit.Hour`

pcadabam:
The implementation parallelizes at a topic level. Do you think we need parallelization at a partition level?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -104,34 +108,50 @@
    */
   @VisibleForTesting
   static final String KAFKA_TOPIC_SPECIFIC_STATE = "kafka.topic.specific.state";
+  public static final String KAFKA_SOURCE_WORK_UNITS_CREATION_THREADS = "kafka.source.work.units.creation.threads"; 
+  public static final int DEFAULT_THREAD_COUNT = 10;
```
pcadabam: We process around 300 topics. I think we can set the thread size to 30. If we hit throttling issues from kafka or heap issues we can bring it down.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
tuGithub: In case that function did not get called in threads pool

tuGithub:
Addressed all the comments, please reviewed the updated version

tuGithub:
@pcadabam ,

The implementation parallelizes at a topic level is much simpler than in partition level. Unless the performance still bad, prefer to do that in topic level

zliu41:
LGTM other than one comment. Please use a more meaningful commit message.

sahilTakiar:
Can you also change the title of this PR to something like `Adding Parallelization to WorkUnit Creation in KafkaSource`

pcadabam:
LGTM

gobblin-metrics/src/main/java/gobblin/metrics/reporter/util/HttpClientFactory.java
```diff
@@ -0,0 +1,20 @@
+package gobblin.metrics.reporter.util;
```
zliu41: please add copyright block

zliu41:
LGTM other than above

PR #780, opened 2016-03-01
Flatten Avro Schema to make it optimal for ORC

abti:
@chavdar Please review

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
+++ b/gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
@@ -1,0 +1,466 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+import org.apache.commons.lang3.StringUtils;
+import org.codehaus.jackson.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/***
+ * This class provides methods to flatten an Avro Schema to make it more optimal for ORC
+ * (Hive does not support predicate pushdown for ORC with nested fields: ETL-7214)
+ *
+ * The behavior of Avro Schema un-nesting is listed below:
+ *
+ * 1. Record within Record (and so on recursively) are flattened into the parent Record
+ * Record R1 {
+ *   fields: {[
+ *      {
+ *        Record R2 {
+ *          fields: {[
+ *              {
+ *                Record R3 {
+ *                  fields: {[
+ *                      {
+ *                        String S2
+ *                      }
+ *                  ]}
+ *                }, {
+ *                  String S3
+ *                }
+ *              }
+ *
+ *          ]}
+ *        }
+ *      }, {
+ *        String S1
+ *      }
+ *   ]}
+ * }
+ * will be flattened to:
+ * Record R1 {
+ *   fields: {[
+ *      {
+ *        String S1
+ *      }, {
+ *        String S2
+ *      }, {
+ *        String S3
+ *      }
+ *   ]}
+ * }
+ *
+ * 2. All fields un-nested from a Record within an Option (ie. Union of the type [null, Record] or [Record, null])
+ * within a Record are moved to parent Record as a list of Option fields
+ * Record R1 {
+ *   fields : {[
+ *      {
+ *        Union : [
+ *          null,
+ *          Record R2 {
+ *            fields : {[
+ *                {
+ *                  String S1
+ *                }, {
+ *                  String S2
+ *                }
+ *            ]}
+ *          }
+ *      }
+ *   ]}
+ * }
+ * will be flattened to:
+ * Record R1 {
+ *   fields : {[
+ *      {
+ *        Union : [ null, String S1]
+ *      }, {
+ *        Union : [ null, String S2]
+ *      }
+ *   ]}
+ * }
+ *
+ * 3. Array or Map will not be un-nested, however Records within it will be un-nested as described above
+ *
+ * 4. All un-nested fields are decorated with a new property "flatten_source" which is a dot separated string
+ * concatenation of parent fields name, similarly un-nested fields are renamed to double-underscore string
+ * concatenation of parent fields name
+ *
+ * 5. Primitive Types are not un-nested
+ */
+public class AvroFlattener {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroFlattener.class);
+
+  private static final String FLATTENED_NAME_JOINER = "__";
+  private static final String FLATTENED_SOURCE_JOINER = ".";
+  private static final String FLATTENED_SOURCE_KEY = "flatten_source";
+
+  private String flattenedNameJoiner;
+  private String flattenedSourceJoiner;
+
+  /***
+   * Flatten the Schema to un-nest recursive Records (to make it optimal for ORC)
+   * @param schema Avro Schema to flatten
+   * @return Flattened Avro Schema
+   */
+  public Schema flatten(Schema schema) {
+    Preconditions.checkNotNull(schema);
+
+    // To help make it configurable later
+    this.flattenedNameJoiner = FLATTENED_NAME_JOINER;
+    this.flattenedSourceJoiner = FLATTENED_SOURCE_JOINER;
+
+    Schema flattenedSchema = flatten(schema, false);
+
+    LOG.debug("Original Schema : " + schema);
+    LOG.debug("Flattened Schema: " + flattenedSchema);
+
+    return flattenedSchema;
+  }
+
+  /***
+   * Flatten the Schema to un-nest recursive Records (to make it optimal for ORC)
+   * @param schema Schema to flatten
+   * @param shouldPopulateLineage is set to true if the field is going to be flattened and moved up the hierarchy -
+   *                              so that lineage information can be tagged to it; which happens when there is a
+   *                              Record within a Record OR Record within Option within Record and so on,
+   *                              however not when there is a Record within Map or Array
+   * @return Flattened Avro Schema
+   */
+  private Schema flatten(Schema schema, boolean shouldPopulateLineage) {
+    Schema flattenedSchema;
+
+    // Process all Schema Types
+    // (Primitives are simply cloned)
+    switch (schema.getType()) {
+      case ARRAY:
+        // Array might be an array of recursive Records, flatten them
+        flattenedSchema = Schema.createArray(flatten(schema.getElementType(), false));
+        break;
+      case BOOLEAN:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case BYTES:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case DOUBLE:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case ENUM:
+        flattenedSchema =
+            Schema.createEnum(schema.getName(), schema.getDoc(), schema.getNamespace(), schema.getEnumSymbols());
+        break;
+      case FIXED:
+        flattenedSchema =
+            Schema.createFixed(schema.getName(), schema.getDoc(), schema.getNamespace(), schema.getFixedSize());
+        break;
+      case FLOAT:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case INT:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case LONG:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case MAP:
+        flattenedSchema = Schema.createMap(flatten(schema.getValueType(), false));
+        break;
+      case NULL:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case RECORD:
+        flattenedSchema = flattenRecord(schema, shouldPopulateLineage);
+        break;
+      case STRING:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case UNION:
+        flattenedSchema = flattenUnion(schema, shouldPopulateLineage);
+        break;
+      default:
+        String exceptionMessage = String.format("Schema flattening failed for \"%s\" ", schema);
+        LOG.error(exceptionMessage);
+
+        throw new AvroRuntimeException(exceptionMessage);
+    }
+
+    // Copy schema metadata
+    copyProperties(schema, flattenedSchema);
+
+    return flattenedSchema;
+  }
+
+  /***
+   * Flatten Record schema
+   * @param schema Record Schema to flatten
+   * @param shouldPopulateLineage If lineage information should be tagged in the field, this is true when we are
+   *                              un-nesting fields
+   * @return Flattened Record Schema
+   */
+  private Schema flattenRecord(Schema schema, boolean shouldPopulateLineage) {
+    Preconditions.checkNotNull(schema);
+    Preconditions.checkArgument(Schema.Type.RECORD.equals(schema.getType()));
+
+    Schema flattenedSchema;
+
+    List<Schema.Field> flattenedFields = new ArrayList<>();
+    if (schema.getFields().size() > 0) {
+      for (Schema.Field oldField : schema.getFields()) {
+        List<Schema.Field> newFields = flattenField(oldField, ImmutableList.<String>of(),
+            shouldPopulateLineage, Optional.<Schema>absent());
+        if (null != newFields && newFields.size() > 0) {
+          flattenedFields.addAll(newFields);
+        }
+      }
+    }
+
+    flattenedSchema = Schema.createRecord(schema.getName(), schema.getDoc(), schema.getNamespace(),
+        schema.isError());
+    flattenedSchema.setFields(flattenedFields);
+
+    return flattenedSchema;
+  }
+
+  /***
+   * Flatten Union Schema
+   * @param schema Union Schema to flatten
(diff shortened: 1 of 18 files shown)


kadaan:
@abti ORC has great support for complex types.  Why would you want to flatten the avro schema when converting the data to ORC?

zliu41:
@kadaan flattening improves query response time for many types of queries by a noticeable margin

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroFlattener.class);
+
+  /***
+   * Flatten the Schema to un-nest recursive Records (to make it optimal for ORC)
+   * @param schema Avro Schema to flatten
+   * @return Flattened Avro Schema
+   */
+  public static Schema flatten(Schema schema) {
+    // Validate input
```
zliu41: There are many unnecessary comments such as this one, line 43, 53, 156, etc. These are obvious from the code so please remove them

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+    Preconditions.checkNotNull(schema);
+
+    // Flatten schema
+    Schema flattenedSchema = flatten(schema, 0);
+
+    LOG.info("Original Schema : " + schema);
```
zliu41: Please change this to debug. A schema can be thousands of characters and if every schema is printed it takes a lot of space.

ibuenros:
@kadaan Hive does not support predicate pushdown for ORC with nested fields:
https://issues.apache.org/jira/browse/HIVE-7214

zliu41:
@abti should this PR include a Converter?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+  private static List<Schema.Field> flattenField(Schema.Field f, LinkedList<String> lineage, int level,
+      boolean shouldWrapInOption) {
+    Preconditions.checkNotNull(f);
+    Preconditions.checkNotNull(f.schema());
+    Preconditions.checkNotNull(f.name());
+
+    List<Schema.Field> flattenedFields = new ArrayList<>();
+
+    // Set lineage
+    if (null == lineage) {
```
zliu41: Doesn't seem possible for lineage to be null.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+          flattenedFields.addAll(flattenField(field, lineage, level+1, false));
+        }
+        lineage.removeLast();
+      }
+    }
+    // If field.Type = OPTION, un-nest its fields and return them
+    else {
+      if (isOfOptionType(f.schema())) {
+        Schema first = f.schema().getTypes().get(0);
+        Schema second = f.schema().getTypes().get(1);
```
zliu41: What if there are more than 2 options?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+        if (null != schema.getTypes() && schema.getTypes().size() > 0) {
+          for (Schema oldUnionMember : schema.getTypes()) {
+            // Its member might still recursively contain records
+            flattenedUnionMembers.add(flatten(oldUnionMember, level + 1));
+          }
+        }
+        flattenedSchema = Schema.createUnion(flattenedUnionMembers);
+        break;
+      default:
+        String exceptionMessage = String.format("Schema flattening failed for \"%s\" ", schema);
```
zliu41: Can you just have `case`s for RECORD, UNION, MAP and ARRAY, and use `default` for all others?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+    List<Schema.Field> flattenedFields = new ArrayList<>();
+
+    // Set lineage
+    if (null == lineage) {
+      lineage = new LinkedList<>();
+    }
+    lineage.add(f.name());
+
+    // If field.Type = RECORD, un-nest its fields and return them
+    if (Schema.Type.RECORD.equals(f.schema().getType())) {
```
zliu41: This seems to be similar logic as the `case RECORD` in line 95. Can you put it in a method `flattenRecord`?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+    // Copy schema metadata
+    copyProperties(schema, flattenedSchema);
+
+    return flattenedSchema;
+  }
+
+  /***
+   * Flatten Record field, and compute a list of flattened fields
+   * @param f Field to flatten
+   * @param lineage Lineage represented as a List of Strings
```
zliu41: Can you explain what `lineage` is using an example. It's not very clear to me.

abti:
@kadaan Like ibuenros mentioned, Hive does not support pushdown for ORC for nested fields 

@zliu41 We have phased the overall effort, Converter/Writer and Source will follow soon

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Yes, it won't be null

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: If there are two fields which are options, both get un-nested

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: We want default to throw exception. Modeled it around an Avro's internal class

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: I will add an example

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: I will remove them, though I like to have a few extra comments :)

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: I will change this.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Moved Record and Union flattening to methods

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+public class AvroFlattener {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroFlattener.class);
+
+  /***
+   * Flatten the Schema to un-nest recursive Records (to make it optimal for ORC)
+   * @param schema Avro Schema to flatten
+   * @return Flattened Avro Schema
+   */
+  public static Schema flatten(Schema schema) {
```
chavdar: Please make the methods instance-level. This will allow us to easily add configs and similar in the future.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+import org.apache.commons.lang3.StringUtils;
+import org.codehaus.jackson.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+public class AvroFlattener {
```
chavdar: Please add a high-level description of how the flattening works, what gets flattened, naming, metadata, etc.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+        break;
+      case NULL:
+        flattenedSchema = Schema.create(schema.getType());
+        break;
+      case RECORD:
+        List<Schema.Field> flattenedFields = new ArrayList<>();
+        if (null != schema.getFields() && schema.getFields().size() > 0) {
+          // Recursively flatten fields
+          for (Schema.Field oldField : schema.getFields()) {
+            LinkedList<String> lineage = new LinkedList<>();
```
chavdar: Shouldn't the lineage be a list of Schemas rather than just names. That give more flexibility, e.g. if you want to concatenate doc strings.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+        break;
+      case UNION:
+        List<Schema> flattenedUnionMembers = new ArrayList<>();
+        if (null != schema.getTypes() && schema.getTypes().size() > 0) {
+          for (Schema oldUnionMember : schema.getTypes()) {
+            // Its member might still recursively contain records
+            flattenedUnionMembers.add(flatten(oldUnionMember, level + 1));
+          }
+        }
+        flattenedSchema = Schema.createUnion(flattenedUnionMembers);
```
chavdar: 1. Union of a [record_type, null] and [null, record_type] should be handled as a special case. It should be flattened to the result of the flattening of record_type and all fields should be optional.
2. [a lower priority, might add as an extension in the future] The case of union of [record_type_1, record_type_2, ...] may be handled by combining the flattened record_type_1, record_type_2, ... and using pre-pending the name of the record types to each field name for disambiguation.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+    return flattenedFields;
+  }
+
+  /***
+   * Check if the Avro Schema is of type OPTION
+   * ie. [null, RECORD] or [RECORD, null]
+   * @param schema Avro Schema to check
+   * @return true if Avro Schema is of type OPTION else false
+   */
+  private static boolean isOfOptionType(Schema schema) {
```
chavdar: nit: it might be worthwhile to change that the method returns the record type if it is optional type as it is needed anyway.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+        lineage.removeLast();
+      }
+    }
+    // If field.Type = OPTION, un-nest its fields and return them
+    else {
+      if (isOfOptionType(f.schema())) {
+        Schema first = f.schema().getTypes().get(0);
+        Schema second = f.schema().getTypes().get(1);
+        Schema record = first;
+        boolean wrapInOption = Schema.Type.NULL.equals(first.getType());
```
chavdar: wrapInOption should be true in the case of [record_type, null] too.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,296 @@
...
+        if (StringUtils.isNotBlank(flattenSource)) {
+          field.addProp("flatten_source", flattenSource);
+        }
+        for (Map.Entry<String, JsonNode> entry : f.getJsonProps().entrySet()) {
+          field.addProp(entry.getKey(), entry.getValue());
+        }
+        flattenedFields.add(field);
+      }
+    }
+    lineage.removeLast();
```
chavdar: Modifying "lineage" and trying to revert it to its original value is very error-prone. Someone will later modify the code, forget to do it and all kinds of bad things may happen. I think it is best to make "lineage" immutable and make copies if necessary to modify it.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,357 @@
...
+   *            "name" : "nestedFieldString",
+   *            "type" : "string"
+   *          }, {
+   *            "name" : "nestedFieldInt",
+   *            "type" : "int"
+   *          } ]
+   *       }
+   *     }]
+   * }
+   * The expected output schema is:
```
zliu41: The output schema doesn't have "nestedRecordName". Is this expected?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
zliu41: Also is there any restrictions, e.g., can it handle recursive schema?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,357 @@
...
+
+  private static Schema flatten(Schema schema, boolean shouldPopulateLineage) {
+    Schema flattenedSchema;
+
+    // Process all Schema Types
+    // (Primitives are simply cloned)
+    switch (schema.getType()) {
+      case ARRAY:
+        // Array might be an array of recursive Records, flatten them
+        flattenedSchema = Schema.createArray(flatten(schema.getElementType(), false));
```
zliu41: Can you explain in the comments why use `false` here. Same for `MAP`.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,357 @@
...
+   * @return Flattened Record Schema
+   */
+  private static Schema flattenRecord(Schema schema, boolean shouldPopulateLineage) {
+    Preconditions.checkNotNull(schema);
+    Preconditions.checkArgument(Schema.Type.RECORD.equals(schema.getType()));
+
+    Schema flattenedSchema;
+
+    List<Schema.Field> flattenedFields = new ArrayList<>();
+    if (null != schema.getFields() && schema.getFields().size() > 0) {
```
zliu41: This line is unnecessary since `schema.getFields()` won't return null. Same for line 132.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,357 @@
...
+   *     }]
+   * }
+   * The expected output schema is:
+   * {
+   *    "type" : "record",
+   *    "name" : "parentRecordName",
+   *    "fields" : [ {
+   *      "name" : "parentFieldRecord__nestedFieldString",
+   *      "type" : "string",
+   *      "flatten_source" : "parentFieldRecord.nestedFieldString"
```
zliu41: Seems `name` already contains all information in `flatten_source`. Why is `flatten_source` needed?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
```diff
@@ -0,0 +1,357 @@
...
+      }
+    }
+    lineage.removeLast();
+
+    return flattenedFields;
+  }
+
+  /***
+   * Check if the Avro Schema is of type OPTION
+   * ie. [null, RECORD] or [RECORD, null]
```
zliu41: Is it possible to have a schema of type [RECORD, String, null]?

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Changed it

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Added

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Changed to Optional<Schema> and made use of it for checking / getting Record

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Changed

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Made it immutable

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Yes, thats expected. Earlier we were adding that to output but after a discussion we removed it.

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: I will add some comments, but in quick summary: shouldPopulateLineage is set to true if the field is going to be flattened and moved up the hierarchy - so that lineage information can be tagged to it. Which happens when there is Record within Record OR Record within Option with Record and so on, however not when there is Record within Map within Record

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Removed

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: It came in as a requirement

gobblin-utility/src/main/java/gobblin/util/AvroFlattener.java
abti: Yes, but we do not consider it as an option

gobblin-utility/src/test/java/gobblin/util/AvroFlattenerTest.java
```diff
@@ -0,0 +1,303 @@
+package gobblin.util;
+
+import org.apache.avro.Schema;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+public class AvroFlattenerTest {
```
chavdar: I think we should have some additional tests:
* Record within Record within Record
* Negative cases involving arrays or records

I think it will be easier to write tests if instead of inlining the schemas as string (lots of " escaping), schemas are loaded from resources: testX() reads the resources X.input.json and X.expected.json and verifies that flatten(X.input) == X.expected. Then we can easily add more tests.

abti: Moved json inputs to resources, and added more test cases (positive and negative)

chavdar:
One minor comment. Other than that, LGTM.

abti:
Thanks for the reviews everyone. Also addressed the minor comment by Chavdar. Merging

PR #783, opened 2016-03-02
Fix for NullReferenceException cause by #739

kadaan:
@sahilTakiar The lombok @synchronized attribute isn't initialized before the super call is done, therefore we needed to replace it with the framework synchronized keyword.


sahilTakiar:
LGTM

PR #793, opened 2016-03-03
Add TokenUtils

zliu41:
@ydai1124 can you review?

### DIFF
--- a/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanTags.java
+++ b/gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanTags.java
@@ -1,3 +1,15 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
 package gobblin.azkaban;
 
 import java.util.Map;
--- a/gobblin-yarn/build.gradle
+++ b/gobblin-yarn/build.gradle
@@ -32,6 +32,7 @@ dependencies {
   compile externalDependency.log4j
   compile externalDependency.commonsCli
   compile externalDependency.gson
+  compile externalDependency.hiveCommon
   compile externalDependency.metricsCore
   compile externalDependency.metricsJvm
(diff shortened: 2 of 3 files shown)


gobblin-azkaban/build.gradle
```diff
@@ -21,12 +21,20 @@ dependencies {
   compile project(":gobblin-utility")
   if (project.hasProperty('useHadoop2')) {
     compile project(":gobblin-yarn")
+    compile externalDependency.hadoopClientCommon
+    compile externalDependency.hadoopYarnCommon
+  } else {
+    compile group: 'org.apache.hadoop', name: 'hadoop-yarn-api', version: '2.3.0'
```
ydai1124: Why is this necessary? ../build.gradle already defines hadoop version.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanTokenUtils.java
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * @author ziliu
+ */
+public class AzkabanTokenUtils {
+
+  private static final Logger LOG = Logger.getLogger(AzkabanTokenUtils.class);
+
+  private static final String USER_TO_PROXY = "user.to.proxy";
+  private static final String OTHER_NAMENODES = "other.namenodes";
+  private static final String KEYTAB_USER = "idpc/eat1-nertzaz01.grid.linkedin.com@GRID.LINKEDIN.COM";
```
ydai1124: Should this be extracted from `state`?

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanTokenUtils.java
```diff
@@ -0,0 +1,259 @@
...
+  private static final String KEYTAB_LOCATION = "idpc.azkaban.service.keytab";
```
ydai1124: same here

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanTokenUtils.java
```diff
@@ -0,0 +1,259 @@
...
+      renewer = new Text("azkaban mr tokens");
+    }
+
+    return renewer;
+  }
+
+  /**
+   * Get a {@link HiveConf} using the specified metastore Uri.
+   */
+  public static HiveConf getHiveConf(String metastoreUri) throws IOException {
```
ydai1124: Is this completed? What about `HiveToken`?

gobblin-azkaban/build.gradle
zliu41: Otherwise it won't build with hadoop1.
I'm going to put this class in the gobblin-yarn submodule so that we don't need to do this.

zliu41:
@ydai1124 updated 7b28049

ydai1124:
LGTM

PR #794, opened 2016-03-04
Read current version of config store from metadata file

pcadabam:
## Changes
- Added reading and writing to store metadata file which will reside in _CONIFG_STORE directory
- Currently only active version is stored but can store any key value pairs
- Keep a back up of old store metadata file if case we want to recover from a failed deployment
- Added more unit tests
- Removed rollback for Deployable interface as the deploy will be used for rollback as well

@sahilTakiar can you review?


gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -471,44 +434,42 @@ public void deploy(FsDeploymentConfig deploymentConfig) throws IOException {
...
-      if (confStream.getInputStream().isPresent()) {
-        // Read the resource as a stream from the classpath and write it to HDFS
-        try (SeekableFSInputStream inputStream = new SeekableFSInputStream(confStream.getInputStream().get());
-            FSDataOutputStream os = this.fs.create(hdsfConfPath, false)) {
-          StreamUtils.copy(inputStream, os);
+        // If an empty directory needs to created it may not have a stream.
+        if (confStream.getInputStream().isPresent()) {
+          // Read the resource as a stream from the classpath and write it to HDFS
+          try (SeekableFSInputStream inputStream = new SeekableFSInputStream(confStream.getInputStream().get());
+              FSDataOutputStream os = this.fs.create(hdsfConfPath, false)) {
```
sahilTakiar: Can you add an explicit check to see if the config file already exists, that way we can control the error message and make it more descriptive than `FileAlreadyExistsException`

sahilTakiar:
Whats the reason for removing the `rollback` method? The problem with just having a `deploy` method is that now deployed version are basically mutable.

The advantage of the rollback method is that it shouldn't need to modify any files, it just changes the symlink file.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
pcadabam: In line 436 we skip the copying if the version already exists. Are you talking about duplicate files in the same version? I am not sure we should handle that case. That will bad data and `FileAlreadyExistsException` is fine in that case.

pcadabam:
A version is NOT mutable. Take a look at line 436 in the `deploy` method, we skip copy a new set of files from classpath only if the version does not exist on HDFS. For instance when we rollback from 0.7 to 0.3 no new configs will be copied to 0.3 directory even if the files have changed in classpath. We only update the metadata file to 0.3

sahilTakiar:
Ok thats fine, but still whats the reason for not having a `rollback` method that just sets the version.

pcadabam:
We do not have the need for that now. Let us add the `rollback` interface when such a requirement comes up.
From the `StoreDeployer` I always call the deploy method with the version to be deployed. It can either be an existing version (a rollback) or a new version.

sahilTakiar:
I'm confused, I thought the point of this PR was to add rollback functionality as stated in the PR description "Removed rollback for Deployable interface as the deploy will be used for rollback as well"

My question is whats the justification for using the `deploy` method for rollbacks instead of having an explicit `rollback` method? The `rollback` method shouldn't be particularly complex as you already have the work to update the symlink file. I think having a `rollback` method makes more sense than using the `deploy` method for rollbacks, at least from a usability perspective.

pcadabam:
Hi @sahilTakiar 
If we have separate methods `rollback` and `deploy` the user has to pass from command line or in the azkaban props the operation he is trying to do. Then the `StoreDeployer` needs to have logic to decide whether to call `SimpleHDFSStore.deploy` or `SimpleHDFSStore.rollback`.
With the way I have it, the user does not have to specify the operation from command line and `StoreDeployer` always calls `SimpleHDFSStore.deploy`
We can discuss offline if you still have concerns.

sahilTakiar:
Can rollbacks just be done from the command line? If its Azkaban there can be a separate job for rollbacks.

We can discuss more on Monday.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStore.java
```diff
@@ -161,36 +152,22 @@ protected SimpleHDFSConfigStore(FileSystem fs, URI physicalStoreRoot, URI logica
...
     try {
-      FileStatus[] fileStatuses = Iterables.toArray(
-          Iterables.filter(Arrays.asList(this.fs.listStatus(configStoreDir)), new FileStatusIsDir()), FileStatus.class);
-      Arrays.sort(fileStatuses, new FileStatusPathNameComparator());
-      return fileStatuses[fileStatuses.length - 1].getPath().getName();
+      return storeMetadata.getCurrentVersion();
     } catch (IOException e) {
-      throw new RuntimeException(
-          String.format("Error while checking current version for configStoreDir: \"%s\"", configStoreDir), e);
+      throw new RuntimeException("Error getting current version", e);
```
sahilTakiar: Any reason to remove the `configStoreDir` from the exception message?

pcadabam: Added back. `configStoreDir` was not used locally in this method

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param config to be serialized
+   */
+  void writeMetadata(Config config) throws IOException {
+
+    Path storeMetadataFileBkpPath = new Path(this.storeMetadataFilePath.getParent(), this.storeMetadataFilePath.getName() + ".bkp");
+
+    try {
+
+      // Delete old backup file if exists
+      if (this.fs.exists(storeMetadataFileBkpPath)) {
```
sahilTakiar: Check `HadoopUtils.deleteIfExists`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+        if (!this.fs.delete(storeMetadataFileBkpPath, true)) {
+          throw new IOException("Failed to delete " + storeMetadataFileBkpPath);
+        }
+      }
+
+      // Rename current storeMetadataFile to backup
+      if (this.fs.exists(this.storeMetadataFilePath)) {
```
sahilTakiar: Check `HadoopUtils.renamePath`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+          throw new IOException(String.format("Failed to create backup file for %s at %s", this.storeMetadataFilePath, storeMetadataFileBkpPath));
+        }
+      }
+
+      // Write new storeMetadataFile
+      try (FSDataOutputStream outputStream = this.fs.create(this.storeMetadataFilePath);) {
+        outputStream.write(config.root().render(ConfigRenderOptions.concise()).getBytes(Charsets.UTF_8));
+      }
+    } finally {
+      this.fs.delete(storeMetadataFileBkpPath, true);
```
sahilTakiar: do we want to try and restore the back up file if there is a problem creating the new file?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+
+import com.google.common.base.Charsets;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigRenderOptions;
+import com.typesafe.config.ConfigValueFactory;
+
+
+/**
+ * A metadata accessor for {@link SimpleHDFSConfigStore}. Every {@link SimpleHDFSConfigStore} will have a file named
```
sahilTakiar: I don't think this needs to be tied to `SimpleHDFSConfigStore` as it is usable by any HDFS-based `ConfigStore`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
pcadabam: Used `HadoopUtils`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
pcadabam: Used `HadoopUtils`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
pcadabam: Yes makes sense. Added logic to restore.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSStoreMetadata.java
pcadabam: Changed javadoc

sahilTakiar:
LGTM

* Had some offline discussion about whether to include the `rollback` feature - sticking to current approach for now to unblock this PR, and will discuss with @chavdar offline

PR #797, opened 2016-03-06
Fixing Fork + Task Retry Logic #776

sahilTakiar:
* Fix for #776
* Change is to cancel all `Future`s launched by a `Task`
    * A `Task` launches a `Future` for each `Fork` it runs, the PR ensures all the `Future`s are cancelled if any exception is thrown by the `Task`

@zliu41 can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/Task.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/Task.java
@@ -14,9 +14,9 @@ package gobblin.runtime;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.concurrent.CompletionService;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
@@ -25,7 +25,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.io.Closer;
 
 import gobblin.Constructs;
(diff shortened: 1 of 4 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -237,6 +233,16 @@ public void run() {
         LOG.error("Failed to close all open resources", t);
       }
 
+      for (Map.Entry<Optional<Fork>, Optional<Future<?>>> forkAndFuture : this.forks.entrySet()) {
```
zliu41: Instead of canceling all forks if one fork fails, can you let all forks finish, record the failed forks and when retrying the task, only re-run the failed forks?

sahilTakiar: I can but its slightly more complicated than it seems. The current PR preserves the current behavior with regards to isolation of failures between `Fork`s, but I don't think it is the ideal behavior.

For now it may be best to stick to the current approach and add better isolation semantics if there is need. The problem is a `Task` doesn't really support re-trying from a known good location, its all or nothing:
* We would have to carry over some state from the failed `Task` to the new `Task`
* We also need to consider which files get deleted if only some `Fork`s fail

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -470,20 +476,23 @@ private void processRecord(Object convertedRecord, ForkOperator forkOperator, Ro
     // until all puts succeed, at which point the task moves to the next record.
     while (!allPutsSucceeded) {
       allPutsSucceeded = true;
-      for (int i = 0; i < branches; i++) {
-        if (succeededPuts[i]) {
+
+      int branch = 0;
+      for (Optional<Fork> fork : this.forks.keySet()) {
+        if (succeededPuts[branch]) {
```
zliu41: branch++ before continue?

gobblin-runtime/src/test/java/gobblin/runtime/TaskTest.java
```diff
@@ -0,0 +1,117 @@
+package gobblin.runtime;
```
zliu41: license header

zliu41:
LGTM other than above

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
sahilTakiar: Yes good catch, fixed

Issue #798, opened 2016-03-06
Add JaCoCo and Coveralls Integration


sahilTakiar:
* Gobblin should have JaCoCo integration for code coverage
* It should also integrate with Coveralls which provides code coverage analysis

sahilTakiar:
Not sure what's wrong but it doesn't seem to be working. The `./gradlew coveralls` command is throwing a `StackOverflowError`, here is the full stack-trace:

```
17:48:09.906 [ERROR] [org.gradle.BuildExceptionReporter] FAILURE: Build failed with an exception.
17:48:09.906 [ERROR] [org.gradle.BuildExceptionReporter]
17:48:09.907 [ERROR] [org.gradle.BuildExceptionReporter] * What went wrong:
17:48:09.908 [ERROR] [org.gradle.BuildExceptionReporter] Execution failed for task ':coveralls'.
17:48:09.909 [ERROR] [org.gradle.BuildExceptionReporter] > java.lang.StackOverflowError (no error message)
17:48:09.911 [ERROR] [org.gradle.BuildExceptionReporter]
17:48:09.912 [ERROR] [org.gradle.BuildExceptionReporter] * Exception is:
17:48:09.914 [ERROR] [org.gradle.BuildExceptionReporter] org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':coveralls'.
17:48:09.915 [ERROR] [org.gradle.BuildExceptionReporter] 	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:69)
17:48:09.916 [ERROR] [org.gradle.BuildExceptionReporter] 	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:46)
17:48:09.916 [ERROR] [org.gradle.BuildExceptionReporter] 	at org.gradle.api.internal.tasks.execution.PostExecutionAnalysisTaskExecuter.execute(PostExecutionAnalysisTaskExecuter.java:35)
17:48:09.921 [ERROR] [org.gradle.BuildExceptionReporter] 	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:64)
[...]

PR #799, opened 2016-03-06
(#798) Adding JaCoCo and Coveralls support for code coverage analysis

sahilTakiar:
* Adding support for running JaCoCo and integration with Coveralls
* JaCoCO can be run using `./gradlew test jacocoTestReport` which will generate JaCoCo reports for each individual Java subproject
    * Output will be in `xml` and `html` format and can be found under `build/[subproject]/jacoco`
* Adding the Gradle Plugin `jacoco-full-report`
    * Docs: https://github.com/palantir/gradle-jacoco-coverage and https://plugins.gradle.org/plugin/com.palantir.jacoco-coverage
    * The default Gradle JaCoCo Plugin does not aggregate all reports into one single report, this is problematic as coveralls expects one aggregated report
    * This plugin automatically does the aggregation and is configured in a similar way to the Gradle `jacoco` plugin
    * It can be run using `./gradlew test jacocoFullReport`
* The Travis build has been modified so that JaCoCo report is uploaded to coveralls
* Coveralls Gobblin Page: https://coveralls.io/github/linkedin/gobblin
* Issue: #798

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,11 +16,15 @@ buildscript {
   }
   dependencies {
     classpath 'gradle.plugin.org.inferred:gradle-processors:1.1.2'
+    classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.6.3'
+    classpath 'gradle.plugin.com.palantir:jacoco-coverage:0.3.0'
   }
 }
 
 apply plugin: 'org.inferred.processors'
 apply plugin: 'idea'
+
+apply from: rootProject.projectDir.path + '/gradle/scripts/jacoco-coveralls-support.gradle'
 
 idea.project {
   ext.languageLevel = JavaVersion.VERSION_1_7
--- a/gradle/scripts/jacoco-coveralls-support.gradle
+++ b/gradle/scripts/jacoco-coveralls-support.gradle
@@ -1,0 +1,63 @@
+// Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at  http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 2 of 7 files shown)


kadaan:
Sweet!!!!!!

sahilTakiar:
Looks like there may be some problems with this, seems Travis is killing the builds because they are taking up too many resources. My guess is that the addition of the `jacoco` plugin is causing this. Will see if I can decrease the number of threads used.

sahilTakiar:
It seems setting `org.gradle.parallel` to `false` helps out. So I am disabling this property from Travis builds only. The build takes a little longer, but I think that should be ok since we want to increase the reliability of Travis builds.

sahilTakiar:
@chavdar can you review this?

sahilTakiar:
Updated so that all logic for JaCoCo and Coveralls support is in a separate `.gradle` file, also add more comments to the script.

sahilTakiar:
@chavdar can you review?

sahilTakiar:
@zliu41 can you review this?

gradle/scripts/jacoco-coveralls-support.gradle
```diff
@@ -0,0 +1,63 @@
...
+ * is why this plugin is necessary. The master report is also needed for Coveralls integration which expects a single
+ * aggregated report in <code>xml</code> format. The aggregated report will be placed under
+ * <code>build/reports/jacocoFullReport/</code>.
+ */
+
+apply plugin: 'com.github.kt3k.coveralls'
+apply plugin: 'com.palantir.jacoco-full-report'
+apply plugin: 'jacoco'
+
+ext.jacocoVersion = '0.7.1.201405082137'
```
zliu41: Is there any reason to use this version? Seems to be quite old.

build.gradle
```diff
@@ -16,12 +16,16 @@ buildscript {
   }
   dependencies {
     classpath 'gradle.plugin.org.inferred:gradle-processors:1.1.2'
+    classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:1.0.2'
```
zliu41: This seems to be an old version.

sahilTakiar:
@zliu41 updated, versions upgraded to the latest available

zliu41:
LGTM

PR #806, opened 2016-03-08
Fix permission of store metadata file to allow read access to all users

pcadabam:
@sahilTakiar can you review this small change?


sahilTakiar:
LGTM

Issue #802, opened 2016-03-07
addShutdownHook is causing a leak


kadaan:
@sahilTakiar As far as I can tell, the addition of shutdown hooks are causing a leak: [oom_heapdump_Leak_Suspects.zip](https://github.com/linkedin/gobblin/files/161907/oom_heapdump_Leak_Suspects.zip)

sahilTakiar:
hmmm, this is Gobblin on YARN right? I'm guessing this was introduced in: https://github.com/linkedin/gobblin/pull/741

`AbstractJobLauncher` adds a new shutdown hook for each job. Since Gobblin on YARN is continuously running these shutdown hooks never get invoked when a job completes. The fix is to move this shutdown hook to the `ApplicationLauncher`. I will take care of this as part of the work for https://github.com/linkedin/gobblin/issues/792 (should have a PR by EOD today)

kadaan:
Yes. It also seems to be important to call `remoteShutdownHook`.

sahilTakiar:
Fixed in #808

Issue #792, opened 2016-03-03
Reporting metrics on Gobblin Yarn causes App_Master to crash.


kadaan:
@zliu41 @sahilTakiar @liyinan926 As far as I can tell, there are major issues around metrics and running gobblin under yarn:
### 1. Reporter Leak - SEVERE

_Metric reporters are constantly added to the system and never removed._

On the app_master when the Quartz timer fires of a job, `AbstractJobLauncher.launchJob(JobListener)` is called.  Right away, this calls `GobblinMetrics.startMetricReporting(Properties)`.  This creates new file, jmx, kafka, and custom reporters.  When these reporters are created, they are added to the collection of reporters stored in the `RootMetricContext`.  Later, when the job finished, it calls `JobMetrics.stopMetricsReporting()`.  This ends up calling `RootMetricContext.stopReporting()` which stops, but **does not remove** all reporters.  Additionally, `GobblinMetrics.metricsReportingStarted` is set to `false`.  This means that a subsequent call to `GobblinMetrics.startMetricReporting(Properties)` will cause the new reporters to be created.

The net result is that the reporter collection in RootMetricContext has unbounded growth, causes excessive numbers of threads to be created, and depending on the reporter can cause the app_master to crash.
### 2. Metric Gaps - MODERATE

_The way reporting is started and stopped can result in gaps in metric reporting._

On the app_master when the Quartz timer fires of a job, `AbstractJobLauncher.launchJob(JobListener)` is called.  Right away, this calls `GobblinMetrics.startMetricReporting(Properties)`.  This calls `RootMetricContext.startReporting()`, which starts are associated reporters.  Later, when the job is done it calls `GobblinMetrics.stopMetricReporting()`, which calls `RootMetricContext.stopReporting()` method that stops all associated reporters.  This means that if the app_master is running two jobs, A & B, that the following scenario will cause a metric gap for Job B:

| Time | Job | Action |
| --- | --- | --- |
| T | A | Start |
| T + 1 | B | Start |
| T + 2 | A | Finish |
| T + 3 | B | **No Metrics Reported** |
| T + N | B | Finish |
### 3. Invalid Data - MILD

_All file reporters are assigned all contexts, which will result in the wrong data in the files._

File metric reporters write to files with filenames which are specific to the job.  Unfortunately, the list of contexts to which the reporters listen is not scoped to a given job.  This means that files may contain metrics from many different jobs.

sahilTakiar:
[...]

Issue #785, opened 2016-03-02
AbstractJobLauncher leaks metrics reporters when job cannot lock


kadaan:
@sahilTakiar AbstractJobLauncher can cause a leak of metrics reports when jobs fail to get the job lock because the metrics reporting is started, but JobException thrown when failing to get the lock causes metrics reporting to not get stopped.

```java
@Override
public void launchJob(JobListener jobListener) throws JobException {
  if (this.jobContext.getJobMetricsOptional().isPresent()) {
    this.jobContext.getJobMetricsOptional().get().startMetricReporting(this.jobProps);
  }
  try {
    if (!tryLockJob()) {
      throw new JobException(String.format("Previous instance of job %s is still running, skipping this scheduled run", this.jobContext.getJobName()));
    }
    ...
  } finally {
    try {
      ...
    } finally {
      unlockJob();
    }
  }
  ...
  // Stop metrics reporting
  if (this.jobContext.getJobMetricsOptional().isPresent()) {
    this.jobContext.getJobMetricsOptional().get().stopMetricsReporting();
    JobMetrics.remove(jobState);
  }
  ...
}
```

sahilTakiar:
Thanks for reporting the issue, looks like a bug.

kadaan:
@sahilTakiar There is more to it than this.  Still looking into it.

kadaan:
@sahilTakiar This issue correctly fixes the described but, but there are a number of others which affect Gobblin on Yarn.  I will open another issue to track those.

sahilTakiar:
Closing as #764 has been merged

PR #808, opened 2016-03-08
(#792, #802) Adding ApplicationLauncher to manage app services, including GobblinMetrics lifecyle

sahilTakiar:
#### Issues

* Partial fix for #792
* Full fix for #802

#### Description

* Adding an `ApplicationLauncher` interface that is responsible for launching Gobblin applications
    * Gobblin Applications are launched in a variety of ways - `AzkabanJobLauncher`, `GobblinApplicationMaster`, `CliLocalJobLauncher`, `CliMRJobLauncher`, `SchedulerDaemon`
* A base implementation of `ApplicationLauncher` called `ServiceBasedApplicationLauncher` defines a `ApplicationLauncher` is a collection of Guava `Service`s that need to be run
    * A core set of `Service`s are always run, and sub-classes can add more `Service`s
* All Gobblin Application Launchers mentioned above now extend `ServiceBasedApplicationLauncher`
    * A lot of common functionality between all these classes has been moved into `ServiceBasedApplicationLauncher`
* A `MetricsReportingService` has been added that starts and stops metrics reporting - this fixes the two issues mentioned above since metrics will only be started and stopped once per application
* The `gobblin-scheduler` submodule has been deleted and its classed moved into `gobblin-runtime` - this way `gobblin-runtime` can use the scheduler classes in a more generic manner
* Added a `JMXReportingService` for collecting JVM stats and reporting them via JMX (this code was duplicated in a few places so I moved it into a separate class)

#### Reviewing this PR

* `ApplicationLauncher` and `ServiceBasedApplicationLauncher` are the best places to start


gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -28,10 +27,20 @@
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
+import org.apache.hadoop.conf.Configuration;
```
kadaan: import ordering: org comes before com

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -125,12 +144,41 @@ public void run()
...
+        stop();
+      } finally {
+        close();
+      }
     }
   }
 
+  @Override
+  public void start() throws ApplicationException {
+    this.applicationLauncher.start();
```
kadaan: Should we add listener events here?  Maybe as an enhancement.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -125,12 +144,41 @@ public void run()
...
+  }
+
+  @Override
+  public void stop() throws ApplicationException {
+    this.applicationLauncher.stop();
```
kadaan: The same as above.  If this was done, job developers could also have resources that are initialized once at job start and cleaned up at job end.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -204,10 +202,6 @@ public void apply(JobListener jobListener, JobContext jobContext) throws Excepti
 
   @Override
   public void launchJob(JobListener jobListener) throws JobException {
-    if (this.jobContext.getJobMetricsOptional().isPresent()) {
```
kadaan: Annoyingly, these changes are going to cause #764 to not merge correctly.  Would be really nice to get #764 in as it was fixing a couple of other bugs.

gobblin-runtime/src/main/java/gobblin/runtime/local/CliLocalJobLauncher.java
```diff
@@ -12,27 +12,86 @@
...
 
-import gobblin.runtime.cli.CliOptions;
-
+import java.io.IOException;
 import java.util.Properties;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+import com.google.common.io.Closer;
```
kadaan: Import ordering

gobblin-runtime/src/main/java/gobblin/runtime/local/CliLocalJobLauncher.java
```diff
@@ -12,27 +12,86 @@
...
+  }
+
+  @Override
+  public void cancelJob(@Nullable JobListener jobListener) throws JobException {
+    this.localJobLauncher.cancelJob(jobListener);
+  }
+
+  @Override
+  public void close() throws IOException {
+    this.close();
```
kadaan: This seems to be calling the wrong thing.  Shouldn't this be `this.closer.close()`?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/CliMRJobLauncher.java
```diff
@@ -12,20 +12,28 @@
 
 package gobblin.runtime.mapreduce;
 
+import java.io.IOException;
 import java.util.Properties;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+import com.google.common.io.Closer;
```
kadaan: Import ordering

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+      LOG.warn("ApplicationLauncher has already stopped");
+      return;
+    } else {
+      this.hasStopped = true;
+    }
+
+    LOG.info("Shutting down the application");
+    try {
+      // Give the services 5 seconds to stop to ensure that we are responsive to shutdown requests
+      this.serviceManager.stopAsync().awaitStopped(5, TimeUnit.SECONDS);
```
kadaan: 5 seconds seems really short.  The `GobblinApplicationMaster` used to wait for 5 minutes.

gobblin-yarn/src/main/java/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -202,17 +166,12 @@ public synchronized void stop() {
 
     // Send a shutdown request to the containers as a second guard in case Yarn could not stop the containers
     sendShutdownRequest();
-
     try {
-      // Give the services 5 minutes to stop to ensure that we are responsive to shutdown requests
-      this.serviceManager.stopAsync().awaitStopped(5, TimeUnit.MINUTES);
-    } catch (TimeoutException te) {
-      LOGGER.error("Timeout in stopping the service manager", te);
+      super.stop();
```
kadaan: Doesn't this mean that `GobblinApplicationMaster` will only wait 5 seconds for shutdown now, rather than 5 minutes?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
```diff
@@ -174,6 +162,7 @@ public GobblinWorkUnitRunner(String applicationName, String helixInstanceName, C
       LOGGER.info("Adding YarnContainerSecurityManager since login is keytab based");
       services.add(new YarnContainerSecurityManager(config, fs, this.eventBus));
     }
+    services.add(new JMXReportingService());
```
kadaan: Why don't we use `ServiceBasedAppLauncher` for the services in the container as well?

kadaan:
Am I correct that this fixes 1 and 2 of #792 as well as fixing #785?

gobblin-runtime/src/main/java/gobblin/runtime/services/MetricsReportingService.java
```diff
@@ -0,0 +1,44 @@
...
+  private final String appId;
+
+  public MetricsReportingService(Properties properties, String appId) {
+    this.properties = properties;
+    this.appId = appId;
+  }
+
+  @Override
+  protected void startUp() throws Exception {
+    GobblinMetrics.get(this.appId).startMetricReporting(this.properties);
```
kadaan: How is this going to work with the `buildFileMetricReporter()` call made from with `startMetricReporting`?  That method constructs an outputstream which is specific for the job_id that is running.

kadaan:
@sahilTakiar I'm running this locally and so far the issues seem to be fixed.  I still have metrics and jmx disabled, but I'll try it with them enabled later today.

sahilTakiar:
@kadaan awesome! good to know. I will work on rebasing this PR and addressing your comments sometime today.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
sahilTakiar: You mean add `ApplicationListener`s to the `start()` and `stop()` method of `ApplicationLauncher`?

kadaan: That was what I was wondering.  It could be useful for users to be able to setup services that need to  start/stop when the application start/stops.

sahilTakiar: Yeah, but the `ServiceBasedAppLauncher.addService` method should be able to handle that. Although it may still be useful for other things so I will still add it.

kadaan: The benefit is that JobListeners can be added just by specifying them in the config files.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
sahilTakiar: 5 minutes seems excessive, I'm going to change it to 1 minute

kadaan: Can this be configurable?  Either in the config file or by the job launcher?

gobblin-runtime/src/main/java/gobblin/runtime/services/MetricsReportingService.java
sahilTakiar: It should just be specific to `this.appId`. The `get(id)` method just constructs a new `GobblinMetrics` instance with a given `id`. This `id` is used when constructing the `OutputStreamMetricReporter`

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
sahilTakiar: Yeah I can make this configurable, with a default of 1 minute

gobblin-yarn/src/main/java/gobblin/yarn/GobblinWorkUnitRunner.java
sahilTakiar: We could be there are services like the Job History Store and Admin UI that aren't really needed. Also the Metrics Service would be different as `GobblinWorkUnitRunner` uses `ContainerMetrics`

sahilTakiar:
@kadaan updated
* Rebased the PR
* `ServiceBasedAppLauncher` can now run a configurable set of `Service`s specified by the config key `app.additional.services`
* Addressed other comments

sahilTakiar:
@zliu41 can you also take a look at this when you have a chance?

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -0,0 +1,242 @@
...
+   * that invokes {@link #stop()} and the {@link #close()} methods. So {@link #stop()} and {@link #close()} need not be
+   * called explicitly; they can be triggered during the JVM shutdown.
+   */
+  @Override
+  public void start() {
+    if (this.hasStarted) {
+      LOG.warn("ApplicationLauncher has already started");
+      return;
+    } else {
+      this.hasStarted = true;
```
zliu41: should set this.hasStopped = false?

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -0,0 +1,242 @@
...
+    // Start the application
+    this.serviceManager.startAsync();
+  }
+
+  /**
+   * Stops the {@link ApplicationLauncher} by stopping all associated services.
+   */
+  @Override
+  public void stop() throws ApplicationException {
+    if (this.hasStopped) {
```
zliu41: should be `if (!this.hasStarted || this.hasStopped)`?

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -0,0 +1,242 @@
...
+      LOG.warn("ApplicationLauncher has already stopped");
+      return;
+    } else {
+      this.hasStopped = true;
```
zliu41: should set  hasStarted = false?

zliu41: actually why not just use a single boolean variable to denote started vs stopped.

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -49,7 +58,7 @@
  *
  * @author Yinan Li
  */
-public class AzkabanJobLauncher extends AbstractJob {
+public class AzkabanJobLauncher extends AbstractJob implements ApplicationLauncher, JobLauncher {
```
zliu41: why does this class need to implement `JobLauncher`? When should one call `launchJob` directly instead of `run`?

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -0,0 +1,242 @@
...
+            LOG.error("Failed to close application", e);
+          }
+        }
+      }
+    });
+
+    LOG.info("Starting the Gobblin application and all its associated Services");
+
+    // Start the application
+    this.serviceManager.startAsync();
```
zliu41: should probably add `this.serviceManager.awaitHealthy()`

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
sahilTakiar: A Guava `Service` cannot be re-started, so the `ServiceBasedAppLauncher` shouldn't be able to restart either, which is why two variables are necessary.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
sahilTakiar: Added

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
sahilTakiar: Its mostly just to declare that `AzkabanJobLauncher` is a `JobLauncher` - that is, it is capable of launching jobs. In this specific scenario there wouldn't really be case where someone would directly call `launchJob`, I think it just makes the code easier to understand.

sahilTakiar:
@zliu41 updated and comments addressed

gobblin-azkaban/src/main/java/gobblin/azkaban/AzkabanJobLauncher.java
zliu41: It may be a little confusing. Because `job.launcher.type` can only be LOCAL, MAPREDUCE or YARN. It cannot be AzkabanJobLauncher. So technically it's not a job launcher, at least not in the same sense as LocalJobLauncher or MRJobLauncher.

sahilTakiar: True, but I think its more like `CliLocalJobLauncher extends LocalJobLauncher` or `CliMRJobLauncher extends MRJobLauncher` - but I wanted to avoid using `extends` since many of these classes already `extend` some other class.

`JobLauncherFactory` only has two publicly available launch types, `LOCAL` and `MAPREDUCE` and provides implementations for both. I could add the `AzkabanJobLauncher` here, but it wouldn't provide any benefit.

I think `JobLauncherFactory` is useful for creating `JobLauncher`s in code, but that shouldn't prevent creation of `JobLauncher`s that are directly invoked using a `main` method.

zliu41:
LGTM

sahilTakiar:
@kadaan any other comments?

kadaan:
@sahilTakiar LGTM (metrics work)

PR #815, opened 2016-03-09
Improve Hive registration performance

zliu41:
Summary:

- Parallelized generating `HiveSpec`s in `HiveRegistrationPublisher` using a `CompletionService`.
- In `HiveMetaStoreBasedRegister`, added methods `needToUpdateTable` and `needToUpdatePartition` to determine whether an existing table/partition needs to be altered. Previously it was using the `equals` method in `Table` and `Partition` but that was not a good choice since Hive automatically adds or modifies properties to existing tables and partitions, such as last access time, location (from raw to fully qualified), retention, isCompressed, etc., which causes `equals` to never return true. This means existing tables and partitions are always altered which is expensive.
- Added `haveSameSchema` method in `HiveSerDeManager`. This helps reduce the number of unnecessary schema updates which is expensive.
- Removed `HiveStripedLocks` since Hive metastore (usually RDBMS) should be able to handle concurrent requests.
- Added `hashCode` and `equals` to `HiveRegistrationUnit`, `HiveTable` and `HivePartition`.


zliu41:
@ydai1124 can you review?

gobblin-hive-registration/src/main/java/gobblin/hive/HivePartition.java
```diff
@@ -38,12 +38,42 @@
...
+  private final List<String> values;
+
   private HivePartition(Builder builder) {
     super(builder);
     this.values = ImmutableList.<String> copyOf(builder.values);
   }
 
-  private List<String> values;
+  @Override
+  public int hashCode() {
```
ibuenros: You should consider using lombok's `equalandhashcode`

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -231,11 +210,9 @@ public boolean existsTable(String dbName, String tableName) throws IOException {
   public boolean existsPartition(String dbName, String tableName, List<Column> partitionKeys,
       List<String> partitionValues) throws IOException {
     IMetaStoreClient client = borrowClient();
-    HiveLock lock = locks.getTableLock(dbName, tableName);
-    lock.lock();
 
     try {
-      client.getPartition(tableName, dbName, partitionValues);
+      client.getPartition(dbName, tableName, partitionValues);
```
ibuenros: Double check this order, I believe it is wrong.

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -302,51 +272,162 @@ private void addOrAlterPartition(IMetaStoreClient client, Table table, Partition
...
+    if (newSd.isSetCols() && !newSd.getCols().equals(existingSd.getCols())) {
+      log.info(
+          logMsgPrefix + String.format("existing columns: %s, new columns: %s", existingSd.getCols(), newSd.getCols()));
+      return true;
+    }
+
+    // Convert locations of both existingPartition and newPartition are raw paths. So that if one path is raw
+    // and the other is fully qualified, they can still be compared.
+    if (existingSd.isSetLocation()) {
+      existingSd.setLocation(new Path(existingSd.getLocation()).toUri().getRawPath());
```
ibuenros: What if the paths are in different namenodes?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -338,6 +351,124 @@ public void setSerDeType(String serDeType) {
...
+    }
+    HiveRegistrationUnit other = (HiveRegistrationUnit) obj;
+    if (this.bucketColumns == null) {
+      if (other.bucketColumns != null) {
+        return false;
+      }
+    } else if (!this.bucketColumns.equals(other.bucketColumns)) {
+      return false;
+    }
+    if (this.columns == null) {
```
ydai1124: What if `this.columns != null` and `other.columns == null`? Similar issue with the rest comparison block.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -338,6 +351,124 @@ public void setSerDeType(String serDeType) {
...
+    result = prime * result + ((this.numBuckets == null) ? 0 : this.numBuckets.hashCode());
+    result = prime * result + ((this.outputFormat == null) ? 0 : this.outputFormat.hashCode());
+    result = prime * result + ((this.serDeType == null) ? 0 : this.serDeType.hashCode());
+    result = prime * result + ((this.tableName == null) ? 0 : this.tableName.hashCode());
+    return result;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
```
ydai1124: What about creating a pluggable `HiveResigrationUnitComparator`?

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
```diff
@@ -139,4 +161,18 @@ public void updateSchema(HiveRegistrationUnit existingUnit, HiveRegistrationUnit
     }
   }
 
+  @Override
+  public boolean haveSameSchema(HiveRegistrationUnit unit1, HiveRegistrationUnit unit2) {
```
ydai1124: What about moving it to `HiveRegistrationUnit`?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -302,51 +272,162 @@ private void addOrAlterPartition(IMetaStoreClient client, Table table, Partition
...
-            client.getPartition(table.getTableName(), table.getDbName(), partition.getValues());
-        partition.setCreateTime(existingPartition.getCreateTime());
-        partition.setLastAccessTime(existingPartition.getLastAccessTime());
+            client.getPartition(table.getDbName(), table.getTableName(), partition.getValues());
+
         if (needToUpdatePartition(existingPartition, partition)) {
           client.alter_partition(table.getDbName(), table.getTableName(), partition);
-          log.info(String.format("Updated partition %s in table %s with location %s", partition, table.getTableName(),
-              partition.getSd().getLocation()));
-        } else {
```
ydai1124: Why removing this?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnit.java
ydai1124: These can be the default properties to compare. But for other cases, like hive distcp, properties like "location" will be customized.

gobblin-hive-registration/src/main/java/gobblin/hive/avro/HiveAvroSerDeManager.java
zliu41: This method needs to be overridden by subclasses of `HiveAvroSerDeManager`

zliu41:
@ydai1124 updated. Removed equals method in `HiveRegistrationUnit`, and added `HiveRegistrationUnitComparator`. fa4d3b5

zliu41:
updated

ydai1124:
LGTM

PR #817, opened 2016-03-09
fix findbug warnings

ydai1124:


ydai1124:
Trivial change. Self approve.

PR #818, opened 2016-03-10
Fixing FindBugs build to pick up the correct excludes file

sahilTakiar:


sahilTakiar:
Trivial change self-approving

PR #824, opened 2016-03-10
Create TimestampBasedCopyableDataset and version selection policy

ydai1124:
Moved RB655435 to github. @chavdar Can you take a look?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.copy;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
+import com.google.common.collect.Lists;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.joda.time.DateTime;
+
+import gobblin.data.management.copy.CopyConfiguration;
+import gobblin.data.management.copy.CopyableDataset;
+import gobblin.data.management.copy.CopyableFile;
+import gobblin.data.management.policy.TimeBasedSelectionPolicy;
+import gobblin.data.management.policy.VersionSelectionPolicy;
+import gobblin.data.management.version.TimestampedDatasetVersion;
+import gobblin.data.management.version.finder.DateTimeDatasetVersionFinder;
+import gobblin.data.management.version.finder.VersionFinder;
+import gobblin.dataset.FileSystemDataset;
+import gobblin.util.filters.HiddenFilter;
+import gobblin.util.ExecutorsUtils;
+import gobblin.util.PathUtils;
+import gobblin.util.executors.ScalingThreadPoolExecutor;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * Implementation of {@link CopyableDataset}. It depends on {@link #datasetVersionFinder} to find dataset versions and
+ * {@link #versionSelectionPolicy} to select the dataset versions for copying. {@link #datasetVersionFinder} is pluggable
+ * and must implement the interface {@link VersionSelectionPolicy<TimestampedDatasetVersion>}.
+ *
+ * The default logic ({@link #isCopyableFile(Path, FileSystem, FileSystem)}) for determining if a file is
+ * {@link CopyableFile} is based on the file existence and modified_timestamp at source and target {@link
+ * FileSystem}s.
+ */
+@Slf4j
+@Getter
+@SuppressWarnings("unchecked")
+public class TimestampBasedCopyableDataset implements CopyableDataset, FileSystemDataset {
+
+  private final Path datasetRoot;
+  private final VersionFinder<TimestampedDatasetVersion> datasetVersionFinder;
+  private final VersionSelectionPolicy<TimestampedDatasetVersion> versionSelectionPolicy;
+  private final ExecutorService executor;
+  private final FileSystem srcFs;
+
+  public static final String DATASET_VERSION_FINDER = "timestamp.based.copyable.dataset.version.finder";
+  public static final String DEFAULT_DATASET_VERSION_FINDER = DateTimeDatasetVersionFinder.class.getName();
+
+  public static final String COPY_POLICY = "timestamp.based.copyable.dataset.copy.policy";
+  public static final String DEFAULT_COPY_POLICY = TimeBasedSelectionPolicy.class.getName();
+
+  public static final String THREADPOOL_SIZE_TO_GET_COPYABLE_FILES = "threadpool.size.to.get.copyable.files";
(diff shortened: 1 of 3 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
```diff
@@ -0,0 +1,227 @@
...
+          (VersionSelectionPolicy<TimestampedDatasetVersion>) copyPolicyClass.getConstructor(Properties.class)
+              .newInstance(props);
+      Class<?> timestampedDatasetVersionFinderClass =
+          Class.forName(props.getProperty(DATASET_VERSION_FINDER, DEFAULT_DATASET_VERSION_FINDER));
+      this.datasetVersionFinder =
+          (VersionFinder<TimestampedDatasetVersion>) timestampedDatasetVersionFinderClass.getConstructor(
+              FileSystem.class, Properties.class).newInstance(this.srcFs, props);
+    } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException
+        | InvocationTargetException exception) {
+      throw new IOException(exception);
```
chavdar: Why not throw a RuntimeException? None of these seem to be I/O related.

gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
```diff
@@ -0,0 +1,227 @@
...
+    this.dateTimeZone =
+        DateTimeZone.forID(props.getProperty(DateTimeDatasetVersionFinder.DATE_TIME_PATTERN_TIMEZONE_KEY,
+            gobblin.data.management.version.finder.DateTimeDatasetVersionFinder.DEFAULT_DATE_TIME_PATTERN_TIMEZONE));
+
+    try {
+      Class<?> copyPolicyClass = Class.forName(props.getProperty(COPY_POLICY, DEFAULT_COPY_POLICY));
+      this.versionSelectionPolicy =
+          (VersionSelectionPolicy<TimestampedDatasetVersion>) copyPolicyClass.getConstructor(Properties.class)
+              .newInstance(props);
+      Class<?> timestampedDatasetVersionFinderClass =
```
chavdar: Are other version finder classes really supported? What are the restrictions? Is it any class that implements VersionFinder<TimestampedDatasetVersion> or should it be a subclass DateTimeDatasetVersionFinder? The code has a dependency on some config properties for DateTimeDatasetVersionFinder. What will happen if we inject some completely different class?

I think it is OK if the version finder is fixed unless you see specific use-cases to be replaced. 

Further, the current implementation re-uses some configuration properties from DateTimeDatasetVersionFinder but it seems like the logic to read those is slightly different (e.g. for DATE_TIME_PATTERN_TIMEZONE_KEY). Can't we just depend on the DateTimeDatasetVersionFinder instance to get those?

gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
```diff
@@ -0,0 +1,227 @@
...
+      for (Path locationToCopy : locationsToCopy) {
+        long timestampFromPath = this.versionDatetime.getMillis();
+        try {
+          for (FileStatus singleFile : this.srcFs.listStatus(locationToCopy, this.filter)) {
+            Path singleFilePath = singleFile.getPath();
+            log.debug("Checking if it is a copyable file: " + singleFilePath);
+            Path relativePath = PathUtils.relativizePath(singleFilePath, datasetRoot);
+            Path targetPath = new Path(targetRoot, relativePath);
+            if (this.isCopyableFile(singleFile, targetPath)) {
+              log.info("Will create workunit for: " + singleFilePath);
```
chavdar: This may create too many log lines if we try to copy many (big) topics.

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
```diff
@@ -0,0 +1,167 @@
...
+.put(TimestampBasedCopyableDataset.COPY_POLICY, TimeBasedCopyPolicyForTest.class.getName());
+    props.put(TimestampBasedCopyableDataset.DATASET_VERSION_FINDER,
+        TimestampedDatasetVersionFinderForTest.class.getName());
+    FileSystem localFs = FileSystem.getLocal(new Configuration());
+    TimestampBasedCopyableDataset copyabledataset =
+        new TimestampBasedCopyableDatasetForTest(localFs, props, new Path("/data/tracking/PVE"));
+    Collection<CopyableFile> copyableFiles = copyabledataset.getCopyableFiles(localFs, null);
+    Assert.assertEquals(copyableFiles.size(), 6);
+    Pattern pattern = Pattern.compile("dummy/[\\d]*/file[12]");
+    for (CopyableFile copyableFile : copyableFiles) {
```
chavdar: This check will return "true" if the getCopyableFiles() method returns the same file 6 times.

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
```diff
@@ -0,0 +1,167 @@
...
+    props.put(DateTimeDatasetVersionFinder.DATE_TIME_PATTERN_KEY, "yyyy/MM/dd");
+    props
+.put(TimestampBasedCopyableDataset.COPY_POLICY, TimeBasedCopyPolicyForTest.class.getName());
+    props.put(TimestampBasedCopyableDataset.DATASET_VERSION_FINDER,
+        TimestampedDatasetVersionFinderForTest.class.getName());
+    FileSystem localFs = FileSystem.getLocal(new Configuration());
+    TimestampBasedCopyableDataset copyabledataset =
+        new TimestampBasedCopyableDatasetForTest(localFs, props, new Path("/data/tracking/PVE"));
+    Collection<CopyableFile> copyableFiles = copyabledataset.getCopyableFiles(localFs, null);
+    Assert.assertEquals(copyableFiles.size(), 6);
```
chavdar: Please describe the test setup (e.g. why 6 is the magical correct number).

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
```diff
@@ -0,0 +1,167 @@
...
+import gobblin.data.management.version.finder.DateTimeDatasetVersionFinder;
+import gobblin.data.management.version.finder.VersionFinder;
+import gobblin.dataset.Dataset;
+
+
+/**
+ * Test for {@link TimestampBasedCopyableDataset}.
+ */
+@Test(groups = { "gobblin.data.management.copy" })
+public class TimestampBasedCopyableDatasetTest {
```
chavdar: Some additional tests: processing config options correctly (whatever property is set does indeed change the state of the object), failure scenarios (e.g. CopyableFileGenerator throws an exception), checking the file comparison logic.

gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
ydai1124: The configuration keys related to `DateTimeDatasetVersionFinder` are leftover (my bad), and they are stick to  `DateTimeDatasetVersionFinder` itself, which are never explicitly used in `TimestampBasedCopyableDataset`. I will remove them.

`TimestampBasedCopyableDataset` supports any class that implements `VersionFinder<TimestampedDatasetVersion>`. In the current code base, both `ModDateTimeDatasetVersionFinder` and `DateTimeDatasetVersionFinder` are supported. I will add more java doc.

gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
ydai1124: Fixed.

gobblin-data-management/src/main/java/gobblin/data/management/copy/TimestampBasedCopyableDataset.java
ydai1124: Changed to debug.

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
ydai1124: Fixed.

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
ydai1124: Fixed.

gobblin-data-management/src/test/java/gobblin/data/management/copy/TimestampBasedCopyableDatasetTest.java
ydai1124: Added more tests.

ydai1124:
@chavdar Updated.

chavdar:
LGTM

PR #826, opened 2016-03-10
(#825) fixing and improving useHadoop2 and hadoopVersion gradle properties

### LINKED ISSUE #825: Gobblin picks up wrong artifacts when specifying Hadoop version
When building Gobblin as follows
`./gradlew clean assemble -PhadoopVersion=2.*`
without specifying the `-PuseHadoop2` flag, the following lines in `build.gradle` do not execute as expected:

```
          compile(externalDependency.hadoopCommon) {
            exclude module: 'servlet-api'
          }
          compile externalDependency.hadoopClientCore
          compile externalDependency.hadoopAnnotations
          if (project.name.equals('gobblin-runtime') || project.name.equals('gobblin-test')) {
            compile externalDependency.hadoopClientCommon
          }
        } else {
          compile externalDependency.hadoop
        }
```

Even though building for Hadoop 2, the else statement will trigger, adding the wrong jars to the build (hadoop 2 uses `hadoop-common`, hadoop 1 uses `hadoop-core`). Explicitly adding the flag `-PuseHadoop2` should fix this, but we should be able to infer whether it's version 1 or 2 from the hadoop version.

sahilTakiar:
* Created a separate `.gradle` file for setting the `useHadoop2` property
* The property is now dynamically set based on the hadoop version specified
* The original behavior of `-PuseHadoop2` is preserved
* The only change is that if `-PhadoopVersion=` a 2.x.x version, then `useHadoop2` does not need to be specified also
* Fixes #825

@ibuenros can you review?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,10 @@ File getEnvironmentScript()
 
 apply from: environmentScript
 
+allprojects {
+  apply from: rootProject.projectDir.path + '/gradle/scripts/hadoop-version.gradle'
+}
+
 ext.publishToMaven = project.hasProperty('publishToMaven')
 if (ext.publishToMaven) {
     plugins.apply('maven')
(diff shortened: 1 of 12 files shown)


sahilTakiar:
@ibuenros can you review?

ibuenros:
LGTM

PR #828, opened 2016-03-10
Add zookeeper based job lock for gobblin yarn

kadaan:
@sahilTakiar Here is the `ZookeeperBasedJobLock` that we discussed in #754.  The only part I'm a little unsure about is the cancellation of a running job if the connection to ZK is lost.  I think it is needed to ensure that duplicate jobs are not running, but I'm not sure if the mean of cancellation is correct.


sahilTakiar:
@kadaan is this PR ready for review? Can you also add some documentation about this under `gobblin-docs/user-guide/Gobblin-on-Yarn.md`

kadaan:
Yes, it is ready for review.  Please consider the following while reviewing:

> The only part I'm a little unsure about is the cancellation of a running job if the connection to ZK is lost. I think it is needed to ensure that duplicate jobs are not running, but I'm not sure if the mean of cancellation is correct.

kadaan:
Docs are coming

kadaan:
@sahilTakiar Added docs

kadaan:
@sahilTakiar Anything else outstanding for this PR?

sahilTakiar:
I just need to find some time to go through the code, will try to take a look today.

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLock.java
```diff
@@ -27,40 +28,53 @@
  *
  * @author Yinan Li
  */
-public interface JobLock {
+public abstract class JobLock implements Closeable {
```
sahilTakiar: can this be an interface instead of an abstract class?

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLockFactory.java
```diff
@@ -0,0 +1,55 @@
...
+  public static JobLock getJobLock(Properties properties, JobLockEventListener jobLockEventListener)
+          throws JobLockException {
+    JobLock jobLock;
+    if (properties.containsKey(ConfigurationKeys.JOB_LOCK_TYPE)) {
+      try {
+        Class<?> jobLockClass = Class.forName(
+                properties.getProperty(ConfigurationKeys.JOB_LOCK_TYPE, FileBasedJobLock.class.getSimpleName()));
+        jobLock = (JobLock) jobLockClass.newInstance();
+      } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+        jobLock = new FileBasedJobLock();
```
sahilTakiar: It may be better to just throw the exception, this way if a user has a typo like in `job.lock.type=gobblin.runtime.locks.MyCustomLock` they know something is wrong, rather than the code silently defaulting to a `FileBasedJobLock`

gobblin-runtime/src/main/java/gobblin/runtime/locks/ZookeeperBasedJobLock.java
```diff
@@ -0,0 +1,231 @@
...
+            switch (connectionState) {
+              case LOST:
+              case SUSPENDED:
+                log.warn("Lost connection with zookeeper");
+                for (Map.Entry<String, JobLockEventListener> lockEventListener : lockEventListeners.entrySet()) {
+                  log.warn("Informing job %s that lock was lost", lockEventListener.getKey());
+                    lockEventListener.getValue().onLost();
+                }
+                break;
+              case CONNECTED:
```
sahilTakiar: might want to add a separate `log.info` statement and `break` statement

gobblin-runtime/src/test/java/gobblin/runtime/locks/JobLockTest.java
```diff
@@ -0,0 +1,145 @@
...
+ * Base for all {@link JobLock} unit tests.
+ *
+ * @author Joel Baranick
+ */
+@Test(groups = {"gobblin.runtime"})
+public abstract class JobLockTest {
+
+    public abstract JobLock getJobLock() throws JobLockException;
+
+    public void testLocalJobLock() throws Exception {
```
sahilTakiar: can you change the name of this method? shouldn't this method be tagged with the `Test` annotation?

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
```diff
@@ -121,6 +120,9 @@ public GobblinHelixJobLauncher(Properties jobProps, HelixManager helixManager, F
     this.stateSerDeRunnerThreads = Integer.parseInt(jobProps.getProperty(ParallelRunner.PARALLEL_RUNNER_THREADS_KEY,
         Integer.toString(ParallelRunner.DEFAULT_PARALLEL_RUNNER_THREADS)));
 
+    URI fsUri = URI.create(jobProps.getProperty(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
+    this.fs = FileSystem.get(fsUri, new Configuration());
```
sahilTakiar: what the reason for moving the `FileSystem` creation into this class?

gobblin-runtime/src/main/java/gobblin/runtime/locks/ZookeeperBasedJobLock.java
```diff
@@ -0,0 +1,231 @@
...
+                log.warn("Regained connection with zookeeper");
+                break;
+            }
+          }
+      });
+      newCuratorFramework.start();
+      curatorFramework = newCuratorFramework;
+    }
+    if (locksReaper == null) {
+      ChildReaper newLocksReaper = new ChildReaper(
```
sahilTakiar: Whats this needed for? Also looking at the JavaDocs if we upgrade to Curator 2.9.0 we don't need this anymore: http://curator.apache.org/apidocs/org/apache/curator/framework/recipes/locks/ChildReaper.html

kadaan: Great. I'll upgrade to 2.9.0 and remove this code

gobblin-runtime/src/test/java/gobblin/runtime/locks/JobLockTest.java
kadaan: I copied this from the original FileBasedJobLockTeat

gobblin-runtime/src/main/java/gobblin/runtime/locks/ZookeeperBasedJobLock.java
kadaan: Sure

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLock.java
kadaan: Yes

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLockFactory.java
kadaan: Yes, I will handle missing and invalid differently

gobblin-yarn/src/main/java/gobblin/yarn/GobblinHelixJobLauncher.java
kadaan: AFAIK the only thing it was used for was the FileBasedJobLock and rather than pass it in just for that, let the JobLock initialize itself from properties.

kadaan:
Updated to address comments.

kadaan:
@sahilTakiar Ended up putting the fix for #891 in here as well as I couldn't build without it.

kadaan:
@sahilTakiar I think this should address all the comments.

kadaan:
@sahilTakiar Is there anything else you want to see changed?

kadaan:
Is there anything blocking this being merged?

sahilTakiar:
hey @kadaan sorry for the delay. I probably don't have time to look at this within the new week. @zliu41 can you review?

kadaan:
@zliu41 Is this something you could review?  It is beginning to be difficult for me to keep all of my PRs up to date and ensure that what we are running doesn't drift too far from upstream gobblin.

kadaan:
@sahilTakiar @zliu41 Do you have time to review this?

build.gradle
```diff
@@ -378,16 +407,18 @@ subprojects {
...
 
     sourceCompatibility = JavaVersion.VERSION_1_7
 
-    findbugs {
-      toolVersion = findBugsVersion
-      ignoreFailures = true
-      effort = "max"
-      // The exclude filter file must be under "ligradle/findbugs/"
-      excludeFilter = file(rootProject.projectDir.path + "/ligradle/findbugs/findbugsExclude.xml")
+    if (project.hasProperty("disableFindBugs")) {
```
sahilTakiar: Shouldn't it be `if (!project.hasProperty("disableFindBugs"))` - also is this necessary? You can always disable findBugs by running `./gradlew assemble` instead of `./gradlew build`

kadaan: Yes it should. I didn't know you can do that.  I will remove this.

build.gradle
```diff
@@ -42,8 +44,33 @@ File getEnvironmentScript()
 
 apply from: environmentScript
 
+// Maven POM generation is not thread safe, so serialize all the Upload tasks we can use `--parallel`.
+// https://issues.gradle.org/browse/GRADLE-2492
+// When we start building with 2.3 and later we should remove this and just add a common output dir for all tasks and let Gradle serialize them
+def lock = new ReentrantLock()
+def available = lock.newCondition()
```
sahilTakiar: Is this condition variable necessary, isn't it just sufficient to use a lock.

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLock.java
```diff
@@ -27,40 +28,51 @@
...
+public interface JobLock extends Closeable {
+
+  /**
+   * Initializes the lock.
+   *
+   * @param properties  the job properties
+   * @param jobLockEventListener the listener for lock events
+   * @throws JobLockException thrown if the {@link JobLock} fails to initialize
+   */
+  void initialize(Properties properties, JobLockEventListener jobLockEventListener)
```
sahilTakiar: Can we have an `abstract` class called `ListenableJobLock` that takes in a `Properties` and a `JobLockEventListener` in the constructor. Would rather avoid adding an `initialize` method to the interface as it is a backwards incompatible change to the API.

gobblin-runtime/src/main/java/gobblin/runtime/locks/ZookeeperBasedJobLock.java
```diff
@@ -0,0 +1,242 @@
...
+  public static final String LOCKS_ACQUIRE_TIMEOUT_MILLISECONDS = "zookeeper.locks.acquire.timeout.milliseconds";
+  public static final String CONNECTION_STRING = "zookeeper.connection.string";
+  public static final String CONNECTION_TIMEOUT_SECONDS = "zookeeper.connection.timeout.seconds";
+  public static final String SESSION_TIMEOUT_SECONDS = "zookeeper.session.timeout.seconds";
+  public static final String RETRY_BACKOFF_SECONDS = "zookeeper.retry.backoff.seconds";
+  public static final String MAX_RETRY_COUNT = "zookeeper.retry.count.max";
+
+  private String lockPath;
+  private long lockAcquireTimeoutMilliseconds;
+  private InterProcessSemaphoreMutex lock;
```
sahilTakiar: Change variable declaration to interface `InterProcessLock`

gobblin-runtime/src/main/java/gobblin/runtime/locks/ZookeeperBasedJobLock.java
```diff
@@ -0,0 +1,242 @@
...
+  private static CuratorFramework curatorFramework;
+  private static ConcurrentMap<String, JobLockEventListener> lockEventListeners = Maps.newConcurrentMap();
+  private static Thread curatorFrameworkShutdownHook;
+
+  public static final String LOCKS_ACQUIRE_TIMEOUT_MILLISECONDS = "zookeeper.locks.acquire.timeout.milliseconds";
+  public static final String CONNECTION_STRING = "zookeeper.connection.string";
+  public static final String CONNECTION_TIMEOUT_SECONDS = "zookeeper.connection.timeout.seconds";
+  public static final String SESSION_TIMEOUT_SECONDS = "zookeeper.session.timeout.seconds";
+  public static final String RETRY_BACKOFF_SECONDS = "zookeeper.retry.backoff.seconds";
+  public static final String MAX_RETRY_COUNT = "zookeeper.retry.count.max";
```
sahilTakiar: We are trying to make any new configuration keys compatible with Typesafe Config - that being said dots should no longer be substitutions for spaces, when defining a new config key. So this configs should read something like `gobblin.locks.zookeeper.retry.max_count` and `gobblin.locks.zookeeper.retry.backoff_seconds`

sahilTakiar:
Mostly a few new minor comments, otherwise LGTM.

gobblin-runtime/src/main/java/gobblin/runtime/locks/JobLock.java
kadaan: I don't think this can be moved to the constructor as we have no way to ensure that implementations still provide a constructor with those arguments.  As a compromise, I will add an `abstract` class called `ConfigurableJobLock` with the `void configure(Properties properties, JobLockEventListener jobLockEventListener)` method.  The factory will check if the `JobLock` extends `ConfigurableJobLock ` and if so it will call the `configure` method.

kadaan:
@sahilTakiar Addressed your comments.

Issue #778, opened 2016-03-01
ParallelRunner.movePath semantics broken by #741


kadaan:
It is no longer a move, it is a copy.

sahilTakiar:
Can this be closed?

kadaan:
Yes

PR #829, opened 2016-03-10
Adds support to `HadoopUtils` for overwriting files

### LINKED ISSUE #718: Publishing fails if same named file exists

kadaan:
@sahilTakiar This can be enabled for data publishing by adding `data.publisher.overwrite.enabled=true` to the gobblin config.  Fixes #718 & #778

### DIFF
--- a/gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
+++ b/gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
@@ -41,6 +41,7 @@ public class FsRenameCommitStep extends CommitStepBase {
   private final Path dstPath;
   private final String srcFsUri;
   private final String dstFsUri;
+  private final boolean overwrite;
   private transient FileSystem srcFs;
   private transient FileSystem dstFs;
 
@@ -55,6 +56,7 @@ public class FsRenameCommitStep extends CommitStepBase {
     this.dstFs = builder.dstFs != null ? builder.dstFs
         : getFileSystem(this.props.getProp(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
     this.dstFsUri = this.dstFs.getUri().toString();
+    this.overwrite = builder.overwrite;
   }
 
   public static class Builder<T extends Builder<?>> extends CommitStepBase.Builder<T> {
@@ -71,6 +73,7 @@ public class FsRenameCommitStep extends CommitStepBase {
     private Path dstPath;
     private FileSystem srcFs;
     private FileSystem dstFs;
+    private boolean overwrite;
 
     @Override
     public T withProps(State props) {
@@ -98,6 +101,12 @@ public class FsRenameCommitStep extends CommitStepBase {
     @SuppressWarnings("unchecked")
     public T withDstFs(FileSystem dstFs) {
       this.dstFs = dstFs;
+      return (T) this;
+    }
+
+    @SuppressWarnings("unchecked")
+    public T overwrite() {
+      this.overwrite = true;
       return (T) this;
     }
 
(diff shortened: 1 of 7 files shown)


gobblin-core/src/main/java/gobblin/commit/FsRenameCommitStep.java
```diff
@@ -131,6 +131,7 @@ public void execute() throws IOException {
       this.dstFs = getFileSystem(this.dstFsUri);
     }
     log.info(String.format("Moving %s to %s", this.srcPath, this.dstPath));
-    HadoopUtils.movePath(this.srcFs, this.srcPath, this.dstFs, this.dstPath, this.dstFs.getConf());
+    boolean overwrite = this.props.getPropAsBoolean(ConfigurationKeys.DATA_PUBLISHER_OVERWRITE_ENABLED, false);
```
sahilTakiar: for this class can you set the `overwrite` property in the builder, since this `CommitStep` may not always be used in the publisher

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -173,13 +187,31 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+   * Moves a src {@link Path} from a srcFs {@link FileSystem} to a dst {@link Path} on a dstFs {@link FileSystem}. If
+   * the srcFs and the dstFs have the same scheme, and neither of them or S3 schemes, then the {@link Path} is simply
+   * renamed. Otherwise, the data is from the src {@link Path} to the dst {@link Path}. So this method can handle copying
+   * data between different {@link FileSystem} implementations.
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the source {@link Path} which will me moved
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to move data to
+   */
```
sahilTakiar: update javadoc to document the `overwrite` property

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -203,13 +235,43 @@ public static void movePath(FileSystem srcFs, Path src, FileSystem dstFs, Path d
...
+   *
+   * <p>
+   *   TODO this method does not handle cleaning up any local files leftover by writing to S3.
+   * </p>
+   *
+   * @param srcFs the source {@link FileSystem} where the src {@link Path} exists
+   * @param src the {@link Path} to copy from the source {@link FileSystem}
+   * @param dstFs the destination {@link FileSystem} where the dst {@link Path} should be created
+   * @param dst the {@link Path} to copy data to
+   */
```
sahilTakiar: same, update javadoc please

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -203,13 +235,43 @@ public static void movePath(FileSystem srcFs, Path src, FileSystem dstFs, Path d
...
+  public static void copyPath(FileSystem srcFs, Path src, FileSystem dstFs, Path dst, boolean overwrite,
+                              Configuration conf) throws IOException {
+
+    copyPath(srcFs, src, dstFs, dst, false, overwrite, conf);
+  }
+
+  private static void copyPath(FileSystem srcFs, Path src, FileSystem dstFs, Path dst, boolean deleteSource,
```
sahilTakiar: i thought we wanted to avoid having a method that can delete the source, even if it is an internal method

sahilTakiar:
Mostly minor comments

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
kadaan: I was more of the mind that any exposed methods shouldn't combine copy and move together.  I'm less concerned if it is a private method.

kadaan:
@sahilTakiar Updated addressing most of your comments.

sahilTakiar:
LGTM

Issue #814, opened 2016-03-09
Gobblin Yarn log size is unbounded


kadaan:
The current solution for handling logs in Gobblin Yarn results in logs files which keep growing.  To handle this, we should support log rotation.  

Additionally, while looking into this, it seems like `LogCopier` may no longer be needed when running [Yarn 2.6.0](https://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-common/yarn-default.xml) or greater (`yarn.nodemanager.log-aggregation.roll-monitoring-interval-seconds`, but it might be slower than the current solution.

liyinan926:
I believe we do have log rotation built in. @zliu41 can comment on this.

zliu41:
There are ways to avoid a single log file to grow unboundedly (`MaxMinutesPerLogFile` and `MaxBytesPerLogFile`) but I don't think there's log rotation support.

kadaan:
@zliu41 It looks like if either of those configuration keys are specified `LogCopier` will rotate the files in the `createNewLogFileIfNeeded()` method.

PR #834, opened 2016-03-11
Enable the maximum log file size for Gobblin Yarn LogCopier to be configured

kadaan:
The new configuration key `gobblin.yarn.log.copier.max.file.size` will allow this to be configured with values such as: 100000, "10M", "512K", etc.  If it is not configured, the log files will not rotate.

Fixes: #814


kadaan:
@zliu41 LogCopier works correctly if maximum file size is specified.  The Web UI is still a pain as it represents it as one large file, making opening, downloading, etc. really slow.  I can imagine that this will only get worse as the Yarn application runs even longer.
screen shot 2016-03-14 at 7 48 40 am

sahilTakiar:
I think you need to rebase your commit, otherwise this generally LGTM.

What UI are you referring to?

kadaan:
I'm referring to the Hadoop Resource Manager web UI.

kadaan:
@sahilTakiar Rebased

PR #836, opened 2016-03-11
Updating README.md to reference ReadTheDocs documentation

sahilTakiar:
* Removing IRC batch and we discussed its no longer going to be supported
* Updating `README.md` file so that it points to new documentation on ReadTheDocs

@pcadabam can you review?


pcadabam:
LGTM

PR #841, opened 2016-03-14
change TaskDone notification for forks to be not racy

Kaiserchen:


gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -91,18 +94,16 @@
...
   private Optional<DataWriter<Object>> writer = Optional.absent();
 
-  // This is used by the parent task to signal that it has done pulling records and this fork
-  // should not expect any new incoming data records. This is written by the parent task and
-  // read by this fork. Since this flag will be updated by only a single thread and updates to
-  // a boolean are atomic, volatile is sufficient here.
-  private volatile boolean parentTaskDone = false;
+  // Submitted to the recordQueue by the parent task to indicated that no
+  // more records are to be expected
+  private static final Object PARENTTASKDONE = new Object();
```
sahilTakiar: Can you change the variable name to `PARENT_TASK_DONE` in order to be consistent with the Gobblin coding style: http://gobblin.readthedocs.org/en/latest/developer-guide/CodingStyle/

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -134,18 +135,9 @@ public Fork(TaskContext taskContext, Object schema, int branches, int index) thr
       buildWriterIfNotPresent();
     }
 
-    this.recordQueue = BoundedBlockingRecordQueue.newBuilder()
-        .hasCapacity(this.taskState.getPropAsInt(
+    this.recordQueue = new LinkedBlockingQueue<Object>(this.taskState.getPropAsInt(
```
sahilTakiar: Any reason to not use the `BoundedBlockingRecordQueue`, I think we should keep it if we can as it provides important statistics such as queue file ratio, mean rate of put attempts, etc.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -319,12 +317,12 @@ public String toString() {
...
 
+  /**
+   * Tell this fork that the parent task is done. This is a second chance call if the parent
+   * task failed and didn't do so through the normal way of calling markParentTaskDone().
+   */ 
   @Override
   public void close() throws IOException {
-    // Tell this fork that the parent task is done. This is a second chance call if the parent
-    // task failed and didn't do so through the normal way of calling markParentTaskDone().
-    this.parentTaskDone = true;
```
sahilTakiar: should this invoke `markParentTaskDone`?

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
```diff
@@ -302,7 +300,7 @@ public int getIndex() {
    *         which means it may be absent if collecting of queue statistics is not enabled.
    */
   public Optional<BoundedBlockingRecordQueue<Object>.QueueStats> queueStats() {
-    return this.recordQueue.stats();
+    return Optional.absent();
```
zliu41: What does this change mean?

zliu41:
It's ok to remove the timeout on getting a record from the queue which is no longer needed, but I think there should still be a timeout on putting a record into the queue. The benefit of having a timeout for put is that, if there are multiple forks in a task, and one fork is slow (hence its queue is often full), the task can timeout the put for this fork and go to the next fork.

Otherwise the entire task is blocked by the slow fork.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
Kaiserchen: Sure thing, will do

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
Kaiserchen: With my current patch, the fork does not keep the state of the ParentTask anymore it will just signal itself to shutdown. The markParentTaskDone method itself only fails when the Task gets interrupted. We could keep state if we did signal already and only signal once, wich is I think the best thing todo

I quickly need to check what this implies all the logic in Task,processRecord()
It looks like one stalled fork will stall the whole task as before, wich is wanted behaviour i guess, and my patch is not changing to much of it, except that fork 1 can't see a record n before fork0 anymore, but fork1 still can't see record n+1 before fork 0 saw record n. Maybe this whole logic can also be removed then. I think the gain to keep it is very small.

Kaiserchen:
@zliu41 the entire task will be blocked by a slow fork at the moment anyhow. Wich is wanted behaviour i guess. IMO the gain of the whole looping thing in Task,processRecord() is very little, it just shifts the place where they see nor record for long time from the position between n to n+1 to n-1 to n. This only has an impact when its the last record. In this case the fast forks will still be done faster then the problematic fork that caused the blocking. I am actually in favour of throwing this `while (!allPutsSucceeded) {` away, as layed out the benefit is not really there IMO.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
Kaiserchen: Yes, I also liked the stats, have to inspect the code more to see if I can get a blocking bahavior out of it. Maybe having Interger.MAX_VALUE as a timeout.

zliu41:
@Kaiserchen that's true. Then can you change the return type of `putRecord` to void.

Kaiserchen:
Changing the queue with statistics to give the wanted behavior is a little more invasive. I also need to touch/ write some unittests then. Will take me a little longer

sahilTakiar:
@Kaiserchen I agree the `while (!allPutsSucceeded)` block in `Task.processRecord`. I believe the main benefit is that for a single record, a single slow `Fork` will not block other `Fork`s from seeing the record. However, the next record will not be processed, until all `Fork`s have accepted the record.

I'm also ok with removing this logic, it should simplify the code also. 

Removing the timeouts from the `BoundedBlockingRecordQueue` also sounds reasonable.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
sahilTakiar: Yes, I see your point, see my comment below.

I think the problem here is that the original behavior of the `Fork` class allows a the `close` method to be invoked before `markParentTaskDone` is invoke. The original `close` implementation will ensure the `while (true)` loop in `Fork.processRecords` exits properly, even if `markParentTaskDone` has not been invoked. I think we need to preserve this behavior.

gobblin-runtime/src/main/java/gobblin/runtime/Fork.java
sahilTakiar: The `BoundedBlockingRecordQueue` class is backed by a `BlockingDeque`; we use the `BlockingDeque.offer(record, timeout, timeunit)` and `BlockingDeque.poll(timeout, timeunit)` methods. This `BlockingDeque` class also has equivalent `put(record)` and `take()` methods that don't require specifying a timeout.

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingDeque.html

kadaan: Additionally, I don't think we want to used LinkedBlockingQueue as it's performance is worse then ArrayBlockingQueue.

Kaiserchen:
Did all the changes that seemed to be required, Junittest seem to deadlock. If someone else want to fix this. Highly appreaciated.

abti:
@shirshanka Can you please review this, since this might be a possible change that may impact streaming?

Kaiserchen:
Hi,

Sorry yall, not sure if i have time to pull it over the finish line. I can
maybe chop of a few hours next week. If not Ill let you know. My company is
only now migrating towards gobblin found this back then while evaluating.

Btw gobblin streaming? Really? :Sad_face:

Am 12.01.2017 04:16 schrieb "Abhishek Tiwari" <notifications@github.com>:

abti:
https://issues.apache.org/jira/browse/GOBBLIN-121 

Please update your PR title with following prefix: [GOBBLIN-121]

PR #843, opened 2016-03-14
Make SimpleHiveSpec.Builder fields accessible via @Getter

zliu41:
Since they need to be accessed in subclasses of `SimpleHiveSpec`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -184,7 +184,7 @@ ext.externalDependency = [
   "lombok":"org.projectlombok:lombok:1.16.8",
   "mockRunnerJdbc":"com.mockrunner:mockrunner-jdbc:1.0.8",
   "xerces":"xerces:xercesImpl:2.11.0",
-  "typesafeConfig": "com.typesafe:config:1.3.0",
+  "typesafeConfig": "com.typesafe:config:1.2.1",
   "byteman": "org.jboss.byteman:byteman:" + bytemanVersion,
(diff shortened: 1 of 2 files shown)


ibuenros:
@zliu41 I don't like this change. Using non-singular fields makes it very hard to create `SimpleHiveSpecs` with activities or predicates. Removing the `@Builder` makes adding new fields in the future error prone and tedious. Why not just make the fields protected?

zliu41:
How to make them protected? `@Singular` will automatically declare them as private, and adding `@Getter` is not allowed.

One way is to add the getter methods in builder manually.

zliu41:
Turned out I only needed to access `path`, `table` and `partition` in the builder, so added `@Getter` annotation to these fields.

build.gradle
```diff
@@ -184,7 +184,7 @@ ext.externalDependency = [
   "lombok":"org.projectlombok:lombok:1.16.8",
   "mockRunnerJdbc":"com.mockrunner:mockrunner-jdbc:1.0.8",
   "xerces":"xerces:xercesImpl:2.11.0",
-  "typesafeConfig": "com.typesafe:config:1.3.0",
+  "typesafeConfig": "com.typesafe:config:1.2.1",
```
ibuenros: Is there a reason for this downgrade?

zliu41: I upgraded it to 1.3.0 in a PR earlier today, turned out 1.3.0 only supports JAVA 8

ibuenros:
LGTM

PR #851, opened 2016-03-15
Update Avro version to 1.8.0

zliu41:
Looks like there are quite a few bug fixes / improvements in 1.8.0. https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310911&version=12323299

Release note mentioned some backward incompatible changes. Tested on Kafka ingestion and compaction and no issue popped up.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,7 +104,7 @@ if (!project.hasProperty('bytemanVersion')) {
   ext.bytemanVersion = '2.2.1'
 }
 
-ext.avroVersion = '1.7.7'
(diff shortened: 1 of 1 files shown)


kadaan:
Will this cause an issue? https://issues.apache.org/jira/browse/AVRO-1586

zliu41:
I believe this just makes Avro build with Hadoop 2 by default; Hadoop 1 is still supported. And indeed avro-mapred-1.8.0-hadoop1.jar is available in Maven.

kadaan:
Great

PR #852, opened 2016-03-15
gobblin retention bug fixes

pcadabam:
Bugs fixed -
1. Support deprecated key gobblin.retention.dataset.profile.class
2. Add null check to array of FileStatus returned by fs.globStatus

@tuGithub can you review?


tuGithub:
LGTM

PR #853, opened 2016-03-15
Sanitize extracts and dataset urns in distpc that are used as path tokens

ibuenros:
@chavdar this should fix the issue with the invalid staging path.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
@@ -32,6 +32,7 @@ import gobblin.source.extractor.extract.AbstractSource;
 import gobblin.source.workunit.Extract;
 import gobblin.source.workunit.WorkUnit;
 import gobblin.util.HadoopUtils;
+import gobblin.util.PathUtils;
 import gobblin.util.RateControlledFileSystem;
 import gobblin.util.WriterUtils;
 import gobblin.util.executors.ScalingThreadPoolExecutor;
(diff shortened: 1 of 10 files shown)


ibuenros:
@sahilTakiar can you review? Thanks.

sahilTakiar:
What are "path tokens" exactly? Why is it necessary to do this to the dataset urn when writing to the `FsStateStore`, it seems like a pretty major restriction to force all dataset urn's to be of a certain length.

It may be better to have a separate class that extends `FsStateStore`, so we can add the method to the new class.

ibuenros:
File token is a single name in a Path (e.g. /data/my/path, the tokens are "data", "my", and "path"). This has to be done because character : is invalid in HDFS, / has special semantics, and HDFS does not allow any token to be longer than 250 bytes.

sahilTakiar:
Is that a bug in HDFS, are there any upstream JIRAs we can link this PR to? Would it be better to throw an exception rather than shorten the token? This way users are confused when they see a table name randomly shortened.

Also, when I tried to test this, the following `-mkdir` command worked, is there some exception that is suppose to be thrown?

```
hdfs dfs -mkdir 57483q9ohjfjdkal90173uy54hmnaioapufhgksayuc87f7193204oaoalhgnbkshquytynzmxhdhfkiqueornvnagvbztqr2416323447809afkufuiudaojklfdabualfdjakluywqruoahfnbyhfuds7a98047234347895jkhgnmfdauimhzvisifeoq9173jk4hg98472jfhnsrvyodaf790231h34jk24759o10rhamvioahfjekda7yiohrje791h42jk4hg8ehrjkyufihjakhfudhsknmzvhjkgfsafyeo89320134745660174812k4h3nfyhauifyhjzvczx042314FDUVNMCMXHJSQOQWHJKDSAJDKJSAKdjhdy7qeiwoqyUMNVCNZMCIIDOQIHEWJQKEWDFFSAFDSAFDASmaoqowuerhfdgxvdftq4w126435769h0y0hpnlhkmjnyuisgzfctedkapqouirhnvnzmxncudif89382910eud5d5dsea3sfxgc7v8d0s0sxzk0a0s98df6f5dkdas09dlzKJdsai0aSDFFa
```

ibuenros:
It is not a bug, see:
https://issues.apache.org/jira/browse/HDFS-6055

Apparently it was introduced in 2.4 for compatibility with local file systems.

ibuenros:
@sahilTakiar any more comments here?

abti:
@ibuenros Please fix the merge conflicts

PR #854, opened 2016-03-15
Fixing coveralls report location

sahilTakiar:


sahilTakiar:
Trivial change self-approving

PR #855, opened 2016-03-15
Hack to coveralls project to apply Java plugin to all Gradle subprojects

sahilTakiar:


sahilTakiar:
self-approving to get coveralls to work

PR #856, opened 2016-03-16
Clean up KafkaSource

zliu41:


gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -526,37 +529,31 @@ public boolean apply(KafkaTopic input) {
...
-        long curTime = System.currentTimeMillis();
-        LOG.info("Using thread pool to create work units for topic " + this.topic.getName());
-        List<WorkUnit> oneTopicWorkUnits =
-            KafkaSource.this.getWorkUnitsForTopic(this.topic, this.state, topicSpecificState);
-        this.allTopicWorkUnits.put(this.topic.getName(), oneTopicWorkUnits);
-        LOG.info("Done creating work unit for " + this.topic.getName() + ", created work unit size is "
-            + oneTopicWorkUnits.size() + ". Total time used for creating this topic is "
-            + (System.currentTimeMillis() - curTime) + " milliseconds");
+        this.allTopicWorkUnits.put(this.topic.getName(),
+            KafkaSource.this.getWorkUnitsForTopic(this.topic, this.state, topicSpecificState));
```
ydai1124: The topic name seems missing from log. Is it printed out anywhere else?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -526,37 +529,31 @@ public boolean apply(KafkaTopic input) {
...
       } catch (Throwable t) {
```
ydai1124: Is it better to catch Exception instead of Throwable here?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
zliu41: The log is removed here. There will be log for creating workunits for each partition, so these two log messages are not really useful.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
zliu41: Since t will be re-thrown (rather than swallowed), it shouldn't matter.

ydai1124:
LGTM

PR #859, opened 2016-03-16
Null Configuration is passed to MRJobLauncher

lbendig:
It seems to me that due to a regression of https://github.com/linkedin/gobblin/pull/808 the MRJobLauncher receives a null Configuration object which results in an NPE at `JobConfigurationUtils#putPropertiesIntoConfiguration`.

The instantiation of `MRJobLauncher` was moved from `run()` to the `CliMRJobLauncher`'s constructor, but here `getConf()` is null as `org.apache.hadoop.conf.Configured()` calls to `setConf(null)`. `org.apache.hadoop.util.ToolRunner#run` would set it, but it's too late as MRJobLauncher is instantiated
earlier.


sahilTakiar:
LGTM, thanks for finding this bug and fixing it

lbendig:
Thanks for meging it.

PR #860, opened 2016-03-16
Bump dependencies in gobblin-mapreduce

lbendig:
Also added log4j-extras jar to the distribution, as client side logging uses `RollingFileAppender`.


sahilTakiar:
LGTM thanks for the fix

lbendig:
Thank you, Sahil!

2016-03-17 2:17 GMT+01:00 Sahil Takiar <notifications@github.com>:

PR #863, opened 2016-03-17
Metric reporters (Graphite, InfluxDB)

lbendig:
Re-implementation of the Graphite and InfluxDB reporters that were removed in https://github.com/linkedin/gobblin/pull/429 due to the changes in Metric API.

Graphite:
- Supports sending through UDP and TCP 
- Metric and/or event reporter can be toggled
- Additional parameter `metrics.reporting.graphite.events.value.as.key`: 
  - For those backends that don't allow strings as event values (e.g: InfluxDB's Graphite protocol)
    In this case the event value will be added to the end of the key

InfluxDB:
- Metric and/or event reporter can be toggled
- Only TCP as influxdb-java 2.1 doesn't support UDP yet

Both were tested on InfluxDB 0.9.4.1.

Refactoring:
- Restructured configuration keys for the metric reporters and classes of metric/event names
- Timing event related event names are now in the `TimingEvent` class
- Metric reporting configurations are moved from `MetricReportReport` to a new base class `ConfiguredScheduledReporter`.
  To report non-serialized metrics one can extend `ConfiguredScheduledReporter`, for serialized Avro records `MetricReportReport` can be subclassed
- Duplicate class: MetricGroup  -> gobblin.runtime.MetricGroup was removed


ibuenros:
Thanks @lbendig . It is a large PR, I'll try to look at it over the weekend.

lbendig:
Thanks @ibuenros !

gobblin-metrics/src/main/java/gobblin/metrics/event/MultiPartEvent.java
```diff
@@ -0,0 +1,87 @@
...
+  TASKSTATE_EVENT(
+      TASK_STATE,
+      METADATA_TASK_START_TIME,
+      METADATA_TASK_END_TIME,
+      METADATA_TASK_WORKING_STATE);
+
+  private String eventName;
+  private String[] metadataFields;
+  
+  private static final Map<String, MultiPartEvent> lookup = Maps.newHashMap();
```
ibuenros: Can you make this an `ImmutableMap` instead? Helps avoid memory leaks.

lbendig: Fixed.

lbendig:
@ibuenros pls just let me know if you need anything regarding the review of this PR.

lbendig:
@ibuenros what do you think, would this PR be a good candidate to add it to Gobblin?

ibuenros:
Sorry about the delay (I know it's ridiculous). I promise to look at it by end of weekend.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -548,6 +557,155 @@ private void buildKafkaMetricReporter(Properties properties) {
...
 
+  private void buildGraphiteMetricReporter(Properties properties) {
+    boolean metricsEnabled =
+        Utils.getPropAsBoolean(properties, ConfigurationKeys.METRICS_REPORTING_GRAPHITE_METRICS_ENABLED_KEY,
+            ConfigurationKeys.DEFAULT_METRICS_REPORTING_GRAPHITE_METRICS_ENABLED);
+    boolean eventsEnabled =
+        Utils.getPropAsBoolean(properties, ConfigurationKeys.METRICS_REPORTING_GRAPHITE_EVENTS_ENABLED_KEY,
+            ConfigurationKeys.DEFAULT_METRICS_REPORTING_GRAPHITE_EVENTS_ENABLED);
+    if (!metricsEnabled && !eventsEnabled) {
+      LOGGER.info("Not reporting metrics and events to Graphite");
```
ibuenros: I know we do it this way right now, but I would prefer instead logging the positive case. E.g. if the graphite metrics are enabled, log "Reporting metrics... Graphite"

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -548,6 +557,155 @@ private void buildKafkaMetricReporter(Properties properties) {
...
+      return;
+    }
+    
+    try {
+      Preconditions.checkArgument(properties.containsKey(ConfigurationKeys.METRICS_REPORTING_GRAPHITE_HOSTNAME),
```
ibuenros: I think we should fail if something critical is missing. Better than not getting metrics.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -548,6 +557,155 @@ private void buildKafkaMetricReporter(Properties properties) {
...
+      Preconditions.checkArgument(properties.containsKey(ConfigurationKeys.METRICS_REPORTING_INFLUXDB_DATABASE),
+          "InfluxDB database name is missing.");
+    } catch (IllegalArgumentException exception) {
+      LOGGER.error("Not reporting to InfluxDB due to missing InfluxDB configuration(s).", exception);
+      return;
+    }
+    
+    String url = properties.getProperty(ConfigurationKeys.METRICS_REPORTING_INFLUXDB_URL);
+    String username = properties.getProperty(ConfigurationKeys.METRICS_REPORTING_INFLUXDB_USER);
+    String password = properties.getProperty(ConfigurationKeys.METRICS_REPORTING_INFLUXDB_PASSWORD);
```
ibuenros: Not necessary, but would it be good to allow encryption of this password? (There is a password manager module in Gobblin).

gobblin-metrics/src/main/java/gobblin/metrics/graphite/GraphiteEventReporter.java
```diff
@@ -0,0 +1,231 @@
...
+    this.emitValueAsKey = builder.emitValueAsKey;
+  }
+
+  @Override
+  public void reportEventQueue(Queue<GobblinTrackingEvent> queue) {
+
+    GobblinTrackingEvent nextEvent;
+    try {
+      while (null != (nextEvent = queue.poll())) {
+        pushEvent(nextEvent);
```
ibuenros: can `pushEvent` fail on certain event, failing to flush previous events?

ibuenros:
It looks pretty good in general. A few comments on the Graphite reporter, which also apply to the corresponding methods for InfluxDB.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
lbendig: Fixed. I also changed the way of logging (positive case) for the other reporters.

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
lbendig: I followed the pattern of the other reporters: catch exception, log it and return. Should any issue in the metric reporting bring down the job?

gobblin-core/src/main/java/gobblin/metrics/GobblinMetrics.java
lbendig: Good idea, fixed.

gobblin-metrics/src/main/java/gobblin/metrics/graphite/GraphiteEventReporter.java
lbendig: I think this can be solved by flushing the events in the catch block.

lbendig:
@ibuenros  Thank you for the review! I addressed your comments and also made changes in the metric key naming for the Influx and Graphite reporters as I found that it was impacted by some commits since the initial pull request. Also downgraded influxdb-java to 2.1 because the api went through some major changes in 2.2.

ibuenros:
Merging. Thanks @lbendig .

lbendig:
Thank you @ibuenros for the review/merge!

PR #867, opened 2016-03-18
Making HiveRegistrationPolicyBase constructor public

sahilTakiar:
* Making the constructor for `HiveRegistrationPolicyBase` public so that it can be initialized in other classes
* Otherwise `HiveRegistrationPolicyBase.getPolicy(state)` throws `NoSuchMethodException: No such accessible constructor on object`


sahilTakiar:
@zliu41 can you review?

zliu41:
LGTM
This PR increases test coverage? what??

sahilTakiar:
that is very weird, not sure why thats happening

PR #874, opened 2016-03-23
Minor revision to create table/add partition logic in HiveMetaStoreBasedRegister

zliu41:
The purpose is to resolve conflict between two Hive registration jobs running simultaneously.

The original logic may cause create table/add partition failure (`MySQLIntegrityConstraintViolationException` - primary key already exists in the metastore table) even if `client.tableExists` returns false.

The new logic first tries to create table/add partition, if unsuccessful, will try to alter table if needed, and throws `IOException` if unsuccessful again.


sahilTakiar:
LGTM

Issue #861, opened 2016-03-17
ExecutorUtils.newThreadFactory should create Error handling Threads


sahilTakiar:
A vanilla Java `Thread` that executes a `Runnable` does not handle `RuntimeException`s thrown by the `Runnable`. We have added `uncaughtExceptionHandler`s to ensure that any `RuntimeException`s are at least logged. This logic is encapsulates in `ExecutorUtils.newThreadFactory`.

However, `uncaughtExceptionHandler`s do not handle `Error`s (e.g. `OutOfMemoryError` or `NoClassDefFoundError`). It simply swallows them and does not log them.

The easiest way to avoid this it to always use `Callable` instead of `Runnable`, as `Callable` handles this; however, legacy projects may be stuck using `Runnable`s. 

One possible solution is to add instrumentation to `ExecutorUtils.newThreadFactory` to produce `Thread`s that catch any `Error`s and throw them as `RuntimeException`s. This is exactly what `Callable` does - if a `Error` is thrown in a `Callable` it gets wrapped in an `ExecutionException`.

sahilTakiar:
Closed, detailed explanation in https://github.com/linkedin/gobblin/pull/875

PR #875, opened 2016-03-23
(#861) ExecutorUtils.newThreadFactory should create Error handling Threads

sahilTakiar:
* Fix for #861 (contains a detailed explanation of the problem)
* `ExecutorUtils.newThreadFactory` now uses a `ErrorHandlingThreadFactory` which creates `ErrorHandlingThread`s
* A `ErrorHandlingThread` wraps a given `Thread`, executes its `run()` method, catches any `Error`s thrown, and converts them to `RuntimeException`s

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ExecutorsUtils.java
@@ -24,12 +24,15 @@ import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+import lombok.AllArgsConstructor;
 
 
 /**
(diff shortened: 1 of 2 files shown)


sahilTakiar:
@zliu41 can you review?

zliu41:
I didn't quite understand the problem you described in #861.

My understanding is that if the user submits either a Runnable or Callable via `ThreadPoolExecutor.submit`, then `UncaughtExceptionHandler` is irrelevant and won't be invoked. Any `Throwable` thrown by the Runnable/Callable will be wrapped into `ExecutionException` in the `Future` returned by `submit`.

If the user submits a Runnable via `ThreadPoolExecutor.execute`, then any `Throwable` thrown by the Runnable can be retrieved in `afterExecute`. 

So I'm not sure why `ErrorHandlingThreadFactory` is needed. In fact if I understand correctly, `UncaughtExceptionHandler` is only useful if you run a thread manually (not via `ExecutorService` so I'm not even sure whether we need `UncaughtExceptionHandler` in the first place).

sahilTakiar:
Yes, I see you are correct. My understanding was the `UncaughtExceptionHandler` would catch any `RuntimeException`s thrown by a `Thread` that is run in an `ExecutorService`. However, testing this locally using the below code indicates otherwise:

```
public static void main(String args[]) throws ExecutionException, InterruptedException {
  final Logger LOG = LoggerFactory.getLogger(ExecutorsUtils.class);

  ExecutorService service = Executors.newSingleThreadExecutor(ExecutorsUtils.newThreadFactory(Optional.of(LOG)));
  service.submit(new Runnable() {
    @Override public void run() {
      LOG.info("Hello World!");
      throw new RuntimeException("Throwing Exception!");
    }
  });
  ExecutorsUtils.shutdownExecutorService(service, Optional.<Logger>absent());
}
```

Output:

```
0 [pool-2-thread-1] INFO gobblin.util.ExecutorsUtils  - Hello World!

Process finished with exit code 0
```

This seems to indicate that the `UncaughtExceptionHandler` is never even invoked in the above code. Interestingly, if `ExecutorService.execute(Runnable)` is used instead of `ExecutorService.submit(Runnable)` then the `UncaughtExceptionHandler` is invoked, and the exception is logged. These stack overflow posts confirms the above findings: http://stackoverflow.com/questions/3929342/choose-between-executorservices-submit-and-executorservices-execute and http://stackoverflow.com/questions/1838923/why-is-uncaughtexceptionhandler-not-called-by-executorservice

My guess is that since `ExecutorService.submit(Runnable)` returns a `Future` the recommended way of handling `Exception`s is via `Future.get()`. Since `ExecutorService.execute(Runnable)` returns `void` it has to rely on the `UncaughtExceptionHandler`. I also confirmed that when invoked, the `UncaughtExceptionHandler` handles `RuntimeException`s and `Error`s properly.

This is problematic as we have followed the above pattern in many projects with the goal of handling any uncaught exceptions (ref internal JIRAs ETL-2559 and ETL-2560). For example, in Lumos we use the above pattern when running `CommitRunner`s in the `VirtualSnapshotJob`. Thats a big problem because we could have uncaught exceptions being thrown that we would never know about.

That being said, I will close this PR and create some internal JIRAs. Ideally, we would use `ExecutorService.submit(Runnable)` and use the returned `Future` object to handle any exceptions, but that is a more invasive change. The easiest fix is to just use `ExecutorService.execute(Runnable)` instead of `ExecutorService.submit(Runnable)`.

PR #876, opened 2016-03-23
New file sets in distcp will only be generated if there is space in t…

ibuenros:
This PR allows `CopyableDatasets` to instead implement `IterableCopyableDataset`, which returns an iterator of `FileSet`s instead of the entire list of `CopyEntities`. The file sets could be created only on demand, avoiding expensive operations that will be unused because the work unit list is full.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
@@ -42,8 +42,6 @@ import gobblin.util.iterators.InterruptibleIterator;
 
 import java.io.IOException;
 import java.net.URI;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -61,10 +59,8 @@ import org.apache.hadoop.fs.FileSystem;
 
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
 
 
 /**
@@ -119,14 +115,14 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
 
       final CopyConfiguration copyConfiguration = CopyConfiguration.builder(targetFs, state.getProperties()).build();
 
-      DatasetsFinder<CopyableDataset> datasetFinder =
+      DatasetsFinder<CopyableDatasetBase> datasetFinder =
           DatasetUtils.instantiateDatasetFinder(state.getProperties(), sourceFs, DEFAULT_DATASET_PROFILE_CLASS_KEY);
 
-      IterableDatasetFinder<CopyableDataset> iterableDatasetFinder =
-          datasetFinder instanceof IterableDatasetFinder ? (IterableDatasetFinder<CopyableDataset>) datasetFinder
+      IterableDatasetFinder<CopyableDatasetBase> iterableDatasetFinder =
+          datasetFinder instanceof IterableDatasetFinder ? (IterableDatasetFinder<CopyableDatasetBase>) datasetFinder
               : new IterableDatasetFinderImpl<>(datasetFinder);
 
-      Iterator<CopyableDataset> copyableDatasets = new InterruptibleIterator<>(iterableDatasetFinder.getDatasetsIterator(),
+      Iterator<CopyableDatasetBase> copyableDatasets = new InterruptibleIterator<>(iterableDatasetFinder.getDatasetsIterator(),
           new Callable<Boolean>() {
         @Override
         public Boolean call()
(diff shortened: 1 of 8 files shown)


ibuenros:
@zliu41 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -200,13 +210,11 @@ private FileSystem getOptionallyThrottledFileSystem(FileSystem fs, State state)
...
-        Collection<? extends CopyEntity> files = this.copyableDataset.getCopyableFiles(this.targetFs, this.copyConfiguration);
-        List<FileSet<CopyEntity>> fileSets = partitionCopyableFiles(this.copyableDataset, files);
-
-        // Sort to optimize the insertion to work units list
-        Collections.sort(fileSets, this.workUnitList.getComparator());
+        Iterator<FileSet<CopyEntity>> fileSets =
+            this.copyableDataset.getFileSetIterator(this.targetFs, this.copyConfiguration);
 
-        for (FileSet<CopyEntity> fileSet : fileSets) {
+        while (fileSets.hasNext() && !this.workUnitList.hasRejectedFileSet()) {
```
zliu41: * Does `this.workUnitList.hasRejectedFileSet()` return true once a fileset is either rejected or evicted, or only if a fileset is rejected? If former is true, probably should rename to `hasRejectedOrEvictedFileSet`. If latter is true, why differentiating rejected vs evicted?
* If `this.workUnitList.hasRejectedFileSet()` returns false here, but becomes true before or while you create the workunit, is this a problem?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetBase.java
```diff
@@ -0,0 +1,11 @@
+package gobblin.data.management.copy;
+
+import gobblin.dataset.Dataset;
+
+
+/**
+ * A common superinterface for {@link Dataset}s that can be operated on by distcp.
+ * Concrete classes must implement a subinterface of this interface ({@link CopyableDataset} or {@link IterableCopyableDataset}).
+ */
+interface CopyableDatasetBase extends Dataset {
```
zliu41: What's the point of this interface? Do you intend to add methods to it later?

gobblin-data-management/src/main/java/gobblin/data/management/copy/IterableCopyableDataset.java
```diff
@@ -0,0 +1,27 @@
...
+public interface IterableCopyableDataset extends CopyableDatasetBase {
+
+  /**
+   * Get an iterator of {@link FileSet}s of {@link CopyEntity}, each one representing a group of files to copy and
+   * associated actions.
+   * @param targetFs target {@link org.apache.hadoop.fs.FileSystem} where copied files will be placed.
+   * @param configuration {@link gobblin.data.management.copy.CopyConfiguration} for this job. See {@link gobblin.data.management.copy.CopyConfiguration}.
+   * @throws IOException
+   */
+  public Iterator<FileSet<CopyEntity>> getFileSetIterator(FileSystem targetFs, CopyConfiguration configuration)
```
zliu41: Is there a reason to not have `sourceFs` in the signature? Does this only support push mode?

gobblin-data-management/src/main/java/gobblin/data/management/copy/IterableCopyableDatasetImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private List<FileSet<CopyEntity>> partitionCopyableFiles(Dataset dataset, Collection<? extends CopyEntity> files) {
+    Map<String, FileSet.Builder<CopyEntity>> partitionBuildersMaps = Maps.newHashMap();
+    for (CopyEntity file : files) {
+      if (!partitionBuildersMaps.containsKey(file.getFileSet())) {
+        partitionBuildersMaps.put(file.getFileSet(), new FileSet.Builder<>(file.getFileSet(), dataset));
+      }
+      partitionBuildersMaps.get(file.getFileSet()).add(file);
+    }
+    return Lists.newArrayList(Iterables.transform(partitionBuildersMaps.values(),
```
zliu41: You should be able to just return `Iterables.transform` rather than creating a new ArrayList out of it.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
ibuenros: * `hasRejectedFileSet` only returns `true` with an actual rejection (false even if work units have been evicted). It is just a way of letting caller know whether it is likely that the new work units will also be rejected. There could be a different method to check if anything has been evicted.
* There is no issue if it returns false and then it becomes true after the work units are created. The new work units will be rejected or accepted in the same way. This is just a really dump heuristic to know when to stop generating work units.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetBase.java
ibuenros: Just to have a superinterface for both `CopyableDataset` and `IterableCopyableDataset`. The alternative is to make `IterableCopyableDataset` to extend `CopyableDataset`, forcing users to implement the method `getCopyableFiles` which will never be used. I don't mind either way.

gobblin-data-management/src/main/java/gobblin/data/management/copy/IterableCopyableDataset.java
ibuenros: The source `FileSystem` is available on the constructor of the `Dataset`. We do support push and pull (and even running distcp on a 3rd, unrelated `FileSystem`).

gobblin-data-management/src/main/java/gobblin/data/management/copy/IterableCopyableDatasetImpl.java
ibuenros: True, I just copied this method from somewhere else.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetBase.java
zliu41: What's wrong with having both of them just extend `Dataset`?

ibuenros: Type awareness. `CopySource` gets datasets of type `CopyableDataset` or `IterableCopyableDataset`, it doesn't care about any other type of dataset.

zliu41:
LGTM

zliu41:
Copyright header needs to be added to new files.

PR #877, opened 2016-03-24
Stream datasets to distcp.

ibuenros:
This PR enables streaming of datasets to distcp using an iterator. It implements an `IterableDatasetFinder` that returns datasets as an iterator. Distcp only consumes as many datasets at a time as copy listing threads are available.
This allows controlling memory footprint (ie datasets are created on demand when needed, whereas before, all datasets needed to exist in memory at the same time).

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
+++ b/gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util.executors;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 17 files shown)


ibuenros:
@ydai1124 can you review?

gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+    this.executor = Executors.newFixedThreadPool(numThreads, threadFactory);
+    this.completionService = new ExecutorCompletionService<>(this.executor);
+  }
+
+  /**
+   * Execute the tasks in the task {@link Iterator}.
+   * @throws InterruptedException
+   */
+  public void execute() throws InterruptedException {
+    List<Future<T>> futures = Lists.newArrayList();
```
ydai1124: Is the `futures` ever checked later?

gobblin-utility/src/main/java/gobblin/util/iterators/InterruptibleIterator.java
```diff
@@ -0,0 +1,62 @@
...
+        return true;
+      }
+      return false;
+    } catch (Exception exception) {
+      throw new RuntimeException(exception);
+    }
+  }
+
+  @Override
+  public T next() {
```
ydai1124: Why do you need `promisedNext`?
If a user calls `hasNext()` which returns true. Then, `interrupt` returns true. Another `next()` is called and `T` is still returned. 
This could be a problem if multiple iterators are explicitly calling `hasNext` first and then 'next()' to add elements to a bounded queue, when the first element gets rejected, the rest should also be rejected, which means the rest `next()` should simply be interrupted.

ibuenros: the idea is that a `true` response to `hasNext` guarantees `next` returns something (contract of `Iterator`). You are right that this guarantee is invalid with multiple threads, but iterators are not thread-safe and the same would be true for other iterators.

gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
ibuenros: Fixed. It is returned now.

ydai1124:
LGTM

PR #878, opened 2016-03-24
Distcp on Hive supports predicates for fast partition skips, and supp…

ibuenros:
…orts copying full directories recursively.

* Allow skipping partitions for Hive copy based on a predicate.
* Allow listing files on partition recursively from base directory instead of using input format.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -12,22 +12,13 @@
 
 package gobblin.data.management.copy.hive;
 
-import com.google.common.collect.Iterators;
-import gobblin.data.management.partition.FileSet;
-import java.util.Iterator;
-import java.util.NoSuchElementException;
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.Getter;
-import lombok.Singular;
-import lombok.extern.slf4j.Slf4j;
-
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


ibuenros:
@tuGithub can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -207,6 +213,16 @@
       this.partitionFilter = Optional.fromNullable(this.dataset.properties.getProperty(COPY_PARTITIONS_FILTER_CONSTANT));
     }
 
+    try {
+      this.fastPartitionSkip = this.dataset.properties.containsKey(FAST_PARTITION_SKIP_PREDICATE) ? Optional.of(
+          GobblinConstructorUtils.invokeFirstConstructor(
+              (Class<Predicate<Partition>>) Class.forName(this.dataset.properties.getProperty(FAST_PARTITION_SKIP_PREDICATE)),
+              Lists.<Object>newArrayList(this), Lists.newArrayList()))
```
tuGithub: the current Predicates<Partition> take only one parameter in the constructor: the helper. Why in here pass in two lists?

gobblin-data-management/src/main/java/gobblin/data/management/copy/predicates/RootDirectoryModtimeSkipPredicate.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public boolean apply(@Nullable Partition input) {
+
+    if (input == null) {
+      return true;
+    }
+
+    try {
+
+      if (PathUtils.isGlob(input.getDataLocation())) {
```
ibuenros: not now, but I don't know if they will change that in the future in Hive. This way we protect ourselves from such a change.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
ibuenros: This will try two constructors, `<init>(HiveCopyEntityHelper)` and `<init>()`. In case someone uses a predicate that doesn't take arguments.

tuGithub:
LGTM

PR #884, opened 2016-03-25
Fix issue with replicating owner and permission of system directories…

ibuenros:
… in distcp.

Distcp was trying to set unnecessary permissions and groups the user was not allowed to set in the staging directory. The cause is that we replicate ancestor permissions, and previously it was replicating all ancestors up to root path. At publish time this is not an issue because paths like "/" and "/data" already exist and we don't overwrite them. However, in the staging directory we recreate this structure and attempt to set the same permissions, so we would create "/path/to/staging/directory/data" with group "hdfs", which user is not allowed to set.

This PR restricts the default permission replication to only directories that don't exist on the target, and for the RecursiveCopyableDataset it only sets permissions between the non-globbed input path and the target path.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
@@ -20,8 +20,6 @@ import gobblin.util.guid.Guid;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
 
 import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
@@ -35,6 +33,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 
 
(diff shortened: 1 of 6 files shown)


ibuenros:
@pcadabam can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -218,42 +219,87 @@ public CopyableFile build() throws IOException {
...
+      CopyConfiguration copyConfiguration) throws IOException {
+
+    PreserveAttributes preserve = copyConfiguration.getPreserve();
+    Optional<FileStatus> originFileStatus = copyConfiguration.getCopyContext().getFileStatus(fs, path);
+
+    if (!originFileStatus.isPresent()) {
+      throw new IOException(String.format("Origin path %s does not exist.", originFileStatus));
+    }
+
+    String group = null;
```
pcadabam: Is it ok if group remains null?

pcadabam:
Overall LGTM. Is there a way to test `resolveReplicatedOwnerAndPermission`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
ibuenros: a null group means it should remain unchanged (that's the Hadoop API).

ibuenros:
Added test for `resolveReplicatedOwnerAndPermission`

gobblin-data-management/src/test/java/gobblin/data/management/copy/CopyableFileTest.java
```diff
@@ -30,6 +32,7 @@
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.util.PathUtils;
 
+import static org.mockito.Mockito.*;
```
pcadabam: Please list out the imports instead *

pcadabam:
One minor comment. Fix and Merge

gobblin-data-management/src/test/java/gobblin/data/management/copy/CopyableFileTest.java
ibuenros: I was going for mockito's suggestion on usage:
http://mockito.org/
This is a test and a static import, so I am not sure it should be covered so strictly by Gobblin's import format. However, I am happy to switch it if needed.

PR #886, opened 2016-03-25
Purge distcp persist directory at the beginning of publish phase.

ibuenros:


ibuenros:
@abti can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/recovery/RecoveryHelper.java
```diff
@@ -119,6 +129,26 @@ public boolean persistFile(State state, CopyableFile file, Path path) throws IOE
...
+      log.info("No persist directory to clean.");
+      return;
+    }
+
+    long retentionMillis = TimeUnit.HOURS.toMillis(this.retentionHours);
+    long now = System.currentTimeMillis();
+
+    for (FileStatus fileStatus : this.fs.listStatus(this.persistDir.get())) {
+      if (now - fileStatus.getModificationTime() > retentionMillis) {
+        this.fs.delete(fileStatus.getPath(), true);
```
abti: minor: should we log if delete() returns false - could happen if somehow permissions mess up or if two executions have a run away condition (though it has a very small probability )

abti:
LGTM 
(Just had a minor comment )

PR #889, opened 2016-03-28
Avro schema modification in distcp is executed only for URLs in the o…

ibuenros:
…rigin schema and authority.


ibuenros:
@tuGithub can you review?

tuGithub:
LGTM

PR #892, opened 2016-03-29
Hive registration: use full path, make it possible to disable

zliu41:
- Use full path in table location
- Make it possible to disable Hive registration by specifying an empty `data.publisher.job.type`


zliu41:
@chavdar can you review?

PR #893, opened 2016-03-29
JDBC Writer

jinhyukchang:
Pull request for JDBC Writer. Will combine to one commit after the review.

Design doc: https://docs.google.com/a/linkedin.com/document/d/1pSWZtgNssxaV1c19xdd1rKft9JaXM2qmqDwwbIb0scM/edit?usp=sharing


kadaan:
Any reason the design doc can't be public?

chavdar:
@kadaan Internally, for reviews, we use a company google docs account. Once we finalize the internal review, we'll share the design.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -102,6 +102,7 @@
   public static final String JOB_JAR_FILES_KEY = "job.jars";
   public static final String JOB_LOCAL_FILES_KEY = "job.local.files";
   public static final String JOB_HDFS_FILES_KEY = "job.hdfs.files";
+  public static final String JOB_JAR_HDFS_FILES_KEY = "job.jars.hdfs";
```
chavdar: Please make name consistent with above property: job.hdfs.jars.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -296,6 +310,17 @@
   public static final boolean DEFAULT_PUBLISH_DATA_AT_JOB_LEVEL = true;
   public static final String PUBLISHER_DIRS = DATA_PUBLISHER_PREFIX + ".output.dirs";
 
+  public static final String JDBC_PUBLISHER_PREFIX = "jdbc.publisher.";
```
chavdar: Please move all JDBC_PUBLISHER properties to the JdbcPublisher class.

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * The implementation of JdbcBufferedInserter for MySQL.
+ * This purpose of buffered insert is mainly for performance reason and the implementation is based on the
+ * reference manual http://dev.mysql.com/doc/refman/5.0/en/insert-speed.html
+ */
+public class MySqlBufferedInserter implements JdbcBufferedInserter {
+  private static final Logger LOG = LoggerFactory.getLogger(MySqlBufferedInserter.class);
+
+  private static final String INSERT_STATEMENT_PREFIX_FORMAT = "INSERT INTO %s.%s (%s) VALUES ";
```
chavdar: Have you considered an SQL DSL library like jooq? It makes logic much more portable.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
jinhyukchang: Done.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
jinhyukchang: Done.

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
jinhyukchang: Synced offline in design meeting and will be researched in next iteration.

gobblin-core/src/main/java/gobblin/converter/filter/AvroFieldsPickConverter.java
```diff
@@ -0,0 +1,182 @@
...
+              ]
+            }
+          ]
+        }
+   * If user wants to only choose name and city, the input parameter should be "name,address.city". Note that it is not user.name as first record is depth zero.
+   * {@inheritDoc}
+   * @see gobblin.converter.AvroToAvroConverterBase#convertSchema(org.apache.avro.Schema, gobblin.configuration.WorkUnitState)
+   */
+  @Override
+  public Schema convertSchema(Schema inputSchema, WorkUnitState workUnit) throws SchemaConversionException {
```
sahilTakiar: Why does this method need to first populate a list of fields to remove, and then remove them remove them each from the schema? Can't it simply create a `Schema` based on the value of `CONVERTER_AVRO_FIELDS_PICK_FIELDS`.

`AvroUtils.getFieldSchema` will fetch the `Schema` of a field given a dot separated path to the field name.

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
```diff
@@ -0,0 +1,354 @@
...
+import gobblin.util.jdbc.DataSourceBuilder;
+import gobblin.writer.commands.JdbcWriterCommands;
+import gobblin.writer.commands.JdbcWriterCommandsFactory;
+
+/**
+ * Converts Avro Schema into JdbcEntrySchema
+ * Converts Avro GenericRecord into JdbcEntryData
+ * Converts Avro field name for JDBC counterpart.
+ *
+ * This converter is written based on Avro 1.8 specification http://avro.apache.org/docs/1.8.0/spec.html
```
sahilTakiar: Gobblin is still on Avro 1.7.7 and I believe Avro 1.8 has some backwards incompatible changes from 1.7.7 - I'm not sure what those changes are, but I would recommend sticking with 1.7.7

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
```diff
@@ -0,0 +1,354 @@
...
+   *  string --> java.lang.String
+   *  null: only allowed if it's within union (see complex types for more details)
+   * 4. Supported Avro complex types
+   *  Records: Only first level depth can have Records type. Basically converter will peel out Records type and start with 2nd level.
+   *  Enum --> java.lang.String
+   *  Unions --> Only allowed if it have one primitive type in it or null type with one primitive type where null will be ignored.
+   *  Once Union is narrowed down to one primitive type, it will follow conversion of primitive type above.
+   * {@inheritDoc}
+   *
+   * 5. In order to make conversion from Avro long type to java.sql.Date or java.sql.Time or java.sql.Timestamp,
```
sahilTakiar: I assume this metadata doesn't change between WorkUnits, so can it just be determined once in the driver and stored somewhere (e.g. the WorkUnit). This avoids having some to make an extra JDBC connection in the `Converter`.

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
```diff
@@ -0,0 +1,84 @@
...
+    return byteSize;
+  }
+
+  /**
+   * Note that it only uses column name as a key
+   * {@inheritDoc}
+   * @see java.lang.Object#hashCode()
+   */
+  @Override
+  public int hashCode() {
```
sahilTakiar: I suggest using Lomboks `@EqualsAndHashCode` annotation to auto-generate these

https://projectlombok.org/features/EqualsAndHashCode.html

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
```diff
@@ -0,0 +1,84 @@
...
+    if (columnName == null) {
+      if (other.columnName != null)
+        return false;
+    } else if (!columnName.equals(other.columnName))
+      return false;
+    return true;
+  }
+
+  @Override
+  public String toString() {
```
sahilTakiar: Also suggest using lombok to auto-generate this method: https://projectlombok.org/features/ToString.html

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
```diff
@@ -0,0 +1,194 @@
...
+import com.github.rholder.retry.WaitStrategies;
+import com.google.common.base.Joiner;
+import com.google.common.collect.Lists;
+
+/**
+ * The implementation of JdbcBufferedInserter for MySQL.
+ * This purpose of buffered insert is mainly for performance reason and the implementation is based on the
+ * reference manual http://dev.mysql.com/doc/refman/5.0/en/insert-speed.html
+ */
+public class MySqlBufferedInserter implements JdbcBufferedInserter {
```
sahilTakiar: Is this using batches to do the buffered insert? If so does it have to be MySQL specific. Can't it be used for all JDBC connections using `PreparedStatement.addBatch()`

gobblin-core/src/main/java/gobblin/writer/JdbcWriterBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+package gobblin.writer;
+
+import gobblin.converter.jdbc.JdbcEntryData;
+import gobblin.converter.jdbc.JdbcEntrySchema;
+
+import java.io.IOException;
+
+import org.apache.avro.Schema;
+
+public class JdbcWriterBuilder extends PartitionAwareDataWriterBuilder<JdbcEntrySchema, JdbcEntryData> {
```
sahilTakiar: I suggest just extending `DataWriterBuilder` instead, we can add the partitioning integration in another PR, as it doesn't seem to be fully done.

gobblin-core/src/main/java/gobblin/Initializer.java
```diff
@@ -0,0 +1,34 @@
...
+   */
+  public void initialize();
+
+  /**
+   * Removed checked exception.
+   * {@inheritDoc}
+   * @see java.io.Closeable#close()
+   */
+  @Override
+  public void close();
```
sahilTakiar: I don't think you need to explicitly override this method

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -235,6 +242,13 @@
   public static final String DEFAULT_WRITER_FILE_PATH_TYPE = "default";
   public static final String SIMPLE_WRITER_DELIMITER = "simple.writer.delimiter";
   public static final String SIMPLE_WRITER_PREPEND_SIZE = "simple.writer.prepend.size";
+  public static final String WRITER_JDBC_INSERT_BATCH_SIZE = WRITER_PREFIX + ".jdbc.batch.size";
+  public static final int DEFAULT_WRITER_JDBC_INSERT_BATCH_SIZE = 30;
+  public static final String WRITER_JDBC_INSERT_BUFFER_SIZE = WRITER_PREFIX + ".jdbc.insert.buffer.size";
+  public static final int DEFAULT_WRITER_JDBC_INSERT_BUFFER_SIZE = 1024 * 1024; //1 MBytes
+  public static final int MAX_WRITER_JDBC_INSERT_BUFFER_SIZE = 10 * 1024 * 1024; //10 MBytes
+  public static final String WRITER_JDBC_MAX_PARAM_SIZE = WRITER_PREFIX + ".jdbc.insert.max.param.size";
+  public static final int DEFAULT_WRITER_JDBC_MAX_PARAM_SIZE = 100000; //MySQL limit
```
sahilTakiar: Can you move all these configuration keys into the classes that use them.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -235,6 +242,13 @@
   public static final String DEFAULT_WRITER_FILE_PATH_TYPE = "default";
   public static final String SIMPLE_WRITER_DELIMITER = "simple.writer.delimiter";
   public static final String SIMPLE_WRITER_PREPEND_SIZE = "simple.writer.prepend.size";
+  public static final String WRITER_JDBC_INSERT_BATCH_SIZE = WRITER_PREFIX + ".jdbc.batch.size";
+  public static final int DEFAULT_WRITER_JDBC_INSERT_BATCH_SIZE = 30;
+  public static final String WRITER_JDBC_INSERT_BUFFER_SIZE = WRITER_PREFIX + ".jdbc.insert.buffer.size";
+  public static final int DEFAULT_WRITER_JDBC_INSERT_BUFFER_SIZE = 1024 * 1024; //1 MBytes
+  public static final int MAX_WRITER_JDBC_INSERT_BUFFER_SIZE = 10 * 1024 * 1024; //10 MBytes
+  public static final String WRITER_JDBC_MAX_PARAM_SIZE = WRITER_PREFIX + ".jdbc.insert.max.param.size";
```
sahilTakiar: We try to avoid use `.` as a a replacement for spaces. So this config key should actually be `writer.jdbc.max_param_insert_size` (or something along those lines)

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class JdbcEntryDatum {
+  private final String columnName;
+  private final Object val;
+  private int byteSize;
+
+  public JdbcEntryDatum(String columnName, Object val) {
+    this.columnName = Preconditions.checkNotNull(columnName);
+    this.val = Preconditions.checkNotNull(val);
+    this.byteSize = val.toString().getBytes().length;
```
sahilTakiar: Is this an accurate way of measuring the bytes of of each datum? How necessary is this? There is an associated cost with serializing the String into a byte array.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -389,6 +394,19 @@ private void addHDFSFiles(String jobFileList) throws IOException {
     }
   }
 
+  private void addHdfsJars(String hdfsJarFileList) throws IOException {
```
sahilTakiar: Why is this needed when `JOB_JAR_FILES_KEY` can be used to specify jar files? The only difference I see is that `JOB_JAR_FILES_KEY` specifies jars on the local filesystem.

gobblin-utility/src/main/java/gobblin/util/jdbc/DataSourceBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+    validateNotEmpty(url, "url");
+    validateNotEmpty(driver, "driver");
+    validateNotEmpty(passWord, "passWord");
+    validateTrue(maxIdleConnections == null || maxIdleConnections > 0, "maxIdleConnections should be a positive integer.");
+    validateTrue(maxActiveConnections == null || maxActiveConnections > 0, "maxActiveConnections should be a positive integer.");
+  }
+
+  private void validateNotEmpty(String s, String name) {
+    if(StringUtils.isEmpty(s)) {
+      throw new IllegalArgumentException(name + " should not be empty.");
```
sahilTakiar: You can use Guava's `Preconditions` to do this

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -249,6 +253,10 @@ public void apply(JobListener jobListener, JobContext jobContext) throws Excepti
             return;
           }
 
+          WriterInitializer writerInitializer = WriterInitializerFactory.newInstace(jobState, workUnits.get());
+          closer.register(writerInitializer);
```
sahilTakiar: You can combine the above two lines as `closer.register` returns the object it registers

gobblin-core/src/test/java/gobblin/writer/jdbc/JdbcWriterInitializerTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+import static org.mockito.Mockito.*;
+
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+
+import junit.framework.Assert;
```
sahilTakiar: Lets stick with using `testng` instead of `junit`

gobblin-core/src/main/java/gobblin/writer/commands/JdbcWriterCommands.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+package gobblin.writer.commands;
+
+import gobblin.converter.jdbc.JdbcType;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.Map;
+
+public interface JdbcWriterCommands extends JdbcBufferedInserter {
```
sahilTakiar: Can you document what this interface is for?

gobblin-core/src/main/java/gobblin/writer/commands/JdbcWriterCommands.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Creates table structure based on the table in parameter. Note that this won't guarantee copy exactly the same as
+   * original table such as restraints, foreign keys, sequences, indices, etc
+   * @param conn
+   * @param fromStructure
+   * @param targetTableName
+   * @throws SQLException
+   */
+  public void createTableStructure(Connection conn, String fromStructure, String targetTableName) throws SQLException;
```
sahilTakiar: All these methods seem to take in a `Connection` is it possible to pass in the `Connection` through the constructor?

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
sahilTakiar: Its not a blocker, but would definitely recommend looking into a library like jooq it should remove the need for having the `JdbcWriterCommands` interface

gobblin-core/src/main/java/gobblin/converter/filter/AvroFieldsPickConverter.java
jinhyukchang: It was just to reduce it to remove problem as removal tool is already there in Gobblin. It seems like it can be easily done by AvroUtils and will update it.

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
jinhyukchang: It seems primitive type and complex type have not changed and I will update the comment.

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
jinhyukchang: Good point. Seems like the use case fits into Initializer. I can have ConverterInitializer just as WriterInitializer. Thoughts?

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
jinhyukchang: Will do.

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
jinhyukchang: Will do.

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
jinhyukchang: The implementation is to use multi values in one insert statement as per MySQL recommendation. I've briefly tested out PreparedStatement.addBatch but it didn't show as performant as passing multi value within insert statement.

gobblin-core/src/main/java/gobblin/writer/JdbcWriterBuilder.java
jinhyukchang: Will do.

gobblin-core/src/main/java/gobblin/Initializer.java
jinhyukchang: This is to remove checked exception. This is more like my preference, as caller anyway force to catch and re-throw anyway. If it preferred to have checked exception, let me know and I will remove it.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
jinhyukchang: Will do.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
jinhyukchang: Will do.

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
jinhyukchang: The main purpose of this is to control buffer size of buffered insert and it's better than having batch size to limit buffer size as each row size is different.
This cannot be 100% accurate as each JVM has its own way to store String and I think it can be used as an measurement. (Performance can be benefit by not having this, but, for now, I don't see any performance issue with the performance test.)

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
jinhyukchang: This is more for Azkaban Gobblin job type as Gobblin has lots of dependency jars and each job execution will upload from local to hdfs which takes more than a minute. Having this already in HDFS dramatically reduces it.

gobblin-utility/src/main/java/gobblin/util/jdbc/DataSourceBuilder.java
jinhyukchang: Will do.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
jinhyukchang: Got it.

gobblin-core/src/test/java/gobblin/writer/jdbc/JdbcWriterInitializerTest.java
jinhyukchang: Will update.

gobblin-core/src/main/java/gobblin/writer/commands/JdbcWriterCommands.java
jinhyukchang: Will do. (It's basically interface to each RDBMS as each RDBMS may uses different SQL syntax)

gobblin-core/src/main/java/gobblin/writer/commands/JdbcWriterCommands.java
jinhyukchang: Good point and let me figure out.

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
jinhyukchang: Jooq definitely looks interesting, and I will come back to it on next iteration (possibly when to support another RDBMS.)

jinhyukchang:
Update the code per PR.

gobblin-core/src/main/java/gobblin/converter/filter/AvroFieldsPickConverter.java
jinhyukchang: I was exploring AvroUtils and was not able to find an easy way to reconstruct Schema mainly because getFieldSchema only provides leaf information which means anyway converter needs to traverse the Schema.
I've used Trie to present dependencies from input and reconstruct the Schema. (Just traversing Schema and add field when it matches was dropped once I learned Schema fields are immutable). Let me know what you think.

jinhyukchang:
Merged from origin to sync up to date.

gobblin-core/src/main/java/gobblin/converter/filter/AvroFieldsPickConverter.java
sahilTakiar: I see, yes its unfortunate that just having the leaf Schemas is not enough. A Trie is one way to do this, but I think any nested data format would work. For example, I think you could do this using JSON, which has the advantage that you can use standard libraries like GSON. However, your current Trie approach works too, so I will leave it up to you.

gobblin-core/src/main/java/gobblin/converter/initializer/AvroToJdbcEntryConverterInitializer.java
```diff
@@ -0,0 +1,99 @@
...
+import gobblin.converter.jdbc.AvroToJdbcEntryConverter;
+import gobblin.converter.jdbc.JdbcType;
+import gobblin.publisher.JdbcPublisher;
+import gobblin.source.workunit.WorkUnit;
+import gobblin.util.ForkOperatorUtils;
+import gobblin.util.jdbc.DataSourceBuilder;
+import gobblin.writer.commands.JdbcWriterCommands;
+import gobblin.writer.commands.JdbcWriterCommandsFactory;
+
+public class AvroToJdbcEntryConverterInitializer implements ConverterInitializer {
```
sahilTakiar: javadoc please

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializer.java
```diff
@@ -0,0 +1,7 @@
+package gobblin.converter.initializer;
+
+import gobblin.initializer.Initializer;
+
+public interface ConverterInitializer extends Initializer {
```
sahilTakiar: javadoc please

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final Splitter COMMA_SPLITTER = Splitter.on(',').omitEmptyStrings().omitEmptyStrings();
+
+  /**
+   * Provides WriterInitializer based on the writer. Mostly writer is decided by the Writer builder (and destination) that user passes.
+   * If there's more than one branch, it will instantiate same number of WriterInitializer instance as number of branches and combine it into MultiWriterInitializer.
+   *
+   * @param state
+   * @return WriterInitializer
+   */
+  public static ConverterInitializer newInstace(State state, Collection<WorkUnit> workUnits) {
```
sahilTakiar: typo, `newInstace` to `newInstance`

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
```diff
@@ -0,0 +1,347 @@
...
+    if(!mapping.isPresent()) {
+      return key;
+    }
+
+    String converted = mapping.get().get(key);
+    return converted != null ? converted : key;
+  }
+
+  @VisibleForTesting
+  public Connection createConnection(State state) throws SQLException {
```
sahilTakiar: is this still needed?

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
sahilTakiar: Is it necessary to control the size of the INSERT query based on the size? For now can't it just be based on the number on the number of values?

I agree its probably not a bottleneck now, but it may be a performance problem in the future. It may just be better to remove the buffer size restriction altogether unless there is a hard need for it. We can always add it back later when we have to time to think through how to implement it.

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
sahilTakiar: I see, but how about for other RDMS? Is there anyway there can be a default `JdbcBufferedInserter` that works for popular RDMS such as Postgres, Oracle, or SQLServer? Currently, it seems that the JDBCWriter won't work with any RDBS besides MySQL.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
sahilTakiar: I see, makes sense.

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+    if(converterClasses.isEmpty()) {
+      return NOOP;
+    }
+
+    List<ConverterInitializer> cis = Lists.newArrayList();
+    JdbcWriterCommandsFactory factory = new JdbcWriterCommandsFactory();
+    for (String converterClass : converterClasses) {
+      if (AvroToJdbcEntryConverter.class.getName().equals(converterClass)) {
+        cis.add(new AvroToJdbcEntryConverterInitializer(state, workUnits, factory, branches, branchId));
```
sahilTakiar: The problem with this class (as well as the `WriterInitializerFactory`) is that every `Converter` / `Writer` that wants to have a corresponding `Initializer` has to modify this method. An alternative approach would be to create the following interface:

```
public interface Initializable {
    public <? extends Initializer> getInitializer()
```

The `AvroToJdbcEntryConverter` could then implement `Initializable` and its `getInitializer` method could return `AvroToJdbcEntryConverterInitializer`.

A similar pattern could be used to `JDBCWriterBuilder`. Thus, enabling the code to derive the `Initializer` class from the `Converter` / `Writer` itself.

@chavdar what do you think?

gobblin-core/src/main/java/gobblin/writer/commands/MySqlBufferedInserter.java
jinhyukchang: It should be easy to build a default one. Will add one.

gobblin-core/src/main/java/gobblin/converter/jdbc/JdbcEntryDatum.java
jinhyukchang: Got it. Let me remove byte size related logic.

gobblin-core/src/main/java/gobblin/converter/initializer/AvroToJdbcEntryConverterInitializer.java
jinhyukchang: Will add.

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializer.java
jinhyukchang: Will add.

gobblin-core/src/main/java/gobblin/converter/jdbc/AvroToJdbcEntryConverter.java
jinhyukchang: Nope, will remove it.

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
jinhyukchang: I think Initializable is a good in a way that it keeps open/closed principle.

The downside of this approach I think is that the class that implements Initializable should always have default constructor to make getInitializer() available. Not having default constructor will face runtime exception and for some classes, and the instance is potentially in invalid state by having default constructor -- e.g: AvroToJdbcEntryConverterInitializer needs to remove final modifier on member variables in order to have default constructor.

In my opinion, it's a typical factory method pattern class where if statement should be straight forward.

@sahilTakiar Thoughts?

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
jinhyukchang: Done

jinhyukchang:
@sahilTakiar Just updated the code addressing PR. Would you take a look when you have a chance?

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
sahilTakiar: Sure, but this method need only be specified in the builder class, not the concrete class itself - e.g. in `JDBCDataWriter`, for `Converter`s Gobblin already enforces each `Converter` to have a default constructor.

chavdar: I like @sahilTakiar idea but we'll have to see how make it generic. I also don't like that workunits are being passed here. The converter should not care about the entire set of workunits. I don't want to block this commit. We'll have to come back to this later.

sahilTakiar:
hey @jinhyukchang I discussed with @chavdar and everything in this PR looks good! I can merge this PR, but I have one last request.

Can you add some documentation to the Gobblin Docs about this feature, you can probably just copy what you have in your Google Doc. If you have time its always nice to add documentation on configuration keys and documentation on the different plugin points. This is generally useful for any new Gobblin user who wants to write to a RDMS (I already know of one user waiting on this PR because they want to use this).

All Gobblin documentation is checked into GitHub under `gobblin-docs`. I suggest putting the documentation in a new file called `gobblin-docs/user-guide/Gobblin-JDBC-Writer.md`. The format of the files is all in Markdown. It's should be a relatively simple process, but if you have more questions this page has more details: http://gobblin.readthedocs.io/en/latest/developer-guide/Contributing/

chavdar:
@jinhyukchang we will feature the JDBC writer as one of the highlights for our next release (0.7.0) so it will be great to have some supporting documentation.

jinhyukchang:
@sahilTakiar , @chavdar ,

Great, thanks for the review and will add documentation shortly.

jinhyukchang:
@sahilTakiar , @chavdar 
Just added documentation. Let me know how it looks.

Seems now there's a conflict with master branch and I will resolve it once documentation is reviewed.

sahilTakiar:
Documentation LGTM!

When you rebase this PR, can you also squash all the commits together into one commit?

jinhyukchang:
Sure, will do that.

PR #894, opened 2016-03-29
Enable multiple db and table names in Hive registration

zliu41:
- Multiple db and table names can be specified for a path via `additional.hive.database.names` and `additional.hive.table.names`. The original properties (`hive.database.name`, etc.) still works for backward compatibility.

- `HiveRegistrationPolicyBase` now uses fully qualified path as table location.


zliu41:
@chavdar

zliu41:
@ibuenros

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -97,10 +109,40 @@ public HiveRegistrationPolicyBase(State props) {
...
+    List<String> databaseNames = Lists.newArrayList();
+
+    Optional<String> databaseName;
+    if ((databaseName = getDatabaseName(path)).isPresent()) {
+      databaseNames.add(databaseName.get());
+    }
+
+    if (!Strings.isNullOrEmpty(this.props.getProp(ADDITIONAL_HIVE_DATABASE_NAMES))) {
+      for (String additionalDbName : this.props.getPropAsList(ADDITIONAL_HIVE_DATABASE_NAMES)) {
+        databaseNames.add(this.dbNamePrefix + additionalDbName + this.dbNameSuffix);
```
ibuenros: All additional dbs and tables must have fixed names? Is it possible for users to give multiple tables as regexes?

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -141,33 +211,45 @@ protected String sanitizeAndValidateName(String name) {
...
-    table.setStorageProps(this.props.getStorageProps());
-    table.setSerDeProps(this.props.getSerdeProps());
-    table.setNumBuckets(-1);
-    table.setBucketColumns(Lists.<String> newArrayList());
-    table.setTableType(TableType.EXTERNAL_TABLE.toString());
-    return table;
+  protected List<HiveTable> getTables(Path path) throws IOException {
+    List<HiveTable> tables = Lists.newArrayList();
+
+    for (String databaseName : getDatabaseNames(path)) {
```
ibuenros: This will do a cross product of all dbs and tables, so registering only `dba.tablea`, `dbb.tableb` is impossible. A suggestion is for users to instead provide the additional registrations as complete names, where db and table are separated by a dot.

zliu41: That will be problematic if one wants to register to `dba.tablea` and `dbb.tablea`, where `tablea` is obtained via a regex.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
zliu41: I don't foresee a use case where one needs to use two or more regexes for db or table names. This class just provides a base implementation; someone who has a fancy use case can extend it.

ibuenros:
LGTM

PR #895, opened 2016-03-29
Distcp hive adds some parameters to Hive registration.

ibuenros:


ibuenros:
@zliu41 can you review? Thanks.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegistrationUnitComparator.java
```diff
@@ -134,6 +134,14 @@ public T compareIsStoredAsSubDirs() {
     return (T) this;
   }
 
+  @SuppressWarnings("unchecked")
+  public T compareParameters() {
+    if (!this.result) {
+      compare(Optional.fromNullable(this.existingUnit.getProps()), Optional.fromNullable(this.newUnit.getProps()));
```
zliu41: Is this method used? This may not work because when creating a table/partition in Hive, Hive may add some parameters. So if you create a table and then retrieves it and compares with itself, it will have more parameters and this will return true.

If this is needed then I'd suggest to return false if existing unit's parameters is a superset of that of the new unit. Also, it should compare `storageProps` and `serDeProps` in addition to `props`. It should also be added to `compareAll`.

ibuenros: Addressed

zliu41:
LGTM

ibuenros:
@chavdar can you review?

PR #898, opened 2016-03-29
Fixed unclosed executor service in IteratorExecutor.

ibuenros:


ibuenros:
Trivial bug fix. Self approving.

PR #903, opened 2016-03-31
The underlying Avro CodecFactory only matches lowercase codecs, so we…

kadaan:
… should make sure they are lowercase before trying to find one.


kadaan:
@sahilTakiar @zliu41 Can you please review this?

sahilTakiar:
LGTM

PR #907, opened 2016-04-01
add log message if store with version already exists

pcadabam:


pcadabam:
Merging trivial change.

PR #912, opened 2016-04-05
Use dfs.replication.max in job.state replication factor

chavdar:
This addresses an issue with https://github.com/linkedin/gobblin/pull/910 . If the replication factor > dfs.replication.max , then it won't be changed. This fix attempts to use that property when setting replication.

@ibuenros @ydai1124 Can you review?


ydai1124:
LGTM

ibuenros:
LGTM

PR #913, opened 2016-04-05
Job state stored in distributed cache in MR mode.

ibuenros:


ibuenros:
@chavdar @sahilTakiar can you review? This addresses https://github.com/linkedin/gobblin/issues/911

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -313,8 +314,10 @@ private void prepareHadoopJob(List<WorkUnit> workUnits) throws IOException {
 
     // Serialize source state to a file which will be picked up by the mappers
     Path jobStateFilePath = new Path(this.mrJobDir, JOB_STATE_FILE_NAME);
-    SerializationUtils.serializeState(this.fs, jobStateFilePath, this.jobContext.getJobState(), 100);
+    SerializationUtils.serializeState(this.fs, jobStateFilePath, this.jobContext.getJobState());
+    DistributedCache.addCacheFile(jobStateFilePath.toUri(), this.job.getConfiguration());
```
sahilTakiar: We should use `MRJobLauncher.addHDFSFiles` which already does this, it also sets up a symlink to the file, this way the additional property `JOB_STATE_DISTRIBUTED_CACHE_NAME` should not longer be necessary.

Using the symlink should also simply the code added to the `setup` method.

ibuenros:
@sahilTakiar I tried using `MRJobLauncher.addHDFSFiles` but I couldn't get it working. `DistributedCache.getLocalCacheFiles(context.getConfiguration())` throws an NPE, and trying to simply load `new File(name)` cannot find the file.
Any suggestions?

ibuenros:
Talked with @sahilTakiar offline. Merging.

PR #914, opened 2016-04-06
Use new version finder and selection policy for retention

pcadabam:
### Changes

This is part-1 for adding retention support for tracking data. The new policies for tracking retention will be in a different pr.

- Allow retention to use `gobblin.data.management.policy.VersionSelectionPolicy` and `gobblin.data.management.version.finder.VersionFinder`. I have deprecated the old retentionPolicy and versionFinder but the code is backward compatible for existing jobs using them
- Add support for retention jobs to specify multiple versions and policy combinations. This is useful for many datasets where each partition needs to have a different retention/selection policy. For instance tracking data, hourly has a different policy than daily. This class is called `MultiVersionCleanableDatasetBase`
- Fix config keys in all the new selection policies and version finders to be more meaningful and work well with typesafe. The policies and finders will have local key paths as opposed to global paths. For ex. in the `NewestKSelectionPolicy` the versions retained key will be `selection.newestK.versionsSelected` instead of `gobblin.retention.selection.newestK.versionsSelected` this will allow us to use the policy for jobs other than retention like distcp-ng. The class instantiating the policy is responsible for converting the global config paths to local paths. In the retention case, `ConfigurableCleanableDataset` passes `config.getConfig("gobblin.retention")` to the constructor of `NewestKSelectionPolicy`.

@ibuenros and @ydai1124 Please review.


gobblin-data-management/src/main/java/gobblin/data/management/policy/AbstractTimeBasedSelectionPolicy.java
```diff
@@ -16,31 +16,38 @@
...
- * An implementation of {@link VersionSelectionPolicy} which returns versions that are newer than a certain time.
+ * An implementation of {@link VersionSelectionPolicy} which returns versions that satisfy a {@link Predicate}
+ * implemented by subclasses {@link #getSelectionPredicate()} method
  */
-public class TimeBasedSelectionPolicy implements VersionSelectionPolicy<TimestampedDatasetVersion> {
+public abstract class AbstractTimeBasedSelectionPolicy implements VersionSelectionPolicy<TimestampedDatasetVersion> {
 
-  public static final String TIME_BASED_SELECTION_LOOK_BACK_TIME_KEY = "gobblin.time.based.selection.lookback.time";
-  public static final String DEFAULT_LOOK_BACK_TIME = "7d";
+  public static final String TIME_BASED_SELECTION_LOOK_BACK_TIME_KEY = "selection.timeBased.lookbackTime";
```
ibuenros: Is there some way of providing backwards compatibility?

gobblin-data-management/src/main/java/gobblin/data/management/policy/CombineSelectionPolicy.java
```diff
@@ -57,8 +59,8 @@
  */
 public class CombineSelectionPolicy implements VersionSelectionPolicy<FileSystemDatasetVersion> {
 
-  public static final String VERSION_SELECTION_POLICIES_PREFIX = "version.selection.policy.class.";
-  public static final String VERSION_SELECTION_COMBINE_OPERATION = "version.selection.combine.operation";
+  public static final String VERSION_SELECTION_POLICIES_PREFIX = "selection.combine.policy.class";
```
ibuenros: `class` -> `classes`

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
```diff
@@ -0,0 +1,294 @@
...
+
+    for (VersionFinderAndPolicy<T> versionFinderAndPolicy : getVersionFindersAndPolicies()) {
+      VersionSelectionPolicy<T> selectionPolicy = versionFinderAndPolicy.getVersionSelectionPolicy();
+      VersionFinder<? extends T> versionFinder = versionFinderAndPolicy.getVersionFinder();
+
+      if (!selectionPolicy.versionClass().isAssignableFrom(versionFinder.versionClass())) {
+        throw new IOException("Incompatible dataset version classes.");
+      }
+
+      this.log.info("Cleaning dataset " + this);
```
ydai1124: Can you add more information here, about using which versionFinder/versionSelectionPolicy?

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
```diff
@@ -0,0 +1,294 @@
...
+   */
+  @Override
+  public void clean() throws IOException {
+
+    if (this.isDatasetBlacklisted) {
+      log.info("Dataset blacklisted. Cleanup skipped for " + datasetRoot());
+      return;
+    }
+
+    for (VersionFinderAndPolicy<T> versionFinderAndPolicy : getVersionFindersAndPolicies()) {
```
ydai1124: What is the use case here? Is it possible that there is race condition among different policies, and the versions returned by different versionFinders have overlap, e.g., they contain some same files?

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/ConfigurableCleanableDataset.java
```diff
@@ -41,46 +44,44 @@
...
+public class ConfigurableCleanableDataset<T extends FileSystemDatasetVersion> extends
+    MultiVersionCleanableDatasetBase<T> {
 
-  public static final String CONFIGURATION_KEY_PREFIX = "gobblin.retention.";
+  public static final String RETENTION_CONFIGURATION_KEY = "gobblin.retention";
+  public static final String CONFIGURATION_KEY_PREFIX = RETENTION_CONFIGURATION_KEY + ".";
   public static final String VERSION_FINDER_CLASS_KEY = CONFIGURATION_KEY_PREFIX + "version.finder.class";
   public static final String RETENTION_POLICY_CLASS_KEY = CONFIGURATION_KEY_PREFIX + "retention.policy.class";
+  public static final String SELECTION_POLICY_CLASS_KEY = CONFIGURATION_KEY_PREFIX + "selection.policy.class";
+  public static final String DATASET_PARTITIONS_LIST_KEY = CONFIGURATION_KEY_PREFIX + "dataset.partitions";
```
ibuenros: Can you explain what this does in a comment?

pcadabam: Added some doc and examples

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
pcadabam: The use case is that a dataset may contain different kind of version. For instance a tracking dataset may contain hourly/daily/daily_deduped etc. The glob pattern within the tracking dataset will be different for hourly vs daily. Also the retention policy may be different for hourly compared to daily.
I do not expect an overlap of versions unless there is a bug in the job config.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
pcadabam: Done added the class names to log

gobblin-data-management/src/main/java/gobblin/data/management/policy/CombineSelectionPolicy.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/policy/AbstractTimeBasedSelectionPolicy.java
pcadabam: I think this property was incorrectly named. Since the selection policies are is new, I do not think we need to worry about this old key. If there are any distcp jobs using the old key I will change the job to use new key.

pcadabam:
@ibuenros and @ydai1124 addressed comment. Please have another look.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
ydai1124: Okay.

ydai1124:
LGTM. Please rebase before merging.

ibuenros:
LGTM

PR #921, opened 2016-04-11
Escaping colons in extract table names

chavdar:
Escaping colons in extract table names to avoid creating invalid HDFS paths (e.g. in gobblin.util.WriterUtils.getDefaultWriterFilePath)

@ibuenros Can you review?


ibuenros:
LGTM

PR #922, opened 2016-04-11
more code refactoring

zliu41:


chavdar:
What is the nature of these change?

zliu41:
Just to make the coding style more consistent, e.g., always access instance fields with `this.`, make a method `static` whenever possible, use `try-with-resource` whenever possible, remove unnecessary `else` block, remove unnecessary type argument for generics, etc.

PR #923, opened 2016-04-11
Simple refactoring of JobContext.commit()

chavdar:
Refactor JobContext.commit() in preparation for parallelization for the commit logic:

* Refactor the commit logic from the for loop into a separate method processDatasetCommit()
* Refactor the check for support for failed WorkUnit handling into a separte method checkForUnpublishedWUHandling .
* Move commitSequenceBuilder from a instance variable to a local variable in processDatasetCommit().

@zliu41 can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
@@ -378,8 +378,7 @@ public class JobContext {
     try (Closer closer = Closer.create()) {
       dataPublisherClass = getJobDataPublisherClass(datasetState)
           .or((Class<? extends DataPublisher>) Class.forName(ConfigurationKeys.DEFAULT_DATA_PUBLISHER_TYPE));
-      success = checkForUnpublishedWUHandling(datasetUrn, datasetState, success, dataPublisherClass,
-                                              closer);
+      success = checkForUnpublishedWUHandling(datasetUrn, datasetState,dataPublisherClass, closer);
     } catch (ReflectiveOperationException roe) {
       this.logger.warn("Unable to find publisher class: " + roe, roe);
       success = false;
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -350,77 +351,105 @@ void commit() throws IOException {
...
-          allDatasetsCommit = false;
-          if (UnpublishedHandling.class.isAssignableFrom(dataPublisherClass)) {
-            DataPublisher publisher = closer.register(DataPublisher.getInstance(dataPublisherClass, datasetState));
-            this.logger
-                .info(String.format("Calling publisher to handle unpublished work units for dataset %s of job %s.",
-                    datasetUrn, this.jobId));
-            ((UnpublishedHandling) publisher).handleUnpublishedWorkUnits(datasetState.getTaskStatesAsWorkUnitStates());
-          }
-          continue;
+      // NOTE: order here is important as we don't want to skip processDatasetCommit if there was
```
zliu41: If you use `&` then order doesn't matter. If use `&&` then order matters.
In fact you can simply use `&=`

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -350,77 +351,105 @@ void commit() throws IOException {
...
+      this.logger.warn(String.format("Not committing dataset %s of job %s with commit policy %s and state %s",
+          datasetUrn, this.jobId, this.jobCommitPolicy, datasetState.getState()));
+      if (UnpublishedHandling.class.isAssignableFrom(dataPublisherClass)) {
+        DataPublisher publisher = closer.register(DataPublisher.getInstance(dataPublisherClass, datasetState));
+        this.logger
+            .info(String.format("Calling publisher to handle unpublished work units for dataset %s of job %s.",
+                datasetUrn, this.jobId));
+        ((UnpublishedHandling) publisher).handleUnpublishedWorkUnits(datasetState.getTaskStatesAsWorkUnitStates());
+      }
+      success = false;
```
zliu41: why success = false here?

chavdar: Preserves existing logic. I'll choose better naming. The return result is if the job can be committed.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
chavdar: Good point. Removed the comment.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
chavdar: I've refactored it to move the canCommitDataset(datasetState) check back processDatasetCommit() as the return value here does not make sense.

zliu41:
LGTM

PR #924, opened 2016-04-12
use new DatasetVersion in CombinePolicy

pcadabam:
- Fix some generics which cause LiCleanableDataset to fail
- Move some code out of `src/test/java/gobblin/data/management/retention/integration/RetentionIntegrationTest.java
` to `gobblin-data-management/src/main/java/gobblin/util/test/RetentionTestHelper.java`

@ibuenros can you review?


ibuenros:
LGTM

PR #926, opened 2016-04-13
Make NewestKSelectionPolicy use Java Generics instead of FileSystemDatasetVersion

sahilTakiar:
* With this approach, one does not have to cast to `FileSystemDatasetVersion` when using `NewestKSelectionPolicy` - any class that extends `FileSystemDatasetVersion` (such as `TimestampedDatasetVersion`) can be used


ibuenros:
LGTM

pcadabam:
I think this should be any class that extends `DatasetVersion` instead of just `FileSystemDatasetVersion`. This selection policy does not use any filesystem specific logic to select version

sahilTakiar:
@pcadabam I agree, but currently `VersionSelectionPolicy` is tied to `FileSystemDatasetVersion`, we would have to change the interface. I think we should make the change, but its will be backwards-incompatible, so there may be cause for concern.

I'm going to merge this for now, lets address any changes to `VersionSelectionPolicy` in another PR.

PR #927, opened 2016-04-14
Oracle reader

jinhyukchang:
Adding OracleExtractor which extracts data from Oracle database. 
OracleExtractor inherits/reuses many codes from JdbcExtractor. JdbcExtractor retrieves data via getRecordSet, where watermark is put into where clause in SELECT statement. As SQL statements and metadata store is differ among RDMSs, Oracle extractor provides specific query statement for extracting metadata (in getSchemaMetadata), extracting data (in getDataMetadata), count (in getCountMetadata), and provides data type mapping for JdbcExtractor.

Passed integration test in local.


sahilTakiar:
Can you provide a little more context for this PR? Are the new classes a simple implementation of the `JDBCSource` and `JDBCExtractor` for Oracle? Does it provide any additional optimizations?

jinhyukchang:
Hey @sahilTakiar , this is a simple implementation of JDBCExtractor & JDBCSource for Oracle. No additional optimization was made.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,297 @@
...
+  @Override
+  public String constructSampleClause() {
+    return "";
+  }
+
+  @Override
+  public String getWatermarkSourceFormat(WatermarkType watermarkType) {
+    String columnFormat = null;
+    switch (watermarkType) {
+      case TIMESTAMP:
```
sahilTakiar: there should be a `break` clause for each `case`, otherwise FindBugs will throw a warning, which will cause the build to fail

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,297 @@
...
+
+  @Override
+  public Map<String, String> getDataTypeMap() {
+    return DATA_TYPE_MAP;
+  }
+
+  @Override
+  public Iterator<JsonElement> getRecordSetFromSourceApi(String schema, String entity, WorkUnit workUnit,
+      List<Predicate> predicateList) throws IOException {
+    return null;
```
sahilTakiar: why does this return `null`?

sahilTakiar:
Can a add a little more explanation in the PR description about how this works? What are the exact queries you are running for fetch the data, watermark, count, etc?

In general, this LGTM pending the few comments and documentation update.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
jinhyukchang: Done

jinhyukchang:
Added explanatory comment on OracleExtractor. Use all_tab_columns instead of user_tab_columns to fetch metadata. Take SID as input parameter.

@chavdar , Could you take a look (or delegate to others) when you have a chance?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
jinhyukchang: Added comment on it (No need to support specific retrieval as getRecordSet in JdbcExtractor is sufficient) and changed it to throw exception.

sahilTakiar:
@jinhyukchang it seems @bjvanov has also created an `OracleSource` and `OracleExtractor`. Can you take a look at https://github.com/linkedin/gobblin/pull/1074 and possibly work together to consolidate your efforts?

sahilTakiar:
This thread on the Google Groups list is also relevant: https://groups.google.com/d/topic/gobblin-users/ljf2FwwvMp4/discussion

abti:
Oracle connector has been added via a consolidated effort in another PR, so closing this.

PR #928, opened 2016-04-14
Substitution of system and env variable in config management

pcadabam:
### Changes
- Added support to resolve system and environment variables in config management. Both in `main.conf` and `includes.conf`
- `Typesafe Config` expects the system/env key to be in the format `${?sys.prop.key}`. Note that `?` signifies system/env variables.
- Added support to resolve system/env variables in `includes.conf`. Since `includes.conf` is in plain text with one include per line, it can not be parsed as `TypeSafe Config`. To workaround that we build a `TypeSafe Config` objects with key `includes` and value a list of include paths. This config object is then resolved with fallback on system/env properties.

E.g. If the includes.conf file looks like

<pre>
/path/to/tag1
/path/to/${?sys.key1}
</pre>

It gets loaded in memory as a `TypeSafe Config` object, 
<pre>
includes = [/path/to/tag1, /path/to/${?sys.key1}]
</pre>

Then resolved with fallback on system/env properties as,
<pre>
includes = [/path/to/tag1, /path/to/sys.value1]
</pre>

- Fixed bug in resolutions. `Config.resolve` should be called only once after all the fallbacks have been configured

@tuGithub and @chavdar please review.

### DIFF
--- a/gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
+++ b/gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
@@ -131,10 +131,10 @@ public class TestConfigClient {
   private void mockupConfigValues(){
     /**
      * each node will have a common key "generalKey" with value as "generalValue_${node}"
-     * this key will be overwrite 
-     * 
+     * this key will be overwrite
+     *
      * each node will have own key "keyOf_${node}" with value "valueOf_${node}"
-     * this key will be inherent 
+     * this key will be inherent
      */
     // mock up the configuration values for root
     Map<String, String> rootMap = new HashMap<>();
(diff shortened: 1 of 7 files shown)


chavdar:
${?sys.prop.key} signifies that the assignment is a no-op if the environment variable is missing. It is certainly usable for properties outside environment variables and Java system properties.

chavdar:
LGTM

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/ConfigStoreBackedValueInspector.java
```diff
@@ -136,21 +127,40 @@ public Config getResolvedConfig(ConfigKeyPath configKey) {
...
       for(ConfigKeyPath p: ownImports){
-        initialConfig = initialConfig.withFallback(this.getResolvedConfig(p));
+        initialConfig = initialConfig.withFallback(this.getResolvedConfigRecursive(p));
       }
     }
-    
+
     // merge with configs from parent for Non root
-    initialConfig = initialConfig.withFallback(this.getResolvedConfig(configKey.getParent())).resolve();
+    initialConfig = initialConfig.withFallback(this.getResolvedConfigRecursive(configKey.getParent()));
```
tuGithub: Why remove the "resolve" function at the end?

gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
```diff
@@ -131,10 +131,10 @@ public void setup(){
   private void mockupConfigValues(){
     /**
      * each node will have a common key "generalKey" with value as "generalValue_${node}"
-     * this key will be overwrite 
-     * 
+     * this key will be overwrite
```
tuGithub: A lot of format changes, please file a separate PR for format changes in the future, then we do NOT need to "really" review those changes and reduce a lot of noise.

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/common/impl/ConfigStoreBackedValueInspector.java
pcadabam: This was a bug to resolve on every recursive call. The resolve has been moved to `getResolvedConfig` after all the recursive fallbacks are set. 
As per typesafe documentation, resolve should be called only once after all the fallbacks are set. 
See unit test `TestConfigStoreValueInspector.testResolveConfigOverridingInChild()` which fails if you resolve on every recursive call.

gobblin-config-management/gobblin-config-client/src/test/java/gobblin/config/client/TestConfigClient.java
pcadabam: Will do. My eclipse auto formats tabs > spaces on save. I will change it.

tuGithub:
LGTM

PR #930, opened 2016-04-15
log resolved retention config

pcadabam:


pcadabam:
Merging trivial change.

PR #931, opened 2016-04-17
IDEA integration improvement

shirshanka:
Fixed out of the box integration with idea. 
./gradlew -PuseHadoop2 idea 
and File > Open in idea works without errors (earlier broke on java 7 language level)

Language level set to java 7, docs updated, codestyle checked in (with import order fixed based on coding style).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,7 @@ apply from: rootProject.projectDir.path + '/gradle/scripts/jacoco-coveralls-supp
 
 idea.project {
   ext.languageLevel = JavaVersion.VERSION_1_7
+  languageLevel = JavaVersion.VERSION_1_7
 }
 
 ext.build_script_dir = "${projectDir.path}/build_script"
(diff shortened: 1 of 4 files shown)


shirshanka:
@sahilTakiar or @ibuenros : can you review that this works with your idea integrations as well?

build.gradle
```diff
@@ -28,6 +28,7 @@ apply from: rootProject.projectDir.path + '/gradle/scripts/jacoco-coveralls-supp
 
 idea.project {
   ext.languageLevel = JavaVersion.VERSION_1_7
```
sahilTakiar: Should we remove this?

ibuenros: Chavdar mentioned at some point that was necessary for some new version of Gradle, so it is there in preparation for the migration.

sahilTakiar:
Should we also specify the `jdkName` - https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.idea.model.IdeaProject.html

Other than that LGTM! Thanks for making this change, I tested this locally and it works.

shirshanka:
Don't want to force a jdkName since people could use any jdk upwards of 1.7 I believe.

build.gradle
sahilTakiar: Yes, I believe that is for the Gradle 2.+ migration. However, the `ext.` prefix is only for global properties that need to be accessed project wide. It's basically a global map. `languageLevel` is specific to the configuration of the `idea` plugin, so its not a global property.

In fact, this commit probably broke the project version specification for idea: https://github.com/linkedin/gobblin/commit/15984d7b16056b00ec9aa27b0bfb789b1ddd174d

Issue #881, opened 2016-03-25
State management after running workunits


zliu41:
Suppose the sizes of the job state and each task state is 0.9M and 0.1M respectively, and there are 1000 tasks.

There are two issues that cause high memory usage:
1. When a task runs, it merges the job state with the properties in the workunit (e.g., in MR mode, this is done in `MRJobLauncher.TaskRunner.map()`. So each task state is 1M. When tasks complete, `TaskStateCollectorService` collects all task states into memory. That takes 1G memory, 900M of which is redundant.
2. In `JobContext.commit`, the first step is to create dataset states. Each dataset state also contains a copy of job state (due to the `addAll` call in `JobState.newDatasetState()`). In the worst case, there are as many dataset states as task states (i.e., 1000), so that takes another 900M of memory, which are also redundant information.

As a result the memory usage is 1.9G, but theoretically we only need 100M.

As can be seen from [this heap dump](http://s11.postimg.org/kydflisj7/image.png), a huge amount of memory is consumed by states, and there's a lot of redundancy.

One possible solution is to never use more than one copy of job state, and either persist job state and task states separately in the state store, or only merge job state with each task state in the last moment before they are persisted.

PR #932, opened 2016-04-18
Separate jobstate from taskstate and datasetstate

zliu41:
This is a solution to #881.
Summary:
- Each `WorkUnitState`/`TaskState` now contains a `JobState` reference, rather than a copy of all properties in `JobState`. `WorkUnitState.getProp()` will first look for the property in its own properties, then in its `WorkUnit`, then in its `JobState`.
- `DatasetState` no longer contains properties in `JobState`, and no longer supports `setProp` or `getProp` methods (will throw `UnsupportedOperationException`). A `DatasetState` can only contain a few specific properties such as `dataset.urn`, and they are accessed via specific setter and getter methods (e.g., `setDatasetUrn()` and `getDatasetUrn()`).
 - The rationale is the following: a `DatasetState` is created in the driver after the job is done. At this point there shouldn’t be a need to look for, change or add any general properties, especially since the driver is not something pluggable like the constructs, where different people can plug in an arbitrarily customized one for their use cases. Also, the driver already has a copy of `JobState`. If it wants to query any properties it can directly query the `JobState`.
 - This means that properties in the `JobState`, which are properties from the job config file and the global property file, will not be persisted into the state store. Since these are static properties rather than runtime properties (runtime properties are in `TaskState`s), there's no need to persist them.

These changes result in a ~90% reduction in memory consumption of a Gobblin job.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
@@ -312,7 +312,10 @@ public class WorkUnitState extends State {
 
     WorkUnitState other = (WorkUnitState) object;
     return ((this.workUnit == null && other.workUnit == null)
-        || (this.workUnit != null && this.workUnit.equals(other.workUnit))) && super.equals(other);
+        || (this.workUnit != null && this.workUnit.equals(other.workUnit)))
+        && ((this.jobState == null && other.jobState == null)
+            || (this.jobState != null && this.jobState.equals(other.jobState)))
+        && super.equals(other);
   }
 
   @Override
--- a/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
@@ -27,6 +27,7 @@ import gobblin.source.extractor.Watermark;
 import gobblin.source.workunit.Extract;
 import gobblin.source.workunit.ImmutableWorkUnit;
 import gobblin.source.workunit.WorkUnit;
+import lombok.Getter;
 
 
 /**
@@ -66,22 +67,35 @@ public class WorkUnitState extends State {
     CANCELLED
   }
 
-  private WorkUnit workunit;
+  private final WorkUnit workUnit;
+
+  @Getter
+  private State jobState;
 
   /**
    * Default constructor used for deserialization.
    */
   public WorkUnitState() {
-    this.workunit = WorkUnit.createEmpty();
+    this.workUnit = WorkUnit.createEmpty();
+    this.jobState = new State();
   }
 
   /**
    * Constructor.
    *
    * @param workUnit a {@link WorkUnit} instance based on which a {@link WorkUnitState} instance is constructed
+   * @deprecated It is recommended to use {@link #WorkUnitState(WorkUnit, State)} rather than combining properties
+   * in the job state into the workunit.
    */
+  @Deprecated
   public WorkUnitState(WorkUnit workUnit) {
-    this.workunit = workUnit;
+    this.workUnit = workUnit;
+    this.jobState = new State();
+  }
+
+  public WorkUnitState(WorkUnit workUnit, State jobState) {
+    this.workUnit = workUnit;
+    this.jobState = jobState;
   }
 
   /**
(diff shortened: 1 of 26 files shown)


zliu41:
@chavdar @sahilTakiar @ydai1124 @ibuenros can you review?

chavdar:
LGTM

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -287,21 +311,22 @@ public boolean equals(Object object) {
     }
 
     WorkUnitState other = (WorkUnitState) object;
-    return ((this.workunit == null && other.workunit == null)
-        || (this.workunit != null && this.workunit.equals(other.workunit))) && super.equals(other);
+    return ((this.workUnit == null && other.workUnit == null)
```
sahilTakiar: Should this method and `hashCode()` be updated to include the `jobState`. Alternatively, we could just use lombok's `@EqualsAndHashCode`

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -564,8 +564,12 @@ public DatasetState newDatasetState(boolean fullCopy) {
...
-   * A subclass of {@link JobState} that is used to represent dataset states. This class is currently
-   * identical to {@link JobState} except that the name is more meaningful and less confusing.
+   * A subclass of {@link JobState} that is used to represent dataset states.
+   *
+   * <p>
+   *   A {@code DatasetState} does <em>not</em> contain any properties. Operations such as {@link #getProp(String)}
+   *   and {@link #setProp(String, Object)} are not supported.
+   * </p>
    */
   public static class DatasetState extends JobState {
```
sahilTakiar: Does `DatasetState` need to extend `JobState` anymore? I'm not sure why it does in the first place, a `DatasetState` should need to know things like `jobId` and doesn't really have a `startTime` or an `endTime`.

sahilTakiar: In fact, given than the `getProp` and `setProp` methods throw `UnsupportedOperationException` does this even need to extend `State`?

gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
```diff
@@ -83,21 +83,20 @@
...
-  public TaskState() {
-  }
+  public TaskState() {}
 
   public TaskState(WorkUnitState workUnitState) {
     // Since getWorkunit() returns an immutable WorkUnit object,
     // the WorkUnit object in this object is also immutable.
-    super(workUnitState.getWorkunit());
+    super(workUnitState.getWorkunit(), workUnitState.getJobState());
     addAll(workUnitState);
```
sahilTakiar: We should also fix that so that a `TaskState` only has a pointer to its `WorkUnitState`, rather than making a full copy of all the properties in the `WorkUnitState`.

On an unrelated note, whats the difference between `TaskState` and `WorkUnitState`? I don't see any functional reason why the two can't be merged, they both seem to try to track some runtime information. But thats a discussion for another day.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
zliu41: Because dataset states need to be persisted into the state store, which, by definition, only takes `State` objects. Also, extending `State` and `JobState` enables it to inherit methods such as `read`,  `write`, `addTaskState`, etc.

gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
zliu41: I agree, I think `WorkUnitState` and `SourceState` are useless, they are the same as `TaskState` and `JobState`, respectively. Also, `Extract` should also be deprecated as it is conceptually identical as `Dataset`. These changes will be part of another PR.

zliu41:
@sahilTakiar updated 53b908f

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
sahilTakiar: Yes it will probably require some more re-factoring, so it can be done later.

sahilTakiar:
LGTM!

PR #941, opened 2016-04-25
Delete empty parent directories on Hive de-registration. Optimize der…

ibuenros:
This PR introduces optimizations to the delete paths logic for distcp-ng.

* User can set `hive.dataset.copy.deregister.deleteLocationRecursively` to true, which will simply delete the root directory location for deregistered partitions (as opposed to deleting the files used by the input format). This is faster, as it involves a single delete. However, if the directory contains data that should not be deleted, this should be set to false.
* Delete step will automatically delete empty parent directories up to the table root.


ibuenros:
@ydai1124 can you review?

ydai1124:
LGTM

PR #942, opened 2016-04-25
Allow disabling state store.

ibuenros:
This PR adds an option to disable state store. (Default is state store is enabled). This is useful when jobs don't use the state store, saving the load of reading / writing it. It also provides a workaround for a memory issue when extracts change name often, as this would cause Gobblin to load previous metadata for many runs, causing OOM.


ibuenros:
@sahilTakiar can you review?

sahilTakiar:
LGTM

PR #944, opened 2016-04-25
Bin pack distcp-ng work units.

ibuenros:
Implemented bin packing for distcp-ng. Made it generic enough to be usable by other Gobblin flows.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnitBinPacker.java
+++ b/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnitBinPacker.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.source.workunit;
+
+import java.util.List;
+
+/**
+ * A bin packing algorithm for packing {@link WorkUnit}s into {@link MultiWorkUnit}s.
+ */
+public interface WorkUnitBinPacker {
+
+  /**
+   * Packs the input {@link WorkUnit}s into {@link MultiWorkUnit}s.
+   * @param workUnitsIn List of {@link WorkUnit}s to pack.
+   * @param weighter {@link WorkUnitWeighter} that provides weights for {@link WorkUnit}s.
+   */
+  public List<WorkUnit> pack(List<WorkUnit> workUnitsIn, WorkUnitWeighter weighter);
+
+}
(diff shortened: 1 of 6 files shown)


ibuenros:
@pcadabam can you review?

pcadabam:
Few comments
- I think bin packing should be in gobblin-api/gobblin.source.workunit

gobblin-utility/src/main/java/gobblin/util/binpacking/WorstFitDecreasingBinPacking.java
```diff
@@ -0,0 +1,147 @@
...
+ *   * Compute the minimum number of {@link MultiWorkUnit}s needed, and create them.
+ *   * For each work unit, find the {@link MultiWorkUnit} with the largest space available, if it fits, add it there,
+ *     otherwise, create a new {@link MultiWorkUnit} and add the work unit there.
+ * </p>
+ */
+public class WorstFitDecreasingBinPacking {
+
+  public static final String TOTAL_MULTI_WORK_UNIT_WEIGHT = "binpacking.multiWorkUnit.totalWeight";
+
+  public List<WorkUnit> pack(List<WorkUnit> workUnitsIn, Weighter weighter, long maxWeigthPerUnit) {
```
pcadabam: `WorstFitDecreasingBinPacking` should implement `WorkUnitBinPacking` interface. The interface can have a `public List<WorkUnit> pack(List<WorkUnit> workUnitsIn, Weighter weighter, long maxWeigthPerUnit)` method

gobblin-utility/src/test/java/gobblin/util/binpacking/WorstFitDecreasingBinPackingTest.java
```diff
@@ -0,0 +1,98 @@
...
+import gobblin.source.workunit.MultiWorkUnit;
+import gobblin.source.workunit.WorkUnit;
+
+
+public class WorstFitDecreasingBinPackingTest {
+
+  public static final String WEIGHT = "weight";
+
+  @Test
+  public void testPack()
```
pcadabam: As a general note it is better to break down a test class into logical test methods. The idea is to have as few Asserts in a test method as possible. This is makes it easy to debug when a test fails.

zliu41:
This may overlap with `KafkaWorkUnitPacker`, is it possible to consolidate these two?

ibuenros:
@zliu41 yes, in fact I pretty much copied it from `KafkaWorkUnitPacker`. However, the algorithm is slightly different (the Kafka bin packer has bounded number of workunits, this algorithms bounds the total weight of each work unit), so I didn't want to change what the `KafkaWorkUnitPacker` uses.
I can either change kafka bin packing to use this new algorithm, or implement the other algorithm as well, although for the latter, I would prefer to do that in a later PR.

ibuenros:
@pcadabam addressed comments. Can you take another quick look? Thanks!

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnitBinPacker.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * A bin packing algorithm for packing {@link WorkUnit}s into {@link MultiWorkUnit}s.
+ */
+public interface WorkUnitBinPacker {
+
+  /**
+   * Packs the input {@link WorkUnit}s into {@link MultiWorkUnit}s.
+   */
+  public List<WorkUnit> pack(List<WorkUnit> workUnitsIn, WorkUnitWeighter weighter);
```
pcadabam: Add `@param` in java doc

pcadabam:
LGTM with that one comment about javadoc

PR #946, opened 2016-04-27
Copyable file performance

ibuenros:
Generating a new `FileSystem` per work unit is very costly (even though Hadoop caches them). This PR stops doing that. 
Also write out cache statistics after creating work units for distcp.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -13,6 +13,7 @@
 package gobblin.data.management.copy.hive;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
@@ -22,6 +23,7 @@ import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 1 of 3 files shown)


ibuenros:
@ydai1124 can you review?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -343,7 +343,6 @@ public void close() throws IOException {
     try {
       super.close();
     } finally {
-      this.clientPool.close();
```
ibuenros: Don't know why this shows up here, this is an old change.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -678,12 +680,24 @@ private void checkPartitionCompatibility(Partition desiredTargetPartition, Parti
...
+
     for (FileStatus status : paths) {
       dataFiles.add(new SourceAndDestination(status, getTargetPath(status.getPath(), this.targetFs, partition, true)));
     }
 
     for (SourceAndDestination sourceAndDestination : dataFiles) {
-      FileSystem actualSourceFs = sourceAndDestination.getSource().getPath().getFileSystem(new Configuration());
+
+      URI uri = sourceAndDestination.getSource().getPath().toUri();
+      if (actualSourceFs == null || !StringUtils.equals(referenceScheme, uri.getScheme())
```
ydai1124: Do you need a cache for the sourceFs, from `uri` to `Fs`?

ibuenros: I think that's overkill. Honestly, most of the time there is a single `FileSystem` needed. Just putting this logic in case there is some super-edge case.

ydai1124: I see.

ydai1124:
LGTM

PR #948, opened 2016-04-28
Newest k files

chavdar:
* Moved the logic of NonNewestKSelectionPolicy to NewestKSelectionPolicy since the name and options for the former are confusing and the implementation is not efficient
* The NonNewestKSelectionPolicy semantics is covered by a new config selection.newestK.versionsNotSelected
* Made NewestKSelectionPolicy non-generic since the type is not used anywhere
* Added tests and extended documentation

@pcadabam or @zliu41 can you review ?


gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
```diff
@@ -28,30 +29,89 @@
...
-    LOGGER.info(String.format("%s will select %d versions of each dataset.", NewestKSelectionPolicy.class.getName(),
+    LOGGER.info(String.format("Will %s %d versions of each dataset.",
+        (this.versionsSelected >= 0 ? "select" : "exclude"),
         this.versionsSelected));
   }
 
   public NewestKSelectionPolicy(Properties props) {
-    this.versionsSelected =
-        Integer.parseInt(props.getProperty(NEWEST_K_VERSIONS_SELECTED_KEY, VERSIONS_SELECTED_DEFAULT));
+    this(getNumVersionsFromProps(props));
```
pcadabam: Seems like duplication of logic in `getNumVersionsFromProps` and `getNumVersionsFromConfig` can you call `getNumVersionsFromConfig(ConfigFactory.parseProps(props))` ?

gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
```diff
@@ -28,30 +29,89 @@
...
+   */
   public static final String NEWEST_K_VERSIONS_SELECTED_KEY = "selection.newestK.versionsSelected";
 
-  public static final String VERSIONS_SELECTED_DEFAULT = Integer.toString(2);
+  /**
+   * The number of newest versions to exclude from the result. Only one of
+   * {@link #NEWEST_K_VERSIONS_SELECTED_KEY} and {@link #NEWEST_K_VERSIONS_NOTSELECTED_KEY} can
+   * be specified. The default is {@link #NEWEST_K_VERSIONS_SELECTED_KEY} with a value of
+   * {@link #VERSIONS_SELECTED_DEFAULT}. Valid values are in the range
+   * [{@link #MIN_VERSIONS_ALLOWED}, {@link #MAX_VERSIONS_ALLOWED}]. Negative values are interpreted
```
pcadabam: I am not completely sure about this. Was there a reason to support Negative values for these keys? This would mean there are 4 ways to configure this class
versionsSelected : positive k
versionsSelected : negative k
versionsExcluded : positive k
versionsExcluded : negative k

2,3 mean the same and 1,4 mean the same. Since the desired behavior can be achieved by using positive k's only, I think we should restrict users to specify positive value. We will have fewer code branches.

<pre>
 public Collection<DatasetVersion> listSelectedVersions(List<DatasetVersion> allVersions) {
  if (versionsSelected > 0) {
       return boundarySafeSublist(allVersions, 0, versionsSelected);
  } else {
       return boundarySafeSublist(allVersions, versionsExcluded, allVersions.size());
  }
}
</pre>
What do you think?

gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
chavdar: Yeah, I was not sure where the props constructor is used. Is there a possibility that the object is instantiated with a full state object with some of known Properties/Config incompatibilities?

gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
chavdar: Yeah, I was divided on this. Fixed.

gobblin-data-management/src/test/resources/retentionIntegrationTest/testCombinePolicy/selection.conf
```diff
@@ -11,10 +11,10 @@ gobblin.retention : {
       combine.operation=INTERSECT
       combine.policy.classes=[
         gobblin.data.management.policy.SelectBeforeTimeBasedPolicy,
-        gobblin.data.management.policy.NonNewestKSelectionPolicy
+        gobblin.data.management.policy.NewestKSelectionPolicy
       ]
       timeBased.lookbackTime=3d
-      newestK.versionsSelected=2
+      newestK.versionsNotSelected=2
```
zliu41: Should this be -2?

gobblin-data-management/src/test/resources/retentionIntegrationTest/testNewestKRetention/selection.conf
```diff
@@ -7,8 +7,8 @@ gobblin.retention : {
     }
 
     selection : {
-      policy.class=gobblin.data.management.policy.NonNewestKSelectionPolicy
-      newestK.versionsSelected=2
+      policy.class=gobblin.data.management.policy.NewestKSelectionPolicy
+      newestK.versionsNotSelected=2
```
zliu41: Ditto

gobblin-data-management/src/test/java/gobblin/data/management/policy/NewestKSelectionPolicyTest.java
```diff
@@ -0,0 +1,182 @@
+package gobblin.data.management.policy;
```
zliu41: License header

zliu41:
LGTM barring above comments.

gobblin-data-management/src/test/resources/retentionIntegrationTest/testCombinePolicy/selection.conf
chavdar: We disabled the negative values for the options. The previous NonNewestKSelectionPolicy with newestK.versionsSelected=K  is now equivalent to NewestKSelectionPolicy with newestK.versionsNotSelected=K

gobblin-data-management/src/test/java/gobblin/data/management/policy/NewestKSelectionPolicyTest.java
chavdar: Fixed.

gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
pcadabam: I don't think so. This is a fairly new class. It should be fine to parse props as config.

pcadabam:
LGTM apart from these two comments
1. parse props as config and use config's constructor 
2. When versionExcluded is set, do we need to read it as a negative number and then again convert it to positive while finding sublist? why not have two member variables versionsSelected and versionsExcluded in the class

gobblin-data-management/src/main/java/gobblin/data/management/policy/NewestKSelectionPolicy.java
chavdar: Ok. Fixed.

chavdar:
@pcadabam Changing the negative value semantics required a bit of a surgery because of dealing with default values. Nothing that can't be solved with another level of indirection.

chavdar:
Superceded by https://github.com/linkedin/gobblin/pull/951

sahilTakiar:
> Made NewestKSelectionPolicy non-generic since the type is not used anywhere

It was made generic as part of https://github.com/linkedin/gobblin/pull/926 so code like this is possible:

```
NewestKSelectionPolicy<FileSystemDatasetVersion> test = new NewestKSelectionPolicy<>(1, true);
Collection<FileSystemDatasetVersion> selectedVersions = test.listSelectedVersions(Lists.<FileSystemDatasetVersion> newArrayList());
```

This is no longer possible with the given changes, unless do you a cast. Why does it need to be removed?

PR #950, opened 2016-04-28
Add delete methods when deregistering Hive partitions in distcp.

ibuenros:
Allows not deleting files when deregistering a Hive partition in distcp.


ibuenros:

PR #952, opened 2016-04-29
Unified way to execute Hive and MR-based compaction jobs

lbendig:
Currently the execution of the Hive and MR-based compactors is not out of the box, needs extra work:
- there's no implementation for MR compaction for CLI execution
- Hive compactor is a standalone package that needs to be built separately

This PR tries to fix these issues. A new shell script **bin/gobblin-compaction.sh** was added with which one can execute both Hive and MR-based compaction jobs from the CLI.
```
Usage: gobblin-compaction.sh [OPTION] --type <compaction type: hive or mr> --conf <compaction configuration file>
Where OPTION can be:
  --projectversion <version>    Gobblin version to be used. If set, overrides the distribution build version
  --logdir <log dir>            Gobblin's log directory: if not set, taken from ${GOBBLIN_LOG_DIR} if present. 
  --help                        Display this help and exit
```

I assumed that the executable gobblin-compaction.jar and tarball creation are no longer needed.

### DIFF
--- a/bin/gobblin-compaction.sh
+++ b/bin/gobblin-compaction.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+
+##############################################################
+############ Run Gobblin Compaction on Hadoop MR #############
+##############################################################
+
+# Set during the distribution build
+GOBBLIN_VERSION=@project.version@
+
+FWDIR="$(cd `dirname $0`/..; pwd)"
+FWDIR_LIB=$FWDIR/lib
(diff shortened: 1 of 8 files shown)


lbendig:
@zliu41 what do you think about this add-on? Can you please review?

zliu41:
Sure, I'll take a look.

zliu41:
LGTM. Thanks for the contribution.

lbendig:
@zliu41 thanks for the review. I'll create a PR for the documentation.

PR #953, opened 2016-04-29
Added instrumentation to Hive copy.

ibuenros:
Adds timers to Hive distcp work unit generation.
Also adds a `MultiTimingEvent` class that emits a single event for timings of different stages in a process.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -43,6 +43,7 @@ import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.io.Closer;
 import com.google.gson.Gson;
 
 import gobblin.commit.CommitStep;
@@ -61,6 +62,8 @@ import gobblin.hive.PartitionDeregisterStep;
 import gobblin.hive.metastore.HiveMetaStoreUtils;
 import gobblin.hive.spec.HiveSpec;
 import gobblin.hive.spec.SimpleHiveSpec;
+import gobblin.metrics.event.EventSubmitter;
+import gobblin.metrics.event.MultiTimingEvent;
 import gobblin.util.PathUtils;
 import gobblin.util.commit.DeleteFileCommitStep;
 import gobblin.util.reflection.GobblinConstructorUtils;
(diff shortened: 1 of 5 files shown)


ibuenros:
@abti can you review?

ibuenros:
@zliu41 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -199,6 +220,10 @@
 
     log.info("Finding copy entities for table " + dataset.table.getCompleteName());
 
+    this.eventSubmitter = new EventSubmitter.Builder(dataset.getMetricContext(), "hive.dataset.copy").build();
+    Closer closer = Closer.create();
```
abti: closer init should be in try block and closer.closer() in finally to be on the safer side

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -434,17 +463,22 @@ private Partition getTargetPartition(Partition originPartition, Path targetLocat
     private final Partition partition;
     private final Properties properties;
     private Optional<Partition> existingTargetPartition;
+    private final EventSubmitter eventSubmitter;
 
     public PartitionCopy(Partition partition, Properties properties) {
       this.partition = partition;
       this.properties = properties;
       this.existingTargetPartition = Optional.fromNullable(targetPartitions.get(this.partition.getValues()));
+      this.eventSubmitter = new EventSubmitter.Builder(dataset.getMetricContext(), "dataset.hive.copy").
```
abti: Second param to EventSubmitter.Builder is inconsistent with HiveEntityCopyHelper; there the second argument was "hive.dataset.copy, here it is "dataset.hive.copy"

gobblin-metrics/src/main/java/gobblin/metrics/event/MultiTimingEvent.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Ends the current stage and submits the event containing the timings of each event.
+   * @param additionalMetadata additional metadata to include in the event.
+   * @throws IOException
+   */
+  public void submit(Map<String, String> additionalMetadata) throws IOException {
+
+    if (this.submitted) {
```
abti: MultiTimingEvent seems thread unsafe during close(), can we call it out in java docs?

abti:
@ibuenros .. a few comments, rest looks good to me

ibuenros:
Fixed. Merging.

PR #956, opened 2016-05-03
Make the default store for SimpleHDFSConfigStoreFactory configurable

chavdar:
Changes:

- Add a default_store_uri option to specify the default store
- Add validation in the constructor if the default config root exists
- Make changes to support use of any Hadoop FS rather than just HDFS.
- Add some unit tests.

@tuGithub @pcadabam Can you review?

### DIFF
--- a/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
+++ b/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
@@ -19,7 +19,6 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.security.UserGroupInformation;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
@@ -127,7 +126,7 @@ public class SimpleHDFSConfigStoreFactory implements ConfigStoreFactory<SimpleHD
    * Gets a default root directory if one is not specified. The default root dir is {@code /jobs/[current-user]/}.
    */
   protected Path getDefaultRootDir() throws IOException {
-    return new Path("/jobs", UserGroupInformation.getCurrentUser().getUserName());
+    return this.defaultStoreFS.getHomeDirectory();
   }
 
   /**
--- a/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
+++ b/gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
@@ -41,19 +41,25 @@ public class SimpleHDFSConfigStoreFactory implements ConfigStoreFactory<SimpleHD
   protected static final String SIMPLE_HDFS_SCHEME_PREFIX = "simple-";
   protected static final String HDFS_SCHEME_NAME = "hdfs";
 
-  public final static String DEFAULT_STORE_URI_KEY = "default_store_uri";
-
-  public SimpleHDFSConfigStoreFactory() {
-    this(ConfigFactory.empty());
-  }
+  public static final String DEFAULT_CONFIG_NAMESPACE = SimpleHDFSConfigStoreFactory.class.getName();
+  public static final String DEFAULT_STORE_URI_KEY = "default_store_uri";
 
   private final Optional<URI> defaultStoreURI;
   private final FileSystem defaultStoreFS;
+
+  /** Instantiates a new instance using standard typesafe config defaults:
+   * {@link ConfigFactory#load()} */
+  public SimpleHDFSConfigStoreFactory() {
+    this(ConfigFactory.load().getConfig(DEFAULT_CONFIG_NAMESPACE));
+  }
 
   public SimpleHDFSConfigStoreFactory(Config factoryConfig) {
     try {
       if (factoryConfig.hasPath(DEFAULT_STORE_URI_KEY)) {
         String uri = factoryConfig.getString(DEFAULT_STORE_URI_KEY);
+        if (Strings.isNullOrEmpty(uri)) {
+          throw new IllegalArgumentException("Default store URI should be non-empty!");
+        }
         Path defStorePath = new Path(uri);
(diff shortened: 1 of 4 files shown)


gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -38,6 +41,47 @@
...
+  }
+
+  private final Optional<URI> defaultStoreURI;
+  private final FileSystem defaultStoreFS;
+
+  public SimpleHDFSConfigStoreFactory(Config factoryConfig) {
+    try {
+      if (factoryConfig.hasPath(DEFAULT_STORE_URI_KEY)) {
+        String uri = factoryConfig.getString(DEFAULT_STORE_URI_KEY);
+        Path defStorePath = new Path(uri);
```
pcadabam: Is it possible for DEFAULT_STORE_URI_KEY to be set to null and we get a NPE here?
You can use `ConfigUtils.hasNonEmptyPath(factoryConfig, DEFAULT_STORE_URI_KEY)` instead of `factoryConfig.hasPath(DEFAULT_STORE_URI_KEY)`

gobblin-config-management/gobblin-config-core/src/test/java/gobblin/config/store/hdfs/SimpleHdfsConfigureStoreFactoryTest.java
```diff
@@ -44,4 +47,49 @@ public void testGetDefaults() throws URISyntaxException, ConfigStoreCreationExce
...
+
+
+  @Test
+  public void testConfiguration() throws Exception {
+    FileSystem localFS = FileSystem.getLocal(new Configuration());
+    Path testRoot = localFS.makeQualified(new Path("dir1"));
+    Path configRoot = localFS.makeQualified(new Path(testRoot, "dir2"));
+    Path configStoreRoot = new Path(configRoot,
+                                    SimpleHDFSConfigStore.CONFIG_STORE_NAME);
+    Assert.assertTrue(localFS.mkdirs(configStoreRoot));
```
pcadabam: Is this deleted after the test?

gobblin-config-management/gobblin-config-core/src/test/java/gobblin/config/store/hdfs/SimpleHdfsConfigureStoreFactoryTest.java
```diff
@@ -44,4 +47,49 @@ public void testGetDefaults() throws URISyntaxException, ConfigStoreCreationExce
...
+    try {
+      Config confConf1 =
+          ConfigFactory.empty().withValue(SimpleHDFSConfigStoreFactory.DEFAULT_STORE_URI_KEY,
```
pcadabam: Nice.I was not aware of this way to create `Config`. I usually use `ConfigFactory.parseMap`

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
```diff
@@ -38,6 +41,47 @@
...
+  public final static String DEFAULT_STORE_URI_KEY = "default_store_uri";
+
+  public SimpleHDFSConfigStoreFactory() {
+    this(ConfigFactory.empty());
+  }
+
+  private final Optional<URI> defaultStoreURI;
+  private final FileSystem defaultStoreFS;
+
+  public SimpleHDFSConfigStoreFactory(Config factoryConfig) {
```
pcadabam: When is this constructor used by the `ConfigClient`? The `ConfigStoreFactory`s are loaded by `ServiceLoader` in `gobblin.config.client.ConfigStoreFactoryRegister`. And it looks like it uses the default no args constructor.

pcadabam: Can you add some java doc for this constructor?

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
chavdar: I'd rather throw an exception rather than just silently ignore it. I've added a check throwing an IllegalArgumentException.

gobblin-config-management/gobblin-config-core/src/test/java/gobblin/config/store/hdfs/SimpleHdfsConfigureStoreFactoryTest.java
chavdar: One of the ancestors is :

    finally {
      localFS.delete(testRoot, true);
    }

gobblin-config-management/gobblin-config-core/src/test/java/gobblin/config/store/hdfs/SimpleHdfsConfigureStoreFactoryTest.java
chavdar: :-) In a branch, I had a helper config builder class to simplify some of the bolier plate. I just to have to find the branch.

gobblin-config-management/gobblin-config-core/src/test/java/gobblin/config/store/hdfs/SimpleHdfsConfigureStoreFactoryTest.java
pcadabam: But finally does not get called if `localFS.mkdirs(configStoreRoot)` throws exception. May be everything should be moved in the try block

gobblin-config-management/gobblin-config-core/src/main/java/gobblin/config/store/hdfs/SimpleHDFSConfigStoreFactory.java
chavdar: Good point. I have modified the default constructor to use systemProperties/resources to configure a global instance. For other instances, we can register specific config using ConfigStoreFactoryRegister.register(). 

Actually, there needs to be a follow-up to modify the scheme handling as currently we can have only one HDFS-based instance.

chavdar:
@pcadabam any more comments?

pcadabam:
LGTM. Please add javadoc to `public SimpleHDFSConfigStoreFactory(Config factoryConfig)` I see only java doc for `public SimpleHDFSConfigStoreFactory()`

PR #957, opened 2016-05-03
Modify QueryBasedSource to allow pulling multiple tables in the same job

zliu41:
This PR modifies `QueryBasedSource` so that it is possible to specify multiple table names in the same job.

Previously a job can only specify a single table name via `source.entity` or `extract.table.name`. Now two new properties are added: `source.entities` and `extract.table.names`.

Related changes:
- Moved the method to obtain topic-specific properties from `KafkaSource` to `DatasetUtils` so that it can be used by `QueryBasedSource` as well.
- Removed the usage of the deprecated watermark methods in `QueryBasedSource` and `QueryBasedExtractor`, and replaced them with the new JSON-based `Watermark`.
- Use a simple round robin method in `QueryBasedSource` to pack `WorkUnit`s into `MultiWorkUnit`s. More sophisticated bin packing methods can be added later if there's a need to do so.


zliu41:
@sahilTakiar @chavdar can you review?

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -140,6 +143,42 @@ public JsonElement getActualHighWatermark() {
...
+   * Get the actual high {@link Watermark}. If the {@code WorkUnitState} does not contain the actual high watermark
+   * (which may be caused by task failures), the low watermark in the corresponding {@link WorkUnit} will be returned.
+   *
+   * @param watermarkClass the watermark class for this {@code WorkUnitState}.
+   * @param gson a {@link Gson} object used to deserialize the watermark.
+   * @return the actual high watermark in this {@code WorkUnitState}. null is returned if this {@code WorkUnitState}
+   * does not contain an actual high watermark, and the corresponding {@code WorkUnit} does not contain a low
+   * watermark.
+   */
+  public <T extends Watermark> T getActualHighWatermark(Class<T> watermarkClass, Gson gson) {
```
sahilTakiar: Can you explain why these new watermark methods are necessary?

gobblin-utility/src/main/java/gobblin/util/dataset/DatasetUtils.java
```diff
@@ -0,0 +1,119 @@
...
+   * </p>
+   *
+   * <p>
+   *   The "dataset" field also allows regular expressions. For example, one can specify key, value
+   *   "dataset" : "myDataset.*". In this case all datasets whose name matches the pattern "myDataset.*" will have
+   *   all the specified config properties added to their {@link WorkUnit}s. If more a dataset matches multiple
+   *   "dataset"s then the properties from all the {@link JsonObject}s will be added to their {@link WorkUnit}s.
+   * </p>
+   */
+  public static final String DATASET_SPECIFIC_PROPS = DATASET + ".specific.props";
```
sahilTakiar: Isn't this what the `ConfigStore` is for? Typically dataset level properties should be specified in the `ConfigStore` and retrieved via the `ConfigStore` APIs.

sahilTakiar:
Can you list out the files that have the actual changes mentioned in the PR description? A lot of the changes seem to be just re-factoring / cleanup.

gobblin-utility/src/main/java/gobblin/util/dataset/DatasetUtils.java
zliu41: Will add an option to obtain table specific properties from the config store, but will also keep the current approach to be backward compatible, and for those who don't have a config store.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
zliu41: Previously since `getActualHighWatermark` returns a `JsonElement`, each `Source` class that uses the new watermark needs to deserialize it using a `Gson` object. It is more convenient to provide these methods that directly return `Watermark`.

zliu41:
@sahilTakiar updated. Code review should focus on `QueryBasedSource`, `QueryBasedExtractor` and `LongWatermark`. The rest is mostly cleanup.

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
sahilTakiar: I see, makes sense

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -88,27 +82,6 @@
...
-   *
-   * <p>
-   *   The "topic.name" field also allows regular expressions. For example, one can specify key, value
-   *   "topic.name" : "myTopic.*". In this case all topics whose name matches the pattern "myTopic.*" will have all the
-   *   specified config properties added to their {@link WorkUnit}s. If more than one topic matches multiple "topic.name"s then
-   *   the properties from all the {@link JsonObject}s will be added to their {@link WorkUnit}s.
-   * </p>
-   */
-  @VisibleForTesting
-  static final String KAFKA_TOPIC_SPECIFIC_STATE = "kafka.topic.specific.state";
```
sahilTakiar: will this property still work with the new approach for adding topic specific state in the `QueryBasedSource`? Since Kafka ingestion is a pretty popular use case, we should maintain backwards compatibility. I believe I have seen at least one person on the google groups refer to using this config.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -12,236 +12,245 @@
...
+      }
+
+      long highWm = previousWus.getActualHighWatermark(LongWatermark.class).getValue();
+      if (!prevActualHighWatermarksByTable.containsKey(table)) {
+        prevActualHighWatermarksByTable.put(table, highWm);
+      } else {
+        prevActualHighWatermarksByTable.put(table, Math.max(prevActualHighWatermarksByTable.get(table), highWm));
+      }
 
-    switch (wmType) {
```
sahilTakiar: do we need to preserve this logic?

sahilTakiar:
Can you add a little more documentation about how `source.entities` and `extract.tables.names` works? It basically seems this is how you specify which tables are going to be pulled in this job?

Ideally the documentation would be in under `gobblin-docs/user-guide/Query-Based-Sources.md` - the new file doesn't have to be a full explanation of the `QueryBasedSource`, just the documentation relevant to these configs, we can fill in the rest of it later.

sahilTakiar:
Just the comments made above, everything else LGTM

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -12,236 +12,245 @@
...
+      tableNameToEntityMap.put(Utils.escapeSpecialCharacters(entity, ConfigurationKeys.ESCAPE_CHARS_IN_TABLE_NAME, "_"),
+          entity);
     }
 
-    TableType tableType = TableType.valueOf(state.getProp(ConfigurationKeys.EXTRACT_TABLE_TYPE_KEY).toUpperCase());
-    long previousWatermark = this.getLatestWatermarkFromMetadata(state);
-
-    Map<Long, Long> sortedPartitions = Maps.newTreeMap();
-    sortedPartitions.putAll(new Partitioner(state).getPartitions(previousWatermark));
+    Map<String, State> tableSpecificPropsMap = shouldObtainTablePropsFromConfigStore(state)
```
chavdar: This looks dangerous. The config store may be needed for some completely different purpose (e.g. in a converter or writer). With this change, the source is forced to support the config store. We want to go there eventually, but we are not there yet.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -161,6 +161,7 @@
   public static final String EXTRACT_TABLE_TYPE_KEY = "extract.table.type";
   public static final String EXTRACT_NAMESPACE_NAME_KEY = "extract.namespace";
   public static final String EXTRACT_TABLE_NAME_KEY = "extract.table.name";
```
chavdar: What is the difference between entity and table? QueryBasedSource seems to be using those interchangeably? Let's agree on terminology and stick to it.

chavdar:
> Use a simple round robin method in QueryBasedSource to pack WorkUnits into MultiWorkUnits.

Are we sure that this works better than randomizing the order of WUs and putting consecutive WUs into the same MultiWU ?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -161,6 +161,7 @@
...
+  public static final String EXTRACT_TABLE_NAMES_KEY = "extract.table.names";
```
chavdar: Can you please document what exactly the format of the property is? A comma-separated list of strings (table names)?

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -12,236 +12,245 @@
...
-   *            Work unit state
-   * @return latest low water mark
+   * TODO: this is currently a simple round-robin packing. More sophisticated bin packing may be necessary
+   * if the round-robin approach leads to mapper skew.
    */
-  private long getLowWatermarkFromWorkUnit(WorkUnitState workUnitState) {
-    String watermarkType = workUnitState
-        .getProp(ConfigurationKeys.SOURCE_QUERYBASED_WATERMARK_TYPE, ConfigurationKeys.DEFAULT_WATERMARK_TYPE);
-    long lowWaterMark = workUnitState.getWorkunit().getLowWaterMark();
+  private static List<WorkUnit> pack(List<WorkUnit> workUnits, int numOfMultiWorkunits) {
```
chavdar: It seems like bin-packing is a common pattern (Kafka, distcp-ng, here). Can we isolate the bin-packing logic and make it reusable?

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -12,236 +12,245 @@
...
+    Map<String, State> tableSpecificPropsMap = shouldObtainTablePropsFromConfigStore(state)
+        ? getTableSpecificPropsFromConfigStore(tableNameToEntityMap.keySet(), state)
+        : DatasetUtils.getDatasetSpecificProps(tableNameToEntityMap.keySet(), state);
+    Map<String, Long> prevWatermarksByTable = getPreviousWatermarksForAllTables(state);
 
-    // Use extract table name to create extract
-    SourceState partitionState = new SourceState();
-    partitionState.addAll(state);
-    Extract extract = partitionState.createExtract(tableType, nameSpaceName, extractTableName);
+    for (String tableName : Sets.union(tableNameToEntityMap.keySet(), prevWatermarksByTable.keySet())) {
```
chavdar: nit: iterating over the key set of a map is inefficient and requires a new set to be allocated. This will also trigger a FindBugs alerts. It is more efficient to iterate over the entry set.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -12,236 +12,245 @@
...
-    workUnits.addAll(previousWorkUnits);
+      long previousWatermark = prevWatermarksByTable.containsKey(tableName) ? prevWatermarksByTable.get(tableName)
+          : ConfigurationKeys.DEFAULT_WATERMARK_VALUE;
 
-    return workUnits;
-  }
+      // If a table name exists in prevWatermarksByTable (i.e., it has a previous watermark) but does not exist
+      // in talbeNameToEntityMap, create an empty workunit for it, so that its previous watermark is preserved.
+      // This is done by overriding the high watermark to be the same as the previous watermark.
+      if (!tableNameToEntityMap.keySet().contains(tableName)) {
```
chavdar: tableNameToEntityMap.containsKey(tableName)

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -47,7 +49,7 @@ public Partitioner(SourceState state) {
    * @return map of partition intervals
    */
   public HashMap<Long, Long> getPartitions(long previousWatermark) {
```
chavdar: We have to consider renaming/refactoring this class later. I find the naming rather misleading. Further, it seems to be tied to the query-based extractor yet it has a generic name and namespace.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
zliu41: Added a different property to indicate whether to use config store in the source, default is false.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
zliu41: Ok, only using the entity key.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
zliu41: Issac has written a `WorkUnitBinPacker` class which has some overlap with `KafkaWorkUnitPacker`. We need to merge them (in a separate PR), and `QueryBasedSource` can use it afterwards.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
zliu41: I need to use keyset here since I need to union two maps' keysets.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zliu41: Yes, the rest API classes definitely need some refactoring.

zliu41:
@chavdar @sahilTakiar updated 0b4c9d6. Added logic in `SalesforceSource` to obtain a list of objects from salesforce Rest API.

For this purpose, two new classes were added: `RestApiConnector` and `SalesforceConnector`. Previously the code for connecting to the Rest API is in `RestApiExtractor` and `SalesforceExtractor`, but now since `SalesforceSource` also needs to do that, these code is moved to the new classes.

gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
```diff
@@ -68,8 +68,8 @@ public DatasetsFinder(State state) {
     this.inputDir = getInputDir();
     this.destDir = getDestDir();
     this.tmpOutputDir = getTmpOutputDir();
-    this.blacklist = getBlacklist();
```
sahilTakiar: Are these changes as well as the ones to `MRCompactor` suppose to be in this PR?

zliu41: Due to the new methods in `DatasetFilterUtils` it is possible to simplify the logic in `DatasetsFinder` and `MRCompactor`, so I put them together.

zliu41:
@sahilTakiar updated 47d8a96

sahilTakiar:
LGTM

PR #959, opened 2016-05-04
Remove checksum from HiveDistcp copy listing.

ibuenros:


ibuenros:

PR #960, opened 2016-05-04
Accelerate path diff in HiveCopyEntityHelper by reusing FileStatus.

ibuenros:


ibuenros:

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveLocationDescriptor.java
```diff
@@ -52,14 +55,16 @@
   protected final FileSystem fileSystem;
   protected final Properties properties;
 
-  public Set<Path> getPaths() throws IOException {
+  public Map<Path, FileStatus> getPaths() throws IOException {
 
     PathFindingMethod pathFindingMethod = PathFindingMethod.valueOf(
         this.properties.getProperty(HIVE_LOCATION_LISTING_METHOD, DEFAULT_HIVE_LOCATION_LISTING_METHOD).toUpperCase());
 
+    Map<Path, FileStatus> result = Maps.newHashMap();
```
ydai1124: nit: you can use `BiMap` here.

ydai1124:
Minor comment. LGTM.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveLocationDescriptor.java
ibuenros: I am concerned with the performance hit that a `BiMap` might introduce without clear benefit. Discussed offline.

Issue #882, opened 2016-03-25
Remove current.jst from FsStateStore, add state-store retention to driver


sahilTakiar:
- The `FsStateStore` creates and updates a `current.jst` to track the most recent version of the job state
- The problem is that for AWS users, the state-store typically has be put on S3
- The problem is that for overwriting data, S3 only provides eventual consistency
- This can cause problems as Gobblin jobs will see an old version of the state-store

A simple solution to this problem would be to:
- Remove the concept of `current.jst` and just let each state-store entry be of the form `job_id.jst`
- The gobblin just does a `ls` on the state-store directory, sorts the contents by file name and picks the most recent one
- File listing time + sorting the listing shouldn't take long, but just in case the state-store retention job should be run as part of the Gobblin core job - either in the `ApplicationLauncher` or the `JobLauncher`

sahilTakiar:
@zliu41 I believe we discussed this briefly while working on #741, any comments on the above approach?

zliu41:
LGTM except that if a job has multiple datasets there will be multiple `current.jst`s so you'll need to find the most recent one for each dataset urn.

kadaan:
I've started working on this.

lakshmanantokbox:
If the consistency is turned on in EMR,“consistent view” for EMRFS(https://blogs.aws.amazon.com/bigdata/post/Tx1WL4KR7SE37YY/Ensuring-Consistency-When-Using-Amazon-S3-and-Amazon-Elastic-MapReduce-for-ETL-W), this problem can be avoided

kadaan:
Correct, but for those use Qubole, this is not the case.

PR #965, opened 2016-05-05
Remove need for current.jst

kadaan:
Fix for #882

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -184,6 +184,12 @@ public class ConfigurationKeys {
   public static final String DATASET_URN_KEY = "dataset.urn";
   public static final String DEFAULT_DATASET_URN = "";
 
+  // The unique id of the dataset state file
+  public static final String DATASET_STATE_ID_KEY = "dataset.state.id";
+  public static final String DEFAULT_DATASET_STATE_ID = "";
+
+  public static final String PREVIOUS_STATE_ID_KEY = "previous.state.id";
+
   /**
    * Work unit related configuration properties.
    */
--- a/gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
+++ b/gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
@@ -109,10 +109,10 @@ public class GobblinClusterKillTest {
   private void setupManager(int id) throws Exception {
     _clusterManagers[id] =
         new GobblinClusterManager(TestHelper.TEST_APPLICATION_NAME, TestHelper.TEST_APPLICATION_ID, _config
-                .withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
-                        ConfigValueFactory.fromAnyRef("Manager_" + id))
-                .withValue(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY,
-                        ConfigValueFactory.fromAnyRef(_testDirPath)),
+            .withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
+                ConfigValueFactory.fromAnyRef("Manager_" + id))
+            .withValue(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY,
+                ConfigValueFactory.fromAnyRef(_testDirPath)),
             Optional.of(new Path(_config.getString("gobblin.cluster.work.dir"))));
 
     _clusterManagers[id].start();
@@ -150,7 +150,7 @@ public class GobblinClusterKillTest {
 
     _config = ConfigFactory.parseURL(url)
         .withValue("gobblin.cluster.zk.connection.string",
-                   ConfigValueFactory.fromAnyRef(_testingZKServer.getConnectString()))
+             ConfigValueFactory.fromAnyRef(_testingZKServer.getConnectString()))
         .resolve();
 
     String zkConnectionString = _config.getString(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY);
--- a/gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
+++ b/gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
@@ -22,6 +22,7 @@ import com.google.common.base.Predicate;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import com.typesafe.config.ConfigValueFactory;
+import gobblin.configuration.ConfigurationKeys;
 import gobblin.testing.AssertWithBackoff;
 import java.io.File;
 import java.io.IOException;
@@ -107,9 +108,11 @@ public class GobblinClusterKillTest {
    */
   private void setupManager(int id) throws Exception {
     _clusterManagers[id] =
-        new GobblinClusterManager(TestHelper.TEST_APPLICATION_NAME, TestHelper.TEST_APPLICATION_ID,
-            _config.withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
-                ConfigValueFactory.fromAnyRef("Manager_" + id)),
+        new GobblinClusterManager(TestHelper.TEST_APPLICATION_NAME, TestHelper.TEST_APPLICATION_ID, _config
+                .withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
+                        ConfigValueFactory.fromAnyRef("Manager_" + id))
+                .withValue(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY,
+                        ConfigValueFactory.fromAnyRef(_testDirPath)),
             Optional.of(new Path(_config.getString("gobblin.cluster.work.dir"))));
 
     _clusterManagers[id].start();
--- a/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
@@ -105,10 +105,11 @@ public abstract class FileBasedSource<S, D> extends AbstractSource<S, D> {
 
     // Get list of files seen in the previous run
     if (!previousWorkunits.isEmpty()
-        && previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
-      prevFsSnapshot =
-          previousWorkunits.get(0).getWorkunit().getPropAsSet(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
-    }
+      state.setProp(ConfigurationKeys.PREVIOUS_STATE_ID_KEY, previousWorkunits.get(0)
+	      .getProp(ConfigurationKeys.DATASET_STATE_ID_KEY, ConfigurationKeys.DEFAULT_DATASET_STATE_ID));
+      if (previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
+        previousWorkunits.get(0).getWorkunit().getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
+      }
 
     List<WorkUnit> workUnits = Lists.newArrayList();
     List<WorkUnit> previousWorkUnitsForRetry = this.getPreviousWorkUnitsForRetry(state);
--- a/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
@@ -104,7 +104,7 @@ public abstract class FileBasedSource<S, D> extends AbstractSource<S, D> {
     Set<String> prevFsSnapshot = Sets.newHashSet();
 
     // Get list of files seen in the previous run
-    if (!previousWorkunits.isEmpty()
+    if (!previousWorkunits.isEmpty()) {
       state.setProp(ConfigurationKeys.PREVIOUS_STATE_ID_KEY, previousWorkunits.get(0)
 	      .getProp(ConfigurationKeys.DATASET_STATE_ID_KEY, ConfigurationKeys.DEFAULT_DATASET_STATE_ID));
       if (previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
--- a/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
@@ -110,6 +110,7 @@ public abstract class FileBasedSource<S, D> extends AbstractSource<S, D> {
       if (previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
         previousWorkunits.get(0).getWorkunit().getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
       }
+	}
 
     List<WorkUnit> workUnits = Lists.newArrayList();
     List<WorkUnit> previousWorkUnitsForRetry = this.getPreviousWorkUnitsForRetry(state);
--- a/gobblin-metastore/src/main/java/gobblin/metastore/DatasetStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/DatasetStateStore.java
@@ -24,15 +24,14 @@ import java.util.Map;
 
 public interface DatasetStateStore<T extends State> extends StateStore<T> {
   String DATASET_STATE_STORE_TABLE_SUFFIX = ".jst";
-  String CURRENT_DATASET_STATE_FILE_SUFFIX = "current";
 
   interface Factory {
     <T extends State> DatasetStateStore<T> createStateStore(Config config);
   }
 
-  public Map<String, T> getLatestDatasetStatesByUrns(String jobName) throws IOException;
+  Map<String, T> getLatestDatasetStatesByUrns(String jobName) throws IOException;
 
-  public T getLatestDatasetState(String storeName, String datasetUrn) throws IOException;
+  T getLatestDatasetState(String storeName, String datasetUrn) throws IOException;
 
-  public void persistDatasetState(String datasetUrn, T datasetState) throws IOException;
+  void persistDatasetState(String datasetUrn, T datasetState) throws IOException;
 }
--- a/gobblin-metastore/src/main/java/gobblin/metastore/DatasetStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/DatasetStateStore.java
@@ -17,10 +17,14 @@
 
 package gobblin.metastore;
 
-import com.typesafe.config.Config;
-import gobblin.configuration.State;
 import java.io.IOException;
 import java.util.Map;
+
+import com.google.common.base.Optional;
+import com.typesafe.config.Config;
+
+import gobblin.configuration.State;
+
 
 public interface DatasetStateStore<T extends State> extends StateStore<T> {
   String DATASET_STATE_STORE_TABLE_SUFFIX = ".jst";
@@ -32,6 +36,8 @@ public interface DatasetStateStore<T extends State> extends StateStore<T> {
   Map<String, T> getLatestDatasetStatesByUrns(String jobName) throws IOException;
 
   T getLatestDatasetState(String storeName, String datasetUrn) throws IOException;
+
+  Map<Optional<String>, String> getLatestDatasetStateTablesByUrn(String jobName) throws IOException;
 
   void persistDatasetState(String datasetUrn, T datasetState) throws IOException;
 }
--- a/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
@@ -24,7 +24,6 @@ import java.net.URI;
 import java.util.Collection;
 import java.util.List;
 
-import org.apache.commons.io.FilenameUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
@@ -43,9 +42,9 @@ import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
 import gobblin.configuration.State;
+import gobblin.metastore.util.StateStoreUtils;
 import gobblin.metastore.util.StateStoreTableInfo;
 import gobblin.util.HadoopUtils;
-import gobblin.util.Id;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -397,30 +396,10 @@ public class FsStateStore<T extends State> implements StateStore<T> {
     String latestStateFilePath = null;
     for (FileStatus fileStatus : stateStoreFileStatuses) {
       String tableName = fileStatus.getPath().getName();
-      if (latestStateFilePath == null) {
-        log.debug("Latest table for {}/{}* set to {}", storeName, tableNamePrefix, tableName);
-        latestStateFilePath = tableName;
-      } else {
-        Id latestId = getId(tableNamePrefix, latestStateFilePath);
-        Id newId = getId(tableNamePrefix, tableName);
-        if (newId.getSequence().compareTo(latestId.getSequence()) > 0) {
-          log.debug("Latest table for {}/{}* set to {} instead of {}", storeName, tableNamePrefix, tableName, latestStateFilePath);
-          latestStateFilePath = tableName;
-        } else {
-          log.debug("Latest table for {}/{}* left as {}. Previous table {} is being ignored", storeName, tableNamePrefix, latestStateFilePath, tableName);
-        }
-      }
+      latestStateFilePath = StateStoreUtils.getLatestStateKey(storeName, tableNamePrefix, tableName, latestStateFilePath);
     }
 
     return latestStateFilePath == null ? null : new Path(new Path(this.storeRootDir, storeName), latestStateFilePath);
-  }
-
-  private Id getId(String tableNamePrefix, String tableName) {
-    String id = FilenameUtils.removeExtension(tableName);
-    if (!Strings.isNullOrEmpty(tableNamePrefix)) {
-      id = id.substring(tableNamePrefix.length() + StateStoreTableInfo.TABLE_PREFIX_SEPARATOR.length() + 1);
-    }
-    return Id.parse(id);
   }
 
   private boolean isCurrentFile(String tableName) {
--- a/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
@@ -24,6 +24,7 @@ import java.net.URI;
 import java.util.Collection;
 import java.util.List;
 
+import org.apache.commons.io.FilenameUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
@@ -44,6 +45,7 @@ import com.google.common.io.Closer;
 import gobblin.configuration.State;
 import gobblin.metastore.util.StateStoreTableInfo;
 import gobblin.util.HadoopUtils;
+import gobblin.util.Id;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -398,15 +400,27 @@ public class FsStateStore<T extends State> implements StateStore<T> {
       if (latestStateFilePath == null) {
         log.debug("Latest table for {}/{}* set to {}", storeName, tableNamePrefix, tableName);
         latestStateFilePath = tableName;
-      } else if (tableName.compareTo(latestStateFilePath) > 0) {
-        log.debug("Latest table for {}/{}* set to {} instead of {}", storeName, tableNamePrefix, tableName, latestStateFilePath);
-        latestStateFilePath = tableName;
       } else {
-        log.debug("Latest table for {}/{}* left as {}. Previous table {} is being ignored", storeName, tableNamePrefix, latestStateFilePath, tableName);
+        Id latestId = getId(tableNamePrefix, latestStateFilePath);
+        Id newId = getId(tableNamePrefix, tableName);
+        if (newId.getSequence().compareTo(latestId.getSequence()) > 0) {
+          log.debug("Latest table for {}/{}* set to {} instead of {}", storeName, tableNamePrefix, tableName, latestStateFilePath);
+          latestStateFilePath = tableName;
+        } else {
+          log.debug("Latest table for {}/{}* left as {}. Previous table {} is being ignored", storeName, tableNamePrefix, latestStateFilePath, tableName);
+        }
       }
     }
 
     return latestStateFilePath == null ? null : new Path(new Path(this.storeRootDir, storeName), latestStateFilePath);
+  }
+
+  private Id getId(String tableNamePrefix, String tableName) {
+    String id = FilenameUtils.removeExtension(tableName);
+    if (!Strings.isNullOrEmpty(tableNamePrefix)) {
+      id = id.substring(tableNamePrefix.length() + StateStoreTableInfo.TABLE_PREFIX_SEPARATOR.length() + 1);
+    }
+    return Id.parse(id);
   }
 
   private boolean isCurrentFile(String tableName) {
--- a/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
@@ -19,8 +19,6 @@ package gobblin.metastore;
 
 import static gobblin.util.HadoopUtils.FS_SCHEMES_NON_ATOMIC;
 
-import com.google.common.base.Predicate;
-import gobblin.util.HadoopUtils;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
@@ -30,16 +28,25 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.io.SequenceFile;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.compress.DefaultCodec;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.io.Closer;
 
 import gobblin.configuration.State;
+import gobblin.metastore.util.StateStoreTableInfo;
+import gobblin.util.HadoopUtils;
+
+import lombok.extern.slf4j.Slf4j;
+
 
 /**
  * An implementation of {@link StateStore} backed by a {@link FileSystem}.
@@ -58,6 +65,7 @@ import gobblin.configuration.State;
  *
  * @author Yinan Li
  */
+@Slf4j
 public class FsStateStore<T extends State> implements StateStore<T> {
 
   public static final String TMP_FILE_PREFIX = "_tmp_";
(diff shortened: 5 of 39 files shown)


kadaan:
@sahilTakiar @zliu41: Can you review this?

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
```diff
@@ -92,14 +92,17 @@ protected void initLogger(SourceState state) {
...
     // Get list of files seen in the previous run
-    if (!previousWorkunits.isEmpty()
-        && previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
-      prevFsSnapshot =
-          previousWorkunits.get(0).getWorkunit().getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
+    if (previousSourceState != null) {
+      state.setProp(ConfigurationKeys.PREVIOUS_STATE_ID_KEY,
+          previousSourceState.getProp(ConfigurationKeys.DATASET_STATE_ID_KEY, ConfigurationKeys.DEFAULT_DATASET_STATE_ID));
+      if (previousSourceState.contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
+        prevFsSnapshot = previousSourceState.getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
```
zliu41: I don't think this works - `DatasetState` (which is `previousSourceState`'s class) no longer supports `getProp` since #932.
Even if it does, I don't understand why this property belongs to `previousSourceState` - it seems it is added to workunits, not the dataset state.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -508,8 +509,8 @@ protected void setup(Context context) {
       try(Closer closer = Closer.create()) {
         this.fs = FileSystem.get(context.getConfiguration());
         this.taskStateStore =
-            new FsStateStore<>(this.fs, FileOutputFormat.getOutputPath(context).toUri().getPath(), TaskState.class);
-
+            new FsStateStore<>(this.fs, SequenceFileOutputFormat.getOutputPath(context).toUri().getPath(),
```
zliu41: Is this needed? I think `SequenceFileOutputFormat` inherits this method from `FileOutputFormat`.

kadaan: I will fix this.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
kadaan: I think this is still needed, so I would love a suggestion on how to accomplish this.  The reason I want this is that it makes it *very* easy to diagnose issues where jobs are running off of the same previous state.  This scenario is particularly insidious as it results in duplicated data.

zliu41:
@kadaan could you please give a brief description of your approach? It seems you are still using `current.jst`, which is a different approach than #882.

kadaan:
`current.jst` is not used.  There is a compromise here so that users of the API aren't broken.  New callers can call `getCurrent` or `getAllCurrent` to get the latest state.  If they want a specific state they can continue to call `get` or `getAll`.  If `current` or `current.jst` is requested when calling `get` or `getAll` it will  return the latest state just like `getCurrent` and `getAllCurrent`.  A precondition will ensure that users of the API are not able to write a file named `current` or `current.jst`.

kadaan:
@zliu41 @sahilTakiar Can you guys finish this review?

kadaan:
@zliu41 @sahilTakiar Can you guys finish this review?

zliu41:
@pcadabam @ibuenros @ydai1124 can you review this PR? Thanks

kadaan:
@pcadabam @ibuenros @ydai1124 Were any of you able to review this?

kadaan:
@abti Can you check this out now?

gobblin-runtime/src/main/java/gobblin/runtime/FsDatasetStateStore.java
```diff
@@ -254,16 +246,47 @@ public void persistDatasetState(String datasetUrn, JobState.DatasetState dataset
...
 
-    datasetUrn = CharMatcher.is(':').replaceFrom(datasetUrn, '.');
-    String tableName = Strings.isNullOrEmpty(datasetUrn) ? jobId + DATASET_STATE_STORE_TABLE_SUFFIX
-        : datasetUrn + "-" + jobId + DATASET_STATE_STORE_TABLE_SUFFIX;
+    Optional<String> sanitizedDatasetUrn = DatasetUrnSanitizer.sanitize(datasetUrn);
+    String tableName = sanitizedDatasetUrn.isPresent() ?
+        sanitizedDatasetUrn.get() + StateStoreTableInfo.TABLE_PREFIX_SEPARATOR + jobId : jobId;
     LOGGER.info("Persisting " + tableName + " to the job state store");
     put(jobName, tableName, datasetState);
-    createAlias(jobName, tableName, getAliasName(datasetUrn));
```
ydai1124: What will happen to the existing `current.jst`? The current implementation (before this PR) is to overwrite `current.jst` every time. With the new changes, they will not be deleted. Will that cause any confusion when loading the latest state?

gobblin-metastore/src/main/java/gobblin/metastore/DatasetStateStore.java
```diff
@@ -21,6 +21,7 @@
 import java.io.IOException;
 import java.util.Map;
 import java.util.Properties;
+import java.util.regex.Pattern;
```
ydai1124: Please remove this.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
```diff
@@ -127,6 +127,18 @@ public T get(String storeName, String tableName, String stateId)
...
   /**
+   * Get a {@link State} with a given state ID from a the current table.
+   *
+   * @param storeName store name
+   * @param stateId state ID
+   * @return {@link State} with the given state ID or <em>null</em>
+   *         if the state with the given state ID does not exist
+   * @throws IOException
+   */
+  public T getCurrent(String storeName, String stateId)
```
ydai1124: Are the two new `getCurrent...` methods really necessary? Plus, we are moving towards `DatasetStateStore`, and it has the access to the latest dataset state.

gobblin-runtime/src/main/java/gobblin/runtime/FsDatasetStateStore.java
kadaan: `current.jst` will not be written to or read from after this change. States are always written to new files and when `current.jst` is requested from the state store, the latest file will be found and returned.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
kadaan: If I remember correctly, the reason for the `getCurrent` vs `getAllCurrent` is that it is mirroring the `get` and `getAll` methods which took table name.

DatasetStateStore has existed for more than a year and this change was compatible with how it worked.  As part of rebasing this changed onto the latest master, I made the new implementations of state store work without a `current` state.

kadaan:
@ydai1124 Can you please review again?

gobblin-runtime/src/main/java/gobblin/runtime/FsDatasetStateStore.java
ydai1124: I see. I thought the `statestore-checker.sh` would be broken, but seems you updated it as well. Thanks.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
ydai1124: I will leave this API change to @htran1  since he is also making some statestore API changes. The rest part LGTM.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
```diff
@@ -64,12 +94,57 @@ public MysqlDatasetStateStore(DataSource dataSource, String stateStoreTableName,
    * @throws IOException if there's something wrong reading the {@link JobState.DatasetState}s
    */
   public Map<String, JobState.DatasetState> getLatestDatasetStatesByUrns(String jobName) throws IOException {
-    List<JobState.DatasetState> previousDatasetStates =
-        getAll(jobName, "%" + CURRENT_DATASET_STATE_FILE_SUFFIX + DATASET_STATE_STORE_TABLE_SUFFIX, true);
+    final String storeName = jobName;
```
htran1: This exposes the querying logic to the MysqlDatasetStateStore and also results in duplication of logic in the result processor.

gobblin-runtime/src/test/java/gobblin/runtime/local/LocalJobLauncherTest.java
```diff
@@ -53,14 +54,14 @@
 
   @BeforeClass
   public void startUp() throws Exception {
-    testMetastoreDatabase = TestMetastoreDatabaseFactory.get();
+    //testMetastoreDatabase = TestMetastoreDatabaseFactory.get();
     this.launcherProps = new Properties();
     this.launcherProps.load(new FileReader("gobblin-test/resource/gobblin.test.properties"));
-    this.launcherProps.setProperty(ConfigurationKeys.JOB_HISTORY_STORE_ENABLED_KEY, "true");
```
htran1: Why is the job history store related items being removed?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -186,6 +186,12 @@
   public static final String DATASET_URN_KEY = "dataset.urn";
   public static final String DEFAULT_DATASET_URN = "";
 
+  // The unique id of the dataset state file
+  public static final String DATASET_STATE_ID_KEY = "dataset.state.id";
+  public static final String DEFAULT_DATASET_STATE_ID = "";
+
+  public static final String PREVIOUS_STATE_ID_KEY = "previous.state.id";
```
kadaan: This is for record keeping, so you can know which was the previous state.

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -73,38 +81,65 @@
...
+   * @param tableName table name
+   * @return if the table is successfully created
+   * @throws IOException
+   */
   @Override
   public boolean create(String storeName, String tableName) throws IOException {
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(storeName), "Store name is null or empty.");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(tableName), "Table name is null or empty.");
+    Preconditions.checkArgument(!isCurrentFile(tableName),
+        String.format("Table name is %s", StateStoreTableInfo.CURRENT_NAME));
```
kadaan: The problem is that without putting an understanding of "CURRENT" in the state store, all calling code will have to change.  This way, any calling code can as for the file named "CURRENT" and they get the latest one.

gobblin-runtime/src/test/java/gobblin/runtime/local/LocalJobLauncherTest.java
kadaan: That was mistakenly removed and will be added back.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
kadaan: I can reduce the duplication between MysqlStateStore and MysqlDatasetStateStore pretty easily.  Can you explain further your problem with exposing the querying logic to MysqlDatasetStateStore?  I don't see any easy way to get the latest dataset states broken out by Urn, unless we allow the MysqlDatasetStateStore  to do the query or push that functionality into MysqlStateStore and allow MysqlDatasetStateStore to delegate the call.

kadaan:
@htran1 In response to your general comment:

> General comment is that StateStore should not know about 'current'...
> The StateStore should exposed APIs used at the DatasetStateStore level to access the last modified state.

The StateStore does expose an API to access current: `getCurrent` and `getAllCurrent`.  Additionally, the `createAlias` API has been removed.  In order to support a seamless upgrade, StateStore, also supports the prior `get*` calls being called with a filename like `current`.  In this case it will perform the same as the `getCurrent`/`getAllCurrent` calls.  The logic behind this is that I cannot be sure that there are no dependencies from third-party libraries on the existing `get*` calls.  By making the change in this way, we can get rid of the `current` aliased files, while still ensuring existing code will work. We can publish in release note that callers need to upgrade to the latest API.

> ... DatasetStateStore should not query directly.

If desired, the logic of `MysqlDatasetStateStore.getLatestDatasetStatesByUrns` can be pushed down into `MysqlStateStore`, if you feel strongly about that.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
htran1: Mainly with the duplication. If there is no interface at the StateStore level that makes sense then having the query logic at the dataset level is fine, but the existing implementation calls getAll() with a pattern. Can't a similar API be provided that will get all items with the pattern and having the most recent modification time?

kadaan: I've removed the duplication with the latest commit.

kadaan: The issue I was facing was that the actual query needs to be different because we are trying to get the latest file for each Urn and the Urn is part of the name.  So, unless we do it in the query, we need to get all files, group by the Urn, and take the latest.

kadaan:
@htran1 Can you review this again?

kadaan:
@htran1 I don't think that test failure is due to my PR.  It seems to be a problem with line 197 in `GobblinClusterKillTest`: ` _clusterWorkers[0].disconnectHelixManager();`.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
```diff
@@ -51,9 +70,20 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(MysqlDatasetStateStore.class);
 
+  private static final String SELECT_JOB_STATE_LATEST_BY_PREFIX_TEMPLATE =
+          "SELECT state, CASE WHEN locate(?, table_name) = 0 THEN '' "
```
htran1: I don't think we have a restriction on the dataset URN not having '-', so not sure if this is a reliable way to parse the name.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -103,6 +104,7 @@ public ServiceBasedAppLauncher(Properties properties, String appName) throws Exc
     this.services = new ArrayList<>();
 
     // Add core Services needed for any application
+    addStateStoreCleaner(properties);
```
htran1: Add the service only if the state store is enabled.

gobblin-runtime/src/test/java/gobblin/test/TestDataPublisher.java
```diff
@@ -30,7 +30,7 @@
  * An implementation of {@link DataPublisher} for integration test.
  *
  * <p>
- *     This is a dummy implementation that exists purely to make
+ *     This is a dummy implementation that storeExists purely to make
```
htran1: Changed by mistake?

gobblin-runtime/src/test/java/gobblin/runtime/local/LocalJobLauncherTest.java
```diff
@@ -21,6 +21,7 @@
 import java.io.IOException;
 import java.util.Properties;
 
+import gobblin.runtime.FsDatasetStateStore;
```
htran1: Not needed?

gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
```diff
@@ -107,9 +108,11 @@ private void setupTestDir() throws IOException {
    */
   private void setupManager(int id) throws Exception {
     _clusterManagers[id] =
-        new GobblinClusterManager(TestHelper.TEST_APPLICATION_NAME, TestHelper.TEST_APPLICATION_ID,
-            _config.withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
-                ConfigValueFactory.fromAnyRef("Manager_" + id)),
+        new GobblinClusterManager(TestHelper.TEST_APPLICATION_NAME, TestHelper.TEST_APPLICATION_ID, _config
+                .withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY,
+                        ConfigValueFactory.fromAnyRef("Manager_" + id))
+                .withValue(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY,
```
kadaan: Good call.  I'm not sure how to tell if the state store is enabled. Ideally, each state store implementation should have a corresponding cleaner that would be automatically included.  I think for now it would be best to just not automatically start the state store cleaner.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
```diff
@@ -103,10 +103,12 @@ protected void initLogger(SourceState state) {
     List<String> prevFsSnapshot = Lists.newArrayList();
 
     // Get list of files seen in the previous run
-    if (!previousWorkunits.isEmpty()
-        && previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
-      prevFsSnapshot =
-          previousWorkunits.get(0).getWorkunit().getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
+    if (previousWorkunits != null) {
```
kadaan: Yes

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -134,28 +180,13 @@ public boolean exists(String storeName, String tableName) throws IOException {
    */
   @Override
   public void put(String storeName, String tableName, T state) throws IOException {
-    String tmpTableName = this.useTmpFileForPut ? TMP_FILE_PREFIX + tableName : tableName;
-    Path tmpTablePath = new Path(new Path(this.storeRootDir, storeName), tmpTableName);
-
-    if (!this.fs.exists(tmpTablePath) && !create(storeName, tmpTableName)) {
-      throw new IOException("Failed to create a state file for table " + tmpTableName);
-    }
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(storeName), "Store name is null or empty.");
```
kadaan: I'll leave it.

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -118,13 +153,24 @@ public boolean create(String storeName, String tableName) throws IOException {
...
-    return this.fs.exists(tablePath);
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(storeName), "Store name is null or empty.");
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(tableName), "Table name is null or empty.");
+
+    Path tablePath = getTablePath(storeName, tableName);
+    return tablePath != null && this.fs.exists(tablePath);
   }
 
-  /**
+    /**
```
kadaan: I'll fix that.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -192,6 +222,25 @@ public T get(String storeName, String tableName, String stateId) throws IOExcept
   }
 
   /**
+   * Get a {@link State} with a given state ID from a the current table.
```
kadaan: I feel like the goal would be to transition all calls to getCurrent/getAllCurrent and deprecate the "current" name.  Would you prefer to stick with the "current" name and not have the more explicit api?

kadaan: One thing I like about your suggestion is that there is less api change.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -321,4 +365,44 @@ private void deserialize(byte[] data, List<T> states, String stateId) throws IOE
...
+    }
+
+    String latestStatePath = null;
+    for (String c : children) {
+      if ((Strings.isNullOrEmpty(tableNamePrefix) ||
+        c.startsWith(tableNamePrefix + StateStoreTableInfo.TABLE_PREFIX_SEPARATOR))) {
+        if (latestStatePath == null) {
+          log.debug("Latest table for {}/{}* set to {}", storeName, tableNamePrefix, c);
+          latestStatePath = c;
+        } else if (c.compareTo(latestStatePath) > 0) {
```
kadaan: It would be better if the sequence number in the task id was padded, otherwise we have to parse the string every time.

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreCleaner.java
```diff
@@ -82,48 +90,79 @@ public StateStoreCleaner(Properties properties) throws IOException {
...
 
     URI fsUri = URI.create(properties.getProperty(
         ConfigurationKeys.STATE_STORE_FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
-    this.fs = FileSystem.get(fsUri, new Configuration());
+    this.fs = FileSystem.get(fsUri, HadoopUtils.getConfFromProperties(properties));
+  }
+
+  @Override
+  protected void startUp() throws Exception {
+    log.info("Starting the " + StateStoreCleaner.class.getSimpleName());
```
kadaan: This was not intended, but a side effect of trying to handle 8 different year old PRs at once.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
kadaan: I am going to remove this until state store cleaner works for all types of state stores.

gobblin-runtime/src/test/java/gobblin/test/TestDataPublisher.java
kadaan: Thought I fixed all of those. Good catch.

gobblin-runtime/src/test/java/gobblin/runtime/local/LocalJobLauncherTest.java
kadaan: Nope, not needed.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
kadaan: Well, if that is the case, this is really tricky.  I think we need to figure out how to deal with this as I believe that using the "current" alias causes a lot of trouble, especially since creating the state and the alias are not an atomic operation.  This can lead to scenarios where the state has been updated and the alias has not been created. Adding a restriction on the dataset URN might be hard because it is already being used and we don't know whether there are values that contain `-`.

@htran1  Do you have any suggestions or ideas?

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreTableInfo.java
```diff
@@ -0,0 +1,41 @@
...
+  private boolean isCurrent;
+
+  public static StateStoreTableInfo get(String tableName) {
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(tableName), "Table name is null or empty.");
+    String name = FilenameUtils.getBaseName(tableName);
+    if (CURRENT_NAME.equalsIgnoreCase(name)) {
+      return new StateStoreTableInfo("", true);
+    }
+    int suffixIndex = name.lastIndexOf(TABLE_PREFIX_SEPARATOR);
+    if (suffixIndex >= 0) {
```
kadaan: Sure

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
kadaan: @htran1 Suggestions?

gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
kadaan: I've removed the addition of the state store cleaner, so this should be solved.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
kadaan: @htran1 I think I have a solution for this issue.  From what I can tell, the `getLatestDatasetStateFilePathsByUrns` is always passed the `jobName` as the `storeName`.  Additionally, we know that `tableName` will be in one of the following two formats:

1. `job_{{ NAME }}_{{ SEQUENCE }}.{{ EXTENSION }}`
2. `{{ URN }}-job_{{ NAME }}_{{ SEQUENCE }}.{{ EXTENSION }}`

What I propose is we look at the table name for the index of `-job_{{ NAME }}`.  In the case where there is no Urn, then it will be at `< 0`. In this case, the Urn will be set to `""`.  Otherwise, it will be `>= 0` and the Urn will be set to `tableName.substring(0, jobPrefixIndex)`.

This should correctly handle the case where the Urn contains characters such as a `-`.  Additionally, this logic should be able to be done in the Mysql query as well.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
kadaan: @htran1 I also think I have a reasonable solution for this.  Given we know the Urn, we can strip that off the front of the `tableName`, remove the extension, and then parse the remaining with the well known Id format.  Then we can directly compare the sequence numbers to see which is the latest.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
htran1: I discussed with @ibuenros and he said there have been discussions about reorganizing the dataset state into subdirectories. That would remove the need for parsing. He also raised concerns about the cost of listing vs being able to read the file directly.

One suggestions to avoid and backward compatibility and performance issue is to support both 'alias-based' and 'time-based' state stores either by abstracting out a new state store type or having it based on configuration within the existing state stores. That approach would leave createAlias() and use it if configured.

kadaan: I would say that we have seen no performance issues with listing vs accessing via an alias and we have been running largely similar code for over a year.  Many of these concerns were discussed and dismissed when #882 was proposed.  Can you also talk to @sahilTakiar about this?

I feel like including both 'alias-based' and 'time-based' complicates the system and will lead to an explosion on state stores.  I think we should try to reduce options where possible so that the system stays clear and maintainable.  Additionally, I believe that the `alias-based` approach is problematic because creating the state and the alias are not an atomic operation.  Additionally, when using S3 as a state store, updates eventual which causes serious problems because all changes to the `current` file are updates.  The effect the old state is used and data is duplicated.

htran1: I think the parsing with -job_ added will make it more reliable. I think we should leave support for 'current.jst' in as an option.

gobblin-cluster/src/test/java/gobblin/cluster/GobblinClusterKillTest.java
kadaan: @htran1 @ibuenros @sahilTakiar StateStoreCleaner desperately needs to support all state store types.  Once it does, we can add this back.

gobblin-runtime/src/main/java/gobblin/runtime/MysqlDatasetStateStore.java
kadaan: @htran1 `current.jst` is an option when calling `get` but we need to move away from writing a `current.jst` entry.

kadaan: I think we can change the `MysqlStateStore.SELECT_JOB_STATE_LATEST_TEMPLATE` query to: 
``` sql
SELECT state 
FROM $TABLE$ WHERE store_name = ?
ORDER BY CAST(REVERSE(SUBSTRING_INDEX(REVERSE(table_name), '_', 1)) AS UNSIGNED) DESC
LIMIT 1
```

This has the benefit of just taking the segment of the table name that appears after the last `_`.  While this may include an extension, the `CAST` correctly ignores it. Then we can order by the sequence number and take the latest.

As far as performance, running this query against my job history db on a task with 168k entries in the DB, I'm able to get a result in 400ms.  I think that is adequate performance for this scenario as this is not called in a tight loop and other tasks will dominate the processing time.

Now, as for the `MysqlDatasetStateStore.SELECT_JOB_STATE_LATEST_BY_PREFIX_TEMPLATE`, I think we can do something like the following:

``` sql
SELECT
    state,
    CASE WHEN locate(?, table_name) = 0 THEN '' ELSE SUBSTRING_INDEX(table_name, ?, 1) END AS urn
FROM $TABLE$ t1
INNER JOIN (
    SELECT
        CASE WHEN locate(?, table_name) = 0 THEN '' ELSE SUBSTRING_INDEX(table_name, ?, 1) END AS urn,
        MAX(CAST(REVERSE(SUBSTRING_INDEX(REVERSE(table_name), ?, 1)) AS UNSIGNED)) AS sequence
    FROM $TABLE$
    WHERE store_name = ?
    GROUP BY urn
) t2 ON CASE WHEN locate(?, table_name) = 0 THEN '' ELSE SUBSTRING_INDEX(table_name, ?, 1) END = t2.urn
     AND CAST(REVERSE(SUBSTRING_INDEX(REVERSE(table_name), ?, 1)) AS UNSIGNED) = t2.sequence
WHERE store_name = ?
```

This has a similar behavior to the above query.

htran1: Instead of specific queries, how about reusing the logic from FS and ZK and just grab out the names then filter and grab the values? If performance is reasonable then we avoid bugs.

kadaan: Sure.  I can do that.

kadaan:
@htran1 Can you review this again?

kadaan:
@htran1 Can you please review this again?

gobblin-runtime/src/test/java/gobblin/runtime/api/TestJobExecutionState.java
```diff
@@ -44,7 +47,10 @@
 
   @Test public void testStateTransitionsSuccess() throws TimeoutException, InterruptedException {
     final Logger log = LoggerFactory.getLogger(getClass().getSimpleName() + ".testStateTransitionsSuccess");
-    JobSpec js1 = JobSpec.builder("gobblin:/testStateTransitionsSuccess/job1").build();
+    JobSpec js1 = JobSpec.builder("gobblin:/testStateTransitionsSuccess/job1")
```
htran1: I think requiring a job name in the test is okay since we check for it in other places, but want to double check with @ibuenros.

kadaan:
@htran1 Any more feedback?

kadaan:
@ibuenros Any feedback?

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Slf4j
+public class StateStoreUtils {
+  private StateStoreUtils() {
+  }
+
+  public static String getLatestStateKey(String storeName, String urn, String tableName, String latestTableName) {
+    if (latestTableName == null) {
+      log.debug("Latest table for {}/{}* set to {}", storeName, urn, tableName);
+        latestTableName = tableName;
```
htran1: Indentation.

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+
+@Slf4j
+public class StateStoreUtils {
+  private StateStoreUtils() {
+  }
+
+  public static String getLatestStateKey(String storeName, String urn, String tableName, String latestTableName) {
```
htran1: Please add javadoc for this method.

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    if (latestTableName == null) {
+      log.debug("Latest table for {}/{}* set to {}", storeName, urn, tableName);
+        latestTableName = tableName;
+    } else {
+      Id latestId = getId(urn, latestTableName);
+      Id newId = getId(urn, tableName);
+      if (newId.getSequence().compareTo(latestId.getSequence()) > 0) {
+        log.debug("Latest table for {}/{}* set to {} instead of {}", storeName, urn, tableName, latestTableName);
+          latestTableName = tableName;
```
htran1: Indentation.

gobblin-metastore/src/main/java/gobblin/metastore/util/StateStoreUtils.java
```diff
@@ -0,0 +1,58 @@
...
+      } else {
+        log.debug("Latest table for {}/{}* left as {}. Previous table {} is being ignored", storeName, urn, latestTableName, tableName);
+      }
+    }
+    return latestTableName;
+  }
+
+  public static Id getId(String urn, String tableName) {
```
htran1: Add javadoc.

gobblin-runtime/src/main/java/gobblin/runtime/util/DatasetStateStoreUtils.java
```diff
@@ -0,0 +1,145 @@
...
+    // The dataset (job) state from the deprecated "current.jst" will be read even though
+    // the job has transitioned to the new dataset-based mechanism
+    if (datasetStatesByUrns.size() > 1) {
+      datasetStatesByUrns.remove(ConfigurationKeys.DEFAULT_DATASET_URN);
+    }
+
+    return datasetStatesByUrns;
+  }
+
+  public static Map<Optional<String>, String> getLatestDatasetStateTablesByUrn(
```
htran1: Javadoc.

gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
```diff
@@ -104,11 +104,13 @@ protected void initLogger(SourceState state) {
...
-    if (!previousWorkunits.isEmpty()
-        && previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
-      prevFsSnapshot =
-          previousWorkunits.get(0).getWorkunit().getPropAsSet(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
-    }
+    if (!previousWorkunits.isEmpty()) {
+      state.setProp(ConfigurationKeys.PREVIOUS_STATE_ID_KEY, previousWorkunits.get(0)
+	      .getProp(ConfigurationKeys.DATASET_STATE_ID_KEY, ConfigurationKeys.DEFAULT_DATASET_STATE_ID));
+      if (previousWorkunits.get(0).getWorkunit().contains(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT)) {
+        previousWorkunits.get(0).getWorkunit().getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT);
```
kadaan: Grrr.  This is a pin of a PR to keep on top of all the changes and make sure nothing get screwed up.  I'll fix that.

kadaan:
@htran1 I understand what you are saying, but I believe that the current.jst is the cause for many problems. It was discussed a year ago and recommended to be removed. One big issue is that depending on the backing FS, the update of the current file contents is "eventual".  This can lead to incorrect state being returned.

Regarding `#1` above, do you have any numbers that show what the performance is for listing the state files?  Also, what state store implementation are you using.  If it is the FS version, what backing FS is it?

For `#2` I would recommend that we have a different mechanism for moving to a prior state.  This is a good feature, but rewriting the current file seems to be the wrong was to go about this.

As for `#3`, it seems that the need to write a current pointer or not is totally dependent on the filesystem which is used to store the state.  Why not get rid of the concept of current outside of the state store and allow implementations to use it or not.  At least that way, I can make an implementation of FsStateStore that doesn't use current and the builtin version can.

htran1:
For 1, we are using FS state store on HDFS. I'm less concerned about this case than case 3 on NFS. As long as we retain only a small number of jst files per dataset it will probably be okay since it will probably add at most a few seconds to the startup of a job.

For 2, I'll have to check with the team since this is the current method of temporarily moving state back.

3 is a concern for us since the NFS system we are on is less scalable than HDFS. I'm seeing delete of small files giving only 50 ops/s. Listing should be faster, but with many concurrent jobs starting it may saturate our NFS filer. We only have a single master node in this mode that does job planning, so saving CPU cycles is more essential.

The base state store interface doesn't know about current.jst (at least before your changes). We have other uses of StateStore that stores non .jst files.  The current.jst is a DatasetStateStore concept. So maybe it makes more sense to add the functionality you need to a new variation of FsDatasetStateStore that doesn't make use of the current file. I think you can even add it as a config option to the current FsDatasetStateStore since it should only be a few blocks that are different. In either case, the mysql and zk dataset state stores won't need to be touched. This allows us to continue using current.jst while users of S3 can use the new approach.

abti:
https://issues.apache.org/jira/browse/GOBBLIN-109 

Please update your PR title with following prefix: [GOBBLIN-109]

PR #967, opened 2016-05-06
Set the job.id and task.id in the MDC for improved logging

kadaan:


kadaan:
@sahilTakiar This removes any compile dependencies to log4j and replaces them with compile dependencies to slf4j and runtime dependencies on logback.  Jobs, Forks, and Tasks now add the job name, job id, and task id to slf4j MDC.  To facilitate this information flowing correctly across threads, Callable and Runnable have been replaced with GobblinCallable and GobblinRunnable.

sahilTakiar:
Can you document why switching to logback is necessary / what its advantages are?

Also, do we need to replace every `Runnable` and `Callable` instance with `GobblinRunnable` and `GobblinCallable` - it is possible to instead add this instrumentation to `ExecutorsUtils.newThreadFactory` - we can have a custom `ThreadFactory` that does all this for us.

kadaan:
The reasons to prefer logback over log4j are [here](http://logback.qos.ch/reasonsToSwitch.html).  The biggest one is the logback implements the SLF4J API natively.  This provides better performance.

Yes, we can have a custom thread factory, but I will have to figure out how to do it.

kadaan:
@sahilTakiar Any comments about this?

kadaan:
@sahilTakiar Any comments about this?

sahilTakiar:
I won't have time to review this, until this weekend.

@chavdar care to take a look?

sahilTakiar:
I still think the `GobblinRunnable` and `GobblinCallable` aren't necessary (its also going to be hard to enforce use of these classes throughout the code base.), moving the logic to `ExecutorsUtils.newThreadFactory` would be preferable.

abti:
@chavdar You had some thoughts around logging story in Gobblin, so can you please review this PR? I will request @kadaan to update the PR and fix merge conflicts / license headers if we have consensus on changing the logging framework.

kadaan:
As chance would have it, I'm reworking this to remove the GobblinCallable and GobblinRunnable in most circumstances for an ExecutorService and ScheduledExecutorService which automatically wraps Runnable and Callable.

--
Joel Baranick
jbaranick@gmail.com

kadaan:
@chavdar Updated the PR removing Logback and switching out `GobblinCallable` and `GobblinRunnable` for custom instances of ExecutorService and ScheduledExecutorService which handle setting and clearing the MDC.

abti:
@chavdar ping

kadaan:
@chavdar Can you please review this?  I'll fix the conflicts today.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -377,11 +373,9 @@ public void runJob(Properties jobProps, JobListener jobListener, JobLauncher job
       return;
     }
 
-    // Populate the assigned job ID
-    jobProps.setProperty(ConfigurationKeys.JOB_ID_KEY, JobLauncherUtils.newJobId(jobName));
-
     // Launch the job
     try (Closer closer = Closer.create()) {
+      closer.register(MDC.putCloseable(ConfigurationKeys.JOB_NAME_KEY, jobName));
```
htran1: Should set JOB_KEY_KEY too.

kadaan: Don't think that is needed here.  It is added as part of the `AbstractJobLauncher.launchJob` call done here: `closer.register(jobLauncher).launchJob(jobListener);`.

gobblin-runtime/src/main/java/gobblin/scheduler/BaseGobblinJob.java
```diff
@@ -0,0 +1,82 @@
...
+   * <code>{@link TriggerListener}s</code> that are watching the job's
+   * execution.
+   * </p>
+   *
+   * @throws JobExecutionException if there is an exception while executing the job.
+   */
+  @Override
+  public void execute(JobExecutionContext context) throws JobExecutionException {
+    Map<String, String> originalContext = MDC.getCopyOfContextMap();
+    if (this.context != null) {
```
kadaan: Sure

gobblin-runtime/src/main/java/gobblin/runtime/AbstractTaskStateTracker.java
```diff
@@ -105,6 +108,7 @@ public TaskMetricsUpdater(Task task) {
 
     @Override
     public void run() {
+      MDC.put(ConfigurationKeys.TASK_KEY_KEY, task.getTaskKey());
```
kadaan: Correct.  When a `Callable` or `Runnable` are run as part of `MDCPropagatingScheduledExecutorService` or `MDCPropagatingExecutorService`, the class is wrapped with code that automatically clone and disposes if the context.  Anything you set into the MDC in this case should automatically go away.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -303,6 +305,7 @@ public boolean apply(@Nullable WorkUnit workUnit) {
   public void launchJob(JobListener jobListener)
       throws JobException {
     String jobId = this.jobContext.getJobId();
+    MDC.put(ConfigurationKeys.JOB_KEY_KEY, this.jobContext.getJobKey());
```
kadaan: JOB_NAME should already be added by JobScheduler before calling `AbstractJobLauncher.launchJob`.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
htran1: Why separate the two? lauchJob() is called in the next line. When lauchJob() is called in a different context then won't you miss the job name unless you push the job name in launchJob() or the callers of launchJob()?

kadaan:
@htran1 The JobId/TaskId stuff is needed.  I pulled in a better version from a different PR.  I changed all the other code to just operate on the string versions, but use the Id.* classes to extract the sequence.  Other than that, I rebuilt the commit in a new branch on master as this one was way too messy and had an excessive number of files in it.

gobblin-cluster/src/test/java/gobblin/cluster/TestHelper.java
```diff
@@ -55,8 +55,10 @@
   public static final String TEST_TASK_RUNNER_ID = "1";
 
   public static final String TEST_JOB_NAME = "TestJob";
-  public static final String TEST_JOB_ID = JobLauncherUtils.newJobId(TEST_JOB_NAME);
-  public static final String TEST_TASK_ID = JobLauncherUtils.newTaskId(TEST_JOB_ID, 0);
+  static final String TEST_JOB_ID = JobLauncherUtils.newJobId(TEST_JOB_NAME);
```
htran1: Why is this one not public?

gobblin-runtime/src/main/java/gobblin/runtime/TaskState.java
```diff
@@ -101,6 +102,7 @@ public TaskState(WorkUnitState workUnitState) {
     addAll(workUnitState);
     this.jobId = workUnitState.getProp(ConfigurationKeys.JOB_ID_KEY);
     this.taskId = workUnitState.getProp(ConfigurationKeys.TASK_ID_KEY);
+    this.taskKey = workUnitState.getProp(ConfigurationKeys.TASK_KEY_KEY);
```
htran1: Can you add a default like "unknown_task_key" in this getProp() in case some code path missed setting it? Also, why is the task_key saved? It looks like in all non-test cases the task key is parsed from the task_id, so we can parse it here instead.

gradle/scripts/dependencyDefinitions.gradle
```diff
@@ -162,7 +158,16 @@ ext.externalDependency = [
     "googleOauthClient": "com.google.oauth-client:google-oauth-client:" + googleVersion,
     "googleApiClient": "com.google.api-client:google-api-client:" + googleVersion,
     "opencsv": "com.opencsv:opencsv:3.8",
-    "hadoopAdl" : "org.apache.hadoop:hadoop-azure-datalake:3.0.0-alpha2"
+    "hadoopAdl" : "org.apache.hadoop:hadoop-azure-datalake:3.0.0-alpha2",
+    "slf4j": [
+        "org.slf4j:slf4j-api:" + slf4jVersion,
+        "org.slf4j:slf4j-log4j12:" + slf4jVersion,
+        "org.slf4j:jcl-over-slf4j:" + slf4jVersion
```
htran1: What requires this JCL dependency?

kadaan: If you don't have `jcl-over-slf4j` then the jcl logs won't go through the slf4j subsystem and be logged by log4j.

kadaan:
@htran1 Still trying to figure out the test issue.  I can't get all the tests to run locally because of hive test failures related to metastore_db, so I don't see the failures that are being reported in Travis.

htran1:
Looks good.

kadaan:
Wooohoooo

--
Joel Baranick
jbaranick@gmail.com

PR #970, opened 2016-05-09
Fixing rest of findbugs warnings, and setting findbugs to fail the build on new warnings

sahilTakiar:
* Related to https://github.com/linkedin/gobblin/pull/968
* Resolving the rest of the FindBugs warnings and setting the build to fail if any new FindBugs warnings are introduced
* Most changes are minor, except for the following list:
    * `HiveCopyEntityHelper` had a dead local store warning because `stepPriority++` wasn't doing anything
    * `SeekableFSInputStream` doesn't seem to implement `seek` properly; its also possible for `InputStream.skip` isn't guaranteed to skip the specified number of bytes

@ibuenros can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -45,6 +45,8 @@ import com.google.common.collect.Maps;
 import com.google.common.io.Closer;
 import com.google.gson.Gson;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import gobblin.commit.CommitStep;
 import gobblin.configuration.State;
 import gobblin.data.management.copy.CopyConfiguration;
@@ -584,6 +586,8 @@ public class HiveCopyEntityHelper {
     return priority;
   }
 
+  // Suppress warnings for "stepPriority++" in the PrePublishStep constructor, as stepPriority may be used later
+  @SuppressFBWarnings("DLS_DEAD_LOCAL_STORE")
   private List<CopyEntity> getCopyEntitiesForUnpartitionedTable() throws IOException {
 
     MultiTimingEvent multiTimer = new MultiTimingEvent(this.eventSubmitter, "TableCopy", true);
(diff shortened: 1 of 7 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -605,7 +605,7 @@ private int addSharedSteps(List<CopyEntity> copyEntities, String fileSet, int in
 
     DeleteFileCommitStep deleteStep = DeleteFileCommitStep.fromPaths(targetFs, diffPathSet.pathsToDelete,
         this.dataset.properties);
-    copyEntities.add(new PrePublishStep(fileSet, Maps.<String, Object>newHashMap(), deleteStep, stepPriority++));
+    copyEntities.add(new PrePublishStep(fileSet, Maps.<String, Object>newHashMap(), deleteStep, stepPriority));
```
ibuenros: What failure was this throwing? `stepPriority` needs to be increased.

sahilTakiar: It throws a http://findbugs.sourceforge.net/bugDescriptions.html#DLS_DEAD_LOCAL_STORE because `stepPriority` is a local variable, and `stepPriority++` will file pass the original value of `stepPriority` to the `PrePublishStep` constructor, and then increment the value of `stepPriority`. So the incremented value of `stepPriority` is never used.

I wasn't sure what the desired behavior was. By removing the `++` it basically does what it always did, but I can add a `+ 1` instead if that is the desired behavior.

ibuenros: I see, it is probably because it is the last use of stepPriority in the method, so yes, it won't be used again, and nothing will break for now. The problem is if later we add new steps, we will probably forget to change this line back. Is it easy to tell it to ignore this warning?

sahilTakiar: Yes, we can ignore it, I will add that in.

sahilTakiar:
@ibuenros updated

ibuenros:
LGTM

PR #972, opened 2016-05-10
Avro2orc Source/Converter/Extractor/Publisher

pcadabam:
### Changes
Detailed design for this change can be found at https://docs.google.com/document/d/1PjvhDzvPxIjsb-gaRk1iiM7R8LyJHU3620DXmfoOeBs/edit?ts=57283d21#heading=h.z2tj4tr1he0q

This change implements everything discussed in the above design.

@abti  and @zliu41 can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.data.management.convertion.hive;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.avro.Schema;
+import org.apache.hadoop.fs.FileSystem;
+
+import com.google.common.collect.Lists;
+
+import gobblin.configuration.WorkUnitState;
+import gobblin.hive.HivePartition;
+import gobblin.hive.HiveRegistrationUnit;
+import gobblin.hive.HiveTable;
+import gobblin.source.extractor.DataRecordException;
+import gobblin.source.extractor.Extractor;
+import gobblin.util.reflection.GobblinConstructorUtils;
+
+
+/**
+ * <p>
+ * Extracts {@link QueryBasedHiveConversionEntity}s. A {@link QueryBasedHiveConversionEntity} can represent a
+ * hive table or a hive partition. Note that this extractor does not extract rows of a partition or a table. Entire
+ * table or partition is considered as a record.
+ * </p>
+ * <p>
+ * This extractor deserializes the {@link HiveTable} or {@link HivePartition} serialized by the {@link HiveSource}
+ * at {@link HiveSource#HIVE_UNIT_SERIALIZED_KEY} to build a {@link QueryBasedHiveConversionEntity}. Uses a {@link HiveAvroSchemaProvider}
+ * to get the avro {@link Schema} of the {@link HiveTable} or {@link HivePartition} being extracted.
+ * </p>
+ */
+public class HiveConvertExtractor implements Extractor<Schema, QueryBasedHiveConversionEntity> {
+
+  private static final String OPTIONAL_HIVE_AVRO_SCHEMA_PROVIDER_CLASS_KEY = "hive.unit.avroSchemaProvider.class";
+  private static final String DEFAULT_HIVE_AVRO_SCHEMA_PROVIDER_CLASS = HdfsBasedSchemaProvider.class.getName();
+
+  /**
+   * Even though each extractor only processes one {@link HiveRegistrationUnit}, We use a list with one {@link HiveRegistrationUnit}
+   * so that {@link #readRecord(QueryBasedHiveConversionEntity)} knows when to stop.
+   */
+  private final List<HiveRegistrationUnit> hiveUnits;
+  private final HiveAvroSchemaProvider hiveAvroSchemaProvider;
+
+  public HiveConvertExtractor(WorkUnitState state, FileSystem fs) {
+    HiveRegistrationUnit hiveUnit =
+        HiveSource.GENERICS_AWARE_GSON.fromJson(state.getProp(HiveSource.HIVE_UNIT_SERIALIZED_KEY),
+            HiveRegistrationUnit.class);
+
+    this.hiveUnits = Lists.newArrayList(hiveUnit);
+    this.hiveAvroSchemaProvider =
+        GobblinConstructorUtils.invokeConstructor(HiveAvroSchemaProvider.class,
+            state.getProp(OPTIONAL_HIVE_AVRO_SCHEMA_PROVIDER_CLASS_KEY, DEFAULT_HIVE_AVRO_SCHEMA_PROVIDER_CLASS), fs);
+
+  }
+
+  @Override
+  public Schema getSchema() throws IOException {
+    return null;
+  }
+
+
+  /**
+   * There is only one record ({@link QueryBasedHiveConversionEntity}) to be read. This {@link QueryBasedHiveConversionEntity} is
+   * removed from {@link #hiveUnits} list after it is read. So when gobblin runtime calls this method the second time, it return a null
+   */
+  @Override
+  public QueryBasedHiveConversionEntity readRecord(QueryBasedHiveConversionEntity reuse) throws DataRecordException,
+      IOException {
+
+    if (this.hiveUnits.isEmpty()) {
+      return null;
+    }
+    HiveRegistrationUnit hiveUnit = this.hiveUnits.remove(0);
+    return new QueryBasedHiveConversionEntity(hiveUnit, this.hiveAvroSchemaProvider.getSchema(hiveUnit));
+  }
+
+  @Override
+  public long getExpectedRecordCount() {
+    return 1;
+  }
+
+  /**
+   * Watermark is not managed by this extractor.
+   */
+  @Override
+  public long getHighWatermark() {
+    return 0;
+  }
+
+  @Override
+  public void close() throws IOException {
+  }
+
+}
--- a/gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertPublisher.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertPublisher.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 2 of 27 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/LongWatermark.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.gson.JsonElement;
+
+import gobblin.source.extractor.Watermark;
+
+/**
+ * A {@link Watermark} that uses a single long value as watermark.
+ */
+@Getter
+@AllArgsConstructor
+public class LongWatermark implements Watermark, Comparable<LongWatermark> {
```
zliu41: Can you either rename this class, or use [this one](https://github.com/zliu41/gobblin/blob/sf/gobblin-core/src/main/java/gobblin/source/extractor/extract/LongWatermark.java)?

Because this watermark class doesn't support incrementing the watermark value, and always returns 0 for `calculatePercentCompletion`, which is counter-intuitive to be called `LongWatermark`.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveSource.java
```diff
@@ -0,0 +1,168 @@
...
+ *{@link WorkUnit}s are created if the previous high watermark of a {@link Partition}
+ * or a {@link Table} are lower than the latest update time.
+ *
+ * <p>
+ * The {@link WorkUnit}s contain a serialized json of the {@link HiveTable} or {@link HivePartition} at {@value #HIVE_UNIT_SERIALIZED_KEY}
+ * This is later deserialized by the extractor.
+ * </p>
+ */
+@Slf4j
+@SuppressWarnings("rawtypes")
```
zliu41: You can use `Source<Schema, QueryBasedHiveConversionEntity>` instead of raw type.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveSource.java
```diff
@@ -0,0 +1,168 @@
...
+      // Find hive tables
+      Iterator<HiveDataset> iterator = datasetFinder.getDatasetsIterator();
+
+      while (iterator.hasNext()) {
+        HiveDataset hiveDataset = iterator.next();
+        LongWatermark expectedDatasetHighWatermark = new LongWatermark(new DateTime().getMillis());
+
+        // Create workunits for partitions
+        if (HiveUtils.isPartitioned(hiveDataset.getTable())) {
+          Map<List<String>, Partition> sourcePartitions =
```
zliu41: Why do you need a Map? It seems the keys are never used.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveSource.java
```diff
@@ -0,0 +1,168 @@
...
+
+              workUnit
+                  .setProp(HIVE_UNIT_SERIALIZED_KEY, GENERICS_AWARE_GSON.toJson(hivePartition, HivePartition.class));
+              workUnit.setWatermarkInterval(new WatermarkInterval(lowWatermark, expectedDatasetHighWatermark));
+              workUnit.setProp(PARTITION_COMPLETE_NAME_KEY, sourcePartition.getValue().getCompleteName());
+              workUnit.setProp(TABLE_COMPLETE_NAME_KEY, hiveDataset.getTable().getCompleteName());
+              workunits.add(workUnit);
+            } else {
+              log.info(String.format(
+                  "Not creating workunit for partition %s as updateTime %s is lesser than low watermark %s",
```
zliu41: Do you not need to create dummy workunit?

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+ * This extractor deserializes the {@link HiveTable} or {@link HivePartition} serialized by the {@link HiveSource}
+ * at {@link HiveSource#HIVE_UNIT_SERIALIZED_KEY} to build a {@link QueryBasedHiveConversionEntity}. Uses a {@link HiveAvroSchemaProvider}
+ * to get the avro {@link Schema} of the {@link HiveTable} or {@link HivePartition} being extracted.
+ * </p>
+ */
+public class HiveConvertExtractor implements Extractor<Schema, QueryBasedHiveConversionEntity> {
+
+  private static final String OPTIONAL_HIVE_AVRO_SCHEMA_PROVIDER_CLASS_KEY = "hive.unit.avroSchemaProvider.class";
+  private static final String DEFAULT_HIVE_AVRO_SCHEMA_PROVIDER_CLASS = HdfsBasedSchemaProvider.class
+      .getCanonicalName();
```
zliu41: `getCanonicalName` -> `getName`

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+  public Schema getSchema() throws IOException {
+    return null;
+  }
+
+  @Override
+  public QueryBasedHiveConversionEntity readRecord(QueryBasedHiveConversionEntity reuse) throws DataRecordException,
+      IOException {
+
+    if (this.hiveUnits.isEmpty()) {
+      return null;
```
zliu41: Can you add some comments: when is `this.hiveUnits` empty, why return null instead of throwing exception. Also it may be better to add a log warning.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+    HiveRegistrationUnit hiveUnit = this.hiveUnits.remove(0);
+    return new QueryBasedHiveConversionEntity(hiveUnit, this.hiveAvroSchemaProvider.getSchema(hiveUnit));
+  }
+
+  @Override
+  public long getExpectedRecordCount() {
+    return 0;
```
zliu41: It is better to explain in comments why `getExpectedRecordCount` and `getHighWatermark` returns 0 since they are non-intuitive.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/TableLevelWatermaker.java
```diff
@@ -0,0 +1,69 @@
...
+import gobblin.configuration.State;
+import gobblin.configuration.WorkUnitState;
+import gobblin.source.extractor.Watermark;
+
+/**
+ * A {@link HiveSourceWatermarker} that manages {@link Watermark} at a per {@link Table} basis.
+ * One {@link Watermark} per table exists.
+ * All {@link Partition}s of a {@link Table} have the same {@link Watermark}.
+ */
+public class TableLevelWatermaker implements HiveSourceWatermarker {
```
zliu41: maker -> marker

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/TableLevelWatermaker.java
```diff
@@ -0,0 +1,69 @@
...
+  public static final Gson GSON = new Gson();
+
+  // Table complete name db@tb - list of previous workunitState
+  private Map<String, LongWatermark> tableWatermarks;
+
+  public TableLevelWatermaker(State state) {
+    this.tableWatermarks = Maps.newHashMap();
+
+    if (state instanceof SourceState) {
```
zliu41: I think this is always true. You should check whether state is instanceof `DatasetState`. If so, skip the block.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertPublisher.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public void initialize() throws IOException {}
+
+  @Override
+  public void publishData(Collection<? extends WorkUnitState> states) throws IOException {
+    for (WorkUnitState wus : states) {
+      wus.setWorkingState(WorkingState.COMMITTED);
+      this.hiveSourceWatermarker.setHighWatermark(wus);
```
zliu41: Do you really need a `HiveSourceWatermarker`? Because each `wus` has an expected highwm. If it is committed, you can simply set the actual highwm to be the same as expected highwm.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/TableLevelWatermaker.java
```diff
@@ -0,0 +1,69 @@
...
+            .get(tableCompleteName)) < 0) || !this.tableWatermarks.containsKey(tableCompleteName)) {
+          this.tableWatermarks.put(tableCompleteName, actualHW);
+        }
+      }
+    }
+  }
+
+  @Override
+  public LongWatermark getPreviousHighWatermark(Table table) {
+    return this.tableWatermarks.getOrDefault(table.getCompleteName(), new LongWatermark(0));
```
zliu41: `getOrDefault` is not available in Java 7.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/LongWatermark.java
pcadabam: Seems like this [LongWatermark](https://github.com/zliu41/gobblin/blob/sf/gobblin-core/src/main/java/gobblin/source/extractor/extract/LongWatermark.java) has not been merged yet. 
Will use it once merged

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveSource.java
pcadabam: The idea is to not tie the `HiveSource` with a single extractor. (that does conversion `QueryBasedHiveConversionEntity`). The source should be unaware of how the extractors and converters will use the workunits it generates. So in theory we can inject the extractor class through properties.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveSource.java
pcadabam: Yes the keys are not used. I was reusing the existing utility function `HiveUtils.getPartitionsMap`. Added `HiveUtils.getPartitions` that returns a list of partitions

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
pcadabam: Done.

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
pcadabam: The list is guaranteed to  contain only 1 record. Add more javadoc

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertExtractor.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/TableLevelWatermaker.java
pcadabam: done

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/HiveConvertPublisher.java
pcadabam: Removed

gobblin-data-management/src/main/java/gobblin/data/management/convertion/hive/TableLevelWatermaker.java
pcadabam: Used the if/else one

pcadabam:
@zliu41 I have addressed all pr comments. Can you take another look?

zliu41:
LGTM

PR #973, opened 2016-05-10
Fix bin packing algorithm when large work units prevent small units f…

ibuenros:
…rom being packed.


ibuenros:

pcadabam:
LGTM

PR #974, opened 2016-05-10
Deprecating and removing support for Hadoop versions other than 2.x.x

sahilTakiar:
* Deprecating and removing support for building against Hadoop version other than 2.x.x
* The build will now throw an error if a user tries to set the `hadoopVersion` to a non-2.x.x version
* Remove the `useHadoop2` build option, as it is no longer applicable
* The Travis build has been updated to only build against Hadoop2, rather than Hadoop2 and Hadoop1
* I did my best to go through all the documentation and remove any mention of support for building against Hadoop 1.x.x
* I did my best to go through the code and remove any support for Hadoop 1.x.x
    * `gobblin-adaptor` has been removed; there was only one class in the there called `TokenUtils`, I moved it into `gobblin-utility`
    * There are a few other small code changes that remove Hadoop1 support
* Updated the Maven Upload scripts and removed the logic for uploading hadoop1 and hadoop2 artifacts

@ibuenros can you review?


gobblin-runtime/src/test/java/gobblin/runtime/mapreduce/MRJobLauncherTest.java
```diff
@@ -228,12 +228,11 @@ public void testLaunchJobWithMultipleDatasets() throws Exception {
...
    * failing and aborting a job if any mapper task fails. Aborting the job causes its working directory
    * to be deleted in {@link GobblinOutputCommitter}, which further fails this test since all the output
-   * {@link gobblin.runtime.TaskState}s are deleted. It works fine in Hadoop1 though by setting
-   * mapred.max.map.failures.percent=100. There may be a bug in Hadoop2's LocalJobRunner.
+   * {@link gobblin.runtime.TaskState}s are deleted. There may be a bug in Hadoop2's LocalJobRunner.
    *
    * Also applicable to the two tests below.
    */
-  @Test(groups = { "Hadoop1Only" })
+  @Test(groups = { "ignore" })
```
ibuenros: Is there any reason to keep these tests?

ibuenros:
Just one comment. LGTM

gobblin-runtime/src/test/java/gobblin/runtime/mapreduce/MRJobLauncherTest.java
sahilTakiar: I'm not sure. According to the comments for the tests, they were disabled for Hadoop 2.x.x because of an apparent bug in Hadoop 2.x.x's LocalJobRunner. Hopefully, one day that bug will be fixed, in which case we can re-enabled these tests.

PR #979, opened 2016-05-11
Initial commit for HTTP Writer APIs

chavdar:
@jinhyukchang can you have a look ?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/http/HttpWriterDecorator.java
+++ b/gobblin-core/src/main/java/gobblin/writer/http/HttpWriterDecorator.java
@@ -1,0 +1,65 @@
+/**
+ *
+ */
+package gobblin.writer.http;
+
+import java.io.IOException;
+
+import org.apache.http.HttpHost;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpUriRequest;
+
+import com.google.common.base.Optional;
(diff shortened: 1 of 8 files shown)


gobblin-core/src/main/java/gobblin/writer/http/HttpWriterDecorator.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * Common parent for {@link AbstractHttpWriter} decorators. Delegates extension methods to another
+ * implementation and simplifies the overriding of only selected methods
+ */
+public abstract class HttpWriterDecorator<D> implements HttpWriterDecoration<D> {
```
jinhyukchang: Seems it doesn't have to be an abstract class. Is it intended to be an abstract class so that it's overridden?

jinhyukchang:
@chavdar LGTM. Just minor comment.

gobblin-core/src/main/java/gobblin/writer/http/HttpWriterDecorator.java
chavdar: Yeah. Although it does not have any abstract methods, I don't see a point of instantiating this class.

PR #980, opened 2016-05-11
Add metadata after completion of job to a specific metadata directory

singhd10:


singhd10:
Add metadata after completion of job to a specific metadata directory

singhd10:
@sahilTakiar  Can you please take a look ?

singhd10:
Can someone merge the code ?

chavdar:
@singhd10 I am not sure what that metadata.txt file is and what it is supposed to contain. If it is something specific to your use case, you create your own DataPublisher implementation which extends the existing publisher.

sahilTakiar:
In general, if you want to get metadata out of a Gobblin job you can take a look at Gobblin Metrics: http://gobblin.readthedocs.io/en/latest/metrics/Gobblin-Metrics/ - specifically the part on `Event`s and `EventReporter`s

singhd10:
@chavdar metadata about the ingestion data e.g. source, schema, timestamp etc. is written to the metadata.txt file. It could be json text or simple text. User can add this in config file. 
Its a no op if these configurations are not set.

IMO, its a good feature to have in gobblin.

Use case 
1. User needs to add information about he source of data
2. User adds json text in the configuration file.
3. Gobblin writes the json in metadata.txt after job completion.

rmschulman:
ping.  any further comments/questions?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -277,6 +277,13 @@
   public static final String DATA_PUBLISHER_PREFIX = "data.publisher";
 
   /**
+   * Metadata configuration properties used internally
+   */
+  public static final String DATA_PUBLISHER_METADATA_OUTPUT_DIR =  "metadata.output.dir";
```
chavdar: Please prefix both properties with "data.publisher." 
The second property should be something along the lines of "data.publisher.metadata.string" .

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -336,12 +341,50 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+      LOG.error("Missing metadata output directory path : "  + ConfigurationKeys.DATA_PUBLISHER_METADATA_OUTPUT_DIR 
+                + " in the config");
+      return;
+    }
+
+    //Write for each branch
+    for (int branchId = 0; branchId < this.numBranches; branchId++) {
+      FileSystem fs = this.metaDataWriterFileSystemByBranches.get(branchId);
 
+      String fileName = ForkOperatorUtils.getPropertyNameForBranch("metadata.txt", this.numBranches, branchId);
```
chavdar: Please use a named constant and add the "data.publisher." prefix. "metadata.txt" suggest a plain-text format. I suggest choosing a name along the lines of "data.publisher.metadata.output_file".

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -336,12 +341,50 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+        }
+
+        Path metaFilepath = new Path(metaDataOutputDirStr, fileName);
+
+        //Delete the file if metadata already exists
+        if (fs.exists(metaFilepath)) {
+          HadoopUtils.deletePath(fs, metaFilepath, false);
+        }
+
+        FSDataOutputStream outputStream = this.closer.register(fs.create(metaFilepath));
```
chavdar: Since Gobblin is Java 1.7-compatible, and the scope of dealing with outputStream is limited to this method, I suggest using a try-resource block.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -336,12 +341,50 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+        outputStream.write(metadataValue.getBytes());
```
chavdar: FindBugs emits a warning for using getBytes() with the default encoding. I suggest using 'UTF-8' explicitly.

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -336,12 +341,50 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
+      } catch (IOException e) {
+        LOG.error("metadata file is not generated", e);
```
chavdar: Nit: It is a useful pattern to do: LOG.error("Some error message:" + e, e)

This will log the "e" message on the same line and it makes it easy to grep for errors.

chavdar:
Sorry about that @rmschulman @singhd10 This has fallen off my radar. I've provided some comments.

singhd10:
@chavdar Incorporated all the changes suggested by you, pleased merge the code.
cc: @rmschulman

chavdar:
LGTM

PR #983, opened 2016-05-12
need to deregister existing table

tuGithub:


tuGithub:
@chavdar , please review it when you have time.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -609,6 +612,41 @@ private int addDeregisterSteps(List<CopyEntity> copyEntities, String fileSet, in
...
     return stepPriority;
   }
+  
+  private int addTableDeregisterSteps(List<CopyEntity> copyEntities, String fileSet, int initialPriority,
+      Table table) throws IOException {
+
+    int stepPriority = initialPriority;
+
+    Collection<Path> tablePaths = Lists.newArrayList();
+    DeregisterFileDeleteMethod deleteMethod = this.dataset.properties.containsKey(DELETE_FILES_ON_DEREGISTER)
```
chavdar: Checking of the property value should happen in the constructor.

chavdar: Also, please access "properties" through a getter.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -609,6 +612,41 @@ private int addDeregisterSteps(List<CopyEntity> copyEntities, String fileSet, in
...
+        ? DeregisterFileDeleteMethod.valueOf(this.dataset.properties.getProperty(DELETE_FILES_ON_DEREGISTER).toUpperCase())
+        : DeregisterFileDeleteMethod.NO_DELETE;
+
+    if (deleteMethod == DeregisterFileDeleteMethod.RECURSIVE) {
```
chavdar: I suggest using "switch" instead of "if" statement. As a safeguard, have a default clause to catch cases where new constants have been added but they are unprocessed.

chavdar:
Could you add unit tests?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
tuGithub: fixed

tuGithub:
Addressed the comments, please review

zliu41:
LGTM. Please use the "squash and merge" option and squash into a single commit with a meaningful commit message.

PR #988, opened 2016-05-13
Adding documentation page for Gobblin Distcp

sahilTakiar:
* Taken from an internal design doc and trimmed down a little bit
* Added some better formatting, otherwise the content is the same

@ydai1124 can you review?


gobblin-docs/adaptors/Gobblin-Distcp.md
```diff
@@ -0,0 +1,275 @@
...
+* Gobblin execution implementation
+* Gobblin publishing implementation
+* Gobblin metrics
+* Gobblin on YARN
+* Exactly once semantics
+* Automatic Hive registration
+
+# Performance, Scalability and Provisioning
+
+There are three components in the flow:
```
ydai1124: But only two components are listed.

ydai1124:
LGTM other than one comment.

PR #989, opened 2016-05-13
Added retention docs

pcadabam:
@sahilTakiar can you review?


sahilTakiar:
Just one high level comment, I would suggest using three tick marks to surround code blocks, rather than using <pre></pre> - the three tick marks are the standard way of creating code blocks, and it supports syntax highlighting automatically - https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

Also "Table of COntents" should be "Table of Contents", and "Intrduction" should be "Introduction"

pcadabam:
Fixed. Merging.

PR #990, opened 2016-05-14
Enable creating services which accept `Properties` for configuration

kadaan:
@sahilTakiar Additional services that are configured via the config file should be able to be passed properties so they can configure themselves.


sahilTakiar:
LGTM

PR #992, opened 2016-05-16
Making kafka metadata read more resillient to issues with the brokers

chavdar:
@zliu41 can you review ?

It seems like the brokers can sometimes return partial partition metadata that can throw NPEs . This should cause those topics to be ignored.

Please see ETL-4495 for more info.


sahilTakiar:
Probably related to https://groups.google.com/d/topic/gobblin-users/zXL0zzDmHbQ/discussion - this PR should fix the issue surfaced in the discussion

zliu41:
LGTM

PR #994, opened 2016-05-17
Tutorial on writing ORC files

sahilTakiar:
* Adding a tutorial on how to write data to ORC files using the `HiveSerDeConverter`
    * Tutorial consists of a new documentation page and a new example `.pull` file
* The tutorial is based on the Wikipedia Example and first converts records to Avro before sending them through the `HiveSerDeConverter`, which converts them to `ORCSerdeRow`s
* Fixed a bug in `HiveWritableHdfsDataWriter` so that the file being written is closed before it gets committed
* Added a new CSS rule for all documentation so that text formatted with the `code` tag looks better

@zliu41 can you review?


zliu41:
Please fix the failed unit test, otherwise LGTM

PR #997, opened 2016-05-18
Fix Build

zliu41:


sahilTakiar:
What's the issue with the `gobblin-tunnel` tests? This PR seems to disable a lot of them, which I agree we should do. The whole `gobblin-tunnel` module can probably have its tests disabled. In fact, `gobblin-tunnel` should probably be moved to a separate repo on GitHub.

zliu41:
Not sure, there were assert failures and NPEs. I'm disabling them to unblock Hudson build. We can re-enable them after we know what's wrong with them.

chavdar:
What are the exceptions in gobblin-tunnel?

chavdar:
What is the list of tests that are being disabled?

chavdar:
At least some failures maybe attributable to missing artifacts in the artifactory:

:gobblin-admin:test/export/home/tester/hudson/data/workspace/ETL_infra-gobblin-proxy-build/gobblin-proxy_trunk/gobblin-github/gobblin-core/src/test/java/gobblin/converter/filter/AvroFieldsPickConverterTest.java:38: error: cannot find symbol
      JSONAssert.assertEquals(expected.toString(), converted.toString(), false);
      ^
  symbol:   variable JSONAssert
  location: class AvroFieldsPickConverterTest

chavdar:
Actually, it is a missing testCompile dependency: https://github.com/linkedin/gobblin/pull/1000

chavdar:
There might be also issues due to the new machine that we got. I see error failing to create files.

PR #999, opened 2016-05-18
distcp-ng throws UnsupportedOperationException

tuGithub:


tuGithub:
@zliu41, @chavdar  , please review it when you have time

zliu41:
LGTM

chavdar:
Can you please explain what the bug was? It seems like a significant change and the previous version of the code was there for a while.

PR #1001, opened 2016-05-18
Setting heaps size for gobblin-runtime tests due to OOM in some cases

chavdar:


chavdar:
Mockito state is causing OOM: http://stackoverflow.com/questions/17437660/mockito-throws-an-outofmemoryerror-on-a-simple-test

PR #1002, opened 2016-05-19
Set explicit 755 permissions to state store

chavdar:
This is required by the distributed cache: http://stackoverflow.com/questions/23903113/mapreduce-error-usergroupinformation-priviledgedactionexception


chavdar:
@zliu41 can you review?

PR #1003, opened 2016-05-19
Added the predicate check to skip full table diff if the existing table's registration time > source table's mod time

tuGithub:


tuGithub:
@chavdar , please review when you have time.

tuGithub:
@ydai1124 , please review it when you have time

tuGithub:
@ibuenros , please review it when you have time

ibuenros:
It looks good in general. Only comment is: the logic for the table fast skip predicate is almost identical to the partition skip predicate, is there any way to avoid the code repetition? Could we use a single configuration key and instance variable for both?

ibuenros:
Discussed offline, LGTM

PR #1008, opened 2016-05-20
ETL-4470: Implementation of http filer puler using Distcp-ng

ydai1124:
https://docs.google.com/document/d/1EmjQLmpvbFcBcFdCQSC46MkQ_EiowGDG8WLt729Zwdg/

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
@@ -23,9 +23,6 @@ import java.util.concurrent.Future;
 
 import javax.annotation.Nullable;
 
-import lombok.AllArgsConstructor;
-import lombok.extern.slf4j.Slf4j;
-
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 
@@ -34,6 +31,9 @@ import com.google.common.base.Optional;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.SourceState;
 import gobblin.configuration.State;
@@ -41,6 +41,8 @@ import gobblin.configuration.WorkUnitState;
 import gobblin.data.management.copy.extractor.EmptyExtractor;
 import gobblin.data.management.copy.extractor.FileAwareInputStreamExtractor;
 import gobblin.data.management.copy.publisher.CopyEventSubmitterHelper;
+import gobblin.data.management.copy.watermark.CopyableFileWatermarkGenerator;
+import gobblin.data.management.copy.watermark.CopyableFileWatermarkHelper;
 import gobblin.data.management.dataset.DatasetUtils;
 import gobblin.data.management.partition.FileSet;
 import gobblin.dataset.Dataset;
@@ -54,14 +56,15 @@ import gobblin.metrics.Tag;
 import gobblin.metrics.event.EventSubmitter;
 import gobblin.metrics.event.sla.SlaEventKeys;
 import gobblin.source.extractor.Extractor;
+import gobblin.source.extractor.WatermarkInterval;
 import gobblin.source.extractor.extract.AbstractSource;
 import gobblin.source.workunit.Extract;
 import gobblin.source.workunit.WorkUnit;
+import gobblin.source.workunit.WorkUnitWeighter;
 import gobblin.util.ExecutorsUtils;
 import gobblin.util.HadoopUtils;
 import gobblin.util.WriterUtils;
 import gobblin.util.binpacking.FieldWeighter;
-import gobblin.source.workunit.WorkUnitWeighter;
 import gobblin.util.binpacking.WorstFitDecreasingBinPacking;
 import gobblin.util.executors.IteratorExecutor;
 import gobblin.util.guid.Guid;
@@ -115,6 +118,7 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
    */
   @Override
   public List<WorkUnit> getWorkunits(final SourceState state) {
+
     this.metricContext = Instrumented.getMetricContext(state, CopySource.class);
 
     try {
@@ -124,6 +128,8 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
       long maxSizePerBin = state.getPropAsLong(MAX_SIZE_MULTI_WORKUNITS, 0);
       long maxWorkUnitsPerMultiWorkUnit = state.getPropAsLong(MAX_WORK_UNITS_PER_BIN, 50);
       final long minWorkUnitWeight = Math.max(1, maxSizePerBin / maxWorkUnitsPerMultiWorkUnit);
+      final Optional<CopyableFileWatermarkGenerator> watermarkGenerator =
+          CopyableFileWatermarkHelper.getCopyableFileWatermarkGenerator(state);
 
       // TODO: The comparator sets the priority of file sets. Currently, all file sets have the same priority, this needs to
       // be pluggable.
@@ -132,17 +138,19 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
 
       final CopyConfiguration copyConfiguration = CopyConfiguration.builder(targetFs, state.getProperties()).build();
 
-      DatasetsFinder<CopyableDatasetBase> datasetFinder =
-          DatasetUtils.instantiateDatasetFinder(state.getProperties(), sourceFs, DEFAULT_DATASET_PROFILE_CLASS_KEY,
-              new EventSubmitter.Builder(this.metricContext, CopyConfiguration.COPY_PREFIX).build());
+      DatasetsFinder<CopyableDatasetBase> datasetFinder = DatasetUtils
+          .instantiateDatasetFinder(state.getProperties(), sourceFs, DEFAULT_DATASET_PROFILE_CLASS_KEY,
+              new EventSubmitter.Builder(this.metricContext, CopyConfiguration.COPY_PREFIX).build(), state);
 
-      IterableDatasetFinder<CopyableDatasetBase> iterableDatasetFinder = datasetFinder instanceof IterableDatasetFinder
-          ? (IterableDatasetFinder<CopyableDatasetBase>) datasetFinder : new IterableDatasetFinderImpl<>(datasetFinder);
+      IterableDatasetFinder<CopyableDatasetBase> iterableDatasetFinder =
+          datasetFinder instanceof IterableDatasetFinder ? (IterableDatasetFinder<CopyableDatasetBase>) datasetFinder
+              : new IterableDatasetFinderImpl<>(datasetFinder);
 
       Iterator<CopyableDatasetBase> copyableDatasets =
           new InterruptibleIterator<>(iterableDatasetFinder.getDatasetsIterator(), new Callable<Boolean>() {
             @Override
-            public Boolean call() throws Exception {
+            public Boolean call()
+                throws Exception {
               return shouldStopGeneratingWorkUnits(workUnitList);
             }
           });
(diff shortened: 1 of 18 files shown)


ydai1124:
@ibuenros Can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -375,22 +454,107 @@ public static void serializeCopyEntity(State state, CopyEntity copyEntity) {
...
+      throws IOException {
     return CopyableDatasetMetadata.deserialize(state.getProp(SERIALIZED_COPYABLE_DATASET));
   }
 
+  /**
+   * Return the mapping from datasetUrn to {@link IncludeExcludeWatermark} based on previous workunitstates if filterByWatermark is enabled and watermarkGenerator is provided.
+   * If all {@link WorkUnitState}s are committed, watermark value will be the highest high watermark among all, and this value should be excluded for next time.
+   * Otherwise, it will be the lowest low watermark among all non-committed {@link WorkUnitState}s, and this value should included for next time.
+   */
+  public Map<String, IncludeExcludeWatermark> getPreivousWatermarkForFilterIfNecessary(boolean filterByWatermark,
```
ibuenros: typo: Preivous -> Previous

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -306,16 +349,49 @@ private boolean shouldStopGeneratingWorkUnits(ConcurrentBoundedWorkUnitList work
...
+      throws IOException {
+    if (!watermarkGenerator.isPresent()) {
+      return Optional.absent();
+    }
+    return watermarkGenerator.get().generateWatermarkIntervalForCopyEntity(copyEntity);
+  }
 
+  protected FileSystem getSourceFileSystem(State state)
+      throws IOException {
+    if (state.contains(CUSTOMIZED_FILESYSTEM)) {
```
ibuenros: Usually this should be done by Hadoop itself. When you provide a fs URI, Hadoop will use the scheme to try to find a FileSystem matching that scheme. To provide custom `FileSystem` implementations you can either use the configuration "fs.<scheme>.impl", or register the implementation as a service provider for `FileSystem` (creating a file `META-INF/service/org.apache.hadoop.fs.FileSystem` containing the name of the implementation).

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -246,13 +280,20 @@ public Void call() {
...
           Extract extract = new Extract(Extract.TableType.SNAPSHOT_ONLY, CopyConfiguration.COPY_PREFIX, extractId);
           List<WorkUnit> workUnitsForPartition = Lists.newArrayList();
           for (CopyEntity copyEntity : fileSet.getFiles()) {
 
             CopyableDatasetMetadata metadata = new CopyableDatasetMetadata(this.copyableDataset);
             CopyEntity.DatasetAndPartition datasetAndPartition = copyEntity.getDatasetAndPartition(metadata);
-
+            Optional<WatermarkInterval> watermarkIntervalOptional =
+                getCopyEntityWatermark(copyEntity, watermarkGenerator);
+            if (shouldSkipDueToWatermark(filterWuByWatermark,
```
ibuenros: Is there a way that the `Dataset` itself can use the watermark? This is doing a late filtering, but the dataset might be able to optimize this filtering, or use the watermark to do a faster listing.

ibuenros:
Would it make sense to move all of the watermark functionality to a helper class? `CopySource` is getting pretty long.

gobblin-data-management/src/main/java/gobblin/data/management/copy/SubsetFilesCopyableDatasetFinder.java
```diff
@@ -0,0 +1,115 @@
...
+
+
+/**
+ * {@link SubsetFilesCopyableDataset} finder that extends {@link ConfigurableGlobDatasetFinder}.
+ *
+ * It lists files on the root path, and uses a user defined pattern to identify {@link SubsetFilesCopyableDataset}s and
+ * their corresponding subset of files.
+ */
+@Slf4j
+public class SubsetFilesCopyableDatasetFinder extends ConfigurableGlobDatasetFinder<CopyableDataset> {
```
ibuenros: Can't this functionality be achieved with `RecursiveCopyableDataset` itself? If I understand correctly, you can just use the identifier pattern to assign files to a file set, and add a filter.

gobblin-data-management/src/main/java/gobblin/data/management/copy/SubsetFilesCopyableDatasetFinder.java
```diff
@@ -0,0 +1,115 @@
...
+  public static final String IDENTIFIER_PATTERN = CopyConfiguration.COPY_PREFIX + ".subsetFilesDatasetIdentifer";
+  public static final String DEFAULT_IDENTIFIER_PATTERN = "([\\w_\\d]*\\.[\\w_\\d]*)\\..*(gz|meta)";
+
+  public static final String SUBSETFILES_REGEX_FILTER = CopyConfiguration.COPY_PREFIX + ".subsetFilesRegexFilter";
+  public static final String DEFAULT_SUBSETFILES_REGEX_FILTER = ".*(gz|meta)";
```
ibuenros: I don't think this is a sensible default for open source.

gobblin-data-management/src/main/java/gobblin/data/management/copy/watermark/StringBasedCopyEntityWatermarkGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+import gobblin.data.management.copy.CopyEntity;
+import gobblin.data.management.copy.CopyableFile;
+import gobblin.source.extractor.ComparableWatermark;
+import gobblin.source.extractor.WatermarkInterval;
+
+
+/**
+ * Implementation of {@link CopyEntityWatermarkGenerator} that uses {@link StringWatermark}.
+ */
+public class StringBasedCopyEntityWatermarkGenerator implements CopyEntityWatermarkGenerator {
```
ibuenros: Can you rename this to something more descriptive? Like `FullPathCopyEntityWatermarkGenerator`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/watermark/StringBasedCopyEntityWatermarkGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.common.base.Optional;
+
+import gobblin.data.management.copy.CopyEntity;
+import gobblin.data.management.copy.CopyableFile;
+import gobblin.source.extractor.ComparableWatermark;
+import gobblin.source.extractor.WatermarkInterval;
+
+
+/**
+ * Implementation of {@link CopyEntityWatermarkGenerator} that uses {@link StringWatermark}.
```
ibuenros: Please explain what string will be returned.

gobblin-utility/src/main/java/gobblin/util/io/HttpInputStream.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+
+import org.apache.hadoop.fs.FSDataInputStream;
+
+
+/**
+ * An extension of {@link FSDataInputStream} that takes a {@link HttpURLConnection} and closes the connection in {@link #close()} method.
+ */
+public class HttpInputStream extends FSDataInputStream {
```
ibuenros: This could be more generic, taking any `Closeable` or even a set of `Closeables` that will al be closed at the same time as the `InputStream`.

gobblin-core/build.gradle
```diff
@@ -36,6 +36,8 @@ dependencies {
   compile externalDependency.commonsIo
   compile externalDependency.hiveExec
   compile externalDependency.hiveSerDe
+  compile externalDependency.htmlparser
+  compile externalDependency.htmllexer
```
ibuenros: What are these being used for?

gobblin-data-management/src/main/java/gobblin/data/management/copy/SubsetFilesCopyableDatasetFinder.java
ydai1124: It is not recursively finding datasets. The non-recursive listing phase is done in `datasetFinder` rather than `dataset` for performance purpose. Limited logic in `RecursiveCopyableDataset` can be reused here.

ydai1124:
@ibuenros Updated.

gobblin-data-management/src/main/java/gobblin/data/management/copy/SubsetFilesCopyableDatasetFinder.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+    return datasets;
+  }
+
+  public CopyableDataset datasetAndPathWithIdentifier(Path path, String identifier)
+      throws IOException {
+    List<Object> args =
+        Lists.newArrayList(fs, path, props, identifier, idToFileStatuses.get(identifier), eventSubmitter, state);
+    try {
+      return GobblinConstructorUtils.invokeLongestConstructor(SubsetFilesCopyableDataset.class, args.toArray());
```
ibuenros: You can just call `GobblinConstructorUtils.invokeLongestConstructor(SubsetFilesCopyableDataset.class, fs, path, props, identifier, idToFileStatuses.get(identifier), eventSubmitter, state)`

gobblin-data-management/src/main/java/gobblin/data/management/copy/SubsetFilesCopyableDatasetFinder.java
ibuenros: There is a lot of repeated logic between this and the recursive copyable dataset, and features that that one would benefit from (like using a pattern to define filesets). Can you merge them as much as possible?

gobblin-data-management/src/main/java/gobblin/data/management/copy/WatermarkBasedCopyableFileFilter.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Implementation of {@link CopyableFileFilter} that filters {@link CopyableFile}s based on previous watermark information.
+ */
+@Slf4j
+public class WatermarkBasedCopyableFileFilter implements CopyableFileFilter {
+  /**
+   * Only create work units that have higher expected high watermark than the previous actual watermark.
+   */
+  public static final String FILTER_BY_WATERMARK = CopyConfiguration.COPY_PREFIX + ".filterWorkUnitsByWatermark";
```
ibuenros: Why is this needed? Is this filter always instantiated or is it user specified? If the latter, wouldn't the user specifying this filter imply that they want the filtering?

gobblin-data-management/src/main/java/gobblin/data/management/copy/WatermarkBasedCopyableFileFilter.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  /**
+   * Return the mapping from datasetUrn to {@link IncludeExcludeWatermark} based on previous workunitstates if filterByWatermark is enabled and watermarkGenerator is provided.
+   * If all {@link WorkUnitState}s are committed, watermark value will be the highest high watermark among all, and this value should be excluded for next time.
+   * Otherwise, it will be the lowest low watermark among all non-committed {@link WorkUnitState}s, and this value should included for next time.
+   */
+  @VisibleForTesting
+  protected Map<String, IncludeExcludeWatermark> getPreviousWatermarkForFilterIfNecessary(boolean filterByWatermark,
```
ibuenros: It seems that a new Map is created for each dataset, even though all maps are the same. Can this be instantiated once and reused?

gobblin-data-management/src/main/java/gobblin/data/management/copy/watermark/CopyEntityWatermarkGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+/**
+ * Watermark generator for {@link CopyEntity}.
+ */
+public interface CopyEntityWatermarkGenerator {
+  /**
+   * Generate optional {@link WatermarkInterval} for a given {@link CopyEntity}.
+   */
+  public Optional<WatermarkInterval> generateWatermarkIntervalForCopyEntity(CopyEntity copyEntity) throws IOException;
```
ibuenros: Above you only use watermarks for `CopyableFile`, is there a point to allowing watermarks for any `CopyEntity`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/WatermarkBasedCopyableFileFilter.java
ydai1124: It is actually only instantiated once. `STATE_TO_PREV_WATERMARK_MAP_CAHCE` stores the mapping from `SourceState` reference to `Map`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/watermark/CopyEntityWatermarkGenerator.java
ydai1124: Not for my case. I cannot really come up a good use case for other `CopyEntity`. Will remove that.

ydai1124:
@ibuenros  Updated.

ibuenros:
LGTM

PR #1015, opened 2016-05-26
Fix hanging IteratorExecutor. HiveDataset will skip bad tables instea…

ibuenros:
…d of failing the entire job.

This PR fixes two issues:
* IteratorExecutor does not shutdown executor on a failure, making job hang.
* HiveDatasetFinder fails if a single table is bad. Fix will skip and report bad tables.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDatasetFinder.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDatasetFinder.java
@@ -18,35 +18,36 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
 
-import javax.annotation.Nullable;
-
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.thrift.TException;
 
-import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
+import com.google.common.collect.AbstractIterator;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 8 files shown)


ibuenros:

gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
```diff
@@ -59,29 +59,32 @@ public IteratorExecutor(Iterator<Callable<T>> runnableIterator, int numThreads,
...
-    while (activeTasks > 0) {
-      this.completionService.take();
-      activeTasks--;
-    }
 
-    this.completionService.poll();
-
-    ExecutorsUtils.shutdownExecutorService(this.executor, Optional.of(log), 10, TimeUnit.SECONDS);
-    this.executed = true;
+      this.completionService.poll();
```
ydai1124: Is `poll` necessary here? `take` seems have taken care of all completion service?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDatasetFinder.java
```diff
@@ -134,20 +157,32 @@ public boolean apply(String table) {
...
-          Table table = client.get().getTable(dbAndTable.getDb(), dbAndTable.getTable());
-          return createHiveDataset(table);
-        } catch (IOException | TException ioe) {
-          throw new RuntimeException(ioe);
+      protected HiveDataset computeNext() {
+        while (this.tables.hasNext()) {
+          DbAndTable dbAndTable = this.tables.next();
+          try (AutoReturnableObject<IMetaStoreClient> client = HiveDatasetFinder.this.clientPool.getClient()) {
+            Table table = client.get().getTable(dbAndTable.getDb(), dbAndTable.getTable());
+            if (HiveDatasetFinder.this.eventSubmitter.isPresent()) {
```
ydai1124: nit: create a method to optionally submit events which checks if `eventSubmitter` is present. So you don't need to have this `if` blocks twice.

ibuenros:
@ydai1124 addressed comments.

ydai1124:
LGTM

PR #1016, opened 2016-05-26
Integration with Confluent Schema Registry, Confluent Deserializers, and Kafka Deserializers

sahilTakiar:
* Adding integration with Confluent's Schema Registry - https://github.com/confluentinc/schema-registry
    * `ConfluentKafkaSchemaRegistry.java`
* Created `KafkaDeserializerExtractor` which wraps Kafka's `Deserializer` API
    * Kafka's `Deserializer` interface provides a generic way for converting `byte[]` to Java Objects
    * The `KafkaDeserializerExtractor` takes in a path to a fully qualified implementation of `Deserializer` and uses it to convert they `byte[]` into a Java Object
    * It has built in support for a few known `Deserializer`s including Confluents `KafkaAvroDeserializer` and `KafkaJsonDeserializer`
* Created a `KafkaGsonDeserializer` that converts `byte[]` to `JsonElement`s
* Updated the `gobblin-docs/case-studies/Kafka-HDFS-Ingestion.md` to document all of these new features
* Unit tests added

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -134,6 +134,7 @@ if (!project.hasProperty('bytemanVersion')) {
 ext.avroVersion = '1.7.7'
 ext.dropwizardMetricsVersion = '3.1.0'
 ext.findBugsVersion = '3.0.0'
+ext.confluentVersion = '2.0.1'
 
 ext.externalDependency = [
   "antlrRuntime": "org.antlr:antlr-runtime:3.5.2",
@@ -178,6 +179,9 @@ ext.externalDependency = [
   "kafka": "org.apache.kafka:kafka_2.11:0.8.2.2",
   "kafkaTest": "org.apache.kafka:kafka_2.11:0.8.2.2:test",
   "kafkaClient": "org.apache.kafka:kafka-clients:0.8.2.2",
+  "confluentSchemaRegistryClient": "io.confluent:kafka-schema-registry-client:" + confluentVersion,
+  "confluentAvroSerializer": "io.confluent:kafka-avro-serializer:" + confluentVersion,
+  "confluentJsonSerializer": "io.confluent:kafka-json-serializer:" + confluentVersion,
   "quartz": "org.quartz-scheduler:quartz:2.2.3",
   "testng": "org.testng:testng:6.9.10",
   "mockserver":"org.mock-server:mockserver-netty:3.10.4",
@@ -446,6 +450,12 @@ subprojects {
 
         // Required to add JDK's tool jar, which is required to run byteman tests.
         testCompile (files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs()))
+      }
+    }
+
+    repositories {
+      maven {
+        url "http://packages.confluent.io/maven/"
       }
     }
 
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaDeserializerExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaDeserializerExtractor.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.source.extractor.extract.kafka;
+
+import java.io.IOException;
+import java.util.Properties;
+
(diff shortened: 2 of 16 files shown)


sahilTakiar:
@zliu41 can you review?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaDeserializerExtractor.java
```diff
@@ -0,0 +1,193 @@
...
+ *     defines the {@link Deserializer} interface. If this property is used a {@link KafkaSchemaRegistry} must also
+ *     be specified using the {@link KafkaSchemaRegistry#KAFKA_SCHEMA_REGISTRY_CLASS} config key</li>
+ *   </ul>
+ * </p>
+ */
+@Getter(AccessLevel.PACKAGE)
+public class KafkaDeserializerExtractor extends KafkaExtractor<Object, Object> {
+
+  public static final String KAFKA_DESERIALIZER_TYPE = "kafka.deserializer.type";
+  public static final String KAFKA_DESERIALIZER_CLASS = "kafka.deserializer.class";
```
zliu41: Using two keys can be confusing, it's not clear which one takes precedence.
Can you combine these two? You can make use of `Enums.getIfPresent`

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaDeserializerExtractor.java
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Constructs a {@link Deserializer}, first looking at the value of {@link #KAFKA_DESERIALIZER_TYPE}, if not set it
+   * uses {@link #KAFKA_DESERIALIZER_CLASS}.
+   */
+  private static Deserializer<?> getDeserializer(Properties props) throws ReflectiveOperationException {
+    Deserializer<?> deserializer;
+    if (props.containsKey(KAFKA_DESERIALIZER_TYPE)) {
+      deserializer = ConstructorUtils
+          .invokeConstructor(Deserializers.valueOf(props.getProperty(KAFKA_DESERIALIZER_TYPE)).getDeserializerClass());
```
zliu41: It's better to add `toUpperCase` here

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleExtractor.java
```diff
@@ -26,8 +27,11 @@
  */
 public class KafkaSimpleExtractor extends KafkaExtractor<String, byte[]> {
```
zliu41: Can this class and `KafkaSimpleJsonExtractor` be deprecated now that there is `KafkaDeserializerExtractor`?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaDeserializerExtractor.java
```diff
@@ -0,0 +1,193 @@
...
+  KafkaDeserializerExtractor(WorkUnitState state, Deserializer<?> kafkaDeserializer,
+      KafkaSchemaRegistry<?, ?> kafkaSchemaRegistry) {
+    super(state);
+    this.kafkaDeserializer = kafkaDeserializer;
+    this.kafkaSchemaRegistry = kafkaSchemaRegistry;
+  }
+
+  @Override
+  protected Object decodeRecord(MessageAndOffset messageAndOffset) throws IOException {
+    return this.kafkaDeserializer.deserialize(this.topicName, getBytes(messageAndOffset.message().payload()));
```
zliu41: How can `deserialize` work if it doesn't take the schema of the message payload as a parameter?

sahilTakiar: Since the `Deserializer` interface doesn't allow passing in a schema registry object, it is up to the `Deserializer` implementation to create a schema registry using some configuration properties.

So in the case of the Confluent `KafkaAvroDeserializer` it will construct and use its own schema registry. Separate from the one used in the `KafkaDeserializerExtractor`. So two different schema registries will be created, which is unfortunately unavoidable in this scenario.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleExtractor.java
sahilTakiar: Yes, done

sahilTakiar:
@zliu41 updated and comments addressed

@abti and @shirshanka care to review?

zliu41:
LGTM

build.gradle
```diff
@@ -134,6 +134,7 @@ if (!project.hasProperty('bytemanVersion')) {
 ext.avroVersion = '1.7.7'
 ext.dropwizardMetricsVersion = '3.1.0'
 ext.findBugsVersion = '3.0.0'
+ext.confluentVersion = '2.0.1'
```
shirshanka: @sahilTakiar Sorry for the late review: why 2.0.1 ? versus 3.0.x

abti:
sorry to have missed this earlier, but since this and #1076 has some overlap I ended up reviewing the classes here too and it looks good

PR #1017, opened 2016-05-26
Fix permissions for distcp-hive. Fix computation of target paths in d…

ibuenros:
…istcp-hive.

This is mostly refactoring of the target path computation logic, previously located in `HiveCopyEntityHelper` it is now in its own class `HiveTargetPathHelper`. The logic is almost the same, except for a fix in the case where new table root is specified.

Also, Hive copy will now specify the desired ancestor owner and permissions (minor change in `HiveCopyEntityHelper`)

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -53,6 +53,7 @@ import gobblin.configuration.State;
 import gobblin.data.management.copy.CopyConfiguration;
 import gobblin.data.management.copy.CopyEntity;
 import gobblin.data.management.copy.CopyableFile;
+import gobblin.data.management.copy.OwnerAndPermission;
 import gobblin.data.management.copy.entities.PostPublishStep;
 import gobblin.data.management.copy.entities.PrePublishStep;
 import gobblin.data.management.copy.hive.avro.HiveAvroCopyEntityHelper;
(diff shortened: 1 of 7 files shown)


ibuenros:
@tuGithub can you review? Thanks.

ibuenros:
@ydai1124 can you review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveTargetPathHelper.java
```diff
@@ -0,0 +1,185 @@
...
+   * </p>
+   *
+   *
+   * @param sourcePath Source path to be transformed.
+   * @param targetFs target {@link FileSystem}
+   * @param partition partition this file belongs to.
+   * @param isConcreteFile true if this is a path to an existing file in HDFS.
+   */
+  public Path getTargetPath(Path sourcePath, FileSystem targetFs, Optional<Partition> partition, boolean isConcreteFile) {
+    if (this.relocateDataFiles) {
```
ydai1124: Are both `RELOCATE_DATA_FILES_KEY` and `COPY_TARGET_TABLE_ROOT` necessary? Can you only keep `COPY_TARGET_TABLE_ROOT`, and relocate the files if `targetTableRoot` is present?

ibuenros: They do different things. If `COPY_TARGET_TABLE_ROOT` and `RELOCATE_DATA_FILES`, then files will be places directly under the new table root (it will look very much like a Hive managed table), while if `RELOCATE_DATA_FILES` is false, we will respect the relative paths of each file, just replace the old table root by the new table root. If both are false, then we will preserve the paths exactly.

ydai1124:
LGTM

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -889,8 +802,20 @@ private static void checkPartitionCompatibility(Partition desiredTargetPartition
         referenceAuthority = uri.getAuthority();
       }
 
+      if (!this.dataset.getTableRootPath().isPresent()) {
```
tuGithub: One performance concern: if there are a lot of files and they are in the same sub directory, if the first file's ancestorOwnerAndPermission been calculated, all other files should not need to re-calculate the ancestorOwnerAndPermission, right?

ibuenros: The big performance issue is the `getFileStatus` call, but that call is cached. We might want to optimize in the future, for now, this is how things have been running so far (this was just refactoring), so I don't have big concerns with performance.

tuGithub:
LGTM

PR #1019, opened 2016-05-30
Fix NPE and mutability of a map in hive target path helper and copy h…

ibuenros:
…elper.


ibuenros:
Minor bug fixes. Self-approving.

PR #1021, opened 2016-05-31
Parallelize commit of datasets.

ibuenros:


ibuenros:
@chavdar @ydai1124 can you review?

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -97,6 +97,8 @@
   public static final String DEFAULT_FORK_OPERATOR_CLASS = "gobblin.fork.IdentityForkOperator";
   public static final String JOB_COMMIT_POLICY_KEY = "job.commit.policy";
   public static final String DEFAULT_JOB_COMMIT_POLICY = "full";
+  public static final String PARALLELIZE_DATASET_COMMIT = "job.commit.parallelize";
```
chavdar: Please add documentation to the new property.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -345,26 +355,48 @@ void finalizeJobStateBeforeCommit() {
...
 
-    if (!allDatasetsCommit) {
-      this.jobState.setState(JobState.RunningState.FAILED);
-      throw new IOException("Failed to commit dataset state for some dataset(s) of job " + this.jobId);
-    }
     this.jobState.setState(JobState.RunningState.COMMITTED);
   }
 
+  private int numCommitThreads(JobState jobState) {
+    return jobState.getPropAsBoolean(ConfigurationKeys.PARALLELIZE_DATASET_COMMIT,
```
chavdar: Configuration properties should be accessed only in the constructor.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -345,26 +355,48 @@ void finalizeJobStateBeforeCommit() {
...
+          new IteratorExecutor<>(Iterables.transform(this.datasetStatesByUrns.get().entrySet(),
+          new Function<Map.Entry<String, DatasetState>, Callable<Void>>() {
+            @Nullable
+            @Override
+            public Callable<Void> apply(@Nullable final Map.Entry<String, DatasetState> entry) {
+              return new Callable<Void>() {
+                @Override
+                public Void call()
+                    throws Exception {
+                  processDatasetCommit(shouldCommitDataInJob, deliverySemantics, entry.getKey(), entry.getValue());
```
chavdar: Currently, processDatasetCommit is an instance method. I think it is very dangerous to call such method unless they are explicitly called out as thread-safe. Can you make an immutable class DatasetCommitter or something for which we can be certain it is thread-safe.

chavdar: Please also add some unit tests, e.g. to verify if failures are handled correctly.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -345,26 +355,48 @@ void finalizeJobStateBeforeCommit() {
...
+              };
+            }
+          }).iterator(), numCommitThreads, ExecutorsUtils.newThreadFactory(Optional.of(this.logger), Optional.of("Commit-thread-%d")))
+          .executeAndGetResults();
+      if (!IteratorExecutor.verifyAllSuccessful(result)) {
+        this.jobState.setState(JobState.RunningState.FAILED);
+        throw new IOException("Failed to commit dataset state for some dataset(s) of job " + this.jobId);
+      }
+    } catch (InterruptedException exc) {
+      throw new IOException(exc);
```
chavdar: nit: this looks more like RuntimeException rather than an IOException.

PR #1023, opened 2016-06-01
Adding parallel test Travis VMs

chavdar:
These changes should create a matrix of Travis VMs so we can run test in parallel and afford more logging.


chavdar:
@ibuenros  can you review?

chavdar:
@kadaan what do you think about this PR?

kadaan:
Seems good, but will make way more builds.

PR #1027, opened 2016-06-02
Added configuration to Hive client for getting credentials.

ibuenros:


ibuenros:
Trivial change, self approving

PR #1030, opened 2016-06-05
Fix a try-catch block in Hive dataset finder.

ibuenros:


ibuenros:
trivial change. Self approving.

PR #1035, opened 2016-06-08
Use table schema and partition schema

pcadabam:
### Changes ###
- Create schema aware extensions to hive standard ```Table``` and ```Partition``` classes
- Create serializable classes for  hive standard ```Table``` and ```Partition```s to be used in ```WorkUnit```s
- ```AvroSchemaManager``` is added which creates schema files for partitions and tables. The schema url is serialized in the Workunit instead of the entire schema.

@abti can you review?


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
```diff
@@ -126,24 +139,27 @@
...
-          List<Partition> sourcePartitions = HiveUtils.getPartitions(hiveDataset.getClientPool().getClient().get(),
-              hiveDataset.getTable(), Optional.<String> absent());
+          List<Partition> sourcePartitions =
+              HiveUtils.getPartitions(hiveDataset.getClientPool().getClient().get(), hiveDataset.getTable(),
+                  Optional.<String> absent());
 
           for (Partition sourcePartition : sourcePartitions) {
             LongWatermark lowWatermark = watermaker.getPreviousHighWatermark(sourcePartition);
+            if (globalLowWatermark.isPresent()) {
+              lowWatermark = globalLowWatermark.get();
```
abti: This will force all datasets to use low water mark even if they have a low water mark and are not bootstrapping / backfilling. 
A way to indicate which datasets the low water mark should be forced upon would be ideal

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/HiveAvroToOrcConverter.java
```diff
@@ -76,8 +72,8 @@ public Schema convertSchema(Schema inputSchema, WorkUnitState workUnit) throws S
     // TODO: Add num of buckets (from config)
 
     // Avro table name and location
-    String avroTableName = conversionEntity.getHiveUnit().getTableName();
-    String avroTableLocation = conversionEntity.getHiveUnit().getLocation().get();
+    String avroTableName = conversionEntity.getHiveTable().getTableName();
+    String avroTableLocation = conversionEntity.getHiveTable().getSd().getLocation();
```
abti: In case that converter is running for a partition, the location will be wrong (of table). We should check if this conversion entity has a partition, if so use its location 
Also, the avroTableLocation can be changed to avroDataLocation or similar, since it can be table or partition location based on what the conversion is happening

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
```diff
@@ -39,54 +45,74 @@
...
   @Override
   public Schema getSchema() throws IOException {
-    return this.hiveAvroSchemaProvider.getSchema(this.hiveUnits.get(0));
-  }
+    if (this.conversionEntities.isEmpty()) {
+      return null;
+    }
 
+    QueryBasedHiveConversionEntity conversionEntity = this.conversionEntities.get(0);
+    if (conversionEntity.getHivePartition().isPresent()) {
```
abti: We want to trust the table schema as the most evolved / latest schema. So, shouldn't we be returning table schema for partition as well?

abti: This will ensure that the older as well as newer partitions both are consistent, and we can handle evolution when we add support for it

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
pcadabam: Yes this is a global setting across all datasets. Dataset level low watermark will be added with config store support.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/HiveAvroToOrcConverter.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
pcadabam: Sure makes sense.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
abti: Cool

abti:
Other than above comments, LGTM

abti:
LGTM

PR #1038, opened 2016-06-09
Made Hive metastore connection auto returnable to connection pool aft…

abti:
…er Hive dataset discovery


abti:

PR #1043, opened 2016-06-09
Fix includes order

pcadabam:
The includes returned are used to build a fallback chain.
With the natural order, if a key found in the first include it is not be overriden by the next include.
By reversing the list, the Typesafe fallbacks are constructed bottom up.

@ibuenros can you review?


ibuenros:
LGTM

PR #1044, opened 2016-06-09
Made HiveAvroToOrcConverter compatible with Hive v0.13 version

abti:
Made HiveAvroToOrcConverter compatible with Hive v0.13 version


abti:
Minor change. Self approving.

PR #1046, opened 2016-06-09
[Avro to ORC] Mark all workunits of a dataset as failed if one task fails

pcadabam:
In the Avro to Orc conversion we use gobblin `DATASET_URN` so Gobblin runtime publishes one dataset at a time by calling `public void publishData(Collection<? extends WorkUnitState> states)` for workunits of each dataset. Since we currently use a partition level watermark all workunits of a dataset need to be failed if any one of them fail.
@abti can you review


abti:
LGTM

PR #1047, opened 2016-06-10
Fixed findbug issues with data management

abti:
Fixed findbug issues with data management


abti:
Minor code optimizations for fixbug. Self approving.

PR #1052, opened 2016-06-11
Fixed test resources for fixing test cases

abti:
Minor fix to reflect updated DDL statements


abti:
Minor change, self approving.

PR #1054, opened 2016-06-15
Converted Hive dereg / registration to post publish steps, fixed miss…

ibuenros:
…ing fileset.

This PR fixes two issues:
* For non-partitioned hive tables, the Hive registration and the file publishing were different file sets, so they could happen out of order.
* Made all table/partition deregistration and registration post-publish steps (used to be pre-publish steps), this ensures that the data is available as soon as the new table gets swapped.


ibuenros:
@ydai1124 can you review?

ydai1124:
LGTM

PR #1055, opened 2016-06-17
Parallelize commit rebased

ibuenros:
Guide to this PR:
* Created a class `SafeDatasetCommit`. Refactored a lot of the commit logic from `JobContext` to this class.
* `JobContext` creates an `IteratorExecutor` to execute dataset commits in parallel.
* Added a `isThreadSafe` method to `DataPublisher`. If this is false (default), Gobblin will automatically fall back to serial commit even if user choses parallel commit.
* Added a unit test to ensure `JobContext` handles publisher failure semantics correctly.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
@@ -15,9 +15,14 @@ package gobblin.runtime;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
+import javax.annotation.Nullable;
 import lombok.Getter;
 
 import org.apache.hadoop.conf.Configuration;
@@ -26,11 +31,13 @@ import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
 import com.google.common.eventbus.Subscribe;
 import com.google.common.io.Closer;
@@ -58,8 +65,11 @@ import gobblin.runtime.commit.FsCommitSequenceStore;
 import gobblin.runtime.util.JobMetrics;
 import gobblin.source.Source;
 import gobblin.source.extractor.JobCommitPolicy;
+import gobblin.util.Either;
+import gobblin.util.ExecutorsUtils;
 import gobblin.util.HadoopUtils;
 import gobblin.util.JobLauncherUtils;
+import gobblin.util.executors.IteratorExecutor;
 
 
 /**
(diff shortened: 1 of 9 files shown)


ibuenros:
@pcadabam @ydai1124 can you review?

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -344,109 +365,61 @@ void finalizeJobStateBeforeCommit() {
...
-    } catch (ReflectiveOperationException roe) {
-      this.logger.warn("Unable to find publisher class: " + roe, roe);
-      success = false;
-    }
-
-    if (!success) {
-      return false;
-    }
+  private int numCommitThreads() {
+    return this.parallelizeCommit ? 20 : 1;
```
ydai1124: Should number of threads be configurable with a default value?

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
```diff
@@ -0,0 +1,275 @@
...
+    Optional<CommitSequence.Builder> commitSequenceBuilder = Optional.absent();
+    try (Closer closer = Closer.create()) {
+      if (this.shouldCommitDataInJob) {
+        log.info(String.format("Committing dataset %s of job %s with commit policy %s and state %s", this.datasetUrn,
+            this.jobContext.getJobId(), this.jobContext.getJobCommitPolicy(), this.datasetState.getState()));
+        if (this.deliverySemantics == DeliverySemantics.EXACTLY_ONCE) {
+          generateCommitSequenceBuilder(this.datasetState);
+        } else {
+          DataPublisher publisher = closer.register(DataPublisher.getInstance(dataPublisherClass, this.jobContext.getJobState()));
+          if (this.isMultithreaded && !publisher.isThreadSafe()) {
```
ydai1124: Why not move this check to `JobContext`, which is aware of publisher type as well? You can decide the real number of threads in `JobContext`, rather than adding `synchronized` committing here.

ibuenros: `JobContext` is aware of the publisher class, but does not have an instance of that class. `isThreadSafe` is an instance method (and should remain so, otherwise it can't be overriden), so `JobContext` can't actually figure out if the publisher is thread safe.

gobblin-runtime/src/test/java/gobblin/runtime/JobContextTest.java
```diff
@@ -0,0 +1,270 @@
...
+      if (this.successPredicate.apply(datasetUrn)) {
+        callable = new ControllableCallable<>(Either.<Void, Exception>left(null), datasetUrn);
+      } else {
+        callable = new ControllableCallable<>(Either.<Void, Exception>right(new RuntimeException("Fail!")), datasetUrn);
+      }
+      this.callablesQueue.add(callable);
+      return callable;
+    }
+  }
+
```
ydai1124: Remove empty lines.

gobblin-utility/src/main/java/gobblin/util/executors/IteratorExecutor.java
```diff
@@ -87,4 +105,56 @@ public IteratorExecutor(Iterator<Callable<T>> runnableIterator, int numThreads,
...
+    });
+  }
+
+  /**
+   * Utility method unpacks the {@link Either} results from {@link IteratorExecutor} to the raw results of the generic
+   * type.
+   * @return List of {@link Callable} results.
+   * @throws ExecutionException if any {@link Callable} failed.
+   */
+  public static <T> List<T> unpackResults(List<Either<T, ExecutionException>> results) throws ExecutionException {
```
ydai1124: Is this used? Also, seems to me that the real content of `results` can be left for users to handle.

ydai1124:
Overall looks good. A few comments.

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
ydai1124: I see. that's true.

ibuenros:
@ydai1124 addressed comments.

ydai1124:
LGTM

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
```diff
@@ -0,0 +1,275 @@
...
+   * Finalize a given {@link JobState.DatasetState} before committing the dataset.
+   *
+   * This method is thread-safe.
+   */
+  private void finalizeDatasetStateBeforeCommit(JobState.DatasetState datasetState) {
+    for (TaskState taskState : datasetState.getTaskStates()) {
+      if (taskState.getWorkingState() != WorkUnitState.WorkingState.SUCCESSFUL
+          && this.jobContext.getJobCommitPolicy() == JobCommitPolicy.COMMIT_ON_FULL_SUCCESS) {
+        // The dataset state is set to FAILED if any task failed and COMMIT_ON_FULL_SUCCESS is used
+        datasetState.setState(JobState.RunningState.FAILED);
```
chavdar: nit: Both the next two lines and the ones on 175-176 lines can be encapsulated in methods to make sure that the state is updated consistently.

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
```diff
@@ -0,0 +1,275 @@
...
+      // Backoff the actual high watermark to the low watermark for each task that has not been committed
+      if (taskState.getWorkingState() != WorkUnitState.WorkingState.COMMITTED) {
+        taskState.backoffActualHighWatermark();
+        if (this.jobContext.getJobCommitPolicy() == JobCommitPolicy.COMMIT_ON_FULL_SUCCESS) {
+          // Determine the final dataset state based on the task states (post commit) and the job commit policy.
+          // 1. If COMMIT_ON_FULL_SUCCESS is used, the processing of the dataset is considered failed if any
+          //    task for the dataset failed to be committed.
+          // 2. Otherwise, the processing of the dataset is considered successful even if some tasks for the
+          //    dataset failed to be committed.
+          datasetState.setState(JobState.RunningState.FAILED);
```
chavdar: case in point: finalizeDatasetStateBeforeCommit() calls "datasetState.incrementJobFailures();" while this does not.

chavdar: I don't even think this is necessary as the  finalizeDatasetStateBeforeCommit() should have already set the dataset state.

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
ibuenros: Note all of this code is just moved from `JobContext`. I can fix it if necessary, but in principle this is outside of the scope of this PR.

chavdar:
LGTM. I have created ETL-4633 as a follow-up for the above code re-factorings.

PR #1056, opened 2016-06-19
Add lastDataPublishTime in hive table/partition properties

pcadabam:
### Changes
- Added ```lastDataPublishTime``` property in hive table properties and partition properties.
- This property store the time when new data was published into a partition or a table.
- Currently there is no way to know when new data gets added to an existing partition, this property will allow applications to use as a watermark to process new data for a partition.
- It provides a guarantee that no new data has been added since this time as the timestamp used is greater than the actual publish time which happens in the task publisher. 

@zliu41 , @ydai1124 can you review?
Spoke to @zliu41 offline about the pr.


zliu41:
LGTM

PR #1057, opened 2016-06-20
AvroToOrc bug fix to not find update time for partitions older than lookback

pcadabam:
@abti can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
@@ -121,7 +121,7 @@ public class HiveSource implements Source {
           if (HiveUtils.isPartitioned(hiveDataset.getTable())) {
             createWorkunitsForPartitionedTable(hiveDataset, client, expectedDatasetHighWatermark);
           } else {
-            createPartitionForNonPartitionedTable(hiveDataset, expectedDatasetHighWatermark);
+            createWorkunitForNonPartitionedTable(hiveDataset, expectedDatasetHighWatermark);
           }
         }
       }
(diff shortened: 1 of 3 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
```diff
@@ -230,6 +220,25 @@ private void createWorkunitsForPartitionedTable(HiveDataset hiveDataset,
     }
   }
 
+  @VisibleForTesting
+  public boolean shouldCreateWorkunitForPartition(Partition partition, long updateTime, LongWatermark lowWatermark) {
```
abti: You don't need partition as param for shouldCreateWorkUnitForPartition(), and when you remove that shouldCreateWorkUnitForPartition() and shouldCreateWorkUnitForTable() become exactly the same, so they can merged to be named shouldCreateWorkUnit()

Also, not a strong preference however: I liked that you had isOlderThanLookback() as part of the logic in shouldCreateWorkUnitForPartition() earlier. But feel free to keep it as is or as earlier

abti:
Minor comment. LGTM

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/HiveSource.java
pcadabam: Made the first change to use ```shouldCreateWorkunit``` instead of ```shouldCreateWorkUnitForPartition``` and ```shouldCreateWorkUnitForTable```. 
I did not move ```isOlderThanLookback``` into ```shouldCreateWorkunit```, I want to avoid reading the update time for partitions older than ```lookbackTime```

PR #1058, opened 2016-06-21
Fix findbugs with gobblin-cluster

abti:
Fix findbugs with gobblin-cluster


abti:
Minor change. Self approving.

PR #1059, opened 2016-06-21
Make kafka buffer size configurable

hibikir:
As mentioned in the mailing list, the existing gobblin configuration means that if any topic configured to have a larger buffer size than the default of 1mb will silently fail to be consumed when a large message is read. The silence is mostly Kafka's fault, but Gobblin should be able to be to support topics with large messages through a configuration parameter.

I am sending this as a preliminary patch, expecting you guys might have problems with it, because to be fair, the situation is a bit tricky: The KafkaOldAPI is the piece that needs this configuration, but it receives no configuration information.

Given that this is theoretically different by topic (although for some reason, KafkaTopicMetaData won't produce this information), I thought it really should be handled as something done by topic, although, for the time being, I am feeding the data from the work unit.

For the call that we make to get the list of kafka topics, I am using the defaults instead, because it doesn't seem like a place where one would change message topics, and trickling that up to a point that can read the configuration seems like quite a few changes.

I'd be happy to take a different approach, but then I'd like guidance on what would make the PR approvable.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -371,6 +371,12 @@ public class ConfigurationKeys {
   public static final String ERROR_MESSAGE_UNDECODABLE_COUNT = "error.message.undecodable.count";
 
   /**
+   * Configuration properties used by both KakfaSource and KafkaExtractor
+   */
+  public static final String KAFKA_BUFFER_SIZE = "kafka.buffer.size";
+  public static final int DEFAULT_KAFKA_BUFFER_SIZE = 1024 * 1024;
+
+  /**
    * Configuration properties for source connection.
    */
   public static final String SOURCE_CONN_USE_AUTHENTICATION = "source.conn.use.authentication";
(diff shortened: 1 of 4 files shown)


ibuenros:
Thanks @hibikir . Is there any way you can warn the user when their buffer size is not large enough?

ibuenros:
@hibikir any updates on the possibility of warning the user if their buffer size is not large enough?

hibikir:
I do not know of how to tell if the buffer isn't large enough: Kafka won't
tell directly at least. We know we had configured message.max.bytes in the
kafka brokers to be a bit higher than the default, so we just tried the
change, and it worked. If there's a way to query the brokers or zookeeper
for the max size, that would work, but I am not aware of a way to do it
remotely and programatically.

shirshanka:
@hibikir: Thanks for taking this on. If you don't mind, I'm going to make the entire KafkaWrapper parameter set configurable. Will cc you on the PR.

abti:
@shirshanka Are the changes in this PR included in your PR? If so, can we close this PR?

shirshanka:
@abti: Closing this PR because changes needed are already merged in at #1218

PR #1060, opened 2016-06-21
MR launcher does not write tasks to the jobstate file in HDFS.

ibuenros:
This PR reduces the size of job.state file by not serializing all of the tasks there, as the mappers read their tasks from separate files either way. This reduces required memory for the mappers.


ibuenros:
@ydai1124 can you review?

ydai1124:
LGTM

PR #1062, opened 2016-06-22
Enable AvroSchemaManager to read schema from Kafka schema registry

abti:
Enable AvroSchemaManager to read schema from Kafka schema registry if the Uri scheme is http(s) or fall back to HDFS

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
@@ -12,11 +12,14 @@
 package gobblin.data.management.conversion.hive;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
-
 import lombok.extern.slf4j.Slf4j;
+
 import org.apache.avro.Schema;
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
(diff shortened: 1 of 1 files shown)


abti:
@pcadabam Please review

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
```diff
@@ -12,11 +12,14 @@
...
 import java.io.IOException;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
-
 import lombok.extern.slf4j.Slf4j;
+
 import org.apache.avro.Schema;
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
```
pcadabam: Can you update the class javadoc for AvroSchemaManager with this new case?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
```diff
@@ -110,7 +113,22 @@ public void cleanupTempSchemas() throws IOException {
   }
 
   public static Schema getSchemaFromUrl(Path schemaUrl, FileSystem fs) throws IOException {
```
pcadabam: I think the argument should not be a ```Path```. It should be changed to a ```URI```
Add javadoc stating the schemaUrl can be with or without scheme and it defaults to hdfs.

pcadabam: Correction: Changing this to ```URI``` will need wider refactor as ```SerializableHivePartition``` and ```SerializableHiveTable``` use ```Path``` as schemaUrl. 
I suggest moving this logic to read from http/https into ```getOrGenerateSchemaFile```. So we would be creating a temp schema file on HDFS by reading the schema from registry. This way we are consistent with other cases and we guarantee schema used in source is same as in extractor.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
abti: As discussed with change in the logic location, its not needed anymore

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/AvroSchemaManager.java
abti: As discussed moved it to other location

pcadabam:
LGTM

PR #1063, opened 2016-06-22
Escape Hive queries column, partition and tables

abti:
Escape Hive queries column, partition and tables to avoid pitfall against users making use of Hive reserver words for column / partition names


abti:
Minor change. Self approving

PR #1064, opened 2016-06-22
Disable flattening for complex types other than nested records and union

abti:
Disable flattening for complex types other than nested records and union, since ORC does not support predicate pushdown for complex types, so even if we flatten nested structure within a map or array - it will not help and complicates conversion process.


abti:
Minor change. Self approving.

PR #1066, opened 2016-06-23
Config file monitor for general file system

autumnust:


ibuenros:
Thanks @autumnust 
I see that you took the apache file monitor and modified to work with `Path` instead. It's a great start, but we probably need to modify it to be more Gobblin-like in design (for example using safer `ThreadPool`s, using scheduled thread instead of `Thread.sleep`, etc.). The functionality will end up being very similar, so any tests for this will likely be useful for the refactoring as well.
Maybe a nice starting point is for you to document the current architecture of the monitor (couple paragraphs, maybe a nice diagram), so we can think from there how to refactor it.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -363,11 +361,11 @@ public void runJob(Properties jobProps, JobListener jobListener, JobLauncher job
   }
 
   /**
-   * Schedule Gobblin jobs configured in general position
+   * Schedule Gobblin jobs in general position
    */
   private void scheduleGeneralConfiguredJobs()
       throws ConfigurationException, JobException, IOException {
-    LOG.info("Scheduling configured jobs");
+    LOG.info("Scheduling locally configured jobs");
```
ibuenros: Why are these locally configured?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -404,46 +402,44 @@ private void scheduleGeneralConfiguredJobs()
...
       throws Exception {
-    LOG.info("Monitor debugging: " + this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY));
-    final File jobConfigFileDir = new File(this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY));
-    FileAlterationListener listener = new FileAlterationListenerAdaptor() {
-      /**
-       * Called when a new job configuration file is dropped in.
-       */
+    final Path jobConfigFileDirPath =
+        new Path(this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY));
+    PathAlterationListener listener = new PathAlterationListenerAdaptor() {
```
ibuenros: Can you refactor this into an actual class (we shouldn't use long anonymous classes)

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -404,46 +402,44 @@ private void scheduleGeneralConfiguredJobs()
...
         }
       }
 
       /**
        * Called when a job configuration file is changed.
        */
       @Override
-      public void onFileChange(File file) {
-        String fileExtension = Files.getFileExtension(file.getName());
+      public void onFileChange(Path path) {
```
ibuenros: Is there an `onFileDelete` method?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -464,13 +460,14 @@ public void onFileChange(File file) {
         }
 
         try {
-          LOG.info("Detected change to job configuration file " + file.getAbsolutePath());
-          Properties jobProps = SchedulerUtils.loadJobConfig(JobScheduler.this.properties, file, jobConfigFileDir);
+          LOG.info("Detected change to job configuration file " + path.toString());
```
ibuenros: Are you checking whether the extension is in `jobConfigFileExtensions`?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -404,46 +402,44 @@ private void scheduleGeneralConfiguredJobs()
...
+    final Path jobConfigFileDirPath =
+        new Path(this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY));
+    PathAlterationListener listener = new PathAlterationListenerAdaptor() {
       @Override
-      public void onFileCreate(File file) {
-        String fileExtension = Files.getFileExtension(file.getName());
-        if (!JobScheduler.this.jobConfigFileExtensions.contains(fileExtension)) {
-          // Not a job configuration file, ignore.
+      public void onFileCreate(Path path) {
+        String fileExntensionFromPath = path.getName().substring(path.getName().lastIndexOf('.') + 1);
```
ibuenros: There is a lot of code repetition between here and `onFileChange`, can they be refactored a bit?

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationMonitor.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+
+public final class PathAlterationMonitor implements Runnable {
+  private final long interval;
+  private volatile boolean running = false;
+//  private Thread thread = null;
```
ibuenros: Remove commented lines.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationMonitor.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+
+public final class PathAlterationMonitor implements Runnable {
```
ibuenros: Please add javadoc.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationMonitor.java
```diff
@@ -0,0 +1,131 @@
...
+  private final long interval;
+  private volatile boolean running = false;
+//  private Thread thread = null;
+  private ScheduledExecutorService executor;
```
ibuenros: should be final

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationMonitor.java
```diff
@@ -0,0 +1,131 @@
...
+  private int initialDelay = 0;
+
+  public PathAlterationMonitor() {
+    this(10000);
+    this.executor = Executors.newScheduledThreadPool(1);
+  }
+
+  public PathAlterationMonitor(final long interval) {
+    this.interval = interval;
+    this.executor = Executors.newScheduledThreadPool(1);
```
ibuenros: Please use a `ThreadFactory` obtained from `ExecutorsUtils.newDaemonThreadFactory`

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationMonitor.java
```diff
@@ -0,0 +1,131 @@
...
+    executionResult.cancel(true) ;
+    running = false;
+    for (final PathAlterationObserver observer : observers) {
+      observer.destroy();
+    }
+  }
+
+  @Override
+  public void run() {
+    while (running) {
```
ibuenros: Shouldn't this run only once?

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,352 @@
...
+   * @param directory the directory to observe
+   * @param fileFilter The file filter or null if none
+   * @param caseSensitivity  what case sensitivity to use comparing file names, null means system sensitive
+   */
+  public PathAlterationObserver(final Path directory, final PathFilter fileFilter, final IOCase caseSensitivity)
+      throws IOException {
+    this(new PathEntry(directory), fileFilter, caseSensitivity);
+  }
+
+  public PathAlterationObserver(final PathEntry rootEntry, final PathFilter pathFilter, final IOCase caseSensitivity)
```
ibuenros: Let's not worry about case sensitivity. In Hadoop everything is case sensitive, it will reduce the necessary code.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,352 @@
...
+      listener.onStart(this);
+    }
+
+    /* fire directory/file events */
+    final Path rootPath = rootEntry.getPath();
+
+    if (rootEntry.isExists()) {
+      checkAndNotify(rootEntry, rootEntry.getChildren(), listPaths(rootPath));
+    } else if (rootEntry.isExists()) {
+      checkAndNotify(rootEntry, rootEntry.getChildren(), EMPTY_PATH_ARRAY);
```
ibuenros: why is this needed?

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,352 @@
...
+   * List the path in the format of PathEntry array
+   * @param path The path to list files for
+   * @param entry the parent entry
+   * @return The child files
+   */
+  private PathEntry[] doListPathsEntry(Path path, PathEntry entry)
+      throws IOException {
+    // Create path array by listing all file status.
+    ArrayList<Path> result = new ArrayList<>();
+    final FileStatus[] tmpFileStatusForPaths = fs.listStatus(path);
```
ibuenros: `tmpFileStatusForPaths` may be null.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,352 @@
...
+    for (FileStatus tmpFileStatusForPath : tmpFileStatusForPaths) {
+      result.add(tmpFileStatusForPath.getPath());
+    }
+    final Path[] paths = result.toArray(new Path[result.size()]);
```
ibuenros: why must `paths` be an array?

gobblin-utility/src/main/java/gobblin/util/filesystem/AbstractPathComparator.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+import org.apache.hadoop.fs.Path;
+
+
+abstract class AbstractPathComparator implements Comparator<Path> {
```
ibuenros: You don't need this anymore when not using case sensitivity. `Path` is comparable.

gobblin-utility/src/main/java/gobblin/util/filesystem/PathComparator.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import org.apache.commons.collections.comparators.ReverseComparator;
+import org.apache.commons.io.IOCase;
+import org.apache.hadoop.fs.Path;
+
+
+public class PathComparator extends AbstractPathComparator implements Serializable {
```
ibuenros: Not needed (see above)

gobblin-utility/src/main/java/gobblin/util/filesystem/PathEntry.java
```diff
@@ -0,0 +1,217 @@
+package gobblin.util.filesystem;
+
+import java.io.IOException;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+
+public class PathEntry {
```
ibuenros: This class is very similar to `FileStatus`, please just use `FileStatus` instead.

ibuenros:
Please also add tests for the `Path` change monitor.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
autumnust: I did this checking in line 461.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
autumnust: But is it only used in hdfs ?

autumnust:
It seems the `JobConfigFileMonitorTest` class already covered the test for Monitor's function.

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
```diff
@@ -147,9 +147,9 @@ private void fetchJobConf()
       if (jobConfigDir.exists()) {
         LOGGER.info("Loading job configurations from " + jobConfigDir);
         final Properties properties = new Properties();
-        properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY, jobConfigDir.getAbsolutePath());
+        properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY, jobConfigDir.getAbsolutePath());
```
abti: Please also change the config property name in conf/aws/application.conf

gobblin-runtime/build.gradle
```diff
@@ -103,6 +103,14 @@ artifacts {
 test {
     workingDir rootProject.rootDir
     maxParallelForks = 1
+  /*
```
ibuenros: Is there a reason to leave these commented lines here? If not, please remove them.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
...
+      try {
+        Properties jobProps =
+            SchedulerUtils.loadGenericJobConfig(JobScheduler.this.properties, path, jobConfigFileDirPath);
+        System.err.println("New finding jobProps has size : " + jobProps.size());
+
+        switch (action) {
+          case SCHEDULE:
+            boolean runOnce = Boolean.valueOf(jobProps.getProperty(ConfigurationKeys.JOB_RUN_ONCE_KEY, "false"));
+            scheduleJob(jobProps, runOnce ? new RunOnceJobListener() : new EmailNotificationJobListener());
+            customizedInfo = "schedule";
```
ibuenros: It seems you should do this before running `scheduleJob`.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
...
+      String customizedInfoAction = "";
+      String customizedInfoResult = "";
+      try {
+        for (Properties jobProps : SchedulerUtils.loadGenericJobConfigs(JobScheduler.this.properties, path,
+            jobConfigFileDirPath)) {
+          try {
+            switch (action) {
+              case RESCHEDULE:
+                rescheduleJob(jobProps);
+                customizedInfoAction = "schedule";
```
ibuenros: Same as above.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
...
+        LOG.error(
+            "Failed to reload job configuration files affected by " + customizedInfoResult + " to " + path.toString(),
+            e);
+      }
+    }
+
+    @Override
+    public void onFileCreate(Path path) {
+
+      System.err.println("detected file creation ! ");
```
ibuenros: Remove debugging lines.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
...
+        }
+      } catch (ConfigurationException | IOException e) {
+        LOG.error(
+            "Failed to reload job configuration files affected by " + customizedInfoResult + " to " + path.toString(),
+            e);
+      }
+    }
+
+    @Override
+    public void onFileCreate(Path path) {
```
ibuenros: What if the new file is a JobProps file?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
...
+    /**
+     * Called when a job configuration file is deleted.
+     */
+    @Override
+    public void onFileDelete(Path path) {
+      String fileExtension = path.getName().substring(path.getName().lastIndexOf('.') + 1);
+      if (fileExtension.equalsIgnoreCase(SchedulerUtils.JOB_PROPS_FILE_EXTENSION)) {
+        LOG.info("Detected deletion to common properties file " + path.toString());
+        // Unschedule all the job using this common properties file
+        loadNewCommonConfigAndHandleNewJob(path, Action.UNSCHEDULE);
```
ibuenros: Does deleting a job props file imply unscheduling all jobs?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -388,7 +391,156 @@ private void scheduleGeneralConfiguredJobs()
   }
 
   /**
-   * Start the job configuration file monitor.
+   * Inner subclass of PathAlterationListenerAdaptor for implementation of Listen's methods,
+   * avoiding anonymous class
+   */
+  public class PathAlterationListenerAdaptorForMonitor extends PathAlterationListenerAdaptor {
```
ibuenros: Please move this class to the top level, it is rather long.

gobblin-utility/src/main/java/gobblin/util/PathUtils.java
```diff
@@ -177,4 +177,14 @@ public static void deleteEmptyParentDirectories(FileSystem fs, Path limitPath, P
...
     }
   }
+
+  /**
+   * Add a prefix into a path, like "file:" when comparing two Paths,
+   * to avoid the case when the actual path are the same while their comparison result turns out to be false.
+   * @param path
+   * @param prefix
+   */
+  public static Path addFilePrefixForComparison( Path path, String prefix ) {
```
ibuenros: I think the correct way is to strip both files of scheme and authority instead, e.g. `PathUtils.getPathWithoutSchemeAndAuthority(path1).equals(PahtUtils.getPathWithoutSchemeAndAuthority(path2))`

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -141,25 +112,27 @@ public boolean accept(Path path) {
...
-   * @param jobConfigFile job configuration file to be loaded
-   * @param jobConfigFileDir root job configuration file directory
+   * @param jobConfigPath job configuration file to be loaded
+   * @param jobConfigPathDir root job configuration file directory
    * @return a job configuration in the form of {@link java.util.Properties}
    */
-  public static Properties loadJobConfig(Properties properties, File jobConfigFile, File jobConfigFileDir)
+  public static Properties loadGenericJobConfig(Properties properties, Path jobConfigPath, Path jobConfigPathDir)
       throws ConfigurationException, IOException {
+    System.err.println("[loadGenericJobConfig]" +jobConfigPath.toString());
```
ibuenros: Remove log lines.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -141,25 +112,27 @@ public boolean accept(Path path) {
...
     List<Properties> commonPropsList = Lists.newArrayList();
-    getCommonProperties(commonPropsList, jobConfigFileDir, jobConfigFile.getParentFile());
+    getCommonProperties(commonPropsList, jobConfigPathDir, jobConfigPath.getParent());
     // Add the framework configuration properties to the end
     commonPropsList.add(properties);
+    System.err.println("Common properties size : " + commonPropsList.size());
```
ibuenros: Ditto.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -168,42 +141,56 @@ public static Properties loadJobConfig(Properties properties, File jobConfigFile
     }
 
     // Then load the job configuration properties defined in the job configuration file
-    jobProps.putAll(ConfigurationConverter.getProperties(new PropertiesConfiguration(jobConfigFile)));
-    jobProps.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_PATH_KEY, jobConfigFile.getAbsolutePath());
+    System.err.println("the path is : " + jobConfigPath);
```
ibuenros: Remove.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -168,42 +141,56 @@ public static Properties loadJobConfig(Properties properties, File jobConfigFile
...
-        return file.isDirectory();
-      }
-    });
+    Configuration conf = new Configuration();
+    try (final FileSystem fileSystem = rootDirPath.getFileSystem(conf)) {
+
+      // List subdirectories under the current root directory, obtain a Path array
+      FileStatus[] subDirsStatus = fileSystem.listStatus(rootDirPath, new PathFilter() {
+        @Override
+        public boolean accept(Path path) {
```
ibuenros: I don't understand what this `PathFilter` is intended to do.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,337 @@
...
+  public void checkAndNotify()
+      throws IOException {
+    /* fire onStart() */
+    for (final PathAlterationListener listener : listeners) {
+      listener.onStart(this);
+    }
+
+    /* fire directory/file events */
+    final Path rootPath = rootEntry.getPath();
+    System.err.println("checking rootPath:" + rootPath);
```
ibuenros: Remove, and next line as well.

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * Fire directory/file created events to the registered listeners.
+   *
+   * @param entry The file entry
+   */
+  private void doCreate(final FileStatusEntry entry) {
+    System.err.println("Trying to create " + entry.getPath() );
+    for (final PathAlterationListener listener : listeners) {
+      if (entry.isDirectory()) {
+        System.err.println("call OnCreate directory");
```
ibuenros: Remove

gobblin-utility/src/main/java/gobblin/util/fileSystem/PathAlterationObserver.java
```diff
@@ -0,0 +1,337 @@
...
+        listener.onDirectoryCreate(entry.getPath());
+      } else {
+        System.err.println("call OnCreate file ");
```
ibuenros: Remove

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,211 @@
...
+import org.apache.hadoop.fs.Path;
+
+
+public class FileStatusEntry extends FileStatus {
+
+  static final FileStatusEntry[] EMPTY_ENTRIES = new FileStatusEntry[0];
+
+  private final FileStatusEntry parent;
+  private FileStatusEntry[] children;
+  private final Path path;
```
ibuenros: The instance variables in lines 18-23 are unnecessary. They are already part of `FileStatus`.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
autumnust: In the original Yinan's implementation there's no such routine. 
I was wondering if we need this, as we may make any configuration file mutable later?

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,211 @@
...
+   */
+  public FileStatusEntry newChildInstance(final Path path) {
+    return new FileStatusEntry(this, path);
+  }
+
+  /**
+   * @return path The path object for this FileStatusEntry.
+   */
+  @Override
+  public Path getPath() {
```
ibuenros: Unnecessary.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
ibuenros: Yes, I think we need it. Also need to check in that case that there was not already a job props file in that directory.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
autumnust: Talked offline about the semantics of deletion.

gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
```diff
@@ -0,0 +1,209 @@
...
+import gobblin.runtime.listeners.RunOnceJobListener;
+import gobblin.util.SchedulerUtils;
+
+
+
+/**
+ * Inner subclass of PathAlterationListenerAdaptor for implementation of Listen's methods,
+ * avoiding anonymous class
+ */
+public class PathAlterationListenerAdaptorForMonitor extends gobblin.util.filesystem.PathAlterationListenerAdaptor {
```
ibuenros: Please import `gobblin.util.filesystem.PathAlterationListenerAdaptor` instead of using the fully qualified class.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
autumnust: To keep it consistent with `OnFilechange`, which doesn't contain routine for file change of `.properties` file. 
I can remove it now since file change of `.properties` file is possible as we discussed.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
autumnust: The reason why I put those fields is that they are private instead protected in `FileStatus` so that subclass don't actually inherit them.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
autumnust: Put them accordingly as reasons mentioned above.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,168 @@
...
+    this(null, path);
+  }
+
+  public FileStatusEntry(final FileStatusEntry parent, final Path path)
+      throws IOException {
+    if (path == null) {
+      throw new IllegalArgumentException("Path is missing");
+    }
+    this.parent = parent;
+    this._fileStatus = Optional.of(path.getFileSystem(new Configuration()).getFileStatus(path));
```
ibuenros: You should be setting `this.exists` here.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
ibuenros: This filter always returns `true`, it is not filtering anything.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,168 @@
...
+    final boolean origExists = exists;
+    if (_fileStatus.isPresent()) {
+      FileStatus oldStatus = _fileStatus.get();
+      Configuration conf = new Configuration();
+
+      try (FileSystem fileSystem = path.getFileSystem(conf)) {
+
+        // refresh the values
+        exists = fileSystem.exists(path);
+        this._fileStatus = Optional.of(fileSystem.getFileStatus(path));
```
ibuenros: Use `Optional.fromNullable` instead

ibuenros: Also, to save an rpc call switch the lines:
```java
this.fileStatus = Optional.fromNullable(fileSystem.getFileStatus(path));
this.exists = this.fileStatus.isPresent();
```

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,168 @@
...
+
+        if ( _fileStatus.isPresent() ) {
+          // Return if there are changes
+          return exists != origExists ||
```
ibuenros: Could you `return this.fileStatus.equals(oldStatus)` instead?

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,168 @@
...
+              _fileStatus.get().getModificationTime() != oldStatus.getModificationTime() ||
+              _fileStatus.get().isDirectory() != oldStatus.isDirectory() ||
+              _fileStatus.get().getLen() != oldStatus.getLen();
+        }else {
+          throw new NullPointerException("Path is missing ");
```
ibuenros: Why does this throw an NPE?

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,168 @@
...
+          return exists != origExists ||
+              _fileStatus.get().getModificationTime() != oldStatus.getModificationTime() ||
+              _fileStatus.get().isDirectory() != oldStatus.isDirectory() ||
+              _fileStatus.get().getLen() != oldStatus.getLen();
+        }else {
+          throw new NullPointerException("Path is missing "); 
+        }
+      }
+    } else {
+      throw new IllegalArgumentException("Path is missing");
```
ibuenros: If the path didn't exist recently, it could exist now. Shouldn't you check that?

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
autumnust: Talked offline about this.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
autumnust: Finally find the cause: In Yinan's original version, the filter is aiming at filtering subdirectories (indicated by the name of `subDirsStatus`).

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
autumnust: A follow-up question, do we always favor  `Optional.fromNullable` over ` Optional.of` ?

autumnust:
@ibuenros Addressed the comments and fix the bugs. Please take a look at it, thanks.

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -168,42 +139,61 @@ public static Properties loadJobConfig(Properties properties, File jobConfigFile
...
-      @Override
-      public boolean accept(File file) {
-        return file.isDirectory();
-      }
-    });
+    Configuration conf = new Configuration();
+    try (final FileSystem fileSystem = rootDirPath.getFileSystem(conf)) {
+
+      // List subdirectories under the current root directory, obtain a Path array
+      FileStatus[] subDirsStatus = fileSystem.listStatus(rootDirPath
```
ibuenros: Instead of doing this, do
```java
FileStatus[] subDirsStatus = fileSystem.listStatus(rootDirPath)
Iterables.filter(Lists.newArrayList(subDirsStatus), new Predicate<FileStatus> {
  public boolean apply(FileStatus status) {
     return status.isDir();
  }
});
```
This reduces the number of rpc calls significantly.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,158 @@
...
+    this.parent = parent;
+    this._fileStatus = Optional.fromNullable(path.getFileSystem(new Configuration()).getFileStatus(path));
+  }
+
+  public boolean refresh(final Path path)
+      throws IOException {
+
+    if (_fileStatus.isPresent()) {
+      Optional<FileStatus> oldStatus = this._fileStatus;
+      Optional<FileStatus> newStatus = Optional.of(path.getFileSystem(new Configuration()).getFileStatus(path));
```
ibuenros: This can throw `FileNotFoundException`, you should handle that correctly.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,158 @@
...
+      this.exists = newStatus.isPresent();
+      return (oldStatus.get().getModificationTime() != newStatus.get().getModificationTime()
```
ibuenros: add `oldStatus.isPresent() != this._fileStatus.isPresent()` at the beginning.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,158 @@
...
+          || oldStatus.get().isDirectory() != newStatus.get().isDirectory()
+          || oldStatus.get().getLen() != newStatus.get().getLen());
+    } else {
+      if (path.getFileSystem(new Configuration()).exists(path)) {
+        _fileStatus = Optional.of(path.getFileSystem(new Configuration()).getFileStatus(path));
```
ibuenros: You should reuse the `FileSystem` instead of creating a new one.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileStatusEntry.java
```diff
@@ -0,0 +1,167 @@
...
+        try {
+          Optional<FileStatus> newStatus = Optional.of(fs.getFileStatus(path));
+          this.exists = newStatus.isPresent();
+          return (oldStatus.isPresent() != this._fileStatus.isPresent()
+              || oldStatus.get().getModificationTime() != newStatus.get().getModificationTime()
+              || oldStatus.get().isDirectory() != newStatus.get().isDirectory() || oldStatus.get().getLen() != newStatus
+              .get()
+              .getLen());
+        }catch (FileNotFoundException e) {
+          _fileStatus.absent();
```
ibuenros:
```java
this.fileStatus = Optional.absent();
this.exists = false;
```

ibuenros:
@autumnust there are test failures (in data management)

PR #1067, opened 2016-06-24
Add a backfill hive source that does not check watermarks

pcadabam:
A backfill source will be used to re run avro to orc conversion for all partitions starting from ```lookBackDays```

@abti can you review?


abti:
LGTM

PR #1069, opened 2016-06-24
Adding security token to Salesforce bulk API login

chavdar:
Adding a security token for access from non-whitelisted IPs


chavdar:
Self-merging hotfix.

PR #1070, opened 2016-06-24
Initial commit for hive retention

ydai1124:
Implementation of hive retention. It uses `HiveCleanableDatasetFinder` to find `HiveCleanableDataset`. Each `HiveCleanableDataset` will contain `HiveTablePartitionVersions`. If a table is not partitioned, it will be single version. Otherwise, it will contain partitions as versions. 
Currently, there are two retention policies. One is `HiveTablePartitionInvalidDataLocRetentionPolicy`, which is based on if the data location exists on filesystem. Another one is `HiveTablePartitionTimeBasedRetentionPolicy`, which is based on some timestamp. Now the default behavior is based on modified timestamp. But it can be overridden. 

Tested on holdem, I have created RB: https://rb.corp.linkedin.com/r/756706/ for sample job configs I used.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.data.management.retention.dataset;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
+import org.apache.hadoop.hive.ql.metadata.Partition;
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.thrift.TException;
+
+import com.google.common.base.Function;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+
+import gobblin.data.management.copy.hive.HiveDataset;
+import gobblin.data.management.retention.policy.CombineRetentionPolicy;
+import gobblin.data.management.retention.policy.HiveTablePartitionInvalidDataLocRetentionPolicy;
+import gobblin.data.management.retention.policy.HiveTablePartitionTimeBasedRetentionPolicy;
+import gobblin.data.management.retention.policy.RetentionPolicy;
+import gobblin.data.management.version.HiveTablePartitionVersion;
+import gobblin.data.management.version.finder.HiveDatasetVersionFinder;
+import gobblin.data.management.version.finder.VersionFinder;
+import gobblin.hive.HiveMetastoreClientPool;
+import gobblin.util.AutoReturnableObject;
+import gobblin.util.Either;
+
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * Extends {@link HiveDataset} and also is cleanable. It uses {@link CombineRetentionPolicy} of {@link HiveTablePartitionTimeBasedRetentionPolicy}
+ * and {@link HiveTablePartitionInvalidDataLocRetentionPolicy} to decide the deletable versions.
+ */
+@Slf4j
+public class HiveCleanableDataset extends HiveDataset implements CleanableDataset {
+  private final VersionFinder<HiveTablePartitionVersion> versionFinder;
+  private final CombineRetentionPolicy<HiveTablePartitionVersion> retentionPolicy;
+  private final boolean simulate;
+
+  public HiveCleanableDataset(FileSystem fs, HiveMetastoreClientPool clientPool, Table table, Properties properties)
+      throws IOException {
+    super(fs, clientPool, table, properties);
+    this.versionFinder = new HiveDatasetVersionFinder(fs, properties);
+    List<RetentionPolicy<HiveTablePartitionVersion>> policies = Lists
+        .newArrayList(new HiveTablePartitionInvalidDataLocRetentionPolicy(fs),
+            new HiveTablePartitionTimeBasedRetentionPolicy(properties, fs));
+    this.retentionPolicy = new CombineRetentionPolicy<HiveTablePartitionVersion>(policies,
+        CombineRetentionPolicy.DeletableCombineOperation.UNION);
+    this.simulate = Boolean.valueOf(properties
+        .getProperty(MultiVersionCleanableDatasetBase.SIMULATE_KEY, MultiVersionCleanableDatasetBase.SIMULATE_DEFAULT));
+  }
+
+  @Override
+  public void clean()
+      throws IOException {
+    Iterable<Either<Table, Partition>> deletableTableOrPartitions = Iterables.transform(
+        this.retentionPolicy.listDeletableVersions(new ArrayList<>(this.versionFinder.findDatasetVersions(this))),
+        new Function<HiveTablePartitionVersion, Either<Table, Partition>>() {
+          @Override
+          public Either<Table, Partition> apply(HiveTablePartitionVersion input) {
+            return input.getVersion();
+          }
+        });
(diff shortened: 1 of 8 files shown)


ydai1124:
@ibuenros @pcadabam  Please take a look.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+ * Extends {@link HiveDataset} and also is cleanable. It uses {@link CombineRetentionPolicy} of {@link HiveTablePartitionTimeBasedRetentionPolicy}
+ * and {@link HiveTablePartitionInvalidDataLocRetentionPolicy} to decide the deletable versions.
+ */
+@Slf4j
+public class HiveCleanableDataset extends HiveDataset implements CleanableDataset {
+  private final VersionFinder<HiveTablePartitionVersion> versionFinder;
+  private final CombineRetentionPolicy<HiveTablePartitionVersion> retentionPolicy;
+  private final boolean simulate;
+
+  public HiveCleanableDataset(FileSystem fs, HiveMetastoreClientPool clientPool, Table table, Properties properties)
```
pcadabam: New classes should use typesafe ```Config``` instead of ```Properties```

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+      throws IOException {
+    super(fs, clientPool, table, properties);
+    this.versionFinder = new HiveDatasetVersionFinder(fs, properties);
+    List<RetentionPolicy<HiveTablePartitionVersion>> policies = Lists
```
pcadabam: This should be configurable

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+          public Either<Table, Partition> apply(HiveTablePartitionVersion input) {
+            return input.getVersion();
+          }
+        });
+    for (Either<Table, Partition> tableOrPartitionToDelete : deletableTableOrPartitions) {
+      this.cleanImpl(tableOrPartitionToDelete);
+    }
+  }
+
+  protected void cleanImpl(Either<Table, Partition> tableOrPartition)
```
pcadabam: Will this delete the data or an external table? I think that should be config to delete on not delete data

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/finder/HiveCleanableDatasetFinder.java
```diff
@@ -0,0 +1,41 @@
...
+import gobblin.data.management.copy.hive.HiveDataset;
+import gobblin.data.management.copy.hive.HiveDatasetFinder;
+import gobblin.data.management.retention.dataset.HiveCleanableDataset;
+
+
+/**
+ * Extends {@link HiveDatasetFinder} and finds {@link HiveCleanableDatasetFinder}.
+ */
+public class HiveCleanableDatasetFinder extends HiveDatasetFinder {
+  public HiveCleanableDatasetFinder(FileSystem fs, Properties properties)
```
pcadabam: Same as above change ```Properties``` to ```Config```

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/HiveTablePartitionInvalidDataLocRetentionPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+import gobblin.util.Either;
+
+import lombok.AllArgsConstructor;
+
+
+/**
+ * List deletable {@link HiveTablePartitionVersion}s by checking if the data location pointed by {@link Table} or {@link Partition} exists.
+ */
+@AllArgsConstructor
+public class HiveTablePartitionInvalidDataLocRetentionPolicy implements RetentionPolicy<HiveTablePartitionVersion> {
```
pcadabam: I though ```RetentionPolicy``` is deprecated in favor of ```SelectionPolicy```. Also can we call this ``HiveUnitNoDataSelectionPolicy```

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/HiveTablePartitionTimeBasedRetentionPolicy.java
```diff
@@ -0,0 +1,81 @@
...
+        throw new RuntimeException("Unable to determine deletable versions", e);
+      }
+    }
+    return deletableVersions;
+  }
+
+  public boolean shouldDeleteTableOrPartition(Either<Table, Partition> tableOrPartition)
+      throws IOException {
+    Path dataLoc =
+        tableOrPartition instanceof Either.Left ? ((Table) ((Either.Left) tableOrPartition).getLeft()).getDataLocation()
```
pcadabam: This part of code to find either table or partition seems to be duplicated several time. Can you model this as ```HiveVersion``` as a base class and ```HivePartitionVersion``` and ```HiveTableVersion``` extending ```HiveVersion``` ?

pcadabam:
While the policy should be runnable as a standalone job we should support running it via config store as well. It is difficult to refactor later to add that support.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+  private final CombineRetentionPolicy<HiveTablePartitionVersion> retentionPolicy;
+  private final boolean simulate;
+
+  public HiveCleanableDataset(FileSystem fs, HiveMetastoreClientPool clientPool, Table table, Properties properties)
+      throws IOException {
+    super(fs, clientPool, table, properties);
+    this.versionFinder = new HiveDatasetVersionFinder(fs, properties);
+    List<RetentionPolicy<HiveTablePartitionVersion>> policies = Lists
+        .newArrayList(new HiveTablePartitionInvalidDataLocRetentionPolicy(fs),
+            new HiveTablePartitionTimeBasedRetentionPolicy(properties, fs));
```
ibuenros: Is there any reason we can't use the general time based retention policy?

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/HiveCleanableDataset.java
```diff
@@ -0,0 +1,115 @@
...
+      throws IOException {
+    try (AutoReturnableObject<IMetaStoreClient> client = this.clientPool.getClient()) {
+      if (tableOrPartition instanceof Either.Right) {
+        Partition paritionToDrop = ((Partition) ((Either.Right) tableOrPartition).getRight());
+        if (this.simulate) {
+          log.info("Simulate mode: will delete partition " + paritionToDrop.getValues() + " for table " + paritionToDrop
+              .getTable().getTableName());
+        }
+
+        client.get().dropPartition(this.table.getDbName(), this.table.getTableName(), paritionToDrop.getValues(), true);
```
ibuenros: Should this be in an `else` clause? If I'm not reading wrong, right now simulate will log the message and then do the deletion either way.

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/HiveTablePartitionTimeBasedRetentionPolicy.java
ibuenros: There is a class `HiveRegistrationUnit` that is intended as a common superclass for `Table` and `Partition`, but it is designed mostly for registration, and doesn't have good access methods.
As an alternative, and in line with Pradhan's suggestion, I could see something like this working with the help of Lombok:
```java
public abstract class HiveEntity {
   public abstract Path getLocation();
   ...
   
   @AllArgsConstructor
   public class Table extends HiveEntity {
      @Delegate
      private final Table table;
   }

   @AllArgsConstructor
   public class Partition extends HiveEntity {
      @Delegate
      private final Partition partition;
   }
}
```

gobblin-data-management/src/main/java/gobblin/data/management/retention/policy/HiveTablePartitionTimeBasedRetentionPolicy.java
```diff
@@ -0,0 +1,81 @@
...
+import gobblin.data.management.version.DatasetVersion;
+import gobblin.data.management.version.HiveTablePartitionVersion;
+import gobblin.util.Either;
+
+
+/**
+ * List deletable {@link HiveTablePartitionVersion}s by checking if the modification timestamp of the data location older than currentTime minus {@link #retention}.
+ * This can be overridden by getting the time information from other ways, e.g., the datetime pattern along the path.
+ */
+public class HiveTablePartitionTimeBasedRetentionPolicy implements RetentionPolicy<HiveTablePartitionVersion> {
```
ibuenros: I think we also need a retention policy that uses information in the partition values itself. For example, if the partition value is `datepartition="2016-01-01"`, we should be able to run retention on that date instead of modtime.

PR #1072, opened 2016-06-27
Implemented a commit step database to reduce size of copy work units.

ibuenros:
- Implemented `CommitStepDB` class using Derby as backing store.
- Added tests for `CommitStepDB` and `CommitStepCopyEntity`.
- Changes required to use `CommitStepDB`.

### DIFF
--- a/gobblin-core/src/test/java/gobblin/configuration/StateTest.java
+++ b/gobblin-core/src/test/java/gobblin/configuration/StateTest.java
@@ -14,9 +14,12 @@ package gobblin.configuration;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.DataInput;
 import java.io.DataInputStream;
+import java.io.DataOutput;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.Map;
 
 import org.testng.Assert;
 import org.testng.annotations.Test;
@@ -119,5 +122,50 @@ public class StateTest {
     Assert.assertEquals(state2.getPropAsInt("int"), Integer.MIN_VALUE);
     Assert.assertEquals(state2.getPropAsDouble("double"), Double.MIN_VALUE);
     Assert.assertEquals(state2.getPropAsBoolean("boolean"), false);
+  }
+
+  @Test
+  public void testInterningOfKeyValues() throws Exception {
+
+    // Prove we can identify interned keys
+    String nonInterned = new String("myKey"); // not interned
+    String interned = new String("myInternedKey").intern(); // interned
+
+    Assert.assertFalse(isInterned(nonInterned));
+    Assert.assertTrue(isInterned(interned));
+
+    State state = new State();
+    state.setProp(new String("someKey"), new String("someValue"));
+
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    DataOutput dataOutput = new DataOutputStream(outputStream);
+    state.write(dataOutput);
+    outputStream.flush();
+
+    ByteArrayInputStream inputStream = new ByteArrayInputStream(outputStream.toByteArray());
+    DataInput dataInput = new DataInputStream(inputStream);
+    State readState = new State();
+    readState.readFields(dataInput);
+    inputStream.close();
+
+    Assert.assertEquals(state, readState);
+
+    assertInterned(state.getProperties(), false);
+    assertInterned(readState.getProperties(), true);
+  }
+
+  public static void assertInterned(Map<Object, Object> map, boolean interned) {
+    for (Map.Entry<Object, Object> entry : map.entrySet()) {
+      if (entry.getKey() instanceof String) {
+        Assert.assertEquals(isInterned((String) entry.getKey()), interned);
+      }
+      if (entry.getValue() instanceof String) {
+        Assert.assertEquals(isInterned((String) entry.getValue()), interned);
+      }
+    }
+  }
+
+  public static boolean isInterned(String str) {
+    return str == str.intern();
   }
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyEntity.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyEntity.java
@@ -121,7 +121,7 @@ public class CopyEntity implements HasGuid {
   /**
    * Used for simulate runs. Should explain what this copy entity will do.
    */
-  public String explain() {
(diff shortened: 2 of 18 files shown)


ibuenros:
@chavdar @pcadabam can you review? Thanks!

pcadabam:
Will look at it today

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepDB.java
```diff
@@ -0,0 +1,206 @@
...
+   * Prints details of an SQLException chain to <code>System.err</code>.
+   * Details included are SQL State, Error code, Exception message.
+   *
+   * @param e the SQLException from which to print details.
+   */
+  public static void printSQLException(SQLException e) {
+    // Unwraps the entire exception chain to unveil the real cause of the
+    // Exception.
+    while (e != null) {
+      System.err.println("\n----- SQLException -----");
```
pcadabam: change to logs

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepDB.java
```diff
@@ -0,0 +1,206 @@
...
+      System.err.println("  SQL State:  " + e.getSQLState());
+      System.err.println("  Error Code: " + e.getErrorCode());
+      System.err.println("  Message:    " + e.getMessage());
+      // for stack traces, refer to derby.log or uncomment this:
+      //e.printStackTrace(System.err);
+      e = e.getNextException();
```
pcadabam: Are you trying to print all the "caused by" exceptions?
Did you check if https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/exception/ExceptionUtils.html has what you need?

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepCopyEntity.java
```diff
@@ -27,20 +28,25 @@
...
-  @Getter
-  private final CommitStep step;
   /** A priority value that can be used for sorting {@link CommitStepCopyEntity}s. Lower values are higher priority.*/
   @Getter
   private final int priority;
+  private final String key;
 
-  public CommitStepCopyEntity(String fileSet, Map<String, Object> additionalMetadata, CommitStep step, int priority) {
+  public CommitStepCopyEntity(String fileSet, Map<String, Object> additionalMetadata, CommitStep step, int priority,
+      CommitStepDB commitStepDB, String stepKey) throws IOException {
```
pcadabam: Do you actually need to pass around ```CommitStepDB`` everywhere? Since all of them connect to the same db and use the same table I think having a singleton class is better.

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepDB.java
ibuenros: I actually just copied this from the Derby examples. It sort of understands `SQLExceptions` better.

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepCopyEntity.java
ibuenros: I don't know how expensive it is to instantiate a connection to the db every time. This allows reusability.

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepDB.java
ibuenros: This runs during a shutdown hook, when loggers are not guaranteed to work, which is why it is using `System.err.println`.

ibuenros:
@pcadabam updated.

pcadabam:
LGTM

gobblin-core/src/test/java/gobblin/configuration/StateTest.java
```diff
@@ -120,4 +123,49 @@ public void testState()
...
+        Assert.assertEquals(isInterned((String) entry.getKey()), interned);
+      }
+      if (entry.getValue() instanceof String) {
+        Assert.assertEquals(isInterned((String) entry.getValue()), interned);
+      }
+    }
+  }
+
+  public static boolean isInterned(String str) {
+    return str == str.intern();
```
pcadabam: This is probably the only place I have seen ```==``` vs ```equals()``` :)

chavdar: This might trigger a FindBugs warning.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyEntity.java
```diff
@@ -121,7 +121,7 @@ public DatasetAndPartition getDatasetAndPartition(CopyableDatasetMetadata metada
   /**
    * Used for simulate runs. Should explain what this copy entity will do.
    */
-  public String explain() {
+  public String explain() throws IOException {
```
chavdar: nit: Why do we need this? toString() does not throw IOException.

gobblin-data-management/src/main/java/gobblin/data/management/copy/entities/CommitStepDB.java
```diff
@@ -0,0 +1,198 @@
...
+
+  private static final String INSERT_STEP_STATEMENT = "INSERT INTO steps VALUES (?, ?)";
+  private static final String GET_STEP_STATEMENT = "SELECT step FROM steps WHERE step_key = ?";
+
+  private static final BasicDataSource BDS = new BasicDataSource();
+  static {
+    initialize();
+  }
+
+  private static void initialize() {
```
chavdar: Please avoid having a class with only static methods. How about having HiveCopyEntityHelper instantiate CommitStepDB and pass it to the steps or invoke the serialization itself?

ibuenros: That was the original design, but Pradhan suggested going the static route instead with a global pool of connections.

chavdar: As discussed, shared static state is bad as it may lead to unpredictable results if I decide to launch two gobblin instances.

abti:
@ibuenros where is this used? Is this still relevant?

PR #1073, opened 2016-06-28
extension of loading config file from general file system

autumnust:
- Added new class `AzkabanGobblinDaemon` for deployment on Azkaban .
- Rewrite the original `loadJobConfig` and related method into general form which is compatible for general file system. 

---

- Temporarily remove some parts of SchedulerUtilsTest.java as there's also no counterpart in SchedulerUtils.java(for general file system, most of them are monitor-related), which leads to decrease in the coverage report.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
+++ b/gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
@@ -116,45 +116,57 @@ public class JobScheduler extends AbstractIdleService {
 
   private final boolean waitForJobCompletion;
 
-  public JobScheduler(Properties properties) throws Exception {
+  public JobScheduler(Properties properties)
+      throws Exception {
     this.properties = properties;
     this.scheduler = new StdSchedulerFactory().getScheduler();
 
-    this.jobExecutor = Executors.newFixedThreadPool(
-        Integer.parseInt(properties.getProperty(ConfigurationKeys.JOB_EXECUTOR_THREAD_POOL_SIZE_KEY,
+    this.jobExecutor = Executors.newFixedThreadPool(Integer.parseInt(
+        properties.getProperty(ConfigurationKeys.JOB_EXECUTOR_THREAD_POOL_SIZE_KEY,
             Integer.toString(ConfigurationKeys.DEFAULT_JOB_EXECUTOR_THREAD_POOL_SIZE))),
         ExecutorsUtils.newThreadFactory(Optional.of(LOG), Optional.of("JobScheduler-%d")));
 
-    this.jobConfigFileExtensions =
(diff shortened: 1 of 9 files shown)


gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -116,45 +116,57 @@
...
+  protected void startUp()
+      throws Exception {
     LOG.info("Starting the job scheduler");
-
     this.scheduler.start();
-    if (this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY)) {
-      scheduleLocallyConfiguredJobs();
-      startJobConfigFileMonitor();
+
+    Preconditions.checkArgument(
```
abti: This should not be made mandatory. It is a breaking change for gobblin-cluster, gobblin-yarn and gobblin-aws modules.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -116,45 +116,57 @@
...
+            ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY),
+        "Error in configuration file: Please check your .pull file");
+
+    if (this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY) && !this.properties.containsKey(
+        ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY)) {
+      this.properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY,
+          "file://" + this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY));
     }
+    scheduleGeneralConfiguredJobs();
+    startJobConfigFileMonitor();
```
abti: This should not be made mandatory. It is a breaking change for gobblin-cluster, gobblin-yarn and gobblin-aws modules. 
Each have their own Job config managers, for eg. gobblin-aws has a one that looks at S3 periodically. 

I am going to make these optionals.

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
ibuenros: @abti was this method changed for gobblin-cluster? On the base that @autumnust was using the key `JOB_CONFIG_FILE_DIR_KEY` was mandatory.

autumnust: @abti The monitor part will break the gobblin-cluster for now but that should affect `monitor-dev` branch. Will talk offline.

Issue #1065, opened 2016-06-23
WorkUnit and WorkUnitState


ggthename:
1. According to your Java Doc of WorkUnit class
2. @deprecated Properties in {@link SourceState} should not be added to a {@link WorkUnit}. Having each
   - {@link WorkUnit} contain a copy of {@link SourceState} is a waste of memory. Use {@link #create(Extract, WatermarkInterval)}.
3. So, QueryBasedSource class is creating a WorkUnit this way.
   WorkUnit.create(extract)
   There is no information about SourceState.
4. But, many extractors(including JdbcExtractor, MysqlExtractor and QueryBasedExtractor) are getting properties this way.  
   this.workUnit.getProp("source.conn.driver")
5. It can return null values..
   So, in my opinion We should get properties from WorkUnitState rather than WorkUnit.

sahilTakiar:
Hello @ggthename,

I'm not sure if I understand your question properly. Is this an actual bug you are seeing?

Some background: 
    * A `WorkUnitState` is just a wrapper around a `WorkUnit`, it just contains a few additional runtime properties
    * A `WorkUnit` defines work that needs to be done in an individual Gobblin `Task`, usually a job will consist of many `WorkUnit`s where each `WorkUnit` consists of some division of work that needs to be done
    * `SourceState` is the global configuration for an entire job, so it is basically a set of configuration properties global to all `WorkUnit`s

This wiki has some more documentation on how this all works: http://gobblin.readthedocs.io/en/latest/user-guide/State-Management-and-Watermarks/#gobblin-state-deep-dive

chosh0615:
As @ggthename mentioned, JdbcExtractor or QuerybasedExtractor classes are getting some properties from WorkUnit and get results getting null values.
For instance, properties about jdbc connection information should be read from WorkUnitState.
Since WorkUnitState.getProp function scans all properties in WorkUnitState itself, WorkUnit, and JobState, we can simply use WorkUnitState.getProp from the JdbcExtractor or QuerybasedExtractor.

abti:
@ggthename Yes, this is a known issue that popped up with a few recent optimizations. It is being tracked here: https://github.com/linkedin/gobblin/issues/1022

ggthename:
@abti - I think that it is a little different from #1022
Because, this issue is related to some propreties being read from WorkUnit not available in it anymore,
whereas the issue #1022 was about the properties being read from WorkUnitState not available in it.

[...]

PR #1074, opened 2016-06-29
Oracle Source/Extractor

bjvanov:
Created OracleSource.java and OracleExtractor.java with heavy inspiration from MySqlSource.java and MySqlExtractor.java.

The classes are largely the same as the MySql ones with the exception of getSchemaMetadata() which needed to be modified to account for Oracle reserved words such as format and comment. Also timestamp formats had to be slightly refactored.

Further different strategies for handling primitive values like booleans in Oracle needed to be accounted for. This resulted in overriding getSchema(). The only difference is that Oracle treats booleans as varchar2(1). I created a simple casting utility castToBoolean() which I added to SqlQueryUtils.java.

Since getSchema requires use of the private static final Gson factory from JDBCExtractor. I created a getter for access. It is accessed in OracleExtractor's getSchema() method like so: this.getGson()

I am able to successfully complete table_append and table_full jobs using this new source and extractor. I did a clean assemble, and it did not break the build. I believe I stayed consistent with documentation and coding format.

Please consider my pull request, and let me know if I need to change anything.

Thanks!
Brian Vanover

P.S. One thing is that in order to use the Oracle source and extractor, ojdbc6.jar must be added as a dependency. I simply copied the jar into goblin-dist/lib, but it may be worthwhile to add it as a dependency. I just wasn't sure how to modify the build scripts to include it.


bjvanov:
Found and resolved bug in implementation. Not all common Oracle datatypes were accounted for in getDataTypeMap

bjvanov:
@sahilTakiar  Ok I will get with @jinhyukchang about merging our code. It seems I can pass the ci tests ok, but I can't seem to have much impact on the testing coverage. Can you or @ibuenros take a look?

What I don't understand is how my testing coverage can decrease for an unmodified file e.g. `BaseDataPublisher.java` Nevertheless, I wrote some tests for the `OracleExtractor`, `SqlQueryUtils`, and the `JdbcExtractor`, but it doesn't seem to have much effect. I'm also wondering why the coveralls test does not show the new tests for Oracle.

I've never used CI or testing coverage suites before, so I'm a little unsure how to interpret my results for the 6d09b10 commit.

sahilTakiar:
I'll take a look into what is wrong with Coveralls. In general, code coverage is not a blocker for merging a Pull Request. You have tests written in this PR, so don't worry about the code coverage numbers for now.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,383 @@
...
+  public Iterator<JsonElement> getRecordSetFromSourceApi(String schema, String entity, WorkUnit workUnit,
+      List<Predicate> predicateList) throws IOException {
+    return null;
+  }
+
+  @Override
+  public String getConnectionUrl() {
+    String host = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
+    String port = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT);
+    String url = "jdbc:oracle:thin:@" + host.trim() + ":" + port + ":" + host.trim().split("\\.")[0];
```
jinhyukchang: Seems SID is derived from first part of hostname. Could it take it from property?

jdbc:oracle:thin:@[HOST][:PORT]:SID

jinhyukchang:
@bjvanov Could you take a look at my comment on SID? Once it's addressed I will use your code to test against my use case. Thanks!

jinhyukchang:
One more note, it seems ojdbc driver is not included in runtime dependency. Could you add one?

(Refer to https://github.com/linkedin/gobblin/pull/927/files#diff-4b2ac60989d93a249b26c6a641ff464dR57 )

bjvanov:
@jinhyukchang I incorporated the changes you suggested based on the template from your code. I added extra timestamp values for the data type map, and I refactored the connection string to retrieve a `SID` config parameter from the `ConfigurationKeys` I agree that this is a better way to build the string. I also added `ojdbc6.jar` as a dependency like you suggested. 

@sahilTakiar  if you think there is anything else we should do, let us know. Otherwise, I'd say this PR is ready to be considered for merging if you guys accept it. :)

jinhyukchang:
@bjvanov Thanks for the update. Will try test it out by EOW and get back to you.

shirshanka:
@bjvanov Thanks for your contribution! While we prepare to merge your changes in, could you please fill out the [Gobblin user survey](https://docs.google.com/a/linkedin.com/forms/d/e/1FAIpQLSeH-8so0m68et6kPvxEiCNqezL7k6cyOlz9W-6eXnk7LEkwiA/viewform)?

shirshanka:
@bjvanov Thanks for the info!

jinhyukchang:
@bjvanov I was testing out with your branch and getting an error where it somehow cannot get connection information from workunit and throwing NPE (when it tries to trim SID). When I tried updating the code to use workunitstate, with same job properties, it was able to get it. Have you experienced this problem while testing?

2016-07-15 10:26:50,098 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.OracleExtractor: this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME): null
2016-07-15 10:26:50,098 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.OracleExtractor: this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT): null
2016-07-15 10:26:50,098 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.OracleExtractor: this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_SID): null

java.lang.NullPointerException
	at gobblin.source.extractor.extract.jdbc.OracleExtractor.getConnectionUrl(OracleExtractor.java:234)
	at gobblin.source.extractor.extract.jdbc.JdbcExtractor.createJdbcSource(JdbcExtractor.java:723)
	at gobblin.source.extractor.extract.jdbc.JdbcExtractor.executePreparedSql(JdbcExtractor.java:682)
	at gobblin.source.extractor.extract.jdbc.JdbcExtractor.extractMetadata(JdbcExtractor.java:296)
	at gobblin.source.extractor.extract.QueryBasedExtractor.build(QueryBasedExtractor.java:243)
	at gobblin.source.extractor.extract.jdbc.OracleSource.getExtractor(OracleSource.java:40)
	at gobblin.runtime.TaskContext.getExtractor(TaskContext.java:119)
	at gobblin.runtime.Task.run(Task.java:127)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

bjvanov:
@jinhyukchang Thanks for testing out this distribution. I encountered the same `NPE` that you mentioned. However, it was because I forgot to specify:

`source.conn.sid=<my-sid>`

If you add this to your `.properties` file, it should run. :)

@shirshanka Thanks for the survey, glad I could help! I think I could also help by expanding on some of the documentation. I'll start seeing what I can come up with.

lbendig:
@jinhyukchang It seems to me that the NPE is because of https://github.com/linkedin/gobblin/issues/1065 .

gobblin-core/build.gradle
```diff
@@ -50,6 +50,8 @@ dependencies {
 
   runtime externalDependency.protobuf
 
+  runtime files('./src/main/resources/lib/ojdbc6.jar')
```
lbendig: I'm not quite sure how this will be looked by the installed Gobblin (extracted tarball).
Do you plan to pack this 3rd party dependency to gobblin-core?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+      List<Predicate> predicateList) throws IOException {
+    return null;
+  }
+
+  @Override
+  public String getConnectionUrl() {
+    String host = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
+    String port = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT);
+    String sid = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_SID).trim();
+    String url = "jdbc:oracle:thin:@" + host.trim() + ":" + port + ":" + /*host.trim().split("\\.")[0]*/ sid;
```
lbendig: What if port is not set? Probably a check would be good here, like: 
`(StringUtils.isEmpty(port) ? "" : ":" + port)`

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+
+  //   String outputQuery = query;
+  //   String inputQuery = query.toLowerCase();
+  //   int limitStartIndex = inputQuery.indexOf(" top ");
+  //   int limitEndIndex = getLimitEndIndex(inputQuery, limitStartIndex);
+  //   if (limitStartIndex > 0) {
+  //     outputQuery = query.substring(0, limitStartIndex) + " " + query.substring(limitEndIndex);
+  //   }
+  //   return outputQuery;
+  // }
```
lbendig: If this is a leftover, I think it could be removed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  // @Override
+  // public String constructSampleClause() {
+  //   long sampleRowCount = this.getSampleRecordCount();
+  //   if (sampleRowCount >= 0) {
+  //     return " top " + sampleRowCount;
+  //   }
+  //   return "";
+  // }
```
lbendig: If this is a leftover, I think it could be removed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+    if (limitStartIndex > 0) {
+      String limitValue = query.substring(query.indexOf("<=") + 3, limitEndIndex);
+      try {
+        recordcount = Long.parseLong(limitValue);
+      } catch (Exception e) {
+        log.error("Ignoring incorrct limit value in input query:" + limitValue);
+      }
+    }
+    return recordcount;
+  }
```
lbendig: I know that queries are currently built / modified by string manipulation but I think it would be less error prone to match the input query against a regex that contains the _where rownum_.. clause. Getting the recordCount would be easier as well as removing the sample clause.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+    int limitEndIndex = getLimitEndIndex(inputQuery, limitStartIndex);
+    
+    if ((limitStartIndex > 0 && !multiPredicate) || (limitStartIndex > 0 && multiPredicate && !leadingLimit)) {
+      outputQuery = query.substring(0, limitStartIndex) + " " + query.substring(limitEndIndex);
+    } else if (limitStartIndex > 0 && multiPredicate && leadingLimit) {
+      outputQuery = query.substring(0, limitStartIndex + 6) + " " + query.substring(limitEndIndex + 4);
+    }
+    return outputQuery.trim().replaceAll(" +", " ");
+  }
+
```
lbendig: Please see my comments above.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+    log.debug("Getting hour predicate for Oracle");
+    String formattedvalue = Utils.toDateTimeFormat(Long.toString(value), valueFormat, HOUR_FORMAT);
+    return Utils.getCoalesceColumnNames(column) + " " + operator + " '" + formattedvalue + "'";
+  }
+
+  @Override
+  public String getDatePredicateCondition(String column, long value, String valueFormat, String operator) {
+    log.debug("Getting date predicate for Oracle");
+    String formattedvalue = Utils.toDateTimeFormat(Long.toString(value), valueFormat, DATE_FORMAT);
+    return Utils.getCoalesceColumnNames(column) + " " + operator + " '" + formattedvalue + "'";
```
lbendig: Don't we need to use the **to_date** function here?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
```diff
@@ -0,0 +1,402 @@
...
+  // TODO: How to handle variable precision datatypes e.g. NUMBER(6)
+  @Override
+  public Map<String, String> getDataTypeMap() {
+    Map<String, String> dataTypeMap = ImmutableMap.<String, String> builder().put("char", "string")
+        .put("varchar2", "string").put("varchar", "string").put("nchar", "string").put("nvarchar2", "string")
+        .put("nclob", "string").put("clob", "string").put("long", "string").put("number", "double")
+        .put("binary_float", "float").put("binary_double", "double").put("date", "date").put("datetime", "timestamp")
+        .put("timestamp", "timestamp").put("timestamp with time zone", "timestamp").put("numeric", "double")
+        .put("timezone with local timezone", "timestamp").put("raw", "string").put("long raw", "string")
+        .put("rowid", "string").put("urowid", "string").put("xmltype", "string").put("float", "float")
```
lbendig: I think the float need to be translated to double.

gobblin-core/build.gradle
bjvanov: @lbendig, I removed the jar from `./src/main/resources/lib/ojdbc6.jar`, and I removed this dependency for my upcoming push. You mention that the jars can be passed in `goblin-mapreduce.sh`. Is the same true for `gobblin-standalone.sh`

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: Good catch. Implementing.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: Removed

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: Removed

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: @lbendig, this is a great suggestion. I refactored my `extractSampleRecordCount` and `removeSampleClause` methods to try and match a static, final compiled pattern for the sample clause. If the find a match, they simply substring on the start/end indices of the matching group. I'll push after I pass my local build.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: Please see my response above.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: @lbendig, Can you elaborate on your comment? I borrowed these methods from `MySqlExtractor`. Let me know how you think `to_date` should be used, and I can refactor these methods accordingly.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
bjvanov: Added!

jinhyukchang:
@bjvanov Did you incorporate the fix mentioned in #1065 when you tested it out?

@lbendig Thanks for figuring this out.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
jinhyukchang: @bjvanov Could you remove .put("float", "float") entry? It's currently throwing out IllegalArgumentException due to the conflict with new addition. Also, it would be great if you can add link to the doc mentioning this float --> double mapping.

jinhyukchang:
@bjvanov It seems it's working well except the case when I use query with sample number in it.
For example, if I use below query, it constructs invalid query and throws SQL syntax error exception. Could you take a look?

SELECT ORDER_ID, DATE_PLACED, GG_MODI_TS FROM ORDERS2 WHERE ROWNUM <= 10000

2016-07-18 15:15:51,214 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.JdbcExtractor: Extract query: SELECT ORDER_ID,DATE_PLACED,GG_MODI_TS FROM ORDERS2 where ('$WATERMARK')
2016-07-18 15:15:51,215 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.JdbcExtractor: Get source record count using JDBC
2016-07-18 15:15:51,215 INFO [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.JdbcExtractor: Executing query:SELECT COUNT(1) FROM (SELECT rownum <= 10000 1 as col FROM ORDERS2 where (1=1))temp
2016-07-18 15:15:51,881 ERROR [TaskExecutor-0] gobblin.source.extractor.extract.jdbc.JdbcExtractor: Failed to execute sql:SELECT COUNT(1) FROM (SELECT rownum <= 10000 1 as col FROM ORDERS2 where (1=1))temp ;error-ORA-00923: FROM keyword not found where expected

java.sql.SQLSyntaxErrorException: ORA-00923: FROM keyword not found where expected

bjvanov:
@jinhyukchang Thanks for testing the extractor so thoroughly. Sorry for not responding earlier, but for some reason, I'm having build problems locally when I try to run the tests. I'm ready to take a look at the issue you raised with the sample number query. So it seems you are pulling using `source.querybased.query` is that correct? What is the query you are using?

jinhyukchang:
Hi @bjvanov ,
Yes, I was using source.querybased.query and here's SQL query commented above.

SELECT ORDER_ID, DATE_PLACED, GG_MODI_TS FROM ORDERS2 WHERE ROWNUM <= 10000

If you look at the log in previous comment, it changes it to 
SELECT COUNT(1) FROM (SELECT rownum <= 10000 1 as col FROM ORDERS2 where (1=1))temp
and throws out exception.

I didn't go through the code to pin point the cause but I was using this query on my testing and you can also take a look at my PR if it helps trouble shooting. I tried to add other where clause into query but didn't have any problem so I guess the problem is limited to this case.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/OracleExtractor.java
lbendig: I meant that alternatively the date range queries can be constructed by applying the **to_date** function, like:
```
Utils.getCoalesceColumnNames(column) + " " + operator 
      + " to_date('" + formattedvalue + "', 'YYYY-MM-DD,HH24-MI-SS')";
```

lbendig:
@bjvanov I also tested your code for full range and for an explicitly given watermark range queries and it works fine. However sampling is broken as @jinhyukchang mentioned. We also have our OracleExtractor version, so if you need help I can look up how we construct the sampling queries.

bjvanov:
@lbendig Thanks also for testing the extractor. That's great to hear that you and @jinhyukchang  have working implementations of the sample query methods. Is there a way we can merge the work you guys have already done into this PR?

jinhyukchang:
@bjvanov It's probably taking more work to make my change merge into yours. If you don't have time working on the fix, could you quickly test out my branch? If it works with you, I can push my change instead.
https://github.com/jinhyukchang/gobblin/tree/oracle-reader
https://github.com/linkedin/gobblin/pull/927

jinhyukchang:
@lbendig I believe you mentioned about moving this forward. Are you still planning to do so?

lbendig:
@jinhyukchang Sorry for the delay. Yes, I haven't forgotten this, will push the changes in the coming days.

jinhyukchang:
Hi @lbendig ,
Do you still have plan to push Oracle source? Thanks!

lbendig:
@jinhyukchang @bjvanov Hi, sorry again. I've just rebased this PR which also contains some parts from your implementation (PR#927) and my addons. Tomorrow I push it after performing tests (with/without date ranges, sampling..etc) .

jinhyukchang:
Thanks, @lbendig

lbendig:
Hi @bjvanov and @jinhyukchang,
I'd like to push my changes.
For this I cloned the fork from @bjvanov, rebased it and applied my changes and pushed it to my Gobblin fork: https://github.com/lbendig/gobblin/tree/pr_oracle_extr. If I create a new PR from here to Gobblin master, then the commit history would show @bjvanov's work.
Please let me know if this is fine with you.

bjvanov:
@lbendig This is fine with me. However, will this also reflect @jinhyukchang as an author as well?

jinhyukchang:
@lbendig It's fine with me. Let me test it out with my use case. Thanks!

lbendig:
@bjvanov I added his name as an author in OracleExtractor as well.
@jinhyukchang PR #1065 is not yet merged, so you need to apply first the "workUnit to workUnitState" changes to avoid NPEs.

PR #1076, opened 2016-06-29
Adding a generic kafka writer, integration with linkedin schema registry

shirshanka:
@abti Can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleExtractor.java
@@ -12,22 +12,28 @@
 
 package gobblin.source.extractor.extract.kafka;
 
-import gobblin.configuration.WorkUnitState;
-
 import java.io.IOException;
 
 import kafka.message.MessageAndOffset;
 
+import gobblin.configuration.WorkUnitState;
+import gobblin.metrics.kafka.KafkaSchemaRegistry;
+import gobblin.metrics.kafka.SchemaRegistryException;
 
 /**
  * An implementation of {@link KafkaExtractor} from which reads and returns records as an array of bytes.
  *
  * @author akshay@nerdwallet.com
+ *
+ * @deprecated use {@link KafkaDeserializerExtractor} and {@link KafkaDeserializerExtractor.Deserializers#BYTE_ARRAY} instead
  */
 public class KafkaSimpleExtractor extends KafkaExtractor<String, byte[]> {
 
+  private final KafkaSchemaRegistry<String, String> kafkaSchemaRegistry;
+
   public KafkaSimpleExtractor(WorkUnitState state) {
     super(state);
+    this.kafkaSchemaRegistry = new SimpleKafkaSchemaRegistry(state.getProperties());
   }
 
   @Override
@@ -43,6 +49,10 @@ public class KafkaSimpleExtractor extends KafkaExtractor<String, byte[]> {
    */
   @Override
   public String getSchema() throws IOException {
-    return this.topicName;
+    try {
+      return this.kafkaSchemaRegistry.getLatestSchemaByTopic(this.topicName);
+    } catch (SchemaRegistryException e) {
+      throw new RuntimeException(e);
+    }
   }
 }
--- a/gobblin-kafka/build.gradle
+++ b/gobblin-kafka/build.gradle
@@ -12,6 +12,7 @@ apply plugin: 'java'
 
 dependencies {
   compile project(":gobblin-api")
+  compile project(":gobblin-core")
   compile project(":gobblin-utility")
 
   compile externalDependency.avro
--- a/gobblin-kafka/build.gradle
+++ b/gobblin-kafka/build.gradle
@@ -1,0 +1,63 @@
+// Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at  http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied.
+
+apply plugin: 'java'
+
+dependencies {
+  compile project(":gobblin-api")
+  compile project(":gobblin-utility")
+
+  compile externalDependency.avro
+  compile externalDependency.jacksonCore
+  compile externalDependency.jacksonMapper
+  compile externalDependency.commonsHttpClient
+  compile externalDependency.commonsPool
+  compile externalDependency.commonsLang3
+  compile externalDependency.guava
+  compile externalDependency.slf4j
+  compile externalDependency.httpclient
+  compile externalDependency.httpcore
+  compile(externalDependency.kafka){
+    exclude group: "com.sun.jmx", module: "jmxri"
+    exclude group: "com.sun.jdmk", module: "jmxtools"
+    exclude group: "javax.jms", module: "jms"
+  }  
+  compile externalDependency.kafkaClient
+  compile externalDependency.scala
+  compile externalDependency.mockito
+  compile externalDependency.lombok
+  compile externalDependency.typesafeConfig
+  compile externalDependency.findBugsAnnotations
+
+  runtime externalDependency.protobuf
+
+  testCompile externalDependency.jsonAssert
+  testCompile externalDependency.testng
+  testCompile(externalDependency.kafkaTest){
+    exclude group: "com.sun.jmx", module: "jmxri"
+    exclude group: "com.sun.jdmk", module: "jmxtools"
+    exclude group: "javax.jms", module: "jms"
+  }
+}
+
+configurations {
+  compile { transitive = false }
+  // Remove xerces dependencies because of versioning issues. Standard JRE implementation should
+  // work. See also http://stackoverflow.com/questions/11677572/dealing-with-xerces-hell-in-java-maven
+  // HADOOP-5254 and MAPREDUCE-5664
+  all*.exclude group: 'xml-apis'
+  all*.exclude group: 'xerces'
+}
+
+test {
+  workingDir rootProject.rootDir
+}
+
+ext.classification="library"
--- a/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
+++ b/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.kafka.schemareg;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.Map;
+
+import com.google.common.base.Preconditions;
+
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * An implementation that wraps a passed in schema registry and caches interactions with it
+ * {@inheritDoc}
+ * */
+@Slf4j
+public class CachingKafkaSchemaRegistry<K,S> implements KafkaSchemaRegistry<K,S> {
+
+  private static final int DEFAULT_MAX_SCHEMA_REFERENCES = 10;
+  private final KafkaSchemaRegistry<K,S> _kafkaSchemaRegistry;
+  private final HashMap<String, Map<S, K>> _namedSchemaCache;
+  private final HashMap<K, S> _idBasedCache;
+  private final int _maxSchemaReferences;
+
+
+  public CachingKafkaSchemaRegistry(KafkaSchemaRegistry kafkaSchemaRegistry)
+  {
+    this(kafkaSchemaRegistry, DEFAULT_MAX_SCHEMA_REFERENCES);
+  }
+
+  /**
+   * Create a caching schema registry.
+   * @param kafkaSchemaRegistry: a schema registry that needs caching
+   * @param maxSchemaReferences: the maximum number of unique references that can exist for a given schema.
+   */
+  public CachingKafkaSchemaRegistry(KafkaSchemaRegistry kafkaSchemaRegistry, int maxSchemaReferences)
+  {
+    Preconditions.checkArgument(kafkaSchemaRegistry!=null, "KafkaSchemaRegistry cannot be null");
+    if (kafkaSchemaRegistry.hasInternalCache())
+    {
+      log.warn(new StringBuilder().append("The passed in KafkaSchemaRegistry ")
+          .append(kafkaSchemaRegistry.getClass().getCanonicalName())
+          .append(" already has a cache. Putting another cache on top isn't going to make things faster!").toString());
+    }
+    _kafkaSchemaRegistry = kafkaSchemaRegistry;
+    _namedSchemaCache = new HashMap<>();
+    _idBasedCache = new HashMap<>();
+    _maxSchemaReferences = maxSchemaReferences;
+  }
+
+  @Override
+  public K register(String name, S schema)
+      throws IOException, SchemaRegistryException {
+
+    Map<S, K> schemaIdMap;
+    if (_namedSchemaCache.containsKey(name))
+    {
+      schemaIdMap = _namedSchemaCache.get(name);
+    }
+    else {
+      // we really care about reference equality to de-dup using cache
+      // when it comes to registering schemas, so use an IdentityHashMap here
+      schemaIdMap = new IdentityHashMap<>();
+      _namedSchemaCache.put(name, schemaIdMap);
+    }
+
+    if (schemaIdMap.containsKey(schema))
+    {
+      return schemaIdMap.get(schema);
+    }
+    else
+    {
+      // check if schemaIdMap is getting too full
+      Preconditions.checkState(schemaIdMap.size() < _maxSchemaReferences, "Too many schema objects for " + name +". Cache is overfull.");
+    }
+    K id = _kafkaSchemaRegistry.register(name, schema);
+    schemaIdMap.put(schema, id);
+    return id;
+  }
+
+  @Override
+  public S getById(K id)
+      throws IOException, SchemaRegistryException {
+    if (_idBasedCache.containsKey(id))
+    {
+      return _idBasedCache.get(id);
+    }
+    else
+    {
+      S schema = _kafkaSchemaRegistry.getById(id);
+      _idBasedCache.put(id, schema);
+      return schema;
+    }
+  }
+
+  /**
+   * This call is not cached because we never want to miss out on the latest schema.
+   * {@inheritDoc}
+   */
+  @Override
+  public S getLatestSchema(String name)
+      throws IOException, SchemaRegistryException {
+    return _kafkaSchemaRegistry.getLatestSchema(name);
+  }
+
+  @Override
+  public boolean hasInternalCache() {
+    return true;
+  }
+}
--- a/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
+++ b/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
@@ -62,7 +62,7 @@ public class CachingKafkaSchemaRegistry<K,S> implements KafkaSchemaRegistry<K,S>
   }
 
   @Override
-  public K register(String name, S schema)
+  synchronized public K register(String name, S schema)
       throws IOException, SchemaRegistryException {
 
     Map<S, K> schemaIdMap;
@@ -88,11 +88,12 @@ public class CachingKafkaSchemaRegistry<K,S> implements KafkaSchemaRegistry<K,S>
     }
     K id = _kafkaSchemaRegistry.register(name, schema);
     schemaIdMap.put(schema, id);
+    _idBasedCache.put(id, schema);
     return id;
   }
 
   @Override
-  public S getById(K id)
+  synchronized public S getById(K id)
       throws IOException, SchemaRegistryException {
     if (_idBasedCache.containsKey(id))
     {
--- a/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/ConfigDrivenMd5SchemaRegistry.java
+++ b/gobblin-kafka/src/main/java/gobblin/kafka/schemareg/ConfigDrivenMd5SchemaRegistry.java
@@ -30,9 +30,12 @@ import gobblin.kafka.serialize.MD5Digest;
 
 
 /**
- * A Mock SchemaRegistry that hands out MD5 based ids
+ * A SchemaRegistry that hands out MD5 based ids based on configuration
+ * Can be configured to be initialized with a single schema name, value pair.
+ * {@see ConfigDrivenMd5SchemaRegistry.ConfigurationKeys} for configuration.
+ *
  */
-public class ConfigDrivenMd5SchemaRegistry implements KafkaSchemaRegistry<MD5Digest> {
+public class ConfigDrivenMd5SchemaRegistry implements KafkaSchemaRegistry<MD5Digest, Schema> {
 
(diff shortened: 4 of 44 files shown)


gobblin-kafka/src/main/java/gobblin/kafka/schemareg/KafkaSchemaRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * An interface for a Kafka Schema Registry
+ * Classes implementing this interface will typically be constructed by a {@link KafkaSchemaRegistryFactory}
+ * and should have a constructor that takes a {@link java.util.Properties} object as a parameter.
+ *
+ * @param <K> : the type of the schema identifier (e.g. int, string, md5, ...)
+ *
+ */
+public interface  KafkaSchemaRegistry<K> {
```
sahilTakiar: What's the relation between this class and `gobblin.metrics.kafka.KafkaSchemaRegistry`?

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,270 @@
...
+    try {
+      schema = new Schema.Parser().parse(schemaString);
+    } catch (Throwable t) {
+      throw new SchemaRegistryException(String.format("Schema with ID = %s cannot be parsed", key), t);
+    }
+
+    return schema;
+  }
+
+  public static void main(String[] args)
```
sahilTakiar: Probably want to remove this

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,270 @@
...
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.kafka.serialize.MD5Digest;
+import gobblin.util.AvroUtils;
+
+
+/**
+ * Integration with LinkedIn's implementation of a schema registry that uses md5-hash for schema ids.
+ */
+public class LiKafkaSchemaRegistry implements KafkaSchemaRegistry<MD5Digest> {
```
sahilTakiar: IIRC there was internal LinkedIn code to handle everything related to LinkedIn's schema registry, and the way it serializers / deserializes Avro records? I think it was in the `gobblin-kafka` repo. Does this PR basically move that code to the open source?

gobblin-kafka/src/main/java/gobblin/kafka/serialize/LiAvroDeserializer.java
```diff
@@ -0,0 +1,100 @@
...
+
+import gobblin.kafka.schemareg.KafkaSchemaRegistry;
+import gobblin.kafka.schemareg.SchemaRegistryException;
+
+
+/**
+ * The LinkedIn Avro Deserializer (works with records serialized by the {@link LiAvroSerializer})
+ */
+@Slf4j
+public class LiAvroDeserializer implements Deserializer<GenericRecord> {
```
sahilTakiar: Is there a better name than `LiAvroDeserializer`, like `DefaultAvroDeserializer`? Should users really need to know that this is the way LinkedIn serializes / deserializes records.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -0,0 +1,127 @@
...
+      }
+    }
+    return strippedProps;
+  }
+
+
+  @Override
+  public void write(D record)
+      throws IOException {
+    this.producer.send(new ProducerRecord<String, D>(topic, record), new Callback() {
```
sahilTakiar: Is it necessary to create a new `Callback` object for each record written. Since the class doesn't keep any state, can you just cache the object.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -0,0 +1,127 @@
...
+
+
+  private static final String KEY_SERIALIZER_CONFIG = "key.serializer";
+  private static final String DEFAULT_KEY_SERIALIZER = "org.apache.kafka.common.serialization.StringSerializer";
+  private static final String VALUE_SERIALIZER_CONFIG = "value.serializer";
+  private static final String DEFAULT_VALUE_SERIALIZER = "org.apache.kafka.common.serialization.ByteArraySerializer";
+
+  public KafkaDataWriter(Properties props) {
+    recordsWritten = new AtomicInteger(0);
+    bytesWritten = new AtomicInteger(0);
```
sahilTakiar: Is there a way to accurately track this when writing records? Right now it doesn't seem to be updated.

sahilTakiar:
You may also be interested in https://github.com/linkedin/gobblin/pull/1016

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
shirshanka: This class is (mostly) a copy of the existing KafkaSchemaRegistry and KafkaAvroSchemaRegistry. 
I'm not sure why this code was there in the first place.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
shirshanka: Essentially.. there will be a few more PR-s later to clean up the other kafka-related code, including KafkaAvroExtractor, I wanted to keep this PR relatively clean (without refactoring existing code) so that the bar for merging was low :)

gobblin-kafka/src/main/java/gobblin/kafka/serialize/LiAvroDeserializer.java
shirshanka: I'm anticipating that the Confluent ser-de integration will add ConfluentAvroDeserializer and Serializer... so don't know if company based naming is really that bad.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
shirshanka: Good catch.. will change.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
shirshanka: What is our current policy regarding these counters if a particular writer cannot keep them accurately maintained?

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/KafkaSchemaRegistry.java
shirshanka: I plan to have this super simple interface replace gobblin.metrics.kafka.KafkaSchemaRegistry (which is a base class with a ton of behavior). 

LiKafkaSchemaRegistry will replace gobblin.metrics.kafka.KafkaSchemaRegistry and gobblin.metrics.kafka.KafkaAvroSchemaRegistry.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
shirshanka: Half of the code (avro serialization) already exists in gobblin-metrics, the other half (the extractor) is in the linkedin-internal repo. 

I plan to change the kafka reporter in gobblin-metrics to be a thin wrapper on the kafka writer.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/ConfigDrivenMd5SchemaRegistry.java
```diff
@@ -0,0 +1,137 @@
...
+import gobblin.kafka.serialize.MD5Digest;
+
+
+/**
+ * A SchemaRegistry that hands out MD5 based ids based on configuration
+ * Can be configured to be initialized with a single schema name, value pair.
+ * {@see ConfigDrivenMd5SchemaRegistry.ConfigurationKeys} for configuration.
+ *
+ */
+public class ConfigDrivenMd5SchemaRegistry implements KafkaSchemaRegistry<MD5Digest, Schema> {
```
shirshanka: need to fix javadocs on this.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/HttpClientFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.commons.pool2.PooledObject;
+import org.apache.commons.pool2.impl.DefaultPooledObject;
+
+
+/**
+ * An implementation of {@link BasePooledObjectFactory} for {@link HttpClient}.
+ *
+ * @author mitu
+ */
+public class HttpClientFactory extends BasePooledObjectFactory<HttpClient>{
```
shirshanka: Uninteresting, copied from gobblin-metrics.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/KafkaSchemaRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.kafka.schemareg;
+
+import java.io.IOException;
+
+import org.apache.avro.Schema;
```
shirshanka: Need to cleanup this unused import.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,247 @@
...
+  private final GenericObjectPool<HttpClient> httpClientPool;
+  private final String url;
+
+  /**
+   * @param props properties should contain property "kafka.schema.registry.url", and optionally
+   * "kafka.schema.registry.max.cache.size" (default = 1000) and
+   * "kafka.schema.registry.cache.expire.after.write.min" (default = 10).
+   */
+  public LiKafkaSchemaRegistry(Properties props) {
+    //super(props);
```
shirshanka: need to delete this commented out code.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,247 @@
...
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.kafka.serialize.MD5Digest;
+import gobblin.util.AvroUtils;
+
+
+/**
+ * Integration with LinkedIn's implementation of a schema registry that uses md5-hash for schema ids.
+ */
+public class LiKafkaSchemaRegistry implements KafkaSchemaRegistry<MD5Digest, Schema> {
```
shirshanka: This class is a merge of KafkaSchemaRegistry and KafkaAvroSchemaRegistry from gobblin-metrics. 
I've removed the internal cache, so we can layer caching on top as needed for any schema registry.

gobblin-kafka/src/main/java/gobblin/kafka/tool/SimpleKafkaConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+import gobblin.kafka.serialize.LiAvroDeserializer;
+import gobblin.kafka.serialize.MD5Digest;
+import gobblin.kafka.writer.KafkaWriterConfigurationKeys;
+
+
+/**
+ * Created by sdas on 6/30/16.
+ */
+@Slf4j
+public class SimpleKafkaConsumer {
```
shirshanka: This is still a draft class.

gobblin-kafka/src/main/java/gobblin/kafka/serialize/LiAvroDeserializer.java
abti: I don't have a strong preference, but a few alternatives can be `ClassicAvroDeserializer` or `GobblinDefaultAvroDeserializer`

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleExtractor.java
```diff
@@ -43,6 +49,10 @@ public KafkaSimpleExtractor(WorkUnitState state) {
    */
   @Override
   public String getSchema() throws IOException {
-    return this.topicName;
```
abti: Why was it returning topicName in getSchema()? Was it even working?

gobblin-kafka/build.gradle
```diff
@@ -0,0 +1,64 @@
...
+  compile externalDependency.httpclient
+  compile externalDependency.httpcore
+  compile(externalDependency.kafka){
+    exclude group: "com.sun.jmx", module: "jmxri"
+    exclude group: "com.sun.jdmk", module: "jmxtools"
+    exclude group: "javax.jms", module: "jms"
+  }  
+  compile externalDependency.kafkaClient
+  compile externalDependency.scala
+  compile externalDependency.mockito
```
abti: Should be testCompile

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/ConfigDrivenMd5SchemaRegistry.java
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Schema getLatestSchema(String name)
+      throws IOException, SchemaRegistryException {
+    return null;
```
abti: throw UnsupportedException?

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    if (schemaIdMap.containsKey(schema))
+    {
+      return schemaIdMap.get(schema);
+    }
+    else
+    {
+      // check if schemaIdMap is getting too full
+      Preconditions.checkState(schemaIdMap.size() < _maxSchemaReferences, "Too many schema objects for " + name +". Cache is overfull.");
```
abti: This will throw IllegalStateException, whereas we should evict older entries or least frequently used. May be something on the lines of sized based cache here: https://github.com/google/guava/wiki/CachesExplained

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,123 @@
...
+   * Create a caching schema registry.
+   * @param kafkaSchemaRegistry: a schema registry that needs caching
+   * @param maxSchemaReferences: the maximum number of unique references that can exist for a given schema.
+   */
+  public CachingKafkaSchemaRegistry(KafkaSchemaRegistry kafkaSchemaRegistry, int maxSchemaReferences)
+  {
+    Preconditions.checkArgument(kafkaSchemaRegistry!=null, "KafkaSchemaRegistry cannot be null");
+    if (kafkaSchemaRegistry.hasInternalCache())
+    {
+      log.warn(new StringBuilder().append("The passed in KafkaSchemaRegistry ")
```
abti: I am of the opinion that we shouldn't allow caching over a registry that has internal caching. It might be a reciepe for disaster.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/CachingKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,123 @@
...
+          .append(" already has a cache. Putting another cache on top isn't going to make things faster!").toString());
+    }
+    _kafkaSchemaRegistry = kafkaSchemaRegistry;
+    _namedSchemaCache = new HashMap<>();
+    _idBasedCache = new HashMap<>();
+    _maxSchemaReferences = maxSchemaReferences;
+  }
+
+  @Override
+  public K register(String name, S schema)
```
abti: We should make register() op thread safe.

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/KafkaSchemaRegistryFactory.java
```diff
@@ -0,0 +1,59 @@
...
+@Slf4j
+public class KafkaSchemaRegistryFactory {
+
+  @SuppressWarnings("unchecked")
+  public static KafkaSchemaRegistry getSchemaRegistry(Properties props) {
+    Preconditions.checkArgument(props.containsKey(KafkaSchemaRegistryConfigurationKeys.KAFKA_SCHEMA_REGISTRY_CLASS),
+        "Missing required property " + KafkaSchemaRegistryConfigurationKeys.KAFKA_SCHEMA_REGISTRY_CLASS);
+
+    boolean tryCache = true;
+    if (props.containsKey(KafkaSchemaRegistryConfigurationKeys.KAFKA_SCHEMA_REGISTRY_CACHE))
```
abti: This is not required if you do properties.getPropertie(key, defaultValue) as done in line 40

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/KafkaSchemaRegistryFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    {
+      tryCache = Boolean.parseBoolean(props.getProperty(KafkaSchemaRegistryConfigurationKeys.KAFKA_SCHEMA_REGISTRY_CACHE, "false"));
```
abti: By this it seems you want to keep tryCache's default value as false, where as it will be left as true (because of line 37 and line 38) if the property is not present in properties object

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,247 @@
...
+  /**
+   * Get the latest schema of a topic.
+   *
+   * @param topic topic name
+   * @return the latest schema
+   * @throws SchemaRegistryException if failed to retrieve schema.
+   */
+  @Override
+  public Schema getLatestSchema(String topic) throws SchemaRegistryException {
+    String schemaUrl = this.url + GET_RESOURCE_BY_TYPE + topic;
```
abti: Urlencode topic name?

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,247 @@
...
+    LOG.debug("Fetching from URL : " + schemaUrl);
+
+    GetMethod get = new GetMethod(schemaUrl);
+
+    int statusCode;
+    String schemaString;
+    HttpClient httpClient = this.borrowClient();
+    try {
+      statusCode = httpClient.executeMethod(get);
+      schemaString = get.getResponseBodyAsString();
```
abti: If statusCode is not SC_OK, we don;t need to fetch response body. So we can move this to line 128

gobblin-kafka/src/main/java/gobblin/kafka/schemareg/LiKafkaSchemaRegistry.java
```diff
@@ -0,0 +1,247 @@
...
+    if (statusCode != HttpStatus.SC_OK) {
+      throw new SchemaRegistryException(
+          String.format("Latest schema for topic %s cannot be retrieved. Status code = %d", topic, statusCode));
+    }
+
+    Schema schema;
+    try {
+      schema = new Schema.Parser().parse(schemaString);
+    } catch (Throwable t) {
+      throw new SchemaRegistryException(String.format("Latest schema for topic %s cannot be retrieved", topic), t);
```
abti: For better logging: String.format("Latest schema for topic %s cannot be parsed: %s", topic, schemaString)

gobblin-kafka/src/main/java/gobblin/kafka/serialize/LiAvroDeserializer.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   *
+   * @param topic topic associated with the data
+   * @param data serialized bytes
+   * @return deserialized object
+   */
+  @Override
+  public GenericRecord deserialize(String topic, byte[] data) {
+    Preconditions.checkState(_schemaRegistry!=null, "Schema Registry is not initialized");
```
abti: _schemaRegistry and _datumReader are private members initialized / assigned in constructor. So, we can just check for Preconditions.notNull() once in constructor. 
At the moment, Preconditions.checkState() would be invoked for all the deserialize() calls rather than once

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -116,45 +116,57 @@
...
+        this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY) || this.properties.containsKey(
+            ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY),
+        "Error in configuration file: Please check your .pull file");
+
+    if (this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY) && !this.properties.containsKey(
+        ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY)) {
+      this.properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY,
+          "file://" + this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY));
     }
+    scheduleGeneralConfiguredJobs();
```
abti: The methods: `scheduleGeneralConfiguredJobs()` and `startJobConfigFileMonitor()` should still be invoked as before within if construct: 
`if(this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY)) { .. }` 
.. because JobScheduler is extended by `GobblinHelixJobScheduler` in `gobblin-cluster` and hence in `gobblin-yarn` and `gobblin-aws` 
These modules schedule jobs using their variants of `JobConfigurationManager`, so making the `JOB_CONFIG_FILE_DIR` property existence and above method calls mandatory will break those modules

gobblin-utility/src/main/java/gobblin/util/SchedulerUtils.java
```diff
@@ -68,13 +75,22 @@ public boolean accept(File dir, String name) {
...
-   * Load job configurations from job configuration files stored under the
-   * root job configuration file directory.
-   *
-   * @param properties Gobblin framework configuration properties
-   * @return a list of job configurations in the form of {@link java.util.Properties}
-   */
+  private static final PathFilter PROPERTIES_PATH_FILTER = new PathFilter() {
+    @Override
+    public boolean accept(Path path) {
+      String fileExtension = path.getName().substring(path.getName().lastIndexOf('.') + 1);
```
abti: Need not change but suggestion: 
`StringUtils.substringAfterLast(str, separator)` .. helps in such cases.

So this line would change to: 
`return JOB_PROPS_FILE_EXTENSION.equalsIgnoreCase(StringUtils.substringAfterLast(path.getName(), "."))` 

Can also use: `return path.getName().toLowerCase().endsWith("." + JOB_PROPS_FILE_EXTENSION)`

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -57,8 +57,11 @@
   // Job configuration file monitor polling interval in milliseconds
   public static final String JOB_CONFIG_FILE_MONITOR_POLLING_INTERVAL_KEY = "jobconf.monitor.interval";
   public static final long DEFAULT_JOB_CONFIG_FILE_MONITOR_POLLING_INTERVAL = 300000;
-  // Directory where all job configuration files are stored
+  // Directory where all job configuration files are stored WHEN ALL confs reside in local FS.
   public static final String JOB_CONFIG_FILE_DIR_KEY = "jobconf.dir";
+
+  // Path where all job configuration files stored
+  public static final String JOB_CONFIG_FILE_GENERAL_PATH_KEY = "jobconf.fullyQualifiedPath" ;
```
abti: How about let user specify fully qualified path as value for existing `JOB_CONFIG_FILE_DIR_KEY` 
.. and in places where you mandatorily need fully qualified path (resolving to local FS) call:  
`jobConfDir.toURI().toString()` or `jobConfDir.toURI().toURL().toString()`
(I believe this returns a string like: `file:/user/foo/bar` when `jobConfDir` is `bar` with parent directory `/user/foo` )

shirshanka:
Discarding this PR since #1089 contains this and more.

PR #1077, opened 2016-06-30
Fixing a findbugs warning in SalesforceExtractor

chavdar:
Fixing a Findbugs warning


chavdar:
Self-merging due to simplicity.

PR #1082, opened 2016-07-04
Updating the README and other outdated docs to encourage use of Gobblin Releases

sahilTakiar:
* Many users try to get started with Gobblin by building the code (e.g. `./gradlew build`), then un-tarring the resulting `gobblin-distribution.tar.gz` file, and then running the Wikipedia example
  * Many of them fail to get the build working, or have problems running the tests
* Most of our documentation suggests getting started with Gobblin by first building it (e.g. the `README.md` and Getting Started Guide both suggest this)
* However, users should just be using the releases here: https://github.com/linkedin/gobblin/releases rather than building the latest code themselves
* This PR updates the `README.md` file and a few other docs to encourage first time users to use the releases rather than building the code themselves
* I also uploaded the `gobblin-distribution.tar.gz` file for the latest release (0.7.0)
  * Going forward I think each release should also have the `gobblin-distribution.tar.gz` file attached, so that users can download it directly rather than building it themselves


sahilTakiar:
@chavdar any thoughts on this? The main thing is that I think each release should include the `gobblin-distribution.tar.gz` file, I have already done this for the 0.7.0 release

sahilTakiar:
Just a few examples of where this happens:

https://github.com/linkedin/gobblin/issues/1084
https://github.com/linkedin/gobblin/issues/1061
https://github.com/linkedin/gobblin/issues/1018
https://github.com/linkedin/gobblin/issues/786
https://github.com/linkedin/gobblin/issues/732

shirshanka:
LGTM!

PR #1088, opened 2016-07-06
Gobblin on AWS 

abti:
## Introduction 

This PR enables Gobblin to run in cluster mode on AWS via a new sub-module called `gobblin-aws` 

* `gobblin-aws` provides support to launch and run Gobblin cluster on AWS 
* `gobblin-aws` inherits from `gobblin-cluster` to make use of its abstraction wrapping Helix clustering capabilities 
* `gobblin-aws` performs all AWS resource management activities and provides for other critical cluster management capabilities not already provided by `gobblin-cluster` such as: High Availability, Elasticity and Failover

## Architecture

gobblin-on-aws-architecture

### AWS Cluster Launcher

`gobblin-aws` is launched using `GobblinAWSClusterLauncher` which upon bootup launches a few services: 
- `AWSClusterSecurityManager` to authenticate with AWS and refresh credentials at configured intervals
- `AWSSdkClient` to coordinate with AWS and launch various resources 

After this, the `GobblinAWSClusterLauncher` computes the cloud-init scripts as well as AWS LaunchConfigurations and requests AWS to launch two `AutoScalingGroups*`:
- Master `AutoScalingGroup` to launch and manage EC2 instances for `GobblinAWSClusterMaster` 
- Workers `AutoScalingGroup` to launch and manage EC2 instances for `GobblinAWSTaskRunner` 

### Master AutoScalingGroup

When an EC2 instance under Master `AutoScalingGroup` boots up, it executes the master cloud-init script that:
- First, creates a `NFS` server and setups work environment on the node (eg. creating all relevant directories) 
- Second, launches a java application to run `GobblinAWSClusterMaster` with relevant classpath and JVM arguments 

The `GobblinAWSClusterMaster` upon bootup: 
- Registers itself as HelixController for the cluster
- Launches a few services:
- 1. `AWSJobConfigurationManager` that runs on configured frequency and polls for new or updated job configurations from S3
- 2. `GobblinHelixJobScheduler` that schedules discovered new or updated jobs for execution

### Workers AutoScalingGroup 

When n-number of EC2 instances under Workers `AutoScalingGroup` boot up, they execute worker's cloud-init script that: 
- First, mounts the `NFS` shared directory by master and setups work environment on the node
- Second, launches a java application to run `GobblinAWSTaskRunner` with relevant classpath and JVM arguments 

The `GobblinAWSTaskRunner` upon bootup: 
- Registers itself as a HelixParticipant in the cluster 
- Registers `GobblinHelixTaskFactory` to listen to Helix and process any tasks 

**AutoScalingGroup*** `AutoScalingGroup` is a service provided by AWS in which:
- Users can specify minimum, maximum and desired number of instances
- Scaling criteria based on which `AutoScalingGroup` automatically launches new instances or shuts them down to scale for load OR launch new instance if any instance dies. 
 
Note: 
1. When new instance launch, they execute the configured cloud-init script and hence register themselves as master or workers
2. `AutoScalingGroup`s help us achieve HA and Elasticity for load 

### Execution Model  
Gobblin cluster on AWS makes use of Helix [Distributed Task Execution Framework](http://helix.apache.org/0.7.1-docs/recipes/task_dag_execution.html) to schedule and run jobs. 
For each Gobblin job run: 
- The `GobblinHelixJobScheduler` on Master node starts a `GobblinHelixJobLauncher` that wraps the Gobblin job into a `GobblinHelixJob` and each Gobblin Task into a `GobblinHelixTask`, which implements the Helix's Task interface so Helix knows how to execute it 
- The `GobblinHelixJobLauncher` on Master node then submits the job to a Helix job queue named after the Gobblin job name, from which the Helix Distributed Task Execution Framework picks up the job and runs its tasks through the live participant worker nodes 

## Upcoming / Pending  
- Replace `NFS` with `EFS` which recently went into GA 
- Multi master support for master HA 
- Support for spot instances 
- More hooks for auto scaling rules / parameters 
- Task execution JVM isolation (though this needs to be baked into `gobblin-cluster` )
- `AMI` for AWS to bootup from a pre-baked Gobblin instance 
- Process watcher daemon running on all nodes (master and workers) to restart `GobblinAWSClusterMaster` and `GobblinAWSTaskRunner` process JVMs if they die 

## Running Gobblin-on-AWS 

Follow these steps to run Gobblin-on-AWS 
- Build Gobblin as you build today, and extract the distribution tar.gz 
- Modify `conf/aws/application.conf` as described later in `Configuration` section 
- Replace secret and access keys with dummy values in `application.conf` and upload it on S3 
- Launch Gobblin cluster on AWS by running: `bin/gobblin-aws.sh start` 
 
A private key file with pem extension should be saved on the disk, you can use this to ssh into Master and Worker nodes as you would in any regular instance on AWS 

### Configuration 

Getting started with Gobblin on AWS is easy. Only the following configurations need to be changed in `conf/aws/application.conf` :
```
# Cluster name
gobblin.aws.cluster.name=GobblinAWSPullRequest

# Zookeeper connection string
# Note: You can use the example below, but the example Zookeeper may or may not be available
# Example: gobblin.cluster.zk.connection.string="ec2-54-149-223-115.us-west-2.compute.amazonaws.com:2181"
gobblin.cluster.zk.connection.string="ec2-54-149-223-115.us-west-2.compute.amazonaws.com:2181"

# AWS Credentials
gobblin.aws.service.access="<AWS Access Key>"
gobblin.aws.service.secret="<AWS Secret Key>"

# Leave as-is for sample wikipedia pull job
# -- OR --
# Specify S3 URI for job configuration:
# Steps to obtain this URI:
# 1. Create a S3 folder in a bucket, say the bucket name is "s3bucket" and folder is "/gobblin-confs/job-conf/"
# 2. Zip your pull files in a zip and upload to this folder. Say your zip file name is myPullFiles.zip
# 3. Specify S3 URI for this zip file in the config below
#    For the description above: "https://s3-region.amazonaws.com/s3bucket/gobblin-confs/job-conf/myPullFiles.zip"
gobblin.aws.job.conf.s3.uri="https://s3-us-west-2.amazonaws.com/gobblin-libs/job-config/wikipedia.zip"

# Specify S3 URI for cluster configuration
# Steps to obtain this URI:
# 1. Create a S3 folder in a bucket, say the bucket name is "s3bucket" and folder is "/gobblin-confs/cluster-conf/"
# 2. After editing this application.conf: upload this application.conf, log4j-aws.properties and quartz.properties to
#    /s3bucket/gobblin-confs/cluster-conf/
#    Note: Make sure to replace AWS credential values with dummy values.
#          We don't read AWS credentials from application.conf in S3
# 3. Specify S3 URI for this folder in the config below
#    For the description above: "https://s3-region.amazonaws.com/s3bucket/gobblin-confs/cluster-conf/"
# Example: gobblin.aws.master.s3.conf.uri="https://s3-us-west-2.amazonaws.com/gobblin-libs/cluster-config/"
gobblin.aws.master.s3.conf.uri="https://s3-region.amazonaws.com/s3bucket/gobblin-confs/cluster-conf/" 
```

## Notes
- **Logs** for master and workers are saved in a common directory in `NFS` and are available in real time
- All classes in this PR have been marked with `Alpha` tag, although they have been tested thoroughly several times 
- No change is required in the way `Metrics` or any jobs are configured 
- More documentation will follow in Wiki PR


abti:
@shirshanka @sahilTakiar Can you please review?

gobblin-aws/build.gradle
```diff
@@ -0,0 +1,105 @@
...
+
+configurations {
+  compile {
+    transitive = false
+  }
+
+  testRuntime {
+    resolutionStrategy {
+      force 'com.google.inject:guice:3.0'
+//      force 'org.apache.hadoop:hadoop-yarn-server-resourcemanager:'+hadoopVersion
```
shirshanka: remove commented line.

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
```diff
@@ -0,0 +1,230 @@
...
+  private Map<String, Properties> jobConfFiles;
+
+  private final long refreshIntervalInSeconds;
+
+  private final ScheduledExecutorService fetchJobConfExecutor;
+
+  public AWSJobConfigurationManager(EventBus eventBus, Config config) {
+    super(eventBus, config);
+    this.jobConfFiles = Maps.newHashMap();
+    if (config.hasPath(GobblinAWSConfigurationKeys.JOB_CONF_REFRESH_INTERVAL_IN_SECONDS)) {
```
shirshanka: Suggestion: support refresh interval as a time value config and support notation like: 30s, 20m .

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
```diff
@@ -0,0 +1,230 @@
...
+
+    // Schedule the job config fetch task
+    this.fetchJobConfExecutor.scheduleAtFixedRate(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          fetchJobConf();
+        } catch (IOException | ConfigurationException e) {
+          LOGGER.error("Failed to fetch job configurations", e);
+          throw Throwables.propagate(e);
```
shirshanka: Any specific reason for preferring this over: 

throw new RuntimeException("Failed to fetch job configurations", e);

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
```diff
@@ -0,0 +1,230 @@
...
+              IOUtils.closeQuietly(out);
+          }
+        }
+      }
+    }
+  }
+
+  @Override
+  protected void shutDown() throws Exception {
+    // Nothing to do
```
shirshanka: Don't you need to shutdown the threadpool / executor?

gobblin-aws/src/main/java/gobblin/aws/AWSShutdownHandler.java
```diff
@@ -0,0 +1,95 @@
...
+      Optional<List<String>> optionalAutoScalingGroupNames) {
+    this.awsSdkClient = awsSdkClient;
+    this.optionalLaunchConfigurationNames = optionalLaunchConfigurationNames;
+    this.optionalAutoScalingGroupNames = optionalAutoScalingGroupNames;
+  }
+
+  @Override
+  public void onTimeOut() {
+    LOGGER.warn("Timeout while waiting for Helix controller and participants shutdown. "
+        + "Moving ahead with ungraceful shutdown of Amazon AutoScaling group");
```
shirshanka: s/ungraceful/forced

gobblin-aws/src/main/java/gobblin/aws/AWSShutdownHandler.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private void shutdownASG() {
+    if (optionalLaunchConfigurationNames.isPresent()) {
+      for (String launchConfigurationName : optionalLaunchConfigurationNames.get()) {
+        try {
+          this.awsSdkClient.deleteLaunchConfiguration(launchConfigurationName);
+        } catch (Exception e) {
+          // Ignore and continue, so that we clean up as many resources as possible
+          LOGGER.error("Issue in deleting launch configuration: " + launchConfigurationName, e);
```
shirshanka: should it be warn or error. 
Also log message should say that it is continuing on.

gobblin-aws/src/main/java/gobblin/aws/CloudInitScriptBuilder.java
```diff
@@ -0,0 +1,254 @@
...
+    // Download jars from S3
+    // TODO: Limit only custom user jars to pulled from S3, load rest from AMI
+    final List<String> awsJars = SPLITTER.splitToList(workerS3JarsFiles);
+    for (String awsJar : awsJars) {
+      cloudInitCmds.append(String.format("wget -P \"${jr}\" \"${jr0}\"%s", awsJar)).append("\n");
+    }
+    classpath.append(":").append(workerJarsDir).append("*");
+
+    // Get a random Helix instance name
+    cloudInitCmds.append("pi=`curl http://169.254.169.254/latest/meta-data/local-ipv4`").append("\n");
```
shirshanka: What does this url: http://169.254.169.254/latest/meta-data/local-ipv4 do?

gobblin-aws/src/main/java/gobblin/aws/CloudInitScriptBuilder.java
```diff
@@ -0,0 +1,254 @@
...
+
+
+/**
+ * Class to generate script for launching Gobblin cluster master and workers via cloud-init
+ * on EC2 instance boot up.
+ *
+ * @author Abhishek Tiwari
+ */
+@Alpha
+public class CloudInitScriptBuilder {
```
shirshanka: Did you consider straight out writing a bash script with variables and then using this class to just set those variables?

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
```diff
@@ -0,0 +1,666 @@
...
+  private String masterLaunchConfigName;
+  private String masterAutoScalingGroupName;
+
+  private String workerLaunchConfigName;
+  private String workerAutoScalingGroupName;
+
+  private final Optional<String> gobblinVersion;
+
+
+  public GobblinAWSClusterLauncher(Config config) throws IOException {
```
shirshanka: The code below seems to assume that all of the configuration is mandatory. 
Should we add defaults to most of the config values?

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
```diff
@@ -0,0 +1,666 @@
...
+    this.config = config;
```
shirshanka: Don't need to hang on to the config instance since you are getting all values out in the constructor.

gobblin-aws/build.gradle
abti: Removed

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
abti: Changed refresh interval to accept values such as 30s, 20m

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
abti: It was remanent of pre-refactor code. However, it has some advantages in some situations: https://github.com/google/guava/wiki/ThrowablesExplained 

However, this doesn't needs it. Changed

gobblin-aws/src/main/java/gobblin/aws/AWSJobConfigurationManager.java
abti: Added shutdown for executor here and in AWSSecurityManager

gobblin-aws/src/main/java/gobblin/aws/AWSShutdownHandler.java
abti: Changed

gobblin-aws/src/main/java/gobblin/aws/AWSShutdownHandler.java
abti: Changed to warn and that its continuing. 

The impact of not being able to delete a launch config is not much. AWS doesn't charges for it (its a simple config object) but has limit of 100 launch configurations unless user specifically requests for more quota. 

However, the cost of not being able to delete an AutoScalingGroup is more, that would mean new instances may continue to get launched and user might still incur charges. So, added log line to indicate to user to manually delete them by logging into AWS console. 
The limit for number of AutoScalingGroups is 20

gobblin-aws/src/main/java/gobblin/aws/CloudInitScriptBuilder.java
abti: The ip 169.254.169.254 is private and available only one the EC2 instance. EC2 provides it as a utility to expose instance metadata to processes that are running locally and are interested to fetch any metadata about the instance. More about it: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

http://169.254.169.254/latest/meta-data/local-ipv4 has the private IP address of the instance. In cases where multiple network interfaces are present, this refers to the eth0 device (the device for which the device number is 0)

gobblin-aws/src/main/java/gobblin/aws/CloudInitScriptBuilder.java
abti: Yes, but cloudinit / userdata has strict limits on number of characters allowed for the script - which gets really bloated after base64 conversion (the only format AWS api accepts it in) 
So, I first had the same model but was forced to move embedded mode to save char size space. The options were to do it like it is done right now or do string substitutions on a template. Since in gobblin-yarn we follow the first approach, I just stuck to it. 

However, this will eventually be changed into a bash script with variable substitution like you propose after we have pre-baked AMI with most of the required jars already baked into the instance (coming soon :) )

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
abti: Changed to add defaults for most of the config values

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
abti: Unfortunately, the email utility we are using needs the config object to construct the state. I plan to change the email utility and reporting going forward in general outside of gobblin-aws

gobblin-aws/src/main/java/gobblin/aws/AWSShutdownHandler.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Class that handles Helix shutdown response and consequently shutdowns Amazon AutoScaling group.
+ *
+ * @author Abhishek Tiwari
+ */
+@Alpha
+public class AWSShutdownHandler extends AsyncCallback {
+  private static final Logger LOGGER = LoggerFactory.getLogger(AWSShutdownHandler.class);
+
+  private static final boolean SHOULD_FORCE_DELETE_ASG_DEFAULT = false;
```
shirshanka: seems a bit forced to declare a constant and use it just once below. 
Do you expect to change this behavior in the future?

abti: No, jt won't change. Removed it

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
```diff
@@ -0,0 +1,669 @@
...
+   *
+   * @throws IOException If there's anything wrong setting up the AWS cluster
+   */
+  @VisibleForTesting
+  String setupGobblinCluster() throws IOException {
+
+    final String uuid = UUID.randomUUID().toString();
+
+    // Create security group
+    // TODO: Make security group restrictive
```
shirshanka: What stops working if the sec group is made restrictive?

abti: Nothing. Today we open more than required ports, however, we need to restrict it to only that are needed by Zookeeper / Helix, UI, and Kibana and other modules going forward

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterMaster.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * <p>
+ *   Note: Shutdown initiated by {@link GobblinAWSClusterLauncher} via a Helix message of subtype
+ *   {@link HelixMessageSubTypes#APPLICATION_MASTER_SHUTDOWN} is handled by super class {@link GobblinClusterManager}
+ * </p>
+ *
+ * @author Abhishek Tiwari
+ */
+@Alpha
+public class GobblinAWSClusterMaster extends GobblinClusterManager {
```
abti: Mostly because it is naturally a form of GobblinClusterManager, wherein it inherits most of the functionalities inbuilt in it and overrides a few methods. Using it as a member variable would need invoking methods on the member variable .. which otherwise are inherited

shirshanka: Does it make sense to call it GobblinAWSClusterManager or just AWSClusterManager in that case?

abti: I don't have a strong preference for either names.

abti: As discussed offline, renamed GobblinAWSClusterMaster to GobblinAWSClusterManager

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -140,6 +144,24 @@ public WikipediaExtractor(WorkUnitState workUnitState) throws IOException {
     this.reader = new WikiResponseReader();
   }
 
+  private String readProp(String key, WorkUnitState workUnitState) {
+    String value = workUnitState.getWorkunit().getProp(key);
+    if (StringUtils.isBlank(value)) {
+      value = workUnitState.getProp(key);
+    }
+    if (StringUtils.isBlank(value)) {
+      value = workUnitState.getJobState().getProp(key);
```
shirshanka: Could there be NPE here?

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -140,6 +144,24 @@ public WikipediaExtractor(WorkUnitState workUnitState) throws IOException {
     this.reader = new WikiResponseReader();
   }
 
+  private String readProp(String key, WorkUnitState workUnitState) {
```
shirshanka: I'm fine with leaving this in, but concerned that this is a hack that hides a more serious problem with workunit state and job state ambiguity.

gobblin-aws/src/main/java/gobblin/aws/CloudInitScriptBuilder.java
shirshanka: Thx!

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
shirshanka: Thx!

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
shirshanka: Thx!

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
shirshanka: Thx!

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
abti: As discussed offline, jobState is never null in all practical purposes. We will however still add a check once we take this bug fix to mainstream: https://github.com/linkedin/gobblin/issues/1022

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
abti: Yes, this is being tracked here: https://github.com/linkedin/gobblin/issues/1022 
More folks have run into it. Its high priority, we will fix it soon.

shirshanka:
LGTM

abti:
@shirshanka Thanks for the review

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSClusterLauncher.java
abti: Thanks

sahilTakiar:
Awesome feature!

May want to consider integration with something like JClouds: https://jclouds.apache.org/ in the future. I'm not sure has an abstraction for auto scaling groups, but it should work well for ec2 and s3.

gobblin-aws/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+  testCompile project(path: ':gobblin-cluster', configuration: 'tests')
+  testCompile project(":gobblin-example")
+
+  testCompile externalDependency.testng
+  testCompile externalDependency.curatorFramework
+  testCompile externalDependency.curatorTest
+  testCompile externalDependency.mockito
+  testCompile externalDependency.powermock
+
+  compile files('./src/main/resources/helix-core-0.6.6-SNAPSHOT.jar')
```
sahilTakiar: do u need this, shouldn't it inherit from `gobblin-cluster`?

gobblin-aws/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+  compile externalDependency.findBugsAnnotations
+
+  testCompile project(path: ':gobblin-cluster', configuration: 'tests')
+  testCompile project(":gobblin-example")
+
+  testCompile externalDependency.testng
+  testCompile externalDependency.curatorFramework
+  testCompile externalDependency.curatorTest
+  testCompile externalDependency.mockito
+  testCompile externalDependency.powermock
```
sahilTakiar: Just curios why powermock is needed?

conf/aws/application.conf
```diff
@@ -0,0 +1,162 @@
...
...,lombok-1.16.8.jar,lz4-1.2.0.jar,mail-1.4.1.jar,maven-scm-api-1.4.jar,maven-scm-provider-svn-commons-1.4.jar,maven-scm-provider-svnexe-1.4.jar,metrics-core-2.2.0.jar,metrics-core-3.1.0.jar,metrics-graphite-3.1.0.jar,metrics-jvm-3.1.0.jar,mina-core-1.1.7.jar,mockito-core-1.10.19.jar,mysql-connector-java-5.1.38.jar,netty-3.2.3.Final.jar,netty-3.7.0.Final.jar,objenesis-2.1.jar,okhttp-2.4.0.jar,okio-1.4.0.jar,opencsv-2.3.jar,paranamer-2.3.jar,parseq-1.3.6.jar,pegasus-common-1.15.9.jar,pentaho-aggdesigner-algorithm-5.1.5-jhyde.jar,plexus-utils-1.5.6.jar,protobuf-java-2.5.0.jar,quartz-2.2.3.jar,r2-1.15.9.jar,reflections-0.9.10.jar,regexp-1.3.jar,restli-client-1.15.9.jar,restli-common-1.15.9.jar,restli-docgen-1.15.9.jar,restli-netty-standalone-1.15.9.jar,restli-server-1.15.9.jar,restli-tools-1.15.9.jar,retrofit-1.9.0.jar,scala-library-2.11.8.jar,scala-parser-combinators_2.11-1.0.2.jar,scala-xml_2.11-1.0.2.jar,servlet-api-2.5-20081211.jar,servlet-api-2.5.jar,slf4j-api-1.7.21.jar,slf4j-log4j12-1.7.21.jar,snappy-0.3.jar,snappy-java-1.1.1.7.jar,stax-api-1.0-2.jar,stax-api-1.0.1.jar,testng-6.9.10.jar,transaction-api-1.1.jar,velocity-1.7.jar,xmlenc-0.52.jar,zkclient-0.5.jar,zookeeper-3.4.6.jar"
```
sahilTakiar: this list seems really long, is there a better way to do this? would this list need to be updated if a new dependency is introduced?

conf/aws/quartz.properties
```diff
@@ -0,0 +1,3 @@
+org.quartz.scheduler.instanceName = LocalJobScheduler
```
sahilTakiar: seems to be the same as the original `quartz.properties` file, can we just use that?

conf/aws/application.conf
```diff
@@ -0,0 +1,162 @@
...
+writer.fs.uri=${fs.uri}
+state.store.fs.uri=${fs.uri}
+
+## Writer related configuration properties
+writer.destination.type=HDFS
+writer.output.format=AVRO
+writer.staging.dir=${gobblin.aws.work.dir}"/task-staging"
+writer.output.dir=${gobblin.aws.work.dir}"/task-output"
+
+## Data publisher related configuration properties
```
sahilTakiar: does data not get published to s3 at the end of the job? if we just add `data.publisher.fs.uri=s3a://gobblin-bucket-name/` it should work

gobblin-aws/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+artifacts {
+  tests testJar
+}
+
+test {
+  workingDir rootProject.rootDir
+  maxParallelForks = 1
+}
+
+clean {
```
sahilTakiar: is this needed?

gobblin-aws/build.gradle
```diff
@@ -0,0 +1,104 @@
...
+  testCompile externalDependency.testng
+  testCompile externalDependency.curatorFramework
+  testCompile externalDependency.curatorTest
+  testCompile externalDependency.mockito
+  testCompile externalDependency.powermock
+
+  compile files('./src/main/resources/helix-core-0.6.6-SNAPSHOT.jar')
+}
+
+task testJar(type: Jar, dependsOn: testClasses) {
```
sahilTakiar: don't think this task and the `tests` configuration are needed

gobblin-aws/src/main/java/gobblin/aws/AWSClusterSecurityManager.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private final Config config;
+
+  private volatile String serviceAccessKey;
+  private volatile String serviceSecretKey;
+  private volatile boolean clientAssumeRole;
+  private volatile String clientRoleArn;
+  private volatile String clientExternalId;
+  private volatile String clientSessionId;
+  private volatile long lastRefreshTimeInMillis;
```
sahilTakiar: can't these all be `final` rather than `volatile`, they all seem to be only set once and derive there values from a `Config` object which is immutable.

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSConfigurationKeys.java
```diff
@@ -0,0 +1,147 @@
...
...lombok-1.16.8.jar,lz4-1.2.0.jar,mail-1.4.1.jar,maven-scm-api-1.4.jar,maven-scm-provider-svn-commons-1.4.jar,maven-scm-provider-svnexe-1.4.jar,metrics-core-2.2.0.jar,metrics-core-3.1.0.jar,metrics-graphite-3.1.0.jar,metrics-jvm-3.1.0.jar,mina-core-1.1.7.jar,mockito-core-1.10.19.jar,mysql-connector-java-5.1.38.jar,netty-3.2.3.Final.jar,netty-3.7.0.Final.jar,objenesis-2.1.jar,okhttp-2.4.0.jar,okio-1.4.0.jar,opencsv-2.3.jar,paranamer-2.3.jar,parseq-1.3.6.jar,pegasus-common-1.15.9.jar,pentaho-aggdesigner-algorithm-5.1.5-jhyde.jar,plexus-utils-1.5.6.jar,protobuf-java-2.5.0.jar,quartz-2.2.3.jar,r2-1.15.9.jar,reflections-0.9.10.jar,regexp-1.3.jar,restli-client-1.15.9.jar,restli-common-1.15.9.jar,restli-docgen-1.15.9.jar,restli-netty-standalone-1.15.9.jar,restli-server-1.15.9.jar,restli-tools-1.15.9.jar,retrofit-1.9.0.jar,scala-library-2.11.8.jar,scala-parser-combinators_2.11-1.0.2.jar,scala-xml_2.11-1.0.2.jar,servlet-api-2.5-20081211.jar,servlet-api-2.5.jar,slf4j-api-1.7.21.jar,slf4j-log4j12-1.7.21.jar,snappy-0.3.jar,snappy-java-1.1.1.7.jar,stax-api-1.0-2.jar,stax-api-1.0.1.jar,testng-6.9.10.jar,transaction-api-1.1.jar,velocity-1.7.jar,xmlenc-0.52.jar,zkclient-0.5.jar,zookeeper-3.4.6.jar";
```
sahilTakiar: is there a more dynamic way of compiling this list?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -217,10 +218,22 @@ private GobblinHelixJobScheduler buildGobblinHelixJobScheduler(Config config, Pa
...
    */
   private JobConfigurationManager buildJobConfigurationManager(Config config) {
-    Optional<String> jobConfPackagePath =
-        config.hasPath(GobblinClusterConfigurationKeys.JOB_CONF_PATH_KEY) ? Optional
-            .of(config.getString(GobblinClusterConfigurationKeys.JOB_CONF_PATH_KEY)) : Optional.<String>absent();
-    return new JobConfigurationManager(this.eventBus, jobConfPackagePath);
+    return create(config);
+  }
+
+  private JobConfigurationManager create(Config config) {
```
sahilTakiar: more descriptive method name

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -113,7 +112,8 @@
 
   protected final String applicationId;
 
-  public GobblinClusterManager(String clusterName, String applicationId, Config config) throws Exception {
+  public GobblinClusterManager(String clusterName, String applicationId, Config config,
+      Optional<Path> appWorkDirOptional) throws Exception {
```
sahilTakiar: what is `appWorkDirOptional` added for?

conf/aws/application.conf
```diff
@@ -0,0 +1,162 @@
...
+
+# Leave as-is for sample wikipedia pull job
+# -- OR --
+# Specify S3 URI for job configuration:
+# Steps to obtain this URI:
+# 1. Create a S3 folder in a bucket, say the bucket name is "s3bucket" and folder is "/gobblin-confs/job-conf/"
+# 2. Zip your pull files in a zip and upload to this folder. Say your zip file name is myPullFiles.zip
+# 3. Specify S3 URI for this zip file in the config below
+#    For the description above: "https://s3-region.amazonaws.com/s3bucket/gobblin-confs/job-conf/myPullFiles.zip"
+gobblin.aws.job.conf.s3.uri="https://s3-us-west-2.amazonaws.com/gobblin-libs/job-config/wikipedia.zip"
```
sahilTakiar: I haven't looked at #1066 but can the job alternation monitor be used here?

conf/aws/application.conf
```diff
@@ -0,0 +1,162 @@
...
+
+#########################################################################
+## Advanced config properties. You need not change anything below      ##
+#########################################################################
+
+## Gobblin version
+# Note:
+# 1. If you change Gobblin version here, make sure that those version jars are
+# .. on S3 url referred by: gobblin.aws.master.s3.jars.uri
+# 2. Version below is available on the S3 at: https://s3-us-west-2.amazonaws.com/gobblin-libs/latest-jars/
```
sahilTakiar: is this link suppose to be publicaly accessible? i can't seem to find it

will this go away when the gobblin ami is built?

build.gradle
```diff
@@ -136,10 +138,23 @@ ext.dropwizardMetricsVersion = '3.1.0'
...
+  imports {
+    mavenBom 'com.amazonaws:aws-java-sdk-bom:1.11.8'
+  }
+}
+
 ext.externalDependency = [
   "antlrRuntime": "org.antlr:antlr-runtime:3.5.2",
   "avro": "org.apache.avro:avro:" + avroVersion,
   "avroMapredH2": "org.apache.avro:avro-mapred:" + avroVersion + ":hadoop2",
+  "awsCore": "com.amazonaws:aws-java-sdk-core:1.11.8",
```
sahilTakiar: may want to define `ext.awsJavaSdkVersion=1.11.8` so that they can all be upgraded together

gobblin-aws/build.gradle
abti: Yes, it can be inherited. We are thinking to upgrade Helix version. Will clean this up along with it.

gobblin-aws/build.gradle
abti: It is for some tests which wip, but redundant at the moment

conf/aws/application.conf
abti: Yes this in not neat. We should be able to generate the dependency list automatically via gradle to some file at build time. Dependency updation is a manual process at the moment

conf/aws/quartz.properties
abti: Yes, will clean that up.

conf/aws/application.conf
abti: In current setup, data is getting written to NFS though via HDFS writer

gobblin-aws/build.gradle
abti: No this is residual from gobblin-cluster, should be cleaned up

gobblin-aws/src/main/java/gobblin/aws/AWSClusterSecurityManager.java
abti: lastRefreshTimeInMillis is being set to System.currentTimeMillis() at line 136 everytime the refresh happens. refreshIntervalInMinutes is being derived from config though

gobblin-aws/src/main/java/gobblin/aws/GobblinAWSConfigurationKeys.java
abti: Yes, a future optimization should be generating the dependency list at build time

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
abti: Will update this

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
abti: The work dir was being specified in gobblin-yarn config but not being used anywhere. This helps us make use of it if specified, though gobblin-yarn still needs to be fixed for it. There was also an issue around it reported by someone on google groups for gobblin-yarn

conf/aws/application.conf
abti: Not verbatim but it can be extended, also we should make use of new JobLauncher pushed by Chavdar - as per my understanding it is very much modeled around gobblin-cluster but I need to dig deeper

conf/aws/application.conf
abti: This should be publically accessible, I will cross check. Yes, this will all go away with AMI

build.gradle
abti: Will change it

abti:
@sahilTakiar Thanks for the review. Took me sometime to get back due to travel and backlog. I will incorporate your suggestions along with changes to gobblin-cluster that we need to do to move to new JobLauncher 
Writing an impl for JClouds is a good idea. I will try to get to that in one of InDays :)

PR #1090, opened 2016-07-07
Teradata JDBC Extractor and Source

ypopov:
Do not required extended access rights  - schema generated from generic JDBC ResultSetMetaData. Works with views and custom queries  (use source.querybased.is.metadata.column.check.enabled=false for derived columns). Tested against Teradata version 13 and 15.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
@@ -198,7 +198,7 @@ public class TeradataExtractor extends JdbcExtractor {
     
     String port = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT,"1025");
     String database = this.workUnit.getProp(ConfigurationKeys.SOURCE_QUERYBASED_SCHEMA);
-    String defaultUrl = urlPrefix + host.trim() + "/DATABASE=" + database.trim() + ",DBS_PORT=" + port.trim() ;
+    String defaultUrl = urlPrefix + host.trim() + "/TYPE=FASTEXPORT,DATABASE=" + database.trim() + ",DBS_PORT=" + port.trim() ;
     // use custom url from source.conn.host if Teradata jdbc url available
     return host.contains(urlPrefix) ? host.trim() : defaultUrl;
   }
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
@@ -1,0 +1,334 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.source.extractor.extract.jdbc;
+
+import java.io.IOException;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
```diff
@@ -0,0 +1,334 @@
...
+      List<Predicate> predicateList) throws IOException {
+    return null;
+  }
+
+  @Override
+  public String getConnectionUrl() {
+    String host = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
+    String port = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT,"1025");
+    String database = this.workUnit.getProp(ConfigurationKeys.SOURCE_QUERYBASED_SCHEMA);
+    String url = "jdbc:teradata://" + host.trim() + "/DATABASE=" + database.trim() + ",DBS_PORT=" + port.trim() ;
```
lbendig: When using the Teradata JDBC driver the `TYPE=FASTEXPORT` parameter could be added to the connection string. Optionally a default charset as well (e.g: `CHARSET=utf16`).

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
```diff
@@ -0,0 +1,334 @@
...
+            .put("varchar", "string")
+            .put("byte", "bytes")
+            .put("varbyte", "bytes")
+            .put("date", "date")
+            .put("time", "time")
+            .put("timestamp", "timestamp")
+            .put("clob", "string")
+            .put("blob", "string")
+            .put("structured udt", "array")
+            .build();
```
lbendig: In our version we have these additional mappings:
```
.put("double precision", "float")
.put("numeric", "double")
.put("real", "float")
.put("character", "string")
.put("char varying", "string")
.put("character varying", "string")
.put("long varchar", "string")
.put("interval", "string")
```

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
```diff
@@ -0,0 +1,334 @@
...
+    if (isNullOrEmpty(watermarkFilter)) {
+      watermarkFilter = "1=1";
+    }
+    query = query.replace(this.getOutputColumnProjection(), columnProjection)
+        .replace(ConfigurationKeys.DEFAULT_SOURCE_QUERYBASED_WATERMARK_PREDICATE_SYMBOL, watermarkFilter);
+    String sampleFilter = this.constructSampleClause();
+    query = query + sampleFilter;
+
+    if (!isNullOrEmpty(sampleFilter)) {
+      query = "SELECT COUNT(1) FROM (" + query.replace(" COUNT(1) ", " 1 at t") + ") temp";
```
lbendig: Teradata count returns an `INT`. For big tables (rows > 2^31-1) this query produces a numerical overflow.
I'd suggest to have `CAST(COUNT(1) AS BIGINT)` instead.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/TeradataExtractor.java
```diff
@@ -188,12 +192,15 @@ public TeradataExtractor(WorkUnitState workUnitState) {
...
+    String urlPrefix = "jdbc:teradata://"; 
     String host = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
+    checkArgument(!isNullOrEmpty(host), "Connectionn host cannot be null or empty at %s", ConfigurationKeys.SOURCE_CONN_HOST_NAME);
+    
     String port = this.workUnit.getProp(ConfigurationKeys.SOURCE_CONN_PORT,"1025");
     String database = this.workUnit.getProp(ConfigurationKeys.SOURCE_QUERYBASED_SCHEMA);
-    String url = "jdbc:teradata://" + host.trim() + "/DATABASE=" + database.trim() + ",DBS_PORT=" + port.trim() ;
-    
-    return url;
+    String defaultUrl = urlPrefix + host.trim() + "/DATABASE=" + database.trim() + ",DBS_PORT=" + port.trim() ;
```
lbendig: I think it's worth to add the `TYPE=FASTEXPORT` connection parameter by default.

lbendig:
Tested the `TeradataExtractor` for full dumps, custom watermark ranges, custom queries and sampling, looks good!

chavdar:
LGTM.

PR #1091, opened 2016-07-10
Confluent schema registry example for kafka writer

shirshanka:
@abti, @sahilTakiar :  this should be a quick one.


abti:
LGTM but Travis build seems to be failing because of Mockito change, you probably need to fix Mockito dependency at a few more places :) 

[Link to Travis](https://travis-ci.org/linkedin/gobblin/builds/143679141 )

shirshanka:
Fixed the build.

chavdar:
For future reference, please make sure that any new dependencies are in the artifactory before adding them. Otherwise, this causes unpredictable build failures for the internal build.

PR #1114, opened 2016-07-15
Config key annotation to auto generate docs

pcadabam:
**Changes**
------------------------------
This pr creates a way to automatically generate documentation for configuration keys used in gobblin.
Every configuration key defined should be annotated with ```@ConfigKey``` annotation. The tool automatically reads the Javadoc associated with this key. All the config keys and docs are finally written to an HTML file.
The tool uses java's doclet APIs to get Javadoc as compiled jars do not have comments or Javadocs. It extends gradle's Javadoc task with custom doclet. The task name is ```generateConfigDocs``` defined in ```gobblin-distribution```. The HTML is generated using Apache ```FreeMarker``` templates

**Files**
---------------------------
- ```ConfigKey``` is the annotation that needs to be added to every ```private static String KEY_NAME="key.something" that has to be documented
- ```ConfigWikiGenerator``` is the custom doclet class that scans all classes for fields annotated with ```ConfigKey``` and writes them to an HTML file (currently /tmp/configDocs.html)
-```configDocs.ftl`` is the Freemarker html template 

The ```ConfigWikiGenerator``` has some example configuration keys that are annotated with ```ConfigKey```. You can see those keys show up in the HTML below.

**Examples**
-------------------------------
<pre>
  /**
   * Root directory where task state files are stored
   */
  @ConfigKey
  public static final String STATE_STORE_ROOT_DIR_KEY = "state.store.dir";
  /**
   * File system URI for file-system-based task store *
   */
  @ConfigKey
  public static final String STATE_STORE_FS_URI_KEY = "state.store.fs.uri";

  /**
   * Enable / disable state store
   */
  @ConfigKey(def = "false")
  public static final String STATE_STORE_ENABLED = "state.store.enabled";

  /**
   * Set this to only log and not delete data
   */
  @ConfigKey(required = false, def = "false", group = "Retention")
  private static final String RETENTION_SIMULATE_KEY = "gobblin.retention.simulate";

  /**
   * Set his to skip trash and delete data
   */
  @ConfigKey(required = false, def = "false", group = "Retention")
  private static final String RETENTION_SKIP_TRASH_KEY = "gobblin.retention.skiptrash";

  /**
   * Name of the destination db
   */
  @ConfigKey(required = true, group = "Hive Conversion")
  private static final String CONVERSION_DESTINATION_DB_NAME_KEY = "gobblin.conversion.destination.dbName";

  /**
   * Name of the destination table
   */
  @ConfigKey(required = true, group = "Hive Conversion")
  private static final String CONVERSION_DESTINATION_TABLE_NAME_KEY = "gobblin.conversion.destination.tableName";

  /**
   * List of tables to be whitelisted of the form [dbpattern.tablepattern1|tablepattern2|...]
   */
  @ConfigKey(required = true, def = "*.*", group = "Hive Distcp")
  private static final String HIVE_DATASET_FINDER_WHITELIST = "hive.dataset.whitelist";

  /**
   * List of tables to blacklisted of the form [dbpattern.tablepattern1|tablepattern2|...]
   */
  @ConfigKey(required = false, def = "-", group = "Hive Distcp")
  private static final String HIVE_DATASET_FINDER_BLACKLIST = "hive.dataset.blacklist";

</pre>

**Generated HTML**
------------------------------------

screen shot 2016-07-15 at 1 51 18 am

The task can be invoked by running ```./gradlew clean gobblin-distribution:generateConfigDocs```. Later we expect this task to always run with ```./gradlew build```

@chavdar @shirshanka @abti please have a look


gobblin-utility/build.gradle
```diff
@@ -12,7 +12,7 @@ apply plugin: 'java'
 
 dependencies {
   compile project(":gobblin-api")
-
+  compile files("${System.properties['java.home']}/../lib/tools.jar")
```
abti: why do we need this?

gobblin-utility/src/main/java/gobblin/util/wiki/ConfigWikiGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+    config.setClassForTemplateLoading(ConfigWikiGenerator.class, "templates");
+    config.setDefaultEncoding("UTF-8");
+    config.setTemplateExceptionHandler(TemplateExceptionHandler.RETHROW_HANDLER);
+
+    Map<String, Object> input = new HashMap<String, Object>();
+
+    input.put("title", "Configuration Keys");
+    input.put("groupConfigKeys", new TreeMap<>(groupConfigKeys.asMap()));
+
+    Template template = config.getTemplate(DOCS_HTML_TEMPLATE_RESOURCE);
```
abti: would be nice to have a template for config file, that generates config file for a give source and its related tightly coupled components (such as extractor)
may be we can leverage grouping information for that or define it as an a new property

gobblin-api/src/main/java/gobblin/configuration/ConfigKey.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+  boolean required() default false;
+  /**
+   * Default value for this config key
+   */
+  String def() default "None";
+  /**
+   * Group of this config key. All keys of the same group will be documented together
+   */
+  String group() default "";
```
abti: one property can be 'sample' value that helps generate a config file for working simple example / case study (for eg. a case study example config for MysqlSource reading from localhost:3306 with user: admin pass: admin can be auto generated if had such property)

gobblin-utility/build.gradle
pcadabam: Because tools.jar is in JRE and needed at compile for the ```Doclet``` APIs

gobblin-utility/src/main/java/gobblin/util/wiki/ConfigWikiGenerator.java
pcadabam: Yes that was the idea behind groups

gobblin-api/src/main/java/gobblin/configuration/ConfigKey.java
pcadabam: The sample is same as ```def()```

abti:
https://issues.apache.org/jira/browse/GOBBLIN-93

Please update your PR title with following prefix: [GOBBLIN-93]

PR #1116, opened 2016-07-16
Fix findbugs failure

pcadabam:


pcadabam:
Self approving trivial change.

PR #1119, opened 2016-07-19
Changed concatenated queries string to list in Hive converter publisher

abti:


abti:
Minor change, self approving

PR #1120, opened 2016-07-19
Added Hive query generation to optionally support explicit database names

abti:
Added Hive query generation to optionally support explicit database names


abti:
Minor change, self approving.

PR #1121, opened 2016-07-19
Made Hive table rename backward compatible with Hive v0.13

abti:
Made Hive table rename backward compatible with Hive v0.13
 
In Hive v0.13, Hive does not support fully qualified Hive table names such as db.table for table rename. So, changed publish commands to specify 'use dbName' as a precursor to the rename


abti:
Minor change, self approving

PR #1122, opened 2016-07-19
Made changes to handle Hive-6129 (inverted exchange partition bug) and corresponding support for backward incompatible changes in Hive

abti:
Made changes to handle Hive-6129 (inverted exchange partition bug) and corresponding support for backward incompatible changes in Hive: 

- Hive-4095 introduced the partition exchange feature for first time in v0.12
- HIVE-6129 found out that the advertised behavior was inverse (source was being treated as target and vice-versa) and a backward incompatible fix was introduced

So, changed Gobblin code to handle all versions of Hive by taking hint from config about the Hive version and creating the correct corresponding Hive statement 

Also, added a few javadocs


abti:
Minor bug fix related to Hive bugs, self approving

PR #1123, opened 2016-07-19
Fix for HIVE-11194: exchange partition does not work properly if data directory for destination partitions pre-exists

abti:
If a partition is being exchanged into a table but the data directory location for the partition already exists, Hive moves data within a sub-directory of it rather than cleanly overwriting it.

More details here: HIVE-11194

So, adding a pre publish hook to cleanup the existing directory (if any) of an existing partition in destination table before issuing exchange partition from staging to destination.


abti:
Minor changes, self approving.

PR #1124, opened 2016-07-19
Added comments to indicate if DDL and DML generation are being done using source or destination schemas in Hive Avro to ORC

abti:
Added comments to indicate if DDL and DML generation are being done using source or destination schemas in Hive Avro to ORC


abti:
Added couple of log lines. Self approving.

PR #1126, opened 2016-07-19
Make distcp publisher safer: renameRecursively fails appropriately, h…

ibuenros:
…ive registration fails if location doesn't exist.


ibuenros:
@chavdar @pcadabam can you review? Thanks!

pcadabam:
LGTM

PR #1129, opened 2016-07-20
launcher refactoring: FSJobCatalog

autumnust:
initial commit, basic implementation, require further commits to complete.


gobblin-runtime/src/main/java/gobblin/refactor/FSJobCatalog.java
```diff
@@ -0,0 +1,85 @@
...
+  private Configuration conf = new Configuration();
+  private FileSystem fs;
+
+  /**
+   * Initialize the JobCatalog, fetch all jobs in jobConfDirPath.
+   * @param jobConfDirPath
+   * @throws Exception
+   */
+  public FSJobCatalog(Path jobConfDirPath)
+      throws Exception {
```
chavdar: Please avoid throwing a generic Exception as it does not tell anything about the nature of exceptions that can be thrown. It defeats the purpose of using checked exceptions. Either use specific exception types or just let the implementation throw RuntimeException.

gobblin-runtime/src/main/java/gobblin/refactor/FSJobCatalog.java
```diff
@@ -0,0 +1,85 @@
...
+        LOGGER.info("Loading job configurations from " + jobConfDirPath);
+        Properties properties = new Properties();
+        properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY, jobConfDirPath.toString());
+        persistedJobs = JobCatalogUtils.jobSpecListToMap(JobCatalogUtils.loadGenericJobConfigs(properties));
+        LOGGER.info("Loaded " + persistedJobs.size() + " job configuration(s)");
+      } else {
+        LOGGER.warn("Job configuration directory " + jobConfDirPath + " not found");
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
```
chavdar: Please use logging.
Do not swallow exceptions. Just let the exception propagate if you can't handle it.

autumnust:
Switch to another PR as I carelessly messed this up.

PR #1130, opened 2016-07-20
Initial commit for new gobblin launch interfaces

chavdar:
@ibuenros @pcadabam @tuGithub @ydai1124 @abti @vasanthrajamani @autumnust Can you review ?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.runtime.api;
+
+import java.net.URI;
+import java.util.Collection;
+
+import gobblin.annotation.Alpha;
+
+/**
+ * A catalog of all the {@link JobSpec}s a Gobblin instance is currently aware of.
+ */
+@Alpha
+public interface JobCatalog {
+  /** Returns an immutable {@link Collection} of {@link JobSpec}s that are known to the catalog. */
+  public Collection<JobSpec> getJobs();
+
+  /** Get a {@link JobSpec} by uri. */
+  public JobSpec getJobSpec(URI uri);
+
+  /**
+   * Adds a {@link JobCatalogListener} that will be invoked upon updates on the
+   * {@link JobCatalog}. Upon registration {@link JobCatalogListener#onAddJob(JobSpec)} will be
+   * invoked for all pre-existing jobs in the JobCatalog.
+   */
+  public void addListener(JobCatalogListener jobListener);
+
+  /**
+   * Removes the specified listener. No-op if not registered.
+   */
+  public void removeListener(JobCatalogListener jobListener);
+}
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
@@ -22,20 +22,20 @@ import gobblin.annotation.Alpha;
 @Alpha
 public interface JobCatalog {
   /** Returns an immutable {@link Collection} of {@link JobSpec}s that are known to the catalog. */
-  public Collection<JobSpec> getJobs();
+  Collection<JobSpec> getJobs();
 
   /** Get a {@link JobSpec} by uri. */
-  public JobSpec getJobSpec(URI uri);
+  JobSpec getJobSpec(URI uri);
 
   /**
    * Adds a {@link JobCatalogListener} that will be invoked upon updates on the
    * {@link JobCatalog}. Upon registration {@link JobCatalogListener#onAddJob(JobSpec)} will be
    * invoked for all pre-existing jobs in the JobCatalog.
    */
-  public void addListener(JobCatalogListener jobListener);
+  void addListener(JobCatalogListener jobListener);
 
   /**
    * Removes the specified listener. No-op if not registered.
    */
-  public void removeListener(JobCatalogListener jobListener);
+  void removeListener(JobCatalogListener jobListener);
 }
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecution.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecution.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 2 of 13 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecution.java
```diff
@@ -0,0 +1,24 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.runtime.api;
+
+import java.net.URI;
+
+/**
+ * Identifies a specific execution of a {@link JobSpec}
+ */
+public interface JobExecution {
```
ibuenros: I think this should be a class. For a distributed job driver (e.g. job driver is a different jvm), this is the object that would be sent over the network, so it is useful to have a set structure.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpec.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Defines a Gobblin Job that can be run once, or multiple times. A {@link JobSpec} is
+ * {@link Configurable} so it has an associated {@link Config}, along with other mandatory
+ * properties such as a uri, description, and version. A {@link JobSpec} is
+ * uniquely identified by its uri (containing group and name).
+ *
+ */
+@Alpha
+public interface JobSpec extends Configurable {
```
ibuenros: Is there a reason for this to be an interface?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpecScheduler.java
```diff
@@ -0,0 +1,22 @@
...
+package gobblin.runtime.api;
+
+import gobblin.annotation.Alpha;
+
+/**
+ *  Placeholder
+ */
+@Alpha
+public interface JobSpecScheduler {
+
```
ibuenros: Does this have any methods?

chavdar: Eventually. :-) It is a placeholder for now.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpec.java
chavdar: Mostly because it is easier to separate what is the public API on which other components can depend on from the implementation details. Don't feel strongly about interface or class.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpec.java
```diff
@@ -0,0 +1,46 @@
...
+ * uniquely identified by its uri (containing group and name).
+ *
+ */
+@Alpha
+@Data
+public class JobSpec implements Configurable {
+  /** Job config as a typesafe config object*/
+  final Config config;
+  /** Job config as a properties collection for backwards compatibility */
+  final Properties configAsProperties;
```
ibuenros: Is this necessary? it could be computed on demand.

ibuenros:
One comment, otherwise LGTM.

abti:
LGTM

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
```diff
@@ -0,0 +1,41 @@
...
+   * Adds a {@link JobCatalogListener} that will be invoked upon updates on the
+   * {@link JobCatalog}. Upon registration {@link JobCatalogListener#onAddJob(JobSpec)} will be
+   * invoked for all pre-existing jobs in the JobCatalog.
+   */
+  public void addListener(JobCatalogListener jobListener);
+
+  /**
+   * Removes the specified listener. No-op if not registered.
+   */
+  public void removeListener(JobCatalogListener jobListener);
```
autumnust: Can the `public` modifier be omitted for method definition in interface, as it is implicitly public ?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpec.java
chavdar: For now, I think it is good to cache the values as I expect them to be frequently used until we have wider adoption of typesafe config.

I'll add a comment there.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
chavdar: Done.

gobblin-runtime/src/main/java/gobblin/runtime/api/package-info.java
```diff
@@ -0,0 +1,57 @@
...
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+/**
+ * The Gobblin Execution Model.
+ *
+ * <p>The Gobblin launcher components:
```
sahilTakiar: How do these interfaces interact with `gobblin.runtime.app.ApplicationLauncher`?

gobblin-runtime/src/main/java/gobblin/runtime/api/StandardRunControls.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+package gobblin.runtime.api;
+
+import java.util.concurrent.TimeoutException;
+
+import gobblin.annotation.Alpha;
+
+/** Defines a standard set of methods for running Gobblin components. */
+@Alpha
+public interface StandardRunControls {
```
sahilTakiar: Seems similar to Guava's `Service` or even `AbstractService`, can those be used instead? They are already extensively used throughout Gobblin.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionLauncher.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+package gobblin.runtime.api;
+
+import gobblin.annotation.Alpha;
+
+/**
+ * Placeholder
+ */
+@Alpha
+public interface JobExecutionLauncher {
```
sahilTakiar: What does this do? Difference between `JobLauncher`?

gobblin-runtime/src/main/java/gobblin/runtime/api/StandardRunControls.java
chavdar: Yes, see https://github.com/linkedin/gobblin/pull/1133

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionLauncher.java
chavdar: Actually, it will replace to JobLauncherFactory but the interface will use higher level of abstractions. It will be in terms of JobExecutions rather than Properties. The lifecycle of a job goes through JobSpec -> JobSchedule -> JobExecution and the interfaces should reflect that.

gobblin-runtime/src/main/java/gobblin/runtime/api/package-info.java
chavdar: In a meeting, we decided to use the terminology "Gobblin Instance" instead of "Gobblin Application" with respect to a instantiation of Gobblin that runs a collection of jobs as those job are not necessarily related.

The interface itself will be deprecated as it looks very generic and does not define (in terms of APIs) what a Gobblin application actually is.

There are 4 implementations of this interface currently: AzkabanJobLauncher, CliLocalJobLauncher, CliMRJobLauncher, and ServiceBasedAppLauncher. The first three implement both ApplicationLauncher and JobLauncher which is kind of strange. Running a Gobblin Application (aka Instance) is more than just executing a job -- it is controlling the entire lifecycle of jobs. These three should focus on their primary responsibility -- executing a single job. The lifecycle management of Jobs should be left to the GobblinInstanceLauncher/GobblinInstanceDriver . We'll probably remove the "implements ApplicationLauncher" part (may keep the start/stop methods for backwards compatibility).

ServiceBasedAppLauncher will become part of a standard implementation of GobblinInstanceLauncher that sets up a standard environment for running Gobblin jobs.

PR #1131, opened 2016-07-21
Gobblin launcher standard implementations

chavdar:
Add the following to :
- Builder for JobSpec
- A helper class DefaultConfigurableImpl for implementations of Configurable
- A helper class JobCatalogListenersList to manage callbacks to JobCatalogListener instances
- A basic implementation of JobCatalog which stores JobSpecs in memory
- Unit tests for the above

@autumnust @ibuenros Can you review?


chavdar:
I will merge that since it contains a fix for the broken build. This should not affect anything. I will apply any review comments additionally.

PR #1133, opened 2016-07-22
Adding caching decorators for JobCatalog

chavdar:
A follow-up on https://github.com/linkedin/gobblin/pull/1131 .

Adds:
- Throws an exception in InMemoryJobCatalog.getJobSpec() if key is missing.
- Mutable and read-only caching decorators for JobCatalogs
- Unit tests.
- A simple change in  KafkaDeserializerExtractorTest to suppress an error in eclipse.

@ibuenros @autumnust Can you review?


chavdar:
@sahilTakiar @ibuenros @autumnust  any comments?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
```diff
@@ -24,7 +24,10 @@
   /** Returns an immutable {@link Collection} of {@link JobSpec}s that are known to the catalog. */
   Collection<JobSpec> getJobs();
 
-  /** Get a {@link JobSpec} by uri. */
+  /**
+   * Get a {@link JobSpec} by uri.
+   * @throws JobSpecNotFoundException if no such JobSpec exists
+   **/
   JobSpec getJobSpec(URI uri);
```
ibuenros: can you declare the `JobSpecNotFoundException`?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpecNotFoundException.java
```diff
@@ -0,0 +1,31 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.runtime.api;
+
+import java.net.URI;
+
+/**
+ * An exception thrown when a JobSpec with a given URI is not found.
+ */
+public class JobSpecNotFoundException extends RuntimeException {
```
ibuenros: Why is this a `RuntimeException`?

gobblin-runtime/src/main/java/gobblin/runtime/job_catalog/CachingJobCatalog.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public JobSpec getJobSpec(URI uri) {
+    try {
+      return _cache.getJobSpec(uri);
+    }
+    catch (RuntimeException e) {
+      return _fallback.getJobSpec(uri);
```
ibuenros: When would this happen? The caching layer, as a listener, is notified of all jobs in the fallback catalog right?

gobblin-runtime/src/test/java/gobblin/runtime/std/TestMutableCachingJobCatalog.java
```diff
@@ -0,0 +1,93 @@
...
+  public void test() throws Exception {
+    InMemoryJobCatalog baseCat =
+        new InMemoryJobCatalog(Optional.<Logger>of(LoggerFactory.getLogger("baseCat")));
+    MutableCachingJobCatalog cachedCat =
+        new MutableCachingJobCatalog(baseCat, Optional.<Logger>of(LoggerFactory.getLogger("cachedCat")));
+
+    JobCatalogListener l = Mockito.mock(JobCatalogListener.class);
+    cachedCat.addListener(l);
+
+    JobSpec js1_1 = JobSpec.builer("test:job1").withVersion("1").build();
```
ibuenros: typo: `builer` -> `builder`

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
chavdar: It's a RuntimeException. I think under normal circumstances, it should rarely happen so I am not sure the API users should be burdened with always doing try-catch. What do you think?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpecNotFoundException.java
chavdar: See above. I don't think this should be a frequent enough occurrence to warrant a checked exception with the required try-catch . I don't feel too strongly about this so I am ok making it a checked exception.

gobblin-runtime/src/main/java/gobblin/runtime/job_catalog/CachingJobCatalog.java
chavdar: Mostly as a safeguard against timing issues for read-after-write scenarios. If the callbacks are asynchronous, it is possible that the job is not in the cache yet.

gobblin-runtime/src/test/java/gobblin/runtime/std/TestMutableCachingJobCatalog.java
chavdar: Fixed.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
ibuenros: My concern is if API users are not aware of the exception, and the exception does show up (eg a race condition where a spec has just been deleted but the notification has not been processed yet), the application will crash. If it is declared, user will be aware of it.

gobblin-runtime/src/main/java/gobblin/runtime/job_catalog/CachingJobCatalog.java
ibuenros: Makes sense.

gobblin-runtime/src/main/java/gobblin/runtime/job_catalog/CachingJobCatalog.java
```diff
@@ -0,0 +1,91 @@
...
+    _log = log.isPresent() ? log.get() : LoggerFactory.getLogger(getClass());
+    _fallback = fallback;
+    _cache = new InMemoryJobCatalog(log);
+    _fallback.addListener(new FallbackCatalogListener());
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Collection<JobSpec> getJobs() {
+    return _cache.getJobs();
```
ibuenros: Should this block until the fallback finishes notifying the cache layer of jobs? Otherwise the return value is not actually representative of `fallback.getJobs()`.

chavdar: There is no such thing as finishing notifications. New updates can come at any time. There is always a race condition and even the fallback getJobs() may become obsolete immediately.

I think we have two options:
- Either risk lagging behind get fallback by using _cache.getJobs(), 
- Or risk being ahead of the cache by using _fallback.getJobs()

I think the first one (the current option) is safer since it is consistent with the getJobs() behavior which may always lag behind the actual state. The only way to be consistent with the actual state is through registering a listener.

chavdar:
@ibuenros any more comments?

ibuenros:
lgtm

PR #1136, opened 2016-07-23
Fix config store enable/disable logic in HiveDatasetFinder

pcadabam:


pcadabam:
trivial change merging.

PR #1138, opened 2016-07-25
Change Hive Avro to ORC publish to use Gobblin constructs instead of Hive exchange partition query

abti:
Change Hive Avro to ORC publish to use Gobblin constructs instead of Hive exchange partition query: 
- Hive exchange partition query was being used to move partitions from staging to final table
- However, this can fail queries in flight because the current model overwrites data location
- To overcome this, we have changed the publish module to use Gobblin constructs 

New publish works as follows: 
- Staging table partition is created in a directory under Gobblin control (dir name prefix is derived from source partition location string; such as if partitionDir.prefixLocationHint is set to 'hourly,daily' and 'hourly' exists as substring in source partition path, then 'hourly' string is prefixed to dir name of the partition) 
- At the time of publish, first the staging partition directory is moved as ORC final table sub directory (if previously any dir existed, its deleted; however here hourly / daily hint helps)
- Previous partition is dropped if any from the destination table, and new partition is created with new partition directory location
- Old partition directory (if different) is left over for retention to clean up

Also, other minor changes: 
- Each staging partition gets it own staging table 
- Code refactoring


abti:
Trivial changes, discussed offline. Merging.

PR #1139, opened 2016-07-25
Added support to escape the Hive nested field names when derived from destination table as raw string

abti:
Added support to escape the Hive nested field names when derived from destination table as raw string


abti:
Minor change. Self approving.

PR #1143, opened 2016-07-25
Removing MRCompactorAvroKeyDedupJobRunnerTest.java

chavdar:
It does not compile and it is not being run


chavdar:
@ibuenros  can you review?

ibuenros:
Do we have another similar test? If not, could you open an issue/ticket to create a test for this? Otherwise LGTM.

chavdar:
@ibuenros I've reverted the test removal and I've fixed it so it passes now.

chavdar:
@ibuenros any comments?

ibuenros:
LGTM

PR #1153, opened 2016-07-26
Make sure maven central repo is first; bug fixes

chavdar:
* We've been seeing gobblin-core failures on Travis. It seems related to aws-java-sdk-1.7.4 being needed by hadoop-aws . It seems like the artifact was being pulled from jcenter rather than Maven Central and there might be an issue with the former. I've reordered the repositoriy registration so Maven comes first.
* I think I've finally figured out an issue with intermittently failing :gobblin-compaction:test . MRCompactorJobRunnerFilenameRecordCountProviderTest was creating a directory under source code tree which was confusing gradle. Moved the directory to /tmp/
* Added log4j configuration for gobblin-rest-server tests to write test output to /tmp/gobblin-rest-server-test.log . Hopefully that helps with intermittent failures.

@abti can you review ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,18 @@
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.locks.ReentrantLock
 
+// Setup environment
+ext.build_script_dir = "${projectDir.path}/build_script"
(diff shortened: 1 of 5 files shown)


abti:
LGTM

PR #1155, opened 2016-07-26
Changed Hive validation to make it compatible with old Hive version with auth turned on, and Hive query generation compile with new Hive version

abti:
Changed Hive validation to make it compatible with old Hive version with auth turned on, and Hive query generation compile with new Hive version:

- Changed Hive conversion validation job to make use of filesystem instead of resultset to make it work with ancient Hive versions that do not play with delegation tokens refer: HIVE-14015
- Tests to check if Hive queries are being generated with backticks at correct places, to compile with new Hive versions
- Changed Hive query generator to apply backticks as restricted by new Hive versions while maintaining backward compatibility


abti:
Minor changes, self approving

PR #1156, opened 2016-07-26
Upgrade wix-embedded-mysql 

chavdar:
Checking if some of the intermittent mysql failures are addressed


chavdar:
Self merging to address failing builds.

PR #1161, opened 2016-07-27
Made Hive schema evolution commands compatible with older Hive versions

abti:
Made Hive schema evolution commands compatible with older Hive versions


abti:
Trivial fix, self approving.

PR #1168, opened 2016-07-29
Initial commit for a skeleton GobblinInstanceDriver

chavdar:
Changes:
- A skeleton GobblinInstanceDriver implementation -- DefaultGobblinInstanceDriverImpl
- ImmediateJobSpecScheduler: Simple JobSpecScheduler implementation that puts the tasks immediately in a separate thread.
- Some default listener implementations
- Fix FindBugs in new launcher APIs
- Unit tests

@ibuenros @autumnust Can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
@@ -11,18 +11,28 @@
  */
 package gobblin.runtime.api;
 
+import java.util.HashMap;
 import java.util.Map;
 
+import lombok.Getter;
+
 /**
- * @author cbotev
+ * TODO
  *
  */
 public class JobExecutionState extends JobExecutionStatus {
-  final JobSpec jobSpec;
+  @Getter final JobSpec jobSpec;
   Map<String, Object> executionMetadata;
 
   public JobExecutionState(JobSpec jobSpec, JobExecution jobExecution) {
     super(jobExecution);
     this.jobSpec = jobSpec;
+    // TODO default implementation
+    this.executionMetadata = new HashMap<>();
   }
+
+  public Map<String, Object> getExecutionMetadata() {
+    return this.executionMetadata;
+  }
+
 }
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
@@ -11,6 +11,7 @@
  */
 package gobblin.runtime.api;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -32,7 +33,7 @@ public class JobExecutionState extends JobExecutionStatus {
   }
 
   public Map<String, Object> getExecutionMetadata() {
-    return this.executionMetadata;
+    return Collections.unmodifiableMap(this.executionMetadata);
   }
 
 }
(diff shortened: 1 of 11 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
```diff
@@ -11,18 +11,28 @@
...
 
   public JobExecutionState(JobSpec jobSpec, JobExecution jobExecution) {
     super(jobExecution);
     this.jobSpec = jobSpec;
+    // TODO default implementation
+    this.executionMetadata = new HashMap<>();
+  }
+
+  public Map<String, Object> getExecutionMetadata() {
+    return this.executionMetadata;
```
ibuenros: Return an immutable copy?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionStatus.java
```diff
@@ -11,23 +11,25 @@
...
-
 import gobblin.runtime.JobState;
 
+import lombok.Getter;
+
 public class JobExecutionStatus {
-  final JobExecution jobExecution;
+  public static final String UKNOWN_STAGE = "unkown";
+
+  @Getter final JobExecution jobExecution;
```
ibuenros: You can put `@Getter` at the class level instead.

gobblin-runtime/src/main/java/gobblin/runtime/instance_driver/DefaultGobblinInstanceDriverImpl.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Override
+    public void run() {
+       _jobLauncher.launchJob(_jobSpec, new JobExecutionListener());
+    }
+  }
+
+  /** Listens to changes in the Job catalog and schedules/un-schedules jobs. */
+  class JobSpecListener extends DefaultJobCatalogListenerImpl {
+
```
autumnust: A question: Why do we prefer put the `JobSpecListener` which extends `DefaultJobCatalogListenerImpl ` as inner class inside `DefaultGobblinInstanceDriverImpl` instead of pass the object as the parameter in?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
chavdar: Yeah, I was thinking about this but we also need way to modify the metadata.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionStatus.java
chavdar: Fixed.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
ibuenros: I think the `JobExecutionObject` should proxy those methods. That way it can also notify the listeners of changes in metadata.

gobblin-runtime/src/main/java/gobblin/runtime/instance_driver/DefaultGobblinInstanceDriverImpl.java
chavdar: DefaultJobCatalogListenerImpl is meant to provide token implementation for all callbacks. Other classes can extend from it and override only the methods they care about. If we use delegation pattern, we'll have to always implement all the methods.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
chavdar: Ok. I'll make it read only.

ibuenros:
LGTM

PR #1169, opened 2016-07-29
Gobblin metrics registers the base schemas instead of inferring them …

ibuenros:
…from events.


ibuenros:
@pcadabam can you review?

pcadabam:
LGTM

PR #1172, opened 2016-08-02
Use camel case for event names

pcadabam:
Trivial change self merging.


pcadabam:
Trivial change self merging.

PR #1174, opened 2016-08-03
Base implementation for Kafka job spec monitor.

ibuenros:


ibuenros:
@chavdar @autumnust can you review?

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -54,6 +54,17 @@ public static Properties configToProperties(Config config) {
   }
 
   /**
+   * @return the subconfig under key "key" if it exists, otherwise returns an empty config.
+   */
+  public static Config getConfigOrEmpty(Config config, String key) {
+    try {
+      return config.getConfig(key);
```
chavdar: you avoid exception handling by using config.hasPath()

gobblin-runtime/src/main/java/gobblin/runtime/kafka/HighLevelConsumer.java
```diff
@@ -0,0 +1,126 @@
...
+    this.consumer = createConsumerConnector(config);
+    this.topic = topic;
+    this.numThreads = numThreads;
+  }
+
+  protected abstract void processMessage(MessageAndMetadata<K, V> message);
+
+  @Override
+  protected void startUp()
+      throws Exception {
```
chavdar: Please get rid of "throws Exception"

gobblin-runtime/src/main/java/gobblin/runtime/kafka/HighLevelConsumer.java
```diff
@@ -0,0 +1,126 @@
...
+  private final int numThreads;
+  private ExecutorService executor;
+
+  public HighLevelConsumer(String topic, Config config, int numThreads) {
+    this.consumer = createConsumerConnector(config);
+    this.topic = topic;
+    this.numThreads = numThreads;
+  }
+
+  protected abstract void processMessage(MessageAndMetadata<K, V> message);
```
chavdar: Please document that the implementation of this method should be thread-safe.

gobblin-runtime/src/main/java/gobblin/runtime/kafka/HighLevelConsumer.java
```diff
@@ -0,0 +1,126 @@
...
+import kafka.consumer.ConsumerIterator;
+import kafka.consumer.KafkaStream;
+import kafka.javaapi.consumer.ConsumerConnector;
+import kafka.message.MessageAndMetadata;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * A high level consumer for Kafka topics. Subclasses should implement {@link HighLevelConsumer#processMessage(MessageAndMetadata)}
```
chavdar: Given that processing of messages is serialized it may be useful to document that this consumer should not be used for high-throughput cases.

gobblin-runtime/src/main/java/gobblin/runtime/job_monitor/KafkaJobMonitor.java
```diff
@@ -0,0 +1,73 @@
...
+  public static final String KAFKA_JOB_MONITOR_PREFIX = "jobSpecMonitor.kafka";
+
+  private final MutableJobCatalog jobCatalog;
+
+  /**
+   * @return A collection of either {@link JobSpec}s to add/update or {@link URI}s to remove from the catalog,
+   *        parsed from the Kafka message.
+   * @throws IOException
+   */
+  public abstract Collection<Either<JobSpec, URI>> parseJobSpec(byte[] message) throws IOException;
```
chavdar: Do you envision a use-case where multiple jobs are in the same message?

gobblin-runtime/src/main/java/gobblin/runtime/kafka/HighLevelConsumer.java
ibuenros: The processing is not serialized. You can have as many threads as partitions there are. I think this is usual for Kafka consumers.

gobblin-runtime/src/main/java/gobblin/runtime/job_monitor/KafkaJobMonitor.java
ibuenros: We could run multiple jobs from a trigger. For example, when receiving a data availability event we may want to launch copies to multiple clusters, as well as a format transformation, etc.

gobblin-runtime/src/main/java/gobblin/runtime/kafka/HighLevelConsumer.java
chavdar: OK but it depends on how processMessage() is made thread-safe.

gobblin-runtime/src/main/java/gobblin/runtime/job_monitor/KafkaJobMonitor.java
chavdar: I see.

ibuenros:
Addressed comments.

gobblin-runtime/src/test/java/gobblin/runtime/TestUtils.java
```diff
@@ -0,0 +1,50 @@
...
+package gobblin.runtime;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+import lombok.extern.slf4j.Slf4j;
+
+
+@Slf4j
+public class TestUtils {
```
chavdar: Have a look at gobblin.testing.AssertWithBackoff

chavdar:
LGTM. One small comment about reusing gobblin.testing.AssertWithBackoff .

PR #1175, opened 2016-08-03
Schema files (schema.avsc) will be written with 774 permission.

ibuenros:
@chavdar very quick one, can you review? Thanks!


pcadabam:
LGTM

PR #1177, opened 2016-08-03
IllegalArgumentException in JdbcPublisher when staging tables are skipped

lbendig:
When publishing to an RDBMS and staging tables are skipped 
(`job.commit.policy=partial` and `publish.data.at.job.level=false`) the following exception is thrown in the tasks:

```
ERROR [TaskExecutor-0] gobblin.runtime.TaskExecutor: Thread Thread[TaskExecutor-0,5,main]
 threw an uncaught exception: java.lang.IllegalArgumentException: Cannot instantiate
    JdbcPublisher since it does not extend SingleTaskDataPublisher
      at com.google.common.base.Preconditions.checkArgument(Preconditions.java:93)
      at gobblin.publisher.SingleTaskDataPublisher.getInstance(SingleTaskDataPublisher.java:67)
      at gobblin.runtime.Task.publishTaskData(Task.java:313)
      at gobblin.runtime.Task.run(Task.java:251)
```
The problem is that `JdbcPublisher extends DataPublisher` however, according to PR https://github.com/linkedin/gobblin/pull/360 task level publishers should extend `SingleTaskDataPublisher`.
With this patch I'd propose to change to `JdbcPublisher extends SingleTaskDataPublisher` and instead of throwing an unchecked exception in its constructor, let just log that the commit has already been done in the tasks.

@zliu41 , @jinhyukchang what do you think?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/JdbcPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/JdbcPublisher.java
@@ -44,7 +44,7 @@ import gobblin.writer.commands.JdbcWriterCommandsFactory;
 /**
  * Publishes data into JDBC RDBMS. Expects all the data has been already in staging table.
  */
-public class JdbcPublisher extends DataPublisher {
+public class JdbcPublisher extends SingleTaskDataPublisher {
   public static final String JDBC_PUBLISHER_PREFIX = "jdbc.publisher.";
   public static final String JDBC_PUBLISHER_DATABASE_NAME = JDBC_PUBLISHER_PREFIX + "database_name";
   public static final String JDBC_PUBLISHER_FINAL_TABLE_NAME = JDBC_PUBLISHER_PREFIX + "table_name";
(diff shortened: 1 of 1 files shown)


jinhyukchang:
@lbendig Seems like there has been some change regarding instantiating publisher as JDBCPublisher is not intended to be instantiated when staging table is skipped -- JdbcPublisher has validate(state) method to check the condition and will throw exception on this condition.
I didn't have time to look at what kind of change brought this regression, but do you know what change brought to instantiate it on this condition?

lbendig:
@jinhyukchang I think changes in #360 and #684.

chavdar:
@ibuenros can you review ?

gobblin-core/src/main/java/gobblin/publisher/JdbcPublisher.java
```diff
@@ -199,5 +179,28 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+    // Nothing to do
+  }
+  
+  @Override
+  public void publishMetadata(WorkUnitState state) throws IOException {
+    // Nothing to do
+  }
+  
+  @Override
+  public void publishData(WorkUnitState state) throws IOException {
```
ibuenros: This method is only called from the tasks. If I understand correctly there are two options:

1. JDBC writer wrote directly to main table. This (for now) is equivalent to `commitPolicy!=COMMIT_ON_FULL_SUCCESS`. In this case, the message should probably simply be something like `Records written directly to table. No publish necessary.`. However, there's the problem of guaranteeing that that was the case, which right now relies on a single key, and a developer may mistakenly change that in the future in the writer but not in the publisher. Instead, I see two options:
1.1. Have a static method in the writer that determines whether it writes directly to main table. Use that method from both writer and publisher.
1.2. (Preferable) Write whether the data is already published directly on the `WorkUnitState`. An existing way of doing this is using the `FinalState` functionality (see for example `FsDataWriter`, specifically the `getFinalState()` method.

2. JDBC writer used a staging table. Currently, this guarantees the commit to be done at the job level, and this method should never be called. However, this may not be the case in the future (again, developer might unknowingly modify the writer). If this method should not have been called, then the method should throw an error and fail the task. If it should have been called, then it should do the appropriate publishing work.

abti:
https://issues.apache.org/jira/browse/GOBBLIN-90

Please update your PR title with following prefix: [GOBBLIN-90]

PR #1181, opened 2016-08-04
TeradataWriter implementation

lbendig:
With this PR I'd like to add the capability of writing to Teradata.
- Writing is done through batch inserts 
- The final table should be created beforehand which is ideally a _MULTISET_ table
  _without a primary index_.
- Writer specific jars (**tdgssconfig** and **terajdbc4**) need to be passed to the job 

Excerpt from a job definition:

> \# Converter
> converter.classes=gobblin.converter.jdbc.AvroToJdbcEntryConverter
> 
> \# Writer
> writer.destination.type=TERADATA
> writer.builder.class=gobblin.writer.JdbcWriterBuilder
> 
> \# Publish
> data.publisher.job.type=gobblin.publisher.JdbcPublisher
> jdbc.publisher.url=jdbc:teradata://host
> jdbc.publisher.driver=com.teradata.jdbc.TeraDriver

Tests:
Export 30M records with and without staging tables to Teradata 15.10
both from HDFS and Kafka sources.

---
Additional fix: class `resourcesBasedTemplate` was renamed to `ResourcesBasedTemplate` which caused Findbug checks to fail

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
@@ -12,33 +12,23 @@
 
 package gobblin.writer.commands;
 
-import gobblin.configuration.State;
-import gobblin.converter.jdbc.JdbcEntryData;
-import gobblin.converter.jdbc.JdbcEntryDatum;
-
 import java.sql.BatchUpdateException;
 import java.sql.Connection;
 import java.sql.ParameterMetaData;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
-
-import lombok.ToString;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.github.rholder.retry.Retryer;
-import com.github.rholder.retry.RetryerBuilder;
-import com.github.rholder.retry.StopStrategies;
-import com.github.rholder.retry.WaitStrategies;
-import com.google.common.base.Joiner;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
+import gobblin.configuration.State;
+import gobblin.converter.jdbc.JdbcEntryData;
+import gobblin.converter.jdbc.JdbcEntryDatum;
+import lombok.ToString;
 
 
 /**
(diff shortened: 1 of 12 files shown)


chavdar:
@pcadabam Can you review?

pcadabam:
@lbendig I started looking into this one

jinhyukchang:
Hi @lbendig,
Does TeradataWriter work with TYPE=FASTLOAD ?

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
```diff
@@ -0,0 +1,201 @@
...
+  @Override
+  public void insert(String databaseName, String table, JdbcEntryData jdbcEntryData) throws SQLException {
+    if (this.columnNames == null) {
+      initializeForBatch(this.conn, databaseName, table, jdbcEntryData);
+    }
+    this.pendingInserts.add(jdbcEntryData);
+
+    if (this.pendingInserts.size() == this.batchSize) {
+      insertBatch(this.insertPstmtForFixedBatch); //reuse pre-computed Preparedstatement.
+      return;
```
pcadabam: Not required

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
```diff
@@ -0,0 +1,201 @@
...
+          pstmt.clearParameters();
+        }
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("Executing SQL " + pstmt);
+        }
+        int[] execStatus = pstmt.executeBatch();
+        // Check status explicitly if driver continues batch insertion upon failure 
+        for (int status : execStatus) {
+          if (status == Statement.EXECUTE_FAILED) {
+            throw new BatchUpdateException("Batch insert failed.", execStatus);
```
pcadabam: If any of the statements failed in the batch we throw an exception? The ones that succeeded will be retried by the retryer. Are the inserts idempotent?

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
```diff
@@ -0,0 +1,201 @@
...
+ *
+ */
+@ToString
+public class TeradataBufferedInserter implements JdbcBufferedInserter {
+  private static final Logger LOG = LoggerFactory.getLogger(TeradataBufferedInserter.class);
+
+  private static final String INSERT_STATEMENT_PREFIX_FORMAT = "INSERT INTO %s.%s (%s) VALUES ";
+  private static final Joiner JOINER_ON_COMMA = Joiner.on(',');
+
+  private List<JdbcEntryData> pendingInserts;
```
pcadabam: `final`

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
```diff
@@ -0,0 +1,201 @@
...
+    this.pendingInserts = Lists.newArrayList();
+
+    this.insertStmtPrefix =
+        String.format(INSERT_STATEMENT_PREFIX_FORMAT, databaseName, table, JOINER_ON_COMMA.join(this.columnNames));
+    this.insertPstmtForFixedBatch = conn.prepareStatement(createPrepareStatementStr(this.insertStmtPrefix));
+    this.columnPosSqlTypes = getColumnPosSqlTypes();
+    LOG.info(String.format("Initialized for %s insert " + this, (this.batchSize > 1) ? "batch" : ""));
+
+    //retry after 2, 4, 8, 16... sec, max 30 sec delay
+    this.retryer = RetryerBuilder.<Boolean> newBuilder().retryIfException()
```
pcadabam: I think it makes sense to make this configurable with defaults

pcadabam: Can this be initialized in the constructor?

pcadabam:
@lbendig 
Left some comments please take a look.

lbendig:
@jinhyukchang  I haven't added the fastload option because due to some its limitations, like:
- it can only insert to an empty table
- the parallell writes shouldn't exceed the number of the AMPs
- there are restrictions on then lenght of naming the columns/tables
- not all the jdbc types are supported, like clobs, blobs
but later I think it would be good to have it, I'll do some tests.

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: Indeed. Will remove the return, also from MySqlBufferedInserter.

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: If an exception happens when the execute is called on the preparedStatement then AFAIK this batch will be rolled back. However, it's up to the JDBC driver implementation whether to throw and exception if any of the inserts failed or just move to the next insert. I just explicitly check the return codes after one particular batch and throw an exception if an error code has been stored for a given record. If the driver just silently moves to the next record without throwing an exception then I think we can't be sure whether all the inserts were successful.

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: You mean to make pendingInserts final? It is currently initialized in TeradataBufferedInserter#initializeForBatch (similarly as in MySqlBufferedInserter)
but only once when columnNames == null.

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: Having a parameter for maxattempt and maxwait? I'll then adapt `MySqlBufferedInserter` as well.

lbendig:
@pcadabam Thanks for the comments. I made my comments inline.

jinhyukchang:
@lbendig Yes, I wondered if fastload requires any additional code change or not -- as long as conditions meet. Thanks!

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
pcadabam: I meant in general for all fields where possible.

private List<String> columnNames;
private Map<Integer, Integer> columnPosSqlTypes;
private String insertStmtPrefix;
private PreparedStatement insertPstmtForFixedBatch;
private Retryer<Boolean> retryer;

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
pcadabam: Ok

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: For this I restructured the *`BufferedInserter` classes. Because some of the methods were just copy pasted from `MySQLBufferedInserter` to the Teradata one, I refactored the
code a bit and now these are in an abstract `BaseJdbcBufferedInserter` class.

gobblin-core/src/main/java/gobblin/writer/commands/TeradataBufferedInserter.java
lbendig: Added new config parameters: 
`writer.jdbc.insert_retry_timeout` (default: 30 sec)
`writer.jdbc.insert_retry_max_attempt `(default: 5)

ydai1124:
@lbendig @pcadabam Have all the comments been addressed?

lbendig:
@ydai1124 Yes, I addressed @pcadabam's comments. I'll open a PR for the docs.

ydai1124:
@lbendig As @chavdar mentioned in #1223 , your change will be included in our next release. We will have more frequent release in the future. Thank you for contributing :)

pcadabam:
LGTM
@lbendig Please fix the merge conflicts and I will merge this.

lbendig:
Hi @pcadabam, conflicts are fixed.

lbendig:
@pcadabam, @ydai1124 thanks for the review!

PR #1183, opened 2016-08-05
change the key name of required.attributes

autumnust:
change the key name of required.attributes to  gobblin.template.required_attributes


ibuenros:
LGTM, but this will conflict with a fix I had to make regarding the name of `ResourcesBasedTemplate`. Please fix that and let me know before committing.

autumnust:
@ibuenros  fixed.

PR #1187, opened 2016-08-08
Changed copy entity metadata to Map<String,String> because otherwise …

ibuenros:
…GSON cannot properly deserialize.


ibuenros:
@ydai1124 can you review?

ydai1124:
LGTM

PR #1191, opened 2016-08-10
Rename init method in BackfillHiveSource

pcadabam:
Simple rename of ```initialize``` method to ```initBackfillHiveSource``` since the name conflicts with method in super class


pcadabam:
Self merging trivial refactor.

PR #1196, opened 2016-08-11
Configurable partitioning in DatePartitionedDailyAvroSource

lbendig:
Currently the `DatePartitionedDailyAvroSource` can only read _daily_ partitioned data.
This PR extends this functionality by introducing `DatePartitionedAvroFileSource` in which the partitioning is configurable by the properties as follows:

> date.partitioned.source.partition.prefix
> date.partitioned.source.partition.suffix
> date.partitioned.source.partition.pattern
> date.partitioned.source.partition.granularity

In this way an arbitrarily partitioned data (e.g written out by `TimeBasedWriterPartitioner`) can be read with the same temporal granularity configuration.


lbendig:
@zliu41 or @sahilTakiar , could you please take a look?

PR #1197, opened 2016-08-11
Kafka-template

autumnust:
- simple kafka template 
- file detector running period (30s) added into gobblin/conf/gobblin-standalone.properties


autumnust:
@ibuenros Could you review this template?

gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
```diff
@@ -186,6 +186,10 @@ private void rescheduleJob(Properties jobProps)
     boolean runOnce = Boolean.valueOf(jobProps.getProperty(ConfigurationKeys.JOB_RUN_ONCE_KEY, "false"));
     // Reschedule the job with the new job configuration
     jobScheduler.scheduleJob(jobProps, runOnce ? new RunOnceJobListener() : new EmailNotificationJobListener());
+    LOG.info("[JobScheduler]"
```
ibuenros: This looks like a debug message. If so, please use `LOG.debug`, also, why is it in multiple lines?

ibuenros:
Just one comment, LGTM otherwise.

gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
autumnust: I guess IntelliJ's formatter doesn't like me :(

PR #1198, opened 2016-08-11
Removing originalJob from JobCatalogListener.onUpdatedJob(); Adding J…

chavdar:
* Removing originalJob from JobCatalogListener.onUpdatedJob() -- this touches a lot of files
* Adding JobLifecycleListener interface and related classes
  - JobLifecycleListenersContainer interface
  - Extending GobblinInstanceDriver with the JobLifecycleListenersContainer interface
  - DefaultJobLifecycleListenerImpl - default no-op implementation
  - JobLifecycleListenersList - default JobLifecycleListenersContainer implementation

@ibuenros @autumnust can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/GobblinInstanceDriver.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/GobblinInstanceDriver.java
@@ -22,8 +22,14 @@ import gobblin.annotation.Alpha;
  * {@link JobSpec}s and run them.
  * */
 @Alpha
-public interface GobblinInstanceDriver extends Service {
+public interface GobblinInstanceDriver extends Service, JobLifecycleListenersContainer {
   JobCatalog getJobCatalog();
+  /**
+   * Returns a mutable instance of the job catalog
+   * (if it implements the {@link MutableJobCatalog} interface). Implementation will throw
+   * ClassCastException if the current catalog is not mutable.
+   */
+  MutableJobCatalog getMutableJobCatalog();
   JobSpecScheduler getJobScheduler();
   JobExecutionLauncher getJobLauncher();
   Logger getLog();
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalogListener.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalogListener.java
@@ -33,7 +33,7 @@ public interface JobCatalogListener {
   /**
    * Invoked when the contents of a JobSpec gets updated in the catalog.
    */
-  public void onUpdateJob(JobSpec originalJob, JobSpec updatedJob);
+  public void onUpdateJob(JobSpec updatedJob);
 
   /** A standard implementation of onAddJob as a functional object */
   public static class AddJobCallback extends Callback<JobCatalogListener, Void> {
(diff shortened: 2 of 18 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/api/GobblinInstanceDriver.java
```diff
@@ -22,8 +22,14 @@
...
 @Alpha
-public interface GobblinInstanceDriver extends Service {
+public interface GobblinInstanceDriver extends Service, JobLifecycleListenersContainer {
   JobCatalog getJobCatalog();
+  /**
+   * Returns a mutable instance of the job catalog
+   * (if it implements the {@link MutableJobCatalog} interface). Implementation will throw
+   * ClassCastException if the current catalog is not mutable.
+   */
+  MutableJobCatalog getMutableJobCatalog();
```
ibuenros: Why not return `Optional`?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalogListener.java
```diff
@@ -33,7 +33,7 @@
   /**
    * Invoked when the contents of a JobSpec gets updated in the catalog.
    */
-  public void onUpdateJob(JobSpec originalJob, JobSpec updatedJob);
+  public void onUpdateJob(JobSpec updatedJob);
```
ibuenros: Also modify `onDeleteJob` to only take the `URI`

gobblin-runtime/src/main/java/gobblin/runtime/api/GobblinInstanceDriver.java
chavdar: The catalog is not optional. If someone is trying to use an Immutable catalog as mutable, that's an error.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalogListener.java
chavdar: Will do.

gobblin-runtime/src/main/java/gobblin/runtime/api/GobblinInstanceDriver.java
ibuenros: What is the use case for this method?

chavdar: Manually submitting a job: driver.getMutableJobCatalog().put(JobSpec).

ibuenros: Can you at least declare the `ClassCastException` so that it is more obvious?

chavdar: ClassCastException is a RuntimeException. It denotes an incorrect usage of the method, not a problem with the business logic.  Typically, unchecked exceptions are not declared because the calling code is not expected to recover from them. In this particular case, someone has instantiated the instance with a read-only catalog and they are trying to write to the catalog. The calling code cannot recover from this.

See also https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

ibuenros:
LGTM

PR #1201, opened 2016-08-12
Fix SchedulerUtils bugs for loading config file on hdfs

autumnust:
bugs appeared when testing on Azkaban. (try-with-resources one and loading a single config file one) Could you help for the review? @ibuenros   

--- 
Bugs detail: 
- In `SchedulerUtils.java: loadGenericJobConfig `  method, the loading of single configuration file process is buggy, as it hard coded  "file://" prefix which is not compatible for file system except local fs. The solution is to change it to use `PropertiesConfiguration` object for loading a `inputStream` instead. 

- The instance of FileSystem will be closed with try-with-resources implementation, which is not expected for periodically running scenario. May need to redesign this later.


chavdar:
For bookkeeping purposes, please provide a description of the bug and how you are fixing it. Any way, this can be unit tested?

Further, the code does not compile.

PR #1202, opened 2016-08-12
Bug fix - Retention does not blacklist dataset

pcadabam:
## Root cause ##
```ConfigurableCleanableDataset``` was calling the incorrect super constructor ```super(fs, jobProps, log)``` instead of ```super(fs, jobProps, config, log)```

## Impact ##
When retention job is configured through config store, each dataset can be blacklisted by setting ```gobblin.retention.is.blacklisted=true```. Since the wrong constructor was being called, this value was never passed to the super class. Hence the dataset was not blacklisted.

@chavdar can you review?


chavdar:
LGTM.

PR #1203, opened 2016-08-12
New std driver2

chavdar:
A lot of files touched with mostly mechanical changes.

Summary of changes:
- Adding weak listeners which will get automatically GCed
- Adding JobSpecSchedulerListener and related classes to listen to scheduling events
- Extending the JobLifecycleListener with JobSpecSchedulerListener and onJobLaunch(JobExecutionDriver jobDriver)
- Adding JobSpecFilter and FilteredJobLifecycleListener 

@ibuenros @autumnust  Can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
@@ -14,6 +14,11 @@ package gobblin.runtime.api;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
@@ -33,11 +38,15 @@ public class JobExecutionState implements JobExecutionStatus {
 
   @Getter private final JobExecution jobExecution;
   private final Optional<JobExecutionStateListener> listener;
-  @Getter private JobState.RunningState runningState;
-  /** Arbitrary execution stage, e.g. setup, workUnitGeneration, taskExecution, publishing */
-  @Getter private String stage;
   @Getter final JobSpec jobSpec;
-  Map<String, Object> executionMetadata;
+  // We use a lock instead of synchronized so that we can add different conditional variables if
+  // needed.
+  private final Lock changeLock = new ReentrantLock();
+  private final Condition runningStateChanged = changeLock.newCondition();
+  private JobState.RunningState runningState;
+  /** Arbitrary execution stage, e.g. setup, workUnitGeneration, taskExecution, publishing */
+  private String stage;
+  private Map<String, Object> executionMetadata;
 
   public JobExecutionState(JobSpec jobSpec, JobExecution jobExecution,
                            Optional<JobExecutionStateListener> listener) {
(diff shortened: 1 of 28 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
```diff
@@ -51,18 +60,51 @@ public JobExecutionState(JobSpec jobSpec, JobExecution jobExecution,
...
+    }
+    finally {
+      this.changeLock.unlock();
+    }
+  }
+
+  @Override public JobState.RunningState getRunningState() {
+    this.changeLock.lock();
+    try {
+      return this.runningState;
```
ibuenros: Do you need a lock for this method?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
```diff
@@ -76,63 +118,165 @@ public void setRunningState(JobState.RunningState runningState) {
...
+          || RunningState.SUCCESSFUL == this.runningState,
+          "unexpected state:" + this.runningState);
+      doRunningStateChange(RunningState.CANCELLED);
+    }
+    finally {
+      this.changeLock.unlock();
+    }
   }
 
+  // This must be called only when holding changeLock
```
ibuenros: You could re-acquire the lock, given that it is reentrant.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
```diff
@@ -76,63 +118,165 @@ public void setRunningState(JobState.RunningState runningState) {
...
+      if (null == this.runningState || !this.runningState.isDone()) {
+        throw new TimeoutException("Timeout waiting for done.");
+      }
+    }
+    finally {
+      this.changeLock.unlock();
+    }
+  }
+
+  public void awaitForState(RunningState targetState, long timeoutMs)
```
ibuenros: Could you refactor a bit to avoid repeating the code from `awaitForDone`?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpecScheduler.java
```diff
@@ -37,6 +37,14 @@
   public JobSpecSchedule scheduleJob(JobSpec jobSpec, Runnable jobRunnable);
 
   /**
+   * Add a Gobblin job for scheduling. Job is guaranteed to run only once.
```
ibuenros: Can you add `regardless of job outcome` or similar?

gobblin-utility/src/main/java/gobblin/util/callbacks/CallbacksDispatcher.java
```diff
@@ -43,6 +45,7 @@
 public class CallbacksDispatcher<L> {
   private final Logger _log;
   private final List<L> _listeners = new ArrayList<>();
+  private final List<WeakReference<L>> _autoListeners = new ArrayList<>();
```
ibuenros: I would prefer a `WeakHashMap` here, which manages itself and you don't have to create weak references.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
chavdar: Yes, both read and write paths need to be protected for consistency. We can try to avoid using lock with "transient" but in some perf experiments I did once, I didn't find it much faster.

Better safe than sorry :-)

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
chavdar: Sure.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobSpecScheduler.java
chavdar: Done.

gobblin-utility/src/main/java/gobblin/util/callbacks/CallbacksDispatcher.java
chavdar: Why a hash map? What are the values of the map? I think managing the weakreferences is trivial.

ibuenros: It is essentially a set (e.g. values are dummies). The set is useful because listeners might be removed often, which is faster in a set. I just want to avoid reinventing the wheel and possibly making mistakes.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionState.java
chavdar: Fixed. I added a common implementation which works on a Predicate<JobExecutionState>.

gobblin-utility/src/main/java/gobblin/util/callbacks/CallbacksDispatcher.java
chavdar: Changed.

ibuenros:
LGTM

PR #1205, opened 2016-08-16
AvroToorc - Implemented a per partition watermark

pcadabam:
### Changes ###
For every Table we create a NoOp watermark WorkUnit that stores watermarks for all its Partitions as a ```Map```. This workunit is identified by the property ```IS_WATERMARK_WORKUNIT_KEY``` being set to ```true```.
This workunit is intended to only carry watermark state and does not contain any business logic to execute hive conversion queries.

The watermark is stored as a ```MultiKeyValueLongWatermark``` which is an extension to ```Map``` that implements gobblin's ```Watermark``` interface. The key is a Partition identifier and value is the watermark for this Partition

The maximum number of Partition watermarks for each Table can be adjusted by the property MAX_PARTITIONS_IN_WATERMARK_PER_DATASET_KEY, The default is 50. This means we only track watermarks for 50 most recently modified partitions of each table. If a previous watermark is not found for as partition, it returns 0 as the watermark

A ```TableLevelWatermarer``` is still supported. It can be used instead of a ```PartitionLevelWatermarker```

Note that the publish logic is still at a per ```Table``` level. We still fail all partitions if any partition fails. Per partition commit is doable but is an optimization we can work on later.

@chavdar and @abti  can you review?

Internal jira ticket - ETL-4785

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
@@ -91,7 +91,7 @@ public abstract class AbstractAvroToOrcConverter extends Converter<Schema, Schem
   /**
    * list of partitions that a partition has replaced. E.g. list of hourly partitons for a daily partition
    */
-  private static final String REPLACED_PARTITIONS_HIVE_METASTORE_KEY = "gobblin.replaced.partitions";
+  public static final String REPLACED_PARTITIONS_HIVE_METASTORE_KEY = "gobblin.replaced.partitions";
 
   /**
    * The dataset being converted.
@@ -459,21 +459,28 @@ public abstract class AbstractAvroToOrcConverter extends Converter<Schema, Schem
     return orcDataLocation + Path.SEPARATOR + orcStagingTableName;
   }
 
-  /**
-   * Parse the {@link #REPLACED_PARTITIONS_HIVE_METASTORE_KEY} from partition parameters to returns DDLs for all the partitions to be
-   * dropped.
-   */
+
   @VisibleForTesting
   public static List<Map<String, String>> getDropPartitionsDDLInfo(QueryBasedHiveConversionEntity conversionEntity) {
     if (!conversionEntity.getHivePartition().isPresent()) {
       return Collections.emptyList();
     }
 
-    Table table = conversionEntity.getHiveTable().getTTable();
-    Partition hivePartition = conversionEntity.getHivePartition().get();
+    return getDropPartitionsDDLInfo(conversionEntity.getHivePartition().get());
 
+  }
+
+  /**
+   * Parse the {@link #REPLACED_PARTITIONS_HIVE_METASTORE_KEY} from partition parameters to returns DDLs for all the partitions to be
+   * dropped.
+   *
+   * @return A {@link List} of partitions to be dropped. Each element of the list is a {@link Map} which maps a partition's
+   * key and value.
+   *
+   */
+  public static List<Map<String, String>> getDropPartitionsDDLInfo(Partition hivePartition) {
     List<Map<String, String>> replacedPartitionsDDLInfo = Lists.newArrayList();
-    List<FieldSchema> partitionKeys = table.getPartitionKeys();
+    List<FieldSchema> partitionKeys = hivePartition.getTable().getPartitionKeys();
 
     if (StringUtils.isNotBlank(hivePartition.getParameters().get(REPLACED_PARTITIONS_HIVE_METASTORE_KEY))) {
 
@@ -493,7 +500,6 @@ public abstract class AbstractAvroToOrcConverter extends Converter<Schema, Schem
         }
       }
     }
-
     return replacedPartitionsDDLInfo;
   }
 
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
@@ -14,7 +14,10 @@ package gobblin.data.management.conversion.hive.extractor;
 import java.io.IOException;
 import java.util.List;
 
+import lombok.extern.slf4j.Slf4j;
+
 import org.apache.avro.Schema;
+import org.apache.avro.Schema.Type;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.Partition;
@@ -25,6 +28,7 @@ import org.apache.thrift.TException;
 import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 
+import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
 import gobblin.data.management.conversion.hive.AvroSchemaManager;
 import gobblin.data.management.conversion.hive.dataset.ConvertibleHiveDataset;
@@ -32,6 +36,7 @@ import gobblin.data.management.conversion.hive.entities.QueryBasedHiveConversion
 import gobblin.data.management.conversion.hive.entities.SchemaAwareHivePartition;
 import gobblin.data.management.conversion.hive.entities.SchemaAwareHiveTable;
 import gobblin.data.management.conversion.hive.source.HiveWorkUnit;
+import gobblin.data.management.conversion.hive.watermarker.PartitionLevelWatermarker;
 import gobblin.data.management.copy.hive.HiveDatasetFinder;
 import gobblin.hive.HiveMetastoreClientPool;
 import gobblin.source.extractor.DataRecordException;
@@ -51,11 +56,17 @@ import gobblin.util.AutoReturnableObject;
  * to get the corresponding hive {@link org.apache.hadoop.hive.ql.metadata.Table} and hive {@link org.apache.hadoop.hive.ql.metadata.Partition}
  * </p>
  */
+@Slf4j
 public class HiveConvertExtractor implements Extractor<Schema, QueryBasedHiveConversionEntity> {
 
   private List<QueryBasedHiveConversionEntity> conversionEntities = Lists.newArrayList();
 
   public HiveConvertExtractor(WorkUnitState state, FileSystem fs) throws IOException, TException, HiveException {
+
+    if (Boolean.valueOf(state.getPropAsBoolean(PartitionLevelWatermarker.IS_WATERMARK_WORKUNIT_KEY))) {
+      log.info("Ignoring Watermark workunit for {}", state.getProp(ConfigurationKeys.DATASET_URN_KEY));
+      return;
+    }
 
     HiveWorkUnit hiveWokUnit = new HiveWorkUnit(state.getWorkunit());
 
@@ -94,7 +105,7 @@ public class HiveConvertExtractor implements Extractor<Schema, QueryBasedHiveCon
   @Override
   public Schema getSchema() throws IOException {
     if (this.conversionEntities.isEmpty()) {
-      return null;
+      return Schema.create(Type.NULL);
     }
 
     QueryBasedHiveConversionEntity conversionEntity = this.conversionEntities.get(0);
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
@@ -18,6 +18,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nonnull;
+
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.commons.lang.StringUtils;
@@ -26,10 +28,12 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
 
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Ordering;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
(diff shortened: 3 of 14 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Getter
+  private final Map<String, Long> watermarks;
+
+  public MultiKeyValueLongWatermark() {
+    this.watermarks = Maps.newConcurrentMap();
+  }
+
+  public MultiKeyValueLongWatermark(Map<String, Long> watermarks) {
+    this.watermarks = watermarks;
```
abti: This contructor provides the opportunity to break the contract of being backed by concurrent hashmap

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/TableLevelWatermarker.java
```diff
@@ -73,4 +75,28 @@ public LongWatermark getPreviousHighWatermark(Table table) {
...
+  public void onGetWorkunitsEnd(List<WorkUnit> workunits) {}
+
+  @Override
+  public LongWatermark getExpectedHighWatermark(Table table, long tableProcessTime) {
+    return new LongWatermark(tableProcessTime);
+  }
+
+  @Override
+  public LongWatermark getExpectedHighWatermark(Partition partition, long tableProcessTime, long partitionProcessTime) {
+    return getExpectedHighWatermark(partition.getTable(), tableProcessTime);
```
abti: Not making use of partitionProcessTime?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
```diff
@@ -91,7 +91,7 @@ public String getConfigPrefix() {
   /**
    * list of partitions that a partition has replaced. E.g. list of hourly partitons for a daily partition
    */
-  private static final String REPLACED_PARTITIONS_HIVE_METASTORE_KEY = "gobblin.replaced.partitions";
+  public static final String REPLACED_PARTITIONS_HIVE_METASTORE_KEY = "gobblin.replaced.partitions";
```
chavdar: It's probably late for this change but dots.are.not.substitutes.for.spaces.

chavdar:
> The maximum number of Partition watermarks for each Table can be adjusted by the property MAX_PARTITIONS_IN_WATERMARK_PER_DATASET_KEY, The default is 50.

Can this value be derived from other properties? If not, please add a description of how one is to determine the correct value. What is so magical about "50"?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
```diff
@@ -51,12 +57,18 @@
...
  */
+@Slf4j
 public class HiveConvertExtractor implements Extractor<Schema, QueryBasedHiveConversionEntity> {
 
   private List<QueryBasedHiveConversionEntity> conversionEntities = Lists.newArrayList();
 
   public HiveConvertExtractor(WorkUnitState state, FileSystem fs) throws IOException, TException, HiveException {
 
+    if (Boolean.valueOf(state.getPropAsBoolean(PartitionLevelWatermarker.IS_WATERMARK_WORKUNIT_KEY))) {
+      log.info(String.format("Ignoring Watermark workunit for %s", state.getProp(ConfigurationKeys.DATASET_URN_KEY)));
```
chavdar: nit: log.info("Ignoring Watermark workunit for %s", state.getProp(ConfigurationKeys.DATASET_URN_KEY)) achieves the same.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
```diff
@@ -83,6 +91,10 @@ public HiveConvertPublisher(State state) throws IOException {
     } catch (SQLException e) {
       throw new RuntimeException(e);
     }
+
+    this.watermarker =
```
chavdar: I think we should encode the fact that you pass a state to instantiate a the watermark class in an interface

interface HiveSourceWatermarkerFactory {
   HiveSourceWatermarker createFromState(State state);
}

What we configure is the factory class which gets created using the default constructor.

I think the above approach documents more explicitly the interface and we can also add different factory methods in the future.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A {@link Watermark} that holds multiple key value watermarks. The key is a {@link String} and value is a {@link Long}.
+ * It is backed by a {@link ConcurrentHashMap}.
+ */
+public class MultiKeyValueLongWatermark implements Watermark {
+
+  private static final Gson GSON = new Gson();
+
+  @Getter
```
chavdar: Are users allowed to modify the map directly? If not, they shouldn't have direct access to the map.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Getter;
+
+import com.google.common.collect.Maps;
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+
+import gobblin.source.extractor.Watermark;
+
+/**
+ * A {@link Watermark} that holds multiple key value watermarks. The key is a {@link String} and value is a {@link Long}.
```
chavdar: > The key is a {@link String} and value is a {@link Long}.

I think this is pretty obvious from the interface. I'd remove this comment as it has little informational content. Are there any restrictions/expectations about the keys and/or values that need to be called out? What is the semantics of the key? E.g. are the watermarks expected to be non-negative? Can they be null?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
```diff
@@ -0,0 +1,54 @@
...
+    this.watermarks = watermarks;
+  }
+
+  @Override
+  public JsonElement toJson() {
+    return GSON.toJsonTree(this);
+  }
+
+  @Override
+  public short calculatePercentCompletion(Watermark lowWatermark, Watermark highWatermark) {
```
chavdar: IS this going to be fixed later? Is there an expectation that it needs to be overridden?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
```diff
@@ -0,0 +1,339 @@
...
+  public static final String IS_WATERMARK_WORKUNIT_KEY = "hive.source.watermark.isWatermarkWorkUnit";
+  public static final String MAX_PARTITIONS_IN_WATERMARK_PER_DATASET_KEY = "hive.source.watermark.maxPartitions";
+  public static final int DEFAULT_MAX_PARTITIONS_IN_WATERMARK_PER_DATASET = 50;
+
+  private static final Joiner PARTITION_VALUES_JOINER = Joiner.on(",");
+
+  /**
+   * Ordering used to retain only the {@link #maxPartitionsPerDataset} allowed
+   */
+  private static final Ordering<Map.Entry<String, Long>> MOST_RECENTLY_UPDATED_PARTITIONS_ORDERING = Ordering.natural()
```
chavdar: If this is MRU, shouldn't it be reverse-sorted?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
```diff
@@ -0,0 +1,339 @@
...
+          return entry.getValue();
+        }
+      });
+
+  private final int maxPartitionsPerDataset;
+
+  /**
+   * Watermarks from previous state
+   */
+  @Getter
```
chavdar: If this is made visible only for testing, then the getter should be package-private.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  /**
+   * Always 0. As this is a {@link PartitionLevelWatermarker}
+   *
+   * {@inheritDoc}
+   * @see gobblin.data.management.conversion.hive.watermarker.HiveSourceWatermarker#getPreviousHighWatermark(org.apache.hadoop.hive.ql.metadata.Table)
+   */
+  @Override
+  public LongWatermark getPreviousHighWatermark(Table table) {
```
chavdar: I am a bit confused here. Are callers supposed to know not to call getPreviousHighWatermark(table) if we are using PartitionLevelWatermarker implementation ? This looks like a break in abstraction.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
pcadabam: Fixed

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/TableLevelWatermarker.java
pcadabam: No. This implementation does not use ```partitionProcessTime```. Other Implementations may choose to use

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
pcadabam: I did limit it to 2 dots. Probably should have been gobblin.replacedPartitions

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/extractor/HiveConvertExtractor.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
pcadabam: Users will have read access. They need to read the actual watermark values. 
E.g. Reading previous high watermarks for a partition. We can make it an ImmutableMap if we don't want users to modify the key/values not sure if that is necessary.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
pcadabam: Fixed.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/MultiKeyValueLongWatermark.java
pcadabam: Gobblin's `Watermark` interface expects you to implement this method. I check gobblin codebase and it does not seem to be used. 
I have now added an implementation in case it gets used later

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
pcadabam: In line `247` we pick the `greatestOf` instead of `leastOf` `MOST_RECENTLY_UPDATED_PARTITIONS_ORDERING.greatestOf(partitionWatermarks.entrySet(), this.maxPartitionsPerDataset)`

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
pcadabam: Callers need not know the kind of `HiveSourceWatermarker` used. The `HiveSource` creates workunits for both partitioned and non partitioned hive tables. This method is used for non partitioned hive tables. 

Updated the java doc on the interface.
<pre>
 /**
   * Get high watermark for a {@link Table}. This API is used by the {@link HiveSource} for Non Partitioned hive tables
   * @param table for which a high watermark needs to be returned
   */
public LongWatermark getPreviousHighWatermark(Table table);
</pre>

Also returning always 0 is incorrect. Fixed it to delegate this call to `TableLevelWatermarker`

pcadabam:
Addressed all comments @chavdar and @abti please have another look.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/watermarker/PartitionLevelWatermarker.java
pcadabam: This code has been removed now.

abti:
LGTM

chavdar:
I think there must have been some misunderstanding with respect to the watermarker factory. Right now the implementation uses a public API (the fact that you need a constructor that takes state to instantiate a HiveSourceWatermarker implementation) that is undocumented and only visible in an obscure code location.

What  I mean was:
a) Keep the HiveSourceWatermarkerFactory interface. 
b) Have at least two implementations PartitionLevelWatermarker.Factory and TableLevelWatermarker.Factory 
c) Have HIVE_SOURCE_WATERMARKER_FACTORY_KEY instead of HIVE_SOURCE_WATERMARKER_CLASS_KEY
d) Always instantiate the factory class from HIVE_SOURCE_WATERMARKER_FACTORY_KEY using the default constructor.
e) Invoke createFromState() on the above instance.

I would even suggest that we go as far as not having the constructor for PartitionLevelWatermarker and TableLevelWatermarker take a state in the constructor but only the parts of the sate that they need. It should be the factory which adapts from the passed state to whatever is needed for the construction of  PartitionLevelWatermarker and TableLevelWatermarker.

Right now the code is full of these opaque reflection invocations and passing of state which is a glorified map. This completely obfuscates what is used where and we keep having this unintended consequence when changing something in the state.

If we want to program in Python, we should program in Python, not abuse Java :-)

chavdar:
LGTM

PR #1206, opened 2016-08-16
Refactored SchedulerUtils into a new PullFileLoader that uses Config …

ibuenros:
…to load pull files.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
+++ b/gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
@@ -1,0 +1,262 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+
+import org.apache.commons.configuration.ConfigurationConverter;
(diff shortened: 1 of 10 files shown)


ibuenros:
@autumnust @ydai1124 can you review? thanks!

gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
```diff
@@ -0,0 +1,257 @@
...
+   * @param path base {@link Path} where pull files should be found recursively.
+   * @param sysProps A {@link Config} used as fallback.
+   * @param loadGlobalProperties if true, will also load at most one *.properties file per directory from the
+   *          {@link #rootDirectory} to the pull file {@link Path} for each pull file.
+   * @return The loaded {@link Config}s.
+   */
+  public Collection<Config> loadPullFilesRecursively(Path path, Config sysProps, boolean loadGlobalProperties) {
+    try {
+      Config fallback = sysProps;
+      if (loadGlobalProperties && PathUtils.isAncestor(this.rootDirectory, path.getParent())) {
```
ydai1124: Why checking against `path.getParent()' instead of `path`?

gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
```diff
@@ -0,0 +1,257 @@
...
+          pullFiles.addAll(loadPullFilesRecursivelyHelper(status.getPath(), fallback, loadGlobalProperties));
+        } else if (this.javaPropsPullFileFilter.accept(status.getPath())) {
+          pullFiles.add(loadJavaPropsWithFallback(status.getPath(), fallback).resolve());
+        } else if (this.hoconPullFileFilter.accept(status.getPath())) {
+          pullFiles.add(loadHoconConfigAtPath(status.getPath()).withFallback(fallback).resolve());
+        }
+      }
+
+      return pullFiles;
+    } catch (IOException ioe) {
```
ydai1124: You can throw the IOException here instead of `catch` it.

gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
ibuenros: Because the global properties of the current directory and all descendants are included automatically, however, we need to check whether we also need to include global properties of the parent directory and further ancestors.

gobblin-utility/src/main/java/gobblin/util/PullFileLoader.java
ibuenros: Hmm, you're right, and that's a problem. The idea is that if a subdirectory fails to load, pull files in other directories will still be loaded. I'll fix that.

autumnust:
LGTM.

PR #1207, opened 2016-08-16
template wiki doc added

autumnust:
@ibuenros @chavdar can you review ? Thanks !

### DIFF
--- a/gobblin-docs/user-guide/Gobblin-genericLoad
+++ b/gobblin-docs/user-guide/Gobblin-genericLoad
@@ -1,0 +1,19 @@
+Table of Contents
+-------------------
+
+[TOC]
+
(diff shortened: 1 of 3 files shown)


ibuenros:
Looks good in general.

Please switch the order of sections a bit, that way people don't have to read implementation in order to use templates.
1. How to use template (include here the example with `wikipedia.pull`, just the short file)
2. Available templates. Also, for each one please include a short description and required options.
3. How to create your own template (in here include the actual `wikiSample.template`)
4. How does template work in Gobblin

Also, you should have a page on how to use pull files in HDFS (e.g. what system property needs to be set, etc.)

ibuenros:
LGTM

PR #1210, opened 2016-08-17
Make topic suffix configurable for lookup in Confluent Schema Registry

lbendig:
Currently only the name of the Kafka topic is used to retrieve the schema from the registry, however schemas are usually registered under `topicName-value` or `topicName-key`. 
Thus the schema will never be retrieved from the registry and the following exception is thrown:
```
...
Caused by: gobblin.metrics.kafka.SchemaRegistryException:
io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException:
Subject not found.; error code: 40401
...
```
This PR adds a new config parameter `kafka.schema_registry.confluent.schema.name.suffix `which defaults to `-value`. This suffix will be used along with the topic name to query the registry.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
@@ -48,7 +48,7 @@ public class ConfluentKafkaSchemaRegistry extends KafkaSchemaRegistry<Integer, S
   public static final String CONFLUENT_MAX_SCHEMAS_PER_SUBJECT =
       "kafka.schema_registry.confluent.max_schemas_per_subject";
 
-  public static final String CONFLUENT_SCHEMA_NAME_SUFFIX = "kafka.schema_registry.confluent.schema.name.suffix";
+  public static final String CONFLUENT_SCHEMA_NAME_SUFFIX = "kafka.schema_registry.confluent.schema_name_suffix";
   
   // Default suffix of the topic name to register / retrieve from the registry
   private static final String DEFAULT_CONFLUENT_SCHEMA_NAME_SUFFIX = "-value";
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
@@ -51,11 +51,13 @@ public class ConfluentKafkaSchemaRegistry extends KafkaSchemaRegistry<Integer, S
   public static final String CONFLUENT_SCHEMA_NAME_SUFFIX = "kafka.schema_registry.confluent.schema.name.suffix";
   
   // Default suffix of the topic name to register / retrieve from the registry
-  private static final String CONFLUENT_SCHEMA_DEFAULT_NAME_SUFFIX = "-value";
+  private static final String DEFAULT_CONFLUENT_SCHEMA_NAME_SUFFIX = "-value";
   
   @Getter
   private final SchemaRegistryClient schemaRegistryClient;
 
+  private final String schemaNameSuffix;
+  
   public ConfluentKafkaSchemaRegistry(Properties props) {
     this(props, new CachedSchemaRegistryClient(props.getProperty(KAFKA_SCHEMA_REGISTRY_URL),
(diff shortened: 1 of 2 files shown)


lbendig:
@shirshanka can you please review?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
```diff
@@ -48,6 +48,11 @@
   public static final String CONFLUENT_MAX_SCHEMAS_PER_SUBJECT =
       "kafka.schema_registry.confluent.max_schemas_per_subject";
 
+  public static final String CONFLUENT_SCHEMA_NAME_SUFFIX = "kafka.schema_registry.confluent.schema.name.suffix";
```
shirshanka: Can you change it to: kafka.schema_registry.confluent.schema_name_suffix  to be consistent in style with the other config keys?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
```diff
@@ -48,6 +48,11 @@
...
+  
+  // Default suffix of the topic name to register / retrieve from the registry
+  private static final String CONFLUENT_SCHEMA_DEFAULT_NAME_SUFFIX = "-value";
```
shirshanka: Naming suggestion: either DEFAULT_CONFLUENT_SCHEMA_NAME_SUFFIX  or CONFLUENT_SCHEMA_NAME_SUFFIX_DEFAULT

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
```diff
@@ -74,7 +79,8 @@ protected Schema fetchSchemaByKey(Integer key) throws SchemaRegistryException {
   @Override
   public Schema getLatestSchemaByTopic(String topic) throws SchemaRegistryException {
     try {
-      return new Schema.Parser().parse(this.schemaRegistryClient.getLatestSchemaMetadata(topic).getSchema());
+      String schemaName = topic + props.getProperty(CONFLUENT_SCHEMA_NAME_SUFFIX, CONFLUENT_SCHEMA_DEFAULT_NAME_SUFFIX);
```
shirshanka: Replace the props.getProperty invocation with a member variable: schemaNameSuffix which is populated during construction of the schema registry from the passed in props. 

String schemaName = topic + schemaNameSuffix;

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
```diff
@@ -88,7 +94,8 @@ public Integer register(Schema schema) throws SchemaRegistryException {
   @Override
   public Integer register(Schema schema, String name) throws SchemaRegistryException {
     try {
-      return this.schemaRegistryClient.register(name, schema);
+      String schemaName = name + props.getProperty(CONFLUENT_SCHEMA_NAME_SUFFIX, CONFLUENT_SCHEMA_DEFAULT_NAME_SUFFIX);
```
shirshanka: Same comment here as above.

shirshanka:
Added a few suggestions and comments. Happy to merge it in if you agree and make the changes.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
lbendig: Fixed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
lbendig: Fixed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/ConfluentKafkaSchemaRegistry.java
lbendig: I tried to be here as "stateless" as possible but this doesn't make sense, fixed.

shirshanka:
LGTM. Thanks!

lbendig:
@shirshanka , thank you for the review!

PR #1213, opened 2016-08-18
Prevent re-scheduling while a job is running; make JobExecutionDriver…

chavdar:
- Prevent re-scheduling while a job is running
- Make JobExecutionDriver like a future so that user can detect easier when a job is complete
- Add JobExecutionResult class that can be used to determine the outcome of a job and (in the future) get access to some metrics, etc.

@ibuenros can you review?

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionResult.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionResult.java
@@ -13,6 +13,8 @@ package gobblin.runtime.api;
 
 import com.google.common.base.Preconditions;
 
+import gobblin.runtime.JobState.RunningState;
+
 import lombok.AccessLevel;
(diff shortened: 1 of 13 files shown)


gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionResult.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Represents the result of the execution of a Gobblin job.
+ */
+@AllArgsConstructor(access=AccessLevel.PROTECTED)
+@Getter
+public class JobExecutionResult {
```
ibuenros: Can we add a running state here?

ibuenros:
Just one comment, LGTM otherwise.

PR #1215, opened 2016-08-19
Fixed silent failures and hung application when a standalone service …

ibuenros:
…fails to initialize.

Issue:
Previously, if `JobScheduler` failed to start, the application would hang and no error would be reported.

Solution:
Added a listener to the `ServiceManager` in `ServiceBasedAppLauncher` that logs failures of the services and shutdowns Gobblin on failure of a service.
Also separated the quartz scheduler as an independent service managed in the `ServiceManager`. This is necessary because otherwise, when `JobScheduler` fails, its `shutDown` method cannot be called, so the scheduler cannot be closed, causing the application to hang. As a separate service, the `ServiceManager` will shutdown the scheduler correctly.


ibuenros:
@chavdar can you review?

chavdar:
LGTM.

chavdar:
Pleas fix the FindBugs warnings.

PR #1216, opened 2016-08-20
Adding a simple console writer to gobblin

shirshanka:
@abti : care to review?


gobblin-core/src/main/java/gobblin/writer/ConsoleWriter.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Slf4j
+public class ConsoleWriter<D> implements DataWriter<D> {
+  private long _recordsWritten = 0;
+  @Override
+  public void write(D record)
+      throws IOException {
+    if (record != null)
+    {
+      System.out.println(record.toString());
```
abti: Might as well add log.info if console o/p stream is not captured

abti:
LGTM

PR #1217, opened 2016-08-20
Fixing console writer example

shirshanka:
Console writer doesn't expect to work with a partitioner. Will add this feature in a future PR if needed.


shirshanka:
Merging as it is a trivial change.

PR #1218, opened 2016-08-21
Making Kafka consumer configurable for Kafka extract

shirshanka:
Global configuration for Kafka consumer for Kafka extracts. 

@hibikir This generalizes the proposal in https://github.com/linkedin/gobblin/pull/1059/

While making the changes, I also discovered that the socketTimeout was being set to the same value as the fetchTimeout in the old code. This resulted in intermittent failures in Kafka extraction when the amount of data in the source topic is less than the default fetchBytes. This was also commonly reported on the mailing lists by users trying out Kafka ingest locally. 

@abti, @ibuenros : Please review!

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
@@ -61,10 +61,6 @@ import kafka.message.MessageAndOffset;
  */
 public class KafkaWrapper implements Closeable {
 
-
-
-
-
   private static final Logger LOG = LoggerFactory.getLogger(KafkaWrapper.class);
 
   private static final String USE_NEW_KAFKA_API = "use.new.kafka.api";
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
@@ -31,9 +31,12 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.net.HostAndPort;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
+import gobblin.util.ConfigUtils;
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
```diff
@@ -58,6 +61,10 @@
  */
 public class KafkaWrapper implements Closeable {
 
+
```
abti: Please delete the extra spacing

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
```diff
@@ -162,16 +182,56 @@ public void close() throws IOException {
...
+    public static final String CONFIG_KAFKA_FETCH_TOPIC_NUM_TRIES = CONFIG_PREFIX + "fetchTopicNumTries";
+    private static final int CONFIG_KAFKA_FETCH_TOPIC_NUM_TRIES_DEFAULT = 3;
+    public static final String CONFIG_KAFKA_FETCH_OFFSET_NUM_TRIES = CONFIG_PREFIX + "fetchOffsetNumTries";
+    private static final int CONFIG_KAFKA_FETCH_OFFSET_NUM_TRIES_DEFAULT = 3;
+
+    private final int socketTimeoutMillis;
+    private final int bufferSize;
+    private final String clientName;
+    private final int fetchCorrelationId;
+    private final int fetchTimeoutMillis;
```
abti: Millis generally go well with long

abti:
Minor comments, rest LGTM

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
shirshanka: The Kafka API-s ask for int, so kept it that way.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/KafkaWrapper.java
shirshanka: Fixed!

shirshanka:
Thanks for the quick review!

aitchnu:
ConfigUtils.propertiesToConfig() method throws exception if two keys are prefixes of one another  which is true in case of fork in state.getProperties().

PR #1220, opened 2016-08-22
Refactored MR mode to use GobblinInputFormat.

ibuenros:
Implements an `InputFormat` for Gobblin that does not require a single file to be read by all mappers.

Fixes the following issue:
Gobblin MR writes a file with extension *.wulist containing a list of *.wu and *.mwu files for mappers to process. Each of these mappers must read this single file, which can cause slowness in HDFS or affect datanodes.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -12,13 +12,9 @@
 
 package gobblin.runtime.mapreduce;
 
-import java.io.BufferedWriter;
 import java.io.DataOutputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
 import java.net.URI;
 import java.util.List;
 import java.util.Properties;
@@ -143,7 +139,9 @@ public class MRJobLauncher extends AbstractJobLauncher {
     this.fs = buildFileSystem(jobProps, this.conf);
 
     this.mrJobDir =
-        new Path(this.jobProps.getProperty(ConfigurationKeys.MR_JOB_ROOT_DIR_KEY), this.jobContext.getJobName());
+        new Path(
+            new Path(this.jobProps.getProperty(ConfigurationKeys.MR_JOB_ROOT_DIR_KEY), this.jobContext.getJobName()),
+            this.jobContext.getJobId());
     if (this.fs.exists(this.mrJobDir)) {
       LOG.warn("Job working directory already exists for job " + this.jobContext.getJobName());
       this.fs.delete(this.mrJobDir, true);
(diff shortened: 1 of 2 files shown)


ibuenros:
@chavdar can you review?

chavdar:
Please add a more detailed description of the change.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -427,20 +429,12 @@ private void addHdfsJars(String hdfsJarFileList)
    * Prepare the job input.
    * @throws IOException
    */
-  private Path prepareJobInput(List<WorkUnit> workUnits)
+  private void prepareJobInput(List<WorkUnit> workUnits)
       throws IOException {
-    // The job input is a file named after the job ID listing all work unit file paths
-    Path jobInputFile = new Path(this.jobInputPath, this.jobContext.getJobId() + WORK_UNIT_LIST_FILE_EXTENSION);
```
chavdar: I think it may be worthwhile to have a file with the work units for debugging purposes. Ideally, the file is in text or JSON format.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinInputFormat.java
```diff
@@ -0,0 +1,156 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+
+/**
+ * An input format for reading Gobblin inputs (work unit and multi work unit files).
+ */
+public class GobblinInputFormat extends InputFormat<LongWritable, Text> {
```
chavdar: Can you use a more descriptive name like GobblinWorkUnitesInputFormat ?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinInputFormat.java
```diff
@@ -0,0 +1,156 @@
...
+      throws IOException, InterruptedException {
+
+    Path[] inputPaths = FileInputFormat.getInputPaths(context);
+    if (inputPaths == null || inputPaths.length == 0) {
+      throw new IOException("No input found!");
+    }
+
+    List<InputSplit> splits = Lists.newArrayList();
+
+    for (Path path : inputPaths) {
```
chavdar: Please document that the input paths the files with individual work units.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ibuenros: What about logging instead? The file seems mostly useless for debugging, especially because it gets deleted on exit.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/GobblinInputFormat.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    public String[] getLocations()
+        throws IOException, InterruptedException {
+      return new String[0];
+    }
+  }
+
+  /**
+   * Returns a single record containing the name of the work unit / multi work unit file to process.
```
chavdar: As a side not, I think we should later change the input format to actually return the work units. This will allow us to store work units in something like HBase or some such.

chavdar:
Minor comments. Feel free to fix and merge.

PR #1221, opened 2016-08-22
Use FileSystem.newInstance instead of FileSystem.get

jinhyukchang:
Use FileSystem.newInstance instead of FileSystem.get as it closes shared resource.
Performed integration test.

https://github.com/linkedin/gobblin/issues/1219

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -182,6 +182,7 @@ public class MRJobLauncher extends AbstractJobLauncher {
         cleanUpWorkingDirectory();
       } finally {
         super.close();
+        fs.close();
       }
     }
   }
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -497,7 +497,7 @@ public class MRJobLauncher extends AbstractJobLauncher {
   private static FileSystem buildFileSystem(Properties jobProps, Configuration configuration)
       throws IOException {
     URI fsUri = URI.create(jobProps.getProperty(ConfigurationKeys.FS_URI_KEY, ConfigurationKeys.LOCAL_FS_URI));
-    return FileSystem.get(fsUri, configuration);
+    return FileSystem.newInstance(fsUri, configuration);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jinhyukchang:
@ibuenros Could you take a look?

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -110,7 +110,7 @@ public FileSystem getProxiedFileSystem(State properties, AuthType authType, Stri
       @Override
       public Void run() throws IOException {
         LOG.debug("Now performing file system operations as :" + UserGroupInformation.getCurrentUser());
-        proxiedFs = FileSystem.get(fsURI, conf);
+        proxiedFs = FileSystem.newInstance(fsURI, conf);
```
ibuenros: I have concerns about this, as this gets called often by other parts of Gobblin, which could impact performance. Ideally, I would want to see something like HDFS and local filesystems are not closed by the fshelper, which would allow us to use `get` in here (since proxied file systems are not generally used for filesystems that should be closed).

jinhyukchang: Got it. Applied the change accordingly.

jinhyukchang:
Synced with @shirshanka and added FileSystem.close() on MRJobLauncher. Passed integration test.

@shirshanka, @ibuenros , Could you take a look?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -182,6 +182,7 @@ public void close()
         cleanUpWorkingDirectory();
       } finally {
         super.close();
+        fs.close();
```
ibuenros: Why is this needed?

jinhyukchang: I think it's generally a good practice to close in order to release any resource that object holds. As it now has newInstance of FileSystem, instead of shared one, I think it makes sense to close it to release resource.

ibuenros: I think it is good practice for non-HDFS file systems. However, this is the fs that supports MR, which must be either HDFS or local file system afaik, so I don't see a benefit to using a new instance of the filesystem.

jinhyukchang: The benefit is to prevent this kind of error from happening again. As MR launcher has almost the longest life span and this can easily happen again if shared object is being closed. As MR launcher is not frequently instantiated, one per process if I understand correctly, I think the downside is minimal.

PR #1226, opened 2016-08-23
Making kafka writer more robust, adding tests

shirshanka:


shirshanka:
@ydai1124: Would like to get this in before 0.8.0 release
@ibuenros: review?

gobblin-kafka/src/main/java/gobblin/kafka/tool/SimpleKafkaConsumer.java
```diff
@@ -59,7 +55,7 @@
   private final ConsumerIterator<byte[], byte[]> iterator;
   private final String topic;
   private final KafkaSchemaRegistry<MD5Digest, Schema> schemaRegistry;
-  private final Deserializer<GenericRecord> deserializer;
+  private final Deserializer deserializer;
 
   public SimpleKafkaConsumer(Properties props, KafkaCheckpoint checkpoint)
```
ibuenros: What is the advantage of this over a console consumer?

gobblin-kafka/src/main/java/gobblin/kafka/tool/SimpleKafkaConsumer.java
```diff
@@ -69,10 +65,17 @@ public SimpleKafkaConsumer(Properties props, KafkaCheckpoint checkpoint)
...
+     * avroSerDeConfig.put("schema.registry.url", "http://localhost:8081");
+     * deserializer = new io.confluent.kafka.serializers.KafkaAvroDeserializer();
+     * deserializer.configure(avroSerDeConfig, false);
+     *
+     **/
 
     Properties consumeProps = new Properties();
     consumeProps.put("zookeeper.connect", zkConnect);
-    consumeProps.put("group.id", "testConsumer2");
+    consumeProps.put("group.id", "gobblin-tool-" + System.nanoTime());
```
ibuenros: Why not let zookeeper manage the checkpoint?

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -13,22 +13,31 @@
...
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 
 import lombok.extern.slf4j.Slf4j;
 
+import gobblin.util.ConfigUtils;
 import gobblin.writer.DataWriter;
 
-
+import static gobblin.kafka.writer.KafkaWriterConfigurationKeys.*;
```
ibuenros: We generally don't do star imports in Gobblin.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -38,51 +47,91 @@
...
-
+    lock = new ReentrantLock();
+    stateChange = lock.newCondition();
     this.topic = config.getString(KafkaWriterConfigurationKeys.KAFKA_TOPIC);
+    this.commitTimeoutInNanos = ConfigUtils.getLong(config, KafkaWriterConfigurationKeys.COMMIT_TIMEOUT_MILLIS_CONFIG,
+        KafkaWriterConfigurationKeys.COMMIT_TIMEOUT_MILLIS_DEFAULT) * MILLIS_TO_NANOS;
     this.producer = producer;
     this.producerCallback = new Callback() {
       @Override
       public void onCompletion(RecordMetadata metadata, Exception exception) {
```
ibuenros: Will this work with high volumes of data? Won't it slow down significantly? Especially with the lock?

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -38,51 +47,91 @@
...
+  private static final long MILLIS_TO_NANOS = 1000 * 1000;
+
   private final KafkaProducer<String, D> producer;
   private final Callback producerCallback;
   private final String topic;
-  private final AtomicInteger recordsWritten;
+  private final long commitTimeoutInNanos;
   private final AtomicInteger bytesWritten;
+  private final AtomicInteger recordsWritten;
+  private long recordsProduced;
```
ibuenros: You can use metrics instead (in particular `Counter`). Just extend `InstrumentedDataWriter` and create your custom metrics.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -38,51 +47,91 @@
...
   }
 
   public KafkaDataWriter(KafkaProducer producer, Config config)
   {
+    recordsProduced = 0;
     recordsWritten = new AtomicInteger(0);
+    recordsFailed = 0;
     bytesWritten = new AtomicInteger(-1);
-
+    lock = new ReentrantLock();
```
ibuenros: what is the use of this lock?

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -104,23 +153,68 @@ private static Properties stripPrefix(Properties props, String prefix) {
...
+        } catch (InterruptedException e) {
+          throw new IOException(e);
+        }
+      }
+      log.debug("Commit done waiting");
+      if (recordsProduced != (recordsWritten.get() + recordsFailed)) // timeout
+      {
+        String message = "Commit timeout: RecordsWritten = " + recordsWritten.get()
+            + "; RecordsFailed = " + recordsFailed;
+        log.debug(message);
```
ibuenros: Shouldn't this be logged even if it didn't time out?

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
```diff
@@ -104,23 +153,68 @@ private static Properties stripPrefix(Properties props, String prefix) {
...
+        throw new IOException(message);
```
ibuenros: If a writer fails to publish all messages, it will cause a failure of the work unit, so the watermark won't get advanced. This would cause Gobblin to keep resending the same messages over and over if even a single one is bad. Could we have different semantics here, where failures of a few messages still cause success? (We have metrics to notify user that there were failed messages).

gobblin-kafka/src/test/java/gobblin/kafka/ErrorManager.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Random;
+import java.util.regex.Pattern;
+
+import com.typesafe.config.Config;
+
+import gobblin.util.ConfigUtils;
+
+
+/**
+ * Created by sdas on 7/17/16.
```
ibuenros: Can you add a better javadoc?

ibuenros:
There are some test failures in this PR. Historically, we've had issues with tests using an embedded Kafka server, especially around timing (Travis nodes are very slow in general).
Is there any way we can avoid using an embedded Kafka server? In metrics we get around this by using a `KafkaPusher` and a `MockKafkaPusher`. There is also a `HighLevelConsumer` and a `MockHighLevelConsumer` for manually pushing messages into something that looks like a Kafka high level consumer.

gobblin-kafka/src/main/java/gobblin/kafka/tool/SimpleKafkaConsumer.java
shirshanka: Much harder to restart consumption. You have to delete offsets separately. 
This is much better if you are debugging issues with pulling from kafka because you don't have to worry about your checkpoint getting committed for you automatically. 

kafka 0.9 onwards has better consumers, but I wanted to be able to run this directly from gobblin, the ide etc.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
shirshanka: not even for statics?

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
shirshanka: Good point. I was optimizing for not busy waiting / poll-based waiting on the commit. 
I'll try a micro-benchmark to see how bad it is. I can change it to a no-lock based callback.

gobblin-kafka/src/main/java/gobblin/kafka/writer/KafkaDataWriter.java
shirshanka: Is this a showstopper or nice-to-have :)

shirshanka:
The failures seem related to gobblin-runtime:test not gobblin-kafka:test. 

I noticed KafkaPusherTest uses an EmbeddedKafkaServer, so was using that as reference.

gobblin-kafka/src/test/java/gobblin/kafka/ErrorManager.java
shirshanka: Primarily to simulate errors thru configuration for the FlakyKafkaProducer. Updating the javadoc.

ibuenros:
The idea is `KafkaPusherTest` is the only test with an embedded kafka, everything else uses a pusher. That way, worst case scenario, if we trust the pusher, we can disable that test in Travis or other unreliable systems.

ydai1124:
@shirshanka Will you be able to have a doc for Kafka writer? So I can reference it in the release.

shirshanka:
@ibuenros : Addressed most of the review comments.
@ydai1124 : will open a separate PR for the doc.

ibuenros:
LGTM

PR #1228, opened 2016-08-23
Templates use config instead of properties.

ibuenros:


ibuenros:

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -149,6 +149,8 @@ protected void startUp()
       throws Exception {
     LOG.info("Starting the job scheduler");
 
+    this.scheduler.awaitRunning();
```
ibuenros: Note: this is a small improvement for a different issue. `JobScheduler` should not start until the `SchedulerService` is running.

ydai1124:
LGTM

PR #1231, opened 2016-08-24
Fix distcp hive doc

ydai1124:


ydai1124:
Trivial doc change. Merging.

PR #1240, opened 2016-08-29
Add backward compatibility in time based retention

pcadabam:
1. Some old jobs are still using `gobblin.retention.minutes.retained` property for `TimeBasedRetentionPolicy`. Added backward compatibility 
2. Removed the default retention of 1 day as it is risky to silently set 1 day as retention

@ydai1124 can you review?


ydai1124:
LGTM

PR #1242, opened 2016-08-30
Added option to specify destination table properties and added Hive properties for tracking in Avro to ORC

abti:
Initially this PR added a way to specify destination compression via configuration but as per discussion changed this PR to: 
- Add a way to specify destination table properties via configuration and;
- Add Hive properties for tracking in Avro to ORC


abti:
@pcadabam Please review.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/dataset/ConvertibleHiveDataset.java
```diff
@@ -136,6 +136,7 @@ public ConvertibleHiveDataset(FileSystem fs, HiveMetastoreClientPool clientPool,
     public static final String DESTINATION_TABLE_KEY = "destination.tableName";
     public static final String DESTINATION_DB_KEY = "destination.dbName";
     public static final String DESTINATION_DATA_PATH_KEY = "destination.dataPath";
+    public static final String DESTINATION_TABLE_COMPRESSION_KEY = "destination.tableCompression";
```
pcadabam: Do you think the property should be ```destination.additionalTableProperties``` instead of ```destination.tableCompression```? This will allow us to add more table properties without any code change.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
```diff
@@ -168,6 +168,13 @@ public String getConfigPrefix() {
             : Optional.of(getConversionConfig().getClusterBy());
     Optional<Integer> numBuckets = getConversionConfig().getNumBuckets();
     Optional<Integer> rowLimit = getConversionConfig().getRowLimit();
+    Optional<String> orcTableCompression = getConversionConfig().getDestinationTableCompression();
+    Optional<Map<String, String>> optionalTblProperties = Optional.<Map<String, String>>absent();
```
pcadabam: This does not need to ```Optional<Map<String, String>>```. It can be an empty may if nothing exists

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/dataset/ConvertibleHiveDataset.java
abti: Changed to table properties.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
abti: Changed to properties.

pcadabam:
LGTM

PR #1243, opened 2016-08-30
Use avro permission for orc data

pcadabam:
- Manually create ORC data location with the same permissions as Avro data
- Note that hive can also automatically create the non-existing directories, but it does not
seem to create it with the desired permissions.
According to hive docs, permissions for newly created directories/files can be controlled using uMask.
I tried the following but it did not honor the umask
```
hive (default)> SET hive.warehouse.subdir.inherit.perms=false;
hive (default)> SET fs.permissions.umask-mode=022;
hive (default)> CREATE EXTERNAL TABLE IF NOT EXISTS `testtable` (`Id` int, `Urn` string) LOCATION '/user/pcadabam/testtable';
OK
Time taken: 6.031 seconds
```
Directory got created with 750 permissions. Expected is 755
@abti can you review?


abti:
LGTM

PR #1244, opened 2016-08-31
Config wrapper for data replication

tuGithub:


tuGithub:
ReplicationConfiguration is the overall config for data replication, including ReplicationMetaData, ReplicationSource, ReplicationReplicas, and DataFlowTopology

ReplicationUtils used to build different java class based on the type safe Config

tuGithub:
@chavdar , @ibuenros , please review it

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,124 @@
...
+
+    this.routes = new ArrayList<CopyRoute>();
+
+    Config routesConfig = builder.topologyConfig.getConfig(ROUTES);
+    // routes should be available for each replica
+    for (ReplicationReplica replica : builder.replicas) {
+      Preconditions.checkArgument(routesConfig.hasPath(replica.getReplicationName()),
+          "can not find route for replica " + replica.getReplicationName());
+
+      CopyRoute route = new CopyRoute();
```
ibuenros: Can you use a `builder` instead and make `CopyRoute` immutable?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,124 @@
...
+      route.copyDestination = replica;
+      route.copyFrom = new ArrayList<ReplicationData<CopyableDataset>>();
+
+      List<String> copyFroms = routesConfig.getStringList(replica.getReplicationName());
+
+      for (String copyFromName : copyFroms) {
+        Preconditions.checkArgument(!copyFromName.equals(route.copyDestination.getReplicationName()),
```
ibuenros: Maybe change this to a warning and only throw error if there are no valid `copyFroms`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,124 @@
...
+        // copy from source
+        if (copyFromName.equals(ReplicationUtils.REPLICATION_SOURCE)) {
+          route.copyFrom.add(builder.source);
+        }
+        // copy from other replicas
+        else if(replicasMap.containsKey(copyFromName)){
+          route.copyFrom.add(replicasMap.get(copyFromName));
+        }
+        else{
+          throw new IllegalArgumentException("can not find replica with name " + copyFromName);
```
ibuenros: As above, would prefer this to be a warning.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+
+      this.routes.add(route);
+    }
+  }
+
+  public static class CopyRoute {
+    @Getter
+    private ReplicationReplica copyDestination;
```
ibuenros: Please make `final`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+
+/**
+ * Class ReplicationConfiguration is used to describe the overall configuration of the replication flow for
+ * a specific {@link ReplicationData}, including:
+ * <ul>
+ *  <li>Meta data
```
ibuenros: Can you use `{@link ...}` to the relevant classes?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public interface ReplicationData<T extends Dataset> {
+
+  public boolean isSource();
+
+  public String getReplicationName();
+
+  public ReplicationLocation getReplicationLocation();
+
+  public DatasetsFinder<T> getDatasetsFinder();
```
ibuenros: Are there multiple `Dataset`s in a single `ReplicationData`? Shouldn't this be `Dataset getDataset()` instead?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
```diff
@@ -0,0 +1,31 @@
...
+ *  <li>Specifying the replication name
+ *  <li>Specifying the replication location {@link ReplicationLocation}
+ *  <li>Specifying the {@link DatasetsFinder} which can find all data sets
+ * </ul>
+ * 
+ * @author mitu
+ *
+ * @param <T>
+ */
+public interface ReplicationData<T extends Dataset> {
```
ibuenros: Should it be `T extends CopyableDataset`?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.Getter;
+
+/**
+ * Class used to represent the meta data of the replication
+ * @author mitu
+ *
+ */
+
+@AllArgsConstructor
+public class ReplicationMetaData {
```
ibuenros: This class seems unnecessarily restrictive, in that it will be hard in the future to add new fields, is there a problem with simply using a map? Otherwise we should really think about what are the correct fields here.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Getter
+  private final Optional<String> jira;
+
+  @Getter
+  private final Optional<String> owner;
```
ibuenros: Is this metadata for the replication or for the dataset?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Class used to represent the meta data of the replication
+ * @author mitu
+ *
+ */
+
+@AllArgsConstructor
```
ibuenros: You can use `@Data` instead, it combines `@AllArgsConstructor`, `@Getter`, and generates a `toString()` method very similar to the one you implemented.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationType.java
```diff
@@ -0,0 +1,41 @@
+package gobblin.data.management.copy.replication;
+
+/**
+ * An enumeration of possible replication types which is the attribute of {@link ReplicationLocation}
+ * 
+ * @author mitu
+ *
+ */
+public enum ReplicationType {
```
ibuenros: What is the use of this enum instead of classes?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationType.java
```diff
@@ -0,0 +1,41 @@
...
+    return this.name;
+  }
+
+  /**
+   * Get a {@link ReplicationType} for the given name.
+   *
+   * @param name the given name
+   * @return a {@link ReplicationType} for the given name
+   */
+  public static ReplicationType forName(String name) {
```
ibuenros: Not really needed, you could do `ReplicationType.valueof(string.toUpperCase())` instead.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+
+  public static final String REPLICATION_SOURCE = "source";
+  public static final String REPLICATION_REPLICAS = "replicas";
+  public static final String REPLICATOIN_REPLICAS_LIST = "list";
+
+  public static final String REPLICATION_LOCATION_TYPE_KEY = "type";
+
+  public static final String DATAFLOWTOPOLOGY = "dataFlowTopology";
+
+  public static ReplicationMetaData buildMetaData(Config config) {
```
ibuenros: Can you move these method to their respective classes? e.g. `ReplicationMetadata.fromConfig(Config)`.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+    Preconditions.checkArgument(config.hasPath(type));
+
+    ReplicationType RType = ReplicationType.forName(type);
+
+    if (RType == ReplicationType.HDFS) {
+      return new HdfsReplicationLocation(config.getConfig(type));
+    }
+
+    // TODO, other replication types implmenetation 
+    return null;
```
ibuenros: Can you instead throw an `UnsupportedOperationException` with the explanation?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+  
+  //TODO
+  /**
+   * Based on the input {@link Config} , build corresponding {@link DatasetsFinder}
+   * @param config
+   * @return
+   */
+  public static DatasetsFinder<CopyableDataset> buildDatasetFinder(Config config){
+    return null;
```
ibuenros: Please use `UnsupportedOperationException` instead.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: Still prefer to have the error failure for miss configs, we can talk offline

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: Still prefer to have the error failure for miss configs, we can talk offline

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: done

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
tuGithub: Each ReplicationData should map to a single dataset ( which contains multiple Filesets)

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
tuGithub: done

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
tuGithub: changed to a Map

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
tuGithub: It is for a replication which used to replicate a dataset

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationMetaData.java
tuGithub: changed, much simpler

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationType.java
tuGithub: prefer to have enum for limited "valid" types

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationType.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
tuGithub: Would still prefer to have all those in an utility class. We can talk offline

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
tuGithub: changed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
tuGithub: It is a place holder, will work on the actual code soon

tuGithub:
please ignore following files as those are not final yet

gobblin-data-management/src/main/java/gobblin/data/management/copy/ConcurrentBoundedWorkUnitList.java
gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableDatasetBase.java
gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
chavdar: In that case, @ibuenros comment holds. We don't need a finder if it is supposed to be a single dataset.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
```diff
@@ -0,0 +1,31 @@
...
+ *  <li>Specifying the replication name
+ *  <li>Specifying the replication location {@link ReplicationLocation}
+ *  <li>Specifying the {@link DatasetsFinder} which can find all data sets
+ * </ul>
+ * 
+ * @author mitu
+ *
+ * @param <T>
+ */
+public interface ReplicationData<T extends CopyableDataset> {
```
chavdar: I find the ReplicationData name a bit vague. How about ReplicationEndPoint ?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public boolean isSource();
```
chavdar: Please add javadocs to interface methods.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationLocation.java
```diff
@@ -0,0 +1,12 @@
+package gobblin.data.management.copy.replication;
+
+/**
+ * Interface ReplicationLocation is used to specify the location of a replication data
+ * 
+ * @author mitu
+ *
+ */
+public interface ReplicationLocation {
```
chavdar: I don't understand this interface. There is nothing that describes an actual location.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public String getReplicationName();
+
+  public ReplicationLocation getReplicationLocation();
```
chavdar: The ReplicationLocation interface does not seem to declare any useful methods. It is unusable without a bunch of if () else statements with typecasts which then invoke implementation-specific behavior. This is not a proper use of interfaces. Interfaces should not require a use of typecasts.

Further, there might be more than one "locations" associated with the copy. For example, when copying Hive tables, we may need to know both the filesystem paths where to write the data files and the Hive DB/Table names.

Finally, "hdfs" is a bad name for a type. We should support any Hadoop FileSystem implementation.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,150 @@
...
+
+import com.google.common.base.Preconditions;
+import com.typesafe.config.Config;
+
+import gobblin.data.management.copy.CopyableDataset;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Class to represent the data flow topology from copy destination to copy sources
```
chavdar: Flows always flow from source to destination even though the actual route specification may use the destination as a key.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/AbstractReplicationData.java
```diff
@@ -0,0 +1,55 @@
...
+    return this.location;
+  }
+  
+  @Override
+  public DatasetsFinder<CopyableDataset> getDatasetsFinder(){
+    return this.datasetFinder;
+  }
+
+  @Override
+  public String toString() {
```
chavdar: Have a look at Objects.toStringHelper . It can help a lot with the implementation of methods like this one.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationReplica.java
```diff
@@ -0,0 +1,16 @@
+package gobblin.data.management.copy.replication;
+
+import gobblin.data.management.copy.CopyableDataset;
+import gobblin.dataset.DatasetsFinder;
+
+public class ReplicationReplica extends AbstractReplicationData {
```
chavdar: This looks like tautology. How about simple "Replica".

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,150 @@
...
+    private List<ReplicationReplica> replicas;
+
+    private Config topologyConfig;
+
+    public Builder withReplicationSource(ReplicationSource source) {
+      this.source = source;
+      return this;
+    }
+
+    public Builder withReplicationReplicas(List<ReplicationReplica> replicas) {
```
chavdar: When dealing with lists in builders, I think it is useful to have a method that can add individual elements to the list rather then require populating the list separately. For example, the API can be "addReplica(Replica replica)".

gobblin-data-management/src/test/java/gobblin/data/management/copy/replication/ReplicationConfigurationTest.java
```diff
@@ -0,0 +1,143 @@
...
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+@Test(groups = {"gobblin.data.management.copy.replication"})
+public class ReplicationConfigurationTest {
+  @Test
+  public void testValidConfigsWithoutMetaData(){
+    URL u = getClass().getClassLoader().getResource("replicationConfigTest/missMetaData.conf");
+    File configFile = new File(u.getFile());
+    Config c = ConfigFactory.parseFile(configFile).resolve();
```
chavdar: Why not use ConfigFactory.parseResources() directly?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,150 @@
...
+      CopyRouteBuilder routeBuilder = new CopyRouteBuilder();
+      routeBuilder.withCopyDestination(replica);
+      routeBuilder.withCopyFrom(copyFromReplica);
+      this.routes.add(routeBuilder.build());
+    }
+  }
+
+  public static class CopyRoute {
+    @Getter
+    private final ReplicationReplica copyDestination;
```
chavdar: nit: it's nice to have the names be consistent: either copyFrom/copyTo or copySource/copyDestination or routeSource/routeDestination.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,150 @@
...
+    private final List<ReplicationData<CopyableDataset>> copyFrom;
+    
+    private CopyRoute(CopyRouteBuilder builder){
+      this.copyDestination = builder.copyDestination;
+      this.copyFrom = builder.copyFrom;
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
```
chavdar: guava Joiner?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,150 @@
...
+    private ReplicationReplica copyDestination;
+    
+    private List<ReplicationData<CopyableDataset>> copyFrom;
+    
+    public CopyRouteBuilder withCopyDestination(ReplicationReplica copyDest) {
+      this.copyDestination = copyDest;
+      return this;
+    }
+
+    public CopyRouteBuilder withCopyFrom(List<ReplicationData<CopyableDataset>> copyFrom) {
```
chavdar: consider having addCopySource() or addRouteSource() instead.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,277 @@
...
+  public static final String CONFIG_STORE_REPLICATION_ROOT = "config.store.replication.root";
+  public static final String CONFIG_STORE_REPLICATION_TAG = "config.store.replication.tag";
+
+  public static void displayAllConfig(Config c){
+    for(Map.Entry<String, ConfigValue> entry :c.entrySet()){
+      log.info("AAA key " + entry.getKey() + " value " + entry.getValue());
+    }
+  }
+
+  private DataFlowTopology.CopyRoute getCopyRoute(DataFlowTopology topology)throws IOException{
```
chavdar: What is this method supposed to do? Does it determine which route to use for a dataset? If yes, it seems to make 2 assumptions:
- We are always running in pull mode (it uses the default cluster)
- There is only one ReplicationEntryPoint per cluster. 

Both of those assumptions are incorrect.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationData.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/AbstractReplicationData.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationReplica.java
tuGithub: I used ReplicationSource and ReplicationReplica in pair, if change to Replica, change ReplicationSource to Source ? I think that is too big a name for "source"

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: fixed

gobblin-data-management/src/test/java/gobblin/data/management/copy/replication/ReplicationConfigurationTest.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
tuGithub: still working on this, please ignore this for now

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -122,82 +122,22 @@
...
-      } catch (InterruptedException ie) {
-        log.error("Retrieval of work units was interrupted. Aborting.");
+      if(!generateWorkunitsBasedOnConfig(state, workUnitList, minWorkUnitWeight)){
         return Lists.newArrayList();
       }
 
       log.info(String.format("Created %s workunits ", workUnitList.getWorkUnits().size()));
 
-      copyConfiguration.getCopyContext().logCacheStatistics();
+      //copyConfiguration.getCopyContext().logCacheStatistics();
```
chavdar: Please remove.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -122,82 +122,22 @@
     this.metricContext = Instrumented.getMetricContext(state, CopySource.class);
 
     try {
-
-      final FileSystem sourceFs = getSourceFileSystem(state);
```
chavdar: I see a lot of code moved around but I don't see any change in logic in CopySource. Is that correct? If not, can you please describe the nature of the change?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/AbstractReplicationEndPoint.java
```diff
@@ -0,0 +1,48 @@
+package gobblin.data.management.copy.replication;
+
+import org.testng.collections.Objects;
+
+import gobblin.data.management.copy.CopyableDataset;
+
+
+/**
+ * Abstract class which has the basic implementations
```
chavdar: This is not a very useful javadoc. Please describe:
- What implementations the class provides
- What implementations are left to the sub-classes.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+@Slf4j
+public class CopySourceWithMultipleSourceLocation extends CopySource {
+
+  public static final String CONFIG_STORE_ROOT = "config.store.root";
+  public static final String CONFIG_STORE_REPLICATION_ROOT = "config.store.replication.root";
+  public static final String CONFIG_STORE_REPLICATION_TAG = "config.store.replication.tag";
+
+  public static void displayAllConfig(Config c) {
+    for (Map.Entry<String, ConfigValue> entry : c.entrySet()) {
+      log.info("AAA key " + entry.getKey() + " value " + entry.getValue());
```
chavdar: Make sure that this does not get merged.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+import gobblin.data.management.copy.ConcurrentBoundedWorkUnitList;
+import gobblin.data.management.copy.CopyConfiguration;
+import gobblin.data.management.copy.CopySource;
+
+
+@Slf4j
+public class CopySourceWithMultipleSourceLocation extends CopySource {
+
+  public static final String CONFIG_STORE_ROOT = "config.store.root";
+  public static final String CONFIG_STORE_REPLICATION_ROOT = "config.store.replication.root";
```
chavdar: Please add documentation to these options.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+    List<Config> configs = this.generateReplicationConfigs(originState);
+
+    for (Config config : configs) {
+      ReplicationConfiguration rc = ReplicationConfiguration.buildFromConfig(config);
+      if (!generateWorkunitsPerReplicationDataset(originState, workUnitList, minWorkUnitWeight, rc)) {
+        return false;
+      }
+    }
+    return true;
+    //    String storeRoot;
```
chavdar: Please remove any un-needed code.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopySource.java
```diff
@@ -122,82 +122,22 @@
...
-          try {
-            future.get();
-          } catch (ExecutionException exc) {
-            log.error("Failed to get work units for dataset.", exc.getCause());
-          }
-        }
-      } catch (InterruptedException ie) {
-        log.error("Retrieval of work units was interrupted. Aborting.");
+      if(!generateWorkunitsBasedOnConfig(state, workUnitList, minWorkUnitWeight)){
         return Lists.newArrayList();
```
chavdar: nit: Collections.emptyList() is slight more efficient since it returns a immutable singleton instance.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+      Collection<URI> importedBy = configClient.getImportedBy(new URI(replicationTag), true);
+
+      List<URI> replicationDataset = new ArrayList<URI>(importedBy);
+      Comparator<URI> pathLengthComparator = new Comparator<URI>() {
+        public int compare(URI c1, URI c2) {
+          return c1.getPath().length() - c2.getPath().length();
+        }
+      };
+
+      // sort the URI based on the path length to make sure the parent path appear before children
```
chavdar: It's hard to review this part since I am not completely sure what CONFIG_STORE_REPLICATION_ROOT is .

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationConfiguration.java
```diff
@@ -0,0 +1,103 @@
...
+  private final SourceEndPoint source;
+
+  @Getter
+  private final List<ReplicaEndPoint> replicas;
+
+  @Getter
+  private final DataFlowTopology topology;
+
+  public static ReplicationConfiguration buildFromConfig(Config config) {
+    if (config == null)
```
chavdar: Shouldn't this be a Precondition.checkNotNull() ?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,157 @@
...
+    return metaData;
+  }
+
+  public static SourceEndPoint buildSource(Config config) {
+    Preconditions.checkArgument(config.hasPath(REPLICATION_SOURCE),
+        "missing required config entery " + REPLICATION_SOURCE);
+
+    Config sourceConfig = config.getConfig(REPLICATION_SOURCE);
+
+    return new SourceEndPoint(ReplicationUtils.buildReplicationLocation(sourceConfig),
```
chavdar: This should be a static method in SourceEndPoint .

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+    log.info("route is " + route);
+
+    // TODO, for now only pick the first copyFrom
+    if (source.getReplicationLocation().getType() == ReplicationLocationType.HADOOPFS) {
+
+      ReplicationLocation copyTo = route.getCopyTo().getReplicationLocation();
+      ReplicationLocation copyFrom = route.getCopyFrom().get(0).getReplicationLocation();
+      stateCp.setProp(ConfigurationKeys.SOURCE_FILEBASED_FS_URI, copyFrom.getFsURI());
+      stateCp.setProp(ConfigurableGlobDatasetFinder.DATASET_FINDER_PATTERN_KEY, copyFrom.getRootPath());
+      log.info("source dir is " + copyFrom.getRootPath());
```
chavdar: Logging should provide enough context what it relates to. If you have multiple datasets being processed in parallel, it will be impossible to determine what this logging relates to.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/CopySourceWithMultipleSourceLocation.java
```diff
@@ -0,0 +1,276 @@
...
+    DataFlowTopology topology = rc.getTopology();
+    for (DataFlowTopology.CopyRoute route : topology.getRoutes()) {
+      log.info("route: " + route);
+    }
+
+    DataFlowTopology.CopyRoute route = this.getCopyRoute(topology);
+    log.info("route is " + route);
+
+    // TODO, for now only pick the first copyFrom
+    if (source.getReplicationLocation().getType() == ReplicationLocationType.HADOOPFS) {
```
chavdar: If we are using interfaces and polymorphism correctly, we should not have a getType() and if/switch statements. This makes the whole approach hard to extend because every time we add a new implementation, we'll have to go change all the places where these if/switch statements exist.

chavdar: In this case, you seem to take a Location and return a bunch of work units. So this should be encoded as a method in the Location interface.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowRoutesPicker.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.typesafe.config.Config;
+
+
+/**
+ * Used to pick preferred {@link DataFlowTopology.CopyRoute} in {@link com.typesafe.config.Config} format
+ * @author mitu
+ *
+ */
+public interface DataFlowRoutesPicker {
```
chavdar: I assume this depends on the source or destination. How is that passed?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,157 @@
...
+      Config allRoutes = config.getConfig(DEFAULT_ALL_ROUTES_KEY);
+      String routePickerClassName = dataflowConfig.hasPath(ROUTES_PICKER_CLASS_KEY)
+          ? dataflowConfig.getString(ROUTES_PICKER_CLASS_KEY) : DEFAULT_ROUTES_PICKER_CLASS_KEY;
+
+      List<Object> args = Lists.newArrayList(allRoutes, source);
+
+      try {
+        Class<?> routePickerClass = Class.forName(routePickerClassName);
+
+        DataFlowRoutesPicker picker =
```
chavdar: A couple of meetings ago, we discussed in that there should not be invocation of non-default constructors through reflection.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowTopology.java
```diff
@@ -0,0 +1,161 @@
...
+
+import gobblin.data.management.copy.CopyableDataset;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+
+/**
+ * Class to represent the data flow topology from copy source to copy destinations
+ * @author mitu
+ *
```
tuGithub: done

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
tuGithub: done refactoring the code by removing ReplicationUtils class

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationUtils.java
```diff
@@ -0,0 +1,157 @@
...
+        "missing required config entery " + REPLICATION_REPLICAS);
+
+    Config replicaConfig = config.getConfig(REPLICATION_REPLICAS);
+    Preconditions.checkArgument(replicaConfig.hasPath(REPLICATOIN_REPLICAS_LIST),
+        "missing required config entery " + REPLICATOIN_REPLICAS_LIST);
+    List<String> replicas = replicaConfig.getStringList(REPLICATOIN_REPLICAS_LIST);
+
+    List<ReplicaEndPoint> result = new ArrayList<ReplicaEndPoint>();
+    for (String replicaName : replicas) {
+      Config subConfig = replicaConfig.getConfig(replicaName);
```
tuGithub: done

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/AbstractReplicationEndPoint.java
tuGithub: done

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationConfiguration.java
tuGithub: done

tuGithub:
Removed the half done class CopySource and CopySourceWithMultipleSourceLocation.java

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationCopyMode.java
```diff
@@ -0,0 +1,44 @@
...
+
+
+/**
+ * Specify the replication copy mode, either Pull or Push
+ * @author mitu
+ *
+ */
+public enum ReplicationCopyMode {
+  PULL("pull"),
+  PUSH("push");
```
chavdar: You may have to do some thinking about the push support with respect to the route selection. In push mode, it is a bit harder because different sources may try to make the decision at the same time. It is possible that some routing policies are not possible in push mode.

tuGithub:
ReplicationConfiguration : overall config for a dataset replication
- ReplicationCopyMode: Push/Pull
- ReplicationMetaData : meta data
- SourceEndPoint : define the original source
- ReplicaEndPoint : a list of replication end points
- DataFlowTopology : the data flow topology

Interface DataFlowRoutesPicker : if topology is not defined using literal, use this interface to pick the preferred route from default routes
- class DataFlowRoutesPickerBySourceCluster : pick route based on the original source location

Interface ReplicationEndPoint : describe the replication dataset information,
- abstract class: AbstractReplicationEndPoint
- actual class: SourceEndPoint/ReplicaEndPoint

Interface ReplicationLocation : describe the location information of the dataset, for dataset on hadoopFs, describe cluster name, colo name, root path etc.
- class HdfsReplicationLocation

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/DataFlowRoutesPicker.java
tuGithub: this only defined the interface , how to pass it refer to detail class DataFlowRoutesPickerBySourceCluster

chavdar: I still don't understand. How do you instantiate a DataFlowRoutesPicker implementation?

tuGithub:
Too messy in this PR, created a new PR: data replication configs #1276

chavdar:
I assume this PR can be closed as it has been superseded by 1276 .

PR #1246, opened 2016-08-31
Various fixes to standalone and additions for boss workshop

ibuenros:
Parts of this PR:
* New converters: `GobblinMetricsFlattenerConverter` (flattens Metric report), `SchemaInjector` (injects a string schema to sources that can't provide a schema), `AvroToJsonStringConverter`, `JsonToStringConverter` (uses Gson to serialize json into string), `StringToBytesConverter` (using UTF8 encoding).
* New/improved sources/extractors: `FixedSchemaKafkaAvroExtractor`: allows ingesting avro topics wiht a fixed schema when schema registry is not available, `SimpleKafkaSource`: single entry point to Kafka extractors.
* `WikipediaSource` refactoring: previously Wikipedia source would simply pull the last 10 revisions over and over, which is not a good way of showing Gobblins capabilities. Modified it so it pulls revisions, remembers last revision pulled per article, and pushes new revisions each run.
* Some fixes to `KafkaAvroReporter`.
* Fixes to `JobScheduler` to improve standalone Gobblin execution: exit Gobblin correctly on failure of services, skip invalid pull files, etc.
* Bug fixes to pull file change detector.
* Updated templates and moved them to `gobblin-runtime`.


ibuenros:
@ydai1124 @chavdar can you review?

chavdar:
There are issues with the build: 

    Execution failed for task ':gobblin-core:findbugsMain'. 
    Execution failed for task ':gobblin-cluster:test'.
    Execution failed for task ':gobblin-utility:compileTestJava'.

gobblin-core/src/main/java/gobblin/converter/GobblinMetricsFlattenerConverter.java
```diff
@@ -0,0 +1,70 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * Flatten {@link gobblin.metrics.MetricReport} for use by Pinot.
+ * The output schema can be found at "gobblin-metrics/src/main/avro/FlatGobblinMetric.avsc".
+ */
+@Slf4j
+public class GobblinMetricsFlattenerConverter extends Converter<Schema, Schema, GenericRecord, GenericRecord> {
```
chavdar: I'd put Pinot in the name since the output format is specific to it.

gobblin-core/src/main/java/gobblin/converter/SchemaInjector.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+
+import gobblin.configuration.WorkUnitState;
+
+
+/**
+ * Injects a string schema into specified by key {@link #SCHEMA_KEY}.
+ */
+public class SchemaInjector<SI, DI> extends Converter<SI, String, DI, DI> {
```
chavdar: StringSchemaInjector?

gobblin-core/src/main/java/gobblin/converter/json/JsonStringToJsonIntermediateConverter.java
```diff
@@ -4,27 +4,34 @@
...
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 
 
+/**
+ * Converts a json string to a {@link JsonObject}.
+ */
 public class JsonStringToJsonIntermediateConverter extends Converter<String, JsonArray, String, JsonObject> {
 
+  private static Logger log = LoggerFactory.getLogger(JsonStringToJsonIntermediateConverter.class);
```
chavdar: final

gobblin-core/src/main/java/gobblin/converter/json/JsonStringToJsonIntermediateConverter.java
```diff
@@ -39,39 +46,6 @@ public JsonArray convertSchema(String inputSchema, WorkUnitState workUnit) throw
...
-        } else {
-          outputRecord.add(expectedColumnName, inputRecord.get(expectedColumnName));
-        }
-      } else {
-        outputRecord.add(expectedColumnName, JsonNull.INSTANCE);
-      }
-
-    }
-    return new SingleRecordIterable<>(outputRecord);
+    return Lists.newArrayList(inputRecord);
```
chavdar: This seems to be changing the semantics of the converter. This is dangerous as it is used quite a lot.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/FixedSchemaKafkaAvroExtractor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.avro.io.Decoder;
+import org.apache.avro.io.DecoderFactory;
+
+import com.google.common.base.Preconditions;
+
+import gobblin.configuration.WorkUnitState;
+
+
+/**
+ * Extract avro records from a Kafka topic using a fixed schema provided by {@link #STATIC_SCHEMA_ROOT_KEY}.
```
chavdar: nit: "fixed Avro schema"

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/SimpleKafkaSource.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Override
+  public Extractor<S, D> getExtractor(WorkUnitState state) throws IOException {
+    Preconditions.checkArgument(state.contains(EXTRACTOR_TYPE), "Missing key " + EXTRACTOR_TYPE);
+
+    try {
+      Class<? extends KafkaExtractor> klazz = ExtractorType.valueOf(state.getProp(EXTRACTOR_TYPE).toUpperCase()).getKlazz();
+
+      return GobblinConstructorUtils.invokeLongestConstructor(klazz, state);
```
chavdar: I wanted to cover this in one of my GOM sessions. I think we should avoid invoking non-default constructors through reflection as this is essentially an undocumented API. We should instead use a factory interface with a default constructor which has a "create(State)" method.

chavdar: You should also check with @pcadabam . He has some helper classes where you can instantiate classes by class name or alias.

gobblin-metrics/src/main/java/gobblin/metrics/kafka/KafkaReporter.java
```diff
@@ -28,22 +27,31 @@
...
   protected final KafkaPusher kafkaPusher;
 
+
   protected KafkaReporter(Builder<?> builder, Config config) throws IOException {
     super(builder, config);
 
-    this.serializer = this.closer.register(
-        createSerializer(new FixedSchemaVersionWriter()));
+    SchemaVersionWriter versionWriter = config.hasPath(SCHEMA_VERSION_WRITER_TYPE) ?
+        SchemaVersionWriter.Type.valueOf(config.getString(SCHEMA_VERSION_WRITER_TYPE)).getVersionWriter() :
```
chavdar: See above comment about using class names or class aliases. The current approach is non-extensible.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -112,6 +112,8 @@ public JobContext(Properties jobProps, Logger logger) throws Exception {
     Preconditions.checkArgument(jobProps.containsKey(ConfigurationKeys.JOB_NAME_KEY),
         "A job must have a job name specified by job.name");
 
+    logger.info(jobProps.toString());
```
chavdar: Doesn't this have the potential to create a very big message?

gobblin-runtime/src/main/resources/templates/gobblin-kafka.template
```diff
@@ -0,0 +1,28 @@
+gobblin.template.required_attributes="topics,job.name,output"
```
chavdar: Please add a comment with a description to the template what it is about.

gobblin-runtime/src/main/resources/templates/kafka-to-kafka.template
```diff
@@ -0,0 +1,28 @@
+gobblin.template.required_attributes="inputTopics,outputTopic,job.name"
```
chavdar: Ditto.

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -476,4 +477,13 @@ public static Path serializeAsPath(GenericRecord record, boolean includeFieldNam
     }
     return new Path(Joiner.on(Path.SEPARATOR).join(tokens));
   }
+
+  /**
+   * Deserialize a {@link GenericRecord} from a byte buffer. This method is not intended for high performance.
```
chavdar: s/byte buffer/byte array/

gobblin-utility/src/main/java/gobblin/util/ResourceBasedTemplate.java
```diff
@@ -37,26 +37,38 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(SchedulerUtils.class);
   private final Config config;
 
+  public static class TemplateException extends IOException {
```
chavdar: Why does TemplateException extend IOException ?

gobblin-utility/src/main/java/gobblin/util/ResourceBasedTemplate.java
```diff
@@ -85,10 +97,10 @@ public Config getRawTemplateConfig() {
    * @return
    */
   @Override
-  public Config getResolvedConfig(Config userProps) throws IOException {
+  public Config getResolvedConfig(Config userProps) throws TemplateException {
     for (String required : this.getRequiredConfigList()) {
       if (!userProps.hasPath(required)) {
-        throw new IOException(String.format("Missing required property %s for template %s.", required, this.templatePath));
+        throw new TemplateException(String.format("Missing required property %s for template %s.", required, this.templatePath));
```
chavdar: Can there be any recovery from this exception?

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -54,58 +64,87 @@
...
     @Override
     public boolean hasNext() {
-      if (!WikipediaExtractor.this.recordsOfCurrentTitle.isEmpty()) {
-        return true;
-      } else if (WikipediaExtractor.this.requestedTitles.isEmpty()) {
+
+      if (WikipediaExtractor.this.maxRevisionsPulled > -1
+          && this.revisionsPulled >= WikipediaExtractor.this.maxRevisionsPulled) {
+        WikipediaExtractor.this.workUnitState.setActualHighWatermark(new LongWatermark(this.latestPulledRevision));
+        LOG.info(String.format("Pulled max number of records %d, final revision pulled %d.", this.revisionsPulled,
```
chavdar: LOG.info("Pulled max number of records {}, final revision pulled {}.", this.revisionsPulled,
 this.latestPulledRevision)

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -54,58 +64,87 @@
...
+  private final int batchSize;
+  private final long lastRevisionId;
+  private Queue<JsonElement> currentBatch;
+  private final ImmutableMap<String, String> baseQuery;
+  private final WorkUnitState workUnitState;
+  private final int maxRevisionsPulled;
 
   private class WikiResponseReader implements Iterator<JsonElement> {
 
+    private long latestPulledRevision;
```
chavdar: final

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -54,58 +64,87 @@
...
+
     @Override
     public boolean hasNext() {
-      if (!WikipediaExtractor.this.recordsOfCurrentTitle.isEmpty()) {
-        return true;
-      } else if (WikipediaExtractor.this.requestedTitles.isEmpty()) {
+
+      if (WikipediaExtractor.this.maxRevisionsPulled > -1
+          && this.revisionsPulled >= WikipediaExtractor.this.maxRevisionsPulled) {
+        WikipediaExtractor.this.workUnitState.setActualHighWatermark(new LongWatermark(this.latestPulledRevision));
```
chavdar: It is not self-evident why pulling more that WikipediaExtractor.this.maxRevisionsPulled means that we have pulled revision this.latestPulledRevision

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -162,15 +250,21 @@ private boolean isPropPresent(String key, WorkUnitState workUnitState) {
...
-    Queue<JsonElement> retrievedRevisions = new LinkedList<>();
+  private JsonElement performHttpQuery(String rootUrl, Map<String, String> query) throws URISyntaxException, IOException {
+
+    List<String> queryTokens = Lists.newArrayList();
+    for (Map.Entry<String, String> entry : query.entrySet()) {
+      queryTokens.add(entry.getKey() + "=" + entry.getValue());
+    }
+    String actualQuery = Joiner.on("&").skipNulls().join(queryTokens);
+
+    URL actualURL = new URL(rootUrl + (rootUrl.contains("?") ? "&" : "?") + actualQuery);
```
chavdar: This looks dangerous. From the javadocs of the URL class:

> The URL class does not itself encode or decode any URL components according to the escaping mechanism defined in RFC2396

I suggest using org.apache.http.client.utils.URIBuilder or some such for proper escaping.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaSource.java
```diff
@@ -41,13 +50,53 @@
...
+      Iterable<LongWatermark> watermarks =
+          Iterables.transform(entry.getValue(), new Function<WorkUnitState, LongWatermark>() {
+        @Override
+        public LongWatermark apply(WorkUnitState wus) {
+          return wus.getActualHighWatermark(LongWatermark.class);
+        }
+      });
+      watermarks = Iterables.filter(watermarks, Predicates.notNull());
+      List<LongWatermark> watermarkList = Lists.newArrayList(watermarks);
+      Collections.sort(watermarkList, Collections.<LongWatermark>reverseOrder());
```
chavdar: This looks like an expensive way to get the max. Why not use Collections.max() ?

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/SimpleKafkaSource.java
pcadabam: Using the `@Alias` annotation.

https://github.com/linkedin/gobblin/blob/master/gobblin-api/src/main/java/gobblin/util/ClassAliasResolver.java

Used here.
https://github.com/linkedin/gobblin/blob/master/gobblin-audit/src/main/java/gobblin/audit/values/sink/DefaultAuditSinkFactory.java
https://github.com/linkedin/gobblin/blob/master/gobblin-audit/src/main/java/gobblin/audit/values/sink/FsAuditSink.java

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -54,58 +64,87 @@
...
+  private final ImmutableMap<String, String> baseQuery;
+  private final WorkUnitState workUnitState;
+  private final int maxRevisionsPulled;
 
   private class WikiResponseReader implements Iterator<JsonElement> {
 
+    private long latestPulledRevision;
+    private long revisionsPulled = 0;
+
+    public WikiResponseReader(long latestPulledRevision) {
```
chavdar: I think "lastPulledRevision" is more accurate.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -54,58 +64,87 @@
...
-          try {
-            WikipediaExtractor.this.recordsOfCurrentTitle = retrievePageRevisions(currentTitle);
-          } catch (IOException e) {
-            LOG.error("IOException while retrieving revisions for title '" + currentTitle + "'");
-          }
-          if (!WikipediaExtractor.this.recordsOfCurrentTitle.isEmpty()) {
-            return true;
-          }
+        if (this.latestPulledRevision >= WikipediaExtractor.this.lastRevisionId) {
+          return false;
```
chavdar: Why isn't state updated here while it is on lines 112-114?

gobblin-core/src/main/java/gobblin/converter/json/JsonStringToJsonIntermediateConverter.java
ibuenros: I will ask Plaban about this. I really don't understand why all of this logic is there, from what I understand it is just replicating the same json object with some exceptions (like destroying arrays). If it is intentional, I'll just make a new converter.

gobblin-utility/src/main/java/gobblin/util/ResourceBasedTemplate.java
ibuenros: Not for this job, but others can and should still run.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
ibuenros: Because `actualHighWatermark` is set in the constructor. It needs to be updated in lines 112-114 only if we don't actually reach that watermark.

gobblin-runtime/src/main/resources/templates/gobblin-kafka.template
ibuenros: Fixed

gobblin-runtime/src/main/resources/templates/kafka-to-kafka.template
ibuenros: Fixed

ibuenros:
@chavdar addressed comments.

gobblin-core/src/main/java/gobblin/source/extractor/extract/kafka/SimpleKafkaSource.java
```diff
@@ -0,0 +1,37 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+
+/**
+ * A {@link KafkaSource} to use with arbitrary {@link KafkaExtractor}. Specify the extractor to use with key
+ * {@link #EXTRACTOR_TYPE}.
+ */
+public class SimpleKafkaSource<S, D> extends KafkaSource<S, D> {
```
ydai1124: We have another `KafkaSimpleSource` existing in the same package. Can you rename this class?

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -124,24 +170,70 @@ public void remove() {
   }
 
   public WikipediaExtractor(WorkUnitState workUnitState) throws IOException {
-    LOG.info("WorkUnitState received: " + workUnitState);
+    //LOG.info("WorkUnitState received: " + workUnitState);
```
ydai1124: Remove this line.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -124,24 +170,70 @@ public void remove() {
...
+
+    try {
+      Queue<JsonElement> lastRevision = retrievePageRevisions(ImmutableMap.<String, String>builder().putAll(this.baseQuery)
+          .put("rvprop","ids").put("titles",this.requestedTitle).put("rvlimit","1").build());
+      this.lastRevisionId = lastRevision.isEmpty() ? -1 : parseRevision(lastRevision.poll());
+    } catch (URISyntaxException use) {
+      throw new IOException(use);
+    }
+
+    long baseRevision = workUnitState.getWorkunit().getLowWatermark(LongWatermark.class, new Gson()).getValue();
```
ydai1124: Do you need to check `null` for the first run?

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
```diff
@@ -282,12 +388,12 @@ public JsonElement readRecord(@Deprecated JsonElement reuse) throws DataRecordEx
 
   @Override
   public long getExpectedRecordCount() {
-    return this.numRequestedTitles * this.revisionsCnt;
+    return 0;
```
ydai1124: What is the point of this change?

gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
```diff
@@ -144,13 +147,25 @@ public JobScheduler(Properties properties, SchedulerService scheduler)
     this.waitForJobCompletion = Boolean.parseBoolean(
         this.properties.getProperty(ConfigurationKeys.SCHEDULER_WAIT_FOR_JOB_COMPLETION_KEY,
             ConfigurationKeys.DEFAULT_SCHEDULER_WAIT_FOR_JOB_COMPLETION));
+
+    if (this.properties.containsKey(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY)) {
```
ydai1124: We should update our Getting-started/Scheduler documents about this cool feature!

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
ibuenros: Base revision is never `null`, the source sets it to `-1` if it can't find any data.

gobblin-example/src/main/java/gobblin/example/wikipedia/WikipediaExtractor.java
ibuenros: We don't know the expected record count because the revisions are not sequential, the only way to know is to actually consume revisions.

ibuenros:
@ydai1124 addressed comments

ydai1124:
LGTM

chavdar:
LGTM

PR #1259, opened 2016-09-08
Bug fixes in job configuration loading

jsavolainen:


gobblin-cluster/src/main/java/gobblin/cluster/JobConfigurationManager.java
```diff
@@ -76,7 +76,7 @@ protected void startUp() throws Exception {
       if (jobConfigDir.exists()) {
         LOGGER.info("Loading job configurations from " + jobConfigDir);
         Properties properties = new Properties();
-        properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY, jobConfigDir.getAbsolutePath());
+        properties.setProperty(ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY, "file://" + jobConfigDir.getAbsolutePath());
```
chavdar: Let me have a deeper look here. This change is not correct because the idea is to able to poll HDFS directories too.

chavdar: Actually, it looks good. Thanks for catching this.

PR #1262, opened 2016-09-08
[AvroToOrc] Fix valiation job

pcadabam:
The list of destination formats is not provided from the job props. It is available from the config store.

@abti can you review?


abti:
LGTM

PR #1263, opened 2016-09-09
Create publisher to split output by timestamp

jack-moseley:
@ydai1124 please review

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -0,0 +1,39 @@
...
+public class TimestampDataPublisher extends BaseDataPublisher {
+
+  public TimestampDataPublisher(State state) throws IOException {
+    super(state);
+  }
+
+  @Override
+  protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId) {
+    Path outputDir = WriterUtils.getDataPublisherFinalDir(workUnitState, this.numBranches, branchId);
+    outputDir = outputDir.suffix(String.valueOf(System.currentTimeMillis()));
```
ydai1124: Can you add one log message for the final `outputDir`?

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class TimestampDataPublisher extends BaseDataPublisher {
+
+  public TimestampDataPublisher(State state) throws IOException {
+    super(state);
+  }
+
+  @Override
+  protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId) {
+    Path outputDir = WriterUtils.getDataPublisherFinalDir(workUnitState, this.numBranches, branchId);
```
ydai1124: You can call ` Path outputDir = super.getPublisherOutputDir(workUnitState, branchId);`, so it is more clear.

chavdar:
Different work units for the same topic may get published to different directories. That is incorrect.

ydai1124:
LGTM. Please squash your commits into one before merging this PR.

PR #1265, opened 2016-09-11
Replace DatePartitionedDailyAvroSource with configurable partitioning

lbendig:
This PR is a continuation of PR https://github.com/linkedin/gobblin/pull/1196 .

Currently the `DatePartitionedDailyAvroSource` can only read daily partitioned data.
This PR extends this functionality by introducing `DatePartitionedAvroFileSource` in which the partitioning is configurable by the properties as follows:

> date.partitioned.source.partition.prefix
> date.partitioned.source.partition.suffix
> date.partitioned.source.partition.pattern
> date.partitioned.source.partition.granularity

This is the _reader equivalent_ of the time-based writer and partitioner, found in `TimeBasedWriterPartitioner` and `TimePartitionedDataPublisher`.
With this add-on new data sets can be read using different temporal granularity, not just as daily partitioned data.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -29,12 +30,16 @@ import org.apache.hadoop.fs.PathFilter;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.DurationFieldType;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Enums;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 
 import gobblin.configuration.ConfigurationKeys;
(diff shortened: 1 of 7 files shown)


lbendig:
Hi @zliu41, @sahilTakiar, could you please take a look?

chavdar:
@pcadabam can you help review? Right now the biggest concern that I have is that we cannot change the name of the source class. This will break any jobs that use that source already.

At a high level, I see two avenues:
- Keep the name with the new extended functionality (the name will be misleading)
- Add a new class DatePartitionedAvroFileSource and have DatePartitionedDailyAvroSource be a specialization. 

I haven't looked deeply in the code but I am leaning toward the second option.

pcadabam:
@lbendig I agree with @chavdar that we should not rename the source class `DatePartitionedDailyAvroSource `. I would suggest the following,
- Create a new class called `DatePartitionedAvroFileSource ` with all the changes you have.
- Make the current class `DatePartitionedDailyAvroSource` extend `DatePartitionedAvroFileSource`.
- In `DatePartitionedDailyAvroSource` override the `init(state)` , modify the `state` to set `DATE_PARTITIONED_SOURCE_PARTITION_PATTERN=yyyy/MM/dd` and then call `super.init(state)`. The rest should work automatically. If people choose to use the class `DatePartitionedDailyAvroSource` they do not need to set the pattern.

gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
```diff
@@ -43,78 +48,98 @@
...
+  public static final String DATE_PARTITIONED_SOURCE_PARTITION_SUFFIX =
+      DATE_PARTITIONED_SOURCE_PREFIX + ".partition.suffix";
+  public static final String DATE_PARTITIONED_SOURCE_PARTITION_PATTERN =
+      DATE_PARTITIONED_SOURCE_PREFIX + ".partition.pattern";
+  public static final String DATE_PARTITIONED_SOURCE_PARTITION_TIMEZONE =
+      DATE_PARTITIONED_SOURCE_PREFIX + ".partition.timezone";
+  public static final String DEFAULT_DATE_PARTITIONED_SOURCE_PARTITION_TIMEZONE = ConfigurationKeys.PST_TIMEZONE_NAME;
+  public static final String DATE_PARTITIONED_SOURCE_PARTITION_GRANULARITY =
+      DATE_PARTITIONED_SOURCE_PREFIX + ".partition.granularity";
+  public static final DatePartitionType DEFAULT_DATE_PARTITIONED_SOURCE_PARTITION_GRANULARITY =
```
pcadabam: Can you make all the `static final` constants `private`? If you want access them from unit tests you can make them package protected and add a `@VisibleForTesting` annotation.

gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
```diff
@@ -43,78 +48,98 @@
...
+      DATE_PARTITIONED_SOURCE_PREFIX + ".min.watermark.value";
 
   /**
-   * The maximum number of files that this job should process.
-   */
-  private static final String DATE_PARTITIONED_SOURCE_MAX_FILES_PER_JOB =
-      DATE_PARTITIONED_SOURCE_PREFIX + "max.files.per.job";
+  * The maximum number of files that this job should process.
+  */
+  public static final String DATE_PARTITIONED_SOURCE_MAX_FILES_PER_JOB =
```
pcadabam: Same as above. Should be private.

gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
```diff
@@ -43,78 +48,98 @@
...
-public class DatePartitionedDailyAvroSource extends FileBasedSource<Schema, GenericRecord> {
-
+public class DatePartitionedAvroFileSource extends FileBasedSource<Schema, GenericRecord> {
+  
   // Configuration parameters
-  private static final String DATE_PARTITIONED_SOURCE_PREFIX = "date.partitioned.source.";
-
+  private static final String DATE_PARTITIONED_SOURCE_PREFIX = "date.partitioned.source";
+  
+  public static final String DATE_PARTITIONED_SOURCE_PARTITION_PREFIX =
```
pcadabam: Please add javadoc for each of these config options.

gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
```diff
@@ -125,15 +150,25 @@
...
   /**
    * Gobblin calls the {@link Source#getWorkunits(SourceState)} method after creating a {@link Source} object with a
    * blank constructor, so any custom initialization of the object needs to be done here.
    */
   private void init(SourceState state) {
     DateTimeZone.setDefault(DateTimeZone
         .forID(state.getProp(ConfigurationKeys.SOURCE_TIMEZONE, ConfigurationKeys.DEFAULT_SOURCE_TIMEZONE)));
-
+    
+    initDatePartitionFromPattern(state);
```
pcadabam: Have one method called `initDatePartition(State state)` that sets the datePattern either from granularity or pattern key.

gobblin-core/src/main/java/gobblin/writer/partitioner/DatePartitionType.java
```diff
@@ -0,0 +1,91 @@
...
+    this.dateTimeField = dateTimeField;
+    this.dateTimePattern = dateTimePattern;
+  }
+
+  /**
+   * @param pattern full partitioning pattern
+   * @return a DateTimeFieldType corresponding to the smallest temporal unit in the pattern. 
+   * E.g for yyyy/MM/dd {@link DateTimeFieldType#dayOfMonth()}
+   */
+  public static DateTimeFieldType getLowestIntervalUnit(String pattern) {
```
pcadabam: Looks like this is a utility function used for any pattern. Not for this type of `DatePartitionType `. Should this reside in gobblin-utility?

lbendig:
Hi @chavdar and @pcadabam, thanks for your inputs. If renaming is an issue, I'll go with the option you described.

chavdar:
@lbendig is that PR still relevant?

lbendig:
Hi @chavdar @pcadabam, thanks for the review, I addressed your comments. Also added a sample job.

lbendig:
Hi @pcadabam, can you please take a look at the changes?

lbendig:
Hi @pcadabam, please let me know if you need any more changes in this PR.

lbendig:
Thanks @chavdar and @ibuenros !

PR #1269, opened 2016-09-13
[AvroToOrc] Silence hive logging

pcadabam:
- The logging is too verbose when more than 5 datasets are onboarded. Moving most of the DDLs/DMLs to debug as the `HiveJDBCConnector` already logs every statement executed. 
- Set hive log level to `WARN`

@abti can you review?


abti:
LGTM

PR #1279, opened 2016-09-22
Add Access control retention action

pcadabam:
## Changes

- Extend gobblin retention to plug-in `RetentionAction`s
- Implement `AccessControlAction` to set permission/user/group for versions
- Implement `SelectBetweenTimeBasedSelectionPolicy` to select versions created/modified in a time range
- Added unit and integration tests

Design doc - https://docs.google.com/document/d/17k8kbxh-OXe6_HVLn51ke4sdRtVx4zP0Jrni1nrfTFg/edit#

Config for Access control action

<pre>
gobblin.retention {
		accessControl {
	        policies = [restricted,noAccess]
	        restricted {
	             selection {
	                policy.class=SelectionBetweenTimeBasedSelectionPolicy
	                timeBased.maxLookbackTime=3y
	                timeBased.minLookbackTime=2y
	             }
	             mode : 770
	              user : u1
	              group : g1
	          }
	         noAccess {
	             selection {
	                policy.class=SelectionBetweenTimeBasedSelectionPolicy
	                timeBased.maxLookbackTime=5y
	                timeBased.minLookbackTime=3y
	             }
	             mode : 700
	             user : u2
	             group : g2
	          }
	     }
}
</pre>

Main classes to review 

1. `RetentionAction` the new API for retention actions like access control, encryption, deletion etc.
2. `RetentionAction.RetentionActionFactory` the factory API for creating new retention actions. The factory has a `RetentionAction.RetentionActionFactory#canCreateWithConfig(config)` method which decides if the `config` object passed has all the required keys to create a new `RetentionAction`. If yes, `RetentionAction.RetentionActionFactory#createRetentionAction(config)` is called. This pattern allows adding new `RetentionAction`s easily. A `RetentionAction` can decide to use whatever config keys it wants which only its facory knows.
3. `AccessControlAction` mostly business logic for setting permissions, user and group

@ibuenros and @chavdar can you review?


gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
```diff
@@ -0,0 +1,109 @@
...
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigRenderOptions;
+
+import gobblin.data.management.retention.dataset.ConfigurableCleanableDataset;
+import gobblin.data.management.version.DatasetVersion;
+import gobblin.util.ConfigUtils;
+
+
+/**
+ * A wrapped around {@link AccessControlAction} that delegates the {@link #execute(List)} call to multiple embedded
```
ibuenros: typo: `wrapped` -> `wrapper`

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
```diff
@@ -0,0 +1,109 @@
...
+ * {@link AccessControlAction}s
+ * Use {@link MultiAccessControlActionFactory} to create new {@link MultiAccessControlAction}s
+ */
+public class MultiAccessControlAction extends RetentionAction {
```
ibuenros: What is the use case for this?

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/ConfigurableCleanableDataset.java
```diff
@@ -96,6 +106,14 @@
   private final List<VersionFinderAndPolicy<T>> versionFindersAndPolicies;
 
   /**
+   * A set of all available {@link RetentionActionFactory}s
+   */
+  private static final Set<Class<? extends RetentionActionFactory>> RETENTION_ACTION_TYPES;
```
ibuenros: Why this instead of `Alias`?

gobblin-data-management/src/main/java/gobblin/data/management/policy/SelectBetweenTimeBasedPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+import com.typesafe.config.Config;
+
+import gobblin.data.management.version.TimestampedDatasetVersion;
+
+/**
+ * Policy used to select versions in a time range.
+ * An implementation of {@link AbstractTimeBasedSelectionPolicy} which returns versions that are newer than a certain
+ * time <code>lookBackPeriod</code> and older than a certain time <code>minLookBackPeriod</code>
+ */
+public class SelectBetweenTimeBasedPolicy extends AbstractTimeBasedSelectionPolicy {
```
ibuenros: Is there a point to extending the `AbstractTimeBasedSelectionPolicy`. It seems that a single selection policy with `olderThan` and `newerThan` can handle all cases.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/RetentionAction.java
```diff
@@ -0,0 +1,77 @@
...
+  public RetentionAction(Config actionConfig, FileSystem fs) {
+    this.fs = fs;
+  }
+
+  /**
+   * Execute the action on all {@link DatasetVersion}s or a subset of {@link DatasetVersion}s. Each {@link Dataset}
+   * uses the {@link VersionFinder} to find all the {@link DatasetVersion}s and calls this method to perform the necessary
+   * action on those {@link DatasetVersion}s
+   * <p>
+   * <b>Note</b> Any kind of {@link VersionSelectionPolicy} has <b>NOT</b> been applied to the list of {@link DatasetVersion}s
```
pcadabam: Each action can have its own `SelectionPolicy` of several `SelectionPolicies`. This give more flexibility for the the action itself to decide what to do with all the versions.

Another approach would have been that the dataset calls `RetentionAction.getSelectionPolicy`, filters versions using this policy and then calls `RetentionAction.execute` on them. This does not work well with `MultiAccessControlAction` as there it has multiple `SelectionPolicy` and permission combination.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/RetentionAction.java
```diff
@@ -0,0 +1,77 @@
...
+    RetentionAction createRetentionAction(Config config, FileSystem fs);
+
+    /**
+     * Method to check if a {@link RetentionAction} can be created/instantiated with the <code>config</code>.
+     * If the method returns <code>true</code>, {@link #createRetentionAction(Config, FileSystem)} can be called
+     *
+     * @param config to use to create the {@link RetentionAction}
+     * @return true if it is safe to call {@link #createRetentionAction(Config, FileSystem)}, false otherwise
+     */
+    boolean canCreateWithConfig(Config config);
```
pcadabam: I think this makes the factory API more explicit. I am open to throwing and catching exception if you feel strongly about it.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public void execute(List<DatasetVersion> allVersions) throws IOException {
+
+    // Select version on which access control actions need to performed
+    for (DatasetVersion datasetVersion : this.selectionPolicy.listSelectedVersions(allVersions)) {
+
+      // Perform action if it is a FileSystemDatasetVersion
+      if (datasetVersion instanceof FileSystemDatasetVersion) {
+        FileSystemDatasetVersion fsDatasetVersion = (FileSystemDatasetVersion) datasetVersion;
+        for (Path path : fsDatasetVersion.getPaths()) {
```
pcadabam: You mean the dataset root? The permissions need to be set on specific versions. In our current implementation each version maps to only one path.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+
+          if (this.fs.exists(path)) {
+
+            // Update permissions if set in config
+            if (this.permission.isPresent()) {
+              FsPermission fsPermission = new FsPermission(this.permission.get());
+              this.fs.setPermission(path, fsPermission);
+              log.info(String.format("Set permissions for %s to %s", path.toString(), fsPermission));
```
pcadabam: Sure. Changed to debug.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+  @SuppressWarnings("unchecked")
+  private static VersionSelectionPolicy<DatasetVersion> createSelectionPolicy(Config selectionConfig) {
+    try {
+      String selectionPolicyKey =
+          StringUtils.substringAfter(ConfigurableCleanableDataset.SELECTION_POLICY_CLASS_KEY,
+              ConfigurableCleanableDataset.CONFIGURATION_KEY_PREFIX);
+      Preconditions.checkArgument(selectionConfig.hasPath(selectionPolicyKey));
+      String className = selectionConfig.getString(selectionPolicyKey);
+      return (VersionSelectionPolicy<DatasetVersion>) GobblinConstructorUtils.invokeFirstConstructor(
+          Class.forName(className), ImmutableList.<Object> of(selectionConfig),
```
pcadabam: `VersionSelectionPolicy` is an existing inteface. Implementations of `VersionSelectionPolicy` are not annotated with `@Alias`.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
pcadabam: Fixed

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
pcadabam: A dataset can have multiple access control policies for its versions. For instance below I have shown `restrictedAccess` for a few versions and `noAccess` for a few others. For the below case we create two `AccessControlAction`s (Note that an `AccessControlAction` has a `SelectionPolicy` and permissions to set for this selection). 
The `MultiAccessControlAction` is a container for these two real access control actions.

<pre>
accessControl {
           
            ### Read by the MultiAccessControlAction and creates a AccessControlAction for each.
            policies = [restricted,noAccess]
            restricted {
                 selection {
                    policy.class=SelectionBetweenTimeBasedSelectionPolicy
                    timeBased.maxLookbackTime=3y
                    timeBased.minLookbackTime=2y
                 }
                 mode : 770
                  user : u1
                  group : g1
              }
             noAccess {
                 selection {
                    policy.class=SelectionBetweenTimeBasedSelectionPolicy
                    timeBased.maxLookbackTime=5y
                    timeBased.minLookbackTime=3y
                 }
                 mode : 700
                 user : u2
                 group : g2
              }
         }
</pre>

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/ConfigurableCleanableDataset.java
pcadabam: `Alias` is not required for `RetentionActionFactory` because it is not injected by the user.

gobblin-data-management/src/main/java/gobblin/data/management/policy/SelectBetweenTimeBasedPolicy.java
pcadabam: Initially I thought of doing an `INTERSECT` of `SelectBeforeTimeBasedPolicy` and `SelectAfterTimeBasedPolicy` to get `SelectBetweenTimeBasedPolicy`. 
But it turns out that both of them use the same key `selection.timeBased.lookbackTime` to set the `lookbackTime`. For the `SelectBetweenTimeBasedPolicy` we use `selection.timeBased.minLookbackTime` and ``selection.timeBased.maxLookbackTime``

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/RetentionAction.java
pcadabam: If we go the exception approach, we should create a new exception class called say `RetentionActionNotApplicableException`. Only this exception should be caught, other exceptions should be rethrown. 
@chavdar what are your thoughts on this?

chavdar: In general, I'd prefer the canCreateWithConfig() method because we should not be using exception for flow control. In this specific case, I don't understand why we need this method. Are there other ways to instantiate a retention action? The code below seems to ignore the "false" case which is bad.

chavdar: Never mind. I misunderstood the intention of the method. I'd consider rewording the method javadoc. Something along the lines if the specific type of retention access has been specified in the configuration. The way I read it was that the method returns true iff the action supports instantiation with config. Then again, English is not my first language. :-)

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/NoOpAction.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+
+import com.typesafe.config.ConfigFactory;
+
+import gobblin.data.management.version.DatasetVersion;
+
+/**
+ * A {@link RetentionAction} that does nothing
+ */
+public class NoOpAction extends RetentionAction {
```
chavdar: Why is this needed?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+
+
+/**
+ * A {@link RetentionAction} that is used to change the permissions/owner/group of a {@link FileSystemDatasetVersion}
+ */
+@Slf4j
+public class AccessControlAction extends RetentionAction {
+
+  /**
+   * Optional - The permission mode to set on selected versions. E.g 750
```
chavdar: The javadoc for FsPermission say that this can be either and octal number or symbolic. I think it is worth mentioning here.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+        FileSystemDatasetVersion fsDatasetVersion = (FileSystemDatasetVersion) datasetVersion;
+        for (Path path : fsDatasetVersion.getPaths()) {
+
+          if (this.fs.exists(path)) {
+
+            // Update permissions if set in config
+            if (this.permission.isPresent()) {
+              FsPermission fsPermission = new FsPermission(this.permission.get());
+              this.fs.setPermission(path, fsPermission);
+              log.debug(String.format("Set permissions for %s to %s", path.toString(), fsPermission));
```
chavdar: log.debug("Set permissions for {} to {}", path, fsPermission).

Thus you don't need to pay the cost of generating the string if logging is disabled.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            // Update owner and group if set in config
+            if (this.owner.isPresent() || this.group.isPresent()) {
+              this.fs.setOwner(path, this.owner.orNull(), this.group.orNull());
+              log.debug(String.format("Set owner and group for %s to %s:%s", path.toString(), this.owner.orNull(),
```
chavdar: See above.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
chavdar: But we can, right?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
```diff
@@ -0,0 +1,109 @@
...
+   * </pre>
+   * @param actionConfig to use while creating a new {@link MultiAccessControlAction}
+   * @param fs
+   */
+  private MultiAccessControlAction(Config actionConfig, FileSystem fs) {
+    super(actionConfig, fs);
+    this.embeddedAccessControlActions = Lists.newArrayList();
+    for (String policy : ConfigUtils.getStringList(actionConfig, POLICIES_KEY)) {
+      Preconditions.checkArgument(actionConfig.hasPath(policy));
+      embeddedAccessControlActions.add(new AccessControlAction(actionConfig.getConfig(policy), fs));
```
chavdar: It may be worthwhile to check if the config is there and throw an exception with a meaningful message.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
```diff
@@ -0,0 +1,109 @@
...
+
+import gobblin.data.management.retention.dataset.ConfigurableCleanableDataset;
+import gobblin.data.management.version.DatasetVersion;
+import gobblin.util.ConfigUtils;
+
+
+/**
+ * A wrapper around {@link AccessControlAction} that delegates the {@link #execute(List)} call to multiple embedded
+ * {@link AccessControlAction}s
+ * Use {@link MultiAccessControlActionFactory} to create new {@link MultiAccessControlAction}s
```
chavdar: Please document what's the behavior in case the selection policies of the nested actions are overlapping.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
```diff
@@ -261,6 +283,14 @@ public void clean() throws IOException {
       Collection<T> deletableVersions = selectionPolicy.listSelectedVersions(versions);
 
       cleanImpl(deletableVersions);
+
+      List<DatasetVersion> allVersions = Lists.newArrayList();
+      for (T ver : versions) {
+        allVersions.add(ver);
+      }
+      for (RetentionAction retentionAction : versionFinderAndPolicy.getRetentionActions()) {
+        retentionAction.execute(allVersions);
```
chavdar: I think we should add some isolation (i.e. catching of exceptions) between the application of different actions.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -0,0 +1,133 @@
...
+   * where necessary.
+   * <p>
+   * This action only available for {@link FileSystemDatasetVersion}. It simply skips the operation if a different type
+   * of {@link DatasetVersion} is passed.
+   * </p>
+   * {@inheritDoc}
+   * @see gobblin.data.management.retention.action.RetentionAction#execute(java.util.List)
+   */
+  @Override
+  public void execute(List<DatasetVersion> allVersions) throws IOException {
```
chavdar: I suggest adding a help abstract class, SingleVersionRetentionAction or some such, which applies an abstract method executeOnVersion(DatasetVersion version) and adds isolation (i.e. catches exceptions) between the application of the action on different versions.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/MultiVersionCleanableDatasetBase.java
pcadabam: done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: The caller `MultiVersionCleanableDatasetBase.cleanImpl` is not aware of a `RetentionAction`'s selection policy. Hence the caller can not select versions and call execute on each one of them.
 Also a `RetentionAction` implementation is free to perform `execute` as a batch operation on all the versions.
For this case I have extracted the logic into a private method `private void executeOnVersion(DatasetVersion datasetVersion)`

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Added

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/NoOpAction.java
pcadabam: Not needed. Added it while testing. Removed.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Sure. I will add the alias support in a subsequent pr.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/RetentionAction.java
pcadabam: Updated the doc.

pcadabam:
@chavdar and @ibuenros I have addressed all your comments. Please take a final look. Thanks.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
ibuenros: `@Alias` fully supports regular class names, is there a reason not to do it in this PR?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
ibuenros: But why not call `execute` on each action independently instead of using a `MultiAccessControlAction`?

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/ConfigurableCleanableDataset.java
ibuenros: What is the use of the enum? When would it be updated?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Yes the `ClassAliasResolver` does not cache aliasToClass mapping. I will have setup a cache so that we don't have to scan the classpath for every `Dataset`.

gobblin-data-management/src/main/java/gobblin/data/management/policy/SelectBetweenTimeBasedPolicy.java
ibuenros: You could get rid of `AbstractTimeBasedSelectionPolicy`, then `SelectBefore...` is just a `SelectBetween...` between beginning of time and user specified time, similarly with `SelectAfter...`.

gobblin-data-management/src/main/java/gobblin/data/management/retention/dataset/ConfigurableCleanableDataset.java
pcadabam: It lists out all the available `RetentionActionFactory`s. It gets updated when we add a new type of `RetentionAction`. 
It could be auto discoverable through reflections or service loader. We can add the auto discover feature when we need to support user supplied `RetentionAction`

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/MultiAccessControlAction.java
pcadabam: This approach fits naturally with our idea of passing local configs to construct an object. 
1. `ConfigurableCleanableDataset` which constructs actions is unaware of the config key gobblin.retention.accessControl
2. `MultiAccessControlAction` is only aware of the key `policies`. Which lists a bunch of embedded `AccessControlAction`s
3. The individual `AccessControlAction`s are only aware of their local configs. Like `restricted` for restricted `AccessControlAction`. Hence I chose this approach.

gobblin-data-management/src/main/java/gobblin/data/management/policy/SelectBetweenTimeBasedPolicy.java
pcadabam: Discussed offline. Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: nm. I included this change in the same pr.

ibuenros:
LGTM

PR #1280, opened 2016-09-23
Create converter for opening GoldenGate envelope schema

jack-moseley:
@ydai1124 please review

### DIFF
--- a/gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
+++ b/gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
@@ -1,0 +1,86 @@
+package gobblin.converter;
+
+import com.google.common.base.Optional;
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.filter.AvroProjectionConverter;
+import gobblin.converter.filter.AvroSchemaFieldRemover;
+import gobblin.metrics.kafka.KafkaAvroSchemaRegistry;
+import gobblin.metrics.kafka.SchemaRegistryException;
+import gobblin.util.AvroUtils;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import javax.xml.bind.DatatypeConverter;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+/**
+ * A converter for opening GoldenGate envelope schema.
+ */
+public class EnvelopeSchemaConverter extends AvroToAvroConverterBase {
+
+  public static final String PAYLOAD_SCHEMA_ID = "envelope.schema.id.field";
+  public static final String PAYLOAD = "envelope.payload.field";
+
+  private Optional<AvroSchemaFieldRemover> fieldRemover;
+  private KafkaAvroSchemaRegistry registry;
+
+  /**
+   * To remove certain fields from the Avro schema or records of a topic/table, set property
+   * {topic/table name}.remove.fields={comma-separated, fully qualified field names} in workUnit.
+   *
+   * E.g., PageViewEvent.remove.fields=header.memberId,mobileHeader.osVersion
+   */
+  @Override
+  public EnvelopeSchemaConverter init(WorkUnitState workUnit) {
+    if (workUnit.contains(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY)) {
+      String removeFieldsPropName = workUnit.getProp(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY) + AvroProjectionConverter.REMOVE_FIELDS;
+      if (workUnit.contains(removeFieldsPropName)) {
+        this.fieldRemover = Optional.of(new AvroSchemaFieldRemover(workUnit.getProp(removeFieldsPropName)));
+      } else {
+        this.fieldRemover = Optional.absent();
+      }
+    }
+    return this;
+  }
+
+  /**
+   * Do nothing, actual schema must be obtained from records.
+   */
+  @Override
+  public Schema convertSchema(Schema inputSchema, WorkUnitState workUnit) throws SchemaConversionException {
+    return inputSchema;
+  }
+
+  /**
+   * Get actual schema from registry and deserialize payload using it.
+   */
+  @Override
+  public Iterable<GenericRecord> convertRecord(Schema outputSchema, GenericRecord inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+    try {
+      if (registry == null) {
+        registry = new KafkaAvroSchemaRegistry(workUnit.getProperties());
+      }
+      String schemaKey = String.valueOf(inputRecord.get(workUnit.getProp(PAYLOAD_SCHEMA_ID)));
+      Schema payloadSchema = registry.getSchemaByKey(schemaKey);
+      byte[] payload = getPayload(inputRecord, workUnit.getProp(PAYLOAD));
+      GenericRecord outputRecord = AvroUtils.slowDeserializeGenericRecord(payload, payloadSchema);
+      if (this.fieldRemover.isPresent()) {
+        payloadSchema = this.fieldRemover.get().removeFields(payloadSchema);
+      }
+      return new SingleRecordIterable<>(AvroUtils.convertRecordSchema(outputRecord, payloadSchema));
+    } catch (IOException | SchemaRegistryException e) {
+      throw new DataConversionException(e);
+    }
+  }
+
+  /**
+   * Get payload field from GenericRecord and convert to byte array
+   */
+  public byte[] getPayload(GenericRecord inputRecord, String payloadFieldName) {
+    ByteBuffer bb = (ByteBuffer) inputRecord.get(payloadFieldName);
+    String hexString = new String(bb.array());
+    return DatatypeConverter.parseHexBinary(hexString);
+  }
+}
--- a/gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
+++ b/gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
@@ -1,3 +1,15 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
 package gobblin.converter;
 
 import com.google.common.base.Optional;
@@ -8,11 +20,13 @@ import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
 import gobblin.converter.filter.AvroProjectionConverter;
 import gobblin.converter.filter.AvroSchemaFieldRemover;
-import gobblin.metrics.kafka.KafkaAvroSchemaRegistry;
+import gobblin.metrics.kafka.KafkaSchemaRegistry;
+import gobblin.metrics.kafka.KafkaSchemaRegistryFactory;
 import gobblin.metrics.kafka.SchemaRegistryException;
 import gobblin.util.AvroUtils;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.concurrent.ExecutionException;
 import javax.xml.bind.DatatypeConverter;
 import org.apache.avro.Schema;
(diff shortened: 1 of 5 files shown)


gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.xml.bind.DatatypeConverter;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+/**
+ * A converter for opening GoldenGate envelope schema.
+ */
+public class EnvelopeSchemaConverter extends AvroToAvroConverterBase {
+
+  public static final String PAYLOAD_SCHEMA_ID = "envelope.schema.id.field";
```
chavdar: 1. All properties should have namespace to prevent name clashes.
2. Dots.are.not.substitutes.for.spaces. Dots should be used as scope separators. E.g. envelope.schemaIdField, envelope.payloadField.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return this;
+  }
+
+  /**
+   * Do nothing, actual schema must be obtained from records.
+   */
+  @Override
+  public Schema convertSchema(Schema inputSchema, WorkUnitState workUnit) throws SchemaConversionException {
+    return inputSchema;
```
chavdar: Let's discuss this on Monday. This is incorrect. The output schema should correspond to the records produced.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Get actual schema from registry and deserialize payload using it.
+   */
+  @Override
+  public Iterable<GenericRecord> convertRecord(Schema outputSchema, GenericRecord inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+    try {
+      if (registry == null) {
```
chavdar: Please be consistent in prefixing object fields with "this."

chavdar: Why isn't the registry initialized in init() ?

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+        registry = new KafkaAvroSchemaRegistry(workUnit.getProperties());
+      }
+      String schemaKey = String.valueOf(inputRecord.get(workUnit.getProp(PAYLOAD_SCHEMA_ID)));
+      Schema payloadSchema = registry.getSchemaByKey(schemaKey);
+      byte[] payload = getPayload(inputRecord, workUnit.getProp(PAYLOAD));
+      GenericRecord outputRecord = AvroUtils.slowDeserializeGenericRecord(payload, payloadSchema);
```
chavdar: Have a look at the implementation of slowDeserializeGenericRecord. You can do several performance optimizations:
- Reuse the BindaryDecoder object.
- Cache GenericDatumReader by schema.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  /**
+   * Get payload field from GenericRecord and convert to byte array
+   */
+  public byte[] getPayload(GenericRecord inputRecord, String payloadFieldName) {
+    ByteBuffer bb = (ByteBuffer) inputRecord.get(payloadFieldName);
+    String hexString = new String(bb.array());
+    return DatatypeConverter.parseHexBinary(hexString);
```
chavdar: So the payload is hex-encoded? Let's talk on Monday about this.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+   * Get actual schema from registry and deserialize payload using it.
+   */
+  @Override
+  public Iterable<GenericRecord> convertRecord(Schema outputSchema, GenericRecord inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+    try {
+      if (registry == null) {
+        registry = new KafkaAvroSchemaRegistry(workUnit.getProperties());
+      }
+      String schemaKey = String.valueOf(inputRecord.get(workUnit.getProp(PAYLOAD_SCHEMA_ID)));
```
chavdar: Can the names of the fields change? At the very least, we should have defaults.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,86 @@
...
+      throw new DataConversionException(e);
+    }
+  }
+
+  /**
+   * Get payload field from GenericRecord and convert to byte array
+   */
+  public byte[] getPayload(GenericRecord inputRecord, String payloadFieldName) {
+    ByteBuffer bb = (ByteBuffer) inputRecord.get(payloadFieldName);
+    String hexString = new String(bb.array());
```
chavdar: array() may not work, depending on the implementation. See the javadoc for ByteBuffer.array().

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
chavdar: If the payload must be in hex, we should rather implement the bytebuffer to byte[] conversion manually. It should not be too hard and performance will be much better.

chavdar:
There are FindBugs warnings:

Execution failed for task ':gobblin-core:findbugsMain'.
> FindBugs rule violations were found. See the report at: file:///home/travis/build/linkedin/gobblin/build/gobblin-core/reports/findbugs/main.html

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public EnvelopeSchemaConverterForTest init(WorkUnitState workUnit) {
+      if (workUnit.contains(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY)) {
+        String removeFieldsPropName = workUnit.getProp(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY) + AvroProjectionConverter.REMOVE_FIELDS;
+        if (workUnit.contains(removeFieldsPropName)) {
+          this.fieldRemover = Optional.of(new AvroSchemaFieldRemover(workUnit.getProp(removeFieldsPropName)));
+        } else {
+          this.fieldRemover = Optional.absent();
+        }
+      }
+      this.registry = new KafkaAvroSchemaRegistryForTest(workUnit.getProperties());
```
chavdar: If I am not mistaken the only different between EnvelopeSchemaConverter and EnvelopeSchemaConverterForTest is the Kafka schema registry implementation. If that is the case, I suggest that you use dependency injection instead of having to copy/paste the code which is error prone and the copies will go out-of-sync next time we change EnvelopeSchemaConverter.init() .


You can define:
public interface gobblin.metrics.kafka.KafkaSchemaRegistryFactory<K, S> {
   KafkaSchemaRegistry<K, S> create(Properties config);
}

Then have two implementations:
- KafkaAvroSchemaRegistryFactory or KafkaAvroSchemaRegistry.Factory
- KafkaAvroSchemaRegistryForTestFactory or KafkaAvroSchemaRegistryForTest.Factory

You can have a new configuration property for EnvelopeSchemaConverter that will contain the factory class. The default value should be KafkaAvroSchemaRegistryFactory/KafkaAvroSchemaRegistry$Factory but you can override to KafkaAvroSchemaRegistryForTestFactory/KafkaAvroSchemaRegistryForTest$Factory for testing purposes.

Then your init method can lookup the above factory and instantiate it dynamically.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.ExecutionException;
+import javax.xml.bind.DatatypeConverter;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.io.DecoderFactory;
+
+/**
+ * A converter for opening GoldenGate envelope schema.
```
chavdar: Please be a bit more descriptive of what happens. In fact, the current description is misleading as it is not the schema that is converted. You need to describe:
- What the input schema is
- What the pre-conditions for the input schema are, e.g. required fields, etc.
- What  the input record is
- What the output record/s is/are
- What the output schema is

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,111 @@
+package gobblin.converter;
```
ydai1124: Missing license header.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
chavdar: Again, using bb.array() is unsafe. The correct code is something along the lines of 

    byte[] payloadBytes;
    if (bb.hasArray()) {
        payloadBytes = bb.array();
    } else {
       payloadBytes = new byte[bb.remaining()];
       bb.get(payloadBytes);
    }
    String hexString = new String(payloadBytes, Charset.UTF_8);

If we avoid the copying of bytes from payloadBytes to hexString to the decoded result array, it will be even better. Writing a simple function read from the ByteBuffer directly and decode on-the-fly should not be that hard.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.io.DecoderFactory;
+
+/**
+ * A converter for opening GoldenGate envelope schema.
+ */
+public class EnvelopeSchemaConverter extends Converter<Schema, String, GenericRecord, GenericRecord> {
+
+  public static final String PAYLOAD_SCHEMA_ID = "envelope.schemaIdField";
```
ydai1124: Can you change these two keys' names, e.g. "EnvelopeSchemaConverter.payloadSchemaIdField"? It is more clear to see where this key belongs to.

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Get actual schema from registry and deserialize payload using it.
+   */
+  @Override
+  public Iterable<GenericRecord> convertRecord(String outputSchema, GenericRecord inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+    try {
+      String schemaIdField = workUnit.contains(PAYLOAD_SCHEMA_ID) ? workUnit.getProp(PAYLOAD_SCHEMA_ID) : "payloadSchemaId";
```
ydai1124: Please make "payloadSchemaId" and "payload" to be constants, and call them default values. For example, `private static final String DEFAULT.EnvelopeSchemaConverter.payloadSchemaIdField="payloadSchemaId"`

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
```diff
@@ -0,0 +1,96 @@
+package gobblin.converter;
```
ydai1124: License header.

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.DecoderFactory;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+
+public class EnvelopeSchemaConverterTest {
```
ydai1124: Please add `@Test(groups = {"gobblin.converter"})`

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
jack-moseley: fixed

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
jack-moseley: fixed

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
jack-moseley: fixed

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
jack-moseley: fixed

gobblin-core/src/test/java/gobblin/converter/EnvelopeSchemaConverterTest.java
jack-moseley: Added a KafkaSchemaRegistryFactory and using it in the init method

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
jack-moseley: updated description

gobblin-core/src/main/java/gobblin/converter/EnvelopeSchemaConverter.java
jack-moseley: Just using your code for now, I assume the decoding on the fly isn't too important since we should receive payload as bytes instead of hex eventually anyway

PR #1281, opened 2016-09-24
Adding GobblinTestEventBusWriter

chavdar:
Adding some instrumentation to allow us to test end-to-end gobblin jobs.

The idea is to write output to a Guava EventBus and then have some utilities to assert the output.

Changes:
- TestingEventBuses - maintains a global list of testing event buses. This is to simplify the sharing of buses between the writer(s) and the asserter(s).
- GobblinTestEventBusWriter - a writer implementation to write to an EventBus managed by TestingEventBuses
- TestingEventBusAsserter - helper functions to do asserts on the incoming events from TestingEventBuses.
- Unit tests.

@ibuenros @abti can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
@@ -48,9 +48,9 @@ public class TestingEventBusAsserter implements Closeable {
 
   @AllArgsConstructor
   public static class StaticMessage implements Function<TestingEventBuses.Event, String> {
-    private final String errorMessage;
+    private final String message;
     @Override public String apply(Event input) {
-      return this.errorMessage;
+      return this.message;
     }
   }
 
--- a/gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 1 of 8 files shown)


gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
```diff
@@ -0,0 +1,169 @@
...
+   * Blocking assert.  */
+  public <T> void assertNextValuesEq(final Collection<T> expected)
+         throws InterruptedException, TimeoutException {
+    final Set<T> remainingExpectedValues = new HashSet<>(expected);
+    final Predicate<T> checkInRemainingAndRemove = new Predicate<T>() {
+      @Override public boolean apply(@Nonnull T input) {
+        if (! remainingExpectedValues.contains(input)) {
+          return false;
+        }
+        remainingExpectedValues.remove(input);
```
abti: What would happen if two duplicate values are produced / asserted?

chavdar: This specific asset implementation will fail. It is meant to check only for permutations of distinct values.

chavdar:
Any more comments?

gobblin-core/src/main/java/gobblin/writer/test/TestingEventBusAsserter.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public class TestingEventBusAsserter implements Closeable {
+  private final BlockingDeque<TestingEventBuses.Event> _events = new LinkedBlockingDeque<>();
+  private final EventBus _eventBus;
+  private long _defaultTimeoutValue = 1;
+  private TimeUnit _defaultTimeoutUnit = TimeUnit.SECONDS;
+
+  @AllArgsConstructor
+  public static class StaticMessage implements Function<TestingEventBuses.Event, String> {
+    private final String errorMessage;
```
ibuenros: Can you rename this to just `message`?

chavdar: Fixed.

chavdar:
@ibuenros 

I have added a sample usage. See
- SimpleHelloWorldJob.jobconf for a sample configuration for using GobblinTestEventBusWriter
- TestStandardGobblinInstanceLauncher.testSubmitToJobCatalog() for refactoring of an old method to validate the output of the job.

I have also made some small improvements:
- writer.GobblinTestEventBusWriter.eventBusId config to set explicitly the eventbus id because using the writer output path is complicated.
- Added a helper method to HelloWorldSource to simplify validation of the output.

PR #1286, opened 2016-09-27
Fix unit test that may not reset system time

pcadabam:
Some of the tests that alter the current time may not set it back if the test fails.
All test that modify system time belong to the group `SystemTimeTests`. We already have an after group method that resets time if the tests themselves has not done it.

@ibuenros can you review?


ibuenros:
LGTM

PR #1287, opened 2016-09-28
Make Quartz initialization backward compatible in SchedulerService

abti:
Changes to SchedulerService were breaking backward compatibility, because irrespective of whether or not properties at initialization time in SchedulerDaemon had quartz related properties, it was being used to initialize SchedulerFactory; which in turn was making the SchedulerFactory to ignore quartz.properties file in classpath and failing to acquire requisite configs to boot up.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/PropertiesUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/PropertiesUtils.java
@@ -17,6 +17,7 @@ import java.util.Properties;
 
 import org.apache.commons.lang3.StringUtils;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableMap;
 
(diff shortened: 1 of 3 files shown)


abti:
@chavdar @pcadabam Can you please review?

gobblin-utility/src/main/java/gobblin/util/PropertiesUtils.java
```diff
@@ -48,4 +51,23 @@ public static Properties combineProperties(Properties... properties) {
...
+
+  /**
+   * Extract all the keys that start with a <code>prefix</code> in {@link Properties} to a new {@link Properties}
+   * instance.
+   *
+   * @param properties the given {@link Properties} instance
+   * @param prefix of keys to be extracted
+   * @return a {@link Properties} instance
+   */
+  public static Properties extractPropertiesWithPrefix(Properties properties, Optional<String> prefix) {
```
pcadabam: Can we have some basic unit tests for this?

abti: Added some basic tests, there seems to be none pre-existing tests for most of the basic util methods. Will create a ticket for it as well.

PR #1288, opened 2016-09-29
Make AccessControlAction to update permissions only when required

pcadabam:
### Changes
This PR looks like a lot of changes but it is not. It has several unrelated changes in a single PR

#### Major changes

1. Major change is in `AccessControlAction` to not make unecessary `FileSystem` calls. With this change, `AccessControlAction` calls `fs.setPermission` and `fs.setOwner` if the permissions and owner of a path are different from what the user wants. To make this change I had to move around some code because the current `FileSystemDatasetVersion` does not know the `FileStatus` of a `DatasetVersion`. It only knows the `Path`.
2. Currently `DatasetCleanerJob` only supports running retention on datasets that import a tag. The only way to run retention on multiple kinds of datasets  that import different tags is to have different jobs. I have added multi tag support in this pr. The same key `gobblin.retention.tag` is now a comma separated list of tags in the config store. (Internal ask from ops ETL-4855)

#### Minor trivial changes

1. Logging in retention was too verbose and useless. Moved some of them to debug. Added `@ToString` to all the version selection policies for more meaningful logs
2. By default the hive based retention used to delete data as well. Looking at the use cases, it seems like the default should to not delete data. Changed it in `CleanableHiveDataset` and a few unit tests.
3. Chagned the `DEFAULT_MAX_CONCURRENT_DATASETS_CLEANED` to 100 from 1000. Since we have more actions like `AccessControl`, I expect more namenode calls. There is no need to stress the namenode because retention is not a time critical job.
4. Added simulate mode for `AccessControlAction`

@ydai1124 and @ibuenros  can you review?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
@@ -17,6 +17,8 @@ import java.util.List;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
@@ -27,6 +29,7 @@ import com.typesafe.config.Config;
 
 import gobblin.data.management.policy.VersionSelectionPolicy;
 import gobblin.data.management.version.DatasetVersion;
+import gobblin.data.management.version.FileStatusAware;
 import gobblin.data.management.version.FileSystemDatasetVersion;
 import gobblin.util.ConfigUtils;
 
@@ -50,7 +53,7 @@ public class AccessControlAction extends RetentionAction {
    */
   private static final String GROUP_KEY = "group";
 
-  private final Optional<String> permission;
+  private final Optional<FsPermission> permission;
   private final Optional<String> owner;
   private final Optional<String> group;
 
@@ -58,14 +61,14 @@ public class AccessControlAction extends RetentionAction {
   @Getter
   private final VersionSelectionPolicy<DatasetVersion> selectionPolicy;
 
-
   @VisibleForTesting
-  AccessControlAction(Config actionConfig, FileSystem fs) {
-    super(actionConfig, fs);
-    this.permission = Optional.fromNullable(ConfigUtils.getString(actionConfig, MODE_KEY, null));
+  AccessControlAction(Config actionConfig, FileSystem fs, Config jobConfig) {
+    super(actionConfig, fs, jobConfig);
+    this.permission = actionConfig.hasPath(MODE_KEY) ? Optional.of(new FsPermission(actionConfig.getString(MODE_KEY))) : Optional
+            .<FsPermission> absent();
     this.owner = Optional.fromNullable(ConfigUtils.getString(actionConfig, OWNER_KEY, null));
     this.group = Optional.fromNullable(ConfigUtils.getString(actionConfig, GROUP_KEY, null));
-    this.selectionPolicy = createSelectionPolicy(actionConfig);
+    this.selectionPolicy = createSelectionPolicy(actionConfig, jobConfig);
 
   }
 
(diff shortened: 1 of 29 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+  }
 
-          // Update owner and group if set in config
-          if (this.owner.isPresent() || this.group.isPresent()) {
-            this.fs.setOwner(path, this.owner.orNull(), this.group.orNull());
-            log.debug("Set owner and group for {} to {}:{}", path.toString(), this.owner.orNull(),
-                this.group.orNull());
-          }
+  private boolean needsPermissionsUpdate(FileStatus fileStatus) {
+    return this.permission.isPresent() && !new FsPermission(this.permission.get()).equals(fileStatus.getPermission());
```
ibuenros: Can we store the actual `FsPermission` object instead of having to parse every time?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+  }
+
+  private boolean needsOwnerUpdate(FileStatus fileStatus) {
+    return this.owner.isPresent() && this.owner!= null && !StringUtils.equals(owner.get(), fileStatus.getOwner());
```
ibuenros: `this.owner.isPresent()` automatically implies `this.owner != null`

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+  }
+
+  private boolean needsGroupUpdate(FileStatus fileStatus) {
+    return this.group.isPresent() && this.group!= null && !StringUtils.equals(group.get(), fileStatus.getGroup());
```
ibuenros: Same as above

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+
+  private void updatePermissionsAndOwner(Path path) throws IOException {
+    if (this.fs.exists(path)) {
+
+      // Update permissions if set in config
+      if (this.permission.isPresent()) {
+        FsPermission fsPermission = new FsPermission(this.permission.get());
+        if (!this.isSimulateMode) {
+          this.fs.setPermission(path, fsPermission);
+          log.debug("Set permissions for {} to {}", path.toString(), fsPermission);
```
ibuenros: No need to call `path.toString()`, just pass `path` in (saves the `toString()` time)

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+  }
+
+  private void updatePermissionsAndOwner(Path path) throws IOException {
+    if (this.fs.exists(path)) {
+
+      // Update permissions if set in config
+      if (this.permission.isPresent()) {
+        FsPermission fsPermission = new FsPermission(this.permission.get());
+        if (!this.isSimulateMode) {
+          this.fs.setPermission(path, fsPermission);
```
ibuenros: Do we maybe want to catch exception here to make sure change of owner still happens even if `setPermission` failed?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
```diff
@@ -91,23 +93,58 @@ private void executeOnVersion(DatasetVersion datasetVersion) throws IOException
...
+        } else {
+          log.info("Simulating set permissions for {} to {}", path.toString(), fsPermission);
+        }
+      }
+
+      // Update owner and group if set in config
+      if (this.owner.isPresent() || this.group.isPresent()) {
+        if (!this.isSimulateMode) {
+          this.fs.setOwner(path, this.owner.orNull(), this.group.orNull());
+          log.debug("Set owner and group for {} to {}:{}", path.toString(), this.owner.orNull(),
```
ibuenros: Same as above

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/DateTimeDatasetVersionFinder.java
```diff
@@ -59,14 +61,29 @@ public Path globVersionPattern() {
...
     if (timestampedDatasetVersion != null) {
       return new TimestampedDatasetVersion(timestampedDatasetVersion);
     }
     return null;
   }
 
+  // This Method will never be called. It exists because the deprecated super class gobblin.data.management.retention.version.finder.DatasetVersionFinder
+  // requires it. getDatasetVersion(Path pathRelativeToDatasetRoot, FileStatus versionFileStatus) will be called instead
+  @Override
+  public TimestampedDatasetVersion getDatasetVersion(Path pathRelativeToDatasetRoot, Path fullPath) {
```
ibuenros: Why not throw `UnsupportedOperationException` instead?

gobblin-data-management/src/main/java/gobblin/data/management/version/FileStatusTimestampedDatasetVersion.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.hadoop.fs.FileStatus;
+import org.joda.time.DateTime;
+
+import com.google.common.collect.Sets;
+
+/**
+ * A {@link TimestampedDatasetVersion} that is also aware of the {@link FileStatus}s of all its paths.
+ */
+public class FileStatusTimestampedDatasetVersion extends TimestampedDatasetVersion implements FileStatusAware {
```
ibuenros: Why create a new class instead of using the existing one?

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Is there a benefit to doing that? If not I will leave it this way.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/finder/DateTimeDatasetVersionFinder.java
pcadabam: Done

gobblin-data-management/src/main/java/gobblin/data/management/version/FileStatusTimestampedDatasetVersion.java
pcadabam: It will be a cascading change and I want to limit the change to this `VersionFinder`. Making `TimestampedDatasetVersion` implement `FileStatusAware` would mean all the `VersionFinder`s that build a `TimestampedDatasetVersion` needs to be changed.

gobblin-data-management/src/main/java/gobblin/data/management/retention/action/AccessControlAction.java
pcadabam: Fixed.

pcadabam:
@ibuenros any more changes here?

PR #1289, opened 2016-09-29
Lazily load reflections in ClassAliasResolver

pcadabam:
Currently the reflections scan happens in `ClassAliasResolver` even if a job does not use `@Alias`. This change is to use https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom to load reflections only when `ClassAliasResolver` is used.

@chavdar can you review?


pcadabam:
@ibuenros mentioned the desired behavior is already there. The static field `REFLECTIONS` will only get initialized when `ClassAliasResovler` is used.
Discarding this  PR.

PR #1290, opened 2016-09-29
Add blacklist to filter out typesafe incompatible keys

ydai1124:
@pcadabam Fix issue ETL-4958. Can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
@@ -16,9 +16,11 @@ import java.io.IOException;
 import java.io.StringReader;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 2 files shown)


pcadabam:
Can we do at the `DatasetCleanerJob` so that all classes get this change?

In `public DatasetCleanerJob(String id, Props props)` we can use `PropertiesUtils.extractPropertiesWithPrefix(props, "gobblin")`

ydai1124:
@pcadabam DatasetCleanerJob needs more properties than the ones with prefix "gobblin". I don't want to apply this check to a wider area because that needs more investigation to all the configs.

ydai1124:
@pcadabam or @ibuenros  Can you take a look?

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -44,6 +47,12 @@
 public class ConfigUtils {
 
   /**
+   * List of keys that should be excluded when converting to typesafe config.
+   * Usually, it is the key that is both the parent object of a value and a value, which is disallowed by Typesafe.
+   */
+  public static final String GOBBLIN_CONFIG_BLACKLIST_KEYS = "gobblin.config.blacklistKeys";
```
pcadabam: `private` ?

PR #1293, opened 2016-09-30
Data replication complete configuration

tuGithub:


chavdar:
There are FindBugs failures.

> * What went wrong:
> Execution failed for task ':gobblin-data-management:findbugsMain'.
>> FindBugs rule violations were found. See the report at: file:///home/travis/build/linkedin/gobblin/build/gobblin-data-management/reports/findbugs/main.html

chavdar:
You seem to be using Java 8 APIs and the Java 7 compilation fails.

> /home/travis/build/linkedin/gobblin/gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationConfiguration.java:9: error: package
> java.util.stream does not exist
> import java.util.stream.Collectors;
>                       ^

tuGithub:
Somehow, the local build did not catch this

Just removed the unused java8 import, and the local build seems fine (without findBug warning)

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/EndPointCopyPair.java
```diff
@@ -0,0 +1,12 @@
+package gobblin.data.management.copy.replication;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@AllArgsConstructor
+@Data
+public class EndPointCopyPair {
```
chavdar: nit: Can we use CopyRoute instead of CopyPair? May be we can have routes with intermediate hops in the future? It's better to have the names describe the concept (the route to copy data from point A to point B) rather than how it is implemented (a pair of from- and to- points).

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/EndPointCopyPairGenerator.java
```diff
@@ -0,0 +1,15 @@
+package gobblin.data.management.copy.replication;
+
+import java.util.List;
+
+import com.google.common.base.Optional;
+
+public interface EndPointCopyPairGenerator {
```
chavdar: See above comment for CopyPair vs CopyRoute

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/EndPointCopyPairGenerator.java
```diff
@@ -0,0 +1,15 @@
...
+  
+  // implied for pull mode
+  public Optional<EndPointCopyPair> getCopyFrom(ReplicationConfiguration rc, EndPoint copyTo);
```
chavdar: nit: How about "getPullRoute" and getPushRoute ?

tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/EndPointCopyPairGenerator.java
tuGithub: fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/EndPointCopyPair.java
tuGithub: fixed

PR #1301, opened 2016-10-05
Gobblin Google Analysis Source / Google Drive Source

jinhyukchang:
Gobblin Google Analysis Source with Google Drive Source
- Supports Google Analysis unsampled report.
- Unit test passed
- Integration passed
- New CSVFileDownloader that conforms RFC 4180 via leveraging OpenCSV
- New CsvToJsonConverterV2 that supports customization.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
+++ b/gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.converter.csv;
+
+import java.util.Iterator;
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.Converter;
+import gobblin.converter.DataConversionException;
+import gobblin.converter.SchemaConversionException;
+import gobblin.converter.SingleRecordIterable;
+
+/**
+ * CsvToJsonConverterV2 accepts already deserialized (parsed) CSV row, String[], where you can use
+ * @see CsvFileDownloader that conforms with RFC 4180 by leveraging Open CSV.
+ *
+ * Converts CSV to JSON. CSV schema is represented by the form of JsonArray same interface being used by CsvToJonConverter.
+ * Each CSV record is represented by a array of String.
+ *
+ * Example of CSV schema:
+ * [
+  {
+    "columnName": "Day",
+    "comment": "",
+    "isNullable": "true",
+    "dataType": {
+      "type": "string"
+    }
+  },
+  {
+    "columnName": "Pageviews",
+    "comment": "",
+    "isNullable": "true",
+    "dataType": {
+      "type": "long"
+    }
+  }
+]
+ */
+public class CsvToJsonConverterV2 extends Converter<String, JsonArray, String[], JsonObject>  {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CsvToJsonConverterV2.class);
+  private static final JsonParser JSON_PARSER = new JsonParser();
+  private static final String COLUMN_NAME_KEY = "columnName";
+  private static final String JSON_NULL_VAL = "null";
+
+  public static final String CUSTOM_ORDERING = "converter.csv_to_json.custom_order";
+
+  private List<String> customOrder;
+  @Override
+  public Converter<String, JsonArray, String[], JsonObject> init(WorkUnitState workUnit) {
+    super.init(workUnit);
+    customOrder = workUnit.getPropAsList(CUSTOM_ORDERING);
+    if (!customOrder.isEmpty()) {
+      LOG.info("Will use custom order to generate JSON from CSV: " + customOrder);
+    }
+    return this;
+  }
+  @Override
+  public JsonArray convertSchema(String inputSchema, WorkUnitState workUnit) throws SchemaConversionException {
+    Preconditions.checkNotNull(inputSchema, "inputSchema is required.");
+    return JSON_PARSER.parse(inputSchema).getAsJsonArray();
+  }
+
+  /**
+   * Converts CSV (array of String) to JSON.
+   * By default, fields between CSV and JSON are mapped in order bases and it validates if both input and output has same number of fields.
+   *
+   * Customization can be achieved by adding custom order where user can define list of indices of CSV fields correspond to output schema user defined.
+   * Use case of customization (custom order):
+   * In custom order, there are three input parameters generates output.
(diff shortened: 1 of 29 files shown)


jinhyukchang:
@abti Can you take a look? Thanks.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleAnalyticsUnsampledExtractor.java
```diff
@@ -0,0 +1,367 @@
...
+   *
+   */
+  @VisibleForTesting
+  UnsampledReport prepareUnsampledReport(final GoogleDriveFsHelper fsHelper) throws IOException {
+    UnsampledReport createdReport = createUnsampledReports();
+
+    final String fileId = createdReport.getDriveDownloadDetails().getDocumentId();
+    LOG.info("Temporary unsampled report created in Google Drive: " + fileId);
+
+    if (wuState.getPropAsBoolean(DELETE_TEMP_UNSAMPLED_REPORT, true)) {
```
abti: A generic feedback: please change to retrieve all config values from state / config in constructor instead of individual methods. This is to enforce fail fast if config is incorrect and helps in

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleAnalyticsUnsampledExtractor.java
```diff
@@ -0,0 +1,367 @@
...
+                     "Elasped so far: " + stopwatch.elapsed(TimeUnit.SECONDS) + " seconds", e);
+            throw e;
+          }
+
+          if (FAILED_STATUS.equals(response.getStatus())) {
+            //Stop retrying if it explicitly failed from server.
+            throw new NonTransientException("Unsampled report has failed to be generated. " + response);
+          }
+
+          if (!COMPLETED_STATUS.equals(response.getStatus())) {
```
abti: I think it might be a better idea to throw on PENDING status rather than not-COMPLETED status, because in future if the API extends to throw say ABORTED or TIMEDOUT instead of FAILED status, we will un-necessarily continue to poll. Similarly it would be a good idea to ensure that COMPLETED status is returned instead of implicitly assuming it that it is completed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleDriveFileSystem.java
```diff
@@ -0,0 +1,311 @@
...
+   */
+  public static String toFileId(Path p) {
+    if (p.isRoot()) {
+      return "";
+    }
+    final String format = "%s" + Path.SEPARATOR + "%s";
+    if (p.getParent() != null && StringUtils.isEmpty(p.getParent().getName())) {
+      return p.getName();
+    }
+    return String.format(format, toFileId(p.getParent()), p.getName());
```
abti: What if there are multiple / in the id, you probably need to recursively call toFileId()

gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
```diff
@@ -0,0 +1,170 @@
...
+import com.google.gson.JsonParser;
+
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.Converter;
+import gobblin.converter.DataConversionException;
+import gobblin.converter.SchemaConversionException;
+import gobblin.converter.SingleRecordIterable;
+
+/**
+ * (Current CsvToJonConverter parses CSV in converter where it simply delimits ",".
```
jinhyukchang: Will remove it.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleAnalyticsUnsampledExtractor.java
jinhyukchang: Agreed. Will make change.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleDriveFileSystem.java
jinhyukchang: Actually, the behavior of "org.apache.hadoop.fs.Path" is to only generates parent once even there are multiples.
Will add unit test case on multiple "/", to validate the result.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleCommon.java
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Provides Google credential either via JSON file or via P12 file.
+   * Note that it distinguishes file by extension (e.g: ".json")
+   *
+   * @param state
+   * @param privateKeyFile
+   * @return
+   */
+  public static Credential newSourceCredential(State state) {
```
jinhyukchang: I see. Actually, it was my first approach to use explicit parameters but switched it state because it became too many. Let me refactor to use explicit parameters and make it neat as well.

jinhyukchang:
@abti I've just updated the code incorporating the comments. Could you take a look?

jinhyukchang:
@abti Could you take a look when you have a chance? Thanks.

abti:
@jinhyukchang Sorry for the delay, been busy with a few support issues - will look at it today / tomorrow.

gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+
+    return outputRecord;
+  }
+
+  @VisibleForTesting
+  JsonObject createOutput(JsonArray outputSchema, String[] inputRecord, List<String> customOrder) {
+
+    Preconditions.checkArgument(outputSchema.size() == customOrder.size(), "# of columns mismatch. Input "
```
abti: As per the example: customOrder can be lesser than outputSchema size. Is my understanding correct? If so this check should be changed

gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
```diff
@@ -0,0 +1,169 @@
...
+    JsonObject outputRecord = new JsonObject();
+    Iterator<JsonElement> outputSchemaIterator = outputSchema.iterator();
+    Iterator<String> customOrderIterator = customOrder.iterator();
+
+    while(outputSchemaIterator.hasNext() && customOrderIterator.hasNext()) {
+      String key = outputSchemaIterator.next().getAsJsonObject().get(COLUMN_NAME_KEY).getAsString();
+      int i = Integer.parseInt(customOrderIterator.next());
+      Preconditions.checkArgument(i < inputRecord.length, "Index out of bound detected in customer order. Index: " + i + " , # of CSV columns: " + inputRecord.length);
+      if (i < 0 || null == inputRecord[i] || JSON_NULL_VAL.equalsIgnoreCase(inputRecord[i])) {
+        outputRecord.add(key, JsonNull.INSTANCE);
```
abti: Since you iterate over and add in same order as outputSchema, aren't you maintaining the same order by with mismatched inputvalue rather? Lets discuss this offline, maybe I am not understanding this correctly.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/GoogleDriveFileSystem.java
abti: oh ok. thanks

gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
jinhyukchang: Synced offline. Will add more documentation to clarify the usage of custom order.

gobblin-core/src/main/java/gobblin/converter/csv/CsvToJsonConverterV2.java
jinhyukchang: Synced offline. Will add more documentation to clarify the usage of custom order.

jinhyukchang:
@abti, just added example pull file and update the java doc comment on the converter.
Will work on pulling from master branch and resolve conflict so that it can be merged.

jinhyukchang:
@abti Just resolved conflict and squashed into 1 commit. Let me know if you need anything else. Thanks.

abti:
Thanks so much @jinhyukchang for the changes / answers. I will merge as soon as travis finishes.

PR #1303, opened 2016-10-06
Added policy in distcp-ng Hive copy to force delete physical data on target if its not managed by the target Hive table or partition

abti:


abti:
@ibuenros Can you please review?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -90,6 +90,11 @@
   public static final String EXISTING_ENTITY_POLICY_KEY =
       HiveDatasetFinder.HIVE_DATASET_PREFIX + ".existing.entity.conflict.policy";
   public static final String DEFAULT_EXISTING_ENTITY_POLICY = ExistingEntityPolicy.ABORT.name();
+
+  public static final String MANAGED_DATA_POLICY_KEY =
+      HiveDatasetFinder.HIVE_DATASET_PREFIX + ".managed.data.conflict.policy";
```
ibuenros: Should this be `unmanaged` instead?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -755,14 +773,26 @@ protected static DiffPathSet fullPathDiff(HiveLocationDescriptor sourceLocation,
...
+          partition.isPresent() ? partition.get().getCompleteName() : helper.getDataset().getTable().getCompleteName(),
           Arrays.toString(desiredTargetExistingPaths.keySet().toArray())));
     }
+    // Unless, the policy requires us to delete such un-managed files - in which case: we will add the leftover files
+    // to the deletion list.
+    else if (desiredTargetExistingPaths.size() > 0) {
+      log.warn("Un-managed files detected in target file system, however deleting them because of the policy: " +
+        ManagedDataPolicy.DELETE_UNMANAGED_DATA);
+      for (Path delete : desiredTargetExistingPaths.keySet()) {
+        log.warn("Un-managed file to be deleted: " + delete);
```
ibuenros: I would prefer not to log every file in a separate line. Can we do a single line with all files to be deleted?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
abti: Yes, unmanaged would be more crisp, will change it.

Issue #1307, opened 2016-10-09
Gobblin Docker Integration


sahilTakiar:
Create some basic Gobblin integration with Docker. Main idea is to run the Gobblin standalone service inside a Docker container. Created a new entry in the docs page called `Docker Integration` which explains everything in depth.

For Gobblin users familiar with Docker, should provide a good way to get Gobblin up and running quickly.

Docker images will be hosted on Docker Hub, I created a Gobblin group on Docker Hub to host the images https://hub.docker.com/u/gobblin/dashboard/ (I've already built and uploaded the images from the Dockerfile present in this PR). Any committers that want to be added to the group, just create an account on Docker Hub, send me your username and I will add you as an owner.

Related Issues:
#1006
#905
#900

abti:
Great! Thanks @sahilTakiar My username is same as github at docker ( @abti ). Please add me too. 
I don't see the PR that you talk about in the description. Are you yet to send it?

sahilTakiar:
@abti added you to the repos on Docker Hub

abti:
This was merged here: https://github.com/linkedin/gobblin/pull/1308

PR #1308, opened 2016-10-09
(#1307) Gobblin integration with Docker for Gobblin standalone

sahilTakiar:
* Created a new properties file for launching jobs via the CliJobLauncher
* All Dockerfiles are under the gobblin-docker folder
* The gobblin-base files add libraries necessary for Gobblin to run
* Each entry in gobblin-base is for a different OS
* Only Ubuntu based images have been added so far
* gobblin-distribution contains files for each Gobblin release
* gobblin-standalone contains file for running the Gobblin SchedulerDaemon
* gobblin-wikipedia contains files for running Gobblin Wikipedia

I built all the images locally and uploaded them to Docker Hub, the
following link has all the repositories:

https://hub.docker.com/u/gobblin/dashboard/

Any committer who wants to become an admin of the Docker Hub Gobblin
repositories, just send me your Docker Hub user name.

Each folder under gobblin-docker is a different repository on Docker Hub.
Repositories are laid out in this fashion to make it easier to add new
Dockerfiles. Adding support for new Gobblin versions and different base
OSs should be easy too

### DIFF
--- a/gobblin-docker/gobblin-wikipedia/ubuntu-gobblin-0.7.0/Dockerfile
+++ b/gobblin-docker/gobblin-wikipedia/ubuntu-gobblin-0.7.0/Dockerfile
@@ -1,0 +1,16 @@
+FROM gobblin/gobblin-distributions:ubuntu-gobblin-0.7.0
+
+ENV RELEASE_VERSION "0.7.0"
+ENV GOBBLIN_WORK_DIR /home/gobblin/work-dir
+
+RUN echo "Downloading Gobblin Wikipedia Configuration Files"
+RUN mkdir -p /etc/opt/gobblin/job-conf
+
+# Downloading wikipedia.pull
+RUN curl -L --progress-bar https://raw.githubusercontent.com/linkedin/gobblin/gobblin_$RELEASE_VERSION/gobblin-example/src/main/resources/wikipedia.pull -o /etc/opt/gobblin/job-conf/wikipedia.pull
+
+# Downloading gobblin-cli.properties since its not in release 0.7.0
+RUN curl -L --progress-bar https://raw.githubusercontent.com/linkedin/gobblin/gobblin-docker/conf/gobblin-cli.properties -o /opt/gobblin/gobblin-dist/conf/gobblin-cli.properties
+
+# Start Gobblin Wikipedia Example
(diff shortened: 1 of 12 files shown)


gobblin-docker/gobblin-wikipedia/ubuntu-gobblin-0.7.0/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM gobblin/gobblin-distributions:ubuntu-gobblin-0.7.0
+
+ENV GOBBLIN_WORK_DIR /home/gobblin/work-dir
+
+RUN echo "Downloading Gobblin Wikipedia Configuration Files"
+RUN mkdir -p /etc/opt/gobblin/job-conf
+
+# Downloading wikipedia.pull
+RUN curl -L --progress-bar https://raw.githubusercontent.com/linkedin/gobblin/master/gobblin-example/src/main/resources/wikipedia.pull -o /etc/opt/gobblin/job-conf/wikipedia.pull
```
abti: wikipedia.pull might change between different Gobblin versions. Isn't it bundled in the gobblin-distribution tarball? If so, we can just copy it from there to be sure that it would work.

chavdar: It's best to use the release tag: https://raw.githubusercontent.com/linkedin/gobblin/gobblin_0.7.0/gobblin-example/src/main/resources/wikipedia.pull

sahilTakiar: It's not in the distribution unfortunately, although it probably should be. Created #1316 to track this.

Updated to use the release tag. I can't use it for the gobblin-cli.properties file unfortunately since it didn't exist in those releases

sahilTakiar:
Updated, comments addressed

gobblin-docker/gobblin-wikipedia/ubuntu-gobblin-0.7.0/Dockerfile
abti: Thanks for the changes Sahil. Can you also please create a placeholder github issue for adding release tag to gobblin-cli.properties download, it might be a while before we make a 0.9.0 release and we will forget to update it by then

PR #1310, opened 2016-10-11
Restored max mappers option for Gobblin MR.

ibuenros:


ibuenros:
@ydai1124 can you review?

gobblin-runtime/src/test/java/org/apache/hadoop/fs/FileSystemTestUtils.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.io.IOException;
+import java.net.URI;
+
+import org.apache.hadoop.conf.Configuration;
+
+
+public class FileSystemTestUtils {
+
+  public static void addFileSystemForTest(URI uri, Configuration conf, FileSystem fs) throws IOException {
```
ydai1124: Why is this necessary as a single line method?

ibuenros: Because the method is protected, so it needs to be in the correct package (`org.apache.hadoop.fs`)

ydai1124: I see.

Issue #1267, opened 2016-09-12
gobblin.source.extractor.filebased.FileBasedSource  191 - Not able to fetch the filename/file modified time to Failed to get modified timestamp due to error Cannot open a channel to SFTP server will not pull any files


abraham0477:
Hi, 
I have gobblin installed on a standalone linux server and my ingestion job is basically connecting to a remote server using sftp extracts the files from the remote server, convert them to avro and publish them out to S3. 

I already tested the job and the test went successful, my issue here is I am seeing an error thrown as "gobblin.source.extractor.filebased.FileBasedSource  191 - Not able to fetch the filename/file modified time to Failed to get modified timestamp for file at path filename due to error Cannot open a channel to SFTP server will not pull any files"

I want to mention out here that the error is throwing after processed a certain amount of files, let's say if I have 100 files, the ingestion processed stops after processed and ingested 10 files.

Here is the log file output, and I appreciate any feedback here to fix this issue?

016-09-12 18:26:33 UTC INFO  [TaskExecutor-0] gobblin.source.extractor.extract.sftp.SftpFsHelper$SftpGetMonitor  317 - Transfer finished 1 src: /home/clduser/poller/archive/mibs-3dfd4c3a-d64e-42d0-99fb-0608ff648579-1473303473.csv.gz dest: ?? in 0 at 0.00
2016-09-12 18:26:33 UTC INFO  [TaskExecutor-1] gobblin.publisher.TaskPublisher  43 - All components finished successfully, checking quality tests
2016-09-12 18:26:33 UTC INFO  [TaskExecutor-1] gobblin.publisher.TaskPublisher  45 - All required test passed for this task passed.
2016-09-12 18:26:33 UTC INFO  [TaskExecutor-1] gobblin.publisher.TaskPublisher  47 - Cleanup for task publisher executed successfully.
2016-09-12 18:26:33 UTC INFO  [TaskExecutor-1] gobblin.runtime.Fork  261 - Committing data for fork 0 of task task_aphealth_ingestion_98.119.181.163_1473704782375_8
[...]

PR #1311, opened 2016-10-12
Close sftp connection after getting mtime

sidarthakarna:
Fix for #1267. 
Sftp connection is not being closed while getting modification time. So, only 10 files can be pulled from sftp source as sshd supports 10 concurrent channels by default. Fix is added to close the connection once we get modification time.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
@@ -451,14 +451,20 @@ public class SftpFsHelper implements TimestampAwareFileBasedHelper {
 
   @Override
   public long getFileMTime(String filePath) throws FileBasedHelperException {
-    try {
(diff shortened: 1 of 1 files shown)


ibuenros:
@pcadabam can you take a look at this one? It is quick.

gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
```diff
@@ -453,7 +453,9 @@ public boolean setIdentity(String privateKey, JSch jsch) {
   public long getFileMTime(String filePath) throws FileBasedHelperException {
     try {
       ChannelSftp channelSftp = getSftpChannel();
-      return channelSftp.lstat(filePath).getMTime();
+      int modificationTime = channelSftp.lstat(filePath).getMTime();
```
chavdar: Good catch!

Could you put the channelSftp.disconnect() in a try-finally block?

sidarthakarna:
Thanks for the review @chavdar. Updated the code as you suggested.

PR #1313, opened 2016-10-13
Basic embedded Gobblin implementation.

ibuenros:
Implements a basic version of embedded Gobblin.

Future PRs will add template aware embedded Gobbblin that is easier to configure (i.e. don't require magically knowing which configurations need to be set).


ibuenros:
@pcadabam @chavdar can you review?

chavdar:
There are findbugs failures:

> Execution failed for task ':gobblin-runtime:findbugsMain'.

PR #1318, opened 2016-10-17
Various changes to job execution to appropriately close services and …

ibuenros:
…executors.

There are three main changes here:
* Make JobLauncherExecutionDriver a `FutureTask` (which also makes it a `RunnableFuture`) and a `ListenableFuture`, and it not longer implements `Service`. Futures have a concrete end, which is appropriate for the job driver. `FutureTask` additionally offers a concrete start, which was the other requirement for the job driver. Finally, `ListenableFuture` allows adding callbacks.
* Make `JobCatalog` a `Service` so it can be shut down allowing it to clean up whatever threads it spawned. Also extended `JobCatalogBase` to handle the service part of the catalog, and implemented `MutableJobCatalogBase` that does the same for `MutableJobCatalog`. Most `JobCatalog` implementations were simplified taking advantage of the base job catalogs.
* Make `CallbackDispatcher` closeable to allow it to shut down its `ExecutorService`. All classes that use `CallbackDispatcher` were updated accordingly.

Additionally, some small, unrelated changes to `EmbeddedGobblin` that make it easier to run from a command line.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
@@ -31,7 +31,7 @@ import lombok.Getter;
  * A catalog of all the {@link JobSpec}s a Gobblin instance is currently aware of.
  */
 @Alpha
-public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable, Service {
+public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable {
   /** Returns an immutable {@link Collection} of {@link JobSpec}s that are known to the catalog. */
   Collection<JobSpec> getJobs();
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 
 import com.codahale.metrics.Gauge;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.Service;
 
 import gobblin.annotation.Alpha;
 import gobblin.instrumented.GobblinMetricsKeys;
@@ -30,7 +31,7 @@ import lombok.Getter;
  * A catalog of all the {@link JobSpec}s a Gobblin instance is currently aware of.
  */
 @Alpha
-public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable {
+public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable, Service {
   /** Returns an immutable {@link Collection} of {@link JobSpec}s that are known to the catalog. */
   Collection<JobSpec> getJobs();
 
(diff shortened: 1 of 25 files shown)


ibuenros:
@chavdar can you review?

gobblin-runtime/src/main/java/gobblin/runtime/api/JobCatalog.java
```diff
@@ -30,7 +31,7 @@
  * A catalog of all the {@link JobSpec}s a Gobblin instance is currently aware of.
  */
 @Alpha
-public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable {
+public interface JobCatalog extends JobCatalogListenersContainer, Instrumentable, Service {
```
chavdar: As discussed offline, let's not force every catalog be a service. StandardGobblinInstanceDriver takes care of dynamically checking for components that are services and starting them automatically.

gobblin-runtime/src/main/java/gobblin/runtime/api/JobExecutionDriver.java
```diff
@@ -42,9 +45,9 @@
   //void stopForcefully();
 
   /** {@inheritDoc} */
-  @Override JobExecutionResult get() throws InterruptedException;
+  @Override JobExecutionResult get() throws InterruptedException, ExecutionException;
```
chavdar: Is adding ExecutionException necessary? The idea is to make the execution exceptions accessible through the errorCause in JobExecutionResult.

gobblin-runtime/src/main/java/gobblin/runtime/instance/DefaultGobblinInstanceDriverImpl.java
```diff
@@ -188,7 +189,7 @@ public void run() {
          JobExecutionDriver driver = _jobLauncher.launchJob(_jobSpec);
          _callbacksDispatcher.onJobLaunch(driver);
          driver.registerStateListener(new JobStateTracker());
-         driver.startAsync();
+        new Thread(driver).start();
```
chavdar: Can you use gobblin.util.ExecutorsUtils.newThreadFactory(Optional<Logger>, Optional<String>) so that we can name the thread for debugging purposes and we log uncaught exceptions?

gobblin-runtime/src/main/java/gobblin/runtime/job_catalog/CachingJobCatalog.java
```diff
@@ -61,6 +63,30 @@ public JobSpec getJobSpec(URI uri) throws JobSpecNotFoundException {
     }
   }
 
+  @Override
+  protected void startUp() {
+    _cache.startAsync();
```
chavdar: The CachingJobCatalog does not own the lifecycle of the cached object so it should not be responsible for starting/stopping it.

ibuenros: do you mean the fallback object? The cached object is a private field created by the `CachingJobCatalog`, so the `CachingJobCatalog` is the only object that can be responsible for starting/stopping it.

chavdar: Yeah, I meant the fallback.

ibuenros:
Addressed comments, merging.

PR #1324, opened 2016-10-18
Fix AvroUtils.getFieldHelper() to support map fields

aditya1105:
Currently AvroUtils.getFieldHelper() assumes that object passed to it is of type Record. Method recursively calls itself based on the field location passed to it as list. If a record contains map, this method fails with ClassCastException: java.util.HashMap cannot be cast to org.apache.avro.generic.GenericData$Record. 

This fix enables support for passing map object to AvroUtils.getFieldHelper(). Also added a test case to test for the same.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/sftp/SftpFsHelper.java
@@ -13,15 +13,24 @@
 package gobblin.source.extractor.extract.sftp;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
 
+import lombok.extern.slf4j.Slf4j;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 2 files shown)


gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -146,12 +148,31 @@
     }
 
     if ((field + 1) == pathList.size()) {
+      if (data instanceof java.util.HashMap) {
```
pcadabam: `data instanceof Map` instead of `data instanceof HashMap`

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -146,12 +148,31 @@
...
+        return getObjectFromMap((HashMap) data, pathList.get(field));
```
pcadabam: Cast it to a `Map` instead of `HashMap`. Since the cast is a raw `Map` you may need to add `@SuppressWarnings("rawtypes")` to the method.

pcadabam:
@aditya1105 It seems like you do not have merge privileges. Let me know when you have pushed your changes. I will merge it for you.

abti:
This PR seems to be empty. What happened here? :)

aditya1105:
I am create a new pull request.

PR #1325, opened 2016-10-18
Enhanced RecursiveCopyableDataset to match all options in Hadoop dist…

ibuenros:
…cp and handle files existing in target.

Most of the changes are a fairly long test.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
@@ -12,22 +12,32 @@
 
 package gobblin.data.management.copy;
 
+import gobblin.commit.CommitStep;
+import gobblin.data.management.copy.entities.PrePublishStep;
 import gobblin.data.management.dataset.DatasetUtils;
 import gobblin.dataset.FileSystemDataset;
 import gobblin.util.PathUtils;
 import gobblin.util.FileListUtils;
+import gobblin.util.commit.DeleteFileCommitStep;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 
 
 /**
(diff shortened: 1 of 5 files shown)


ibuenros:
@tuGithub can you review? Thanks.

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -51,29 +73,90 @@ public RecursiveCopyableDataset(final FileSystem fs, Path rootPath, Properties p
...
 
     Path nonGlobSearchPath = PathUtils.deepestNonGlobPath(this.glob);
+    Path targetPath = new Path(configuration.getPublishDir(), PathUtils.relativizePath(this.rootPath, nonGlobSearchPath));
 
-    List<FileStatus> files = FileListUtils.listFilesRecursively(this.fs, this.rootPath, this.pathFilter);
+    Map<Path, FileStatus> filesInSource = createPathMap(getFilesAtPath(this.fs, this.rootPath, this.pathFilter), this.rootPath);
+    Map<Path, FileStatus> filesInTarget = createPathMap(getFilesAtPath(targetFs, targetPath, this.pathFilter), targetPath);
 
-    List<CopyableFile> copyableFiles = Lists.newArrayList();
+    Set<Path> inSourceNotInTarget = Sets.difference(filesInSource.keySet(), filesInTarget.keySet());
```
tuGithub: Create 3 sets in this way will loop through filesInSource and filesInTarget 3 times? Can we do this in one loop?

iterate filesInSource
    if In target, put to set3 and remove that entry in filesInTarget
    if NOT in target, put to set2

the left over of filesInTarget is set1

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -51,29 +73,90 @@ public RecursiveCopyableDataset(final FileSystem fs, Path rootPath, Properties p
     this.pathFilter = DatasetUtils.instantiatePathFilter(properties);
     this.copyableFileFilter = DatasetUtils.instantiateCopyableFileFilter(properties);
     this.glob = glob;
+
+    this.update = Boolean.parseBoolean(properties.getProperty(UPDATE_KEY));
```
tuGithub: should we add the default value if not specified?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
```diff
@@ -51,29 +73,90 @@ public RecursiveCopyableDataset(final FileSystem fs, Path rootPath, Properties p
...
+
+    if (!toDelete.isEmpty()) {
+      List<FileStatus> statusesToDelete = Lists.newArrayList();
+      for (Path path : toDelete) {
+        statusesToDelete.add(filesInTarget.get(path));
+      }
+
+      CommitStep step =
+          new DeleteFileCommitStep(targetFs, statusesToDelete, this.properties,
+              this.deleteEmptyDirectories ? Optional.of(targetPath) : Optional.<Path>absent());
```
tuGithub: how is deleteEmptyDirectories been used? I thought somewhere, we need to check that FileStatus is empty dir, right?

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
ibuenros: `Boolean.parseBoolean` always defaults to false if the argument is `null`

gobblin-data-management/src/main/java/gobblin/data/management/copy/RecursiveCopyableDataset.java
tuGithub: why need to make filesInTarget as immutable? I thought we only iterate filesInSource

ibuenros: Because we later use those same maps to get the appropriate file statuses for either creating copy work units, determining whether a collision file is the same or not, and scheduling deletes.

tuGithub: how about this:

Map<Path, FileStatus> toDelete = new HashMap<Path, FileStatus>();

for (Path path: filesInSource){
	FileStatus statusInTarget = filesInTarget.remove(path);

	// both exists
	if(statusInTarget!=null){
		if (!sameFile(filesInSource.get(path), statusInTarget)) {
         toCopy.add(path);
         toDelete.put(path, statusInTarget);
       }
	}
	else{
		toCopy.add(path);
		inSourceNotInTarget.add(path); // may not need it anymore
	}

	
}

if(this.delete){
	// left over need to be deleted
	toDelete.addAll(filesInTarget);
}

List<FileStatus> statusesToDelete = toDelete.values();

ibuenros:
@tuGithub addressed comments

PR #1326, opened 2016-10-19
Fix AvroUtils.getFieldHelper() to support map fields. Minor changes to cast hashMap to Map

aditya1105:
Currently AvroUtils.getFieldHelper() assumes that object passed to it is of type Record. Method recursively calls itself based on the field location passed to it as list. If a record contains map, this method fails with ClassCastException: java.util.HashMap cannot be cast to org.apache.avro.generic.GenericData$Record.

This fix enables support for passing map object to AvroUtils.getFieldHelper(). Also added a test case to test for the same.


pcadabam:
Copy of https://github.com/linkedin/gobblin/pull/1324
Already reviewed. Merging.

Issue #1322, opened 2016-10-18
JdbcWriter sets wrong property


chosh0615:
I have an issue when retrying JdbcWriter.

In the JdbcWriter, it appends FORK_BRANCH_ID_KEY property, but nowhere in the source code it is read as a list. I think it should be modified to use setProp method instead of appendToListProp method.

`this.state.appendToListProp(ConfigurationKeys.FORK_BRANCH_ID_KEY, Integer.toString(builder.branch));`

If I'm correct, I can make a pull request.

abti:
@jinhyukchang Do you have a quick answer for this?

jinhyukchang:
@chosh0615 Thanks for pointing it out. Could you make a pull request?

abti:
PR merged, this issue is fixed.

PR #1327, opened 2016-10-19
Edited JdbcWriter to use setProp rather than appendToListProp method …

chosh0615:
Changed JdbcWriter to use setProp rather than appendToListProp when setting the property FORK_BRANCH_ID_KEY.
Please refer to the issue #1322 .


abti:
@chosh0615 Thanks for the PR

PR #1328, opened 2016-10-19
Data replication in push mode

tuGithub:
Use Properties to control whether that job is a pull or push mode
Only pick up dataset which mode is in sync with Properties config
For push mode, only pick up the copyTo cluster which is the same as the writer.fs.uri


tuGithub:
@chavdar , @ibuenros , please review the pr when you have time

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedMultiDatasets.java
```diff
@@ -32,54 +34,101 @@
...
+    // PUSH mode
+    CopyRouteGenerator cpGen = rc.getCopyRouteGenerator();
+    List<EndPoint> replicas = rc.getReplicas();
+    List<EndPoint> pushCandidates = new ArrayList<EndPoint>(replicas);
+    pushCandidates.add(rc.getSource());
+    
+    for(EndPoint pushFrom: pushCandidates){
+      if(needGenerateCopyEntity(pushFrom, executionClusterURI)){
+        Optional<List<CopyRoute>> copyRoutes = cpGen.getPushRoutes(rc, pushFrom);
+        if(!copyRoutes.isPresent()) return;
```
chavdar: Please add a logging message that no routes have been found for the dataset.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedMultiDatasets.java
```diff
@@ -32,54 +34,101 @@
...
+              this.datasets.add(new ConfigBasedDataset(rc, this.props, cr));
+            }
+          }
+        }// inner for loops ends
+      }
+    }// outer for loop ends
+  }
+
+  private void generateDatasetInPullMode(ReplicationConfiguration rc, URI executionClusterURI){
+    if(rc.getCopyMode()== ReplicationCopyMode.PUSH){
```
chavdar: Please replace with Preconditions check

tuGithub: There are actually two ways to separate pull/pull flows

1. All datasets in Push mode import push tag, all datasets in Pull mode import pull tag, then in the job config, whitelist pull tag for pull job, whitelist push tag for push job

2. Or All datasets import data-derived tag, job lists/gets all the datasets out and based on the job config to filter out push or pull datasets

I would prefer the option2 which will not change tag once the mode be updated.
if option2, we should not use Preconditions

chavdar: Ok. In that case, can you just some logging to note why the dataset was skipped.

PR #1329, opened 2016-10-20
Fix AvroUtils.getFieldSchemaHelper() to support map and union types in schema

aditya1105:
Currently AvroUtils.getFieldSchemaHelper() doesn't support map and union types in schema passed to it as parameters. This fix is to support map completely in schema. Further it will support union with primitive data types. A union with complex data types will result in AvroRuntimeException.


pcadabam:
@abti can you review?

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -98,19 +99,31 @@
...
-    if (schema.getField(pathList.get(field)) == null) {
+    if (schema.getType() == Type.RECORD && schema.getField(pathList.get(field)) == null) {
       return Optional.absent();
     }
-    if ((field + 1) == pathList.size()) {
-      return Optional.fromNullable(schema.getField(pathList.get(field)).schema());
+    switch (schema.getType()) {
+      case UNION:
+        throw new AvroRuntimeException("Union of complex types cannot be handled : " + schema);
+      case MAP:
```
abti: Can you please include a couple of test cases: 
1. That retrieves a schema that has map in its path; and 
2. That goes the happy path of nested records (as was the only supported case) since there isn't already a test case for it

gobblin-utility/src/test/java/gobblin/util/AvroUtilsTest.java
```diff
@@ -242,6 +244,56 @@ public void testSwitchName() {
...
+        Schema.Type.STRING);
+  }
+
+  @Test(expectedExceptions = AvroRuntimeException.class)
+  public void testComplexTypesInUnionNotSupported() {
+    final String TEST_OBJECT = "testUnionObject";
+    final String TEST_FIELD_LOCATION = TEST_OBJECT + "." + this.TEST_FIELD1;
+
+
+    String testSchema = "{" +
```
abti: Since the schema is getting bigger, can you please move it to a resource file and read from it (refer to gobblin-data-management test cases for quick reference of tests that do that)

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
aditya1105: We have a test AvroUtilsTest.testGetObjectFromMap(), it does check for retrieving value as well as schema.

abti: Thanks, had missed it earlier.

gobblin-utility/src/test/java/gobblin/util/AvroUtilsTest.java
aditya1105: Sure

aditya1105:
@abti I made the required changes. Can you review it again?

PR #1330, opened 2016-10-20
Add publisher for hive registration of timestamp paths

jack-moseley:
@ydai1124 please review


gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisherWithHiveRegistration.java
```diff
@@ -0,0 +1,46 @@
...
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.publisher;
+
+import gobblin.configuration.ConfigurationKeys;
```
ydai1124: Can you fix the import order.

PR #1331, opened 2016-10-20
Implemented embedded distcp.

ibuenros:


ibuenros:
@abti can you review?

PR #1332, opened 2016-10-20
Add more metadata to compaction events

pcadabam:
@ydai1124 can you review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
@@ -41,6 +41,8 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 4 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -499,8 +499,24 @@ private void deleteAdditionalInputPaths() throws IOException {
...
-    CompactionSlaEventHelper.populateState(this.dataset, Optional.of(job), this.fs);
-    new SlaEventSubmitter(this.eventSubmitter, "CompactionCompleted", this.dataset.jobProps().getProperties()).submit();
+    try {
+      CompactionSlaEventHelper
+          .getEventSubmitterBuilder(this.dataset, Optional.of(job), this.fs)
+          .eventSubmitter(this.eventSubmitter)
+          .eventName(CompactionSlaEventHelper.COMPACTION_COMPLETED_EVENT_NAME)
+          .additionalMetadata(
+              CompactionRecordCountEvent.LATE_RECORD_COUNT,
+              Long.toString(this.lateOutputRecordCountProvider.getRecordCount(this.getApplicableFilePaths(this.dataset
```
ydai1124: `getApplicableFilePaths` make recursive `listFileStatus` calls, which can be expensive. Is that possible to cache the late record count for sending the event?

PR #1336, opened 2016-10-21
Change log messages to be generic, fix javadocs. Fixes #1335

### LINKED ISSUE #1335: Misleading error message in gobblin.source.extractor.extract.jdbc.JdbcExtractor
Error messages in class gobblin.source.extractor.extract.jdbc.JdbcExtractor:
line 791 - "Failed to get schema from Mysql - Resultset has no record" - mentions MySQL, while this class isn't MySql specific.
The  same issue repeats for many times in this class.

abti:


abti:
Minor changes. Self approving.

PR #1338, opened 2016-10-22
Changed LiKafkaSchemaRegistry to use correct id from MD5Digest

abti:
Fixed LiKafkaSchemaRegistry to use MD5Digest asString method that provides correct hex digest of data for schema id instead of toString which is not overridden over default impl


abti:
Minor fix. Self approving.

PR #1339, opened 2016-10-23
Added functionality to gracefully skip a work unit

aditya1105:
This fix is to allow a work unit to be skipped gracefully without creating it's corresponding task. The state corresponding to work unit will be saved. 
A new working state "SKIP" is added. 
Note that it is still the publisher's responsibility to handle such a work unit. A publisher can still mark it as failed or committed or simply skip.
Similarly source implementation will decide whether to create a new work unit or not , if a skip working state is found among previous work units.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
@@ -67,7 +67,8 @@ public class WorkUnitState extends State {
     SUCCESSFUL,
     COMMITTED,
     FAILED,
-    CANCELLED
+    CANCELLED,
+    SKIPPED
   }
 
   private final WorkUnit workUnit;
--- a/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
+++ b/gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
@@ -179,6 +179,13 @@ public class WorkUnit extends State {
   }
 
   /**
+   * This method will allow a work unit to be skipped if needed.
+   */
+  public void skip() {
+    this.setProp(ConfigurationKeys.WORK_UNIT_SKIP_KEY, true);
+  }
+
+  /**
    * Get the low {@link Watermark} as a {@link JsonElement}.
    *
    * @return a {@link JsonElement} representing the low {@link Watermark} or
--- a/gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
@@ -16,6 +16,7 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -122,6 +123,8 @@ public class JobState extends SourceState {
   private RunningState state = RunningState.PENDING;
   private int taskCount = 0;
   private final Map<String, TaskState> taskStates = Maps.newLinkedHashMap();
+  // Skipped task states shouldn't be exposed to publisher, but they need to be in JobState and DatasetState so that they can be written to StateStore.
+  private final Map<String, TaskState> skippedTaskStates = Maps.newLinkedHashMap();
 
   // Necessary for serialization/deserialization
   public JobState() {}
@@ -306,6 +309,31 @@ public class JobState extends SourceState {
     this.taskStates.put(taskState.getTaskId(), taskState);
   }
 
+  public void addSkippedTaskState(TaskState taskState) {
+    this.skippedTaskStates.put(taskState.getTaskId(), taskState);
+  }
+
+  public void removeTaskState(TaskState taskState) {
+    this.taskStates.remove(taskState.getTaskId());
+    this.taskCount--;
+  }
+
+  /**
+   * Filter the task states corresponding to the skipped work units and add it to the skippedTaskStates
+   */
+  public void filterSkippedTaskStates() {
+    List<TaskState> skippedTaskStates = new ArrayList<>();
+    for (TaskState taskState : this.taskStates.values()) {
+      if (taskState.getWorkingState() == WorkUnitState.WorkingState.SKIPPED) {
+        skippedTaskStates.add(taskState);
+      }
+    }
+    for (TaskState taskState : skippedTaskStates) {
+      removeTaskState(taskState);
+      addSkippedTaskState(taskState);
+    }
+  }
+
   /**
    * Add a collection of {@link TaskState}s.
    *
@@ -314,6 +342,12 @@ public class JobState extends SourceState {
   public void addTaskStates(Collection<TaskState> taskStates) {
     for (TaskState taskState : taskStates) {
       this.taskStates.put(taskState.getTaskId(), taskState);
+    }
+  }
+
+  public void addSkippedTaskStates(Collection<TaskState> taskStates) {
+    for (TaskState taskState : taskStates) {
+      addSkippedTaskState(taskState);
     }
   }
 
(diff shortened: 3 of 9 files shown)


pcadabam:
"Note that it is still the publisher's responsibility to handle such a work unit. A publisher can still mark it as failed or committed or simply skip."
Why should the the publisher be aware of skipped workunits? They are filtered by the `AbstractJobLauncher` right?

Can you please add some tests that verify the following -
1. The skipped workunits are persisted in the state store
2. The publisher does not receive these workunits
3. The job completes with state `SUCCESSFUL` even if few workunits has `SKIP` state
4. The skipped workunits can be read using `SourceState.getPreviousWorkunitStates()`

aditya1105:
TaskLevelPublisher will be unaware of skipped work units as we are avoiding task creation.

JobState consists of WorkUnitState of all WorkUnits including skipped WorkUnits. And all the WorkUnitStates from JobState are exposed to Publisher. Hence a job level publisher is aware of skipped work units. 

We have 2 options from here: 
1) Let the publisher decide if a skipped work unit should be marked as failed or successful or remain as skipped, depending on the use case.

2) Don't expose WorkUnitStates of skipped WorkUnits to the publisher.

aditya1105:
Added the test cases, specified above. Made changes in the job state, so that the skipped work units won't be exposed. Will add support for skipping multi work units if needed later.

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnit.java
```diff
@@ -229,7 +236,7 @@ public JsonElement getExpectedHighWatermark() {
 
   /**
    * Get the expected high {@link Watermark}.
-   *
```
agepati: Please remove the whitepace changes.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -306,6 +308,31 @@ public void addTaskState(TaskState taskState) {
...
+    this.taskStates.remove(taskState.getTaskId());
+    this.taskCount--;
+  }
+
+  /**
+   * Filter the task states corresponding to the skipped work units and add it to the skippedTaskStates
+   */
+
+  public void filterSkippedTaskStates() {
+    List<TaskState> taskStates = new ArrayList<>();
```
agepati: Might be better to rename to skippedTaskStates, in the context of the following removeTaskState invocation.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -122,6 +123,7 @@ public boolean isRunningOrDone() {
   private RunningState state = RunningState.PENDING;
   private int taskCount = 0;
   private final Map<String, TaskState> taskStates = Maps.newLinkedHashMap();
+  private final Map<String, TaskState> skippedTaskStates = Maps.newLinkedHashMap();
```
agepati: Could you put in a brief comment here about why this was needed. As in this is required to avoid reaching the publisher, but still allow state persistence.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -369,6 +402,16 @@ public int getCompletedTasks() {
...
     }
 
+    for (TaskState taskState : this.skippedTaskStates.values()) {
+      String datasetUrn = taskState.getProp(ConfigurationKeys.DATASET_URN_KEY, ConfigurationKeys.DEFAULT_DATASET_URN);
+      if (!datasetStatesByUrns.containsKey(datasetUrn)) {
+        DatasetState datasetState = newDatasetState(false);
+        datasetState.setDatasetUrn(datasetUrn);
+        datasetStatesByUrns.put(datasetUrn, datasetState);
+      }
+      datasetStatesByUrns.get(datasetUrn).addSkippedTaskState(taskState);
```
agepati: This code looks duplicated, could you please refactor it into a common method?

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -317,6 +344,12 @@ public void addTaskStates(Collection<TaskState> taskStates) {
     }
   }
 
+  public void addSkippedTaskStates(Collection<TaskState> taskStates) {
+    for (TaskState taskState : taskStates) {
+      this.skippedTaskStates.put(taskState.getTaskId(), taskState);
```
agepati: It might be simpler to inline this code in the initial iteration itself. Else, make it a method that works on a single element rather than the entire Collection so that it can be called from the first iteration.

gobblin-runtime/src/test/java/gobblin/GobblinRuntimeTest.java
```diff
@@ -0,0 +1,88 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import gobblin.runtime.local.LocalJobLauncher;
+
+
+/**
+ * Created by adsharma on 11/22/16.
+ */
+
+@Slf4j
```
agepati: This doesn't look like its being used?

gobblin-runtime/src/test/java/gobblin/GobblinRuntimeTest.java
```diff
@@ -0,0 +1,88 @@
...
+      throws Exception {
+    try (LocalJobLauncher localJobLauncher = new LocalJobLauncher(properties)) {
+      localJobLauncher.launchJob(null);
+    }
+  }
+
+  public Properties getJobProperties(String fileProperties)
+      throws IOException {
+    Properties jobProperties = new Properties();
+    InputStream inputStream = getClass().getClassLoader().getResourceAsStream(fileProperties);
```
agepati: Can possibly inline inputStream.

gobblin-runtime/src/test/java/gobblin/GobblinRuntimeTest.java
```diff
@@ -0,0 +1,88 @@
...
+   * @throws Exception
+   */
+
+  @Test
+  public void testSkippedWorkUnitsEvadePublisher()
+      throws Exception {
+    Properties jobProperties = getJobProperties("gobblinRuntimeTest/skipWorkUnitsTest.properties");
+    jobProperties.setProperty("data.publisher.type", "gobblin.skipWorkUnits.TestPublisher");
+    invokeLocalJobLauncher(jobProperties);
+    cleanDir();
```
agepati: If all the tests are invoking it, should probably mark this as a tearDown/@AfterTest

gobblin-api/src/main/java/gobblin/configuration/WorkUnitState.java
```diff
@@ -67,7 +67,8 @@
     SUCCESSFUL,
     COMMITTED,
     FAILED,
-    CANCELLED
+    CANCELLED,
+    SKIP
```
agepati: SKIPPED would seem to be more in line with convention.

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
agepati: Call addSkippedTaskState(taskState) instead?

gobblin-runtime/src/main/java/gobblin/runtime/JobState.java
```diff
@@ -306,6 +309,30 @@ public void addTaskState(TaskState taskState) {
...
+    this.taskCount--;
+  }
+
+  /**
+   * Filter the task states corresponding to the skipped work units and add it to the skippedTaskStates
+   */
+  public void filterSkippedTaskStates() {
+    List<TaskState> skippedTaskStates = new ArrayList<>();
+    for (TaskState taskState : this.taskStates.values()) {
+      if(taskState.getWorkingState() == WorkUnitState.WorkingState.SKIP)
```
agepati: Space after if, and even a single statement should be surrounded in braces.

gobblin-runtime/src/test/java/gobblin/skipWorkUnits/TestPublisher.java
```diff
@@ -0,0 +1,45 @@
+package gobblin.skipWorkUnits;
```
agepati: Why not gobblin.runtime? or gobblin.runtime.test?

pcadabam:
LGTM Fix and merge
- Please fix failing travis tests before merging
- Delete .keep files if you don't need them.

PR #1340, opened 2016-10-24
Gobblin cli

ibuenros:
This PR implements a Gobblin CLI for easily running arbitrary run-once applications (and to have a framework to add more applications in the future).

Usage:
`bin/gobblin run distcp -from /data/source -to /data/target`

Suggested order:
1. `CliEmbeddedGobblin` instantiates an embedded Gobblin and runs it. Does a lot of dynamic option generation based on the `EmbeddedGobblin` class that the user wants to run. Most of the business logic is here.
2. `EmbeddedGobblinCliOption`, `EmbeddedGobblinCliSupport`, `NotOnCli`: annotations used to control how `CliEmbeddedGobblin` generates dynamic options for Gobblin applications.
3. `GobblinCli` instantiates `CliApplication` and runs it.
4. `bin/gobblin.sh` and `bin/gobblin` used to launch Gobblin.
5. `EmbeddedGobblin` and `EmbeddedGobblinDistcp`, updated to support calling from CLI.


ibuenros:
@abti @chavdar can you review?

gobblin-runtime/src/main/java/gobblin/runtime/cli/CliEmbeddedGobblin.java
```diff
@@ -0,0 +1,290 @@
...
+  private static final List<String> BLACKLISTED_FROM_CLI = ImmutableList.of(
+      "getClass", "hashCode", "notify", "notifyAll", "toString", "wait"
+  );
+
+  @Override
+  public void run(String[] args) {
+
+    int startOptions = 1;
+
+    Class<? extends EmbeddedGobblin> klazz;
```
chavdar: please have a more descriptive name.

gobblin-runtime/src/main/java/gobblin/runtime/cli/CliEmbeddedGobblin.java
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * Usage:
+ * java -jar gobblin.jar run [app] [OPTIONS]
+ *
+ * The cli can run arbitrary applications by either setting each configuration manually (using the -setConfiguration
+ * option), or with the help of a template (using the -setTemplate option). However, since this requires knowing
+ * the exact configurations needed for a job, developers can implement more specific applications with simpler options
+ * for the user.
+ *
+ * {@link CliEmbeddedGobblin} will load any extension of {@link EmbeddedGobblin} using an alias or a class name. It will
```
chavdar: This looks a bit too restrictive on what the input parameters can be. For example, it will be hard to model a standard "cp source1 ... sourceN dest" command line. Why not let the application/tool take control of the command line? For example, we can have a factory instead of annotation

interface CliAccess {
   Options getOptions();
   void run(CommandLine cli) throws CliParseException, CliRunException;
}

Applications can provide an implementation of the interface that can take full use of the apache commons parsing capabilities like long/short names, argument names, multiple arguments, required parameters with option names, etc. You can move some of the annotation magic to a helper class that CliAccess implementations can use to create the command-line options list.

gobblin-runtime/src/main/java/gobblin/runtime/cli/Dummy.java
```diff
@@ -0,0 +1,19 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.runtime.cli;
+
+/**
+ * A dummy class used by {@link CliEmbeddedGobblin}.
+ */
+class Dummy {
```
chavdar: why is this needed?

gobblin-runtime/src/main/java/gobblin/runtime/cli/CliEmbeddedGobblin.java
abti: +1 Feels a bit restrictive, and trying to add more bells and whistle might feel like reinventing the wheel over apache commons

bin/gobblin.sh
```diff
@@ -0,0 +1,19 @@
...
+  case "$1" in
+    "classpath")
+      classpath
+      exit
+  esac
+done
+
+CLASSPATH=$(classpath)
+
+java -cp "$CLASSPATH" $GOBBLIN_OPTS gobblin.runtime.cli.GobblinCli $@
```
abti: We should let user specify JVM args, possibly work dir and log dir as well.

ibuenros: JVM args are specified with GOBBLIN_OPTS (this is the way Hadoop does it as well). work dir and log dir can be set already using something like `-configuration gobblin.wor.dir:/my/work/dir`. This is intended as a basic, first iteration of the script, but I can add this functionality now if you think it is important.

gobblin-runtime/src/main/java/gobblin/runtime/cli/CliEmbeddedGobblin.java
ibuenros: Implemented Chavdar's suggestion.

gobblin-runtime/src/main/java/gobblin/runtime/cli/Dummy.java
ibuenros: No longer needed. Removed.

bin/gobblin.sh
ibuenros: Added also a way to set log4j configuration easily.

ibuenros:
@chavdar @abti addressed comments.

gobblin-runtime/src/main/java/gobblin/runtime/cli/GobblinCli.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Instantiates a {@link CliApplication} and runs it.
+ */
+public class GobblinCli {
+
+  public static void main(String[] args) {
+    ClassAliasResolver<CliApplication> resolver = new ClassAliasResolver<>(CliApplication.class);
+
+    if (args.length < 1) {
```
chavdar: nit: can you also check for "-h"/"--help"?

gobblin-runtime/src/main/java/gobblin/runtime/cli/GobblinCli.java
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+
+    String alias = args[0];
+
+    try {
+      CliApplication application = resolver.resolveClass(alias).newInstance();
+      application.run(args);
+    } catch (ReflectiveOperationException roe) {
+      throw new RuntimeException("Could not find an application with alias " + alias);
```
chavdar: nit: Can you print to stderr and exit() instead? Printing out stacktraces is ugly.

PR #1344, opened 2016-10-26
Various fixes to embedded gobblin to allow it to run on mr mode.

ibuenros:
Major changes:
- Embedded Gobblin will automatically add the correct classpath to distribute cache for a basic Gobblin run, and allow users to add their own jars.
- Correctly set sys config in embedded mode.
- Mr mode will only upload jars to distributed cache when setting up the job. This avoids having to upload them if there are no work units. Additionally, mr mode can now reuse a common jar directory to avoid new uploads for every run.
- bin/gobblin.sh shell script will automatically load Hadoop classpath if HADOOP_HOME is defined.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -36,7 +36,6 @@ import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.Mapper;
 import org.apache.hadoop.mapreduce.filecache.DistributedCache;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
-import org.apache.hadoop.mapreduce.lib.input.NLineInputFormat;
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -108,6 +107,7 @@ public class MRJobLauncher extends AbstractJobLauncher {
   private final FileSystem fs;
   private final Job job;
   private final Path mrJobDir;
+  private final Path jarsDir;
   private final Path jobInputPath;
   private final Path jobOutputPath;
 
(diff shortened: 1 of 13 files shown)


ibuenros:
@chavdar can you review?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -249,37 +248,37 @@ protected void executeCancellation() {
...
    * Add dependent jars and files.
    */
-  private void addDependencies()
+  private void addDependencies(Configuration conf)
       throws IOException {
     TimingEvent distributedCacheSetupTimer =
         this.eventSubmitter.getTimingEvent(TimingEvent.RunJobTimings.MR_DISTRIBUTED_CACHE_SETUP);
 
-    Path jarFileDir = new Path(this.mrJobDir, JARS_DIR_NAME);
+    Path jarFileDir = this.jarsDir;
```
chavdar: I think this is  dangerous as jarsDir will be shared across different jobs/runs. This may cause jar with mismatched versions to be stored. 

I think you have two options:
- Use both a job-specific and global directory. Use the latter if the name and size match, otherwise upload to the job-specific dir.
- Have a more complex layout for the jar cache which allows different versions to be stored (like the gradle cache).

ibuenros:
@chavdar addressed comments.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
chavdar: I am not sure I understand. Aren't all embedded gobblin jobs re-use the same path?

ibuenros: Line 362 in `MRJobLauncher` is this: `       DistributedCache.addFileToClassPath(destJarFile, this.conf, this.fs);
`, so the jars get added individually to the distributed classpath, the directory is just a container for all jars.

chavdar: Never mind. If the jars are already versioned, it should work.

PR #1346, opened 2016-10-27
Ensure clusterIdentifier tag is included in the job metric context

agepati:
The clusterIdentifier tag is missing in the Gobblin events being emitted.

This is happening because the JobContext, and MetricsContext is being initialized before the clusterIdentifier tag is added into the JobState. Once the JobContext is already initialized minus the clusterIdentifier tag, further invocations of Instrumented.getMetricContext are unable to pick up the tag unless extra tags are explicitly added in again using GobblinMetrics.getCustomTagsFromState.


pcadabam:
LGTM

agepati:
Thank you! I don't yet have write access, could you please merge the changes in?

chavdar:
Done.

PR #1347, opened 2016-10-28
Initial commit for speculative execution in MR mode

ydai1124:
This is the first PR of https://docs.google.com/document/d/1Vf5A7pHbFKwr7-Bbw2X3zBBVZSvT7HXEPE6bsi9qGco/
Main changes:
1. It has the new `GobblinMultiTaskAttempt`, which does not contain any real execution but provides the layout.
Gobblin `Task` has new `commit` method, which is empty in this PR.
1. It has the new interface `SpeculativeAttemptAwareConstruct`. Some most commonly used gobblin writers, and writer wrappers, have implemented this interface.

@chavdar @ibuenros Can you review?

### DIFF
--- a/gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
+++ b/gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
@@ -34,6 +34,7 @@ public abstract class DataWriterBuilder<S, D> {
   protected S schema;
   protected int branches;
   protected int branch;
+  protected String writerAttemptId;
 
   /**
    * Tell the writer the destination to write to.
@@ -98,6 +99,15 @@ public abstract class DataWriterBuilder<S, D> {
    */
   public DataWriterBuilder<S, D> forBranch(int branch) {
     this.branch = branch;
+    return this;
+  }
+
+  /**
+   * Attempt Id for this writer. There could be two duplicate writers with the same {@link #writerId},
+   * their writerAttemptId should be different.
+   */
+  public DataWriterBuilder<S, D> withAttemptId(String attemptId) {
+    this.writerAttemptId = attemptId;
     return this;
   }
 
--- a/gobblin-core/src/main/java/gobblin/commit/SpeculativeAttemptAwareConstruct.java
+++ b/gobblin-core/src/main/java/gobblin/commit/SpeculativeAttemptAwareConstruct.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.commit;
+
+import gobblin.annotation.Alpha;
+
+
+/**
+ * A declaration by any Gobblin construct to claim whether it is safe to have multiple speculative attempts.
+ * For example, if any {@link gobblin.writer.DataWriter} implements {@link SpeculativeAttemptAwareConstruct}
+ * and returns true in {@link #isSpeculativeAttemptSafe()}, then multiple attempts of one {@link gobblin.writer.DataWriter}
+ * should not cause conflict among them.
+ */
+@Alpha
+public interface SpeculativeAttemptAwareConstruct {
+
+  /**
+   * @return true if it is safe to have multiple speculative attempts; false, otherwise.
+   * To avoid inheritance issue, the suggested pattern would be "return this.class == MyClass.class".
+   */
+  public boolean isSpeculativeAttemptSafe();
+}
--- a/gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
@@ -30,6 +30,7 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.io.Closer;
 
+import gobblin.commit.SpeculativeAttemptAwareConstruct;
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.State;
 import gobblin.instrumented.writer.InstrumentedDataWriterDecorator;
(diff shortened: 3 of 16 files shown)


gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -162,16 +172,20 @@ public long bytesWritten() throws IOException {
...
-  private DataWriter<D> createPartitionWriter(GenericRecord partition) throws IOException {
+  private DataWriter<D> createPartitionWriter(GenericRecord partition)
+      throws IOException {
     if (!this.builder.isPresent()) {
       throw new IOException("Writer builder not found. This is an error in the code.");
     }
-    return this.builder.get().forPartition(partition).withWriterId(this.baseWriterId + "_" + this.writerIdSuffix++)
+    DataWriter dataWriter =  this.builder.get().forPartition(partition).withWriterId(this.baseWriterId + "_" + this.writerIdSuffix++)
         .build();
+    this.isSpeculativeAttemptAware = this.isDataWriterForPartitionSafe(dataWriter);
```
chavdar: You should "and" the safeness as there might be multiple writers instantiated for different partitions.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -58,49 +59,55 @@
   private final Optional<PartitionAwareDataWriterBuilder> builder;
   private final boolean shouldPartition;
   private final Closer closer;
+  private boolean isSpeculativeAttemptAware;
```
chavdar: nit: i think the correct name for the variable is "isSpeculativeAttemptSafe"

gobblin-runtime/src/main/java/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -0,0 +1,127 @@
...
+
+    try {
+      List<Future<Void>> futures = new IteratorExecutor<>(callableIterator, this.getTaskCommitThreadPoolSize(),
+          ExecutorsUtils.newDaemonThreadFactory(Optional.of(log), Optional.of("Task-committing-pool-%d"))).execute();
+
+      for (Future<Void> future : futures) {
+        try {
+          future.get();
+        } catch (ExecutionException exc) {
+          log.error("Failed to commit a task.", exc.getCause());
```
chavdar: Shouldn't we fail the whole attempt if one of the tasks fails?

gobblin-runtime/src/main/java/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+      }
+    } catch (InterruptedException ie) {
+      log.error("Committing of tasks interrupted. Aborting.");
+      throw new RuntimeException(ie);
+    }
+
+    for (CommitStep additionalCommitStep: this.additionalCommitSteps) {
```
chavdar: What is the semantics for "additionalCommitSteps" with respect to partial attempt failures?

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -46,7 +47,7 @@
  * @param <D> record type.
  */
 @Slf4j
-public class PartitionedDataWriter<S, D> implements DataWriter<D>, FinalState {
+public class PartitionedDataWriter<S, D> implements DataWriter<D>, FinalState, SpeculativeAttemptAwareConstruct {
```
chavdar: Please add unit tests for speculative execution awareness of this class

gobblin-runtime/src/main/java/gobblin/runtime/GobblinMultiTaskAttempt.java
ydai1124: It is actually the "clean up" steps, which should be always executed. I will put it in a finally cause.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
ydai1124: Added.

gobblin-core/src/main/java/gobblin/writer/PartitionedDataWriter.java
ydai1124: Fixed.

gobblin-runtime/src/main/java/gobblin/runtime/GobblinMultiTaskAttempt.java
ydai1124: Fixed.

ydai1124:
@chavdar Comments addressed.

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
```diff
@@ -34,6 +34,7 @@
   protected S schema;
   protected int branches;
   protected int branch;
+  protected String writerAttemptId;
```
ibuenros: Do we need a default for this?

gobblin-core/src/main/java/gobblin/commit/SpeculativeAttemptAwareConstruct.java
```diff
@@ -0,0 +1,31 @@
...
+ * A declaration by any Gobblin construct to claim whether it is safe to have multiple speculative attempts.
+ * For example, if any {@link gobblin.writer.DataWriter} implements {@link SpeculativeAttemptAwareConstruct}
+ * and returns true in {@link #isSpeculativeAttemptSafe()}, then multiple attempts of one {@link gobblin.writer.DataWriter}
+ * should not cause conflict among them.
+ */
+@Alpha
+public interface SpeculativeAttemptAwareConstruct {
+
+  /**
+   * @return true if it is safe to have multiple speculative attempts; false, otherwise.
```
ibuenros: Can you add to the javadoc the suggested pattern of this? e.g. `return this.class == MyClass.class`

gobblin-runtime/src/main/java/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -0,0 +1,142 @@
...
+                  throws Exception {
+                task.commit();
+                return null;
+              }
+            };
+          }
+        });
+
+    try {
+      List<Future<Void>> futures = new IteratorExecutor<>(callableIterator, this.getTaskCommitThreadPoolSize(),
```
ibuenros: You can use `executeAndGetResults` instead, then run `IteratorExecutor.verifyAllSuccessful`. If you want to log the failures, you can use `IteratorExecutor.logFailures`.

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
ydai1124: It can be null if not specified when building a writer.

ydai1124:
@ibuenros comments addressed.

PR #1349, opened 2016-10-29
Configurable http clients

chavdar:
Added instrumentation to allow customization of the instantiation of HttpClient objects:
- Enabled configurability through the job State/Config
- Enabled the adding of decorations to the HttpClient

At the core is the HttpClientConfigurator interface which allows to:
- Extract relevant configuration for the HttpConfig from the job Config/State (configure() methods)
- Create decorations of the HttpClients (getHttpClient())

DefaultHttpClientConfigurator is the standard implementation of HttpClientConfigurator. HttpClientConfiguratorLoader can be used to inject different implementations through job State/Config.

I've refactored RestApiConnector and AbstractHttpWriterBuilder to use the above instrumentation. Next is the Wikipedia extractor.

I've also done some minor code hygiene changes to clean up code/remove compiler warnings.

@ydai1124 @jinhyukchang can you review?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/http/DefaultHttpClientConfigurator.java
+++ b/gobblin-core/src/main/java/gobblin/http/DefaultHttpClientConfigurator.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+package gobblin.http;
+
+import java.util.regex.Matcher;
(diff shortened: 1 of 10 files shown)


gobblin-core/src/main/java/gobblin/http/DefaultHttpClientConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+  // IMPORTANT: don't change the values for PROXY_URL_KEY and PROXY_PORT_KEY as they are meant to
+  // be backwards compatible with SOURCE_CONN_USE_PROXY_URL and SOURCE_CONN_USE_PROXY_PORT when
+  // the statePropertiesPrefix is "source.conn."
+  /** The hostname of the HTTP proxy to use */
+  public static final String PROXY_URL_KEY = "use.proxy.url";
+  /** The port of the HTTP proxy to use */
+  public static final String PROXY_PORT_KEY = "use.proxy.port";
+  /** Similar to {@link #PROXY_URL_KEY} and {@link #PROXY_PORT_KEY} but allows you to set it on
+   * one property as <host>:<port> */
+  public static final String PROXY_HOSTPORT_KEY = "proxyHostport";
```
ydai1124: Shall we mark it explicitly that setting of `PROXY_HOSTPORT_KEY` will override other two keys if they are all used?

chavdar: Fixed.

PR #1351, opened 2016-10-31
Distcp will report copy speed metrics.

ibuenros:


ibuenros:
@ydai1124 can you review?

PR #1352, opened 2016-10-31
Embedded docs

ibuenros:
Changes:
* Added embedded wikipedia example, can now run wikipedia example from CLI.
* Added documentation for embedded Gobblin, CLI, and updated getting started with CLI.
* Allow specifying that output of `BaseDataPublisher` should go directly under `DATA_PUBLISHER_FINAL_DIR` (instead of adding the extract as subdirectories, with a lot of job_id ugliness).
* Add an option to CLI Gobblin to use logger instead of stdout.
* Refactored CLI factories a bit to allow easier use of the helper classes (inferring options from public methods and constructors).
* Gobblin CLI does not use state store by default.
* Gobblin CLI allows adding custom jars to classpath.


ibuenros:
@abti can you review? Thanks!

ibuenros:
@pcadabam can you review?

PR #1354, opened 2016-11-02
ETL-5076: Delta support in gobblin compaction

ydai1124:
Provide two ways of getting delta field in reducer. One is conf based, and another is from the record schema.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroDeltaFieldNameProvider.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroDeltaFieldNameProvider.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.compaction.mapreduce.avro;
+
+import java.util.List;
+
+import org.apache.avro.generic.GenericRecord;
+
+
+/**
+ * Provides the delta field names for {@link AvroKeyDedupReducer}.
+ */
+public interface AvroDeltaFieldNameProvider {
+  /**
+   * Return the delta fields using the given {@link GenericRecord}.
+   */
+  public List<String> getDeltaFieldNames(GenericRecord record);
+}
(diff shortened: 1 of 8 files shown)


ydai1124:
@ibuenros Can you take a look? Thanks

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroDeltaFieldNameProvider.java
```diff
@@ -0,0 +1,53 @@
...
+            return getDeltaFieldNamesForNewSchema(schema);
+          }
+        });
+  }
+
+  public abstract List<String> getDeltaFieldNamesForNewSchema(Schema originalSchema);
+
+  public List<String> getDeltaFieldNames(GenericRecord record) {
+    try {
+      return recordSchemaToDeltaSchemaCache.get(record.getSchema());
```
ibuenros: This call probably requires at least one call to `Schema.equals(Schema)`, which is probably relatively slow. If the PK is small and the Schema is large, this might dominate the runtime of the reducer, and I think it might be overkill to assume multiple Schemas in the same reducer.

My suggestion would be to use instantiate an arbitrary `Comparator<GenericRecord>` in the reducer. We can have simple comparators that simply extract a static field and compare, as well as more complex comparators that do essentially what this class is doing.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyDedupReducer.java
```diff
@@ -28,31 +35,50 @@
...
     int numVals = 0;
 
+    AvroValue<GenericRecord> valueToRetain = null;
+
     for (AvroValue<GenericRecord> value : values) {
-      this.outKey.datum(value.datum());
+      if (valueToRetain == null) {
+        valueToRetain = value;
+      } else if (this.deltaComparatorOptional.isPresent()) {
+        valueToRetain = this.deltaComparatorOptional.get().compare(valueToRetain, value) > 0 ? valueToRetain : value;
```
ibuenros: Nit: if there are multiple records that are equal, this algorithm will chose the first record if `deltaComparatorOptional` is missing, but it will chose the last record if it is present. For consistency sake, can you change the condition to `>=`?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroDeltaFieldNameProvider.java
ydai1124: Made changes to the `ConfBasedDeltaFieldProvider` so that it simply returns the field from job config without checking the record schema. I think it achieves the same effect as a simple comparator.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyDedupReducer.java
ydai1124: True. Fixed.

ydai1124:
@ibuenros Comments addressed.

PR #1356, opened 2016-11-02
Fixed a bug on CsvToJsonConverterV2 throwing NPE when custom order is…

jinhyukchang:
Fixed a bug on CsvToJsonConverterV2 throwing NPE when custom order is not provided. Updated corresponding unit test.


jinhyukchang:
@abti Could you take a look when you have a chance?

PR #1360, opened 2016-11-03
Capability to register and update Hive view over converted Hive datasets

abti:
This PR adds the following capabilities: 

- Create if not exists Hive view over the converted Hive table / dataset if configured 
- Based on configuration: update view every time Avro to ORC conversion happens OR only when schema evolution happens

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
@@ -192,6 +192,10 @@ public abstract class AbstractAvroToOrcConverter extends Converter<Schema, Schem
     Optional<Table> destinationTableMeta = destinationMeta.getLeft();
 
     // Optional
+    // wrapperViewName          : If specified view with 'wrapperViewName' is created if not already exists
+    //                            over destination table
+    // isUpdateViewAlwaysEnabled: If false 'wrapperViewName' is only updated when schema evolves; if true
+    //                            'wrapperViewName' is always updated (everytime publish happens)
     Optional<String> wrapperViewName = getConversionConfig().getDestinationViewName();
     boolean shouldUpdateView = getConversionConfig().isUpdateViewAlwaysEnabled();
     Optional<List<String>> clusterBy =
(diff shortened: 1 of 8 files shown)


abti:
@pcadabam Please review.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/dataset/ConvertibleHiveDataset.java
```diff
@@ -215,6 +219,11 @@ private ConversionConfig(Config config, Table table, String destinationFormat) {
       this.destinationDataPath = resolveTemplate(config.getString(DESTINATION_DATA_PATH_KEY), table);
 
       // Optional
+      if (config.hasPath(DESTINATION_VIEW_KEY)) {
+        this.destinationViewName = Optional.of(resolveTemplate(config.getString(DESTINATION_VIEW_KEY), table));
```
pcadabam: nit: you can use `ConfigUtils.getString(config, path, defualt)`

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
```diff
@@ -192,6 +192,8 @@ public AbstractAvroToOrcConverter() {
     Optional<Table> destinationTableMeta = destinationMeta.getLeft();
 
     // Optional
+    Optional<String> wrapperViewName = getConversionConfig().getDestinationViewName();
```
pcadabam: Can you document what a `wrapperView` means and how/when is it used?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/converter/AbstractAvroToOrcConverter.java
```diff
@@ -192,6 +192,8 @@ public AbstractAvroToOrcConverter() {
...
+    boolean shouldUpdateView = getConversionConfig().isUpdateViewAlwaysEnabled();
```
pcadabam: Why do you need another flag. If `wrapperViewName`  means the user wants to update view right?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/dataset/ConvertibleHiveDataset.java
abti: Didn't use it because resolveTemplate cannot support nulls, but added null support there

PR #1361, opened 2016-11-03
Added Kerberos keytab authentication to embedded gobblin.

ibuenros:


ibuenros:
@chavdar can you review? Thanks!

PR #1306, opened 2016-10-07
Data replication for pull mode

tuGithub:

PR #1363, opened 2016-11-03
Second commit of speculative execution in MR mode

ydai1124:
Design doc: https://docs.google.com/document/d/1Vf5A7pHbFKwr7-Bbw2X3zBBVZSvT7HXEPE6bsi9qGco/edit#

What this PR contains:
1. Modified `Task` to contain the real `run` and `commit` method.
2. Modified `AbstractJobLauncher` to create `GobblinMultiTaskAttempt`, and use the `run` and `commit` steps in `GobblinMultiTaskAttempt`.
3. Modified `MRJobLauncher` to accommodate the situation with speculative execution enabled.
4. Modified `TaskStateTracker` to have two callback functions: `onTaskRunCompletion` and`onTaskCommitCompletion`.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
@@ -12,6 +12,7 @@
 
 package gobblin.data.management.copy.extractor;
 
+import gobblin.configuration.WorkUnitState;
 import gobblin.data.management.copy.CopyableFile;
 import gobblin.data.management.copy.FileAwareInputStream;
 import gobblin.source.extractor.DataRecordException;
@@ -39,13 +40,19 @@ public class FileAwareInputStreamExtractor implements Extractor<String, FileAwar
 
   private final FileSystem fs;
   private final CopyableFile file;
+  private final WorkUnitState state;
   /** True indicates the unique record has already been read. */
   private boolean recordRead;
 
-  public FileAwareInputStreamExtractor(FileSystem fs, CopyableFile file) {
+  public FileAwareInputStreamExtractor(FileSystem fs, CopyableFile file, WorkUnitState state) {
     this.fs = fs;
     this.file = file;
+    this.state = state;
     this.recordRead = false;
+  }
+
+  public FileAwareInputStreamExtractor(FileSystem fs, CopyableFile file) {
+    this(fs, file, null);
   }
 
   /**
@@ -53,7 +60,8 @@ public class FileAwareInputStreamExtractor implements Extractor<String, FileAwar
    * @throws IOException
    */
   @Override
-  public String getSchema() throws IOException {
+  public String getSchema()
+      throws IOException {
     return FileAwareInputStream.class.getName();
   }
 
@@ -62,11 +70,13 @@ public class FileAwareInputStreamExtractor implements Extractor<String, FileAwar
       throws DataRecordException, IOException {
 
     if (!this.recordRead) {
+      Configuration conf =
+          this.state == null ? HadoopUtils.newConfiguration() : HadoopUtils.getConfFromState(this.state);
+      FileSystem fsFromFile = this.file.getOrigin().getPath().getFileSystem(conf);
       this.recordRead = true;
-      return new FileAwareInputStream(this.file, this.fs.open(this.file.getFileStatus().getPath()));
+      return new FileAwareInputStream(this.file, fsFromFile.open(this.file.getFileStatus().getPath()));
     }
     return null;
-
   }
 
   @Override
@@ -80,5 +90,7 @@ public class FileAwareInputStreamExtractor implements Extractor<String, FileAwar
   }
 
   @Override
-  public void close() throws IOException {}
+  public void close()
+      throws IOException {
+  }
 }
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
@@ -62,10 +62,8 @@ public class FileAwareInputStreamExtractor implements Extractor<String, FileAwar
       throws DataRecordException, IOException {
 
     if (!this.recordRead) {
-      Configuration conf = HadoopUtils.newConfiguration();
-      FileSystem fsFromFile = this.file.getOrigin().getPath().getFileSystem(conf);
       this.recordRead = true;
-      return new FileAwareInputStream(this.file, fsFromFile.open(this.file.getFileStatus().getPath()));
+      return new FileAwareInputStream(this.file, this.fs.open(this.file.getFileStatus().getPath()));
     }
     return null;
 
--- a/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
@@ -120,6 +120,20 @@ public abstract class AbstractJobLauncher implements JobLauncher {
   // A list of JobListeners that will be injected into the user provided JobListener
   private final List<JobListener> mandatoryJobListeners = Lists.newArrayList();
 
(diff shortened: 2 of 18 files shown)


ydai1124:
@chavdar @ibuenros  can you take a look? Thanks!

chavdar:
There are FindBugs failures:

> Execution failed for task ':gobblin-core:findbugsMain'

ydai1124:
@chavdar Couldn't reproduce locally. Re-launched the travis build and it succeeded.

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
```diff
@@ -62,10 +62,8 @@ public FileAwareInputStream readRecord(@Deprecated FileAwareInputStream reuse)
       throws DataRecordException, IOException {
 
     if (!this.recordRead) {
-      Configuration conf = HadoopUtils.newConfiguration();
-      FileSystem fsFromFile = this.file.getOrigin().getPath().getFileSystem(conf);
       this.recordRead = true;
-      return new FileAwareInputStream(this.file, fsFromFile.open(this.file.getFileStatus().getPath()));
+      return new FileAwareInputStream(this.file, this.fs.open(this.file.getFileStatus().getPath()));
```
ibuenros: This will break pull mode from multiple `FileSystem`s. Why is this change needed?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -575,7 +529,69 @@ private void addConstructsFinalStateToTaskState(InstrumentedExtractorBase<?, ?>
    * 3. Check whether to publish data in task.
    */
   public void commit() {
-    // TODO- refactor the current Task.run() method to run & commit phases, and add commit content here.
+    try {
+      // Check if all forks succeeded
+      boolean allForksSucceeded = true;
```
ibuenros: shouldn't the success check happen in the `run()` method?

gobblin-data-management/src/main/java/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractor.java
ydai1124: This change was introduced in #1306 and broke lumos-ng puller. The source filesystem cannot be got from an empty `configuration`. 

I was unaware of the multiple `FileSystem` mode. One possible solution may be instead of using `path.getFs`, will `copyfile.getOriginFs` work for the multiple `FileSystem` mode? Another solution could be creating one more constructor for `FileAwareInputStreamExtractor`, which can pass the job `properties`/`state` in. Then `path.getFs` can have a `FileSystem` instance for `http` uri.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
ydai1124: This check is checking if `fork.commit()` succeeds. The main job of `fork.commit()` is to run `writer.commit()` method to move data from staging to writers' final output. 

The `commit` phase, especially in speculative execution mode, should be something executed only once. So I moved `fork.commit` to the `commit` phase.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -117,13 +121,22 @@ public Task(TaskContext context, TaskStateTracker taskStateTracker, TaskExecutor
         closer.register(new InstrumentedExtractorDecorator<>(this.taskState, this.taskContext.getExtractor()));
 
     this.converter = closer.register(new MultiConverter(this.taskContext.getConverters()));
-
+    try {
+      this.rowChecker = closer.register(this.taskContext.getRowLevelPolicyChecker());
```
chavdar: Why is this necessary? Shouldn't the converter instantiation be also in the try-catch block?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -575,7 +529,69 @@ private void addConstructsFinalStateToTaskState(InstrumentedExtractorBase<?, ?>
...
+        }
+      }
+
+      if (allForksSucceeded) {
+        // Set the task state to SUCCESSFUL. The state is not set to COMMITTED
+        // as the data publisher will do that upon successful data publishing.
+        this.taskState.setWorkingState(WorkUnitState.WorkingState.SUCCESSFUL);
+      } else {
+        LOG.error(String.format("Not all forks of task %s succeeded", this.taskId));
+        this.taskState.setWorkingState(WorkUnitState.WorkingState.FAILED);
```
chavdar: please use failTask() with a exception specifying which forks failed. This will ensure consistent behavior about what failure state is maintained in the task.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -623,61 +625,42 @@ private EventSubmitter buildEventSubmitter(List<? extends Tag<?>> tags) {
    * @param taskExecutor a {@link TaskExecutor} for task execution
    * @param taskStateStore a {@link StateStore} for storing {@link TaskState}s
    * @param logger a {@link Logger} for logging
+   * @param shouldCommitTasks whether should commit {@link GobblinMultiTaskAttempt}
```
chavdar: I think this parameter name and javadoc are misleading. Ultimately, all successful tasks should commit. The questions is whether they should commit immediately after each run or at the end. Please think about improving the naming. Using a enum instead of boolean can also improve readability.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -577,17 +581,53 @@ protected void setup(Context context) {
...
+              .put(context.getTaskAttemptID().toString(), gobblinMultiTaskAttempt);
+        }
       } finally {
-        this.cleanup(context);
+        CommitStep cleanUpCommitStep = new CommitStep() {
+
+          @Override
+          public boolean isCompleted()
+              throws IOException {
+            throw new UnsupportedOperationException("Not supported.");
```
chavdar: What is the expected behavior here? Isn't it better to return "false" instead of throwing an exception?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -577,17 +581,53 @@ protected void setup(Context context) {
...
+          public void execute()
+              throws IOException {
+            LOG.info("Starting the clean-up steps.");
+            try {
+              serviceManager.stopAsync().awaitStopped(5, TimeUnit.SECONDS);
+            } catch (TimeoutException te) {
+              // Ignored
+            } finally {
+              if (jobMetrics.isPresent()) {
+                // TODO: Add "committed" tag to final metrics.
```
chavdar: Is this going to be fixed?

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ydai1124: It is used by "exactly once" semantics to check if a `commitstep` from a previous execution has been executed or not. Change to `return !serviceManager.isHealthy();` to return if `serviceManager` has stopped.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
ydai1124: Yes. Next and last PR will be adding this tag, which is currently not supported by Gobblin metrics.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
ydai1124: Because only `getRowLevelPolicyChecker` throws exception in its method signature.

ydai1124:
@chavdar @ibuenros comments addressed.

PR #1364, opened 2016-11-04
Wikipedia httpclient

chavdar:
Originally, the Wikipedia extract used a plain Java HttpURLConnection. For example, it does not support going through a HTTP proxy.

This is a rather rudimentary solution. I've switched the client to the new configurable HttpClients facility.

@pcadabam  can you review?


pcadabam:
LGTM

PR #1365, opened 2016-11-04
Fixed some tests using wrong temporary directory.

ibuenros:


ibuenros:
@pcadabam can you review?

PR #1366, opened 2016-11-04
Making KafkaDataWriter support any class that implements the Producer…

shirshanka:
… interface. 

Should have done this earlier. Currently the KafkaDataWriter expects classes to extend KafkaProducer which is not an interface. Producer is the better interface to expect.


shirshanka:
@abti : can you review?

PR #1368, opened 2016-11-07
Update lib versions for gobblin-{compaction,mapreduce}.sh

klyr:
Versions in AWS related plugins were not updated as the corresponding S3 directory don't have the new libraries


chavdar:
Thank you @klyr !

PR #1369, opened 2016-11-08
Add sanitization of job property keys which are prefixes of others

chavdar:
We have repeated issues with converting Properties to Config because of keys which are prefixes of other keys.

The approach I've taken is to detect Properties collections which have such keys and sanitize the names for such properties.

- ConfigUtils.propertiesToConfig() will detect such keys and append .ROOT_VALUE at the end of the offending property
- ConfigUtils.configToProperties() will detect properties which end in .ROOT_VALUE and remove it transparently.

For legacy jobs, constructs we typically have the flow: Job pull file (Properties) -> Job config (Config) -> State properties. PullFileLoader/SchedulerUtils use the ConfigUtils.propertiesToConfig() to do the first transition and JobLauncherExecutionDriver uses ConfigUtils.configToProperties() to do the second transition, the offending properties are handled transparently.

Further, when we move to fully Config-based job files, we can still explicitly add the .ROOT_VALUE suffix and work with legacy components.

I've refactored a bunch of classes to use ConfigUtils.propertiesToConfig() instead of ConfigFactory.parseProperties() to make use of the santiization.

@ydai @pcadabam  Can you review?

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
@@ -17,14 +17,14 @@ import java.io.StringReader;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.commons.lang3.StringUtils;
-
-import com.opencsv.CSVReader;
 
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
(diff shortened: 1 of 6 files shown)


gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -149,16 +188,33 @@ public static Config propertiesToConfig(Properties properties, Optional<String>
...
+
+    Set<String> fullPrefixKeys = findFullPrefixKeys(properties, prefix);
+
     ImmutableMap.Builder<String, Object> immutableMapBuilder = ImmutableMap.builder();
     for (Map.Entry<Object, Object> entry : properties.entrySet()) {
       String entryKey = entry.getKey().toString();
-      if (StringUtils.startsWith(entryKey, prefix.or(StringUtils.EMPTY)) && !blacklistedKeys.contains(entryKey)) {
+      if (StringUtils.startsWith(entryKey, prefix.or(StringUtils.EMPTY)) &&
+          !blacklistedKeys.contains(entryKey)) {
+        if (fullPrefixKeys.contains(entryKey)) {
```
pcadabam: This should never happen but in a rare case where user's key already has a substring `.ROOT_VALUE` we may want to throw some exception saying `.ROOT_VALUE` is a keyword

PR #1371, opened 2016-11-09
Added new construct of VersionCleaner with pre and post activity hook in Retention. Also added capability in Hive retention to swap-in partitions as post cleanup activity

abti:
This PR adds: 

A. New construct of VersionCleaner in Retention with pre and post activity hook. 

B. Capability in Hive retention to swap-in a partition from another table in lieu of the partition being deleted from the current table as post cleanup activity. 

This is important for use-cases such as hybrid tables (with Avro and ORC based partitions), where we want to replace ORC partitions with Avro partitions when certain retention predicate become true (such as retain ORC partitions for 30 days, and then replace them with Avro partitions from another table which has the same data in Avro format) 

To illustrate its functioning lets assume that: 
1. We want to delete ORC partition (date="2016-01-01-00") from table db_orc.prod1_table_orc
2. We want to register Avro partition (date="2016-01-01-00") from table db_avro.prod2_table_avro to table db_orc.prod1_table_orc at the same time (since it contains same data but in different physical format ie. avro)

So, this is achieved as follows:
1. Following configuration is set in config store for this dataset (or parent level dataset, refer to config store setup details to learn more in Gobblin wiki): 
- hive.datasetNamePattern = $LOGICAL_DB_orc.prod1_$LOGICAL_TABLE_orc
- hive.replacementHiveDbName = $LOGICAL_DB_avro
- hive.replacementHiveTableName = prod2_$LOGICAL_TABLE_avro
- hive.shouldReplacePartition = true (to enable partition replacement) 
Note: $LOGICAL_DB and $LOGICAL_TABLE are recognized tokens by the code 
In HiveDataset, by applying hive.datasetNamePattern on current Hive dataset (db.table): 
$LOGICAL_DB is auto-resolved to: db 
$LOGICAL_TABLE is auto-resolved to: table 
hive.replacementHiveDbName is auto-resolved to: db_avro 
hive.replacementHiveTableName is auto-resolved to: prod2_table_avro 
2. The HiveVersionCleaner in its clean() method, drops the partition (date="2016-01-01-00") from db_orc.prod1_table_orc
3. The HiveVersionCleaner in its postCleanAction() method then registers the corresponding partition (date="2016-01-01-00") from db_avro.prod2_table_avro to db_orc.prod1_table_orc table 

Please note: This cross-registration of the partition does not duplicates the data, rather the partition registration is only a reference to the data from source table

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
@@ -190,7 +190,7 @@ public class HiveDataset implements PrioritizedCopyableDataset {
    * @return  Parsed logical DB and Table.
    */
   @VisibleForTesting
-  public static DbAndTable parseLogicalDbAndTable(String datasetNamePattern, DbAndTable dbAndTable,
+  protected static DbAndTable parseLogicalDbAndTable(String datasetNamePattern, DbAndTable dbAndTable,
       String logicalDbToken, String logicalTableToken) {
     Preconditions.checkArgument(StringUtils.isNotBlank(datasetNamePattern), "Dataset name pattern must not be empty.");
 
@@ -222,7 +222,7 @@ public class HiveDataset implements PrioritizedCopyableDataset {
    * @return Extracted token value from the source entity.
    */
   @VisibleForTesting
-  public static String extractTokenValueFromEntity(String sourceEntity, String sourceTemplate, String token) {
+  protected static String extractTokenValueFromEntity(String sourceEntity, String sourceTemplate, String token) {
     Preconditions.checkArgument(StringUtils.isNotBlank(sourceEntity), "Source entity should not be blank");
     Preconditions.checkArgument(StringUtils.isNotBlank(sourceTemplate), "Source template should not be blank");
     Preconditions.checkArgument(sourceTemplate.contains(token), String.format("Source template: %s should contain token: %s", sourceTemplate, token));
@@ -245,7 +245,7 @@ public class HiveDataset implements PrioritizedCopyableDataset {
    * @return Resolved config object.
    */
   @VisibleForTesting
-  public static Config resolveConfig(Config datasetConfig, DbAndTable realDbAndTable, DbAndTable logicalDbAndTable) {
+  protected static Config resolveConfig(Config datasetConfig, DbAndTable realDbAndTable, DbAndTable logicalDbAndTable) {
     Preconditions.checkNotNull(datasetConfig, "Dataset config should not be null");
     Preconditions.checkNotNull(realDbAndTable, "Real DB and table should not be null");
     Preconditions.checkNotNull(logicalDbAndTable, "Logical DB and table should not be null");
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
@@ -17,6 +17,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import lombok.Getter;
@@ -28,11 +29,16 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.ql.metadata.Table;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.base.Optional;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigValue;
+import com.typesafe.config.ConfigValueType;
 
 import gobblin.annotation.Alpha;
 import gobblin.configuration.State;
@@ -46,6 +52,7 @@ import gobblin.hive.HiveMetastoreClientPool;
 import gobblin.instrumented.Instrumented;
 import gobblin.metrics.MetricContext;
 import gobblin.metrics.Tag;
+import gobblin.util.ConfigUtils;
 import gobblin.util.PathUtils;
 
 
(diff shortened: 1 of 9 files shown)


abti:
@pcadabam Please review

abti:
Updated PR based on discussions offline.

gobblin-data-management/src/main/java/gobblin/data/management/retention/version/HiveDatasetVersionCleaner.java
```diff
@@ -0,0 +1,250 @@
...
+    // For post cleanup activity:
+    // Determine source db / table name from which partition has to be replaced-in for the target partition being deleted.
+    // This is done as follows:
+    // 1. The db and table from which partition is being deleted is assumed to be in the following format:
+    //    {prefix}{name_element}{postfix}
+    // 2. The template (format) as in #1 comes in from config and using it the {name_element} is extracted.
+    //    If no template is specified, the db / table name is assumed as the {name_element} in entirety.
+    // 3. Another template is used to determine the db / table from which replacement partition has to come in.
+    //    This template can be of the format {newPrefix}{name_element}{postFix}
+    String extractedDbName = extractNameFromTemplate(cleanableHiveDataset.getDatasetConfig(),
```
pcadabam: I recommend we use the following.  Because gobblin already uses the tokens `$DB` and `$TABLE` for 'real' database and table name.

Let's say `$DB` = events_prod and `$TABLE` = LoginEvent_Avro

```
datasetNamePattern = $LOGICAL_DB_prod.$LOGICAL_TABLE_Avro
replacementDb = $LOGICAL_DB_something_else
replacementTable = $LOGICAL_TABLE_something_else

```
From the above config we compute `$LOGICAL_DB = events` and `$LOGICAL_TABLE = LoginEvent`. `$LOGICAL_DB` and `$LOGICAL_DB` can now be used in all config keys which `HiveDataset` automatically resolves

Also the logical name computation can be moved to `gobblin.data.management.copy.hive.HiveDataset` as this will be useful to other kinds of HiveDatasets

You can have getters in `HiveDataset` to get logical db and table names which default to the real db and table name when `datasetNamePattern` is not provided

abti: As discussed offline, made real and logical db and table names citizen of HiveDataset. Also, added auto-resolution of all config properties with real and logical db and table names. 

As well as, added integration tests.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -151,4 +172,105 @@ public static String resolveTemplate(String rawString, Table table) {
...
+   * Parsed Logical DB and Table  : dbName.tableName
+   *
+   * @param datasetNamePattern    Dataset name pattern.
+   * @param dbAndTable            Source DB and Table.
+   * @param logicalDbToken        Logical DB token.
+   * @param logicalTableToken     Logical Table token.
+   * @return  Parsed logical DB and Table.
+   */
+  @VisibleForTesting
+  public static DbAndTable parseLogicalDbAndTable(String datasetNamePattern, DbAndTable dbAndTable,
```
pcadabam: If this is just for testing can you make it package protected?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -151,4 +172,105 @@ public static String resolveTemplate(String rawString, Table table) {
...
+   * Token          : $LOGICAL_TABLE
+   * Extracted Value: tableName
+   *
+   * @param sourceEntity      Source entity (typically a table or database name).
+   * @param sourceTemplate    Source template representing the source entity.
+   * @param token             Token representing the value to extract from the source entity using the template.
+   * @return Extracted token value from the source entity.
+   */
+  @VisibleForTesting
+  public static String extractTokenValueFromEntity(String sourceEntity, String sourceTemplate, String token) {
```
pcadabam: change to package protected

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
abti: Can be protected.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -151,4 +172,105 @@ public static String resolveTemplate(String rawString, Table table) {
...
+   * @param logicalTableToken     Logical Table token.
+   * @return  Parsed logical DB and Table.
+   */
+  @VisibleForTesting
+  public static DbAndTable parseLogicalDbAndTable(String datasetNamePattern, DbAndTable dbAndTable,
+      String logicalDbToken, String logicalTableToken) {
+    Preconditions.checkArgument(StringUtils.isNotBlank(datasetNamePattern), "Dataset name pattern must not be empty.");
+
+    List<String> datasetNameSplit = Lists.newArrayList(SPLIT_ON_DOT.split(datasetNamePattern));
+    Preconditions.checkArgument(datasetNameSplit.size() == 2, "Dataset name pattern must of the format: "
```
abti: You can have LoginEvent_prod and $LOGICAL_TABLE_prod pattern. 
Splitting by token will always yield two parts (because notice we do not emit empty strings), so here the splits will be prefix: "" and postfix: "_prod" which will be removed from LoginEvent_prod yielding LoginEvent. There is a test to cover this as well.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -58,14 +65,20 @@
 @ToString
 public class HiveDataset implements PrioritizedCopyableDataset {
 
+  private static Splitter SPLIT_ON_DOT = Splitter.on(".").omitEmptyStrings().trimResults();
+
   public static final String REGISTERER = "registerer";
   public static final String REGISTRATION_GENERATION_TIME_MILLIS = "registrationGenerationTimeMillis";
+  public static final String DATASET_NAME_PATTERN_KEY = "hive.datasetNamePattern";
```
abti: As discussed offline, we will have it as it is.

pcadabam:
LGTM

PR #1373, opened 2016-11-09
Address unit test flakiness

chavdar:
- ETL-5122: HadoopUtilsTest.testSafeRenameRecursively: currently, the test waits for the shutdown of executorService without actually shutting it down.
- ETL-5123: Increase timeouts in TestTunnelWithArbitraryTCPTraffic

@htran1 can you have a look?


chavdar:
Self-approving since all tests pass.

PR #1381, opened 2016-11-10
Google webmaster

enjoyear:
Implemented a connector for Google webmaster API. Google Webmaster API enables you to download data from Google Search Console for search analytics of the verified sites. See more here https://developers.google.com/webmaster-tools/. This gobblin job partitions the whole task into sub-tasks for each day. Each sub-task is handled by a GoogleWebmasterExtractor for that date, and each GoogleWebmasterExtractor holds a queue of GoogleWebmasterExtractorIterators, each of which does the query task for each filter(Currently, only the country filter is supported.) on that date. GoogleWebmasterClientImpl implements the business logic of performing API calls to download data from Google Search console.

TODO: P1. Add test coverage. P1. Add documentation for code/configurations. P2. Extend use cases.


enjoyear:
Migrated to another pull request "Chen/google webmaster copy#1422". Close this one.

PR #1386, opened 2016-11-11
Writer to launch compaction job using hive metadata

jack-moseley:
Writer to launch an `MRCompactor` job from a local gobblin job using info obtained from hive.

A workunit is created for each hive table. The extractor gets primary key field, delta key field, and file location from the hive metastore and passes these to the writer, which adds them to config for compaction. Compaction is then run on the entire folder for the table, since watermark management is not added yet.

Changes:

- Created extractor, writer, writerBuilder, compaction entity
- Modified `HiveSource` to make extractor configurable
- Fix bug with compaction tmp folder being used in two places by adding a subfolder in `DatasetsFinder`
- Move `HiveJdbcConnector` to gobblin-util

Config for use:
Source: `HiveSource`
Extractor: `HiveMetadataForCompactionExtractor`
DataWriterBuilder: `CompactionLauncherWriterBuilder`
Converter: `IdentityConverter`
Publisher: `NoopPublisher`
DatasetFinder: `SimpleDatasetsFinder`

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
@@ -48,6 +48,7 @@ public abstract class DatasetsFinder implements gobblin.dataset.DatasetsFinder<D
   public static final char DATASETS_WITH_DIFFERENT_RECOMPACT_THRESHOLDS_SEPARATOR = ';';
   public static final char DATASETS_WITH_SAME_RECOMPACT_THRESHOLDS_SEPARATOR = ',';
   public static final char DATASETS_AND_RECOMPACT_THRESHOLD_SEPARATOR = ':';
+  public static final String TMP_OUTPUT_SUBDIR = "output";
 
   protected final State state;
   protected final Configuration conf;
@@ -107,7 +108,7 @@ public abstract class DatasetsFinder implements gobblin.dataset.DatasetsFinder<D
 
   private String getTmpOutputDir() {
     return new Path(this.state.getProp(MRCompactor.COMPACTION_TMP_DEST_DIR,
-        MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR), "output").toString();
+        MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR), TMP_OUTPUT_SUBDIR).toString();
   }
 
   private FileSystem getFileSystem() {
--- a/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
@@ -106,7 +106,8 @@ public abstract class DatasetsFinder implements gobblin.dataset.DatasetsFinder<D
   }
 
   private String getTmpOutputDir() {
-    return this.state.getProp(MRCompactor.COMPACTION_TMP_DEST_DIR, MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR);
+    return new Path(this.state.getProp(MRCompactor.COMPACTION_TMP_DEST_DIR,
+        MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR), "output").toString();
   }
 
   private FileSystem getFileSystem() {
--- a/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Properties;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Joiner;
 import gobblin.compaction.listeners.CompactorListener;
 import gobblin.compaction.mapreduce.MRCompactor;
 import gobblin.metrics.Tag;
@@ -29,7 +30,7 @@ public class CompactionLauncherWriter implements DataWriter<MRCompactionEntity> 
     Properties props = new Properties();
     props.putAll(compactionEntity.getProps());
     props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY,
-        compactionEntity.getDeltaList().toString().replace("[", "").replace("]", ""));
+        Joiner.on(',').join(compactionEntity.getDeltaList()));
     props.setProperty(MRCompactor.COMPACTION_INPUT_DIR, compactionEntity.getDataFilesPath().toString());
 
     MRCompactor compactor = new MRCompactor(props, list, Optional.<CompactorListener>absent());
--- a/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
@@ -28,8 +28,9 @@ public class CompactionLauncherWriter implements DataWriter<MRCompactionEntity> 
 
     Properties props = new Properties();
     props.putAll(compactionEntity.getProps());
-    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY, compactionEntity.getDeltaInfo());
-    props.setProperty(MRCompactor.COMPACTION_INPUT_DIR, compactionEntity.getLocation());
+    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY,
+        compactionEntity.getDeltaList().toString().replace("[", "").replace("]", ""));
+    props.setProperty(MRCompactor.COMPACTION_INPUT_DIR, compactionEntity.getDataFilesPath().toString());
 
     MRCompactor compactor = new MRCompactor(props, list, Optional.<CompactorListener>absent());
     compactor.compact();
--- a/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
@@ -16,6 +16,8 @@ import gobblin.compaction.mapreduce.avro.ConfBasedDeltaFieldProvider;
 /**
  * {@link DataWriter} that launches an {@link MRCompactor} job given an {@link MRCompactionEntity} specifying config
  * for compaction
+ *
+ * Delta field is passed using {@link ConfBasedDeltaFieldProvider}
  */
 public class CompactionLauncherWriter implements DataWriter<MRCompactionEntity> {
   @Override
@@ -24,11 +26,12 @@ public class CompactionLauncherWriter implements DataWriter<MRCompactionEntity> 
 
     List<? extends Tag<?>> list = new ArrayList<>();
 
-    Properties props = compactionEntity.getProps();
-    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY, compactionEntity.getDelta());
+    Properties props = new Properties();
+    props.putAll(compactionEntity.getProps());
+    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY, compactionEntity.getDeltaInfo());
     props.setProperty(MRCompactor.COMPACTION_INPUT_DIR, compactionEntity.getLocation());
 
-    MRCompactor compactor = new MRCompactor(compactionEntity.getProps(), list, Optional.<CompactorListener>absent());
+    MRCompactor compactor = new MRCompactor(props, list, Optional.<CompactorListener>absent());
     compactor.compact();
   }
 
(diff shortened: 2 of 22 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -330,8 +331,10 @@ else if (partition.getTPartition().isSetParameters()
   @Override
   public Extractor getExtractor(WorkUnitState state) throws IOException {
     try {
-      return new HiveConvertExtractor(state, getSourceFs());
-    } catch (TException | HiveException e) {
+      return (Extractor) Class.forName(state.getProp(HIVE_SOURCE_EXTRACTOR_TYPE, DEFAULT_HIVE_SOURCE_EXTRACTOR_TYPE))
+          .getConstructor(WorkUnitState.class, FileSystem.class)
```
chavdar: Please do not non-default extractors through reflection as this creates an undocumented api. The recommended approach is to create a factory interface with a create method that takes the necessary parameters.

For example:

  interface HiveConvertExtractorFactory {
     HiveConvertExtractor createExtractor(WorkUnitState state, FileSystem sourceFs)
  } 

Then you can use something like:

  (new ClassAliasResolver<>(HiveConvertExtractorFactory.class)).resolveClass(state.getProp(HIVE_SOURCE_EXTRACTOR_TYPE, DEFAULT_HIVE_SOURCE_EXTRACTOR_TYPE)).newInstance().create(state, getSourceFs())

You can have a look at HttpClientConfiguratorLoader/HttpClientConfigurator for an example.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Properties;
+import lombok.Getter;
+
+
+/**
+ * Entity that stores information required for launching an {@link gobblin.compaction.mapreduce.MRCompactor} job
+ */
+@Getter
+public class MRCompactionEntity {
+  private final String primaryKey;
```
chavdar: Can you be more descriptive what these strings are? Are these column names? If yes, isn't it possible to have more than one primary key and/or delta column?

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
```diff
@@ -0,0 +1,23 @@
...
+  private final String delta;
+  private final String location;
```
chavdar: It's hard to understand what these "location" and "props" are.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
```diff
@@ -0,0 +1,91 @@
...
+  private MRCompactionEntity compactionEntity;
+  private boolean extracted = false;
+
+  public HiveMetadataForCompactionExtractor(WorkUnitState state, FileSystem fs) throws IOException, TException, HiveException {
+    if (Boolean.valueOf(state.getPropAsBoolean(PartitionLevelWatermarker.IS_WATERMARK_WORKUNIT_KEY))) {
+      log.info("Ignoring Watermark workunit for {}", state.getProp(ConfigurationKeys.DATASET_URN_KEY));
+      return;
+    }
+
+    HiveWorkUnit hiveWorkUnit = new HiveWorkUnit(state.getWorkunit());
```
chavdar: I see copy/paste code from HiveConvertExtractor. You should refactor and reuse code (extend HiveConvertExtractor). Ask yourself, what makes a Hive extractor different than any other extractor.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
```diff
@@ -0,0 +1,53 @@
...
+ * for compaction
+ */
+public class CompactionLauncherWriter implements DataWriter<MRCompactionEntity> {
+  @Override
+  public void write(MRCompactionEntity compactionEntity) throws IOException {
+    Preconditions.checkNotNull(compactionEntity);
+
+    List<? extends Tag<?>> list = new ArrayList<>();
+
+    Properties props = compactionEntity.getProps();
```
chavdar: Please clone the Properties collection. Modifying the existing collection creates side-effects which can lead to hard to find bugs.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -330,8 +331,10 @@ else if (partition.getTPartition().isSetParameters()
   @Override
   public Extractor getExtractor(WorkUnitState state) throws IOException {
     try {
-      return new HiveConvertExtractor(state, getSourceFs());
-    } catch (TException | HiveException e) {
+      return (Extractor) Class.forName(state.getProp(HIVE_SOURCE_EXTRACTOR_TYPE, DEFAULT_HIVE_SOURCE_EXTRACTOR_TYPE))
```
pcadabam: This seems like a good feature for all gobblin sources. Should this be moved to `gobblin.runtime.Task` instead?

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
pcadabam: +1

gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
```diff
@@ -106,7 +106,8 @@ private String getDestDir() {
   }
 
   private String getTmpOutputDir() {
-    return this.state.getProp(MRCompactor.COMPACTION_TMP_DEST_DIR, MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR);
+    return new Path(this.state.getProp(MRCompactor.COMPACTION_TMP_DEST_DIR,
+        MRCompactor.DEFAULT_COMPACTION_TMP_DEST_DIR), "output").toString();
```
ydai1124: Making "output" a constant.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
```diff
@@ -0,0 +1,53 @@
...
+    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY, compactionEntity.getDelta());
```
ydai1124: Since you are using `ConfBasedDeltaFieldProvider`, please add it to the java doc.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
ydai1124: Besides Chavdar's comment, if `primaryKey` is present,  do you need to set `COMPACTION_JOB_DEDUP_KEY`? Otherwise, the compactor will use the default `ALL` option for dedupe.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
jack-moseley: Added `HiveBaseExtractorFactory`, `HiveMetadataForCompactionExtractorFactory`, and `HiveConvertExtractorFactory` and using them to create extractor now

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
jack-moseley: Updated javadoc. Multiple primary/delta columns will be passed by comma delimited list

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
jack-moseley: Updated javadoc

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
jack-moseley: Added `HiveBaseExtractor` as a parent method for both extractors. However it is kind of awkward to try to reuse code in the constructors since there are several differences. Do you think it's worth adding several one line helper methods (e.g. for getting `HiveWorkUnit`) to avoid reusing code?

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
jack-moseley: Fixed

gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
jack-moseley: Fixed

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
jack-moseley: Added

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
jack-moseley: We will have to set that key, but I don't think there is much point now since passing primary keys as a string isn't implemented yet

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
```diff
@@ -0,0 +1,28 @@
...
+ * Entity that stores information required for launching an {@link gobblin.compaction.mapreduce.MRCompactor} job
+ *
+ * {@link #primaryKeyInfo}: Comma delimited list of fields to use as primary key
+ * {@link #deltaInfo}: Comma delimited list of fields to use as deltaInfo
+ * {@link #location}: Location of files associated with table
+ * {@link #props}: Other properties to be passed to {@link gobblin.compaction.mapreduce.MRCompactor}
+ */
+@Getter
+public class MRCompactionEntity {
+  private final String primaryKeyInfo;
```
chavdar: I suggest that you use make the object stronger typed. If primaryKeyInfo and deltaInfo are lists of field names, then declare them as a List<String>, if location is a Path, then declare it as such. The fact that the current implementation of MRCompactor takes a comma-separated list is an implementation detail. In the future, if we move it to using typesafe config, then we can pass directly lists.

chavdar: Further, a name like primaryKeysList is more descriptive then primaryKeyInfo. Same for dataFilesPath vs location.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -330,8 +333,10 @@ else if (partition.getTPartition().isSetParameters()
   @Override
   public Extractor getExtractor(WorkUnitState state) throws IOException {
     try {
-      return new HiveConvertExtractor(state, getSourceFs());
-    } catch (TException | HiveException e) {
+      return (new ClassAliasResolver<>(HiveBaseExtractorFactory.class))
```
jack-moseley: fixed

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/MRCompactionEntity.java
jack-moseley: Updated names and types

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
```diff
@@ -28,8 +28,9 @@ public void write(MRCompactionEntity compactionEntity) throws IOException {
 
     Properties props = new Properties();
     props.putAll(compactionEntity.getProps());
-    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY, compactionEntity.getDeltaInfo());
-    props.setProperty(MRCompactor.COMPACTION_INPUT_DIR, compactionEntity.getLocation());
+    props.setProperty(ConfBasedDeltaFieldProvider.DELTA_FIELDS_KEY,
+        compactionEntity.getDeltaList().toString().replace("[", "").replace("]", ""));
```
chavdar: Joiner.on(',').join(compactionEntity.getDeltaList());

chavdar:
One minor comment, other than that looks good.

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/CompactionLauncherWriter.java
jack-moseley: fixed

PR #1387, opened 2016-11-11
Completeness verification events

pcadabam:
@ydai1124 can you review?


gobblin-compaction/src/main/java/gobblin/compaction/verify/DataCompletenessVerifier.java
```diff
@@ -153,10 +155,19 @@ public Results(Iterable<Result> results) {
 
       private final Dataset dataset;
       private final Status status;
+      /** Data used to compute this result **/
```
chavdar: This is very generic description. Can you please clarify.

pcadabam: The exact data used for verification is implementation specific. The `DataCompletenessVerifier` does not know how/what a specific verifier used to compute this result.

PR #1388, opened 2016-11-11
Fix a race condition in IteratorExecutorTest

chavdar:
Previously, endOneCallable could signal multiple times before endCall could be processed.


ibuenros:
Nice catch!

PR #1389, opened 2016-11-12
Implemented the source and extractor for Google Webmaster connector

enjoyear:
Implemented a connector for Google webmaster API. Google Webmaster API enables you to download data from Google Search Console for search analytics of the verified sites. See more here https://developers.google.com/webmaster-tools/. This gobblin job partitions the whole task into sub-tasks for each day. Each sub-task is handled by a GoogleWebmasterExtractor for that date, and each GoogleWebmasterExtractor holds a queue of GoogleWebmasterExtractorIterators, each of which does the query task for each filter(Currently, only the country filter is supported.) on that date. GoogleWebmasterClientImpl implements the business logic of performing API calls to download data from Google Search console.


chavdar:
Can you have a look at the travis results? There are unit test failures:

   java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at gobblin.source.extractor.extract.google.webmaster.GoogleWebmasterExtractorIteratorTest.testIterator(GoogleWebmasterExtractorIteratorTest.java:79)

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterClientImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    //GoogleNetHttpTransport.newTrustedTransport(),
+    //JacksonFactory.getDefaultInstance(),
+    Webmasters service =
+        new Webmasters.Builder(new NetHttpTransport(), new JacksonFactory(), credential).setApplicationName(appName)
+            .build();
+    _analytics = service.searchanalytics();
+  }
+
+  @Override
+  public List<String> getPages(String siteProperty, String date, GoogleWebmasterFilter.Country country, int rowLimit,
```
chavdar: I think it may be an easy and worthwhile change to return a Callable<> which does the actual call. Thus we can later parallelize the calls in the fetcher getAllPages().

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterClientImpl.java
```diff
@@ -0,0 +1,98 @@
...
+        pages.add(row.getKeys().get(pageIndex));
+      }
+    } else {
+      LOG.info(String.format("0 pages fetched for market-%s on %s.", country, date));
+    }
+    return pages;
+  }
+
+  @Override
+  public SearchAnalyticsQueryResponse searchAnalyticsQuery(String siteProperty, String date,
```
chavdar: As before, consider returning a Callable&lt;SearchAnalyticsQueryResponse&gt;

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,291 @@
...
+    List<ApiDimensionFilter> filters = new LinkedList<>();
+    filters.add(countryFilter);
+    filters.add(null); //placeholder for page filter
+
+    Set<String> uniquePages = new HashSet<>();
+    Deque<String> toProcess = new LinkedList<>(pagePrefixes);
+    while (!toProcess.isEmpty()) {
+      String prefix = toProcess.remove();
+      LOG.info("Current page prefix is " + prefix);
+      filters.set(1, GoogleWebmasterFilter.pageFilter(GoogleWebmasterFilter.FilterOperator.CONTAINS, prefix));
```
chavdar: The overhead of creating a new filters array is not that high (compared to making the getPages() remote call). I think it is safer to create a new instance. That is necessary if we want to parallelize the getPages() calls in the future.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -0,0 +1,112 @@
...
+   */
+  public static final String KEY_REQUEST_HYPERCRITICAL = "source.google_webmasters.request.isHypercritical";
+  /**
+   * Give a Google API service scope.
+   * For Webmaster, only two scopes are supported. WebmastersScopes.WEBMASTERS_READONLY and WebmastersScopes.WEBMASTERS
+   */
+  public static final String KEY_API_SCOPE = "source.google.api.scope";
+  /**
+   * The filters that will be passed to all your API requests.
+   * Filter format is [GoogleWebmasterFilter.Dimension].[DimensionValue]
```
chavdar: Any reason not to use the regex [GoogleWebmasterFilter.Dimension](=)[DimensionValue] ? It's going to be easier to add new predicates later, especially if they are not single character.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterFilter.java
```diff
@@ -0,0 +1,317 @@
...
+  private static ApiDimensionFilter build(String dimension, String operator, String expression) {
+    return new ApiDimensionFilter().setDimension(dimension).setOperator(operator).setExpression(expression);
+  }
+
+  static ApiDimensionFilter pageFilter(FilterOperator op, String expression) {
+    //Operator string is case insensitive
+    return build(Dimension.PAGE.toString(), op.toString(), expression);
+  }
+
+  static ApiDimensionFilter countryFilter(Country country) {
```
chavdar: I suggest using countryEqFilter to make the predicate explicit.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterExtractorIterator.java
```diff
@@ -0,0 +1,103 @@
...
+      _totalPages = _cachedPages.size();
+      _pageCheckPoint = Math.max(1, (int) Math.round(Math.ceil(_totalPages / 20.0)));
+    }
+  }
+
+  public String[] next() throws IOException {
+    if (hasNext()) {
+      return _cachedQueries.remove();
+    }
+    return null;
```
chavdar: The convention for iterators is to throw NoSuchElementException .

enjoyear:
Hi Chavdar,

The unit test failure was caused by a non-determined order of HashMap
values. I've fixed the failure by adding a customized Matcher for
CollectionEquals.

class CollectionEquals extends ArgumentMatcher<Collection> {

  private final Collection _expected;

  public CollectionEquals(Collection expected) {
    _expected = expected;
  }

  @Override
  public boolean matches(Object actual) {
    return CollectionUtils.isEqualCollection((Collection) actual, _expected);
  }
}


It took me a while to figure out how to fix this. I feel like other people
may experience the same issue. So this could be moved to a better place.

Thanks,
Chen

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -0,0 +1,112 @@
...
+      JsonElement jsonElement = schemaJson.get(i);
+      String columnName = jsonElement.getAsJsonObject().get("columnName").getAsString().toUpperCase();
+      columnPositionMap.put(columnName, i);
+    }
+    ValidateRequests(columnPositionMap, requestedDimensions, requestedMetrics);
+
+    return new GoogleWebmasterExtractor(state, columnPositionMap, requestedDimensions, requestedMetrics);
+  }
+
+  private void ValidateRequests(Map<String, Integer> columnPositionMap,
```
enjoyear: fixed

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebMasterSource.java
enjoyear: An example of the filter in the configuration file is Country.USA, where Country is GoogleWebmasterFilter.Dimension, USA is the value for the Country Dimension. Not quite sure what do you mean by GoogleWebmasterFilter.Dimension[DimensionValue]

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -0,0 +1,112 @@
...
+import gobblin.source.extractor.Extractor;
+import gobblin.source.extractor.extract.QueryBasedSource;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+public class GoogleWebMasterSource extends QueryBasedSource<String, String[]> {
```
enjoyear: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterClient.java
```diff
@@ -0,0 +1,46 @@
...
+   * @return Return all pages given all constraints.
+   * @throws IOException
+   */
+  List<String> getPages(String siteProperty, String date, GoogleWebmasterFilter.Country country, int rowLimit,
+      List<GoogleWebmasterFilter.Dimension> requestedDimensions, List<ApiDimensionFilter> filters, int startRow)
+      throws IOException;
+
+  /**
+   * Perform the api call for search analytics query
+   * @param siteProperty
```
enjoyear: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterFilter.java
enjoyear: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,291 @@
...
+   * 2. Now we need to know a full list of unique page prefixes/root paths. We will keep asking for the maximum number(5000) of rows until it comes to the last block, which is indicated by response giving less than 5000 rows.
+   * 3. Remove duplicated pages got in last step.
+   * 4. Parse out the list of keywords from last step.
+   * 5. Send another request by adding the filters to exclude any pages/keywords in last step.
+   * 6. Repeat steps 4 to 5 until we get a full union list of keywords, or prefixes, or root paths(e.g. www.property.com/path1/, www.property.com/path2/)
+   * 7. For each page root path, apply the logic in pagesWithRootPaths to get all pages with that page root path/prefix.
+   *    In detail, for example, you want all pages for your https://www.linkedin.com/ property. This property has several page prefixes such as https://www.linkedin.com/topic/, https://www.linkedin.com/jobs/, https://www.linkedin.com/groups/ etc.. For each prefix, you request for all pages starting with that prefix.
+   * 8. Union all pages from last step and return.
+   *
+   * @param date
```
enjoyear: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Good point. Fixed.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,291 @@
...
+        for (char c = 'a'; c <= 'z'; ++c) {
+          toProcess.add(prefix + c);
+        }
+        for (int num = 0; num <= 9; ++num) {
+          toProcess.add(prefix + num);
+        }
+        toProcess.add(prefix + "%");
+        toProcess.add(prefix + "_");
+        toProcess.add(prefix + "-");
+        toProcess.add(prefix + "/");
```
enjoyear: This is not an exhaustive list. In reality, only prefix + "%" is enough. Checking for the exhaustive list is too expensive, so I only checked four most likely characters. I may need to think about a different way and rewrite the whole piece here.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterClientImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    if (pageRows != null) {
+      LOG.info(String.format("%d pages fetched for market-%s on %s. The last page has %.1f clicks.", pageRows.size(),
+          country, date, pageRows.get(pageRows.size() - 1).getClicks()));
+
+      int pageIndex = requestedDimensions.indexOf(GoogleWebmasterFilter.Dimension.PAGE);
+      for (ApiDataRow row : pageRows) {
+        pages.add(row.getKeys().get(pageIndex));
+      }
+    } else {
+      LOG.info(String.format("0 pages fetched for market-%s on %s.", country, date));
```
enjoyear: This LOG.info won't be called often, maybe 4~5 times per market per day. And the information is kind of valuable for now. Instead, I've removed logging in doQuery() at GoogleWebmasterDataFetcherImpl. They are too trivial to keep.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebmasterClientImpl.java
enjoyear: As before.

gobblin-core/src/main/java/gobblin/source/extractor/extract/google/webmaster/GoogleWebMasterSource.java
chavdar: It seems like github mis-formatted my comment. I suggested using a regex to match the predicate so that it is extensible, say:

\(\w+\)\(=\)\(.*\)
The first group corresponds to the dimension, the second corresponds to the predicate and the third one corresponds to the constant.

enjoyear:
Migrated to another pull request "Chen/google webmaster copy#1422". Close this one.

PR #1391, opened 2016-11-14
Ignore lookback in avro to orc backfill source

pcadabam:
@abti can you reivew?

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/BackfillHiveSource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/BackfillHiveSource.java
@@ -74,4 +74,15 @@ public class BackfillHiveSource extends HiveAvroToOrcSource {
     // If no whitelist is set, all partitions of a dataset are backfilled
     return true;
   }
+
+  @Override
+  public boolean isOlderThanLookback(Partition partition) {
+    // If partition whitelist is provided, ignore lookback
+    if (!this.partitionsWhitelist.isEmpty()) {
+      return false;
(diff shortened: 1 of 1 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/BackfillHiveSource.java
```diff
@@ -74,4 +74,15 @@ public boolean shouldCreateWorkunit(Partition sourcePartition, LongWatermark low
     // If no whitelist is set, all partitions of a dataset are backfilled
     return true;
   }
+
+  @Override
+  public boolean isOlderThanLookback(Partition partition) {
```
abti: Seems like a confusing behavior. Lookback can / should be applicable for whitelisted partition(s) as well.

pcadabam: Lookback is used to select partitions for conversion. If the user provides list of partition via this whitelist, the lookback becomes irrelevant.

chavdar:
@abti @pcadabam  What is the status of this?

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/BackfillHiveSource.java
abti: As discussed offline, when specific partitions are whitelisted no discovery / partition filtering needs to happen since we already know what partitions to process. Hence ignoring lookback makes sense. Thanks for the clarification.

PR #1399, opened 2016-11-17
Catch socket exception and suppress connection reset error to prevent…

htran1:
… the TalkFirstDoubleEchoServer from aborting when the client socket is closed.  There is a test case that closes the client socket before receiving the server response.


htran1:
@chavdar can you review?

PR #1400, opened 2016-11-18
Fix bytesWritten in FsDataWriter

ydai1124:
Current `AvroHdfsDataWriter` checks `outputFile` length to get the `bytesWritten`. This is a issue when `TaskPublisher` is enabled, because `outputFile` is moved by `TaskPublisher`, and `bytesWritten` always returns 0 or throws `FileNotFoundException`.

To fix this issue, `FileStatus` is recorded in `FsDataWriter`, and so is `bytesWritten`. Also, `synchronized` in the method `bytesWritten()` is no longer necessary here, which is added in #494.


ydai1124:
@ibuenros  or @chavdar  can you review?

PR #1401, opened 2016-11-18
Get rid of fixed ZK ports in unit tests

chavdar:
TestingServer allows using of port -1 which means -- pick a random port. We just have to make sure that job configs are modified dynamically.

@abti Can you verify?


abti:
I couldn't find documentation around port = -1, but creating a TestServer without specifying any port number seems to create a Zookeeper with random port: http://netflix.github.io/curator/doc/com/netflix/curator/test/TestingServer.html#TestingServer() 

However, we will have to check that if the random port which TestingServer decides to use is already bound, then does: 
- TestingServer fails (thus causing potentially flaky intermittently failing tests)
- OR, TestingServer tries to fallback to another random port

chavdar:
I found the behavior by looking at the code. Internally, using new TestingServer(-1), uses new ServerSocket(0) to allocate an available port. There is a small race condition between we find the port and the time it is used by ZK but there is not much we can do about it.

PR #1405, opened 2016-11-18
Retention based Data replication 

tuGithub:


tuGithub:
1. ReplicationDataRetentionCategory specify whether the copy is "Sync", "limited snapshots copy", "limited daily partitions copy" or "limited hourly partitions copy"
2. ReplicationDataValidPathPicker will use GlobModTimeDatasetVersionFinder or DateTimeDatasetVersionFinder to find the valid paths to copy from source to target

chavdar:
There are FindBugs warnings:

> Execution failed for task ':gobblin-data-management:findbugsMain'

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/HadoopFsEndPointDataset.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import gobblin.dataset.FileSystemDataset;
+import gobblin.util.HadoopUtils;
+
+
+public class HadoopFsEndPointDataset implements FileSystemDataset{
```
chavdar: CAn you please explain/document this class?

tuGithub: Tried to use the VersionFinder to select valid sub directory to replicate. The AbstractDatasetVersionFinder.findDatasetVersions function take Dataset as input. So I created this class as the Dataset wrapper for HadoopFsEndPoint

chavdar: It is useful to add the above as part of the javadocs.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationDataRetentionCategory.java
```diff
@@ -0,0 +1,78 @@
...
+ *  <li>finite_daily_partition : only limited daily partitions in source need to be copied to target
+ *  <li>finite_daily_partition : only limited hourly partitions in source need to be copied to target
+ *  <li>finite_snapshot: only limited snapshot directories in source need to be copied to target
+ * </ul>
+ * @author mitu
+ *
+ */
+@AllArgsConstructor
+public class ReplicationDataRetentionCategory {
+  enum Type{
```
chavdar: Why do we need the Type enum instead of implementations of the same interface? Can't this functionality be part of the source dataset implementation?

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationDataValidPathPicker.java
```diff
@@ -0,0 +1,85 @@
...
+import gobblin.dataset.FileSystemDataset;
+
+/**
+ * Used to pick the valid Paths for data replication based on {@link #ReplicationDataRetentionCategory.Type}
+ * @author mitu
+ *
+ */
+public class ReplicationDataValidPathPicker {
+
+  public static Collection<Path> getValidPaths(HadoopFsEndPoint hadoopFsEndPoint) throws IOException{
```
chavdar: The current implementation supports a few hand-picked DatasetVersionFinders and selection policies for the ones to be copied.

The retention management supports many different DatasetVersionFinders (and new ones can be added) plus different selection policies (top K,  before, after date, etc.) It is worthwhile to talk to @pcadabam and see how we can use those in the case of replication in a more generic manner.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationDataRetentionCategory.java
tuGithub: Removed that class and have config based selection

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ReplicationDataValidPathPicker.java
tuGithub: have the config based selection

chavdar:
There seem to be unit test failures in ReplicationConfigurationTest

chavdar:
Please make sure that unit tests pass. Other than that, LGTM

tuGithub:
fixed the unit testing

PR #1406, opened 2016-11-19
Fixed bit and boolean column types in JDBC extractor.

ibuenros:


ibuenros:
@chavdar can you review?

PR #1408, opened 2016-11-19
Handle race conditions involving rename in LocalFileSystem.

htran1:
One race condition is when a directory D1/a/b/c/d is renamed to D2/a/b/c/d after another thread created D2/a/b/c/d.
This resulted in D2/a/b/c/d/d. Handle this by using the Java renameTo() method to avoid the LocalFileSystem fallback of copy.
The copy fallback also results in a race condition when a file is opened for copy in one thread and the parent directory is createed in another thread. A file not found exception is encountered in this case. This case is handled by catching the exception and retrying a few times.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
+++ b/gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
@@ -16,6 +16,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -46,6 +47,7 @@ import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.FileUtil;
+import org.apache.hadoop.fs.LocalFileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.io.IOUtils;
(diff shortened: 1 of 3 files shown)


htran1:
@chavdar can you review?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -549,6 +552,26 @@ public static boolean unsafeRenameIfNotExists(FileSystem fs, Path from, Path to)
       if (!fs.exists(to.getParent())) {
         fs.mkdirs(to.getParent());
       }
+
+      boolean renamed;
```
chavdar: This does not seem to be used. Please remove.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -549,6 +552,26 @@ public static boolean unsafeRenameIfNotExists(FileSystem fs, Path from, Path to)
...
+        if (fs instanceof LocalFileSystem) {
+          File srcFile = ((LocalFileSystem) fs).pathToFile(from);
+          File dstFile = ((LocalFileSystem) fs).pathToFile(to);
+
+          // failure to rename directory occurs when target name already exists
+          return srcFile.renameTo(dstFile);
+        }
+        else if (fs instanceof RateControlledFileSystem &&
+              ((RateControlledFileSystem) fs).getDecoratedObject() instanceof LocalFileSystem) {
+          return fs.rename(from, to);
```
chavdar: Why not use the above method?

gobblin-utility/src/main/java/gobblin/util/RateControlledFileSystem.java
```diff
@@ -139,8 +141,24 @@ public boolean mkdirs(Path path, FsPermission permission) throws IOException {
 
   @Override
   public boolean rename(Path path0, Path path1) throws IOException {
+    boolean renamed;
+
     this.acquirePermit();
-    return this.fs.rename(path0, path1);
+
+    // Use Java rename when renaming a directory for LocalFileSystem.
```
chavdar: This code appears in two places. Please consider refactoring it into a function.

PR #1410, opened 2016-11-21
Change instance check to class name check since returned instance fro…

htran1:
…m cache is not guaranteed to match.


htran1:
@chavdar please review.

PR #1412, opened 2016-11-21
Implemented Avro field replacer converter.

ibuenros:
This implements a converter that can replace string fields in Avro records for other string, supporting custom replacements. Also implemented two default replacements: constant and sha1.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/converter/avro/replacer/AvroFieldReplacer.java
+++ b/gobblin-core/src/main/java/gobblin/converter/avro/replacer/AvroFieldReplacer.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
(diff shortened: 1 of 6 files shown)


ibuenros:
@pcadabam can you review?

chavdar:
I'd like to see a more detailed description of the design. 

- How do we make this dataset config driven?
- Do we allow "." notation of fields (i.e. traverse avro hierarchy)
- How do we handle different field types?

shirshanka:
We seem to have accumulated quite a few different converters for doing field level operations on records. Instead of chaining these converters by hand and specifying config independently for each, maybe we should have an uber converter for supporting config-driven transformation of the record.

gobblin-core/src/main/java/gobblin/converter/avro/replacer/AvroFieldReplacer.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
```
abti: Please change to use the Apache 2.0 License header without copyright

abti:
https://issues.apache.org/jira/browse/GOBBLIN-74

Please update your PR title with following prefix: [GOBBLIN-74]

PR #1417, opened 2016-11-22
Moving Kafka dependencies out into version specific modules

shirshanka:
@ibuenros @pcadabam : This is probably not the final PR, but would like to get a review to see if things look similar to what we discussed. 

Tests are passing.


chavdar:
@shirshanka can you describe the change at a high level?

shirshanka:
@chavdar Travis seems to be happy.

shirshanka:
@chavdar : Care to review now? 
I have tried to extract as much code as possible into gobblin-kafka-common, leaving bare minimum in the 08 and 09 directories. 

I'm not sure why the coverage build is failing. 
I do want to add an integration test for 09 to make sure that the writer works with an 09 cluster.

shirshanka:
High-level summary of changes: 

1) A top-level directory called gobblin-modules which will be used to store any datasystem specific integrations such as kafka. 
2) gobblin-modules/gobblin-kafka-common: a module that contains classes that are related to Kafka integration but do not depend on Kafka. 
3) gobblin-modules/gobblin-kafka-08: classes that allow reading, writing to Kafka 0.8 and reporting metrics to Kafka 0.8
4) gobblin-modules/gobblin-kafka-09: future home of readers, writers, reporters for Kafka 0.9. Currently only hosts the KafkaWriter.
5) gobblin-core-base: datasystem independent classes extracted from gobblin-core that help in creating datasystem integrations
6) gobblin-metrics : slimmed down and no longer depends on kafka or bundles the kafka reporter
7) gobblin-core: to preserve backwards compatibility, pulls in gobblin-kafka-08 using the standard flavored build for gobblin. This module can only depend on either gobblin-kafka-08 or gobblin-kafka-09. This can be controlled through the -DgobblinFlavor build argument and the respective gradle file (e.g. gobblin-core/gobblin-flavor-standard.gradle, gobblin-core/gobblin-flavor-full.gradle, ..)

shirshanka:
@abti : What better time to check in massive changes than Fri midnight?

build.gradle
```diff
@@ -47,6 +47,15 @@ idea.project {
   languageLevel = JavaVersion.VERSION_1_7
 }
 
+// Gobblin build customization
+if (!project.hasProperty('gobblinFlavor')) {
```
abti: It seems there are a few distinct supported flavors, can we document them somewhere to be explicit about what version does what?

gobblin-metrics/src/main/java/gobblin/metrics/GobblinMetrics.java
```diff
@@ -728,33 +680,43 @@ private void buildCustomMetricReporters(Properties properties) {
...
+        com.codahale.metrics.ScheduledReporter scheduledReporter = this.codahaleReportersCloser
+            .register(customCodahaleReporterFactory.newScheduledReporter(RootMetricContext.get(), properties));
+        String reporterSinkMsg = reporterSink.isPresent()?"to " + reporterSink.get():"";
+        LOGGER.info("Will start reporting metrics " + reporterSinkMsg + " using " + reporterClass);
+        this.codahaleScheduledReporters.add(scheduledReporter);
+
+      } else if (CustomReporterFactory.class.isAssignableFrom(clazz)) {
+        CustomReporterFactory customReporterFactory = ((CustomReporterFactory) clazz.getConstructor().newInstance());
+        customReporterFactory.newScheduledReporter(properties);
+
```
abti: nit: a log message for easier tracking that this custom reporter was initialized

build.gradle
shirshanka: Probably makes sense to add this in build documentation and integration docs for kafka etc. 
Will open separate PR for that later.

PR #1421, opened 2016-11-23
Interleave gc() calls with memory allocations to improve reliability …

htran1:
…of garbage collection test.


htran1:
@chavdar can you review?

PR #1422, opened 2016-11-23
Chen/google webmaster copy

enjoyear:
This pull request is a copy of my branch chen/google-webmaster. Previously it was merged using "git rebase" and got duplicated commits. Now I am merging with "git merge", so recreate a new pull request here.


shirshanka:
@enjoyear : Can you move the code under gobblin-modules/google-webmaster ? 

We're trying to not have gobblin-core be a kitchen sink of integrations for every data system. 

You can look at the gobblin-modules/gobblin-kafka-08  module as an example.

enjoyear:
@shirshanka This is done. Please review my last commit(7a5fe1f).

chavdar:
@enjoyear google-ingestion/ is a bit too generic a name. As @shirshanka suggested, google-webmaster or google-source-webmaster are more descriptive.

enjoyear:
@chavdar this is just a beginning. You just see gobblin.ingestion.google.webmaster for now. There are more coming. For example, the next one is gobblin.ingestion.google.adwords.

chavdar:
I see. Do they share code/libraries? If yes, it makes sense to keep them in the same module.

enjoyear:
@chavdar Actually, only the credential/login part share quite similar code. The rest of the code depends on different APIs/jars provided by google.

enjoyear:
@chavdar By the way, the code under gobblin-core/src/main/java/gobblin/source/extractor/extract/google should be moved under this google-ingestion module as well. Ideally, this refactoring can start after this pull request is finished.

chavdar:
YEs, we are in the process of refactoring the code and taking dependencies out.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+      while (!toProcess.isEmpty()) {
+        submitJob(toProcess.poll(), countryFilter, startDate, endDate, dimensions, es, allPages, nextRound);
+        try {
+          Thread.sleep(275); //Submit roughly 4 jobs per second.
+        } catch (InterruptedException ignored) {
+        }
+      }
+      //wait for jobs to finish and start next round if necessary.
+      try {
+        es.shutdown();
```
chavdar: ExecutorUtils has a helper method for shutting down ExecutorServices.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+    ConcurrentLinkedDeque<String> allPages = new ConcurrentLinkedDeque<>();
+    Random random = new Random();
+    //we need to retry many times because the path may be very long
+    final int retry = 120;
+    int r = 0;
+    while (r <= retry) {
+      ++r;
+      LOG.info(String.format("Get pages at round %d with size %d.", r, toProcess.size()));
+      ConcurrentLinkedDeque<Pair<String, FilterOperator>> nextRound = new ConcurrentLinkedDeque<>();
+      ExecutorService es = Executors.newFixedThreadPool(10, new ThreadFactory() {
```
chavdar: As before, check out ExecutorUtils

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+  private Collection<String> getPages(String startDate, String endDate, List<Dimension> dimensions,
+      ApiDimensionFilter countryFilter, Queue<Pair<String, FilterOperator>> toProcess) throws IOException {
+    String country = GoogleWebmasterFilter.countryFilterToString(countryFilter);
+
+    ConcurrentLinkedDeque<String> allPages = new ConcurrentLinkedDeque<>();
+    Random random = new Random();
+    //we need to retry many times because the path may be very long
+    final int retry = 120;
+    int r = 0;
+    while (r <= retry) {
```
chavdar: This seems to be fairly complicated with a few arbitrary constants. I don't also think that the concept of rounds is necessary. I suggest having a look at https://github.com/google/guava/wiki/ListenableFutureExplained .

Here is one option:
- You keep two global counters: runningJobs and failedJobs and a Condition variable jobFinished.
- Convert the implementation submitJob to a Callable or Runnable object with all current parameters plus a retry counter. The object also implements FutureCallback:
     - onSuccess, if there are new jobs to be submitted, it will call submitJob() recursively for them; then it decreases the runningJobs counter; signal jobFinished
     - onFailure, it will increase the retry counter; if too, many retries, it will increase the failedJobs counter, decrease the running jobsCounter. If there are more, retries, it re-submits the job; signal jobFinished
- submitJob itself will:
   - Increase the runningJobs counter
   - Submit a ListenableFuture with the above Callable/Runnable and add the future as a callback to itself.
- getPages() submits the initial job(s) to the executor service and then just waits on the condition jobFinished until either 0 == runningJobs or 0 < failedJobs .

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+
+  private int getPagesSize(final String startDate, final String endDate, final String country,
+      final List<Dimension> requestedDimensions, final List<ApiDimensionFilter> apiDimensionFilters)
+      throws IOException {
+
+    int startRow = 0;
+    int count = 4;
+    final ExecutorService es = Executors.newFixedThreadPool(count);
+    int r = 0;
+    while (r < 50) {
```
enjoyear: No. Those are not random numbers. Those empirical number took me a long time to achieve the best balance among speed, retries and reliability. Check details at this PPT https://iwww.corp.linkedin.com/wiki/cf/display/DWH/Google+Search+Console+Data+Ingestion#GoogleSearchConsoleDataIngestion-PPTPresentation

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -0,0 +1,178 @@
...
+ * The minimum unit of querying range is date. Change the range by configuring "source.querybased.start.value" and "source.querybased.end.value". Note that the analytics data for Google Search Console has a delay or 3 days. So cap your configuration of "source.querybased.append.max.watermark.limit" by "CURRENTDATE-3". See the documentation details of each configuration in the GoogleWebMasterSource fields.
+ *
+ */
+abstract class GoogleWebMasterSource extends QueryBasedSource<String, String[]> {
+
+  /**
+   * Must Provide.
+   * Provide the property site URL whose google search analytics data you want to download
+   */
+  public static final String KEY_PROPERTY = "source.google_webmasters.property";
```
enjoyear: Done

gobblin-modules/google-ingestion/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+  compile externalDependency.typesafeConfig
+  compile externalDependency.findBugsAnnotations
+
+  testCompile project(":gobblin-test-utils")
+  testCompile externalDependency.jsonAssert
+  testCompile externalDependency.mockito
+  testCompile externalDependency.testng
+}
+
+configurations {
```
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: In short, there is a quota limit. You need to control the speed of your requests. Otherwise, you will run into "disasters".

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+                  return pages.size() + start;
+                } else {
+                  //The size is not determined.
+                  return -1;
+                }
+              } catch (IOException e) {
+                LOG.info(String.format("Getting page size from %s failed. Retrying...", start));
+              }
+              try {
+                Thread.sleep(200);
```
enjoyear: No. This will not retry forever. Check this line **Integer integer = result.get(2, TimeUnit.MINUTES);**. The job will timeout in two minutes. But I will improve the thread exit based on interruption. And in this case, I don't need exponential back-off. Linear is good enough for me. Thank you for pointing that out, I will utilize that library in the future when I need exponential back-off.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+          Thread t = Executors.defaultThreadFactory().newThread(r);
+          t.setDaemon(true);
+          return t;
+        }
+      });
+
+      while (!toProcess.isEmpty()) {
+        submitJob(toProcess.poll(), countryFilter, startDate, endDate, dimensions, es, allPages, nextRound);
+        try {
+          Thread.sleep(275); //Submit roughly 4 jobs per second.
```
enjoyear: This is not a google-mandated limit, but a good number that I realized during testing. It can suppress a lot of retries.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: As before, ExecutorUtils doesn't seem to provide fixed daemon thread pool.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: I tried to use the helper method, however, it doesn't support customized error/warning message for logs. So I cannot use it here.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterExtractor.java
```diff
@@ -0,0 +1,221 @@
...
+    return new ArrayList<>();
+  }
+
+  private static Credential getCredential(WorkUnitState wuState) {
+    String scope = wuState.getProp(GoogleCommonKeys.API_SCOPES, WebmastersScopes.WEBMASTERS_READONLY);
+    Preconditions.checkArgument(
+        Objects.equals(WebmastersScopes.WEBMASTERS_READONLY, scope) || Objects.equals(WebmastersScopes.WEBMASTERS,
+            scope), "The scope for WebMaster must either be WEBMASTERS_READONLY or WEBMASTERS");
+
+    String credentialFile = wuState.getProp(SOURCE_CONN_PRIVATE_KEY);
```
enjoyear: The SOURCE_CONN_PRIVATE_KEY here is a path to the private key file, which will be kept at a place with limited access. So does the master password file in your case. But thank you for sharing this because my next project will need this feature.

chavdar: In that case, can you just use a different property instead, say, SOURCE_CONN_PRIVATE_KEY_PATH ?

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,366 @@
...
+    return jobs;
+  }
+
+  private int getPagesSize(final String startDate, final String endDate, final String country,
+      final List<Dimension> requestedDimensions, final List<ApiDimensionFilter> apiDimensionFilters)
+      throws IOException {
+
+    int startRow = 0;
+    int count = 4;
+    final ExecutorService es = Executors.newFixedThreadPool(count);
```
chavdar: Executors.newFixedThreadPool(count, ExecutorUtils.newDaemonThreadFactory(..))

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
chavdar: It's still not completely clear for me but you've done your homework. Please make those named constants and if possible add some intuition why they are what they are. Some of those magic constants my be driven by the Google implementation which may change in the future.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
chavdar: I am still not 100% sure what the code optimizes for. Is it number of concurrent requests, request rate/thread, total request rate, etc? I'll try to read your presentation more carefully but adding some comments in the code will help.

If you want to limit the number of concurrent requests, a fixed thread pool will do.
If you want to control the request submission rate, I'd claim that single job queue is a better approach. 

The rate of job submission in the thread-pool need not be == the request submission rate. You can use a gobblin.util.limiter.RateBasedLimiter before sending the request. That will make sure that you don't submit requests at a higher rate. That will guarantee you that you don't go over the limit yet you'll have more consistent flow of requests. I find the current Sleep-based approach fragile and the submission rate can vary greatly depending on the latency to send a request.

In the current, round-based approach, a single round can be held up by a single request that is taking longer to process on the Google side. During that time, you are not sending any requests.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
chavdar: It gives you a thread factory that you can pass to the thread pool.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
chavdar: Ok

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Oops, my bad... Thank you.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcher.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public static List<String[]> convertResponse(List<Metric> requestedMetrics, SearchAnalyticsQueryResponse response) {
+    List<ApiDataRow> rows = response.getRows();
+    if (rows == null || rows.isEmpty()) {
+      return new ArrayList<>();
+    }
+    List<String[]> ret = new ArrayList<>(rows.size());
+    for (ApiDataRow row : rows) {
+      List<String> keys = row.getKeys();
+      String[] data = new String[keys.size() + 4];
```
abti: requestedMetrics might be lesser or more than 4 now / or in future. So, initializing data[] with keys.size() + 4 might run into an issue. how about either making it dynamic or applying a Precondition and documenting this

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,383 @@
...
+                  return -1;
+                }
+              } catch (IOException e) {
+                LOG.info(String.format("Getting page size from %s failed. Retrying...", start));
+              }
+              try {
+                Thread.sleep(200);
+              } catch (InterruptedException e) {
+                LOG.error(e.getMessage());
+                LOG.error(String.format(
```
abti: Please merge the two LOG.error in one, logs are mostly interleaved with each other due to multiple threads logging simultaneously

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,383 @@
...
+    return jobs;
+  }
+
+  private int getPagesSize(final String startDate, final String endDate, final String country,
+      final List<Dimension> requestedDimensions, final List<ApiDimensionFilter> apiDimensionFilters)
+      throws IOException {
+    int REQUESTS_COUNT_EACH_ROUND = 4;
+    //Each round will give you 20K(4*5000) pages. Doing 100 rounds can give 2 million pages.
+    //Normally 1 week has less than 200K pages. So 2 million pages are more than enough.
+    int MAXIMUM_ROUNDS = 100;
```
abti: Can we please make requests_count_each_round and maximum_rounds configurable, so that we can tweak velocity and request counts per round - should anything change (quota / network / google's API limitations) without re-deployment

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,383 @@
...
+                return -1;
+              }
+            }
+          }
+        });
+
+        results.add(submit);
+        try {
+          //Send 4 jobs per second.
+          Thread.sleep(250);
```
abti: Please make rate ie. number of request / sec configurable too, also I think you might be able to reuse RateLimiter here: gobblin.util.limiter.RateLimiter (for its sample usage: gobblin.util.limiter.RateBasedLimiterTest )

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,383 @@
...
+      LOG.info(String.format("Get pages at round %d with size %d.", r, toProcess.size()));
+      ConcurrentLinkedDeque<Pair<String, FilterOperator>> nextRound = new ConcurrentLinkedDeque<>();
+      ExecutorService es = Executors.newFixedThreadPool(10,
+          ExecutorsUtils.newDaemonThreadFactory(Optional.of(LOG), Optional.of(this.getClass().getSimpleName())));
+
+      while (!toProcess.isEmpty()) {
+        submitJob(toProcess.poll(), countryFilter, startDate, endDate, dimensions, es, allPages, nextRound);
+        try {
+          //Not a random number on a whim, this is a good practical number
+          Thread.sleep(275); //Submit roughly 4 jobs per second.
```
abti: same suggestion as above (for RateLimiter)

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
abti: Jin recently added a Retryer factory within Gobblin code base that supports exponential as well as linear retries: gobblin.retry.RetryerFactory 
He has made use of it here: gobblin.source.extractor.extract.google.GoogleAnalyticsUnsampledExtractor method: requestUnsampledReport()

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
```diff
@@ -0,0 +1,383 @@
...
+  private Collection<String> getPages(String startDate, String endDate, List<Dimension> dimensions,
+      ApiDimensionFilter countryFilter, Queue<Pair<String, FilterOperator>> toProcess) throws IOException {
+    String country = GoogleWebmasterFilter.countryFilterToString(countryFilter);
+
+    ConcurrentLinkedDeque<String> allPages = new ConcurrentLinkedDeque<>();
+    Random random = new Random();
+    //we need to retry many times because
+    // 1. the shared prefix path may be very long
+    // 2. the number of retries
+    final int retry = 120;
```
abti: As suggested above please also make this configurable too for easier control

chavdar:
@abti : how serious are the comments you made ? are you OK with merging the current implementation and having follow-up commit later?

abti:
@chavdar My comments are mostly complementary to what you had, but there are no blockers. I am okay with merging the current impl and have follow up commits later.

chavdar:
@enjoyear Some documentation in gobblin-docs/ will be greatly appreciated. I'd like to include this in the highlights of our 0.9 release notes.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSourceWeekly.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class GoogleWebMasterSourceWeekly extends GoogleWebMasterSource {
+  private static DateTimeFormatter _hmsFormatter = DateTimeFormat.forPattern("yyyyMMddHHmmss");
+
+  @Override
+  GoogleWebmasterExtractor createExtractor(WorkUnitState state, Map<String, Integer> columnPositionMap,
+      List<GoogleWebmasterFilter.Dimension> requestedDimensions,
+      List<GoogleWebmasterDataFetcher.Metric> requestedMetrics) throws IOException {
+
+    long lowWatermark = state.getWorkunit().getLowWatermark(LongWatermark.class).getValue();
```
enjoyear: Under what condition will this be null? QueryBasedSource guarantees workunit to have low water mark at line 130. I thought only work units with water marks could be created. If it would have a null exception, what could I do? Wrap it in an IOException? Which IOException should I use?

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -0,0 +1,178 @@
...
+ * The minimum unit of querying range is date. Change the range by configuring "source.querybased.start.value" and "source.querybased.end.value". Note that the analytics data for Google Search Console has a delay or 3 days. So cap your configuration of "source.querybased.append.max.watermark.limit" by "CURRENTDATE-3". See the documentation details of each configuration in the GoogleWebMasterSource fields.
+ *
+ */
+abstract class GoogleWebMasterSource extends QueryBasedSource<String, String[]> {
+
+  /**
+   * Must Provide.
+   * Provide the property site URL whose google search analytics data you want to download
+   */
+  public static final String KEY_PROPERTY = "source.google_webmasters.property_url";
```
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSourceWeekly.java
```diff
@@ -0,0 +1,51 @@
+package gobblin.ingestion.google.webmaster;
+
+import gobblin.configuration.WorkUnitState;
```
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterClientImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static gobblin.source.extractor.extract.google.GoogleCommonKeys.*;
+
+
+public class GoogleWebmasterClientImpl extends GoogleWebmasterClient {
+
+  private final static Logger LOG = LoggerFactory.getLogger(GoogleWebmasterClientImpl.class);
```
enjoyear: Done

enjoyear:
@chavdar Is the documentation about job configuration, talking about the meaning of each key? Or is it about the hierarchy of the code, talking about relationships among different classes and the responsibility of different classes?

chavdar:
Mostly user-facing: usage, configuration, etc. Dev is good too if you
expect other people to extend/contribute.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: Done

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebmasterDataFetcherImpl.java
enjoyear: After I apply the retryer, I feel like the code becomes a little bit harder to read. Please compare the following.
With retryer:
`List<Future<Integer>> results = new ArrayList<>((int) groupSize);

    Map<String, Object> configMap = ImmutableMap.<String, Object>builder()
        .put(RetryerFactory.RETRY_TIME_OUT_MS, TimeUnit.MINUTES.toMillis(TIME_OUT))
        .put(RetryerFactory.RETRY_INTERVAL_MS, Math.round(1000 / REQUESTS_PER_SECOND))
        .put(RetryerFactory.RETRY_TYPE, RetryerFactory.RetryType.FIXED.name()).build();
    Config retryConfig = ConfigFactory.parseMap(configMap);

    final Retryer<Integer> retryer = RetryerFactory.newInstance(retryConfig);

    while (true) {
      for (int i = 0; i < groupSize; ++i) {
        startRow += GoogleWebmasterClient.API_ROW_LIMIT;
        final int start = startRow;
        final String interruptedMsg = String
            .format("Interrupted while trying to get the size of all pages for %s. Current start row is %d.", country,
                start);

        Future<Integer> submit = es.submit(new Callable<Integer>() {
          @Override
          public Integer call() {
            log.info(String.format("Getting page size from %s...", start));

            Callable<Integer> retryCallable = new Callable<Integer>() {
              @Override
              public Integer call()
                  throws Exception {
                try {
                  LIMITER.acquirePermits(1);
                  List<String> pages = _client
                      .getPages(_siteProperty, startDate, endDate, country, GoogleWebmasterClient.API_ROW_LIMIT,
                          requestedDimensions, apiDimensionFilters, start);
                  if (pages.size() < GoogleWebmasterClient.API_ROW_LIMIT) {
                    return pages.size() + start;  //Figured out the size
                  } else {
                    return -1;
                  }
                } catch (InterruptedException e) {
                  log.error("RateBasedLimiter: " + interruptedMsg, e);
                  return -1;
                }
              }
            };

            try {
              return retryer.call(retryCallable);
            } catch (ExecutionException e) {
            } catch (RetryException e) {
              throw new RuntimeException(String
                  .format("Exceeding the timeout of %d minutes while getting the total size of all pages.", TIME_OUT),
                  e);
            }
          }
        });
        results.add(submit);
      }
      //Check the results group in order. The first non-negative count indicates the size of total pages.
      for (Future<Integer> result : results) {
        try {
          Integer integer = result.get();
          if (integer >= 0) {
            es.shutdownNow();
            return integer;
          }
        } catch (InterruptedException | ExecutionException e) {
        }
      }
      results.clear();
    }` 

Previous version without retryer:
`List<Future<Integer>> results = new ArrayList<>((int) groupSize);

    while (true) {
      for (int i = 0; i < groupSize; ++i) {
        startRow += GoogleWebmasterClient.API_ROW_LIMIT;
        final int start = startRow;
        final String interruptedMsg = String
            .format("Interrupted while trying to get the size of all pages for %s. Current start row is %d.", country,
                start);

        Future<Integer> submit = es.submit(new Callable<Integer>() {
          @Override
          public Integer call() {
            log.info(String.format("Getting page size from %s...", start));
            while (true) {
              if (Thread.interrupted()) {
                log.error(interruptedMsg);
                return -1;
              }
              try {
                LIMITER.acquirePermits(1);
                List<String> pages = _client
                    .getPages(_siteProperty, startDate, endDate, country, GoogleWebmasterClient.API_ROW_LIMIT,
                        requestedDimensions, apiDimensionFilters, start);
                if (pages.size() < GoogleWebmasterClient.API_ROW_LIMIT) {
                  return pages.size() + start;  //Figured out the size
                } else {
                  return -1;
                }
              } catch (IOException e) {
                log.info(String.format("Getting page size from %s failed. Retrying...", start));
              } catch (InterruptedException e) {
                log.error("RateBasedLimiter: " + interruptedMsg, e);
                return -1;
              }
            }
          }
        });
        results.add(submit);
      }
      //Check the results group in order. The first non-negative count indicates the size of total pages.
      for (Future<Integer> result : results) {
        try {
          Integer integer = result.get(GET_PAGE_SIZE_TIME_OUT, TimeUnit.MINUTES);
          if (integer >= 0) {
            es.shutdownNow();
            return integer;
          }
        } catch (InterruptedException | ExecutionException e) {
        } catch (TimeoutException e) {
          throw new RuntimeException(String
              .format("Exceeding the timeout of %d minutes while getting the total size of all pages.",
                  GET_PAGE_SIZE_TIME_OUT), e);
        }
      }
      results.clear();
    }`

I actually feel that the version without retryer is better in this case.

For other cases, I cannot use the retryer because the job can evolve based on the results, but the retryer only supports retrying static jobs.

PR #1427, opened 2016-11-29
Update timestamp publisher output path

jack-moseley:
Modify TimestampDataPublisher so that the output path is in the correct format.

Before:
`outputDir/timestamp/dbname_tablename_xxxxx`

After:
`outputDir/dbname.tablename/timestamp`

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
@@ -23,7 +23,11 @@ import gobblin.configuration.WorkUnitState;
 
 
 /**
- * Write output to {finaldir}/{dbname.tablename}/{currenttimestamp}
+ * Path expected from writer:
+ * {writerfinaldir}/{topicname}/{dbname_tablename_xxxxx}
+ *
+ * Publisher output path:
+ * {publisherfinaldir}/{dbname.tablename}/{currenttimestamp}
  */
 public class TimestampDataPublisher extends BaseDataPublisher {
 
--- a/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
@@ -13,15 +13,16 @@
 package gobblin.publisher;
 
 import java.io.IOException;
-
+import java.util.Set;
 import org.apache.hadoop.fs.Path;
-
+import gobblin.util.ParallelRunner;
+import gobblin.util.WriterUtils;
 import gobblin.configuration.State;
 import gobblin.configuration.WorkUnitState;
 
 
 /**
- * Write output to {finaldir}/{currentTimestamp}
+ * Write output to {finaldir}/{dbname.tablename}/{currenttimestamp}
  */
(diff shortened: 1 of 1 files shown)


jack-moseley:
@ydai1124 @chavdar please review

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -32,10 +33,51 @@ public TimestampDataPublisher(State state) throws IOException {
...
+   */
   @Override
-  protected Path getPublisherOutputDir(WorkUnitState workUnitState, int branchId) {
-    Path outputDir = super.getPublisherOutputDir(workUnitState, branchId);
-    outputDir = new Path(outputDir, timestamp);
-    return outputDir;
+  protected void movePath(ParallelRunner parallelRunner, State state, Path src, Path dst, int branchId)
+      throws IOException {
+
+    String outputDir = dst.toString().substring(0, dst.toString().lastIndexOf('/'));
```
ydai1124: You can use `dst.getName()` and `dst.getParent().getName()` to replace the parsing of `/` here.

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -32,10 +33,51 @@ public TimestampDataPublisher(State state) throws IOException {
...
+    super.movePath(parallelRunner, state, src, newDst, branchId);
+  }
+
+  /**
+   * Translate schema name to "dbname.tablename" to use in path
+   *
+   * @param schemaName In format "dbname_tablename_xxxxx"
+   * @return db and table name in format "dbname.tablename"
+   */
+  private String getDbTableName(String schemaName) {
```
ydai1124: Can you add a precondition check here to enforce the expected format?

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
jack-moseley: fixed

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
jack-moseley: Added a regex check

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -13,15 +13,17 @@
...
+import com.google.common.base.Preconditions;
+import gobblin.util.ParallelRunner;
+import gobblin.util.WriterUtils;
 import gobblin.configuration.State;
 import gobblin.configuration.WorkUnitState;
 
 
 /**
- * Write output to {finaldir}/{currentTimestamp}
+ * Write output to {finaldir}/{dbname.tablename}/{currenttimestamp}
```
chavdar: Please have a more detailed description of:
- Preconditions: what is the writer output expected to be
- Post conditions: what does published output look like

jack-moseley: updated javadoc

PR #1431, opened 2016-12-01
YARN app launcher test fixes.

htran1:
Pass dynamic configuration such as ZK and RM ports through a new configuration file.
Package required libraries into a fat jar for distribution to the app master and workers.
Add verification and waits for state changes and handle race conditions causing intermittent failures.

### DIFF
--- a/gobblin-yarn/build.gradle
+++ b/gobblin-yarn/build.gradle
@@ -63,6 +63,33 @@ task testJar(type: Jar, dependsOn: testClasses) {
   from sourceSets.test.output
 }
 
+// create a single Jar with all dependencies
+task fatJar(type: Jar) {
+  zip64 true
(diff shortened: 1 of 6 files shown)


htran1:
@chavdar please review.

gobblin-yarn/build.gradle
```diff
@@ -63,6 +63,33 @@ task testJar(type: Jar, dependsOn: testClasses) {
   from sourceSets.test.output
 }
 
+// create a single Jar with all dependencies
+task fatJar(type: Jar) {
```
chavdar: There is a gradle plugin to create fat jars. You can have a look li-gobblin:li-gobblin-fat/build.gradle and li-gobblin:build.gradle

PR #1436, opened 2016-12-06
Test cleanup fixes to avoid issues when prior run did not clean up.

htran1:


htran1:
@chavdar can you review?

PR #1438, opened 2016-12-07
Run only coverage

chavdar:


chavdar:
Self-approving build fix.

PR #1441, opened 2016-12-08
Fix default plugins

pcadabam:


pcadabam:
Trivial bug fix self merging

PR #1445, opened 2016-12-08
Increase timeout for service manager stop.

htran1:


htran1:
@chavdar can you review?

PR #1452, opened 2016-12-09
Workaround for inconsistent behavior of MysqlExtractor with tinyint(1)

chavdar:
… columns

See ETL-5244

There is a bug in the MysqlExtractor where tinyint columns are always treated as ints.
There are MySQL jdbc driver setting (tinyInt1isBit=true and transformedBitIsBoolean=false) that
 can cause tinyint(1) columns to be treated as BIT/BOOLEAN columns. The default behavior is to
 treat tinyint(1) as BIT.
 
 Currently, {@link MysqlExtractor#getDataTypeMap()} uses the information_schema to check types.
 That does not do the above conversion. {@link #parseColumnAsString(ResultSet, ResultSetMetaData, int)}
 which does the above type mapping.
 
 On the other hand, SqlServerExtractor treats BIT columns as Booleans. So we can be in a bind
 where sometimes BIT has to be converted to an int (for backwards compatibility in MySQL) and
 sometimes to a Boolean (for SqlServer).

The fix is to add gobblin.source.extractor.extract.jdbc.JdbcExtractor.convertBitToBoolean() which returns true for everything but MySQL.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
@@ -970,13 +970,35 @@ public abstract class JdbcExtractor extends QueryBasedExtractor<JsonArray, JsonE
     return baString;
   }
 
+  /**
+   * HACK: there is a bug in the MysqlExtractor where tinyint columns are always treated as ints.
+   * There are MySQL jdbc driver setting (tinyInt1isBit=true and transformedBitIsBoolean=false) that
+   * can cause tinyint(1) columns to be treated as BIT/BOOLEAN columns. The default behavior is to
+   * treat tinyint(1) as BIT.
+   *
+   * Currently, {@link MysqlExtractor#getDataTypeMap()} uses the information_schema to check types.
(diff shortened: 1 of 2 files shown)


chavdar:
@ydai1124 can you review?

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -970,13 +970,35 @@ private String readBlobAsString(Blob logBlob) throws SQLException {
...
+  }
+
   private String parseColumnAsString(final ResultSet resultset, final ResultSetMetaData resultsetMetadata, int i)
       throws SQLException {
 
     if (isBlob(resultsetMetadata.getColumnType(i))) {
       return readBlobAsString(resultset.getBlob(i));
     }
-    if (resultsetMetadata.getColumnType(i) == Types.BIT || resultsetMetadata.getColumnType(i) == Types.BOOLEAN) {
+    if ((resultsetMetadata.getColumnType(i) == Types.BIT
```
ydai1124: Should this be `(resultsetMetadata.getColumnType(i) == Types.BIT && convertBitToBoolean()) || resultsetMetadata.getColumnType(i) == Types.BOOLEAN`

PR #1455, opened 2016-12-10
Add compaction complete listerner for entire work flow

yukuai518:
1. Made method signature change and constructor change for compaction completion listener
2. Made attributes public in TimeBasedSubDirDatasetsFinder

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/listeners/SimpleCompactorCompletionListener.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/listeners/SimpleCompactorCompletionListener.java
@@ -1,24 +1,48 @@
+/*
+ * Copyright (C) 2016-2018 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
 package gobblin.compaction.listeners;
 
+import java.util.Set;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/listeners/SimpleCompactorCompletionListener.java
```diff
@@ -1,24 +1,29 @@
...
+import org.joda.time.DateTime;
 
 import gobblin.annotation.Alias;
-import gobblin.compaction.conditions.RecompactionCombineCondition;
-import gobblin.compaction.mapreduce.MRCompactor;
+import gobblin.configuration.State;
+import gobblin.compaction.dataset.Dataset;
 
 
 public class SimpleCompactorCompletionListener implements CompactorCompletionListener {
```
pcadabam: Add java doc and license

gobblin-compaction/src/main/java/gobblin/compaction/listeners/CompactorCompletionListener.java
```diff
@@ -12,10 +12,11 @@
...
-
-import gobblin.compaction.mapreduce.MRCompactor;
+import java.util.Set;
+import org.joda.time.DateTime;
+import gobblin.compaction.dataset.Dataset;
 
 
 public interface CompactorCompletionListener {
-  void onCompactionCompletion(MRCompactor compactor);
+  void onCompactionCompletion (DateTime intializeTime, Set<Dataset> datasets);
```
pcadabam: Why do we need `initializeTime`? I think having `MRCompactor` in the signature is useful. Can't you get `datasets` from `MRCompactor`?

gobblin-compaction/src/main/java/gobblin/compaction/listeners/SimpleCompactorCompletionListener.java
yukuai518: Done

gobblin-compaction/src/main/java/gobblin/compaction/listeners/CompactorCompletionListener.java
yukuai518: Change back to MRCompactor.

PR #1456, opened 2016-12-12
Fix some issues for compaction launched from writer

jack-moseley:
Changes:

In `CompactionLauncherWriter`
- Add table name as a subdir to compaction tmp path to avoid workunits using conflicting tmp paths
- Add table name and timestamp to compaction destination path

In `HiveMetadataForCompactionExtractor`
- Convert db/table name to "db.table" format in the path to match `TimestampDataPublisher` output

In `HiveSource`
- Add config key to allow a partitioned table to have one workunit per table instead of one workunit per partition

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
@@ -11,12 +11,10 @@ import org.apache.thrift.TException;
 import com.google.common.base.Splitter;
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.WorkUnitState;
-import gobblin.data.management.conversion.hive.avro.AvroSchemaManager;
 import gobblin.data.management.conversion.hive.watermarker.PartitionLevelWatermarker;
 import gobblin.source.extractor.Extractor;
 import gobblin.util.AutoReturnableObject;
 import gobblin.data.management.conversion.hive.extractor.HiveBaseExtractor;
-import gobblin.publisher.TimestampDataPublisher;
 import lombok.extern.slf4j.Slf4j;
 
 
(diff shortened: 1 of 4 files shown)


jack-moseley:
@ydai1124 please review

gobblin-compaction/src/main/java/gobblin/compaction/hivebasedconstructs/HiveMetadataForCompactionExtractor.java
```diff
@@ -49,8 +50,9 @@ public HiveMetadataForCompactionExtractor(WorkUnitState state, FileSystem fs) th
       String deltaString = table.getParameters().get(state.getProp(COMPACTION_DELTA));
       List<String> deltaList = Splitter.on(',').omitEmptyStrings().trimResults().splitToList(deltaString);
 
-      String topicName = AvroSchemaManager.getSchemaFromUrl(this.hiveWorkUnit.getTableSchemaUrl(), fs).getName();
-      Path dataFilesPath = new Path(table.getSd().getLocation(), topicName);
+      String dbTableName = TimestampDataPublisher.getDbTableName(
+          AvroSchemaManager.getSchemaFromUrl(this.hiveWorkUnit.getTableSchemaUrl(), fs).getName());
+      Path dataFilesPath = new Path(table.getSd().getLocation(), dbTableName);
```
jack-moseley: Fixed, registration policy needs to be updated.

gobblin-core/src/main/java/gobblin/publisher/TimestampDataPublisher.java
```diff
@@ -82,7 +82,7 @@ protected void movePath(ParallelRunner parallelRunner, State state, Path src, Pa
    * @param schemaName In format "dbname_tablename_xxxxx"
    * @return db and table name in format "dbname.tablename"
    */
-  private String getDbTableName(String schemaName) {
+  public static String getDbTableName(String schemaName) {
```
jack-moseley: fixed

PR #1457, opened 2016-12-13
Add token based extractor

ydai1124:
Added a simple token based extractor, and a simple converter. @ibuenros Can you take a look?

### DIFF
--- a/gobblin-core/src/main/java/gobblin/converter/string/StringSplitterToListConverter.java
+++ b/gobblin-core/src/main/java/gobblin/converter/string/StringSplitterToListConverter.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package gobblin.converter.string;
+
+import java.util.List;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.Converter;
+import gobblin.converter.DataConversionException;
+import gobblin.converter.SchemaConversionException;
+import gobblin.converter.SingleRecordIterable;
+import gobblin.util.ForkOperatorUtils;
+
+
+/**
+ * Splits a {@link String} record into a record that is {@link List} of {@link String}, based on
+ * {@link ConfigurationKeys#CONVERTER_STRING_SPLITTER_DELIMITER}.
+ */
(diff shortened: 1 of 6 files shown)


gobblin-core/src/main/java/gobblin/converter/string/StringSplitterToListConverter.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public Converter<String, String, String, List<String>> init(WorkUnitState workUnit) {
+    String stringSplitterDelimiterKey =
+        ForkOperatorUtils.getPropertyNameForBranch(workUnit, ConfigurationKeys.CONVERTER_STRING_SPLITTER_DELIMITER);
+    Preconditions.checkArgument(workUnit.contains(stringSplitterDelimiterKey),
+        "Cannot use " + this.getClass().getName() + " with out specifying "
+            + ConfigurationKeys.CONVERTER_STRING_SPLITTER_DELIMITER);
+
+    this.splitter = Splitter.on(workUnit.getProp(stringSplitterDelimiterKey));
```
ibuenros: Should we support trims, omit empty tokens, etc.?

gobblin-core/src/main/java/gobblin/converter/string/StringSplitterToListConverter.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public String convertSchema(String inputSchema, WorkUnitState workUnit)
+      throws SchemaConversionException {
+    return inputSchema;
+  }
+
+  @Override
+  public Iterable<List<String>> convertRecord(String outputSchema, String inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+    return new SingleRecordIterable<>(this.splitter.splitToList(inputRecord));
```
ibuenros: Might be worth also offering an explode mode, each token becomes a record. Not necessary for this PR, but fairly easy to add here.

gobblin-core/src/main/java/gobblin/converter/string/StringSplitterToListConverter.java
ydai1124: Will do.

ydai1124:
@ibuenros Comments addressed.

gobblin-core/src/test/java/gobblin/converter/string/StringSplitterToListConverterTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.WorkUnitState;
+import gobblin.converter.DataConversionException;
+
+import avro.shaded.com.google.common.collect.Lists;
```
ibuenros: wrong import, please use `com.google...`

gobblin-core/src/test/java/gobblin/source/extractor/filebased/TokenizedFileDownloaderTest.java
```diff
@@ -0,0 +1,51 @@
...
+package gobblin.source.extractor.filebased;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+
+import org.testng.annotations.Test;
+
+import avro.shaded.com.google.common.base.Joiner;
+import junit.framework.Assert;
```
ibuenros: The above two imports are incorrect, should use `testng` assert, and non-shaded guava

PR #1458, opened 2016-12-13
Default gobblin instance job launcher correctly uses templates.

ibuenros:


ibuenros:
@pcadabam can you review?

PR #1463, opened 2016-12-15
MySQL DB state store.

htran1:


htran1:
@abti please take a look.

htran1:
@chavdar please take a look.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -47,6 +47,13 @@
   public static final String STATE_STORE_FS_URI_KEY = "state.store.fs.uri";
   // Enable / disable state store
   public static final String STATE_STORE_ENABLED = "state.store.enabled";
+  // DB state store configuration
+  public static final String STATE_STORE_DB_ENABLED_KEY = "state.store.db.enabled";
```
chavdar: We should have a more generic way of instantiating the state store rather than just having two options: file system and mysql. We certainly want to be able to support other SQL stores (e.g. Oracle or posgres) or other types of stores (like Espresso).

I suggest we have a Factory based approach. Something along the lines of:

public interface StateStoreFactory {
   StateStore createStateStore(Config storeConfig);
}

Then we ca have a config property, say, state.store.type that takes a factory class or a human readable mnemonic. You can have a look at gobblin.http.HttpClientConfiguratorLoader.getConfiguratorClass(Optional<String>) for a pattern on how to support both syntaxes. createStateStore() can then just instantiate the factory class and create the store.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -159,16 +161,29 @@ public JobContext(Properties jobProps, Logger logger) throws Exception {
...
+    boolean stateStoreEnabled = !jobProps.containsKey(ConfigurationKeys.STATE_STORE_ENABLED)
+        || Boolean.parseBoolean(jobProps.getProperty(ConfigurationKeys.STATE_STORE_ENABLED));
+
+    if (stateStoreEnabled &&
+        Boolean.parseBoolean(jobProps.getProperty(ConfigurationKeys.STATE_STORE_DB_ENABLED_KEY, "false"))) {
+      MysqlDataSource mySqlDs = new MysqlDataSource();
+
+      mySqlDs.setURL(jobProps.getProperty(ConfigurationKeys.STATE_STORE_DB_URL_KEY));
+      mySqlDs.setUser(jobProps.getProperty(ConfigurationKeys.STATE_STORE_DB_USER_KEY));
+      mySqlDs.setPassword(jobProps.getProperty(ConfigurationKeys.STATE_STORE_DB_PASSWORD_KEY));
```
lbendig: We may support encrypted passwords here:
`PasswordManager.getInstance(properties).readPassword(properties.getProperty(ConfigurationKeys.STATE_STORE_DB_PASSWORD_KEY));`

gobblin-metastore/src/main/java/gobblin/metastore/DbStateStore.java
```diff
@@ -0,0 +1,343 @@
...
+ *     The database table row is keyed by the store name and table name.
+ *     State keys are state IDs (see {@link State#getId()}), and values are objects of {@link State} or
+ *     any of its extensions. Keys will be empty strings if state IDs are not set
+ *     (i.e., {@link State#getId()} returns <em>null</em>). In this case, the
+ *     {@link DbStateStore#get(String, String, String)} method may not work.
+ * </p>
+ *
+ * @param <T> state object type
+ **/
+public class DbStateStore<T extends State> implements StateStore<T> {
```
chavdar: nit: Can you rename that to MySQLStateStore

gobblin-runtime/src/main/java/gobblin/runtime/DbDatasetStateStore.java
```diff
@@ -0,0 +1,146 @@
...
+ *   backward compatibility.
+ * </p>
+ *
+ */
+public class DbDatasetStateStore extends DbStateStore<JobState.DatasetState>
+    implements DatasetStateStore<JobState.DatasetState> {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DbDatasetStateStore.class);
+
+  @Alias("MySqlDbDatasetStateStore")
```
chavdar: nit: I'd change it to just "mysql" . The alias name scope is per interface so there should not be conflicts.

gobblin-runtime/src/main/java/gobblin/runtime/FsDatasetStateStore.java
```diff
@@ -52,13 +59,42 @@
...
+      return (DatasetStateStore<JobState.DatasetState>) Class.forName(className)
+          .getConstructor(FileSystem.class, String.class)
+          .newInstance(stateStoreFs, stateStoreRootDir);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
 
-  public static final String CURRENT_DATASET_STATE_FILE_SUFFIX = "current";
+  @Alias("FsDatasetStateStore")
```
chavdar: I suggest using "fs"

gobblin-runtime/src/main/java/gobblin/runtime/NoopDatasetStateStore.java
```diff
@@ -28,6 +34,14 @@
  */
 public class NoopDatasetStateStore extends FsDatasetStateStore {
 
+  @Alias("NoopDatasetStateStore")
```
chavdar: I suggest using "noop"

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -53,13 +53,19 @@
...
+      "state.store.db.conn.min.evictable.idle.time";
+  public static final long DEFAULT_STATE_STORE_DB_CONN_MIN_EVICTABLE_IDLE_TIME = 5 * 60 * 1000;
   public static final String STATE_STORE_DB_URL_KEY = "state.store.db.url";
   public static final String STATE_STORE_DB_USER_KEY = "state.store.db.user";
   public static final String STATE_STORE_DB_PASSWORD_KEY = "state.store.db.password";
   public static final String STATE_STORE_DB_TABLE_KEY = "state.store.db.table";
   public static final String DEFAULT_STATE_STORE_DB_TABLE = "gobblin_job_state";
-  public static final String STATE_STORE_DB_COMPRESS_KEY = "state.store.db.compress";
-  public static final boolean DEFAULT_STATE_STORE_DB_COMPRESS = true;
+  public static final String STATE_STORE_DB_COMPRESS_VALUE_KEY = "state.store.db.compress.value";
```
chavdar: As I like to say, dots.are.not.substitutes.for.spaces. We should use dots only for scoping. Unfortunately, we have a bunch of legacy settings that use that pattern but we should not keep using it.

I suggest using state.store.db.compressedValues or state.store.db.enableCompressedValues

PR #1470, opened 2016-12-18
Document OracleExtractor connection property: source.conn.sid

lbendig:


lbendig:
@ibuenros Can you please review?

PR #1472, opened 2016-12-19
Move one flavor of AbstractJobLauncher.runWorkUnits

chavdar:
Move one flavor of AbstractJobLauncher.runWorkUnits to GobblinMultiTaskAttempt since the latter is the only invoker. Convert the method to an instance method.

@ydai1124 Can you review ?


chavdar:
@ydai1124 I also move runAndOptionallyCommitTaskAttempt . Could you have a look again?

PR #1473, opened 2016-12-19
Add support for alternate input paths in gobblin-compaction

pcadabam:
Currently gobblin-compaction supports inputs from one base directory. `a/b/hourly/2016/12/10/{00..23}`. Sometimes few hours may need to be used from a different location say `c/d/hourly/2016/12/10/{00..23}`. This PR add support to have multiple input directories for a `Dataset`

@ydai1124 , @yukuai518 can you please review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
@@ -15,7 +15,6 @@ package gobblin.compaction.dataset;
 import java.io.IOException;
 import java.net.URI;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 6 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/dataset/TimeBasedSubDirDatasetsFinder.java
```diff
@@ -102,55 +107,84 @@ private String getDatasetName(String path, String basePath) {
...
+
+      @Override
+      public Dataset apply(Builder b) {
+        return b.build();
+      }
+    }));
+
+  }
+
+  private void buildDatasetForInput(Map<DateTime, Dataset.Builder> datasetsBuilderMap, String datasetName,
```
ydai1124: Can you add unit test for this?

gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
```diff
@@ -51,6 +46,7 @@
   protected final Configuration conf;
   protected final FileSystem fs;
   protected final String inputDir;
+  protected final Optional<String> alternateInputDir;
```
ydai1124: Is this input `alternate` or more like `additional`?

gobblin-compaction/src/main/java/gobblin/compaction/dataset/TimeBasedSubDirDatasetsFinder.java
pcadabam: Removed this code. Will be extending this finder and creating a li-finder. li-finder will have tests

gobblin-compaction/src/main/java/gobblin/compaction/dataset/DatasetsFinder.java
pcadabam: Removed. Again this will be handled in li-finder. `additional` is already used in gobblin-compaction. I did not want to overload the term.

PR #1476, opened 2016-12-20
Include ThrottleWriter state in FinalState

agepati:
Missed including ThrottleTime in the ThrottleWriter FinalState. Correct, and include tests.

Also, correct casing to be aligned with the existing conventions.


agepati:
@pcadabam Could you please review?

PR #1477, opened 2016-12-20
Query based source will reset to low watermark if previous run did no…

ibuenros:
…t process any data for that table.


ibuenros:
@chavdar can you review? Thanks

PR #1478, opened 2016-12-21
Remove empty AvroKafkaDataWriter

chavdar:


chavdar:

PR #1482, opened 2016-12-21
Move google sources out of gobblin-core

chavdar:
Move google sources from gobblin-core to gobblin-modules/google-ingestion


chavdar:
@abti can you review ?

abti:
There seems to be a conflict with gradle file

PR #1485, opened 2016-12-22
Removing .keep files

aditya1105:
Modified GobblinRuntimeTest and removed .keep files which were added in the first place to add empty directories in the git repo


agepati:
@pcadabam Could you please review?

PR #1487, opened 2016-12-22
Added interfaces for Gobblin broker.

ibuenros:
This PR just adds the interfaces for Gobblin broker.


ibuenros:
@chavdar @ydai1124 can you review? Thanks!

gobblin-utility/src/main/java/gobblin/broker/iface/ScopedSharedResourceFactory.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * A {@link SharedResourceFactory} that creates {@link Scope} specific objects.
+ * @param <T> type of objects this {@link SharedResourceFactory} creates.
+ * @param <K> type of {@link SharedResourceKey} this factory uses.
+ */
+public interface ScopedSharedResourceFactory<T, K extends SharedResourceKey> extends SharedResourceFactory<T, K> {
+  /**
+   * Create an object for the provided {@link SharedResourceKey}, with the provided configuration.
+   */
+  T createResource(ScopedGobblinBroker broker, KeyedScopedConfigView<?, K> config);
```
chavdar: Why does the factory need an instance of the broker? Can't it just get a scope?

gobblin-utility/src/main/java/gobblin/broker/iface/SharedResourceFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+package gobblin.broker.iface;
+
+/**
+ * A superclass of {@link ScopedSharedResourceFactory} and {@link UnscopedSharedResourceFactory}.
+ * {@link SharedResourceFactory}s create objects that a {@link GobblinBroker} can manage. The factory is required to be stateless.
+ * @param <T> The type of objects built by this factory.
+ * @param <K> The key used to create objects by this factory.
+ */
+public interface SharedResourceFactory<T, K extends SharedResourceKey> {
```
chavdar: This interface is strange. It does not define any "factory" methods and does not use any of the generic parameters.

gobblin-utility/src/main/java/gobblin/broker/iface/UnscopedSharedResourceFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package gobblin.broker.iface;
+
+/**
+ * A {@link SharedResourceFactory} that creates global objects (not specific to any {@link Scope}).
+ * @param <T> type of objects this {@link SharedResourceFactory} creates.
+ * @param <K> type of {@link SharedResourceKey} this factory uses.
+ */
+public interface UnscopedSharedResourceFactory<T, K extends SharedResourceKey> extends SharedResourceFactory<T, K> {
```
chavdar: Can you think if we can avoid the parallel class hierarchies for scoped and unscoped?

gobblin-utility/src/main/java/gobblin/broker/iface/KeyedConfigView.java
```diff
@@ -0,0 +1,23 @@
...
+
+package gobblin.broker.iface;
+
+import com.typesafe.config.Config;
+
+
+/**
+ * Provides a view of a {@link Config} specific to a {@link SharedResourceKey} and {@link SharedResourceFactory}.
+ */
+public interface KeyedConfigView<K extends SharedResourceKey> extends ConfigView {
```
ydai1124: What is the interface for? What are other use cases besides `KeyedScopedConfigView`?

gobblin-utility/src/main/java/gobblin/broker/iface/GobblinBroker.java
```diff
@@ -0,0 +1,90 @@
...
+   * @param scopeType the {@link ScopeType} requested.
+   * @return a {@link GobblinBroker} at the specified scope
+   * @throws NoSuchScopeException
+   */
+  ScopedGobblinBroker<S> scopedBroker(S scopeType) throws NoSuchScopeException;
+
+  /**
+   * @return The lowest {@link Scope} for which this {@link GobblinBroker} can provide a {@link ScopedGobblinBroker}.
+   */
+  Scope<S> leafScope();
```
ydai1124: Is `leafScope()` used for the callers to check if the requested `scopeType` is provided? Then why not simply provide a collection of all the `scopeType`s that are supported?

gobblin-utility/src/main/java/gobblin/broker/iface/ScopedSharedResourceFactory.java
ibuenros: Because the factory may itself need a shared resource. For example, a limiter factory needs to acquire a restli client which is shared by all limiters.

ibuenros:
@chavdar @ydai1124 addressed all comments.
I simplified the interfaces by removing scoped parallels of many things (although scoped and unscoped factories are still separate).

gobblin-utility/src/main/java/gobblin/broker/iface/ScopeType.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collection;
+
+import javax.annotation.Nullable;
+
+
+/**
+ * Represents a DAG of scope types.
+ *
+ * For example, the topology of a distributed application might have scopes as follows:
+ * GLOBAL -> INSTANCE --> JOB       --> TASK
```
chavdar: nit: It is useful to enclose that in a &lt;pre&gt; tag otherwise, HTML will mess up the alignment.

PR #1489, opened 2016-12-23
ZK/HelixPropertyStore state store

htran1:
@chavdar please take a look.


gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -64,8 +66,8 @@
   public static final String STATE_STORE_DB_PASSWORD_KEY = "state.store.db.password";
   public static final String STATE_STORE_DB_TABLE_KEY = "state.store.db.table";
   public static final String DEFAULT_STATE_STORE_DB_TABLE = "gobblin_job_state";
-  public static final String STATE_STORE_DB_COMPRESSED_VALUES_KEY = "state.store.db.compressedValues";
-  public static final boolean DEFAULT_STATE_STORE_DB_COMPRESSED_VALUES = true;
+  // ZooKeeper state store configuration
```
chavdar: Please move the property definition to ZkStateStore as it is specific to it.

gobblin-modules/gobblin-helix/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+  compile externalDependency.commonsLang
+  compile externalDependency.commonsLang3
+  compile externalDependency.lombok
+  compile externalDependency.reflections
+  compile externalDependency.zkClient
+
+  testCompile externalDependency.testng
+  testCompile externalDependency.curatorTest
+
+  compile files('../../gobblin-cluster/src/main/resources/helix-core-0.6.6-SNAPSHOT.jar')
```
chavdar: Please use a dependency on the 0.6.6 Maven artifact.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -0,0 +1,279 @@
...
+   * @param stateClass The type of state being stored
+   * @throws IOException
+   */
+  public ZkStateStore(String connectString, String storeRootDir, boolean compressedValues, Class<T> stateClass)
+      throws IOException {
+    this.compressedValues = compressedValues;
+    this.stateClass = stateClass;
+
+    ZkSerializer serializer = new ByteArraySerializer();
+    propStore = new ZkHelixPropertyStore<byte[]>(connectString, serializer, storeRootDir);
```
chavdar: Do we allow customization of the configuration of Helix property store or ZK? For example, ZK has connect and session timeouts which may need to be configured.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  @Override
+  public boolean exists(String storeName, String tableName) throws IOException {
+    String path = formPath(storeName, tableName);
+
+    return propStore.exists(path, 0);
+  }
+
+  private void addStateToOs(DataOutput dataOutput, T state) throws IOException {
```
chavdar: What does "Os" stand for? Could you re-word?

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -0,0 +1,279 @@
...
+
+    for (T state : states) {
+      addStateToOs(dataOutput, state);
+    }
+
+    if (!exists(storeName, tableName) && !create(storeName, tableName)) {
+      throw new IOException("Failed to create a state file for table " + tableName);
+    }
+
+    dataOutput.close();
```
chavdar: this should be in a try-resouce block to guarantee the calling of the close() method.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -0,0 +1,279 @@
...
+   * Deserialize data into a list of {@link State}s.
+   * @param data byte array
+   * @param states output list of states
+   * @param stateId optional key filter. Set to null for no filtering.
+   * @throws IOException
+   */
+  private void deserialize(byte[] data, List<T> states, String stateId) throws IOException {
+    if (data != null) {
+      ByteArrayInputStream bais = new ByteArrayInputStream(data);
+      InputStream is = compressedValues ? new GZIPInputStream(bais) : bais;
```
chavdar: I'd consider storing of the compressedValues flag in ZK. Thus we can avoid error scenarios where previosly compressed values are being read with compressedValues == false and vice versa.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -0,0 +1,279 @@
...
+  /**
+   * Deserialize data into a list of {@link State}s.
+   * @param data byte array
+   * @param states output list of states
+   * @param stateId optional key filter. Set to null for no filtering.
+   * @throws IOException
+   */
+  private void deserialize(byte[] data, List<T> states, String stateId) throws IOException {
+    if (data != null) {
+      ByteArrayInputStream bais = new ByteArrayInputStream(data);
```
chavdar: needs to be in a try-resource block and closed at the end.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
chavdar: Is there any expectation that storeRootDir exists or will ZkHelixPropertyStore create it automatically?

htran1: There is no additional configuration of the timeouts allowed with the Helix property store.  It creates the ZkClient with session timeout of 30 seconds and connect timeout of 60 seconds.
    this._zkclient = new ZkClient(zkAddress, 30000, '\uea60', serializer);
    this._zkclient.waitUntilConnected(60000L, TimeUnit.MILLISECONDS);

htran1: ZkHelixPropertyStore will create storeRootDir automatically when the first entry is created.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
htran1: How about we use the following GZipCompressionUtil.isCompressed to check the data before decompressing?  This is what HelixPropertyStore uses for ZNRecords.
  public static boolean isCompressed(byte[] bytes) {
    return bytes != null && bytes.length >= 2?bytes[0] == 31 && bytes[1] == -117:false;
  }

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
htran1: Os is for OutputStream.  I'll reword this.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
htran1: The close here is to close and flush the compression stream so that the byte array in the put() is complete.  In case of exception we don't need the close.

htran1:
Created a new PR off of a separate branch since streaming branch is being shared.

PR #1492, opened 2016-12-29
Better async management, couchbase writer improvements

shirshanka:
@abti: Please take a look when you get a chance.


gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+    }
+
+    @Override
+    public void run() {
+      while (true)
+      {
+        try {
+          Attempt attempt = _retryQueue.take();
+          if (attempt != null) {
+            attemptWrite(attempt);
```
abti: Do you intend to add some sort of retry back-off to it?

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+  private void attemptWrite(final Attempt attempt)
+  {
+    final long startTime = System.nanoTime();
+    this.recordsAttempted.mark();
+    this.asyncDataWriter.write(attempt.record, new WriteCallback() {
+
+      @Override
+      public void onSuccess(WriteResponse writeResponse) {
+
+        recordsSuccess.mark();
```
abti: If this is a retry attempt, should the recordsFailed meter be decremented? If not, recordsFailed will continue to grow and will soon yield infinite while loop in line 252

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+      throw new IOException("Irrecoverable failure on async write", cachedWriteException);
+    }
+  }
+
+  private void attemptWrite(final Attempt attempt)
+  {
+    final long startTime = System.nanoTime();
+    this.recordsAttempted.mark();
+    this.asyncDataWriter.write(attempt.record, new WriteCallback() {
+
```
abti: mayBeThrow() should be invoked here as well (or only here should work as well), because after all the regular writes are done (ie. no more calls to write() ), and only retry queue only is being attempted (by invoking attemptWrite() directly from retry thread) - then we will miss the check to see if we have crossed the failure tolerance threshold

gobblin-core-base/src/main/java/gobblin/writer/WriteResponse.java
```diff
@@ -0,0 +1,57 @@
...
+   * @return
+   */
+  String getStringResponse();
+
+  /**
+   * The number of bytes written as part of this write.
+   * @return The number of bytes written as part of this write.
+   * -1 if this value is unknown. 0 if nothing was written.
+   */
+  long bytesWritten();
```
abti: Like discussed, since we will move away from hard dependency on bytesWritten() .. if you are not changing it, please create an issue so that we dont forget

gobblin-core-base/src/test/java/gobblin/writer/FutureWrappedWriteCallback.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+      else
+      {
+        return _writeResponse;
+      }
+    }
+    else
+    {
+      //TODO: umm...
+      return null;
```
abti: If it ends up in else, Java is broken, throw exception :)

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/converter/AnyToCouchbaseJsonConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Takes any object in and converts it into a CouchBase compatible
+ * {@link com.couchbase.client.java.document.RawJsonDocument}
+ * It expects to be configured to pick out the String key field from the incoming Object
+ */
+public class AnyToCouchbaseJsonConverter extends Converter<String, String, Object, RawJsonDocument> {
+
+  private static final Gson GSON = new Gson();
+  private String keyField = "key";
```
abti: Make this configurable?

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/converter/AnyToCouchbaseJsonConverter.java
```diff
@@ -0,0 +1,79 @@
...
+      throws SchemaConversionException {
+    //TODO: Use the schema and config to determine which fields to pull out
+    return "";
+  }
+
+  @Override
+  public Iterable<RawJsonDocument> convertRecord(String outputSchema, Object inputRecord, WorkUnitState workUnit)
+      throws DataConversionException {
+
+    JsonElement jsonElement = GSON.toJsonTree(inputRecord);
```
abti: Seems like we are converting from Json to Couchbase, why are we calling it AnyToCouchbase converter?

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/writer/CouchbaseWriter.java
```diff
@@ -103,38 +128,132 @@ public CouchbaseWriter(CouchbaseEnvironment couchbaseEnvironment, Config config)
...
   {
     return _bucket;
   }
 
+
   @Override
-  public void cleanup()
-      throws IOException {
+  public Future<WriteResponse> write(final D record, final WriteCallback callback) {
+    if (record instanceof TupleDocument) {
```
abti: Can record be any other type that TupleDocument? If not, should this throw an exception in else? If it throws exception, will Manager keep it retrying?

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/writer/CouchbaseWriter.java
```diff
@@ -103,38 +128,132 @@ public CouchbaseWriter(CouchbaseEnvironment couchbaseEnvironment, Config config)
...
+          } else {
+            return getWriteResponseorThrow(writeResponseThrowablePair);
+          }
+        }
+      };
+
+      observable.timeout(_operationTimeout, _operationTimeunit).subscribe(new Subscriber<D>() {
+        @Override
+        public void onCompleted() {
+        }
```
abti: why not handle onCompleted?

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/writer/CouchbaseWriter.java
```diff
@@ -103,38 +128,132 @@ public CouchbaseWriter(CouchbaseEnvironment couchbaseEnvironment, Config config)
...
+          callback.onFailure(e);
+        }
+
+        @Override
+        public void onNext(D doc) {
+          callbackFired.set(true);
+          WriteResponse writeResponse = new GenericWriteResponse<D>(doc);
+          writeResponseQueue.add(new Pair<WriteResponse, Throwable>(writeResponse, null));
+          callback.onSuccess(writeResponse);
+          if (doc instanceof TupleDocument) {
```
abti: What if doc is not of TupleDocument type? Also, wondering if content().value1().release() method chain can yield potential NPE?

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: recordsFailed is only incremented when retry attempts have failed. Line 333 below.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: yeah, this is a very simple POC. Eventually this area will turn into a delayed scheduler with a few retry strategies, linear and exponential at the very least.

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/converter/AnyToCouchbaseJsonConverter.java
shirshanka: Converts any Gson supported complex type to Couchbase RawJsonDocument

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/writer/CouchbaseWriter.java
shirshanka: JsonDocument works as well.  Point taken about understanding the behavior of this when non-compliant types are sent down.

gobblin-modules/gobblin-couchbase/src/main/java/gobblin/couchbase/writer/CouchbaseWriter.java
shirshanka: onCompleted is redundant if you handle onNext and onError.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+      } catch (InterruptedException e) {
+        Throwables.propagate(e);
+      }
+    }
+    recordsIn.mark();
+    if (this.retriesEnabled) {
+      attemptWrite(new Attempt(record, 1, null));
+    } else {
+      recordsAttempted.mark();
+      this.asyncDataWriter.write(record, writeCallback);
```
chavdar: I think it will be cleaner if we use attemptWrite both when retries are enabled and disabled. The difference is that  in the latter case the max number of retries is 0. This will guarantee that handling of success and failures is consistent in terms of metrics and other (e.g. the isFailureFatal() aspect).

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+  public long bytesWritten()
+      throws IOException {
+    return bytesWritten.getCount();
+  }
+
+  @Override
+  public void close()
+      throws IOException {
+    log.debug("Close called");
+    this.closer.close();
```
chavdar: As a precaution, close needs to make sure that there are no outstanding attempts and potentially call flush()/commit().

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  @Override
+  public void commit()
+      throws IOException {
+    log.debug("Commit called, will wait for commitTimeout : " + commitTimeoutInNanos / MILLIS_TO_NANOS + "ms");
```
chavdar: log.debug("Commit called, will wait for commitTimeout : {} ms", commitTimeoutInNanos / MILLIS_TO_NANOS) will not have overhead if debug writing is disabled. Further, I think this logging statement may be good for info level.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+    {
+      log.warn("Timeout waiting for all writes to be acknowledged. Missing " + unacknowledgedWrites
+          + " responses out of " + recordsProducedFinal);
+    }
+    if (totalFailures > 0 && recordsProducedFinal > 0)
+    {
+      String message = "Commit failed to write " + totalFailures
+          + " records (" + recordsFailedFinal + " failed, " + unacknowledgedWrites + " unacknowledged) out of "
+          + recordsProducedFinal + " produced.";
+      double failureRatio = (double)totalFailures / (double)recordsProducedFinal;
```
chavdar: This should use the same logic as isFailureFatal()

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+        throw new IOException("Interrupted while waiting for commit to complete", e);
+      }
+    }
+    log.debug("Commit done waiting");
+    long recordsProducedFinal = recordsIn.getCount();
+    long recordsWrittenFinal = recordsSuccess.getCount();
+    long recordsFailedFinal = recordsFailed.getCount();
+    long unacknowledgedWrites  = recordsProducedFinal - recordsWrittenFinal - recordsFailedFinal;
+    long totalFailures = unacknowledgedWrites + recordsFailedFinal;
+    if (unacknowledgedWrites > 0) // timeout
```
chavdar: I think we have to be conservative and treat the unacknowledged writes as errors.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  @Override
+  public void commit()
+      throws IOException {
+    log.debug("Commit called, will wait for commitTimeout : " + commitTimeoutInNanos / MILLIS_TO_NANOS + "ms");
+    long commitStartTime = System.nanoTime();
+    asyncDataWriter.flush();
+    while (((System.nanoTime() - commitStartTime) < commitTimeoutInNanos)  &&
+        (recordsIn.getCount() != (recordsSuccess.getCount() + recordsFailed.getCount())))
```
chavdar: nit: please be consistent in prefixing the member variables with "this." if not using the "_" prefix.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+
+  public static AsyncWriterManagerBuilder builder() {
+    return new AsyncWriterManagerBuilder();
+  }
+
+  public static class AsyncWriterManagerBuilder {
+    private Config config = ConfigFactory.empty();
+    private long commitTimeoutInNanos = COMMIT_TIMEOUT_IN_NANOS_DEFAULT;
+    private long commitStepWaitTimeMillis = COMMIT_STEP_WAITTIME_MILLIS_DEFAULT;
+    private double failureAllowance = FAILURE_ALLOWANCE_DEFAULT;
```
chavdar: Please be more descriptive in the naming. If this is a ratio, then use "failureAllowanceRatio"; if this is a percentage, then use "failureAllowancePct" or some such.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+    private boolean retriesEnabled = RETRIES_ENABLED_DEFAULT;
```
chavdar: I'd rather have one setting (numRetries); numRetries == 0, means that retries are disabled.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+     * b) Use a bounded queue to store call attempts, write the returned futures into it, remove the future on callback success.
+     * Problem: More state to keep in memory.
+     * How do you deal with retries.. how do you ensure there is no starvation on retries?
+     *
+     * Doing a) for now.
+     */
+
+    maybeThrow();
+    int spinNum = 0;
+    while (recordsIn.getCount() - (recordsSuccess.getCount() + recordsFailed.getCount()) > maxOutstandingWrites)
```
chavdar: i'd refactor the check in a method. It is conceivable that we may want to add a size-based limit in the future (that may be more useful than a count-based limit).

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+
+  public static AsyncWriterManagerBuilder builder() {
+    return new AsyncWriterManagerBuilder();
+  }
+
+  public static class AsyncWriterManagerBuilder {
+    private Config config = ConfigFactory.empty();
+    private long commitTimeoutInNanos = COMMIT_TIMEOUT_IN_NANOS_DEFAULT;
```
chavdar: I'd consider avoiding having two more somewhat arbitrary nobs to turn. We can keep track of the average write latency and use it as a proxy for commitStepWaitTimeMillis. We can use commitStepWaitTimeMillis * remaining retries as a proxy for commitTimeoutInNanos . Further, having one setting in milliseconds and the other in nanoseconds is somewhat confusing. I think nanoseconds is an overkill.

gobblin-core-base/src/main/java/gobblin/writer/WriteResponse.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package gobblin.writer;
+
+
+public interface WriteResponse<T> {
+
+  /**
+   * Get the raw response returned by the underlying write system
+   * @return
```
chavdar: nit: please remove empty javadoc tags

gobblin-core-base/src/main/java/gobblin/writer/AsyncDataWriter.java
```diff
@@ -10,25 +10,33 @@
  * CONDITIONS OF ANY KIND, either express or implied.
  */
 
-package gobblin.kafka.writer;
+package gobblin.writer;
```
chavdar: I'd consider putting this in a separate package, say, gobblin.writer.async and adding a package-info.java class with the high-level picture description.

chavdar:
Having a high-level overview of the PR is extremely useful.

gobblin-core-base/src/main/java/gobblin/writer/WriteResponse.java
```diff
@@ -0,0 +1,57 @@
...
+ *  * Unless required by applicable law or agreed to in writing, software distributed
+ *  * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ *  * CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+
+package gobblin.writer;
+
+
+public interface WriteResponse<T> {
```
chavdar: I am not complete sure about the reason for this interface. Can you add some javadocs?

gobblin-core-base/src/main/java/gobblin/writer/WriteResponseMapper.java
```diff
@@ -0,0 +1,19 @@
...
+ *  *
+ *  * Unless required by applicable law or agreed to in writing, software distributed
+ *  * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ *  * CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+
+package gobblin.writer;
+
+public interface WriteResponseMapper<InnerType> {
```
chavdar: Why is this necessary?

gobblin-core-base/src/test/java/gobblin/writer/FutureWrappedWriteCallback.java
chavdar: An assertion error.

gobblin-core-base/src/test/java/gobblin/writer/FutureWrappedWriteCallback.java
```diff
@@ -0,0 +1,124 @@
...
+      return get(-1, TimeUnit.MILLISECONDS);
+    } catch (TimeoutException e) {
+      throw new ExecutionException(e);
+    }
+  }
+
+  @Override
+  public WriteResponse get(long timeout, TimeUnit unit)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    synchronized (_lock)
```
chavdar: I'd consider using "this" or  a regular lock object and a Condition variable.

gobblin-core-base/src/test/java/gobblin/writer/FutureWrappedWriteCallback.java
```diff
@@ -0,0 +1,124 @@
...
+  @Override
+  public void onSuccess(WriteResponse writeResponse) {
+    synchronized (_lock) {
+      _writeResponse = writeResponse;
+      _callbackFired = true;
+      if (_innerCallback != null)
+      {
+        _innerCallback.onSuccess(writeResponse);
+      }
+      _lock.notify();
```
chavdar: I think you have to use notifyAll()

gobblin-core-base/src/test/java/gobblin/writer/FutureWrappedWriteCallback.java
```diff
@@ -0,0 +1,124 @@
...
+
+
+  @Override
+  public void onSuccess(WriteResponse writeResponse) {
+    synchronized (_lock) {
+      _writeResponse = writeResponse;
+      _callbackFired = true;
+      if (_innerCallback != null)
+      {
+        _innerCallback.onSuccess(writeResponse);
```
chavdar: What is the contract if onSuccess throws an exception? It's probably best to catch, log and ignore.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: I don't see this being done in other writers: e.g. 
gobblin-core/src/main/java/gobblin/writer/JdbcWriter.java
gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java

Are you concerned about this particular class because it is interfacing with async writers?

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: That's what I'm doing here.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: Ok to do this later when isFailureFatal actually evolves to being able to do more sophisticated analysis? If I just move this logic to isFailureFatal now, it will run the risk of early termination if there are a large number of failures early (e.g. if the 2nd write fails, we're already at 50% failures). Also unacknowledged writes are treated as failures after timeout on commit.. cannot do the same when writes are still in flight.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: I was trying to save the object creation cost (for Attempt objects and callback) if retries are not enabled, agree that it makes the code harder to make consistent. Will remove this path.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+    this.writeCallback = new WriteCallback() {
+      @Override
+      public void onSuccess(WriteResponse writeResponse) {
+        recordsSuccess.mark();
+        log.debug("Received success", writeResponse.getStringResponse());
+      }
+
+      @Override
+      public void onFailure(Throwable throwable) {
+        recordsFailed.mark();
```
shirshanka: Will remove the default write callback for now.

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
```diff
@@ -0,0 +1,519 @@
...
+ * 1. Calculate metrics for: number of records in, attempted, successfully written, failed, latency.
+ * 2. Wait for a specified amount of time on commit for all pending writes to complete.
+ * 3. Do not proceed if a certain failure threshold is exceeded.
+ * 4. Support a fixed number of retries on failure of individual records (TODO: retry strategies)
+ * 5. Support a max number of outstanding / unacknowledged writes
+ * 6. TODO: Support ordered / unordered write semantics
+ *
+ *
+ */
+@Slf4j
```
shirshanka: Should I then remove the @slf4j annotation? 
Is the recommendation to hand create a logger if one isn't passed down?

gobblin-core-base/src/main/java/gobblin/writer/AsyncWriterManager.java
shirshanka: If there are no more calls to write() then we don't have any way to indicate to the calling thread (the Fork) that we need to throw an exception. 

The next place where this can happen is in the commit() where SLA-s are getting checked. 
What I can do is wrap the cached write exception at that point and throw it with the IOException that gets thrown.

gobblin-core-base/src/main/java/gobblin/writer/AsyncDataWriter.java
shirshanka: Where would you recommend putting the AsyncWriterManager (a class that bridges the sync and async worlds).

gobblin-core-base/src/main/java/gobblin/writer/WriteResponse.java
shirshanka: Will create an issue.

gobblin-core-base/src/main/java/gobblin/writer/GenericWriteResponse.java
```diff
@@ -0,0 +1,45 @@
+/*
+ *
+ *  * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
```
abti: Please change to use the Apache 2.0 License header without copyright

gobblin-core-base/src/main/java/gobblin/writer/GenericWriteResponseWrapper.java
```diff
@@ -0,0 +1,23 @@
+/*
+ *
+ *  * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
```
abti: ditto

shirshanka:
Committing, as I've addressed all the issues, and opened issues for the ones I haven't.

PR #1493, opened 2017-01-03
Error out if mvn is unavailable to build mock-couchbase

agepati:
In case mvn is not installed, the build progresses and CouchbaseWriterTest fails with a "Server is not up!" message.

This fix errors out the build in case mvn is not available to build the mock-couchbase package.

### DIFF
--- a/gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
+++ b/gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
@@ -10,5 +10,9 @@ if [ -d mock-couchbase/.git ];
     git -C mock-couchbase checkout b929a9f99a31a233c43bd94b6a696ad966206e02
   fi
 pushd mock-couchbase
-mvn package -Dmaven.test.skip=true
-popd
+if mvn package -Dmaven.test.skip=true; then
+  popd
+else
+  echo "Error building mock-couchbase"
+  exit 1
+fi
--- a/gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
+++ b/gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
@@ -10,9 +10,8 @@ if [ -d mock-couchbase/.git ];
     git -C mock-couchbase checkout b929a9f99a31a233c43bd94b6a696ad966206e02
   fi
 pushd mock-couchbase
-if mvn package -Dmaven.test.skip=true; then
-  popd
-else
+if ! mvn package -Dmaven.test.skip=true; then
   echo "Error building mock-couchbase"
   exit 1
 fi
+popd


agepati:
@shirshanka Could you please review?

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
```diff
@@ -10,5 +10,9 @@ if [ -d mock-couchbase/.git ];
     git -C mock-couchbase checkout b929a9f99a31a233c43bd94b6a696ad966206e02
   fi
 pushd mock-couchbase
-mvn package -Dmaven.test.skip=true
-popd
+if mvn package -Dmaven.test.skip=true; then
+  popd
```
shirshanka: nit: popd should happen regardless of whether mvn build succeeded.

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
```diff
@@ -10,5 +10,9 @@ if [ -d mock-couchbase/.git ];
     git -C mock-couchbase checkout b929a9f99a31a233c43bd94b6a696ad966206e02
   fi
 pushd mock-couchbase
-mvn package -Dmaven.test.skip=true
```
shirshanka: Should we explicitly check whether mvn is installed? That would allow us to provide a meaningful error message below. [mvn missing versus mvn build failed]

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
```diff
@@ -10,5 +10,9 @@ if [ -d mock-couchbase/.git ];
...
-popd
+if mvn package -Dmaven.test.skip=true; then
+  popd
+else
+  echo "Error building mock-couchbase"
```
shirshanka: Should we provide a more meaningful error message so the user can figure out why the build failed?

shirshanka:
Thanks for doing this, it was on my pending list. I added a few comments. We should be good to go after those are resolved.

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
agepati: I could see multiple failures possible: 
 - mvn might not be installed
 - mvn could be installed and misconfigured in which case I saw a 'Could not find main class...'  error
 - The downloading of the modules could cause disk to get filled up
 - There could be a transient network issue
... maybe more.

So I thought since we needed the command to fully succeed, we could conditionalize progress on the entire command to return 0.

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
agepati: The output already had the error message:

./scripts/install_test_deps.sh: line 13: mvn: command not found

that is now followed up with the new message on the next line: Error building mock-couchbase

Since there was no check on the return value earlier, popd followed and execution continued without stopping.

gobblin-modules/gobblin-couchbase/scripts/install_test_deps.sh
agepati: Oh okay, thanks! When I ran 'gradlew build', the end directory seemed to be undisturbed.

PR #1496, opened 2017-01-03
ZK/HelixPropertyStore based state store.

htran1:


htran1:
@chavdar please take a look.

PR #1497, opened 2017-01-03
Basic implementation for Gobblin broker.

ibuenros:
This PR has an implementation of a shared resources broker.

Suggested order of review:

1. Small changes to interfaces: Removed `AutoScopedSharedResourceFactory` and integrated its functionality into `SharedResourceFactory`. This simplifies usage a bit. For auto scoping, the broker will use the method `getAutoScope` to determine the scope at which a requested resource should be created. A simple factory will simply return `broker.leafScope()`.

2. `DefaultBrokerCache` stores the shared objects keyed by factory, key, and scope using guava cache. It is thread safe and handles all accesses to the cache.

3. `SharedResourcesBrokerImpl` is the actual implementation of `SharedResourcesBroker`. It uses a backing `DefaultBrokerCache` to get the shared objects. It also has a specific way of creating new brokers.

4. `KeyedScopedConfigViewImpl` is the implementation of `ConfigView` and `ScopedConfigView`. It knows how to access correct keys in a `Config`.

5. The other main classes are fairly straightforward.

6. A bunch of tests for the broker.


ibuenros:
@ydai1124 @chavdar can you review?

gobblin-utility/src/main/java/gobblin/broker/SharedResourcesBrokerImpl.java
```diff
@@ -0,0 +1,282 @@
...
+ * Instances of this broker must be created as follows:
+ * <pre>
+ *  SharedResourcesBrokerImpl<MyScopes> rootBroker = SharedResourcesBrokerFactory.createDefaultTopLevelBroker(myConfig);
+ *  SharedResourcesBrokerImpl<MyScopes> scopeBroker = topBroker.newSubscopedBuilder(scope, "scopeId").build();
+ * </pre>
+ */
+public class SharedResourcesBrokerImpl<S extends ScopeType<S>> implements SharedResourcesBroker<S> {
+
+  private final DefaultBrokerCache<S> brokerCache;
+  private final ScopeImpl<S> leafScope;
```
chavdar: nit: i find the naming slightly confusing since there is only one leaf. How about "selfScope" ?

gobblin-utility/src/main/java/gobblin/broker/SharedResourcesBrokerImpl.java
```diff
@@ -0,0 +1,282 @@
...
+  private final List<ScopedConfig<S>> scopedConfigs;
+  private final ImmutableMap<S, ScopeImpl<S>> scopeMap;
```
chavdar: please make the name a bit more specific, say, ancestorScopesByType .

ibuenros:
@ydai1124 addressed comments

PR #1498, opened 2017-01-04
Hive source watermark fixes

pcadabam:
## Bug description ##
`HiveSource` creates workunits even when there are no updates in the input

## Root Cause ##
`PartitionLevelWatermarker` creates dummy workunits that contain a vector of watermarks for every partition of a table. In case of _non-partitioned_ tables,`HiveSource` does not read previous watermarks from this vector. It uses `sourceState.getPreviousWorkUnitStatesByDatasetUrns()`. This causes a table to be processed every alternate execution. See below example

**Run 1**
Previous HighWatermark is read from dummy workunit which is 0 (watermark is null). Hence creates a real workunit with table's modTime as expected high watermark.

**Run 2**
Let's say there were no updates between Run 1 and Run 2
No real workunits are created and only a dummy workunit is created with watermark 0

**Run 3** 
Same as Run 1

## Fix ##
Fix is to disable dummy workunit creation for _non-partitioned_ tables

This PR also removes the usage of hive partition createTime as watermark check. This was a hack to fix a bug in distcp-ng. The current watermaking logic in `HiveSource` does not need this hack.

@abti can you review?


chavdar:
There are unit tests failures in PartitionLevelWatermarkerTest and HiveSourceTest

pcadabam:
Reopening.

PR #1499, opened 2017-01-04
Move kafka module dependency from gobblin-core to gobblin-distribution

chavdar:
This is to avoid circular dependencies in IDEs for gobblin-kafka-08 .


chavdar:
@shirshanka can you review?

shirshanka:
LGTM!

PR #1500, opened 2017-01-04
Make certain static variables final in gobblin-admin

chavdar:


chavdar:

PR #1501, opened 2017-01-05
Refactor static initialization in AvroToJdbcEntryConverter

chavdar:


chavdar:

PR #1502, opened 2017-01-05
Refactoring static usage in SalesForceRestWriterBuilder

chavdar:


chavdar:

PR #1505, opened 2017-01-05
Create and access WorkUnits and TaskStates through the StateStore int…

htran1:
…erface to remove dependency on having a shared file system.


htran1:
@chavdar please take a look.

chavdar:
Can you please add a short description of what the changes are that can guide me through doing the review?

htran1:
WorkUnits are presently serialized using SerializationUtils.serializeState() which writes using an Hadoop FS object. TaskState is written using an FsStateStore.  Changes are to store both of these state types using StateStores that can be backed by a non-filesystem storage.

StateStores store data with a key of (root, storeName, tableName). The root may be a path prefix in file system and ZK state stores.  For DB state store it is a DB table. storeName is the next level of the hierarchy. tableName is the last level and what is traditionally the file name, such as work_unit_name.wu is used for this field.

WorkUnits are stored at appWorkdir/_workunits/<job_id>/<work_unit_id>.wu.
MultiWorkUnits are stored at  appWorkdir/_workunits/<job_id>/<work_unit_id>.mwu
TaskStates are stored at  appWorkdir/_taskstates/<job_id>/<task_id>.tst

The path used in the test GobblinHelixTaskTest was changed to conform to the format above.

Three StateStores are instantiated through GobblinClusterUtils.StateStores for storing and accessing these states.  The StateStore type comes from the configuration key 'state.store.type'.
The configuration of 'state.store.root.dir' and 'state.store.db.table' are set before creating each StateStore to configure the proper root directory and root table.

The ParallelRunner is used to serialize and deserialize in parallel. A method that accepts a Callable was added to allow submitting the StateStore serialization and deserialization tasks. This generic interface was required due to avoid a build dependency issue.

StateStore changes:
Changed the ZK state put calls to create with data atomically when the znode does not exist.
Added a getStateNames() method for listing. This is required by the TaskStateCollectorService to find data to read.
Added factory method to StateStore and refactored some code shared between the DatasetStateStore factory and StateStore factory.

htran1:
@ibuenros please take a look.

gobblin-runtime/src/main/java/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -115,30 +115,32 @@ protected void shutDown() throws Exception {
...
+        LOGGER.debug("Found output task state file " + taskStateName);
         // Deserialize the TaskState and delete the file
-        stateSerDeRunner.deserializeFromSequenceFile(Text.class, TaskState.class, status.getPath(), taskStateQueue,
-            true);
+        stateSerDeRunner.submitCallable(new Callable<Void>() {
+          @Override
+          public Void call() throws Exception {
+            TaskState taskState = taskStateStore.getAll(outputTaskStateDir.getName(), taskStateName).get(0);
+            taskStateQueue.add(taskState);
+            taskStateStore.delete(outputTaskStateDir.getName(), taskStateName);
```
ibuenros: This is a bit risky because certain key-value structures (for example an in-memory map) do not allow concurrent modification. Ideally, the delete should be done after the loop has completed.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -163,8 +165,14 @@ public MRJobLauncher(Properties jobProps, Configuration conf)
     this.parallelRunnerThreads = Integer.parseInt(jobProps.getProperty(ParallelRunner.PARALLEL_RUNNER_THREADS_KEY,
         Integer.toString(ParallelRunner.DEFAULT_PARALLEL_RUNNER_THREADS)));
 
+    // StateStore interface uses the following key (rootDir, storeName, tableName)
+    // The state store base is the root directory and the last two elements of the path are used as the storeName and
+    // tableName. Create the state store with the root at jobOutputPath. The task state will be stored at
+    // jobOutputPath/output/taskState.tst, so output will be the storeName.
+    taskStateStore = new FsStateStore<>(this.fs, jobOutputPath.toString(), TaskState.class);
```
ibuenros: why does this not support arbitrary state stores?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterUtils.java
```diff
@@ -57,4 +64,52 @@ public static Path getAppWorkDirPath(FileSystem fs, String applicationName, Stri
...
     return applicationName + Path.SEPARATOR + applicationId;
   }
+
+  /**
+   * state stores used for storing work units and task states
+   */
+  public static class StateStores {
+    public final StateStore<TaskState> taskStateStore;
+    public final StateStore<WorkUnit> wuStateStore;
+    public final StateStore<MultiWorkUnit> mwuStateStore;
```
ibuenros: why do we need different stores for wu and mwu?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinHelixTask.java
```diff
@@ -113,10 +111,15 @@ public TaskResult run() {
       Path workUnitFilePath =
           new Path(this.taskConfig.getConfigMap().get(GobblinClusterConfigurationKeys.WORK_UNIT_FILE_PATH));
 
-      WorkUnit workUnit =
-          workUnitFilePath.getName().endsWith(AbstractJobLauncher.MULTI_WORK_UNIT_FILE_EXTENSION) ? MultiWorkUnit
-              .createEmpty() : WorkUnit.createEmpty();
-      SerializationUtils.deserializeState(this.fs, workUnitFilePath, workUnit);
+      String fileName = workUnitFilePath.getName();
```
ibuenros: There are still a lot of references to file system style storage (e.g. file, directory, etc.) that might be confusing in the future. How hard would it be to refactor into the correct terms for a generic store?

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -92,6 +95,29 @@ public FsStateStore(String storeUrl, Class<T> stateClass) throws IOException {
...
+      }
+
+      try {
+        String stateStoreFsUri = props.getProperty(ConfigurationKeys.STATE_STORE_FS_URI_KEY,
+            ConfigurationKeys.LOCAL_FS_URI);
+        FileSystem stateStoreFs = FileSystem.get(URI.create(stateStoreFsUri), conf);
+        String stateStoreRootDir = props.getProperty(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY);
+
+        return new FsStateStore(stateStoreFs, stateStoreRootDir, stateClass);
+      } catch (Exception e) {
```
ibuenros: You should be catching a more concrete exception here if possible.

gobblin-runtime/src/main/java/gobblin/runtime/TaskStateCollectorService.java
htran1: This is a delete on the taskStateStore, not the taskStateNames being iterated over.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterUtils.java
htran1: This is due to the deserialization. The extension in the state name of wu or mwu indicates which type to deserialize to and this logic is external to the StateStore.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
htran1: MR has a shared file system, so there is less of a need for non-fs state store, but I guess it doesn't hurt to have the flexibility, so I'll add the code for arbitrary store type.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
```diff
@@ -143,6 +149,17 @@ public T get(String storeName, String tableName, String stateId)
...
 
   /**
+   * Get state names under the storeName
+   *
+   * @param storeName store name
+   * @param predicate only returns names matching predicate
+   * @return (possibly empty) list of state names from the given store
+   * @throws IOException
+   */
+  public List<String> getStateNames(String storeName, Predicate<String> predicate)
```
chavdar: What is a "state name"? Currently, there is no such concept. State id?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterUtils.java
```diff
@@ -57,4 +64,52 @@ public static Path getAppWorkDirPath(FileSystem fs, String applicationName, Stri
   public static String getAppWorkDirPath(String applicationName, String applicationId) {
     return applicationName + Path.SEPARATOR + applicationId;
   }
+
+  /**
+   * state stores used for storing work units and task states
+   */
+  public static class StateStores {
```
chavdar: I don't see anything cluster specific in this class. Eventually, we may want to use something like that in non-cluster use-cases. Further, I think it is important enough to be a top-level class.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
```diff
@@ -34,6 +36,10 @@
  */
 public interface StateStore<T extends State> {
 
+  interface Factory {
+    <T extends State> StateStore<T> createStateStore(Properties props, Class<T> stateClass);
```
chavdar: I know I've missed this in the previous commit (for DatasetStateStore.Factory), but is it too late to switch to typesafe Config instead of Properties? Overall, we've been trying to move away from the latter and to the former  .

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
```diff
@@ -81,6 +84,24 @@ public ZkStateStore(String connectString, String storeRootDir, boolean compresse
     propStore = new ZkHelixPropertyStore<byte[]>(connectString, serializer, storeRootDir);
   }
 
+  @Alias("zk")
+  public static class Factory implements StateStore.Factory {
```
chavdar: It is a bit of a subtlety but it can come to byte us later. We've been moving away from Properties and toward typesafe Config. Overall, the former supports the Properties file format with some caveats. If we don't use the alias but we decide to use the class FQN, it has to be something like: gobblin.metastore.ZkStateStore$Factory.

In properties format, one can write
state.store.type= gobblin.metastore.ZkStateStore$Factory

That does not work if we try to read this with typesafe Config since it will try to expand Factory as a variable reference. So we have to write 

state.store.type='gobblin.metastore.ZkStateStore$Factory'

Reading that as a Properties collection will include the quotes.

Instead, if we have a top level class gobblin.metastore.ZkStateStoreFactory, we can have 

state.store.type=gobblin.metastore.ZkStateStoreFactory

This is compatible with both.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
htran1: Changed this to getTableNames to match the existing StateStore terminology.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterUtils.java
htran1: Moved to gobblin.runtime.util.StateStores.

gobblin-modules/gobblin-helix/src/main/java/gobblin/metastore/ZkStateStore.java
htran1: Created top-level classes.

gobblin-metastore/src/main/java/gobblin/metastore/StateStore.java
htran1: Changed to typesafe config.

gobblin-runtime/src/main/java/gobblin/runtime/mapreduce/MRJobLauncher.java
htran1: I spent some time working on this, but hit some issues with the file format change for the workunits which are not currently stored in a StateStore. Since this is not essential and likely won't be used, I would prefer to leave the code alone to avoid inadvertently introducing bugs or compatibility issues.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinHelixTask.java
htran1: Yes, we need to decide on some generic names, but I think this will require a broader discussion and more global cleanup. It seems like most hierarchical storage systems will be similar to a file system, so the file system terminology may mostly work. The current StateStore has the following terms: storeRootDir, stateStoreTableName, storeName, and tableName. TableName was probably meant to be a generic term instead of calling it fileName, but it is confusing for both FS and DB state stores. TableName refers to the last level of the key. FileName would be the closest match for FS and maybe itemName would be a generic equivalent.

gobblin-cluster/src/test/java/gobblin/cluster/GobblinHelixTaskTest.java
```diff
@@ -96,7 +98,7 @@ public void setUp() throws IOException {
   }
 
   @Test
-  public void testPrepareTask() throws IOException {
+  public void testPrepareTask() throws Exception {
```
ibuenros: Please throw a more concrete exception.

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStoreFactory.java
```diff
@@ -0,0 +1,53 @@
...
+
+    try {
+      String stateStoreFsUri = ConfigUtils.getString(config, ConfigurationKeys.STATE_STORE_FS_URI_KEY,
+          ConfigurationKeys.LOCAL_FS_URI);
+      FileSystem stateStoreFs = FileSystem.get(URI.create(stateStoreFsUri), conf);
+      String stateStoreRootDir = config.getString(ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY);
+
+      return new FsStateStore(stateStoreFs, stateStoreRootDir, stateClass);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
ibuenros: Please include a more precise message with the Exception. E.g. `Failed to start FsStateStoreFactory`.

gobblin-metastore/src/main/java/gobblin/metastore/MysqlStateStore.java
```diff
@@ -299,6 +338,29 @@ public T get(String storeName, String tableName, String stateId) throws IOExcept
...
+      try (ResultSet rs = queryStatement.executeQuery()) {
+        while (rs.next()) {
+          String name = rs.getString(1);
+          if (predicate.apply(name)) {
+            names.add(name);
+          }
+        }
+      }
+    } catch (SQLException e) {
+      throw new IOException(e.getMessage());
```
ibuenros: Same, please put a message in exception.

gobblin-metastore/src/main/java/gobblin/metastore/MysqlStateStoreFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    BasicDataSource basicDataSource = MysqlStateStore.newDataSource(config);
+    String stateStoreTableName = ConfigUtils.getString(config, ConfigurationKeys.STATE_STORE_DB_TABLE_KEY,
+        ConfigurationKeys.DEFAULT_STATE_STORE_DB_TABLE);
+    boolean compressedValues = ConfigUtils.getBoolean(config, ConfigurationKeys.STATE_STORE_COMPRESSED_VALUES_KEY,
+            ConfigurationKeys.DEFAULT_STATE_STORE_COMPRESSED_VALUES);
+
+    try {
+      return new MysqlStateStore(basicDataSource, stateStoreTableName, compressedValues, stateClass);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
ibuenros: Please put message in exception.

PR #1507, opened 2017-01-05
Changed shared resource broker scopes to be less restrictive.

ibuenros:
This PR changes `ScopeInstance` to make them less restrictive.

* `ScopeInstance` is now an even more generic interface, and the broker only interacts with it using `equals`, `hashCode`, `getType`, and `toString` methods. This allows scopes to store additional metadata, for example a job name and job id, instead of a single generic id.
* `ScopeType` now has a method `rootScope()` that returns the root of the DAG of scope types.
* Brokers with `null` scope are no longer allowed. The top level broker is now at the `rootScope()` level.

The effective changes in code are as follows:
* `ScopeInstanceImpl` becomes `ScopeWrapper` which is used internally to store a `ScopeInstance` and references to its ancestors. A lot of changes are just renaming this.
* Public methods on the broker still return `ScopeInstance`. `SharedResourcesBrokerImpl` additionally has package-private methods to get the wrapped scope.
* Added `SubscopedBrokerBuilder` interface (equivalent to the existing `SharedResourcesBrokerImpl.SubscopedBrokerBuilder`) as the only correct way of creating sub brokers.
* `GobblinScopes` became `GobblinScopeTypes`. Added `GobblinScopeInstance`, `JobScopeInstance`, and `TaskScopeInstance` that will have job and task metadata respectively (as needed).
* Changes in tests are just updates to the slightly different way of creating brokers.
* Added a test for `GobblinScopeInstance`.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/broker/DefaultBrokerCache.java
+++ b/gobblin-utility/src/main/java/gobblin/broker/DefaultBrokerCache.java
@@ -47,7 +47,7 @@ import lombok.extern.slf4j.Slf4j;
 class DefaultBrokerCache<S extends ScopeType<S>> {
 
   private final Cache<RawJobBrokerKey, Object> sharedResourceCache;
-  private final Cache<RawJobBrokerKey, ScopeImpl<S>> autoScopeCache;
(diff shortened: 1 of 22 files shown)


ibuenros:
@ydai1124  can you review?

PR #1514, opened 2017-01-07
Changed shared resource broker scopes to be less restrictive.

ibuenros:


ibuenros:
Same as PR 1507, which is approved.

PR #1515, opened 2017-01-07
gobblin-metastore tests on travis use container mysql

chavdar:
Hotfix

### DIFF
--- a/gobblin-metastore/build.gradle
+++ b/gobblin-metastore/build.gradle
@@ -77,6 +77,9 @@ artifacts {
 
 test {
     workingDir rootProject.rootDir
+    error {
+       showStandardStreams true
+    }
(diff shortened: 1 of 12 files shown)


gobblin-metastore/build.gradle
```diff
@@ -77,6 +77,9 @@ artifacts {
 
 test {
     workingDir rootProject.rootDir
+    if (project.hasProperty('usePreinstalledMysql') && Boolean.parseBoolean(project.usePreinstalledMysql)) {
+      jvmArgs '-Dgobblin.metastore.testing.embeddedMysqlEnabled=false'
```
abti: how are pre installed credentials picked up? are the hardcoded in the tests and expected to be the same? what if the state (test tables, etc) are not cleaned up from pre-installed mysql if tests fail / jvm crashes?

chavdar: This is only for Travis. The container sets up a mysql instance with the passwordless root user.

PR #1516, opened 2017-01-07
Add com.linkedin.gobblin to alias resolver

pcadabam:
@ibuenros can you review?

### DIFF
--- a/gobblin-api/src/main/java/gobblin/util/ClassAliasResolver.java
+++ b/gobblin-api/src/main/java/gobblin/util/ClassAliasResolver.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
 
 import org.reflections.Reflections;
+import org.reflections.util.ConfigurationBuilder;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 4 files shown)


gobblin-api/src/test/java/com/linkedin/gobblin/TestAlias.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) 2014-2016 LinkedIn Corp. All rights reserved.
```
abti: Please change to use the Apache 2.0 License header without copyright

gobblin-api/src/main/java/gobblin/util/ClassAliasResolver.java
```diff
@@ -38,16 +39,17 @@
...
  * </b>
  */
 @Slf4j
 public class ClassAliasResolver<T> {
 
-  // Scan all packages in the classpath with prefix gobblin when class is loaded.
+  // Scan all packages in the classpath with prefix gobblin, com.linkedin.gobblin when class is loaded.
   // Since scan is expensive we do it only once when class is loaded.
-  private static final Reflections REFLECTIONS = new Reflections("gobblin");
+  private static final Reflections REFLECTIONS = new Reflections(new ConfigurationBuilder().forPackages("gobblin",
```
abti: This should be configurable. Also, why are we scanning classes within com.linkedin.gobblin package, when the packages are gobblin.* in our code base?

pcadabam: Linkedin internal packages are `com.linkedin.gobblin`

kadaan:
It would be ideal if the classpath scanner package list could be configured.  Trying to add a new implementation of a class that is loaded via Alias doesn't work currently unless I put it within the gobblin package.  The same is true when trying to use this in another context.

abti:
https://issues.apache.org/jira/browse/GOBBLIN-65

Please update your PR title with following prefix: [GOBBLIN-65]

PR #1517, opened 2017-01-07
Fix for compacting partitioned folders

treff7es:
Compaction did not worked for me when I used the following timebased folder pattern:
compaction.timebased.folder.pattern='dt='yyyy-MM-dd/'hr'=HH

I checked the code and the getFolderStructure method and it only replaces a-zA-Z0-9 but not any signs like ', = or - which was in my folder pattern to achieve the following directory pattern: /dt=2017-01-01/hr=12/

I added those characters to the regexp and now it works fine for me.

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/dataset/TimeBasedSubDirDatasetsFinder.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/dataset/TimeBasedSubDirDatasetsFinder.java
@@ -17,11 +17,12 @@
 
 package gobblin.compaction.dataset;
 
-import java.io.IOException;
-import java.util.Set;
-
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Sets;
+import gobblin.compaction.mapreduce.MRCompactor;
+import gobblin.configuration.State;
+import gobblin.util.DatasetFilterUtils;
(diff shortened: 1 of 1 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/dataset/TimeBasedSubDirDatasetsFinder.java
```diff
@@ -140,7 +136,7 @@ protected String getDatasetName(String path, String basePath) {
             try {
               folderTime = getFolderTime(jobInputPath, inputPath);
             } catch (RuntimeException e) {
-              log.warn(jobInputPath + " is not a valid folder. Will be skipped.");
+              log.warn(jobInputPath + " is not a valid folder. Will be skipped. Error: "+e);
```
shirshanka: Try this: 
log.warn("{} is not a valid folder. Will be skipped due to exception.", jobInputPath, e);

http://www.slf4j.org/faq.html#paramException

treff7es:
Thanks @shirshanka, I rewrote the warning message.

shirshanka:
@treff7es: Thanks! Can you rebase the branch and push so the base builds on Travis turn green? The full build is broken due to tests and we'll fix it tomorrow. 

@ibuenros : Can you confirm that the regex change looks good for our usage as well?

PR #1519, opened 2017-01-08
Watermark awareness for console writer, better watermark mgmt helpers

shirshanka:


shirshanka:
Self-approving as changes are trivial and improve on the build breakage situation.

PR #1521, opened 2017-01-09
Added skipping feature to MultiWorkUnits

aditya1105:
This change will allow MultiWorkUnits to be skipped. If some WorkUnits are marked are skipped in MultiWorkUnit they can be skipped too. Modified the test for skipping of MultiWorkUnits.


abti:
Why would we create workunits if we intend to skip them?

aditya1105:
https://github.com/linkedin/gobblin/issues/1314

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -245,18 +247,20 @@ public void apply(JobListener jobListener, JobContext jobContext)
     final List<WorkUnit> skippedWorkUnits = new ArrayList<>();
     for (WorkUnit workUnit : workUnits.get()) {
       if (workUnit instanceof MultiWorkUnit) {
```
agepati: Would it be possible to push the removeSkippedWorkUnits() method down to the WorkUnit, and the MultiWorkUnit classes, so that this code can read:
```java
for (WorkUnit wu: workUnits.get()) {
    wu.removeSkippedWorkUnits();
}
```
or somewhat similar? and then the particular logic to skip can move into the WorkUnit classes?

gobblin-runtime/src/test/java/gobblin/TestSkipWorkUnitsSource.java
```diff
@@ -69,7 +85,8 @@ public void testSkipWorkUnitPersistence(SourceState state) {
         skipCount++;
       }
     }
-    Assert.assertEquals(skipCount, NUMBER_OF_SKIP_WORKUNITS,
+    // Plus one for MultiWorkUnit1, and plus 2 for MultiWorkUnit2
```
agepati: Is there any chance we can have separate tests for skipping MultiWorkUnits (altogether, and specific WorkUnits within MultiWorkUnit?)

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
aditya1105: 1) At multiple places, list of WorkUnits will be passed and invoked. It's essential to remove those WorkUnits from the list.
2) In order to support skipping of some WorkUnit inside of a MultiWorkUnits, this is essential.

gobblin-runtime/src/test/java/gobblin/TestSkipWorkUnitsSource.java
aditya1105: Since state store only store taskstates, a total count will suffice. It would take an extra run of Gobblin-Job to separate them out.

gobblin-api/src/main/java/gobblin/source/workunit/MultiWorkUnit.java
```diff
@@ -152,4 +152,14 @@ public int hashCode() {
   public static MultiWorkUnit createEmpty() {
     return new MultiWorkUnit();
   }
+
+  public void removeWorkUnits(WorkUnitStream workUnitsToRemove) {
+    for (WorkUnit workUnit : workUnitsToRemove.getMaterializedWorkUnitCollection()) {
```
ibuenros: Please don't use `getMaterializedWorkUnitCollection`, it prevents optimizations in the Gobblin flow. Instead, you should use a lambda on the `WorkUnitStream`.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -280,31 +281,44 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
+   * If complement is set to true, then the filter will be reversed
+   *
+   * If a MultiWorkUnit is marked as skipped all WorkUnits contained in it will be skipped.
+   * If a MultiWorkUnit contains some skipped WorkUnits, only those will be skipped.
    */
   @RequiredArgsConstructor
+  @AllArgsConstructor
   private static class SkippedWorkUnitsFilter implements Predicate<WorkUnit> {
     private final JobState jobState;
+    private boolean complement = false;
```
ibuenros: Use `Predicates.not(Predicate)` instead where necessary.

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -280,31 +281,44 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
 
     @Override
     public boolean apply(WorkUnit workUnit) {
       if (workUnit instanceof MultiWorkUnit) {
-        Preconditions.checkArgument(!workUnit.contains(ConfigurationKeys.WORK_UNIT_SKIP_KEY),
-            "Error: MultiWorkUnit cannot be skipped");
-        for (WorkUnit wu : ((MultiWorkUnit) workUnit).getWorkUnits()) {
-          Preconditions.checkArgument(!wu.contains(ConfigurationKeys.WORK_UNIT_SKIP_KEY),
-              "Error: MultiWorkUnit cannot contain skipped WorkUnit");
+        WorkUnitStream wus = ((MultiWorkUnit) workUnit).getWorkUnitsStream();
```
ibuenros: Is it necessary to use a `WorkUnitStream` for this? I think you could just do
```java
workUnit.getWorkUnits.stream().filter(new SkippedWorkUnitsFilter(jobState)).collect();
```
or using guava
```java
Iterables.filter(workUnit.getWorkUnits, Predicates.not(new SkippedWorkUnitsFilter(jobState)));
```

gobblin-runtime/src/main/java/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -676,6 +690,10 @@ protected void forWorkUnit(WorkUnit workUnit) {
       workUnit.setId(taskId);
       workUnit.setProp(ConfigurationKeys.TASK_ID_KEY, taskId);
       workUnit.setProp(ConfigurationKeys.TASK_KEY_KEY, Long.toString(Id.Task.parse(taskId).getSequence()));
+      // In case of MultiWorkUnit, WorkUnit(s) should be assigned different ids
+      if (workUnit instanceof MultiWorkUnit) {
+        ((MultiWorkUnit) workUnit).getWorkUnitsStream().transform(this);
```
ibuenros: Is it valid to have a `MultiWorkUnit` inside another `MultiWorkUnit`? I believe this case is already taken care of by `MultiWorkUnitForEach`.

abti:
https://issues.apache.org/jira/browse/GOBBLIN-64

Please update your PR title with following prefix: [GOBBLIN-64]

PR #1524, opened 2017-01-10
Enabling current Gobblin cluster code to run as a standalone cluster

abti:
Added various scripts and fixed bugs to enable current Gobblin cluster code to run as a standalone cluster. 

Steps to run Gobblin standalone cluster on your *nix based machines: 

**A. Build** 
1. Checkout the code and build (or use Gobblin distribution after this is released)
2. Build using: ./gradlew clean build 
3. A distribution gz will be built with the name of: gobblin-distribution-0.9.0-*.tar.gz

**B. Zookeeper Setup**
1. Download Zookeeper from: http://www.apache.org/dyn/closer.cgi/zookeeper/ and extract it to say /foo/bar/zk/
2. Run Zookeeper: /foo/bar/zk/bin/zkServer.sh start 
(default endpoint is 2181, which is what is set in Gobblin conf as well, so you need not edit zoo.cfg)
Bonus: You may want to run Exhibitor to observe your Zookeeper (really helpful to see how the job execution is progressing later) 

**C. Gobblin Setup**
1. Extract distribution: gobblin-distribution-0.9.0-*.tar.gz to /foo/bar; it will create a sub directory: /foo/bar/gobblin-dist/ 
2. Create a directory for your job configurations (such as wikipedia.pull) and copy your pull files there, say: /foo/bar/job-conf/ 
3. Edit /foo/bar/gobblin-dist/conf/standalone/application.conf and set gobblin.cluster.job.conf.path to /foor/bar/job-conf/ 
4. Minor hack (pending bug fix): Copy /foo/bar/gobblin-dist/conf/standalone/log4j-cluster.properties to /foo/bar/gobblin-dist/ 
Bonus: Defaults should work fine, but if you prefer you can edit conf files (application.conf and reference.conf within /foo/bar/gobblin-dist/conf/standalone/) to change work dir, etc. 

**D. Run Standalone Cluster**
Within directory: /foo/bar/gobblin-dist/
1. Start master by running: ./bin/gobblin-cluster-master.sh start 
2.1. Start one worker by running: ./bin/gobblin-cluster-worker.sh start
2.2. OR Start multiple workers by running: 
./bin/gobblin-cluster-worker.sh start --workerid 1 
./bin/gobblin-cluster-worker.sh start --workerid 2 
.. and so on
Note: Each instance will get its own log file, such as: master.out, worker.single.out (if no workerid is specified) or worker.1.out, worker.2.out (if workerid is specified), etc.  
3. In the logs you will see your job / task execution details and the job output in respective directories as per your pull file 

**Mini runbook:** 
1. Build fails for random test failure: Try building again, Gobblin suffers from transient test failures which has mostly been resolved but still is work in progress. 
2. Gobblin fails to connect to Zookeeper and you see Connection Refused in the logs: check if Zookeeper is bind on port 2181, if not check Zookeeper conf and logs, and make sure /foo/bar/gobblin-dist/conf/standalone/application.conf has the right zookeeper connection string 
3. Everything seems correct but job was not run: check if you have set the gobblin.cluster.job.conf.path correctly to where your job configuration files reside 
4. You added new pull file / changed existing pull file to job conf directory and it was not picked up: Its expected behavior, the standalone cluster only reads all the pull files at the time of bootup. So, you will have to restart the cluster


abti:
Changes tested and used with Streaming MVP, hence merging.

PR #1526, opened 2017-01-10
Integrated broker into Gobblin ingestion.

ibuenros:
This PR integrates a `SharedResourcesBroker` into gobblin execution, which is available through `WorkUnitState.getTaskBroker()`.

Changes:
* Moved a few classes from `gobblin-utility` to `gobblin-api` because `WorkUnitState` is in api.
* Added instance variable `taskBroker` to `WorkUnitState`, as well as getter. This broker is only guaranteed to be available from within a task. In particular, on deserialization of the `TaskState` on the driver, the task broker is not populated. 
* Fixed a bug in creation of sub scoped brokers, which was using the wrong ancestor brokers for scopes with multiple parents (for multi_task_attempt has as parents both container and job, the creation of a multi_task_attempt scope instance was failing).
* Added the needed brokers to the entire instance, job, task creation of the gobblin framework.
* Added a test to verify that the task broker can be acquired and that it is indeed shared as expected.


ibuenros:
@chavdar can you review? Thanks!

PR #1527, opened 2017-01-10
Add isFirstPublish metadata in successful data publish event for tracking

abti:
Add isFirstPublish metadata in successful data publish event for tracking


abti:
@pcadabam Can you please review?

PR #1539, opened 2017-01-13
Increase a timeout in TestQuartzJobSpecScheduler

chavdar:


chavdar:
Self approving a trivial change.

PR #1540, opened 2017-01-13
Making gobblin-admin a module

chavdar:
- Removed compile dependencies on gobblin-admin -- the main challenge was the use of the ADMIN_SERVER_ENABLED_KEY flag rather than dependency injection through config. I had to do the latter but adding a factory interface to gobblin-api and adding ADMIN_SERVER_FACTORY_CLASS_KEY . Eventually, I think the whole implementation has to refactored into a GobblinInstancePlugin that can be specified when the instance is configured.
- Made gobblin-admin as part of the "full" flavored build

@abti can you review ?

Main changes:
- New interface AdminWebServerFactory
- New default implementation DefaultAdminWebServerFactory which preserves existing functionality
- Adding new dependency injection for the implementation through configs: ConfigurationKeys , ServiceBasedAppLauncher, GobblinYarnAppLauncher,


chavdar:
+ @kadaan for review

chavdar:
Ping.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -603,15 +603,20 @@
...
   public static final String REST_SERVER_ADVERTISED_URI_KEY = "rest.server.advertised.uri";
 
-  /**
+  /*
    * Admin server configuration properties.
    */
   public static final String ADMIN_SERVER_ENABLED_KEY = "admin.server.enabled";
+  /** The name of the class with the admin interface. The class must implement the
+   * AdminWebServerFactory interface .*/
+  public static final String ADMIN_SERVER_FACTORY_CLASS_KEY = "admin.server.type";
```
kadaan: Does the naming of this match how other factories are configured in gobblin?  For instance it seems to make more sense if the key is: `admin.server.factory.type`.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
```diff
@@ -224,13 +226,30 @@ private void addJobExecutionServerAndAdminUI(Properties properties) {
...
+        addService(createAdminServer(properties, executionInfoServer.getAdvertisedServerUri()));
       }
     } else if (adminUiServerEnabled) {
       LOG.warn("Not launching the admin UI because the job execution info server is not enabled");
     }
   }
 
+  public static Service createAdminServer(Properties properties,
+                                          URI executionInfoServerURI) {
+    String factoryClassName = properties.getProperty(ConfigurationKeys.ADMIN_SERVER_FACTORY_CLASS_KEY,
```
kadaan: On a side note, in a Spring app this could be a lot cleaner as it would be able to be autoconfigured and conditional on the class being in the classpath.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
chavdar: Fixed.

gobblin-runtime/src/main/java/gobblin/runtime/app/ServiceBasedAppLauncher.java
chavdar: How does that work?

kadaan: I can explain more if you are interested, but here is the doc for now: http://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-auto-configuration.html

abti:
https://issues.apache.org/jira/browse/GOBBLIN-61

Please update your PR title with following prefix: [GOBBLIN-61]

PR #1542, opened 2017-01-13
Create cluster build

chavdar:
- Created the "cluster" build flavor and made gobblin-cluster, gobblin-yarn, gobblin-aws part of it.
- Enhanced the build for gobblin-distribution to include the build flavor in the name of the artifact if not "standard". For example, the "cluster" build will generate gobblin-distribution-cluster artifact.

For now, I haven't moved the above modules to the gobblin-modules directory but this can be changed in the future. 

@abti @kadaan  can you review?


chavdar:
Ping.

gobblin-distribution/gobblin-flavor-cluster.gradle
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compile project(':gobblin-aws')
+  compile project(':gobblin-modules:gobblin-azkaban')
```
abti: why is gobblin-azkaban part of this?

chavdar:
@kadaan Here are the docs: http://gobblin.readthedocs.io/en/latest/developer-guide/GobblinModules/

gobblin-distribution/gobblin-flavor-cluster.gradle
chavdar: gobblin-aws//src/main/java/gobblin/aws/GobblinAWSConfigurationKeys.java seems to reference azkaban

kadaan: @abti Ping

abti: That seems to only reference to the jar name to download from s3 and use in classpath when launching the master / worker, but does not has any real dependency on Azkaban.

abti:
https://issues.apache.org/jira/browse/GOBBLIN-60

Please update your PR title with following prefix: [GOBBLIN-60]

kadaan:
@abti Is this still something we want to merge?

abti:
@kadaan yes, but we need to resolve conflicts

Issue #1538, opened 2017-01-13
Intermittent couchbase tests failures


chavdar:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/191542925/log.txt

chavdar:
This seems to be happening fairly frequently.

shirshanka:
This is resolved by 
https://github.com/linkedin/gobblin/commit/d36d9cf6f23948ad45aef4cc2d947e626a751b82

PR #1546, opened 2017-01-13
Fix flaky tests described in #1538

shirshanka:
Will let people know when this is ready for review.


shirshanka:
Merging in as remaining test failures are unrelated.

PR #1548, opened 2017-01-13
Adding some skeleton docs for sources and writers

chavdar:


chavdar:
Self-approving documentation change

PR #1551, opened 2017-01-15
Fixing documentation build breakages

shirshanka:
There were broken links.


shirshanka:
Self-approving docs change.

Issue #840, opened 2016-03-11
Losing events due to unlikely race in Fork::processRecords()?


Kaiserchen:
It looks like there is an unlikely race in the metioned method in a situation where the extractor would stop pulling records for some time triggering the fork runner to time out on the queue, then publish some records and then calling markParentTaskDone before the forkrunner checks it? 

Isn't something else cleaner like publishing a done message into the queue or something? Am I missing something that makes it not racy? 

I am really looking forward to migrate away from Camus, Gobblin looks really really good :) thank you for all this great software. I am actually quite tired right now. I hope there are not to many more typos :(

sahilTakiar:
hey @Kaiserchen 

Yeah that seems like a possible race condition, in the `Fork.processRecords` method. To provide more details here a list of steps as to how this could happen:

Task Thread launches Fork Thread that listens for records on the `recordQueue` and processes them
Task Thread extracts 1 record from the `Extractor`
Task Thread puts the record on the `recordQueue`
Fork Thread processes the record and writes it, and then invokes `recordQueue.get()` to wait for the next record
Task Thread tries to extract a record, but the extractor stalls waiting for an I/O event
Fork Thread's `recordQueue.get()` method times out, and the operation returns null
Task Thread finally gets the extractor to return a record, the record is put on the queue
Task Thread's Extractor returns null, and it tells the Fork there are no more records coming
Fork Thread then looks at the record of `recordQueue.get` and see that the result is `null` because the `get()` timed out, it then sees that the `Task` is also done processing all records, so it thinks there is no more work to be done and then it returns

In this case the last record will never get written (resulting in missing data) and there will be no failure of any kind.

Kaiserchen:
The Io event could potentially block fetching many records that get "cached in the extractor" so it could start a burst squeezing even more events in, making it more unlikly that it makes it ways to set the done flag, but no gurantees that even many records might be missed. Shall i prepare a patch where the mark done flag puts a static object in as record and in the process i check if it was the done record? also would make the done notification immidiate ( without waiting).

Kaiserchen:
I submitted a pull request #841 maybe someone can take a look?

sahilTakiar:
@Kaiserchen yes this approach looks good, thanks for the PR. I just made a few minor comments. Overall it looks good.

Kaiserchen:
See comment in #841. I would be glad if someone else picks this up. I don't have the time for that currently.

PR #1552, opened 2017-01-17
Optimize tasks with only one branch by removing the need for an intermediary queue.

kadaan:
@sahilTakiar Can you get this to the right person for review?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,6 +42,7 @@ buildscript {
     classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:1.0.2'
     classpath 'gradle.plugin.com.palantir:jacoco-coverage:0.3.0'
     classpath 'io.spring.gradle:dependency-management-plugin:0.5.4.RELEASE'
+    classpath 'me.champeau.gradle:jmh-gradle-plugin:0.3.1'
   }
 }
 
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -680,4 +680,11 @@ public class ConfigurationKeys {
    */
   public static final String TASK_EXECUTION_MODE = "gobblin.task.executionMode";
   public static final String DEFAULT_TASK_EXECUTION_MODE = "BATCH";
+
+  /**
+   * Configuration properties related to optimizations for single branch tasks
+   */
+  public static final String TASK_IS_SINGLE_BRANCH_SYNCHRONOUS = "gobblin.task.is.single.branch.synchronous";
+  public static final String DEFAULT_TASK_IS_SINGLE_BRANCH_SYNCHRONOUS = Boolean.toString(false);
+
 }
--- a/gobblin-runtime/build.gradle
+++ b/gobblin-runtime/build.gradle
@@ -16,6 +16,7 @@
  */
 
 apply plugin: 'java'
+apply plugin: 'me.champeau.gradle.jmh'
 
 dependencies {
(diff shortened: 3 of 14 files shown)


sahilTakiar:
@shirshanka @chavdar does someone have time to review this? I think we should consider removing the intermediate queue all together. It makes the logic complex and can lead to race conditions (e.g. #840). Unless others have seen concrete scenarios where it makes a performance difference? Last I remember I hadn't seen any such cases.

kadaan:
I think it "could" help when there are multiple forks if each fork pulls independently. That said, I'm not sure how many forks are needed to see a perfect gain over SyncFork.

kadaan:
BTW, you can now run JMH benchmarks via `./gradlew jmh`

kadaan:
I believe failing test is due to the shutdown taking slightly longer than the given 1s.  The streaming code has a 1s timeout on the poll loop and when shutdown is called that time + time for the stack to unwind end up being longer than the test waits.  To quote gitter:

> Joel Baranick @kadaan Jan 20 10:27
> I changed to code in that test to:
> long startTime = System.currentTimeMillis();
> boolean success = task.awaitShutdown(10000);
> Assert.assertTrue(success, "Task should shutdown in 10 seconds");
> log.info("Task done waiting to shutdown {} after {}ms", success, System.currentTimeMillis() - startTime);
> what I see in the logs is: 2017-01-20 10:26:44 PST INFO [main] gobblin.runtime.TaskContinuousTest 223 - Task done waiting to shutdown true after 1007ms
> So it likely fails due to taking slightly longer than the 1s requested.
> Am I correct that the assumption that the task will shutdown in 1s is because of this in Task: long commitIntervalMillis = 1000; // TODO: Configure
> 
> Joel Baranick @kadaan Jan 20 10:35
> Also, is there a reason we need to verify that it shuts down in 1s?
> 
> Joel Baranick @kadaan Jan 20 10:46
> What about asserting a longer shutdown time, but verifying that the Task pulled at most one more record after being told to shutdown.
>  long indexBeforeShutdown = continuousExtractor.index;
> task.shutdown();
> long shutdownStartTime = System.currentTimeMillis();
> log.info("Shutting down task now");
> boolean success = task.awaitShutdown(10000);
> long shutdownEndTime = System.currentTimeMillis();
> long indexAfterShutdown = continuousExtractor.index;
> Assert.assertTrue(success, "Task should shutdown in 10 seconds");
> Assert.assertTrue(indexAfterShutdown - indexBeforeShutdown <= 1, "Task should have pulled at most one more record after calling shutdown");
> log.info("Task done waiting to shutdown {} after {}ms", success, shutdownEndTime - shutdownStartTime);
> 
> Shirshanka Das @shirshanka Jan 20 12:00
> hmm… ok I think I know what might be going on
> in the case of an orderly shutdown.. I was sending down a shutdown record
> and the fork was using that to shut itself down quickly
> it is possible that your change for sync fork / async fork might have changed that behavior slightly
> I haven’t looked at your PR yet… let me see what happens in the Task.shutdown code path now
> 
> Joel Baranick @kadaan Jan 20 12:37
> It is possible that it has changed, but from my understanding the async fork should perform exactly like the previous code. If that is not that case, it is a bug.
> If feels to me that the 1s time limit is too tight, as the streaming code blocks for a second while reading, right?
> That would explain why it was shutting down in 1007ms.
> Like:
> 2017-01-20 10:48:13 PST INFO  [main] gobblin.runtime.TaskContinuousTest 227 - Task done waiting to shutdown true after 1008ms
> 
> Shirshanka Das @shirshanka Jan 20 15:37
> ok, let me check whether the master build also fails intermittently due to this

kadaan:
@shirshanka Ping

PR #1553, opened 2017-01-17
Use `FileSystem.copyFromLocalFile` rather than `FileUtil.copy` when the source filesystem is local.

kadaan:


kadaan:

PR #1562, opened 2017-01-18
Ensure that files in the tasks being retried are included in the previous snapshot

kadaan:


kadaan:

kadaan:
@chavdar Can you please review this?

chavdar:
It's probably best for @ibuenros to review. I am not sure what ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT is .

kadaan:
`ConfigurationKeys.SOURCE_FILEBASED_FS_SNAPSHOT` is the list of files seen at a given time on the source.  Basically, when a job runs it gets a list of all files on the source.  This becomes the `SOURCE_FILEBASED_FS_SNAPSHOT` for this run.  Then it checks for a previous run.  If one exists, it looks for the previous run's `SOURCE_FILEBASED_FS_SNAPSHOT` and diffs it with the current one.  Any files that are in the current one and not in the previous are files which will be ingested.

There are two changes here:

1. Ensure that when we log the `SOURCE_FILEBASED_FILES_TO_PULL` that we don't produce an exceedingly large string.  This is done by limiting the number of files we will print to 2000.
1. Ensure that the files to pull from workunits we are retrying are in the previous snapshot

kadaan:
@ibuenros Can you please review this?

ibuenros:
LGTM, will close and re-open to re-run tests.

kadaan:
@ibuenros ty!  Can't wait till we don't get random test failures!!!

PR #1564, opened 2017-01-19
Cleanup of build.gradle

ibuenros:
This PR is a clean up of build.gradle. It is mostly moving the various parts of build.gradle into separate script files, and calling them in the correct order.

Additionally, created a skeleton for restli projects.


gobblin-restli/api.gradle
```diff
@@ -0,0 +1,50 @@
...
+  dataModel externalDependency.pegasus.restliCommon
+
+  compile externalDependency.pegasus.data
+  compile externalDependency.pegasus.restliClient
+  compile externalDependency.pegasus.restliCommon
+  compile externalDependency.pegasus.pegasusCommon
+}
+
+artifacts {
+  //archives mainRestClientJar
```
htran1: Should be uncommented? I see it there in the examples.

gobblin-restli/client.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+apply plugin: 'pegasus'
+apply plugin: 'java'
+
+archivesBaseName = project.getParent().getName() + "-client"
+buildDir = rootProject.buildDir.toString() + "/" + archivesBaseName
+
+dependencies {
+  compile project(path: project.getParent().getPath() + ":api", configuration: 'dataTemplate')
+  compile project(path: project.getParent().getPath() + ':api', configuration: 'restClient')
```
ibuenros: The idea is we'll add them as needed.

gobblin-restli/api.gradle
ibuenros: Talked offline. This is a harder issue that we will address later.

PR #1567, opened 2017-01-22
Gobblin compaction bug fixes

pcadabam:
- Use `Set` of inputPaths instead of `List` because paths are meant to be unique and order of paths does not matter
- Use `CopyOnWriteArraySet` for inputPaths to prevent concurrent modifications

@htran1 can you review?


pcadabam:
@htran1, made changes discussed offline. Please take another look

PR #1569, opened 2017-01-23
Fix bug, can not add elements to immutable list

rsimiciuc:


ibuenros:
@rsimiciuc can you explain how this fixes the bug? It seems we're no longer trying to add additional input paths?

pcadabam:
The bug has been fixed in https://github.com/linkedin/gobblin/pull/1567

rsimiciuc:
getInputPaths does it. Bug fixed anyway, thanks

PR #1574, opened 2017-01-24
Close kafkaCounsumer in KafkaSource and KafkaExtractor

pcadabam:
@ydai1124 can you review?

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -42,7 +42,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
-import com.google.common.io.Closer;
 
 import gobblin.configuration.ConfigurationKeys;
 import gobblin.configuration.SourceState;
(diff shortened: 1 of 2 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -117,52 +114,58 @@
...
-        state.getPropAsInt(ConfigurationKeys.MR_JOB_MAX_MAPPERS_KEY, ConfigurationKeys.DEFAULT_MR_JOB_MAX_MAPPERS);
-    return KafkaWorkUnitPacker.getInstance(this, state).pack(workUnits, numOfMultiWorkunits);
+      int numOfMultiWorkunits =
+          state.getPropAsInt(ConfigurationKeys.MR_JOB_MAX_MAPPERS_KEY, ConfigurationKeys.DEFAULT_MR_JOB_MAX_MAPPERS);
+      return KafkaWorkUnitPacker.getInstance(this, state).pack(workUnits, numOfMultiWorkunits);
+    } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+      throw new RuntimeException(e);
+    } finally {
+      try {
+        this.kafkaConsumerClient.close();
```
ydai1124: I think it is better to retain the original `closer` to register the closable kafkaConsumerClient and then close it.

pcadabam: `Source.shutdown()` is called when the driver finishers. We don't need to keep kafkaConsumer open till then. The client can be closed as soon as we finish building workunits

PR #1575, opened 2017-01-24
Implementation of Limiter using a remote Rest.li throttling service t…

ibuenros:
…o acquire permits.


ibuenros:
@htran1 Can yo review?

PR #1576, opened 2017-01-24
CopyableDatasetRequestor will catch any Dataset exception and skip (p…

ibuenros:
…reviously would only catch IOExceptions causing copy failure).


ibuenros:
@ydai1124 can you review?

PR #1580, opened 2017-01-25
Fix gradle

ibuenros:


ibuenros:
Trivial gradle changes. Merging.

PR #1581, opened 2017-01-25
Fix NPE in KafkaSource

pcadabam:


pcadabam:
Trivial fix self merging

PR #1582, opened 2017-01-25
Added pull file example on JDBC writer

jinhyukchang:
Added pull file example on JDBC writer


jinhyukchang:
@ibuenros  Could you take a look?

PR #1583, opened 2017-01-25
Making AsynchronousFork honor shutdown record

shirshanka:
This is a short-term fix to ensure that tasks shutdown right away. 
Will do a more thorough refactoring later.


shirshanka:
Tested that the unit-test succeeds immediately.

shirshanka:
Merged in since the change was trivial.

PR #1586, opened 2017-01-27
KafkaExtractor should support schema evolution

lbendig:
A job using
```
source.class=gobblin.source.extractor.extract.kafka.KafkaDeserializerSource
kafka.deserializer.type=CONFLUENT_AVRO
writer.destination.type=HDFS
writer.output.format=AVRO
writer.builder.class=gobblin.writer.AvroDataWriterBuilder
```

can't extract data when backward compatible changes have been made in the schema. 
It fails with the exception as follows:

```
2017-01-27 19:49:42,398 INFO [ForkExecutor-0] gobblin.runtime.fork.Fork-0: Wrapping writer gobblin.writer.PartitionedDataWriter@5f3d5169
2017-01-27 19:49:42,497 WARN [ForkExecutor-0] gobblin.writer.RetryWriter: Caught exception. This may be retried.
org.apache.avro.file.DataFileWriter$AppendWriteException: java.lang.ArrayIndexOutOfBoundsException: 2
  at org.apache.avro.file.DataFileWriter.append(DataFileWriter.java:308)
  at gobblin.writer.AvroHdfsDataWriter.write(AvroHdfsDataWriter.java:84)
  at gobblin.writer.AvroHdfsDataWriter.write(AvroHdfsDataWriter.java:51)
  at gobblin.instrumented.writer.InstrumentedDataWriterDecorator.writeImpl(InstrumentedDataWriterDecorator.java:76)
  at gobblin.instrumented.writer.InstrumentedDataWriterBase.write(InstrumentedDataWriterBase.java:165)
  at gobblin.instrumented.writer.InstrumentedDataWriterDecorator.write(InstrumentedDataWriterDecorator.java:70)
  at gobblin.writer.PartitionedDataWriter.write(PartitionedDataWriter.java:127)

Caused by: java.lang.ArrayIndexOutOfBoundsException: 2
  at org.apache.avro.generic.GenericData$Record.get(GenericData.java:211)
  at org.apache.avro.generic.GenericData.getField(GenericData.java:666)
  at org.apache.avro.generic.GenericData.getField(GenericData.java:681)
  at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:151)
  at org.apache.avro.generic.GenericDatumWriter.writeRecord(GenericDatumWriter.java:143)
```

The problem is that the record was read with its own schema, however when it's being written out in `AvroHdfsDataWriter#write`, the `DataFileWriter` uses the most recent schema, taken from
`KafkaDeserializerExtractor#getSchema`. This schema mismatch causes the exception.

Solution: 
When we assume that schema changes in the dataset are backward compatible, we can always use the most recent schema (taken from the schema registry) as a reader schema during deserialization.
Confluent's Schema registry 2.0.1 doesn't support reader schemas (see `io.confluent.kafka.serializers.KafkaAvroDeserializer`), this has been added to 3.0.1 however, `KafkaDeserializerExtractor#decodeRecord` always calls the default deserializer method without passing the reader schema.

To support reader schemas regardless of the _confluentVersion_ being used, I subclassed 
`KafkaAvroDeserializer` and ported the relevant codes from the Confluent implementation.

According to the tests with different Schema Registry versions this works fine, however I'm not quite sure  whether there exists a simpler solution which doesn't require codes ported from 3rd party dependencies.


lbendig:
@sahilTakiar @ibuenros Can you please review?

lbendig:
Could anyone please take a look at this issue? It looks like that after a schema change the corresponding Kafka topic cannot be ingested anymore.

ibuenros:
@lbendig what kind of schema change caused this issue? Any chance you can provide the old and new schema?

lbendig:
@ibuenros I added a full test example which leads to this error:
[error_reproduce.zip](https://github.com/linkedin/gobblin/files/759407/error_reproduce.zip)

shirshanka:
@lbendig : we'll get this resolved this week.

lbendig:
@shirshanka Thank you!

shirshanka:
@lbendig : Here is a simpler approach to accomplish this. I looked at KafkaAvroExtractor and followed the pattern from there. 

First in KafkaDeserializerExtractor: 
 cache the latest schema for the topic in a private member variable (say this.latestSchema) using the same strategy as KafkaAvroExtractor. 

Then in KafkaDeserializerExtractor::decodeRecord

First get the native deserialized object from the deserializer
Object deserialized = this.kafkaDeserializer.deserialize(this.topicName, messageAndOffset.getMessageBytes());

Then coerce this object to be compatible with the latest schema using gobblin.util.AvroUtils
You might have to do an instanceof GenericRecord check before this.

GenericRecord upconvertedRecord = gobblin.util.AvroUtils.convertRecordSchema((GenericRecord) deserialized, this.latestSchema)
return upconvertedRecord;


This should result in a much smaller code patch and hopefully accomplishes the same results.

shirshanka:
@lbendig : Thoughts on the above? 
Do you think you will have cycles to try out the approach I'm proposing?

lbendig:
@shirshanka Thanks for your suggestion, I think it should yields to the same result but in a easier way. I'll implement it this week.

lbendig:
@shirshanka Sorry for the delay, I'll get back to this issue this week.

lbendig:
Hi @shirshanka , I've changed the implementation to your suggested logic. Tested both when reading to / writing from Kafka.

lbendig:
@shirshanka Thanks for the review.

PR #1595, opened 2017-01-31
Upgrade to latest wix-embedded-mysql so that build continues to work.

kadaan:
Fixes problem where the build fails because mysql cannot be downloaded as the URL has changed.


kadaan:
@chavdar Can you assign this to the correct reviewer?

PR #1599, opened 2017-02-06
Support HA of standalone cluster manager by supporting multiple helix…

htran1:
… controllers. The leader controller is the active instance and the other controllers are idle standbys.

### DIFF
--- a/gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterConfigurationKeys.java
+++ b/gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterConfigurationKeys.java
@@ -33,12 +33,16 @@ public class GobblinClusterConfigurationKeys {
   // General Gobblin Cluster application configuration properties.
   public static final String APPLICATION_NAME_OPTION_NAME = "app_name";
   public static final String STANDALONE_CLUSTER_MODE = "standalone_cluster";
+  public static final String STANDALONE_CLUSTER_MODE_KEY = GOBBLIN_CLUSTER_PREFIX + "standaloneMode";
+  public static final boolean DEFAULT_STANDALONE_CLUSTER_MODE = false;
+
 
   // Helix configuration properties.
   public static final String HELIX_CLUSTER_NAME_KEY = GOBBLIN_CLUSTER_PREFIX + "helix.cluster.name";
   public static final String ZK_CONNECTION_STRING_KEY = GOBBLIN_CLUSTER_PREFIX + "zk.connection.string";
   public static final String WORK_UNIT_FILE_PATH = GOBBLIN_CLUSTER_PREFIX + "work.unit.file.path";
   public static final String HELIX_INSTANCE_NAME_OPTION_NAME = "helix_instance_name";
+  public static final String HELIX_INSTANCE_NAME_KEY = GOBBLIN_CLUSTER_PREFIX + "helixInstanceName";
   // The number of tasks that can be running concurrently in the same worker process
   public static final String HELIX_CLUSTER_TASK_CONCURRENCY = GOBBLIN_CLUSTER_PREFIX + "helix.taskConcurrency";
   public static final int HELIX_CLUSTER_TASK_CONCURRENCY_DEFAULT = 40;
--- a/gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
+++ b/gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
@@ -17,6 +17,7 @@
 
 package gobblin.cluster;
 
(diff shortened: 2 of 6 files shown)


htran1:
@abti, can you review?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterConfigurationKeys.java
```diff
@@ -33,12 +33,16 @@
   // General Gobblin Cluster application configuration properties.
   public static final String APPLICATION_NAME_OPTION_NAME = "app_name";
   public static final String STANDALONE_CLUSTER_MODE = "standalone_cluster";
+  public static final String STANDALONE_CLUSTER_MODE_KEY = GOBBLIN_CLUSTER_PREFIX + "standaloneMode";
```
abti: We already have a config that is set for standalone mode, lets converge them in one. Either use this or that (we can make this choice because its not yet eligible for backward compatibility)

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -149,15 +167,109 @@ public GobblinClusterManager(String clusterName, String applicationId, Config co
   }
 
   /**
+   * Handle leadership change.
+   * The applicationLauncher is only started on the leader.
+   * The leader cleans up existing jobs before starting the applicationLauncher.
+   * @param changeContext notification context
+   */
+  private void handleLeadershipChange(NotificationContext changeContext) {
+    if (this.helixManager.isLeader()) {
```
abti: Should we ensure that the current leader has stopped before starting services here?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -149,15 +167,109 @@ public GobblinClusterManager(String clusterName, String applicationId, Config co
...
+
+        for (String r : resourceNames) {
+          LOGGER.info("resource {} found", r);
+
+          // try to infer queue and job names from the resource name.
+          // the resource name is in <queue name>_<jobid> format where the job id may start with _job or _
+          int separatorIndex = r.indexOf("_job");
+
+          if (separatorIndex == -1) {
+            separatorIndex = r.indexOf('_');
```
abti: use lastIndexOf(), there could be now or in future _ in names

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -149,15 +167,109 @@ public GobblinClusterManager(String clusterName, String applicationId, Config co
...
+          }
+
+          String queue = r.substring(0, separatorIndex);
```
abti: nit: Apache string utils might be helpful here

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -149,15 +167,109 @@ public GobblinClusterManager(String clusterName, String applicationId, Config co
...
+
+        this.applicationLauncher.start();
+        isLeader = true;
+      }
+    } else {
+      // stop application launcher if lost leadership role
+      if (isLeader) {
+        isLeader = false;
+        try {
+          this.applicationLauncher.stop();
```
abti: is this call blocking?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterConfigurationKeys.java
htran1: Do you mean the 'standalone_cluster' config? That is a command line argument.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
htran1: That is in another process/node, so we can't check that. Helix has taken leadership away though, so that process should start shutting down when it receives the notification.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
htran1: lastIndexOf() also won't be reliable since the job name is repeated. What we need is to compare against the queue name, but that is not available until we upgrade to the newer helix.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinClusterManager.java
htran1: Yes, this call will block until the timeout.

PR #1600, opened 2017-02-06
Ignore single Json record if size is larger than 256KB which is size limit for eventhub

yukuai518:
Tolerance is chosen to be 95% so the internal size limit will be (256K * 0.95)

### DIFF
--- a/gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatch.java
+++ b/gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatch.java
@@ -39,7 +39,7 @@ public class EventhubBatch extends Batch<JsonObject>{
   private final long creationTimestamp;
   private final long memSizeLimit;
   private final long ttlInMilliSeconds;
-  public static final int overheadSizeInByte = 15;
+  public static final int OVERHEAD_SIZE_IN_BYTES = 15;
 
   public EventhubBatch (long memSizeLimit, long ttlInMilliSeconds) {
     this.creationTimestamp = System.currentTimeMillis();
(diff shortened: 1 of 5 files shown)


gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatch.java
```diff
@@ -39,6 +39,7 @@
   private final long creationTimestamp;
   private final long memSizeLimit;
   private final long ttlInMilliSeconds;
+  public static final int overheadSizeInByte = 15;
```
ibuenros: `static final` variables should have names in all caps `OVERHEAD_SIZE_IN_BYTES`

gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatchAccumulator.java
```diff
@@ -85,17 +93,27 @@ public long getExpireInMilliSecond () {
...
-      EventhubBatch batch = new EventhubBatch(this.batchSizeLimit, this.expireInMilliSecond);
+      EventhubBatch batch = new EventhubBatch(this.memSizeLimit, this.expireInMilliSecond);
       LOG.info ("Batch " + batch.getId() + " is generated");
       Future<RecordMetadata> future = batch.tryAppend(record, callback);
+
+      // Even single record can exceed the size limit from one batch
+      // Ignore the record because Eventhub can only accept payload less than 256KB
+      if (future == null) {
+        LOG.warn ("Batch " + batch.getId() + " is marked as complete because it contains a huge record: " + record.toString());
+        future = new RecordFuture(new CountDownLatch(0), 0);
```
ibuenros: You can use an `ImmediateFailedFuture` here (or `ImmeduateSuccessfulFuture`)

yukuai518: Done

gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatch.java
yukuai518: Done

yukuai518:
Test failed in gobblin-cluster module which is not related to this commit.

PR #1605, opened 2017-02-07
Integrating Azure Data Lake Filesystem

jinhyukchang:
- Added:
  Runtime dependency of Azure Data Lake Filesystem
- Changed:
  Pass through job property, with ability to decrypt, to Configuration to instantiate FileSystem.
  

Pull file snippet for Dist cp case:
from=/from/directory
to=/to/directory

writer.fs.uri=adl://mydatalakestore.azuredatalakestore.net/

dfs.adls.oauth2.access.token.provider.type=ClientCredential
dfs.adls.oauth2.refresh.url=https://[end point token retrieval]
dfs.adls.oauth2.client.id=[client ID]
writer.encrypted.dfs.adls.oauth2.credential=[encrypted ADL secret key]
encrypt.key.loc=/passphrase/file/path


jinhyukchang:
@ibuenros Can you take a look? Thanks.

gobblin-core/build.gradle
```diff
@@ -57,6 +57,7 @@ dependencies {
   compile externalDependency.hadoopHdfs
 
   runtime externalDependency.protobuf
+  runtime externalDependency.hadoopAdl
```
ibuenros: Why is this needed?

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -56,7 +56,7 @@
 import gobblin.publisher.DataPublisher;
 import gobblin.publisher.UnpublishedHandling;
 import gobblin.util.HadoopUtils;
-
+import gobblin.util.WriterUtils;
```
ibuenros: Add empty line after this one

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -68,7 +69,6 @@
 import gobblin.util.limiter.Limiter;
 import gobblin.util.limiter.broker.SharedLimiterFactory;
 import gobblin.writer.DataWriter;
-
```
ibuenros: This line should not be removed

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -38,6 +38,7 @@
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
 import com.google.common.collect.Iterators;
+import com.google.common.collect.Lists;
```
ibuenros: Is this needed?

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -462,4 +464,34 @@ public static boolean verifySubset(Config superConfig, Config subConfig) {
...
+   * e.g:
+   *  encConfigPath = enc.conf
+   *  - Before : { enc.conf.secret_key : ENC(rOF43721f0pZqAXg#63a) }
+   *  - After  : { secret_key : decrypted_val }
+   *
+   * @param config
+   * @param encConfigPath
+   * @return
+   */
+  public static Config resolveEncrypted(Config config, Optional<String> encConfigPath) {
```
ibuenros: What's the point of optional `encConfigPath` if we'll just return `config` if it is not present?

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
```diff
@@ -462,4 +464,34 @@ public static boolean verifySubset(Config superConfig, Config subConfig) {
...
+    if (!encConfigPath.isPresent() || !config.hasPath(encConfigPath.get())) {
+      return config;
+    }
+
+    Config encryptedConfig = config.getConfig(encConfigPath.get());
+
+    PasswordManager passwordManager = PasswordManager.getInstance(configToProperties(config));
+    Map<String, String> tmpMap = Maps.newHashMap();
+    for (Map.Entry<String, ConfigValue> entry : encryptedConfig.entrySet()) {
```
ibuenros: If the config is deep (e.g. `enc.conf.my.secret.key`) `encryptedConfig.entrySet()` will return a weird object for key `my`. What you want to do here instead is `ConfigUtils.configToProperties(encryptedConfig).entrySet()`.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -110,6 +113,7 @@ public static Configuration newConfiguration() {
     // Add a new custom filesystem mapping
     conf.set("fs.sftp.impl", "gobblin.source.extractor.extract.sftp.SftpLightWeightFileSystem");
     conf.set("fs.sftp.impl.disable.cache", "true");
+    conf.set("fs.adl.impl", "org.apache.hadoop.fs.adl.AdlFileSystem");
```
ibuenros: This should not be needed. `FileSystem` uses service providers, so discovery happens automatically.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -656,9 +660,37 @@ public static void safeRenameRecursively(FileSystem fileSystem, Path from, Path
...
+   *
+   * Common use case for these encryptedPath:
+   *   When there's have encrypted credential in job property but you'd like Filesystem to get decrypted value.
+   *
+   * @param srcConfig source config.
+   * @param encryptedPath Optional. If provided, config that is on this path will be decrypted. @see ConfigUtils.resolveEncrypted
+   *                      Note that config on encryptedPath will be included in the end result even it's not part of includeOnlyPath
+   * @return Hadoop Configuration.
+   */
+  public static Configuration getConfFromState(State state, Optional<String> encryptedPath) {
```
ibuenros: Should we instead have a static path for encrypted keys used in Hadoop configurations?

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
```diff
@@ -78,7 +78,7 @@ public void setProxiedFileSystem(FileSystem currentProxiedFs) {
    * @throws URISyntaxException
    * @return proxiedFs
    */
-  public FileSystem getProxiedFileSystem(State properties, AuthType authType, String authPath, String uri)
+  public FileSystem getProxiedFileSystem(State properties, AuthType authType, String authPath, String uri, final Configuration conf)
```
ibuenros: Please add a backwards compatible method (it can use this one internally).

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
jinhyukchang: Will do.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
jinhyukchang: Will remove it.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
jinhyukchang: Will do.

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
jinhyukchang: I think it's just two different pattern for caller:
 a) config = resolveEncrypted(config, Optional.fromNullable(val))
     or config = resolveEncrypted(config, Optional.absent())
 b) if (val != null) { config = resolveEncrypted(config); }

I chose a, but don't have strong preference. Thoughts?

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
jinhyukchang: When I tested out, encryptedConfig.entrySet() provided a full path on the key. Let me update the test case to test nested key as well.

gobblin-utility/src/main/java/gobblin/util/ProxiedFileSystemWrapper.java
jinhyukchang: Will do.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
jinhyukchang: You mean to have static path for Hadoop Config instead of getting it from caller?
The reason I've delegated this prefix out to the caller instead of having a static path in HadoopUtils is for the case of naming collision.
For example, if we need different credential for source Filesystem and target Filesystem, it would be hard to disambiguate using static path.
e.g:
File system accepts property "credential"
prefix for Source: src.encrypted.conf
prefix for Writer: writer.encrypted.conf

gobblin-core/build.gradle
jinhyukchang: This is to include two jars to use ADL FileSystem in runtime. Do we want to place it in different process?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
jinhyukchang: Yes, it supposed to pick it up in runtime without this configuration, but it's not working with AdlFileSystem. I looked at their jar and they seemed misplaced service information under "META-INF/org.apache.hadoop.fs.FileSystem", not in "META-INF/services/org.apache.hadoop.fs.FileSystem"

gobblin-core/build.gradle
ibuenros: Yes, this should not be in `gobblin-core`. You could create a module for ADL, but I expect it would be almost empty, otherwise, just make documentation specifying the artifacts that need to be added for running ADL pipeline.

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
ibuenros: Sure, just make it clear that passing `Optional.absent()` will not change anything, as opposed to attempting to decrypt everything.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
ibuenros: I see, if you add an ADL module you could instead use the service provider pattern with your own resource. Have you considered submitting a PR to their project fixing this?

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
ibuenros: Sounds good

gobblin-core/build.gradle
jinhyukchang: With the comment below regarding service provider, I think it's better to add a module for ADL. Will remove this line and add a module for ADL.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
jinhyukchang: Will add a ADL module as suggested. I am having trouble finding their github page. It's not part of hadoop project. Let me either do the PR or create an issue on this.

gobblin-utility/src/main/java/gobblin/util/ConfigUtils.java
jinhyukchang: Sure, will mention it in Java doc.

jinhyukchang:
@ibuenros Just addressed PR comments. Could you take a look when you have a chance? Thanks!

gobblin-distribution/build.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
   compile project(':gobblin-yarn')
   compile project(':gobblin-aws')
   compile project(':gobblin-runtime-hadoop')
+  compile project(':gobblin-azure-datalake')
```
ibuenros: This should not be part of gobblin-distribution. Instead, add it to `gobblin-distribution/gobblin-flavor-full.gradle`

jinhyukchang: Will move it and will take a look at test failures.

PR #1606, opened 2017-02-07
Combining all the sequential test sources into a single one

shirshanka:
Merged the different Sequential Test sources into one. Created a utility json converter.


shirshanka:
Trivial change.. self-approving.

PR #1607, opened 2017-02-07
Encryption skeleton: capabilities and encoders

eogren:
@shirshanka @ibuenros 

Sorry this so long - wanted to implement the concepts in a few places to make sure the abstractions didn't immediately break down, but that ends up with there being a lot of lines changed.

 2 main pieces in this commit:

- Support of capability checking and verification. Two design goals here:
  - If a job is configured with encryption and the writer doesn't support it, we must blow up
  - There a bunch of cases where we want to query objects for their capability. Annotations aren't necessarily sufficient enough because there may be other pieces of config attached - eg encryption with AES vs something else. Methods seem the right way to go. 
 - Wrap StreamEncoders around FileStreamBasedWriters, FsDataWriterBuilders, and KafkaWriters to support encryption. SimpleEncryptor is a proof of concept.

Other details:

- Since only Writers support capability checks right now, the Fork is responsible for verifying that the writer can be properly built. I could eventually see this being refactored out into a broader object that can validate all objects in the config. (Eg catch errors where a Writer expects a GenericRecord type but will always be passed a String because of the converter chain)
- WriterBuilders are asked if they know how to build a writer with the correct capability. Right now that information tends to be hardcoded in the builders, but you could see this eventually going one level further and having builders query info on the writers they know about. Not implemented here because it would add more complexity, and all builders only instantiate 1 type of writer right now anyways
- Refactor the FsWriters a little bit and introduce a FsStreamBasedWriter middle class - the Hive Output format writer uses Hadoop OutputFormat and RecordWriters so cannot support encryption without additional work
- Adding encryption to the Kafka writers means Gobblin serializes the incoming record to byte[] itself; it will manually call the ValueSerializer specified in the kafka-writer config vs letting Kafka do it for us.


gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+  private KeyRecord currentKey;
+  private List<KeyRecord> keyRecords;
+
+  /**
+   * Create a new encryptor
+   * @param credentialStore Credential store where keys can be found
+   */
+  public RotatingAESEncryptor(CredentialStore credentialStore) {
+    this.credentialStore = credentialStore;
+    this.random = new SecureRandom();
```
ibuenros: No need to do `SecureRandom` here. The value chosen is not sensitive (as it is written in plaintext either way), and it will just slow down record processing.

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Implementation of an encryption algorithm that works in the following way:
+ *
+ * 1. A credentialStore is provisioned with a set of AES keys
+ * 2. When wrapOutputStream() is called, an AES key will be picked at random
+ * 3. A header will be written [keyId][ivLength][base64 encoded iv]
+ * 4. Ciphertext will be base64 encoded and written out
+ */
+public class RotatingAESEncryptor implements StreamEncoder {
```
ibuenros: This class is not thread safe and I don't see any reason why it couldn't be. Please make it thread safe, or if it's actually very hard, make it very clear it isn't.

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+import javax.crypto.spec.SecretKeySpec;
+import javax.xml.bind.DatatypeConverter;
+
+
+/**
+ * Implementation of an encryption algorithm that works in the following way:
+ *
+ * 1. A credentialStore is provisioned with a set of AES keys
+ * 2. When wrapOutputStream() is called, an AES key will be picked at random
+ * 3. A header will be written [keyId][ivLength][base64 encoded iv]
```
ibuenros: Please explain what iv is

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+      throw new IllegalStateException("Can't rekey stream before wrapOutputStream() has been called!");
+    }
+
+    try {
+      currentKey = getNewKey();
+
+      cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
+      cipher.init(Cipher.ENCRYPT_MODE, currentKey.getSecretKey());
+      byte[] iv = cipher.getIV();
+      base64Iv = DatatypeConverter.printBase64Binary(iv);
```
ibuenros: Why not use `java.util.Base64` instead?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+    this.random = new SecureRandom();
+    this.headerWritten = false;
+  }
+
+  @Override
+  public OutputStream wrapOutputStream(OutputStream origStream) {
+    this.origStream = origStream;
+
+    rekey();
+    final Base64OutputStream base64OutputStream = new Base64OutputStream(origStream);
```
ibuenros: Why is this needed?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+    final CipherOutputStream encryptedStream = new CipherOutputStream(base64OutputStream, cipher);
```
ibuenros: A different way of doing the same (I think) is to do
```java
writeHeader(origStream);
return new CipherOutputStream(base64OutputStream, cipher);
```

gobblin-core-base/src/main/java/gobblin/crypto/SimpleEncryptor.java
```diff
@@ -0,0 +1,54 @@
...
+
+import gobblin.configuration.State;
+import gobblin.writer.StreamEncoder;
+
+
+/**
+ * Simple encryption algorithm that just increments every byte sent
+ * through it by 1. Useful for unit tests or proof of concept, but is not actually secure.
+ */
+public class SimpleEncryptor implements StreamEncoder {
```
ibuenros: Why not call it `InsecureShiftEncryptor` to make it very clear what it is.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -62,7 +61,7 @@
   protected final int branchId;
   protected final String fileName;
   protected final FileSystem fs;
-  protected final Path stagingFile;
+  protected Path stagingFile;
```
ibuenros: Why this change?

gobblin-example/src/main/java/gobblin/example/wikipedia/EncryptedEmbeddedWikipediaExample.java
```diff
@@ -0,0 +1,103 @@
...
+import gobblin.writer.WriterOutputFormat;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import org.apache.commons.cli.CommandLine;
+
+
+/**
+ * Embedded Gobblin to run Wikipedia example.
+ */
+public class EncryptedEmbeddedWikipediaExample extends EmbeddedGobblin {
```
ibuenros: I think it's better to use the same wikipedia example with an encryption option.

gobblin-modules/gobblin-kafka-08/src/main/java/gobblin/kafka/writer/Kafka08DataWriter.java
```diff
@@ -74,30 +78,33 @@ public long bytesWritten() {
...
-  }
 
-  public Kafka08DataWriter(Producer producer, Config config)
-  {
-    this.topic = config.getString(KafkaWriterConfigurationKeys.KAFKA_TOPIC);
-    this.producer = producer;
+  @SuppressWarnings("unchecked")
+  public Kafka08DataWriter(Properties props, List<StreamEncoder> encoders) {
+    this.producer = getKafkaProducer(props);
+    this.valueSerializer = (Serializer<D>)KafkaWriterHelper.getValueSerializer(Serializer.class, props);
```
ibuenros: Please check with @shirshanka, I believe he wanted to get rid of explicit serializers.

gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/kafka/writer/FailureFuture.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+
+/**
+ * Future that always throws an exception when queried
+ * @param <T> Type of variable that would be returned
+ */
+public class FailureFuture<T> implements Future<T> {
```
ibuenros: There is a guava method for this: `Futures.immediateFailedFuture(Throwable)`

gobblin-api/src/main/java/gobblin/writer/StreamEncoder.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.OutputStream;
+
+import gobblin.annotation.Alpha;
+
+
+/**
+ * Interface for an object that can encrypt a bytestream.
+ */
+@Alpha
+public interface StreamEncoder {
```
eogren: I actually have a separate interface StreamDecoder sitting unchecked in - figure might as well allow them to be separate concepts

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
eogren: It's java 8 only, and core still have to support 7 right?

gobblin-core-base/src/main/java/gobblin/crypto/KeystoreCredentialStore.java
```diff
@@ -0,0 +1,134 @@
...
+  enum CreationOptions {
+    NONE, CREATE_IF_MISSING
+  };
+
+  public KeystoreCredentialStore(String path, String passwordStr) throws IOException {
+    this(path, passwordStr, EnumSet.noneOf(CreationOptions.class));
+  }
+
+  public KeystoreCredentialStore(String path, String passwordStr, EnumSet<CreationOptions> options) throws IOException {
+    this(FileSystem.get(new Configuration()), new Path(path), passwordStr, options);
```
eogren: In this case path is a String so can't do much - should I change the method below to only take a Path and extract the filesystem from it though?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
eogren: this was part of the encryption spec - all output should be ascii so we need to base64 encode the ciphertext

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
eogren: I originally wrote it that way - I'm not sure if I'm being overly paranoid, but I figured it is possible that  there will be cases later when the full set of OutputStreams have not been chained together when wrapOutputStream() is called. 

Not writing the header until the first write() call is done means we aren't actually doing I/O until a specific write is requested which I think is the safer way to go. Otherwise it's not super intuitive that calling wrapOutputStream(os) may actually end up doing I/O behind the scenes

gobblin-core-base/src/main/java/gobblin/crypto/SimpleEncryptor.java
eogren: ah much better name - thanks

gobblin-api/src/main/java/gobblin/writer/StreamEncoder.java
ibuenros: The advantage of having them together is that it is very obvious, for each encoding, how to decode it. Even if the names are similar, I could see someone spending too much time figuring that out.

gobblin-api/src/main/java/gobblin/writer/StreamEncoder.java
```diff
@@ -0,0 +1,25 @@
...
+   * encrypts the bytes flowing through.
+   * @param origStream Stream to wrap
+   * @return A wrapped stream for encryption
+   */
+  OutputStream wrapOutputStream(OutputStream origStream);
+
+  /**
+   * Get tag/file extension associated with encoder
+   */
+  String getTag();
```
ibuenros: Is this intended to be used as extension for the file? If so, can you specify that in the javadoc?

gobblin-core-base/src/main/java/gobblin/crypto/KeystoreCredentialStore.java
ibuenros: you can also do `new Path(path).getFileSystem(...)`. The advantage of doing this is that the user can specify a file in a different file system.

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
ibuenros: sounds good

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
ibuenros: That makes sense.

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
ibuenros: ok, wasn't aware of that

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESEncryptor.java
```diff
@@ -0,0 +1,168 @@
...
+ * 2. When wrapOutputStream() is called, an AES key will be picked at random
+ * 3. A header will be written [keyId][ivLength][base64 encoded iv]
+ * 4. Ciphertext will be base64 encoded and written out
+ */
+public class RotatingAESEncryptor implements StreamEncoder {
+  private static final int AES_KEY_LEN = 16;
+
+  private final SecureRandom random;
+  private final CredentialStore credentialStore;
+  private boolean headerWritten;
```
ibuenros: This needs to be `volatile` (also `keyRecords`)

ibuenros:
@eogren what is the status of this PR?

abti:
https://issues.apache.org/jira/browse/GOBBLIN-51

Please update your PR title with following prefix: [GOBBLIN-51]

PR #1608, opened 2017-02-08
Adding record with metadata

shirshanka:
and a few converters.


shirshanka:
Self-approving, trivial change.

PR #1611, opened 2017-02-09
Fixing tests and findbugs issues

shirshanka:


shirshanka:
Self-approving trivial change.

PR #1613, opened 2017-02-09
Helix upgrade to 0.6.7.

htran1:
Created customized Helix TaskDriver, MessagingService, and CriteriaEvaluator classes to workaround bug, missing functionality, and behavior changes.
Issues that we had to workaround.
1. Messaging service does not allow sending messages to instances
2. Deletion of job is not allowed on a running queue
3. Cannot stop a queue
4. Shutdown message is not defined
5. Creation of queue fails with generic exception if queue already exists


htran1:
@sdas, @abti, please take a look.

htran1:
@kishoreg, please take a look.

gobblin-cluster/src/main/java/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -147,42 +141,6 @@ public GobblinHelixJobLauncher(Properties jobProps, final HelixManager helixMana
...
-               }
-               **/
-
-              // HACK to ensure that concurrent tasks per instance is set.
-              // TODO: Remove this when we upgrade to a fixed Helix version
-              final String taskResourceName = TaskUtil.getNamespacedJobName(jobContext.getJobName(), jobContext.getJobId());
-              int jobConcurrency = ConfigUtils.getInt(jobConfig, GobblinClusterConfigurationKeys.HELIX_CLUSTER_TASK_CONCURRENCY,
-                  GobblinClusterConfigurationKeys.HELIX_CLUSTER_TASK_CONCURRENCY_DEFAULT);
-              HelixConfigScope resourceScope =
-                  new HelixConfigScopeBuilder(HelixConfigScope.ConfigScopeProperty.RESOURCE).forCluster(clusterName).forResource(taskResourceName).build();
```
shirshanka: When we delete this code, do we know if concurrent jobs per instance setting will be honored?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinHelixMessagingService.java
```diff
@@ -0,0 +1,202 @@
...
+  @Override
+  public Map<InstanceType, List<Message>> generateMessage(final Criteria recipientCriteria,
+      final Message message) {
+    Map<InstanceType, List<Message>> messagesToSendMap = new HashMap<InstanceType, List<Message>>();
+    InstanceType instanceType = recipientCriteria.getRecipientInstanceType();
+
+    if (instanceType == InstanceType.CONTROLLER) {
+      List<Message> messages = generateMessagesForController(message);
+      messagesToSendMap.put(InstanceType.CONTROLLER, messages);
+      // _dataAccessor.setControllerProperty(PropertyType.MESSAGES,
```
shirshanka: remove commented code?

gobblin-cluster/src/main/java/gobblin/cluster/GobblinHelixJobLauncher.java
htran1: I checked and confirmed that it is now set to 100, so the call jobConfigBuilder.setNumConcurrentTasksPerInstance(100); is now working.
Here is the job config from zookeeper showing the value of ConcurrentTasksPerInstance.
"simpleFields":{ "Command":"GobblinTaskFactory" ,"ConcurrentTasksPerInstance":"100"

PR #1614, opened 2017-02-09
Change eventhub to accept byte array as input

yukuai518:


gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatchAccumulator.java
```diff
@@ -107,7 +107,8 @@ public long getExpireInMilliSecond () {
       // Even single record can exceed the size limit from one batch
       // Ignore the record because Eventhub can only accept payload less than 256KB
       if (future == null) {
-        LOG.warn ("Batch " + batch.getId() + " is marked as complete because it contains a huge record: " + record.toString());
+        LOG.warn ("Batch " + batch.getId() + " is marked as complete because it contains a huge record: "
+                + new String(record, Charsets.UTF_8));
```
ibuenros: This may fail (record may not be valid UTF_8). Please print the base64 instead.
Also, the record is huge, it should be written only in `debug` mode.

yukuai518: Fixed.

yukuai518:
Change eventhub datawriter to be byte array type as input format

PR #1615, opened 2017-02-09
PasswordManager to support encrypt.key.val

jinhyukchang:
Added a support on "encrypt.key.val" key will be fetched from state directly. This is to support a case
to relieve burden of placing key file when plain text crypto key in job property can be safely protected.


jinhyukchang:
@htran1 Could you take a look? Thanks!

PR #1616, opened 2017-02-09
Encryption support for distcp only

eogren:
@ibuenros - simpler version

  - Add concept of StreamCodec which can encode/decode streams flowing through them
  - InsecureShiftCodec which just increments everything going through it by 1
  - RotatingAESCodec with CredentialStore that picks a random AES session key from a CredentialStore and uses it for encryption
  - Update FileAwareInputStreamDataWriter to wrap the output stream with a StreamCodec if encryption is specified in config

Stripped out:
 - capability concept, encode/decode support for kafka + record-based hdfs writers

### DIFF
--- a/gobblin-api/src/main/java/gobblin/crypto/CredentialStore.java
+++ b/gobblin-api/src/main/java/gobblin/crypto/CredentialStore.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package gobblin.crypto;
+
+import java.util.Map;
+
+
+/**
+ * Interface for a simple CredentialStore that simply has a set of byte-encoded keys. Format
+ * of the underlying keys is left to the implementor.
+ */
+public interface CredentialStore {
+  /**
+   * Get the binary encoded key with the given key
+   * @param id Key to lookup
+   * @return null if the key does not exist; encoded-key if it does
+   */
+  byte[] getEncodedKey(String id);
+
+  /**
+   * List all binary encoded keys in the credential store
+   * @return all binary encoded keys in the credential store
+   */
+  Map<String, byte[]> getAllEncodedKeys();
+}
--- a/gobblin-api/src/main/java/gobblin/writer/StreamCodec.java
+++ b/gobblin-api/src/main/java/gobblin/writer/StreamCodec.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gobblin.writer;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import gobblin.annotation.Alpha;
+
+
+/**
+ * Interface for an object that can encrypt a bytestream.
+ */
+@Alpha
+public interface StreamCodec {
+  /**
+   * Wrap a bytestream and return a new stream that encodes the bytes written to it and writes
+   * the result into origStream
+   * @param origStream Stream to wrap
+   * @return A wrapped stream for encoding
+   */
+  OutputStream encodeOutputStream(OutputStream origStream) throws IOException;
+
+  /**
+   * Wrap a bytestream and return a new stream that will decode
+   * any bytes from origStream when read from.
+   * @param origStream Stream to wrap
+   * @return A wrapped stream for decoding
+   */
+  InputStream decodeInputStream(InputStream origStream) throws IOException;
+
+  /**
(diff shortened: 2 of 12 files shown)


gobblin-api/src/main/java/gobblin/crypto/CredentialStore.java
```diff
@@ -0,0 +1,39 @@
...
+   * @param id Key to lookup
+   * @return null if the key does not exist; encoded-key if it does
+   */
+  byte[] getEncodedKey(String id);
+
+  /**
+   * List all binary encoded keys in the credential store
+   * @return all binary encoded keys in the credential store
+   */
+  Map<String, byte[]> getAllEncodedKeys();
```
ibuenros: What is the use of this method? It seems the Credential store could have a very large number of keys.

gobblin-api/src/main/java/gobblin/writer/StreamCodec.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@Alpha
+public interface StreamCodec {
+  /**
+   * Wrap a bytestream and return a new stream that encodes the bytes written to it and writes
+   * the result into origStream
+   * @param origStream Stream to wrap
+   * @return A wrapped stream for encoding
+   */
+  OutputStream wrapOutputStream(OutputStream origStream) throws IOException;
```
ibuenros: nit: rename to `encodeOutputStream`, `decodeInputStream`

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Extract encryption related information from taskState
+ */
+public class EncryptionConfigParser {
+  private static final String ENCRYPT_PREFIX = ".encrypt";
+
+  // TODO these are dupe with ConfigurationKeys - this is to eventually allow for encryption plugins down the road
+  // that just get handed a set of config and can decide to define their own config below the branch prefix if
+  // they want
+  public static final String ENCRYPTION_ALGORITHM_KEY = "algorithm";
```
ibuenros: These also exist in `ConfigurationKeys`, is that intentional?

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
```diff
@@ -0,0 +1,102 @@
...
+  public static String getEncryptionType(Map<String, Object> properties) {
+    return (String)properties.get(ENCRYPTION_ALGORITHM_KEY);
+  }
+
+  public static String getKeystorePath(Map<String, Object> properties) {
+    return (String)properties.get(ENCRYPTION_KEYSTORE_PATH_KEY);
+  }
+
+  public static String getKeystorePassword(Map<String, Object> properties) {
+    return (String)properties.get(ENCRYPTION_KEYSTORE_PASSWORD_KEY);
```
ibuenros: Should we use `PasswordManager` for this?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  @Override
+  public OutputStream wrapOutputStream(OutputStream origStream) {
+    return new StreamInstance(random, getKeyRecords(), origStream).wrapOutputStream();
+  }
+
+  @Override
+  public InputStream wrapInputStream(InputStream origStream) throws IOException {
+    throw new AssertionError("not implemented yet");
```
ibuenros: What about `UnsupportedOperationException`?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
```diff
@@ -0,0 +1,221 @@
...
+ * 3. A header will be written [keyId][ivLength][base64 encoded iv]
+ * 4. Ciphertext will be base64 encoded and written out. We do not insert linebreaks.
+ */
+@Slf4j
+public class RotatingAESCodec implements StreamCodec {
+  private static final int AES_KEY_LEN = 16;
+
+  private final Random random;
+  private final CredentialStore credentialStore;
+  private List<KeyRecord> keyRecords;
```
ibuenros: I think this needs to be `volatile` for thread safety.

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  /**
+   * Helper class that keeps state around for a wrapped output stream. Each stream will have a different
+   * selected key, IV, and cipher state.
+   */
+  static class StreamInstance {
+    private final OutputStream origStream;
+    private final List<KeyRecord> keyRecords;
```
ibuenros: Why is `keyRecords` and `random` needed here? `initCipher` could be done by the parent class.

gobblin-api/src/main/java/gobblin/crypto/CredentialStore.java
eogren: The AES Encryptor needs it because it needs to pick a key at random

gobblin-api/src/main/java/gobblin/writer/StreamCodec.java
eogren: sure

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
eogren: it is - the ConfigurationKeys are there just for documentation. eventually I could see the getting removed and docs somehow autogenerated from the keys in this class

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
eogren: passwords in passwords in passwords :) But yes, will update - didn't realize PasswordManager existed

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
eogren: will add, although I thought the synchronized block around getKeyRecords() means we don't need to make it volatile?

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
eogren: sure

gobblin-core-base/src/main/java/gobblin/crypto/RotatingAESCodec.java
eogren: Hm, you are right - will pull it out. I do plan to leave the initCipher() logic inside of StreamInstance because it initializes the IV and cipher state which have to be per stream, but can move getNextKey() out and have the key passed into it by the outer class.

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
ibuenros: We no longer strictly put configuration keys in `ConfigurationKeys` (otherwise that class would get huge). I'd recommend removing them from there.

PR #1620, opened 2017-02-13
RecordWMetadata -> EncryptedRecordWMetadata converter

eogren:
for record by record encryption

I relaxed the emptyOrNull constraint in ForkOperatorUtils.getPropertyNameForBranch() so I can call getPropertyName("", <branchInfo>) to get the suffix that would be added on to all config parameters for a given branch


gobblin-core-base/src/main/java/gobblin/converter/RecordWithMetadataToEncryptedRecordWithMetadataConverter.java
```diff
@@ -19,44 +19,55 @@
...
-    //TODO: Encrypt this
     try {
-      byte[] serialized = GSON.toJson(inputRecord.getRecord()).getBytes("UTF-8");
-      SerializedRecordWithMetadata serializedRecordWithMetadata = new SerializedRecordWithMetadata(serialized,
+      if (!(inputRecord.getRecord() instanceof byte[])) {
+        throw new DataConversionException(
+            "RecordWithMetadataToEncryptedRecordWithMetadataConverter requires input record to be of type byte[]");
+      }
+
+      Map<String, Object> encryptionConfig = EncryptionConfigParser.getConfigForBranch(workUnit);
```
shirshanka: Converters support an init method that you can override, where you can do workunit-level instantiation of configuration.

gobblin-core-base/src/main/java/gobblin/crypto/EncryptionConfigParser.java
```diff
@@ -86,15 +110,14 @@ public static String getKeystorePassword(Map<String, Object> properties) {
...
+   * @param branchSuffix Suffix for all config properties
    * @return Transformed properties as described above
    */
   private static Map<String, Object> extractPropertiesForBranch(
-      Properties properties, String prefix, int numBranches, int branch) {
+      Properties properties, String prefix, String branchSuffix) {
 
     Map<String, Object> ret = new HashMap<>();
-    String branchSuffix = (numBranches > 1) ? String.format(".%d", branch) : "";
+    //String branchSuffix = (numBranches > 1) ? String.format(".%d", branch) : "";
```
shirshanka: nit: remove commented code.

eogren:
Updated with new SerializedRecord (w content-type concept). One thing that is kind of confusing: SerializedRecordWithMetadata can be read two different ways: SerializedRecord With Metadata, which is what it actually is, or Serialized (RecordWithMetadata). Don't know if there is a better naming scheme so it's obvious how to break the phrases up.

gobblin-core-base/src/main/java/gobblin/converter/AnyToJsonRecordWithMetadataConverter.java
```diff
@@ -0,0 +1,91 @@
...
+      @SuppressWarnings("unchecked")
+      RecordWithMetadata<JsonElement> jsonRecord = (RecordWithMetadata) inputRecord;
+
+      return new SingleRecordIterable<>(jsonRecord);
+    } else if (inputRecord instanceof JsonElement) {
+      JsonElement jsonElement = (JsonElement) inputRecord;
+      Map<String, Object> metadata;
+
+      if (recordTypeName != null) {
+        metadata = ImmutableMap.<String, Object>of(RecordWithMetadata.RECORD_NAME, recordTypeName);
```
shirshanka: Looks like you're initializing metadata to be immutable map. 
Is there a specific reason you went this way? 
e.g. It would be nice if downstream converters could potentially add more metadata to the record as it passes through.

gobblin-core-base/src/main/java/gobblin/converter/AnyToJsonRecordWithMetadataConverter.java
```diff
@@ -0,0 +1,91 @@
...
+      recordTypeName = ((Schema) inputSchema).getFullName();
+    }
+    return "";
+  }
+
+  @Override
+  public Iterable<RecordWithMetadata<JsonElement>> convertRecord(String outputSchema, Object inputRecord,
+      WorkUnitState workUnit)
+      throws DataConversionException {
+    if (inputRecord instanceof RecordWithMetadata && ((RecordWithMetadata) inputRecord)
```
shirshanka: Realized something that we should probably make a pattern for these input-type-adjusting converters. 

Since converter chains are generally static, you can lock down the input type and therefore the cast pattern after the first record passes through. This will allow us to skip instanceof checks after the first record passes through. 

Maybe with a config flag to revert to the unoptimized behavior if the pipeline author has set up a pipeline that can generate different types of records. 

We don't have to address this in this PR, but might be worth picking up as a follow-up PR. Thoughts?

gobblin-core-base/src/main/java/gobblin/converter/AnyToSerializedRecordConverter.java
```diff
@@ -0,0 +1,100 @@
...
+
+
+/**
+ * Converts multiple source objects to a SerializedRecord object.
+ */
+public class AnyToSerializedRecordConverter extends Converter<String, String, Object, SerializedRecord> {
+
+  private final static List<String> BINARY_CONTENT_TYPE = ImmutableList.of("application/octet-stream");
+  private final static List<String> JSON_CONTENT_TYPE = ImmutableList.of("application/json");
+  private final static Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
```
shirshanka: Any specific reason for turning off HTML escaping? 

We have different ways of creating new Gson objects in the codebase... should all of them disable HTML escaping?

gobblin-core-base/src/main/java/gobblin/converter/AnyToJsonRecordWithMetadataConverter.java
eogren: Good point - I just tend to mark everything as immutable out of habit, but agree that it doesn't make sense here.

gobblin-core-base/src/main/java/gobblin/converter/AnyToJsonRecordWithMetadataConverter.java
eogren: I think the pattern is useful just so people aren't typing the same code out all the time, yeah - I thought about trying to add that as part of this change but didn't want things to snowball too much. 

I suspect it won't matter much performance wise  - the cost of JSON parsing, Base64 encoding etc far dominates the instanceof/cast checks from what I've seen.

gobblin-core-base/src/main/java/gobblin/converter/AnyToSerializedRecordConverter.java
eogren: If I didn't disable html encoding, Gson was turning the string "abc==" to "abc\u003d\u003d", so it was wasting a bunch of space. It's possible they all should; it's also possible there should just be one global Gson object across the system. I don't know if there's a way to do that and let you register custom serializers though...

PR #1622, opened 2017-02-14
Add system config to job config when loading jobs in the cluster JobC…

htran1:
…onfigurationManager. This is required to propagate system config such as the master key file location.


htran1:
@abti can you review?

PR #1625, opened 2017-02-14
Avro Combine input format will no longer open every avro file.

ibuenros:
This PR removes the need to open every file to read the schema in a compaction job, and simplifies the Avro compaction input job as a consequence.


gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
```diff
@@ -80,6 +81,13 @@
 
   private static final String AVRO_SUFFIX = ".avro";
 
+  public static class AvroPathFilter implements PathFilter {
```
ydai1124: Is this used?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/avro/AvroKeyRecursiveCombineFileInputFormat.java
```diff
@@ -69,15 +71,14 @@
...
 
   @Override
   public List<InputSplit> getSplits(JobContext cx) throws IOException {
-    setSplitSize(cx);
-    List<InputSplit> splits = getSplits(cx, Arrays.asList(getInputPaths(cx)));
-    return cleanSplits(splits);
+    Job modifiedJob = Job.getInstance(cx.getConfiguration());
+    setSplitSize(modifiedJob);
+    FileInputFormat.setInputDirRecursive(modifiedJob, true);
+    return cleanSplits(super.getSplits(modifiedJob));
```
ydai1124: Where do you filter out non-Avro files in the inputSplits?

ibuenros: They aren't. Hidden files are automatically filtered, the assumption here is that all non-hidden files are avro files (normally a reasonable assumption for MR jobs).

gobblin-utility/src/main/java/gobblin/util/AvroUtils.java
ibuenros: No, it was until I realized it broke the input format. I figures I'd leave it there for the future.

PR #1626, opened 2017-02-14
Removed base64 encoding inside EventhubDataWriter

yukuai518:
Remove base64 encoding inside EventhubDataWriter because we assume the converter already has the encoded string


gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubDataWriter.java
```diff
@@ -70,8 +68,6 @@
  *
  * Also this class supports sending multiple records in a batch manner.
  *
- * This class use Base64 to encode any byte array, so the consumer side should use the corresponding
- * decoder to recover the original bytes.
  */
 @Slf4j
 public class EventhubDataWriter implements SyncDataWriter<byte[]>, BatchAsyncDataWriter<byte[]> {
```
ibuenros: Can you make this take Strings as inputs instead?

PR #1627, opened 2017-02-15
Make KafkaSimpleStreamingSource to use Config

ydai1124:
The real `state` class in the source is actually `JobState`, which does not support `getProp`. Change it to Typesafe config.

### DIFF
--- a/gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingSource.java
+++ b/gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingSource.java
@@ -17,32 +17,34 @@
 
 package gobblin.source.extractor.extract.kafka;
 
-import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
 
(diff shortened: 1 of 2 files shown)


gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingSource.java
```diff
@@ -114,28 +121,30 @@ static public Consumer getKafkaConsumer(State state) {
...
-        LOG.debug("Partition info is {}", topicPartition);
-        WorkUnit workUnit = WorkUnit.create(extract);
-        setTopicNameInState(workUnit, topicPartition.topic());
-        workUnit.setProp(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY, topicPartition.topic());
-        setPartitionId(workUnit, topicPartition.partition());
-        workUnits.add(workUnit);
-      }
-      return workUnits;
+    Config config = ConfigUtils.propertiesToConfig(state.getProperties());
+    Consumer<String, byte[]> consumer = getKafkaConsumer(state);
```
shirshanka: You can replace this with getKafkaConsumer(config)  and change getKafkaConsumer to take in Config as parameter

shirshanka:
LGTM!

PR #1631, opened 2017-02-15
Fix over-eager EncryptionConfigParser

eogren:
@jinhyukchang 

- EncryptionConfigParser parser was picking up config settings it shouldn't care about

- Don't return a half-filled in config if mandatory parameters (like algorithm) aren't present


jinhyukchang:
LGTM. Thanks, @eogren

PR #1633, opened 2017-02-16
Modified Getting Started in docs to make it clearer how to obtain a G…

ibuenros:
…obblin distribution.


ibuenros:
Doc change, self- merging.

PR #1634, opened 2017-02-16
Unwrap value from Config when setting the value as a string in a Conf…

htran1:
…iguration object.


htran1:

PR #1635, opened 2017-02-18
Fix NullPointerException when Deserializer is not properly initialized

erwa:


gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -333,35 +334,39 @@ private static State getSerDeProps(SerDeInfo si) {
...
 
-      try {
-        SerDeUtils.initializeSerDe(deserializer, hiveConf, props, null);
-      } catch (SerDeException e) {
-        LOG.warn("Failed to initialize serde " + serde + " with properties " + props);
-      }
+    try {
+      SerDeUtils.initializeSerDe(deserializer, hiveConf, props, null);
+    } catch (SerDeException e) {
+      LOG.warn("Failed to initialize serde " + serde + " with properties " + props + " for table " + unit.getDbName() +
```
erwa: Adding more logging to print database and table.

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -333,35 +334,39 @@ private static State getSerDeProps(SerDeInfo si) {
...
+          "." + unit.getTableName());
+      return null;
```
erwa: This was the missing return statement. Without it, we might return an improperly initialized deserializer.

gradle/scripts/defaultBuildProperties.gradle
```diff
@@ -28,7 +28,7 @@ def BuildProperties BUILD_PROPERTIES = new BuildProperties(project)
     .register(new BuildProperty("doNotSignArtifacts", false, "Do not sight Maven artifacts"))
     .register(new BuildProperty("gobblinFlavor", "standard", "Build flavor (see http://gobblin.readthedocs.io/en/latest/developer-guide/GobblinModules/)"))
     .register(new BuildProperty("hadoopVersion", "2.3.0", "Hadoop dependencies version"))
-    .register(new BuildProperty("hiveVersion", "1.0.1", "Hive dependencies version"))
+    .register(new BuildProperty("hiveVersion", "1.1.0", "Hive dependencies version"))
```
erwa: Need to update Hive version for this patch: https://issues.apache.org/jira/browse/HIVE-7868

Previously, if there was an invalid schema, AvroSerDe would return a dummy error schema, and SerDeUtils.initializeSerDe() would return successfully. Now SerDeUtils.initializeSerDe throws a SerDeException instead.

erwa:
Internal JIRA: ETL-5440

PR #1636, opened 2017-02-20
Migrating from Junit to TestNG Assert in few test classes

abti:
Migrating from Junit to TestNG Assert in few test classes


abti:
Minor changes, self approving

PR #1637, opened 2017-02-20
Add junit dependency that is required for kafka-08 tests.

htran1:
Update hadoop version to 2.6.0 since that is required by the hive 1.1.0 version.


htran1:
@abti, please take a look.

PR #1638, opened 2017-02-20
Add job execution mode for Gobblin Cluster to submit jobs to the jobE…

htran1:
…xecutor in the underlying jobScheduler instead of occupying the scheduler thread.


htran1:
@abti, please review.

PR #1639, opened 2017-02-21
Revert "PasswordManager to support encrypt.key.val"

jinhyukchang:
Reverts linkedin/gobblin#1615

Reverting supporting of "encrypt.key.val" as state can be persisted.

@htran1 , could you take a look? Thanks.


htran1:
Looks fine.

PR #1645, opened 2017-02-21
AsyncWriterManager not to be retried by RetryWriter

jinhyukchang:
(Integration test passed.)

AsyncWriterManager is not designed to be retried by RetryWriter because AsyncWriterManager has its own retry implementation or it can delegate to the writer itself.

Throwing NonTransientException, RetryWriter won't retry.


jinhyukchang:
@shirshanka Could you take a look when you have a chance? Thanks.

PR #1646, opened 2017-02-22
Method to add topic specific properties to work unit. Made dataset name and topic name pattern matching case insensitive

aditya1105:


aditya1105:
Trivial change. Merging myself.

PR #1647, opened 2017-02-22
Watermark storage cli : honor zk address from command line

shirshanka:


shirshanka:

PR #1648, opened 2017-02-22
Add some CLI options for keystore generation

eogren:
CLI utilities to generate a JCEKSKeystore file and count how many keys are inside of it (and implicitly check the password of the store)


eogren:
self-merging, CLI only

PR #1650, opened 2017-02-23
Made GobblinServiceManager optional based via config

abti:
Made GobblinServiceManager optional based via config


abti:
Minor change. Self approving.

PR #1653, opened 2017-02-24
Changed access modifier for constructor of FlowConfigsResource to public for RestLI runtime initialization

abti:
Changed access modifier for constructor of FlowConfigsResource to public for RestLI runtime initialization


abti:
Minor change, self approving.

PR #1654, opened 2017-02-24
Made FlowCatalog named injectable in FlowConfigResources, and added hooks for RestLI bootstrap

abti:
Made FlowCatalog named injectable in FlowConfigResources, and added hooks for RestLI bootstrap


abti:
Minor changes, self approving

PR #1655, opened 2017-02-24
Add decryption support and a CLI command to use it

eogren:
@ibuenros or @shirshanka , can you please take a look?


eogren:
self-merging since these are utility functions

PR #1660, opened 2017-02-27
Fix to stop travis build from failing

aditya1105:
Travis build was failing because of two reasons. If something else fails I will add it in this PR
1) Package junit was not found for modules gobblin-data-management and gobblin-ingestion
  Fix is to add junit as a dependency.
2) Once above is fixed it starts failing due to findbugs
Bug type NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS (click for details) 
In class gobblin.cluster.ScheduledJobConfigurationManager
In method gobblin.cluster.ScheduledJobConfigurationManager.fetchJobSpecs()
Called method gobblin.cluster.JobConfigurationManager.postDeleteJobConfigArrival(String, Properties)
At ScheduledJobConfigurationManager.java:[line 138]
Argument 2 is definitely null but must not be null
Definite null passed to dangerous method call target gobblin.cluster.JobConfigurationManager.postDeleteJobConfigArrival(String, Properties)

Fix is to add new Properties() instead of passing null

### DIFF
--- a/gobblin-cluster/src/main/java/gobblin/cluster/ScheduledJobConfigurationManager.java
+++ b/gobblin-cluster/src/main/java/gobblin/cluster/ScheduledJobConfigurationManager.java
@@ -20,6 +20,7 @@ package gobblin.cluster;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


aditya1105:
Merging

PR #1662, opened 2017-02-28
Add compression and encryption support to SimpleDataWriter

eogren:
The FsDataWriterBuilder changes look a little weird with encoders being created only in the build() method -- however, we can't do it in constructor because the Builder isn't really actually fully initialized until build() is called, and we need to cache the encoder state in the Builder itself because getFileName() exists in the builder and isn't passed down to the writer at construct time.

### DIFF
--- a/gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
+++ b/gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package gobblin.compression;
+
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
+
+import gobblin.configuration.ConfigurationKeys;
+import gobblin.configuration.State;
+import gobblin.util.ForkOperatorUtils;
+
+
+/**
+ * Functions for parsing compression configuration
+ */
+public class CompressionConfigParser {
+  private static final String COMPRESSION_TYPE_KEY = "type";
+
+  /**
+   * Retrieve configuration settings for a given branch.
+   * @param taskState Task state
+   * @param numBranches # of branches in the state
+   * @param branch Branch to retrieve
+   * @return Map of properties for compression
+   */
+  public static Map<String, Object> getConfigForBranch(State taskState, int numBranches, int branch) {
+    String typePropertyName =
+        ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.WRITER_CODEC_TYPE, numBranches, branch);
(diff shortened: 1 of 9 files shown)


gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final String COMPRESSION_TYPE_KEY = "type";
+
+  /**
+   * Retrieve configuration settings for a given branch.
+   * @param taskState Task state
+   * @param numBranches # of branches in the state
+   * @param branch Branch to retrieve
+   * @return Map of properties for compression
+   */
+  public static Map<String, Object> getConfigForBranch(State taskState, int numBranches, int branch) {
```
ibuenros: What other kinds of properties will be in the map? It seems overkill for the moment, but there may be some use in the future.

gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Return compression type
+   * @param properties Compression config settings
+   * @return String representing compression type, null if none exists
+   */
+  public static String getCompressionType(Map<String, Object> properties) {
+    return (String) properties.get(COMPRESSION_TYPE_KEY);
+  }
+
+  private CompressionConfigParser() {
```
ibuenros: Does this compile?

gobblin-core-base/src/main/java/gobblin/compression/CompressionFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package gobblin.compression;
+
+import java.util.Map;
+
+import gobblin.writer.StreamCodec;
+
+
+public class CompressionFactory {
```
ibuenros: Please add javadoc

gobblin-core-base/src/main/java/gobblin/compression/CompressionFactory.java
```diff
@@ -0,0 +1,38 @@
...
+  public static StreamCodec buildStreamCompressor(Map<String, Object> properties) {
+    String type = CompressionConfigParser.getCompressionType(properties);
```
ibuenros: I'd prefer to use aliases here instead of a static switch.

gobblin-core/src/main/java/gobblin/writer/SimpleDataWriterBuilder.java
```diff
@@ -34,7 +49,33 @@
...
+      encoders.add(EncryptionFactory.buildStreamEncryptor(encryptionConfig));
+    }
   }
 
+  @Override
+  public String getFileName(State properties) {
+    StringBuilder filenameBuilder = new StringBuilder(super.getFileName(properties));
+    for (StreamCodec codec : encoders) {
+      filenameBuilder.append('.');
+      filenameBuilder.append(codec.getTag());
```
ibuenros: You could add the additional extensions in the `Writer` and avoid the weirdness, can't you?

gobblin-core-base/src/main/java/gobblin/compression/CompressionFactory.java
ibuenros: What's the advantage of having this separate from `EncryptionFactory`?

gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
eogren: Yeah, it's solely for future proofing. The only thing I could think of off the bat is if we wanted to tweak the compression level - that said figured doing it this way leaves us with maximum room for extensibility later. It is slightly overkill but don't think it makes things really that much more complex.

gobblin-core-base/src/main/java/gobblin/compression/CompressionConfigParser.java
eogren: Why wouldn't it? The idea is that you shouldn't instantiate a ConfigParser of these as the API is all static methods

PR #1664, opened 2017-02-28
Added metrics to rest.li limiter.

ibuenros:
Changes:

* Created a `MetricContextFactory` for broker, along with a `MetricContextKey` and a unit test.
* Rest.li Limiter supports emitting metrics. The factory automatically acquires a metric context from the broker.

* In Gobblin runtime, `WorkUnitState` gets a broker builder instead of a pre-built broker so it can inject its own custom configurations (not done currently). Eventually, I would want to replace all `MetricContext`s in gobblin runtime for ones acquired from the broker.
* Added a scope id to the `ScopeInstance` interface.
* Minor changes to generics in broker (fixes compiler/ide warnings)


ibuenros:
@yukuai518 can you review? Thanks!

PR #1666, opened 2017-03-01
Remove data pull upper bound for the last work unit of query based task

zxcware:
As between the time when data query was created and that was executed, there might be some new data generated in the source. Removing the upper bounds will help us grab the new data.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
@@ -188,14 +188,8 @@ public abstract class QueryBasedExtractor<S, D> implements Extractor<S, D>, Prot
 
     // Don't remove if user specifies one or is recorded in previous run
     if (this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_END_VALUE) != null ||
+        this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_APPEND_MAX_WATERMARK_LIMIT) != null ||
         this.workUnitState.getProp(ConfigurationKeys.WORK_UNIT_STATE_ACTUAL_HIGH_WATER_MARK_KEY) != null) {
-      return false;
-    }
-
-    ExtractType extractType =
-        ExtractType.valueOf(this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_EXTRACT_TYPE).toUpperCase());
-    // Don't remove if the run is for "APPEND"
-    if (extractType != ExtractType.SNAPSHOT) {
       return false;
     }
 
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
@@ -17,6 +17,7 @@
 
 package gobblin.source.extractor.extract;
 
+import com.google.common.annotations.VisibleForTesting;
 import gobblin.source.extractor.DataRecordException;
 import gobblin.source.extractor.Extractor;
 import gobblin.source.extractor.exception.ExtractPrepareException;
@@ -68,11 +69,11 @@ public abstract class QueryBasedExtractor<S, D> implements Extractor<S, D>, Prot
   private S outputSchema;
   private long sourceRecordCount = 0;
   private long highWatermark;
-  private long estimatedActualHighWatermark = ConfigurationKeys.DEFAULT_WATERMARK_VALUE;
 
   private Iterator<D> iterator;
   protected final List<String> columnList = new ArrayList<>();
-  private final List<Predicate> predicateList = new ArrayList<>();
+  @VisibleForTesting
+  protected final List<Predicate> predicateList = new ArrayList<>();
 
   private S getOutputSchema() {
     return this.outputSchema;
@@ -149,7 +150,6 @@ public abstract class QueryBasedExtractor<S, D> implements Extractor<S, D>, Prot
 
         if (shouldRemoveDataPullUpperBounds()) {
           this.removeDataPullUpperBounds();
-          this.setEstimatedActualHighWatermark();
         }
         this.iterator = this.getIterator();
       }
@@ -174,8 +174,7 @@ public abstract class QueryBasedExtractor<S, D> implements Extractor<S, D>, Prot
   /**
    * Check if it's appropriate to remove data pull upper bounds in the last work unit, fetching as much data as possible
    * from the source. As between the time when data query was created and that was executed, there might be some
-   * new data generated in the source. Removing the upper bounds will help us grab the new data. It might result in a
-   * conservative high watermark, which is mitigated by {@link #setEstimatedActualHighWatermark()}
+   * new data generated in the source. Removing the upper bounds will help us grab the new data.
    *
    * Note: It's expected that there might be some duplicate data between runs because of removing the upper bounds
    *
@@ -187,14 +186,16 @@ public abstract class QueryBasedExtractor<S, D> implements Extractor<S, D>, Prot
       return false;
     }
 
-    // Don't remove if the run is for 'APPEND'
-    if (!this.isFullDump()) {
-      return false;
-    }
-
     // Don't remove if user specifies one or is recorded in previous run
     if (this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_END_VALUE) != null ||
         this.workUnitState.getProp(ConfigurationKeys.WORK_UNIT_STATE_ACTUAL_HIGH_WATER_MARK_KEY) != null) {
+      return false;
+    }
+
+    ExtractType extractType =
+        ExtractType.valueOf(this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_EXTRACT_TYPE).toUpperCase());
+    // Don't remove if the run is for "APPEND"
+    if (extractType != ExtractType.SNAPSHOT) {
       return false;
     }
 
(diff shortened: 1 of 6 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
```diff
@@ -166,6 +172,72 @@ public D readRecord(@Deprecated D reuse) throws DataRecordException, IOException
...
+    return true;
+  }
+
+  /**
+   * Remove all upper bounds in the predicateList used for pulling data
+   */
+  private void removeDataPullUpperBounds() {
+    log.info("Removing data pull upper bound for last work unit");
+    Iterator<Predicate> it = predicateList.iterator();
+    while (it.hasNext()) {
```
htran1: You can use `for (Predicate predicate : predicateList)` instead.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -78,6 +78,7 @@
   public static final boolean DEFAULT_SOURCE_OBTAIN_TABLE_PROPS_FROM_CONFIG_STORE = false;
   private static final String QUERY_BASED_SOURCE = "query_based_source";
   public static final String WORK_UNIT_STATE_VERSION_KEY = "source.querybased.workUnitState.version";
+  public static final String IS_LAST_WORK_UNIT = "is.last.work.unit";
```
htran1: How about "source.querybased.workUnitState.isLastWorkUnit" and IS_LAST_WORK_UNIT_KEY?

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
```diff
@@ -166,6 +172,72 @@ public D readRecord(@Deprecated D reuse) throws DataRecordException, IOException
...
+        it.remove();
+      }
+    }
+  }
+
+  /**
+   * Set an estimated actual high watermark when the work unit is the last one
+   * and its upper bounds to fetch data are removed
+   */
+  private void setEstimatedActualHighWatermark() {
```
htran1: I think this function should be removed. Our local timestamp may not be in sync with the remote source, so this is not safe. The this.highWatermark value when computed from the source will give a timestamp that is appropriate for the source. Since we are removing the upper bound on the last partition, we may pull more data relative to this computed high watermark. We can handle dedup of duplicates when merging snapshots and appends.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
zxcware: Discussed offline. It will be named as 'source.querybased.is.last.work.unit'

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
zxcware: We'll do a remove on a predicate. Using iterator here is fine

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
zxcware: Good point. I'll remove the function

zxcware:
@ibuenros Please review. Thanks.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -78,6 +79,7 @@
   public static final boolean DEFAULT_SOURCE_OBTAIN_TABLE_PROPS_FROM_CONFIG_STORE = false;
   private static final String QUERY_BASED_SOURCE = "query_based_source";
   public static final String WORK_UNIT_STATE_VERSION_KEY = "source.querybased.workUnitState.version";
+  public static final String IS_LAST_WORK_UNIT = "source.querybased.is.last.work.unit";
```
htran1: Please use source.querybased.isLastWorkUnit. "." should separate namespace units and not a general replacement for space.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -33,26 +39,49 @@
...
 import gobblin.source.extractor.watermark.WatermarkType;
 
-
 /**
  * An implementation of default partitioner for all types of sources
  */
 public class Partitioner {
   private static final String WATERMARKTIMEFORMAT = "yyyyMMddHHmmss";
   private static final Logger LOG = LoggerFactory.getLogger(Partitioner.class);
+  public static final String HAS_USER_SPECIFIED_HIGH_WATERMARK = "has.user.specified.high.watermark";
```
htran1: Add a namespace and use camel case, so something like "partitioner.hasUserSpecifiedHighWatermark".

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -33,26 +39,49 @@
...
+      if (p1 == null && p2 == null) {
+        return 0;
+      }
+      if (p1 == null) {
+        return -1;
+      }
+      if (p2 == null) {
+        return 1;
+      }
+      return p1.getLowWatermark() > p2.getLowWatermark() ? 1 : -1;
```
htran1: Should add p1.getLowWatermark() == p2.getLowWatermark() case for completeness. I think that can just be merged with the first if check.

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedSource.java
zxcware: Good point. Dot is to separate each meaningful name, not word.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zxcware: "partitioner.hasUserSpecifiedHighWatermark" is a better string. I'll use it.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zxcware: For compactness, I intentionally merge the '=' case into '<'.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
```diff
@@ -0,0 +1,24 @@
+package gobblin.source.extractor.partition;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public class Partition {
```
ibuenros: Please add javadoc

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
```diff
@@ -0,0 +1,24 @@
...
+  @Getter private long lowWatermark;
+  @Getter private long highWatermark;
+
+  // Indicate if the Partition highWatermark is set as user specifies, not computed on the fly
```
ibuenros: Use javadoc, not comment

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
```diff
@@ -0,0 +1,24 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public class Partition {
+  @Getter private long lowWatermark;
```
ibuenros: Make variables final whenever that is the correct semantic

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -33,26 +39,50 @@
...
+      if (p1 == null && p2 == null) {
+        return 0;
+      }
+      if (p1 == null) {
+        return -1;
+      }
+      if (p2 == null) {
+        return 1;
+      }
+      // For compactness, treat the '==' case as '<'
```
ibuenros: This is incorrect, comparison must be reflective (e.g. a < b implies b > a)

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -33,26 +39,50 @@
...
+      return p1.getLowWatermark() > p2.getLowWatermark() ? 1 : -1;
```
ibuenros: You can do `Long.compare(p1.getLowWatermark(), ...)`

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -95,11 +125,44 @@ public Partitioner(SourceState state) {
...
+     */
+    Long highestWatermark = Collections.max(partitionMap.values());
+
+    for (Map.Entry<Long, Long> entry : partitionMap.entrySet()) {
+      Long partitionHighWatermark = entry.getValue();
+      // Apply hasUserSpecifiedHighWatermark to the last partition, which has highestWatermark
+      if (partitionHighWatermark.equals(highestWatermark)) {
+        partitions.add(new Partition(entry.getKey(), partitionHighWatermark, hasUserSpecifiedHighWatermark));
+      } else {
+        partitions.add(new Partition(entry.getKey(), partitionHighWatermark));
```
ibuenros: Please use the three argument constructor for clarity.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
zxcware: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
zxcware: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zxcware: Fixed

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zxcware: Nice!

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
zxcware: Fixed

PR #1668, opened 2017-03-02
Add mdc key for storing the job id to distinguish log entries generat…

htran1:
…ed in gobblin cluster.


htran1:
@abti please review.

htran1:
Looks like this overlaps with #967 and currently doesn't handle the thread pool in the task runner.

PR #1596, opened 2017-02-01
Watermarking changes for streaming

shirshanka:
Reworked watermark management to use fine-grained accounting to handle forks / converters correctly. 
Made AsyncWriterManager watermark aware. 
Added a Zk-based Watermark store using the existing ZkStateStore. 
Added sample pull files.

PR #1673, opened 2017-03-06
Forks: Making primitive types copyable by default, tests and bug fixes

shirshanka:
Currently, record types must implement Copyable to participate in fork-based pipelines. 
This PR relaxes this requirement by adding support for byte[] and immutable types like String etc. 

There are a few notable un-filed issues that were discovered that this PR also fixes. 
a) Fork specific config (typically prefixed by numerical branch ids e.g. writer.staging.dir.0) were being mangled due to transformation from Properties -> TypeSafe Config -> Properties. TypeSafe Config surrounds numeric looking tokens with quotes, leading to these config keys looking like writer.staging.dir."0" in TaskState. This PR works around this by stripping all quotes from keys in the Properties bag when transiting through ConfigUtils::configToProperties

b) #1596 inadvertently introduced a bug in Task.java where forked records weren't being sent into the correct fork.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
+++ b/gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
@@ -25,12 +25,10 @@ import org.apache.commons.configuration.ConfigurationException;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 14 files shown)


gobblin-runtime/src/main/java/gobblin/scheduler/PathAlterationListenerAdaptorForMonitor.java
```diff
@@ -30,6 +30,8 @@
 
 import com.google.common.collect.Maps;
 
+import lombok.extern.slf4j.Slf4j;
```
shirshanka: done.

shirshanka:
Build + tests pass locally. Travis is slooow, merging this in.

PR #1678, opened 2017-03-07
Batch level metadata support

eogren:
- Each writer keeps track of some GlobalMetadata (which for now is only TransferEncoding).
- BaseDataPublisher merges all of these metadata objects together using the GlobalMetadataMerger, and writes the merged data out to disk if the appropriate config is set
- Add a JsonUtil class to hold Jackson objects so we aren't constructing them all over the place. This will also let us hang config settings off there if we figure out there are good defaults to set.

To follow in future PRs:
 - Support to set GlobalMetadata from config/an arbitrary retriever
 - Allow sources + converters to optionally set global metadata
 - Per-record metadata

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -19,13 +19,16 @@ package gobblin.publisher;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 
-import gobblin.util.HadoopUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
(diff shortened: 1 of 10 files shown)


eogren:
@jinhyukchang, wondering if you feel comfortable reviewing this?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -133,13 +143,68 @@ public BaseDataPublisher(State state) throws IOException {
...
+    for (int i = 0; i < numBranches; i++) {
+      if (shouldPublishMetadataForBranch(i)) {
+        metadataMergers.add(buildPublisherForBranch(i));
+        mergersConfigured = true;
+      } else {
+        metadataMergers.add(null);
+      }
+    }
+
+    return mergersConfigured ? metadataMergers : null;
```
shirshanka: Can use Collections.EMPTY_LIST as the default instead of null?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -133,13 +143,68 @@ public BaseDataPublisher(State state) throws IOException {
...
+  }
+
+  private MetadataMerger<String> buildPublisherForBranch(int branchId) {
+    String keyName = ForkOperatorUtils.getPropertyNameForBranch(ConfigurationKeys.DATA_PUBLISH_WRITER_METADATA_MERGER_NAME_KEY,
+        this.numBranches, branchId);
+    String className = this.getState().getProp(keyName, ConfigurationKeys.DATA_PUBLISH_WRITER_METADATA_MERGER_NAME_DEFAULT);
+
+    try {
+      Class<?> mdClass = Class.forName(className);
```
shirshanka: Check if GobblinConstructorUtils can be used here.

gobblin-core/src/main/java/gobblin/writer/SimpleDataWriter.java
```diff
@@ -118,7 +119,8 @@ public long recordsWritten() {
    * @return number of bytes written
    */
   @Override
-  public long bytesWritten() throws IOException {
+  public long bytesWritten()
```
shirshanka: Seems like there were no real changes to this class?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
eogren: I originally had it that way but decided that a check for (mergers == null) later on makes more sense than mergers.size() == 0. Thoughts? I don't feel super strongly either way.

gobblin-core/src/main/java/gobblin/writer/SimpleDataWriter.java
eogren: Oops, thought I reverted all these - will fix

PR #1681, opened 2017-03-07
Bug fixes for Spec Producer and Consumer in Service and Cluster

abti:
Bug fixes for Spec Producer and Consumer in Service and Cluster as well as a few SerDe changes (that are placeholders for now, and will be deleted with Avro based SerDe)


abti:
Pair reviewed. Merging in.

PR #1686, opened 2017-03-08
Gobblin Json module

treff7es:
Adding gobblin json module with TimeBasedJsonWriter, SimpleJsonCompressionWriter
Adding additional retries to WriterUtils and MRCompactorJobRunner which can fail on eventual consistency on S3
Making JobStateToJsonConverter to work with mysql state store

### DIFF
--- a/gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
+++ b/gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
@@ -17,10 +17,10 @@
 
 package gobblin.util;
 
-import java.util.Collection;
-
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.hadoop.fs.Path;
+
+import java.util.Collection;
 
 
 /**
@@ -37,7 +37,7 @@ public abstract class RecordCountProvider {
    * Convert a {@link Path} from another {@link RecordCountProvider} so that it can be used
    * in {@link #getRecordCount(Path)} of this {@link RecordCountProvider}.
    */
-  public Path convertPath(Path path, RecordCountProvider src) {
+  public Path convertPath(Path path, String extension, RecordCountProvider src) {
     if (this.getClass().equals(src.getClass())) {
       return path;
     }
--- a/gobblin-modules/gobblin-json/src/main/java/gobblin/mapreduce/JsonKeyMapper.java
+++ b/gobblin-modules/gobblin-json/src/main/java/gobblin/mapreduce/JsonKeyMapper.java
@@ -1,0 +1,54 @@
+package gobblin.mapreduce;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.mapreduce.Mapper;
+
+import java.io.IOException;
+
+/**
+ * Mapper for json logs.
+ * If you specify {@link CONF_KEY_TIMESTAMP_FIELD} then it will be added as prefix for the mapper key. In this way
+ * file can be sorted.
+ * Created by tamasnemeth on 12/11/16.
+ */
+public class JsonKeyMapper extends Mapper<LongWritable, Text, Text, Text> {
+
+  private static final String CONF_KEY_TIMESTAMP_FIELD = "compaction.timestamp.field";
+
+  private String timestampField;
+
+  public enum EVENT_COUNTER {
+    RECORD_COUNT
+  }
+
+  @Override
+  protected void setup(Context context) throws IOException, InterruptedException {
+    super.setup(context);
+
+    timestampField = context.getConfiguration().get(CONF_KEY_TIMESTAMP_FIELD);
+
+  }
+
+  @Override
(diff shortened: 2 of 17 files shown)


treff7es:
This is my initial pull request for better json support for Gobblin. I created the pull request as a conversation starter if this make sense for you in the current form and how to change to fit into your codebase if it make sense for you.
We started to use this module in our environment where logs are in json format.

gobblin-modules/gobblin-json/src/main/java/gobblin/writer/partitioner/TimeBasedJsonWriterPartitioner.java
```diff
@@ -0,0 +1,108 @@
...
+  private long getRecordTimestamp(Optional<JsonElement> writerPartitionColumnValue) {
+    int multiplier = 1;
+
+    //Unix timestamp is second precision and java timestamp is millisecond
+    if (this.isUnixTimestamp) {
+        multiplier = UNIX_TIMESTAMP_MULTIPLIER;
+    }
+
+    return writerPartitionColumnValue.isPresent() ? (Long) writerPartitionColumnValue.get().getAsLong()*multiplier
+        : System.currentTimeMillis();
```
abti: Defaulting to System.currentTimeMillis in case the writerPartitionColumn is absent is tricky, if I understand correctly this means that the same json would yield into arbitrarily different partitions every time the compaction is run?

gobblin-modules/gobblin-json/src/main/java/gobblin/writer/SimpleJsonCompressionDataWriter.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * An implementation of {@link DataWriter} that writes json objects directly to HDFS.
+ *
+ * This class accepts one new configuration parameters:
+ * <ul>
+ * <li>{@link SIMPLE_WRITER_COMPRESSION_CODEC} accepts a codec name, which implements the
+ * {@link org.apache.hadoop.io.compress.CompressionCodec} interface like {@link org.apache.hadoop.io.compress.GzipCodec}
+ * </ul>
+ */
+public class SimpleJsonCompressionDataWriter extends FsDataWriter<JsonElement> {
```
abti: Just wondering if this can be CompressionDataWriter for generic FsDataWriter<? implements Serializable> rather than being specific to JsonElement?

gobblin-modules/gobblin-json/src/main/java/gobblin/mapreduce/JsonKeyMapper.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override
+  protected void map(LongWritable key, Text value, Context context) throws IOException, InterruptedException {
+    Text outKey = value;
+
+    JsonParser jsonParser = new JsonParser();
+    JsonObject inputRecord = (JsonObject) jsonParser.parse(value.toString());
+
+    if (StringUtils.isNotEmpty(this.timestampField)) {
```
abti: here isNotBlank is more applicable, because isNotEmpty(" ") would return true

gobblin-api/src/main/java/gobblin/util/RecordCountProvider.java
```diff
@@ -37,7 +37,7 @@
    * Convert a {@link Path} from another {@link RecordCountProvider} so that it can be used
    * in {@link #getRecordCount(Path)} of this {@link RecordCountProvider}.
    */
-  public Path convertPath(Path path, RecordCountProvider src) {
+  public Path convertPath(Path path, String extension, RecordCountProvider src) {
```
treff7es: It's used in MRCompactorJobRunner as .avro was hardwired.
https://github.com/linkedin/gobblin/pull/1686/commits/f8fc1d890f74ef6329d0b5c12802d5136833354a#diff-dfd13665b702c59b24278904778cb5cdL326

gobblin-utility/src/main/java/gobblin/util/WriterUtils.java
```diff
@@ -258,10 +268,26 @@ public static void mkdirsWithRecursivePermission(FileSystem fs, Path path, FsPer
...
+      try {
+        LOG.warn("Path "+path+" does not exist. Will retry after 5 second. Attempt: "+ current);
+        current++;
+        TimeUnit.SECONDS.sleep(5);
+      } catch (InterruptedException e) {
+        throw new IOException(e);
+      }
+    }
+
+    if (maxRetry <= current) {
```
treff7es: RetryerFactory is currently in gobblin-core and gobblin-core depends on gobblin-utility. Would you mind if I move RetryerFactory to gobblin-utility?

abti: sure. I think that should be fine.

gobblin-modules/gobblin-json/src/main/java/gobblin/writer/partitioner/TimeBasedJsonWriterPartitioner.java
abti: Yes, I think that should be fine. For better understanding, please add a couple of lines about in class description.

abti:
Issue: https://issues.apache.org/jira/browse/GOBBLIN-47

Please update your PR title with following prefix: [GOBBLIN-47]

PR #1687, opened 2017-03-08
Added metrics to throttling server.

ibuenros:


ibuenros:
@yukuai518 can you review? Thanks!

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/gobblin/metrics/broker/MetricContextFactory.java
```diff
@@ -60,6 +61,19 @@ public String getName() {
       ScopedConfigView<S, MetricContextKey> config) throws NotConfiguredException {
 
     try {
+
+      if (config.getKey() instanceof SubTaggedMetricContextKey) {
+        SubTaggedMetricContextKey key = (SubTaggedMetricContextKey) config.getKey();
+
+        MetricContext parent = broker.getSharedResource(this, new MetricContextKey());
```
yukuai518: Is the "new MetricContextKey()" always the root key?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/gobblin/metrics/broker/MetricContextFactory.java
```diff
@@ -68,7 +82,12 @@ public String getName() {
         parentMetricContext = broker.getSharedResourceAtScope(this, config.getKey(), parentScope);
       }
 
-      MetricContext.Builder builder = parentMetricContext.childBuilder(broker.selfScope().toString());
+      // If this is the root scope, append a UUID to the name. This allows having a separate root context per broker.
+      String metricContextName = parents == null ?
+          config.getScope().name() + "_" + UUID.randomUUID().toString() :
```
yukuai518: Does that mean each time we call createResource against root scope broker, we will have different metricContextName generated? What is the purpose of this?

PR #1688, opened 2017-03-09
Check if delegated `JobListener`s implement `Closeable` and if they do, call `close()`.

kadaan:


gobblin-runtime/src/main/java/gobblin/runtime/listeners/JobListeners.java
```diff
@@ -172,6 +173,25 @@ public void close() throws IOException {
...
+    }
+
+    private void closeJobListeners() throws IOException {
+      IOException exception = null;
+      for (JobListener jobListener : this.jobListeners) {
+        if (jobListener instanceof Closeable) {
+          try {
+            ((Closeable) jobListener).close();
+          } catch (IOException e) {
+            if (exception == null) {
```
abti: nit: add log, also only the last exception in series of exceptions will be cached

kadaan: I'll add a log message, but this code will store the first caught exception and re-throw it once all the listeners are closed. This is the same behavior as not catching the exception, but ensures all listeners are closed.

PR #1690, opened 2017-03-10
Extract modules

eogren:
Refactor the Crypto, Codec, and Metadata concepts out into their own smaller modules so we aren't bloating core and core-base; some of these modules could potentially be included in other projects later.

New modules:

- `gobblin-metadata` contains Metadata related classes
- `gobblin-codecs` contains non-crypto related codecs
- `gobblin-crypto` contains crypto code
- `gobblin-crypto-provider` is a tiny module that links `gobblin-runtime` with `gobblin-crypto`. It's super tiny but exists so that `gobblin-crypto` can retain a very small dependency tree.

`EncryptionFactory` is changed to use `java.util.ServiceLoader` to dynamically register encryption plugins. This is a pattern that probably will be re-used in other cases.

A ton of files have changed, but most of the changes are mechanical:
 - Fixing-up import statements
 - Replace a few hardcoded creations of RotatingAESCodec with factory calls
 - Replaced `@Slf4j` annotations with specific instantiations so we don't need to take a Lombok dependency
 - Replace `SingleRecordIterable` with `Collections.singleton()` so we don't take dependencies on `gobblin-core`


eogren:
@ibuenros @abti @htran1 Any comments you guys have especially on the module structure definitely appreciated.

PR #1692, opened 2017-03-13
Throttle underlying input streams.

ibuenros:
The main point of this PR is that we should throttle underlying streams instead of the streams that Gobblin replication sees. For example, if a replication flow does on-the-fly compression / decompression, we want to limit the bytes per second that are actually going through the network, not the output of compression / decompression.

Few changes:
* Introduce metered streams to measure bytes processed in an input/output stream.
* Introduce throttled input stream that is transparently throttled, and a `StreamThrottler` to create the throttled input streams.
* Added the concept of a BrokerView that allows a higher scoped immutable view of a broker. This is necessary for correct functionality of some factories.
* There is no reason replication should use `FSDataInputStream`, so changed to use `InputStream` (this is a backwards incompatible change, but it was overly restrictive).

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
@@ -17,10 +17,9 @@
 
 package gobblin.data.management.copy.converter;
 
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
-
-import org.apache.hadoop.fs.FSDataInputStream;
 
 import com.google.common.base.Function;
 
@@ -46,9 +45,9 @@ public abstract class DistcpConverter extends Converter<String, String, FileAwar
   }
 
   /**
-   * @return A {@link Function} that transforms the {@link FSDataInputStream} in the {@link FileAwareInputStream}.
+   * @return A {@link Function} that transforms the {@link InputStream} in the {@link FileAwareInputStream}.
    */
-  public abstract Function<FSDataInputStream, FSDataInputStream> inputStreamTransformation();
+  public abstract Function<InputStream, InputStream> inputStreamTransformation();
 
   /**
    * @return A list of extensions that should be removed from the output file name, which will be applied in order.
@@ -76,7 +75,7 @@ public abstract class DistcpConverter extends Converter<String, String, FileAwar
   }
 
   /**
-   * Applies the transformation in {@link #inputStreamTransformation} to the {@link FSDataInputStream} in the
+   * Applies the transformation in {@link #inputStreamTransformation} to the {@link InputStream} in the
    * {@link FileAwareInputStream}.
    */
   @Override
@@ -85,7 +84,7 @@ public abstract class DistcpConverter extends Converter<String, String, FileAwar
 
     modifyExtensionAtDestination(fileAwareInputStream.getFile());
     try {
-      FSDataInputStream newInputStream = inputStreamTransformation().apply(fileAwareInputStream.getInputStream());
+      InputStream newInputStream = inputStreamTransformation().apply(fileAwareInputStream.getInputStream());
       return new SingleRecordIterable<>(new FileAwareInputStream(fileAwareInputStream.getFile(), newInputStream));
     } catch (RuntimeException re) {
       throw new DataConversionException(re);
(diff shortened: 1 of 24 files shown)


ibuenros:

gobblin-utility/src/main/java/gobblin/util/io/ThrottledInputStream.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public synchronized void reset() throws IOException {
+    super.reset();
+    this.prevCount = this.meter.getMeter().getCount();
+  }
+
+  private void blockUntilPermitsAvailable() {
+    try {
+      long currentCount = this.meter.getMeter().getCount();
```
ydai1124: Should this be `count` or `rate`? We want throttling on rate not the total bytes, right?

gobblin-utility/src/main/java/gobblin/util/io/BatchedMeterDecorator.java
```diff
@@ -0,0 +1,70 @@
...
+import gobblin.util.Decorator;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+
+/**
+ * A decorator to a {@link Meter} that batches updates for performance.
+ */
+@NotThreadSafe
+public class BatchedMeterDecorator implements Decorator {
```
ibuenros: Two reasons: it is slightly harder to make it thread-safe (requires some locks), and this is intended as a high performance class, locks / synchronized blocks reduce performance.

gobblin-utility/src/main/java/gobblin/util/io/ThrottledInputStream.java
ibuenros: the throttling is provided by the limiter, not this section of code. The interaction with the limiter is to acquire a permits which correspond to total bytes, not rate.

gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
```diff
@@ -17,6 +17,8 @@
 
 package gobblin.data.management.copy;
 
+import java.io.InputStream;
+
 import lombok.AllArgsConstructor;
 import lombok.Getter;
```
ibuenros: The import is not there (it was changed for the `InputStream` import.

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
```diff
@@ -17,6 +17,7 @@
 
 package gobblin.data.management.copy.converter;
 
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
```
ibuenros: Same as above.

gobblin-data-management/src/main/java/gobblin/data/management/copy/FileAwareInputStream.java
ibuenros: Fixed

gobblin-data-management/src/main/java/gobblin/data/management/copy/converter/DistcpConverter.java
ibuenros: Fixed

gobblin-utility/src/main/java/gobblin/util/io/MeteredInputStream.java
```diff
@@ -0,0 +1,93 @@
...
+  public static Optional<MeteredInputStream> findWrappedMeteredInputStream(InputStream is) {
+    if (is instanceof FilterInputStream) {
+      try {
+        Optional<MeteredInputStream> meteredInputStream =
+            findWrappedMeteredInputStream(FilterStreamUnpacker.unpackFilterInputStream((FilterInputStream) is));
+        if (meteredInputStream.isPresent()) {
+          return meteredInputStream;
+        }
+      } catch (IllegalAccessException iae) {
+        log.warn("Cannot unpack input stream due to SecurityManager.");
```
ibuenros: Fixed

gobblin-utility/src/main/java/gobblin/util/io/MeteredStream.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+/**
+ * A {@link java.io.InputStream} or {@link java.io.OutputStream} that measures bytes processed.
+ */
+public interface MeteredStream {
+  /**
+   * @return The {@link Meter} measuring the bytes processed.
+   */
+  Meter getMeter();
```
ibuenros: Fixed

PR #1698, opened 2017-03-15
Added restore to gobblin-compliance

aditya1105:
Added restore with two restore policies.
Added skeleton for compliance validation.

### DIFF
--- a/gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/ComplianceRestoreJob.java
+++ b/gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/ComplianceRestoreJob.java
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package gobblin.compliance.restore;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.thrift.TException;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
+
+import gobblin.compliance.ComplianceConfigurationKeys;
+import gobblin.compliance.ComplianceEvents;
+import gobblin.compliance.ComplianceJob;
+import gobblin.compliance.utils.ProxyUtils;
+import gobblin.configuration.State;
+import gobblin.dataset.Dataset;
+import gobblin.dataset.DatasetsFinder;
+import gobblin.util.ExecutorsUtils;
+import gobblin.util.reflection.GobblinConstructorUtils;
+
+
+/**
+ * Job for restoring {@link gobblin.compliance.HivePartitionDataset} from the backups
+ *
+ * @author adsharma
+ */
+@Slf4j
+public class ComplianceRestoreJob extends ComplianceJob {
+
+  public ComplianceRestoreJob(Properties properties) {
+    super(properties);
+    initDatasetFinder(properties);
+    try {
+      ProxyUtils.cancelTokens(new State(properties));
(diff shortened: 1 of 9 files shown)


gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/LKGRestorePolicy.java
```diff
@@ -0,0 +1,96 @@
...
+    List<HivePartitionVersion> nonRestorableVersions = new ArrayList<>();
+
+    for (HivePartitionVersion version : versions) {
+      if (!isRestorable(dataset, version)) {
+        nonRestorableVersions.add(version);
+      }
+    }
+    versions.removeAll(nonRestorableVersions);
+    Preconditions.checkArgument(!versions.isEmpty(), "No versions to restore dataset " + dataset.datasetURN());
+    Collections.sort(versions);
```
htran1: Noticed that the version compare does a computation with current time. This means the computation can drift. Why not compare using absolution time instead of relative time?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/ComplianceRestoreJob.java
```diff
@@ -0,0 +1,144 @@
...
+import gobblin.util.reflection.GobblinConstructorUtils;
+
+
+/**
+ * Job for restoring {@link gobblin.compliance.HivePartitionDataset} from the backups
+ *
+ * @author adsharma
+ */
+@Slf4j
+public class ComplianceRestoreJob extends ComplianceJob {
```
htran1: This looks almost the same as ComplianceRetentionJob. Would it make sense to share the code through another class?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDataset.java
```diff
@@ -0,0 +1,183 @@
...
+        GobblinConstructorUtils.invokeConstructor(HivePartitionRestorePolicy.class, restorePolicyClass, this.state);
+    try {
+      this.datasetToRestore = (HivePartitionDataset) this.restorePolicy.getDatasetToRestore(this);
+      log.info("Found dataset to restore with " + this.datasetToRestore.datasetURN());
+    } catch (IOException e) {
+      Throwables.propagate(e);
+    }
+    this.datasetOwner = getOwner();
+    this.datasetToRestoreOwner = this.datasetToRestore.getOwner();
+    Preconditions.checkArgument(this.state.contains(ComplianceConfigurationKeys.TRASH_OWNER),
```
htran1: Why not do this earlier before the call to getDatasetToRestore?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDataset.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private String getCompleteTableName() {
+    return StringUtils.join(Arrays.asList(getDbName(), getTableName()), ComplianceConfigurationKeys.DBNAME_SEPARATOR);
+  }
+
+  private String getTrashTableLocation() {
+    Preconditions.checkArgument(this.state.contains(ComplianceConfigurationKeys.TRASH_DIR),
+        "Missing required property " + ComplianceConfigurationKeys.TRASH_DIR);
+    return this.state.getProp(ComplianceConfigurationKeys.TRASH_DIR) + getCompleteTableName();
```
ydai1124: Is it better to use `new Path(this.state.getProp(ComplianceConfigurationKeys.TRASH_DIR), getCompleteTableName())`?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDatasetFinder.java
```diff
@@ -0,0 +1,69 @@
...
+import gobblin.configuration.State;
+import gobblin.util.WriterUtils;
+
+
+/**
+ * A finder class for finding Restorable {@link HivePartitionDataset}
+ *
+ * @author adsharma
+ */
+@Slf4j
```
ydai1124: Is this used?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDataset.java
aditya1105: getTrashTableLocation is used at other places where String is required.

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDatasetFinder.java
aditya1105: Yes. It is a type of dataset finder specifically for restore

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/LKGRestorePolicy.java
aditya1105: You are right,  I will compare using timestamp

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/RestorableHivePartitionDataset.java
aditya1105: to get datasetToRestoreOwner, we need datasetToRestore. I am reordering others.

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/restore/ComplianceRestoreJob.java
aditya1105: Yes. I tried to abstract out common parts in ComplianceJob which is being used at multiple places.

PR #1701, opened 2017-03-16
Functionality for record-by-record pipelines / extraction of code

eogren:
There are a lot of lines changed, but almost all of this is moving files around / refactoring common functionality into their own place

- Extract a bunch of logic that was in FsDataWriter out to a GlobalMetadataCollector object which can be re-used in other places
- Create a MetadataWriterWrapper that surrounds other writer objects to collect metadata records that flow through
- Create a few Metadata wrapper for existing converters
- Fix a config issue where both the Encryption converter and the SimpleDataWriter were looking at writer.encrypt.* to figure out if they should do anything or not; change the converter class to look at a different set of config


gobblin-example/src/main/resources/avro-eventhub.job
```diff
@@ -3,13 +3,13 @@ gobblin.workDir=gobblin
 
 launcher.type=LOCAL
 
-
+# FIXME!
```
eogren: Ah, this was because a converter class name changed and I wanted to update this file.

gobblin-modules/gobblin-metadata/src/main/java/gobblin/metadata/types/GlobalMetadata.java
```diff
@@ -41,22 +45,44 @@
...
 
   @JsonProperty("file")
   private final Map<String, Map<String, Object>> fileLevel;
 
+  @JsonProperty("id")
+  private String cachedId;
+
+  private transient boolean markedImmutable;
+
   private final static String DATASET_URN_KEY = "Dataset-URN";
```
eogren: Hmm, I can stash them in an enum instead to make the code a little cleaner. What do you think adding a new explicit MetadataName type does? Possible extension / self-documentation later?

gobblin-core/src/main/java/gobblin/writer/MetadataAwareWriter.java
```diff
@@ -20,7 +20,12 @@
...
- * Represents a Writer that is metadata aware
+ * Represents a Writer that is metadata aware. Metadata aware writers
+ * specify default {@link GlobalMetadata} that will be added to all records that flow
+ * through the pipeline they are a part of.
+ *
+ * This allows a writer to communicate with the MetadataWriterWrapper that surrounds it
+ * and is the one responsible for processing any metadata that flows through the conversion pipeline.
  */
 public interface MetadataAwareWriter {
   GlobalMetadata getDefaultMetadata();
```
shirshanka: Can you add a comment on the getDefaultMetadata method? 

What does default metadata mean?  Is it per dataset? 
Should the caller call this once and cache or call multiple times..

PR #1702, opened 2017-03-17
Enable partitioning based on source info + partitioned metadata writes

eogren:
@ibuenros @shirshanka 

- Fix an issue in DatePartitionedAvroFileSource where the code would not
resolve paths properly if the prefix config property was empty
- Enhance the AvroFileSource to send along the partition it extracted
in WorkUnitState
- Create a WorkUnitStateWriterPartitioner that will partition an
incoming record solely based on WorkUnitState
- Update writers to set any partition paths they use in an internal
PARTITION_PATH_KEY
- Update BaseDataPublisher to collect partition information and publish
metadata files in each partition directory if the metadata output
directory is not specifically configured somewhere else.
- Update BaseDataPublisher to publish metadata _after_ publishing data.
If this isn't done - and partitions are enabled - then the code that
publishes all of the data files ends up publishing them as
subdirectories into the partition path. (Eg you get
/out/2017-01-01/2017-01-01/data.json).

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
+++ b/gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
@@ -62,57 +62,57 @@ import gobblin.source.workunit.WorkUnit;
 /**
  * Implementation of {@link gobblin.source.Source} that reads over date-partitioned Avro data.
  * This source can be regarded as the reader equivalent of {@link TimeBasedAvroWriterPartitioner}.
- * 
+ *
  * <p>
  * The class will iterate through all the data folders given by the base directory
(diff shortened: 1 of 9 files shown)


gobblin-core/src/main/java/gobblin/source/DatePartitionedAvroFileSource.java
```diff
@@ -62,57 +62,57 @@
 /**
  * Implementation of {@link gobblin.source.Source} that reads over date-partitioned Avro data.
  * This source can be regarded as the reader equivalent of {@link TimeBasedAvroWriterPartitioner}.
- *
```
eogren: Sorry for all of these junk whitespace changes, I couldn't figure out how to make IntelliJ not make these updates

PR #1703, opened 2017-03-17
Added validation core files to gobblin-compliance

aditya1105:


gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/validation/ValidableDataset.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.IOException;
+
+import gobblin.dataset.Dataset;
+
+
+/**
+ * @author adsharma
+ */
+public interface ValidableDataset extends Dataset {
```
htran1: Should the name be ValidatableDataset?

PR #1708, opened 2017-03-17
Refactor RestApiConnector and RestApiExtractor

howu:
To extend RestApiConnector and RestApiExtractor for other data source, some methods need to be able to override

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/extract/restapi/RestApiExtractor.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/extract/restapi/RestApiExtractor.java
@@ -73,6 +73,27 @@ public abstract class RestApiExtractor extends QueryBasedExtractor<JsonArray, Js
 
   protected abstract RestApiConnector getConnector(WorkUnitState state);
 
+  protected void buildQuery(String inputQuery, String entity, JsonArray columnArray) {
+    if (inputQuery == null && this.columnList.size() != 0) {
+      // if input query is null, build the query from metadata
+      this.updatedQuery = "SELECT " + Joiner.on(",").join(this.columnList) + " FROM " + entity;
+    } else {
+      // if input query is not null, build the query with intersection of columns from input query and columns from Metadata
+      if (inputQuery != null) {
+        String queryLowerCase = inputQuery.toLowerCase();
+        int columnsStartIndex = queryLowerCase.indexOf("select ") + 7;
+        int columnsEndIndex = queryLowerCase.indexOf(" from ");
+        if (columnsStartIndex > 0 && columnsEndIndex > 0) {
+          String givenColumnList = inputQuery.substring(columnsStartIndex, columnsEndIndex);
+          this.updatedQuery = inputQuery.replace(givenColumnList, Joiner.on(",").join(this.columnList));
+        } else {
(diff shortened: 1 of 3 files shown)


ibuenros:
@zxcware can you take a look?

gobblin-core/src/main/java/gobblin/source/extractor/extract/restapi/RestApiExtractor.java
```diff
@@ -73,6 +73,27 @@ public RestApiExtractor(WorkUnitState state) {
 
   protected abstract RestApiConnector getConnector(WorkUnitState state);
 
+  protected void buildQuery(String inputQuery, String entity, JsonArray columnArray) {
```
zxcware: The name is too general, which makes it a function to build any kind of query. Some name like `buildExtractMetadataQuery` is better.  

It's also better to return the query string instead of setting `updatedQuery` inside the function. The value assignment can be done in `extractMetadata`.

zxcware: `columnArray` is not used in the function. Remove? I guess what you want is actually `List<String> columnList`.

gobblin-core/src/main/java/gobblin/source/extractor/extract/restapi/RestApiExtractor.java
```diff
@@ -73,6 +73,27 @@ public RestApiExtractor(WorkUnitState state) {
...
+        int columnsEndIndex = queryLowerCase.indexOf(" from ");
+        if (columnsStartIndex > 0 && columnsEndIndex > 0) {
+          String givenColumnList = inputQuery.substring(columnsStartIndex, columnsEndIndex);
+          this.updatedQuery = inputQuery.replace(givenColumnList, Joiner.on(",").join(this.columnList));
+        } else {
+          this.updatedQuery = inputQuery;
+        }
+      }
+    }
+    log.info("Updated input query: " + this.updatedQuery);
```
zxcware: The log message can be more specific. Something like `Extract metadata query: `

gobblin-core/src/main/java/gobblin/source/extractor/extract/restapi/RestApiExtractor.java
howu: Fixed by renaming `buildQuery()` to `buildDataQuery()` and change return type to `String`
I don't know why `updatedQuery` was named in this way.
By tracing code, `updatedQuery` is used in `getDataMetadata()`, which will prepare query that is used to call GET command and get data from source.
Therefore I rename to data query in both method name and log.
However I can't rename `updatedQuery` to  `dataQuery` since it will affect subclass of `RestApiExtractor`

howu: fixed by removing 3rd parameter `columnArray` in `buildDataQuery()`

howu:
@ibuenros 
Test failed in Travis CI but I don't think that relates to my code.
Do I need to do something to trigger CI again?

zxcware:
@howu Yes. Try triggering CI again.

htran1:
The travis hang does not look related to the code changes and is happening in other PRs.

PR #1709, opened 2017-03-18
Can run custom tasks in Gobblin.

ibuenros:
This PR allows running custom tasks in Gobblin. 

To use custom tasks, implement a `TaskFactory` and use `TaskUtils.setTaskFactoryClass(WorkUnitState, Class<? extends TaskFactory>` in the `Source`. Tasks need to implement `TaskIFace`.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
+++ b/gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
@@ -17,14 +17,11 @@
 
 package gobblin.runtime;
 
-import java.io.IOException;
-import java.util.Collection;
-import java.util.concurrent.Callable;
-
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
 import com.google.common.io.Closer;
-
 import gobblin.commit.CommitSequence;
 import gobblin.commit.CommitStep;
 import gobblin.commit.DeliverySemantics;
@@ -34,9 +31,16 @@ import gobblin.publisher.CommitSequencePublisher;
 import gobblin.publisher.DataPublisher;
 import gobblin.publisher.UnpublishedHandling;
 import gobblin.runtime.commit.DatasetStateCommitStep;
+import gobblin.runtime.task.TaskFactory;
+import gobblin.runtime.task.TaskUtils;
 import gobblin.source.extractor.JobCommitPolicy;
-
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
 import lombok.AllArgsConstructor;
+import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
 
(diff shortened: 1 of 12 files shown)


ibuenros:
@ydai1124 @htran1 can you take a look?

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
```diff
@@ -129,27 +150,62 @@ public Void call()
...
+  }
+
+  @Data
+  private static class TaskFactoryWrapper {
+    private final TaskFactory taskFactory;
 
-    // Set the dataset state to COMMITTED upon successful commit
-    datasetState.setState(JobState.RunningState.COMMITTED);
+    public boolean equals(Object other) {
+      if (other instanceof TaskFactoryWrapper) {
```
ydai1124: call `canEqual` instead?

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -141,6 +143,29 @@
   private final AtomicBoolean shutdownRequested;
   private final CountDownLatch shutdownLatch;
 
+  public Task() {
```
ydai1124: You can use `@NoArgsConstructor(force = true)`

gobblin-runtime/src/main/java/gobblin/runtime/task/TaskIFaceWrapper.java
```diff
@@ -0,0 +1,169 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gobblin.runtime.task;
+
+import com.google.common.base.Optional;
```
ydai1124: import order.

gobblin-runtime/src/main/java/gobblin/runtime/SafeDatasetCommit.java
```diff
@@ -129,27 +150,62 @@ public Void call()
...
 
   /**
-   * Synchronized version of {@link #commitDataset(JobState.DatasetState, DataPublisher)} used when publisher is not
+   * Synchronized version of {@link #commitDataset(Collection, DataPublisher)} used when publisher is not
    * thread safe.
    */
-  private synchronized void safeCommitDataset(JobState.DatasetState datasetState, DataPublisher publisher) {
-    commitDataset(datasetState, publisher);
+  private void safeCommitDataset(Collection<TaskState> taskStates, DataPublisher publisher) {
+    synchronized (GLOBAL_LOCK) {
```
ibuenros: Publishers are not necessarily thread safe. In line 108 we check whether the publisher is thread-safe, if it is we don't use syncronization, but if it isn't we do.

Previously the synchronization was mistakenly done only within a single dataset, which is a mistake. This just ensures global synchronization when the publisher is not thread safe.

ibuenros:
@ydai1124 thanks for the comments. Addressed them.

PR #1713, opened 2017-03-21
Add metrics for certain operations in instrumentedHDFSFilesystem

yukuai518:


gobblin-core/src/main/java/gobblin/filesystem/InstrumentedHDFSFileSystem.java
```diff
@@ -75,4 +147,143 @@ public void close()
     super.close();
     // Should print out statistics here
   }
+
+  /**
+   * Add timer metrics to {@link DistributedFileSystem#mkdir(Path, FsPermission)}
+   */
+  public boolean mkdir(Path f, FsPermission permission) throws IOException {
+    try (Closeable context = new TimerContextWithLog(mkdirTimer.time(), "mkdir", f, permission)) {
```
ibuenros: nit: can simplify this as
```java
try (Closeable context = new TimerContextWithLog(mkdirTimer.time(), "mkdir", f, permission)) {
   return super.mkdir(f, permission);
}
```

Same for all the other methods.

gobblin-core/src/main/java/gobblin/filesystem/InstrumentedHDFSFileSystem.java
```diff
@@ -75,4 +147,143 @@ public void close()
...
+    } catch (IOException e) {
+      throw e;
+    }
+  }
+
+  /**
+   * Indirect call to {@link InstrumentedHDFSFileSystem#mkdirs(Path, FsPermission)}
+   */
+  public boolean mkdirs(Path f) throws IOException {
+    return super.mkdirs(f);
```
ibuenros: Please remove methods that just call `super`.

gobblin-core/src/main/java/gobblin/filesystem/InstrumentedHDFSFileSystem.java
```diff
@@ -40,8 +53,67 @@
...
-
+  public static final String HDFS_METRIC_CONTEXT_NAME = "hdfsMetricContext";
   private static final String HDFS_SCHEME = "hdfs";
+  private MetricContext metricContext;
+
+  protected final Closer closer;
+
+  // Below are HDFS metrics
+  @VisibleForTesting
+  protected final Timer listStatusPathTimer;
```
ibuenros: I think all of the `listStatus` type calls should use the same timer.

yukuai518: There are a few listStatus may duplicate the timer count. 
listStatus(path[]) -> listStatus(path[], filter) ---> listStatus(path)
listStatus(path, filter) ________________________/

If we only care about the final operation listStatus(path), then only listStatusPathTimer should be inspected.

PR #1715, opened 2017-03-21
More metrics on batch permit requester.

ibuenros:


ibuenros:

PR #1717, opened 2017-03-21
Streamed sources implementation.

ibuenros:


ibuenros:
@ydai1124 can you review?

gobblin-api/src/main/java/gobblin/source/StreamedSource.java
```diff
@@ -0,0 +1,39 @@
...
+ * A {@link Source} that produces a {@link WorkUnitStream}.
+ *
+ * Streamed sources may offer certain advantages over non-streamed sources if the job launcher supports them:
+ *  * Low memory usage: If the {@link WorkUnitStream} is backed by a generator, the job launcher may optimize memory
+ *    usage by never materializing all work units in memory.
+ *  * Eager processing of slow sources: If the source is slow at producing work units, the job launcher may start
+ *    processing available work units while future work units are still being computed.
+ *  * Infinite work unit streams: in the future, some job launchers will support infinite streams of work units.
+ */
+public interface StreamedSource<S, D> extends Source<S, D> {
```
shirshanka: This is an overloaded and confusing term... need to think of a better name... source that produces stream of work-units versus a source of streaming data...

ibuenros:
@shirshanka addressed the comments around using object attributes instead of instance checks. Also, do you have a suggestion on a better name for `StreamedSource`?

gobblin-api/src/main/java/gobblin/source/workunit/WorkUnitStream.java
```diff
@@ -0,0 +1,144 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+
+/**
+ * A stream of {@link WorkUnit}s, allows for working with large numbers of work units in a memory-efficient way, as
+ * well as processing infinite streams of work units.
+ */
+public class WorkUnitStream {
```
shirshanka: Given that this will be a pretty important Type for all Sources to implement (eventually), should it be an interface?

ibuenros:
@shirshanka made the `WorkUnitStream` an interface.

gobblin-runtime/src/test/java/gobblin/runtime/TestStreamedSource.java
```diff
@@ -0,0 +1,191 @@
...
+import gobblin.source.workunit.WorkUnit;
+import gobblin.source.workunit.WorkUnitStream;
+import gobblin.task.EventBusPublishingTaskFactory;
+import gobblin.writer.test.TestingEventBuses;
+
+import lombok.extern.slf4j.Slf4j;
+
+
+@Slf4j
+public class TestStreamedSource {
```
shirshanka: For consistency, this should be TestWorkUnitStreamSource

PR #1720, opened 2017-03-22
Add field Rename to GobblinTrackingEvents converter

ydai1124:
@htran1 can you take a look?


ydai1124:
@htran1 Fixed. Thanks!

PR #1721, opened 2017-03-23
Add batch list size upper bound to prevent oom issue in eventhub datawriter

yukuai518:


gobblin-modules/gobblin-eventhub/src/main/java/gobblin/eventhub/writer/EventhubBatchAccumulator.java
```diff
@@ -163,28 +184,68 @@ public void remove(Batch batch) {
...
-        if (dq.size() == 0)
-            return false;
-        EventhubBatch first = dq.peekFirst();
-        if (first.isTTLExpire()) {
-            LOG.info ("Batch " + first.getId() + " is expired");
-            return true;
-        }
-
-        return false;
+        return true;
```
htran1: This is kind of strange to always return true and seems hacky.

gobblin-core-base/src/main/java/gobblin/writer/BatchAccumulator.java
```diff
@@ -91,16 +92,19 @@ public boolean isClosed () {
    * When close is invoked, all new coming records will be rejected
    * Add a busy loop here to ensure all the ongoing appends are completed
    */
-  public final void close () {
-    closed = true;
+  public void close () {
     while (appendsInProgress.get() > 0) {
       LOG.info("Append is still going on, wait for a while");
```
htran1: Should consider adding a small sleep here like 100 milliseconds.

PR #1722, opened 2017-03-24
Add runtime table properties(e.g. Kafka topic) into Hive Registration

autumnust:
Adding upstream data src(specifically "topic.name" for kafka data in this PR, temporarily) info into registered hive table, as one of the table properties.

@ibuenros @htran1  Could you help review ? Thanks.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
@@ -17,8 +17,6 @@
 
 package gobblin.publisher;
 
-import com.google.common.base.Splitter;
-import gobblin.hive.metastore.HiveMetaStoreUtils;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Set;
@@ -33,6 +31,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.fs.Path;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Splitter;
 import com.google.common.collect.Sets;
 import com.google.common.io.Closer;
 
@@ -43,6 +42,7 @@ import gobblin.hive.HiveRegProps;
 import gobblin.hive.HiveRegister;
 import gobblin.hive.policy.HiveRegistrationPolicy;
 import gobblin.hive.policy.HiveRegistrationPolicyBase;
+import gobblin.hive.metastore.HiveMetaStoreUtils;
 import gobblin.hive.spec.HiveSpec;
 import gobblin.util.ExecutorsUtils;
 
@@ -103,7 +103,7 @@ public class HiveRegistrationPublisher extends DataPublisher {
     // runtime.props are comma-separated props collected in runtime.
     Set<String> pathsToRegisterFromSingleState = Sets.newHashSet() ;
     for (State state:states) {
-      State taskSpecificState = super.state;
+      State taskSpecificState = state;
       if (state.contains(ConfigurationKeys.PUBLISHER_DIRS)) {
 
         // Upstream data attribute is specified, need to inject these info into superState as runtimeTableProps.
--- a/gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
@@ -17,6 +17,8 @@
 
 package gobblin.publisher;
 
+import com.google.common.base.Splitter;
+import gobblin.hive.metastore.HiveMetaStoreUtils;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Set;
@@ -63,6 +65,7 @@ import lombok.extern.slf4j.Slf4j;
 public class HiveRegistrationPublisher extends DataPublisher {
 
   private static final String DATA_PUBLISH_TIME = HiveRegistrationPublisher.class.getName() + ".lastDataPublishTime";
+  private static final Splitter LIST_SPLITTER_COMMA = Splitter.on(",").trimResults().omitEmptyStrings();
   private final Closer closer = Closer.create();
   private final HiveRegister hiveRegister;
   private final ExecutorService hivePolicyExecutor;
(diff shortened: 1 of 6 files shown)


gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -92,23 +92,29 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
-    for (final String path : pathsToRegister) {
-      completionService.submit(new Callable<Collection<HiveSpec>>() {
-
-        @Override
-        public Collection<HiveSpec> call() throws Exception {
-          return policy.getHiveSpecs(new Path(path));
+    // Each state in states is task-level State, while superState is the Job-level State.
+    // Using both State objects to distinguish each HiveRegistrationPolicy so that
+    // they can carry task-level information to pass into Hive Partition and its corresponding Hive Table.
+    for (State state:states) {
```
ibuenros: I don't understand what this change is doing. It seems it is doing exactly the same as before, except not deduplicating the paths to register?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -107,6 +107,10 @@ public static Table getTable(HiveTable hiveTable) {
     if (table.getTableType().equals(TableType.EXTERNAL_TABLE.toString())) {
       table.getParameters().put(EXTERNAL, Boolean.TRUE.toString().toUpperCase());
     }
+    if (hiveTable.getProps().contains(HiveConstants.TOPIC_NAME)) {
```
ibuenros: Please don't add such over-specific logic into a more generic class.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicy.java
```diff
@@ -17,14 +17,15 @@
...
 
+
 import java.io.IOException;
 import java.util.Collection;
 
 import org.apache.hadoop.fs.Path;
 
 import gobblin.annotation.Alpha;
 import gobblin.hive.spec.HiveSpec;
-
```
ibuenros: Is this needed?

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -308,7 +308,14 @@ protected HiveTable getTable(Path path, String dbName, String tableName) throws
 
     table.setLocation(this.fs.makeQualified(getTableLocation(path)).toString());
     table.setSerDeProps(path);
-    table.setProps(this.props.getTablePartitionProps());
+
+    // Setting table-level props.
+    State tableProps = new State(this.props.getTablePartitionProps());
+    if (this.props.contains(HiveConstants.TOPIC_NAME)) {
```
ibuenros: Don't use specific logic in generic classes.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
```diff
@@ -373,4 +380,36 @@ public static HiveRegistrationPolicy getPolicy(State props) {
           "Unable to instantiate " + HiveRegistrationPolicy.class.getSimpleName() + " with type " + policyType, e);
     }
   }
+
+  /**
+   *
+   * @param props Job-level state, props A {@link State} object that contains property,
+   * {@link #HIVE_REGISTRATION_POLICY}, which is the class name of the desired policy.
+   * This policy class must have a constructor that takes a {@link State} object.
+   * @param runtimeExtraState There are occasions where some runtime attributes need to be added into Hive Table's
```
ibuenros: Please rephrase this doc, it is hard to understand. Is this parameter needed though? Can't both set of properties be combined in one.

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
autumnust: Previously it generates all Paths to register through one function call and all Paths are registered based on a single policy. 
The code change here is trying to iterate through all states and create different policy objects for state, so that different policy object can carry the topic information for each path to register.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicyBase.java
autumnust: This method will be removed.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicy.java
autumnust: Removed.

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -92,23 +93,38 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+          superState.appendToListProp(HiveConstants.RUNTIME_PROPS,
+              HiveConstants.TOPIC_NAME + ":" + state.getProp(HiveConstants.TOPIC_NAME));
         }
-      });
-
+        final HiveRegistrationPolicy policy = HiveRegistrationPolicyBase.getPolicy(superState);
+        Set<String> pathsToRegisterFromSingleState = Sets.newHashSet() ;
+        pathsToRegisterFromSingleState.addAll(state.getPropAsList(ConfigurationKeys.PUBLISHER_DIRS));
+        for ( final String path : pathsToRegisterFromSingleState ) {
+          completionService.submit(new Callable<Collection<HiveSpec>>() {
```
ibuenros: What happens if multiple states have the same `PUBLISHER_DIR`?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -107,6 +109,15 @@ public static Table getTable(HiveTable hiveTable) {
     if (table.getTableType().equals(TableType.EXTERNAL_TABLE.toString())) {
       table.getParameters().put(EXTERNAL, Boolean.TRUE.toString().toUpperCase());
     }
+    if (hiveTable.getProps().contains(HiveConstants.RUNTIME_PROPS)) {
+      String runtimePropsString = hiveTable.getProps().getProp(HiveConstants.RUNTIME_PROPS);
+      for (String propValue : LIST_SPLITTER.splitToList(runtimePropsString)) {
+        List<String> tokens = LIST_SPLITTER.splitToList(propValue);
```
ibuenros: I am probably not getting something here. If `runtimePropsString` was already split, it seems `propValue` would not contain any commas?

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
```diff
@@ -54,6 +54,8 @@
   private final State storageProps;
   private final State serdeProps;
 
+  private Optional<String> runtimeExtraTableProps;
```
ibuenros: What is the justification for this name?

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -92,23 +93,38 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+    // they can carry task-level information to pass into Hive Partition and its corresponding Hive Table.
 
-        @Override
-        public Collection<HiveSpec> call() throws Exception {
-          return policy.getHiveSpecs(new Path(path));
+    // Here all runtime task-level props are injected into superstate which installed in each Policy Object.
+    // runtime.props are comma-separated props collected in runtime.
+    for (State state:states) {
+      if (state.contains(ConfigurationKeys.PUBLISHER_DIRS)) {
+        if (state.contains(HiveConstants.TOPIC_NAME)) {
```
ibuenros: This is still overly specific. Any reference to TOPIC should only be on the Kafka side of things.

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
autumnust: Yes. Wrong splitter used here.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegProps.java
autumnust: There's already tableProps exists which represent the static table properties obtained from .job file, if any. 

Here there are some table props obtained from runtime. May be `runtimeTableProps` is better?

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
autumnust: `pathsToRegisterFromSingleState` is a set no should be no duplication.

ibuenros: But this is a state local variable, it will not deduplicate across states.

autumnust: Yes, definitely a bug here.  Should add one more global path set to deduplicate.

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -92,23 +95,47 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+              return policy.getHiveSpecs(new Path(path));
+            }
+          });
+        }
+      }
+      else continue;
     }
-
-    for (int i = 0; i < pathsToRegister.size(); i++) {
+    for (int i = 0; i < numberOfPathsToRegister; i++) {
```
yukuai518: I guess you don't need to calculate count here.

gobblin-hive-registration/src/main/java/gobblin/hive/policy/HiveRegistrationPolicy.java
```diff
@@ -25,7 +25,6 @@
 import gobblin.annotation.Alpha;
 import gobblin.hive.spec.HiveSpec;
 
-
```
yukuai518: You don't need to include these files without changes.

gobblin-core/src/main/java/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -92,23 +95,46 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+    Set<String> pathsToRegisterFromSingleState = Sets.newHashSet() ;
+    for (State state:states) {
+      if (state.contains(ConfigurationKeys.PUBLISHER_DIRS)) {
+
+        // Upstream data attribute is specified, need to inject these info into superState as runtimeTableProps.
+        if (this.hiveRegister.getProps().getUpstreamDataAttrName().isPresent()) {
+          for (String attrName:
+              LIST_SPLITTER_COMMA.splitToList(this.hiveRegister.getProps().getUpstreamDataAttrName().get())){
+            if (state.contains(attrName)) {
+              superState.appendToListProp(HiveMetaStoreUtils.RUNTIME_PROPS,
```
ibuenros: It seems you are mutating common state here? It is better to create a local copy of the state.

ibuenros: Actually, doesn't the `WorkUnitState` contain the job state already? Why can't you just use the `WorkUnitState`?

autumnust: `State` here comes from constructor and `WorkUnitState` comes from `publishData`'s parameters. Is that necessary they both contain job state? Is the task state automatically inherit all props from job state?

PR #1726, opened 2017-03-27
Throttling server refuses extremely large requests, can provide fewer…

ibuenros:
… than requested permits.

This PR improves the throttling server:
* Previously, large requests would just block for a long time, eventually timing out on the client. This is obviously not useful.
* Introduced a few heuristics to the server:
** If a request cannot be satisfied within timeout, return immediately.
** If a request has a minimum useful permits lower than the ideal requested permits, and the ideal requested permits cannot be satisfied quickly, try to return a lower number of permits.
** If the throttler is under-utilized, e.g. there are a lot of stored permits, then it may return more permits than requested.
* Since Guava `RateLimiter` does not allow extension or fine control, implemented a custom `TokenBucket` algorithm.
* No longer use straight `Limiter`s in the server, as the needs are different than in a client. Instead, use `ThrottlingPolicies` (which may or may not wrap a `Limiter`).
* Throttling client sends an ideal and minimum number of requested permits.

### DIFF
--- a/gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-server/src/main/java/gobblin/restli/throttling/DynamicTokenBucket.java
+++ b/gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-server/src/main/java/gobblin/restli/throttling/DynamicTokenBucket.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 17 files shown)


ibuenros:

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-server/src/main/java/gobblin/restli/throttling/DynamicTokenBucket.java
```diff
@@ -0,0 +1,83 @@
...
+   * @param timeoutMillis the maximum wait the calling process is willing to wait for tokens.
+   * @return the number of allocated tokens.
+   */
+  public long getPermits(long requestedPermits, long minPermits, long timeoutMillis) {
+
+    try {
+      long storedTokens = this.tokenBucket.getStoredTokens();
+
+      long eagerTokens = storedTokens / 2;
+      if (eagerTokens > requestedPermits && this.tokenBucket.getTokens(storedTokens, 0, TimeUnit.MILLISECONDS)) {
```
htran1: How come the get is for storedTokens, but eagerToken is returned?

PR #1728, opened 2017-03-27
Added fail-fast option to CopyableDatasetRequestor (disabled by default)

jinhyukchang:
Currently when there's failure getting dataset in CopyableDatasetRequestor, it logs the error and continue to perform distcp and makes job status success overall. 
This change will have user to choose to make the job fail fast so that failure from CopyableDatasetRequestor will make the job fail immediately.

- Integration test passed.

job property snippet:
gobblin.copy.CopyableDatasetRequestor.failFast=true

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/partition/CopyableDatasetRequestor.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/partition/CopyableDatasetRequestor.java
@@ -53,9 +53,6 @@ import lombok.extern.slf4j.Slf4j;
 @AllArgsConstructor
 @Getter
 public class CopyableDatasetRequestor implements PushDownRequestor<FileSet<CopyEntity>> {
-  private final static String CONF_PREFIX = CopyConfiguration.COPY_PREFIX + "." + CopyableDatasetRequestor.class.getSimpleName();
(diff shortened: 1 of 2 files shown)


jinhyukchang:
@ibuenros Could you take a look when you have a chance?

gobblin-data-management/src/main/java/gobblin/data/management/partition/CopyableDatasetRequestor.java
```diff
@@ -52,6 +53,8 @@
 @AllArgsConstructor
 @Getter
 public class CopyableDatasetRequestor implements PushDownRequestor<FileSet<CopyEntity>> {
+  private final static String CONF_PREFIX = CopyConfiguration.COPY_PREFIX + "." + CopyableDatasetRequestor.class.getSimpleName();
```
ibuenros: Can you put the configuration key in `CopyConfiguration`? Also, a more descriptive and less specific key might be better. A suggestion might be `gobblin.copy.abortOnSingleDatasetFailure`.

jinhyukchang: Will do.

jinhyukchang:
@ibuenros Thanks for the review and I've addressed PR comments.

PR #1729, opened 2017-03-28
Enable KafKaSource to accpet table type and namespace configuration

zxcware:
- Accept configured table type and namespace instead of fixed values
- Add configuration key `gobblin.kafka.extract.allowTableTypeAndNamspaceCustomization` to turn on the changes, avoid breaking existing jobs


gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -92,6 +92,8 @@
   public static final String AVG_RECORD_SIZE = "avg.record.size";
   public static final String AVG_RECORD_MILLIS = "avg.record.millis";
   public static final String GOBBLIN_KAFKA_CONSUMER_CLIENT_FACTORY_CLASS = "gobblin.kafka.consumerClient.class";
+  public static final String GOBBLIN_KAFKA_EXTRACT_ALLOW_TABLE_TYPE_NAMESPACE_CUSTOMIZATION =
```
ibuenros: Is this needed?

zxcware: Yes. I described the reasons in detail in the solution document. The main point is to avoid breaking existing kafka jobs, which specify the `extract.namespace` configuration but it's ignored by the current KafkaSource.

PR #1735, opened 2017-03-29
Add one more option for distcping Hive-Registered table 

autumnust:
Previously the only flow is : Copy the data into new Path on filesystem, deregister the original hive table and create a new one.  

Add one more option for the case where table deregistration is not preferable, but change the corresponding data location on filesystem only.  

Status: Require validation from Hive side, e.g. Changing the permission/privilege in the original table, run the new flow, see whether the hive table still keep those permission/privilege setting as before, although the data location has been updated.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -171,8 +171,6 @@ public class HiveCopyEntityHelper {
   private final DeregisterFileDeleteMethod deleteMethod;
 
   private final Optional<CommitStep> tableRegistrationStep;
-  private final Optional<CommitStep> dataLocationChangeStep;
-
   private final Map<List<String>, Partition> sourcePartitions;
   private final Map<List<String>, Partition> targetPartitions;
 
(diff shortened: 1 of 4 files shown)


ibuenros:
This PR includes the `topic.name` commits as well. Can you please split them?

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -314,15 +317,26 @@
...
         HiveSpec tableHiveSpec = new SimpleHiveSpec.Builder<>(targetPath)
             .withTable(HiveMetaStoreUtils.getHiveTable(this.targetTable.getTTable())).build();
-        CommitStep tableRegistrationStep = new HiveRegisterStep(this.targetURI, tableHiveSpec, this.hiveRegProps);
-
+        CommitStep tableRegistrationStep = new HiveRegisterStep(this.targetURI, tableHiveSpec, this.hiveRegProps, false);
         this.tableRegistrationStep = Optional.of(tableRegistrationStep);
 
+        // Constructing CommitStep object for table alteration, not necessary to use.
+        if (this.existingTargetTable.isPresent()) {
+          HiveSpec alteredTableHiveSpec = new SimpleHiveSpec.Builder<>(targetPath)
```
ibuenros: It is possible that attributes other than location need to be changed.
I think the fix is a bit simpler than this. If you provide a new table to register, `HiveRegister` will automatically alter the table if it already exists. All that you need to do is, if user specifies `UPDATE_TABLE`, then don't delete the existing table.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/UnpartitionedTableFileSet.java
```diff
@@ -65,23 +65,33 @@ public UnpartitionedTableFileSet(String name, HiveDataset dataset, HiveCopyEntit
     Optional<Table> existingTargetTable = this.helper.getExistingTargetTable();
     if (existingTargetTable.isPresent()) {
       if (!this.helper.getTargetTable().getDataLocation().equals(existingTargetTable.get().getDataLocation())) {
-        if (this.helper.getExistingEntityPolicy() != HiveCopyEntityHelper.ExistingEntityPolicy.REPLACE_TABLE) {
+
+        if (this.helper.getExistingEntityPolicy() == HiveCopyEntityHelper.ExistingEntityPolicy.UPDATE_TABLE) {
```
ibuenros: Can you use a switch statement instead.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/UnpartitionedTableFileSet.java
```diff
@@ -65,27 +65,25 @@ public UnpartitionedTableFileSet(String name, HiveDataset dataset, HiveCopyEntit
...
+            log.warn("Source table will not be deregistered while file locaiton has been changed, update source table's"
+                + " file location to" + this.helper.getTargetTable().getDataLocation());
+            break ;
+          case REPLACE_TABLE:
+            // Required to de-register the original table.
+            log.warn("Source and target table are not compatible. Will override target table " + existingTargetTable.get()
+                .getDataLocation());
+            stepPriority = this.helper.addTableDeregisterSteps(copyEntities, fileSet, stepPriority, this.helper.getTargetTable());
+            break ;
+          case ABORT:
```
ibuenros: Please make this `default:` instead

PR #1736, opened 2017-03-29
Fix the multithreading bug in TimestampWatermark

enjoyear:
Fix the bug that the shared object SimpleDateFormat will cause issues in multi-threads. The scenario is that when all pull jobs start at the same time, the partition calculation for all jobs will be messed up.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
@@ -232,7 +232,7 @@ public class Partitioner {
    * @param diff the amount of difference
    * @return the adjusted watermark value
    */
-  synchronized private static long adjustWatermark(String baseWatermark, WatermarkType watermarkType, int diff) {
+  private static long adjustWatermark(String baseWatermark, WatermarkType watermarkType, int diff) {
     SimpleDateFormat watermarkTimeParser = new SimpleDateFormat(WATERMARKTIMEFORMAT);
     Date date;
     long result = ConfigurationKeys.DEFAULT_WATERMARK_VALUE;
--- a/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
@@ -41,6 +41,7 @@ import gobblin.source.extractor.utils.Utils;
 import gobblin.source.extractor.watermark.WatermarkPredicate;
 import gobblin.source.extractor.watermark.WatermarkType;
 
+
 /**
  * An implementation of default partitioner for all types of sources
  */
@@ -48,8 +49,6 @@ public class Partitioner {
   private static final String WATERMARKDATEFORMAT = "yyyyMMdd";
   private static final String WATERMARKHOURFORMAT = "yyyyMMddHH";
   private static final String WATERMARKTIMEFORMAT = "yyyyMMddHHmmss";
-
-  private static final SimpleDateFormat WATERMARKTIMEPARSER = new SimpleDateFormat(WATERMARKTIMEFORMAT);
   private static final Logger LOG = LoggerFactory.getLogger(Partitioner.class);
   public static final String HAS_USER_SPECIFIED_HIGH_WATERMARK = "partitioner.hasUserSpecifiedHighWatermark";
   public static final String HAS_USER_SPECIFIED_PARTITIONS = "partitioner.hasUserSpecifiedPartitions";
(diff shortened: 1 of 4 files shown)


gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
```diff
@@ -37,7 +37,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(TimestampWatermark.class);
   // default water mark format(input format) example: 20140301050505
   private static final String INPUTFORMAT = "yyyyMMddHHmmss";
-  private static final SimpleDateFormat INPUTFORMATPARSER = new SimpleDateFormat(INPUTFORMAT);
+  private final SimpleDateFormat _inputFormatParser;
```
zxcware: Per the coding styles ([Google style guide](https://google.github.io/styleguide/javaguide.html) or sun java guide) in Gobblin,  names should not begin with '_'. In your case, a name `inputFormatParser` is good enough.

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
```diff
@@ -46,6 +47,7 @@
   public TimestampWatermark(String watermarkColumn, String watermarkFormat) {
     this.watermarkColumn = watermarkColumn;
     this.watermarkFormat = watermarkFormat;
+    _inputFormatParser = new SimpleDateFormat(INPUTFORMAT);
```
zxcware: This approach doesn't solve the multithreading issue. What about two threads concurrently executing `getIntervals`? The right fix is to define it as a local variable or use `synchronize` primitive. I think local variable is a better way.

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
```diff
@@ -37,7 +37,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(TimestampWatermark.class);
   // default water mark format(input format) example: 20140301050505
   private static final String INPUTFORMAT = "yyyyMMddHHmmss";
-  private static final SimpleDateFormat INPUTFORMATPARSER = new SimpleDateFormat(INPUTFORMAT);
+  private final SimpleDateFormat _inputFormatParser;
+
```
zxcware: Remove empty line.

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
zxcware: FYI, this issue exists in `DateWatermark`, `HourWatermark`, etc.

enjoyear: I think it should work because the getIntervals is already a synchronized method. What do you think?

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
enjoyear: Done

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
enjoyear: Done

gobblin-core/src/main/java/gobblin/source/extractor/watermark/TimestampWatermark.java
enjoyear: Fixed in other classes as well.

zxcware: Oh yeah. I didn't notice it's protected by `synchronized`.

zxcware: There is another one in `Partitioner`. Could you help me fix it? Thanks. :)

zxcware: The line affected in `Partitioner` is:
`private static final SimpleDateFormat WATERMARKTIMEPARSER = new SimpleDateFormat(WATERMARKTIMEFORMAT);`

enjoyear: Fixed. Created a local variable this time because the method is static.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -237,6 +233,7 @@ public Partitioner(SourceState state) {
    * @return the adjusted watermark value
    */
   synchronized private static long adjustWatermark(String baseWatermark, WatermarkType watermarkType, int diff) {
+    SimpleDateFormat watermarkTimeParser = new SimpleDateFormat(WATERMARKTIMEFORMAT);
```
enjoyear: Removed the synchronized, and yes, I am using the gobblin code style here https://gobblin.readthedocs.io/en/latest/developer-guide/files/codestyle-intellij-gobblin.xml

PR #1739, opened 2017-03-30
Add commonProps to reduce FsStateStore memory usage when loading, and…

ydai1124:
… also parallize get datasetStateStores


ydai1124:
@htran1 Comments addressed. Also did a rebase.

PR #1740, opened 2017-03-30
Fix partition gaps in Partitioner

zxcware:
- All but last partition is set as [lowWatermark, highWatermark) , last partition will be [lowWatermark, highWatermark]
- Fixing tests is in progress

### DIFF
--- a/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
+++ b/gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
@@ -15,6 +15,9 @@ import lombok.Getter;
  * <p>
  *   A {@link Source} partitions its data into a collection of {@link Partition}, each of which will be used to create
  *   a {@link WorkUnit}.
+ *
+ *   By default, {@link #isLowWatermarkInclusive} is true, {@link #isHighWatermarkInclusive} is false,
+ *   {@link #hasUserSpecifiedHighWatermark} is false
  * </p>
  *
  * @author zhchen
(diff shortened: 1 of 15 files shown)


zxcware:
@htran1 @ibuenros Please review. I will patch another commit on fixing tests.

gobblin-core/src/main/java/gobblin/source/extractor/watermark/SimpleWatermark.java
```diff
@@ -67,11 +67,11 @@ public int getDeltaNumForNextWatermark() {
     long startNum = lowWatermarkValue;
     long endNum = highWatermarkValue;
     boolean longOverflow = false;
-    while (startNum <= endNum && !longOverflow) {
+    while (startNum < endNum && !longOverflow) {
       longOverflow = (Long.MAX_VALUE - interval < startNum);
-      nextNum = longOverflow ? Long.MAX_VALUE : startNum + interval;
-      intervalMap.put(startNum, (nextNum <= endNum ? nextNum : endNum));
-      startNum = nextNum + deltaForNextWatermark;
+      nextNum = longOverflow ? Long.MAX_VALUE : Math.min(startNum + interval, endNum);
```
htran1: This is not the same logic in the overflow case. The old logic would put endNum instead of Long.MAX_VALUE for overflow case.

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partition.java
```diff
@@ -15,23 +15,44 @@
...
  *
  * @author zhchen
  */
 @AllArgsConstructor
 @EqualsAndHashCode
 public class Partition {
-  @Getter private final long lowWatermark;
-  @Getter private final long highWatermark;
+  public static final String IS_LOWWATERMARK_INCLUSIVE = "partition.isLowWatermarkInclusive";
+  public static final String IS_HIGHWATERMARK_INCLUSIVE = "partition.isLowWatermarkInclusive";
```
zxcware: Fixed by introducing se/de of Partition from WorkUnit

gobblin-core/src/main/java/gobblin/source/extractor/watermark/SimpleWatermark.java
zxcware: Good catch. Fixed!

PR #1742, opened 2017-03-30
Add feature to QueryBasedExtractor to be able to retry from the first iterator

enjoyear:
Current behavior for failed QueryBasedExtractor task is that it always retries from the failed iterator. But in some cases, for example, downloading chunks from a long lasting stream with connection, the retry task cannot restart from the middle(failed iterator), it has to restart from the beginning(Rebuild the connection, and re-download in this example.). The PR adds this feature to the QueryBasedExtractor, and also corrects/resets the downloaded records count for the retry task.


gobblin-runtime/src/main/java/gobblin/runtime/local/LocalTaskStateTracker.java
```diff
@@ -88,6 +88,10 @@ public void onTaskRunCompletion(Task task) {
       // it has not reached the maximum number of retries
       WorkUnitState.WorkingState state = task.getTaskState().getWorkingState();
       if (state == WorkUnitState.WorkingState.FAILED && task.getRetryCount() < this.maxTaskRetries) {
+        boolean restartFromFirstIterator = task.getTaskState().getPropAsBoolean(ConfigurationKeys.SOURCE_QUERYBASED_RETRY_RESETITERATOR, false);
```
zxcware: Both `Task` and `LocalTaskStateTracker` are generic. They shouldn't be affected by any source specific configuration.

We're making the change to fix the retry semantic. I don't think you need a configuration key.

enjoyear: Please check my updated changes, which should be trivial.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
```diff
@@ -296,6 +295,7 @@ public String getProgress() {
   @Override
   @SuppressWarnings("unchecked")
   public void run() {
+    recordsPulled.set(0L);
```
zxcware: For retry from the beginning, setting `recordsPulled` to 0 makes sense. But for retry from failed point, this logic is wrong.  We do need a configuration key to identify which retry mode is favored, making sure each is implemented correctly (all states should be cleared for retry from beginning, keep previous states for retry from failed point).

gobblin-core/src/main/java/gobblin/source/extractor/extract/QueryBasedExtractor.java
```diff
@@ -39,17 +27,30 @@
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.MDC;
 
+import com.google.common.annotations.VisibleForTesting;
```
enjoyear: There is no unused imports.

gobblin-runtime/src/main/java/gobblin/runtime/Task.java
enjoyear: Added a configuration key and an enum WorkUnitRetryMode to control this.

gobblin-core-base/src/main/java/gobblin/source/extractor/WorkUnitRetryPolicy.java
```diff
@@ -22,7 +22,7 @@
 
 
 /**
- * An enumeration of supported work unit retry policies.
+ * An enumeration of retry policies determining under what conditions the failed work unit should retry
```
zxcware: I think the right syntax is: An enumeration of retry policies determining under what conditions could the failed work unit retry

PR #1743, opened 2017-03-31
Added property with scheduler service startup timeout

wikp:
Hello,

I'm trying to run Gobblin in standalone cluster mode on our dev environment. I don't want to use Azkaban or Oozie as a scheduler, so I configured Quartz in cluster mode with MySQL backend. Because of not so good performance of mysql cluster on dev environment or physical distance between current mysql master and Gobblin installation, I've encountered `TimeoutException`s during job scheduler startup on Gobblin master. After diving in the code, I've noticed that in `startUp` method of `JobScheduler` there is such a piece of code: `this.scheduler.awaitRunning(1, TimeUnit.SECONDS);`. That was not enough to establish the connection by Quartz which produces  stacktrace below:

```
Exception in thread "GobblinHelixJobScheduler STARTING" java.lang.IllegalStateException: Scheduler service is not running.
	at gobblin.scheduler.JobScheduler.startUp(JobScheduler.java:178)
	at gobblin.cluster.GobblinHelixJobScheduler.startUp(GobblinHelixJobScheduler.java:82)
	at com.google.common.util.concurrent.AbstractIdleService$2$1.run(AbstractIdleService.java:54)
	at com.google.common.util.concurrent.Callables$3.run(Callables.java:93)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.util.concurrent.TimeoutException: Timed out waiting for  [STARTING] to reach the RUNNING state. Current state: STARTING
	at com.google.common.util.concurrent.AbstractService.awaitRunning(AbstractService.java:295)
	at com.google.common.util.concurrent.AbstractIdleService.awaitRunning(AbstractIdleService.java:184)
	at gobblin.scheduler.JobScheduler.startUp(JobScheduler.java:176)
	... 4 more
```

Introducing new property: `jobscheduler.startup.timeout` with default value of `1` was the simplest solution that fixes the issue. Please let me know if it meets your standards :)

### DIFF
--- a/conf/gobblin-standalone-v2.properties
+++ b/conf/gobblin-standalone-v2.properties
@@ -62,4 +62,8 @@ task.status.reportintervalinms=5000
 # Unit in milliseconds, configurable.
 jobconf.monitor.interval=30000
 
+# Startup timeout of JobScheduler. You may want to make it larger while using built-in Quartz scheduler in clustered
+# mode. (seconds)
+jobscheduler.startup.timeout=1
+
 task.maxretries=0
--- a/conf/gobblin-standalone.properties
+++ b/conf/gobblin-standalone.properties
@@ -66,3 +66,7 @@ task.status.reportintervalinms=5000
 # Unit in milliseconds, configurable.
 jobconf.monitor.interval=30000
 
+# Startup timeout of JobScheduler. You may want to make it larger while using built-in Quartz scheduler in clustered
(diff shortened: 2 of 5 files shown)


wikp:
Closing this because there is much more to do to allow clustered Quartz setup, for example: some DAG of dependencies between services starting up in `ServiceBasedAppLauncher`.

PR #1744, opened 2017-03-31
TIMESTAMP fields of the initial schema in gobblin-metastore

wikp:
As of mysql 5.6.6 `TIMESTAMP` columns are `NOT NULL` by default which means we should assign a `DEFAULT` value to them. My understanding of the modified code was that they could be `NULL`, so I've made appropriate modifications to the schema.

Without this modification, there is an error during creating tables.


eogren:
Hi Piotr - 

Thanks for this contribution!

I haven't looked at the code so this may not be strictly needed, but if we want to keep backwards compatibility I think the right behavior is actually to set:

```sql
start_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
end_time TIMESTAMP DEFAULT 0
```

Is it possible for you to verify this on MySQL 5.7 and make sure everything is still working as expected? thanks.

wikp:
Yes, I'll check it :)

But one question - is `ON UPDATE CURRENT TIMESTAMP` needed in `start_time` field? If record is created after start then updating its `state` column from `RUNNING` to `SUCCEEDED` will change the `start_time` - which is, I think, not a desired behaviour ;) But I'll read the code and check every possible scenario.

Thanks for your comment, Eric :)

wikp:
So I've checked mysql behaviour with create statements from modified files and made some changes.

Default `sql_mode` variable in MySQL 5.7 contains `NO_ZERO_DATE` and `STRICT_TRANS_TABLES` modifiers: 

```
ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
```

Because of that, MySQL produces an error during creating tables.

`DatabaseJobHistoryV100` class is responsible for inserting values to `gobblin_{job,task}_executions` tables. It checks for `start_time` and `end_time` passed by "user" and converts `null` to `new Timestamp(1)` with UTC `Calendar`. 
I made a change that ensures UTC timezone in the whole db init script and sets appropriate defaults explicitly. Please note that I cannot set `0` as the default value because of `NO_ZERO_DATE` being on by default in MySQL 5.7.

Hope it will be OK :)

eogren:
Ah, interesting! Thanks for following up, lgtm.

PR #1747, opened 2017-04-04
Respect retry delay in permit requests (throttling).

ibuenros:


ibuenros:

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-client/src/main/java/gobblin/util/limiter/BatchedPermitsRequester.java
```diff
@@ -239,6 +256,12 @@ public void onError(Throwable exc) {
         }
 
         BatchedPermitsRequester.this.retries++;
+        try {
```
htran1: Can be moved back to finally block.

PR #1748, opened 2017-04-04
Throttled FileSystem through broker.

ibuenros:
This is a large PR covering throttled (and in general) instrumented file systems obtained from a `SharedResourcesBroker`. Luckily, a lot of the code is boilerplate and refactoring, so it's not actually THAT big.

Main Changes
==========

Implicit Broker
-------------
Introduced the concept of an implicit `SharedResourcesBroker`. This is necessary for using brokers without having to explicitly define them in the application structure, which might be too invasive. Basically, there is a singleton implicit broker whose configuration is loaded from environment variables, java properties, a configuration file in the classpath, and possibly Hadoop configuration. Furthermore, applications can inject a custom implicit broker which is thread-local and is inherited by children threads.

Instrumented File Systems
-------------------------
Changed the previous `Instrumented*FileSystem` for a slightly different structure.

An `InstrumentedFileSystem` is a `FileSystem` decorator that instantiates a regular `FileSystem` and decorates it using a variety of `FileSystemInstrumentationFactory`s in the classpath (for example, metrics, throttling, etc.). This involves a slightly convoluted structure:
* `FileSystemDecorator` is a base decorator for `FileSystem`s. It allows optionally switching the schemes in the uri. The file is very large, but it is mostly mechanical.
* `FileSystemInstrumentation` is just an extension of `FileSystemDecorator` that does not switch scheme. It is intended as a base for other instrumentations.
* `FileSystemInstrumentationFactory` is the actual class that should be extended for instrumentations. 
* `FileSystemFactory` is a `SharedResourceFactory` that builds an underlying `FileSystem`, then decorates it with any `FileSystemInstrumentationFactory`s that exist in the classpath.
* `InstrumentedFileSystem` and subclasses (`InstrumentedHDFSFileSystem`, etc.) are a different convenient way of getting a `FileSystem` with instrumentations. Basically, users can either use an instrumented scheme (e.g. `instrumented-hdfs`), or use the property `fs.<scheme>.impl` to one of the instrumented file systems. The `InstrumentedFileSystem` will use the implicit broker in that context to return an instrumented `FileSystem`.

There are currently two `FileSystemInstrumentationFactory`s: `MetricsFileSystemInstrumentation` which adds metrics to all fs calls, and `ThrottledFileSystem` which adds throttling.


gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemFactory.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * Equivalent to {@link FileSystem#get(URI, Configuration)}, but uses the input {@link SharedResourcesBroker} to configure
+   * add-ons to the {@link FileSystem} (e.g. throttling, instrumentation).
+   */
+  public static <S extends ScopeType<S>> FileSystem get(URI uri, Configuration configuration, SharedResourcesBroker<S> broker)
+    throws IOException {
+    try {
+      return broker.getSharedResource(new FileSystemFactory<S>(), new FileSystemKey(uri, configuration));
```
ydai1124: Do you need a new instance of FileSystemKey created every time? Can you cache it based on the fs uri ?

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -63,4 +64,17 @@ public static Object resolveUnderlyingObject(Object obj) {
 
     return Lists.reverse(lineage);
   }
+
+  /**
+   * Determines whether any object in the decorator lineage satisfies the input predicate.
+   */
+  public static boolean applyPredicateOnLineage(Object obj, Predicate<Object> predicate) {
```
ydai1124: Is this used?

gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemDecorator.java
```diff
@@ -0,0 +1,533 @@
...
+
+import static gobblin.util.filesystem.InstrumentedFileSystemUtils.*;
+
+
+/**
+ * This is a decorator for {@link FileSystem} that allows optionally changing scheme.
+ *
+ * Note subclasses must set the underlying {@link FileSystem} at {@link #underlyingFs} as necessary.
+ */
+class FileSystemDecorator extends FileSystem implements Decorator {
```
ydai1124: Are most methods auto-generated in this class? Lots of `throws java.io.IOException` instead of `throws IOException`. Also missing `@override` for most methods.

gobblin-utility/src/main/java/gobblin/broker/SharedResourcesBrokerFactory.java
```diff
@@ -44,8 +78,76 @@
...
+   * Get the implicit {@link SharedResourcesBroker} in the callers thread. This is either a singleton broker configured
+   * from environment variables, java options, and classpath configuration options, or a specific broker injected
+   * elsewhere in the application.
+   *
+   * In general, it is preferable to explicitly pass around {@link SharedResourcesBroker}s, as that allows better
+   * control over the scoping. However, in cases where it is hard to do so, this method provides an alternative to
+   * method of acquiring a configured broker.
+   */
+  public static SharedResourcesBroker<?> getImplicitBroker() {
+    SharedResourcesBroker<?> threadLocal = threadLocalBroker.get();
```
ibuenros: `ThreadLocal` variables are thread-safe in the sense that each thread will use a different variable.

gobblin-utility/src/main/java/gobblin/util/HadoopUtils.java
```diff
@@ -912,4 +913,10 @@ public static boolean hasContent(FileSystem fs, Path path)
     }
     return content;
   }
+
+  public static void addGobblinSite() {
+    Configuration.addDefaultResource("gobblin-site.xml");
+    Configuration configuration = new Configuration();
+    log.info("Configuration: " + configuration);
```
ibuenros: This was actually only there for debugging. Removed it.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemFactory.java
ibuenros: Is there a disadvantage to creating a new instance? Given the class is stateless, this reduces cache management.

gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemDecorator.java
ibuenros: The methods are auto-generated.

gobblin-utility/src/main/java/gobblin/broker/SharedResourcesBrokerFactory.java
```diff
@@ -17,20 +17,54 @@
...
+ * * Java properties of the current JVM.
+ * * Environment variables of the current shell.
+ * * Hadoop configuration obtained via {@link Configuration#Configuration()}. This can be disabled setting
+ *   {@link #LOAD_HADOOP_CONFIGURATION} to false.
  */
 public class SharedResourcesBrokerFactory {
 
+  public static final String LOAD_HADOOP_CONFIGURATION = BrokerConstants.GOBBLIN_BROKER_CONFIG_PREFIX + ".loadHadoopConfiguration";
+
+  public static final String BROKER_CONF_FILE_KEY = "gobblin.broker.configuration";
```
ibuenros: Fixed

gobblin-utility/src/main/java/gobblin/broker/SharedResourcesBrokerFactory.java
```diff
@@ -44,8 +78,76 @@
...
+    String brokerConfPath = DEFAULT_BROKER_CONF_FILE;
+    if (tmpConfig.hasPath(BROKER_CONF_FILE_KEY)) {
+      brokerConfPath = tmpConfig.getString(BROKER_CONF_FILE_KEY);
+    }
+
+    Config resourceConfig = ConfigFactory.parseResources(SharedResourcesBrokerFactory.class, brokerConfPath);
+
+    config = config.withFallback(resourceConfig).withFallback(systemConfig);
+
+    if (!config.hasPath(LOAD_HADOOP_CONFIGURATION) || config.getBoolean(LOAD_HADOOP_CONFIGURATION)) {
```
ibuenros: Fixed

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
ibuenros: Removed

gobblin-api/src/main/java/gobblin/util/DecoratorUtils.java
```diff
@@ -19,6 +19,7 @@
 
 import java.util.List;
 
+import com.google.common.base.Predicate;
 import com.google.common.collect.Lists;
```
ydai1124: Not used. Can you remove it?

PR #1749, opened 2017-04-05
Added scheduler and master-worker model in Gobblin Service

abti:
Added scheduler and master-worker model in Gobblin Service. 

@htran1 please review.


gobblin-service/src/main/java/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -169,45 +116,59 @@ public void onDeleteSpec(URI deletedSpecURI, String deletedSpecVersion) {
...
+
+      // Schedule all compiled JobSpecs on their respective Executor
+      for (Map.Entry<Spec, SpecExecutorInstanceProducer> specsToExecute : specExecutorInstanceMap.entrySet()) {
+        // Run this spec on selected executor
+        SpecExecutorInstanceProducer producer = null;
+        try {
+          producer = specsToExecute.getValue();
+          Spec jobSpec = specsToExecute.getKey();
+          producer.addSpec(jobSpec);
+        } catch(Exception e){
```
htran1: Can specific exception be caught here?

gobblin-service/src/main/java/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -169,45 +116,59 @@ public void onDeleteSpec(URI deletedSpecURI, String deletedSpecVersion) {
...
-              _log.warn(String.format("Did not delete Spec with URI: %s and Version: %s on SpecExecutorInstance: %s",
-                  deletedSpecURI, deletedSpecVersion, topologySpec.getSpecExecutorInstanceProducer().getUri()), e);
-            }
-          }
-        }
       }
-    } catch (SpecNotFoundException e) {
-      _log.info(String.format("Spec with URI: %s was not found in TopologyCatalog, going to try FlowCatalog",
-          deletedSpecURI));
+    } catch (RuntimeException e) {
```
htran1: There is risk of suppressing other runtime exceptions.

gobblin-service/src/main/java/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -0,0 +1,293 @@
...
+  public void onDeleteSpec(URI deletedSpecURI, String deletedSpecVersion) {
+    _log.info("Spec deletion detected: " + deletedSpecURI + "/" + deletedSpecVersion);
+
+    if (!isActive && helixManager.isPresent()) {
+      HelixUtils.sendUserDefinedMessage(ServiceConfigKeys.HELIX_FLOWSPEC_REMOVE, deletedSpecURI.toString() + ":" +
+              deletedSpecVersion, UUID.randomUUID().toString(), InstanceType.CONTROLLER, helixManager.get(), _log);
+      return;
+    }
+
+    // To determine if the call was made via TopologyCatalog or FlowCatalog, iterate over StackTrace
```
htran1: Comment doesn't seem to match code.

gobblin-service/src/main/java/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -0,0 +1,293 @@
...
+      scheduleJob(jobProps, jobListener, additionalJobDataMap, GobblinServiceJob.class);
+    } catch (Exception e) {
+      throw new JobException("Failed to schedule job " + jobProps.getProperty(ConfigurationKeys.JOB_NAME_KEY), e);
+    }
+  }
+
+  @Override
+  public void runJob(Properties jobProps, JobListener jobListener) throws JobException {
+    try {
+      Spec flowSpec = this.scheduledFlowSpecs.get(jobProps.getProperty(ConfigurationKeys.JOB_NAME_KEY));
```
htran1: Schedule should be removed before submitting to producer.

gobblin-service/src/main/java/gobblin/service/modules/orchestration/Orchestrator.java
abti: Removed. Earlier Orchestrator was listener to both Flow and Topology Catalog creating a situation where it cannot distinguish between a Flow and Topology URI - hence no way of knowing if missing Spec runtime exception are really an issue or not. 
However, with recent changes Orchestrator only listens to Topology Catalog. So it is not needed anymore.

gobblin-service/src/main/java/gobblin/service/modules/orchestration/Orchestrator.java
abti: producer.addSpec does not throw a typed exception, so catching generic Exception to log it and not crash the thread.

gobblin-service/src/main/java/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
abti: Removed schedule while compiling in compiler, because I think its more of a property of compiler to decide if the schedule should be passed along in the JobSpec or not. Any contrary thoughts?

gobblin-service/src/main/java/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
abti: Removed

gobblin-service/src/main/java/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -250,44 +258,60 @@ private Path getServiceWorkDirPath(FileSystem fs, String serviceName, String ser
    * @param changeContext notification context
    */
   private void handleLeadershipChange(NotificationContext changeContext) {
-    if (this.helixManager.isLeader()) {
-      LOGGER.info("Leader notification for {} HM.isLeader {}", this.helixManager.getInstanceName(),
-          this.helixManager.isLeader());
+    if (this.helixManager.isPresent() && this.helixManager.get().isLeader()) {
```
htran1: Should we ever get notifications if there is no helix manager?

gobblin-service/src/main/java/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -319,9 +343,13 @@ public void stop() throws ApplicationException {
   @VisibleForTesting
   void connectHelixManager() {
     try {
-      this.helixManager.connect();
-      this.helixManager.getMessagingService().registerMessageHandlerFactory(
-          Message.MessageType.USER_DEFINE_MSG.toString(), getUserDefinedMessageHandlerFactory());
+      if (this.helixManager.isPresent()) {
```
htran1: Check is already done before calling this method, so it can be removed from one of the places.

gobblin-service/src/test/java/gobblin/service/modules/core/GobblinServiceHATest.java
```diff
@@ -0,0 +1,501 @@
...
+        .setTemplateUris(TEST_TEMPLATE_URI_2).setSchedule(TEST_SCHEDULE_2).setRunImmediately(true)
+        .setProperties(new StringMap(flowProperties));
+
+    // Try create on both nodes
+    long schedulingStartTime = System.currentTimeMillis();
+    try {
+      this.node1FlowConfigClient.createFlowConfig(flowConfig1);
+      this.node2FlowConfigClient.createFlowConfig(flowConfig2);
+    } catch (Exception e) {
+
```
htran1: Assert if exception?

gobblin-service/src/test/java/gobblin/service/modules/core/GobblinServiceHATest.java
```diff
@@ -0,0 +1,501 @@
...
+    try {
+      this.node2FlowConfigClient.updateFlowConfig(flowConfig);
+      Assert.fail("Get should have raised a 404 error");
+    } catch (RestLiResponseException e) {
+      Assert.assertEquals(e.getStatus(), HttpStatus.NOT_FOUND_404);
+    }
+  }
+
+  @Test (dependsOnMethods = "testBadUpdate")
```
htran1: Might need to disable this test on Travis.

gobblin-service/src/test/java/gobblin/service/modules/orchestration/OrchestratorTest.java
```diff
@@ -263,9 +263,9 @@ public void deleteFlowSpec() throws Exception {
     }
     // Make sure FlowCatalog has the previously added Flow
     Assert.assertTrue(specs.size() == 1, "Spec store should contain 1 Flow that was added in last test");
-    // Make sure FlowCatalog Listener knows about the previously added Flow
+    // Orchestrator is a no-op listener for any new FlowSpecs, so no FlowSpecs should be around
     int specsInSEI = ((List)(sei.listSpecs().get())).size();
-    Assert.assertTrue(specsInSEI == 1, "SpecExecutorInstanceProducer should contain 1 "
+    Assert.assertTrue(specsInSEI == 0, "SpecExecutorInstanceProducer should contain 1 "
```
htran1: Should message be 'contain 0'?

gobblin-service/src/test/java/gobblin/service/modules/core/GobblinServiceHATest.java
abti: That was left behind mistakenly, removed it. 
I was simulating restart tests by not cleaning up previous state, ignoring 'already exists' exceptions and running same tests multiple times :)

gobblin-service/src/test/java/gobblin/service/modules/core/GobblinServiceHATest.java
abti: why?

gobblin-service/src/test/java/gobblin/service/modules/orchestration/OrchestratorTest.java
abti: Fixed

gobblin-service/src/test/java/gobblin/service/modules/core/GobblinServiceHATest.java
htran1: Travis is slow, so that may affect the failover time. First see if it can run reliably on travis in a reasonable amount of time.

abti: Seems to be passing on Travis for now. But we can disable it if starts failing transiently?

gobblin-service/src/main/java/gobblin/service/modules/core/GobblinServiceManager.java
abti: yeah we would not get a notification, but Intellij does it not likes it if I invoke helixManager.get() without checking helixManager.isPresent()

gobblin-service/src/main/java/gobblin/service/modules/core/GobblinServiceManager.java
abti: Intellij to blame here too. Was forced to add check for isPresent().

PR #1751, opened 2017-04-05
Gsc multiple properties

enjoyear:
The change is to support fetching multiple google search console properties in one gobblin job. Also include the source property as the last column of the schema.


gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -47,7 +50,12 @@
    * Must Provide.
    * Provide the property site URL whose google search analytics data you want to download
    */
-  public static final String KEY_PROPERTY = SOURCE_GOOGLE_WEBMASTER_PREFIX + "property_url";
+  public static final String KEY_PROPERTY = SOURCE_GOOGLE_WEBMASTER_PREFIX + "property_urls";
+  /**
+   * Optional. Default to true.
```
abti: I think typical use-case would be to honor source schema, so any additional columns added at runtime should not be a default behavior

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -159,14 +167,17 @@
   // =============================================
 
   private final static Splitter splitter = Splitter.on(",").omitEmptyStrings().trimResults();
+  private final static String SOURCE_PROPERTY_COLUMN_NAME = "source";
```
abti: What if source schema has "source" column name? Should we make this configurable?

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
```diff
@@ -47,7 +50,12 @@
    * Must Provide.
    * Provide the property site URL whose google search analytics data you want to download
    */
-  public static final String KEY_PROPERTY = SOURCE_GOOGLE_WEBMASTER_PREFIX + "property_url";
+  public static final String KEY_PROPERTY = SOURCE_GOOGLE_WEBMASTER_PREFIX + "property_urls";
+  /**
+   * Optional. Default to true.
+   * Determine whether to add source property as the last column to your configured schema
+   */
+  public static final String KEY_INCLUDE_SOURCE_PROPERTY = SOURCE_GOOGLE_WEBMASTER_PREFIX + "include_source_property";
```
abti: Do you also intend to include more columns to schema at runtime in future? If so, its better to move to a model where additional columns are added to schema via converter. If not, we can stick to what you have.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSourceDaily.java
```diff
@@ -45,7 +46,7 @@
   @Override
   GoogleWebmasterExtractor createExtractor(WorkUnitState state, Map<String, Integer> columnPositionMap,
       List<GoogleWebmasterFilter.Dimension> requestedDimensions,
-      List<GoogleWebmasterDataFetcher.Metric> requestedMetrics)
+      List<GoogleWebmasterDataFetcher.Metric> requestedMetrics, JsonArray schemaJson)
```
abti: I see that you have changes some contracts (constructors), are the interfaces / classes in these connectors still evolving. If so, we should mark the classes here with @ Alpha annotation

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
enjoyear: Okay. I will change default behavior to false.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
enjoyear: No, I think that's the only usecase I can think of now.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSource.java
enjoyear: I will make it configurable.

gobblin-modules/google-ingestion/src/main/java/gobblin/ingestion/google/webmaster/GoogleWebMasterSourceDaily.java
enjoyear: Will do.

enjoyear:
@abti , ready for another review. Thanks!

PR #1753, opened 2017-04-06
Fixed hanging embedded Gobblin when initialization fails and metrics.

ibuenros:


ibuenros:
@abti can you review? Thanks!

abti:
Please check both default run group tests are failing: https://travis-ci.org/linkedin/gobblin/builds/219388176

eogren:
I don't know if you were looking at the same thing, but I noticed EmbeddedDistCp was hanging for me (I was running via the CLI). 

The one-line fix I had was to put a `super.close()` call inside `FileAwareInputStreamDataWriter.close()` -- `InstrumentedDataWriterBase` starts a metrics thread which never gets closed. I don't know why this doesn't break on MR, YARN though. Also not sure if that metrics thread should just be started as a daemon thread so it can't hang the process if not cleaned up properly.

ibuenros:
@eogren No, this is a different case of hanging. Is the one you mentioned fix as well?

eogren:
not yet! see #1757

PR #1756, opened 2017-04-07
Catching UncheckedExecutionException in HiveSource

aditya1105:
During creation of WorkUnits in HiveSource, failure to create a single WorkUnit shouldn't lead to failure of Job.
Some exceptions in this case (eg. UpdateNotFoundException) are often wrapped as UncheckedExecutionException leading to failure of Job.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
@@ -16,6 +16,7 @@
  */
 package gobblin.data.management.conversion.hive.source;
 
+import com.google.common.util.concurrent.UncheckedExecutionException;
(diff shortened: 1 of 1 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -333,6 +334,8 @@ private void createWorkunitsForPartitionedTable(HiveDataset hiveDataset, AutoRet
       } catch (SchemaNotFoundException e) {
         log.error(String.format("Not Creating workunit for %s as schema was not found. %s",
             sourcePartition.getCompleteName(), e.getMessage()));
+      } catch (UncheckedExecutionException e) {
+        log.warn(e.getMessage());
```
agepati: Please log an error level message.

PR #1757, opened 2017-04-07
Fix distcp from CLI hang

eogren:
@ibuenros 

`bin/gobblin run distcp foo bar` was hanging on exit whenever any file was actually copied. The root cause was that `FSAwareInputDataStreamWriter` overrode `close()` without calling the superclass method -- this caused the metric updater thread from `InstrumentedDataWriterBase` to hang around forever because it was never shutdown.

This fix is correct, but I'm not sure if we should also be started the metric writer thread as a daemon thread vs one that can hang the process -- thoughts?

- Remove shadowed closer variable since the superclass defines one already
- Remove overridden close() method - it wasn't calling super.close()
which was causing bugs, and is now irrelevant since the parent close
will close resources properly


ibuenros:
I think it is a good idea to make the thread daemon as well.

PR #1758, opened 2017-04-07
Revert "Gsc multiple properties"

ibuenros:
Reverts linkedin/gobblin#1751


ibuenros:
@enjoyear We have to revert this PR because Gobblin does not work with Guava 20. Let's talk about how to maintain compatibility.

PR #1759, opened 2017-04-07
Revert "Google adwords"

ibuenros:
Reverts linkedin/gobblin#1587


ibuenros:
@enjoyear reverting due to guava 20.

PR #1762, opened 2017-04-07
Add salesorce day based dynamic partitioning

zxcware:
- First attempt on dynamic partitioning.
- We'll generalize it in the future.


gobblin-salesforce/src/main/java/gobblin/salesforce/SalesforceSource.java
```diff
@@ -61,6 +93,200 @@
...
+      values.put("year", Integer.toString(year));
+      String query = sub.replace(DAY_PARTITION_QUERY_TEMPLATE);
+      log.info("Histogram query: " + query);
+
+      try {
+        String soqlQuery = SalesforceExtractor.getSoqlUrl(query);
+        List<Command> commands = RestApiConnector.constructGetCommand(connector.getFullUri(soqlQuery));
+        CommandOutput<?, ?> response = connector.getResponse(commands);
+        histogram.add(parseHistogram(response));
+      } catch (Exception e) {
```
htran1: Please catch specific exceptions if possible.  Also, shouldn't the exception be propagated instead of suppressed?

gobblin-core/src/main/java/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -83,6 +84,21 @@ public Partitioner(SourceState state) {
     hasUserSpecifiedHighWatermark = false;
   }
 
+  public Partition getGobalPartition(long previousWatermark) {
```
zxcware: Done

gobblin-salesforce/src/main/java/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -700,6 +696,12 @@ public boolean bulkApiLogin() throws Exception {
         log.info("Waiting for bulk resultSetIds");
       }
 
+      if (this.bulkBatchInfo.getState() == BatchStateEnum.Failed) {
+        log.info("Bulk batch failed: " + bulkBatchInfo.toString());
```
zxcware: Fixed

gobblin-salesforce/src/main/java/gobblin/salesforce/SalesforceSource.java
zxcware: Fixed

gobblin-salesforce/src/main/java/gobblin/salesforce/SalesforceSource.java
```diff
@@ -217,8 +222,10 @@ private Histogram getHistogram(String entity, String watermarkColumn, SourceStat
         List<Command> commands = RestApiConnector.constructGetCommand(connector.getFullUri(soqlQuery));
         CommandOutput<?, ?> response = connector.getResponse(commands);
         histogram.add(parseHistogram(response));
+      } catch (DataRecordException e) {
+        throw e;
       } catch (Exception e) {
```
htran1: How about specifying the following list of specific exceptions and throwing either a runtime exception or DataRecordException?
      } catch (RestApiClientException | RestApiProcessingException | DataRecordException e) {
        throw new RuntimeException("Fail to get data of year " + year + " from salesforce", e);
      }

zxcware: Fixed

PR #1765, opened 2017-04-11
Gsc multiple properties

enjoyear:
The previous code was reverted due to Guava version issue. This PR removes all code about Google AdWords, now only contains reverted changes related to Google Search Console.


ibuenros:
Can you squash the commits?

PR #1767, opened 2017-04-11
Fixes to InstrumentedFileSystem on MR jobs and BatchedPermitsContainer.

ibuenros:
This PR does the following:

* Add an `onClose` trigger for `FileSystemInstrumentations`. Metrics instrumentation writes out a report on this trigger. This required updating the return type of `MetricContext.timer`.
* Fix `BatchedPermitsRequester` to correctly handle multiple batches with the same expiration.
* A few small fixes.

### DIFF
--- a/gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemInstrumentation.java
+++ b/gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemInstrumentation.java
@@ -17,6 +17,8 @@
 
 package gobblin.util.filesystem;
 
+import java.io.IOException;
+
 import org.apache.hadoop.fs.FileSystem;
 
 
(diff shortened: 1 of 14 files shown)


gobblin-utility/src/main/java/gobblin/util/filesystem/FileSystemInstrumentation.java
```diff
@@ -25,9 +27,36 @@
  */
 public class FileSystemInstrumentation extends FileSystemDecorator {
 
+  protected boolean closed = false;
```
yukuai518: Does this need to be thread safe or each object instance will have a different decorator instance?

ibuenros: It is generally an error to call `closed` twice from different threads. However, I think it's easy enough and potentially useful to make this thread safe, so I'll do it.

PR #1772, opened 2017-04-12
Support shorten dataset state store name

ydai1124:
Created `DatasetUrnStateStoreNameParser` for providing the mapping between datasetUrn and state store name.
It currently has two implementations. The default: `SimpleDatasetUrnStateStoreNameParser` and SHA1 based guid: `GuidDatasetUrnStateStoreNameParser`.
`FsDatasetStateStore` has this feature enabled. It is backwards compatible.
Other `DatasetStateStore` can also have this feature. 
@htran1 or @ibuenros  can you take a look?

### DIFF
--- a/gobblin-metastore/src/main/java/gobblin/metastore/nameParser/GuidDatasetUrnStateStoreNameParser.java
+++ b/gobblin-metastore/src/main/java/gobblin/metastore/nameParser/GuidDatasetUrnStateStoreNameParser.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gobblin.metastore.nameParser;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Collection;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileSystem;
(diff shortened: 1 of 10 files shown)


gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
```diff
@@ -17,10 +17,6 @@
```
htran1: Can this file be unbranched? I don't see any changes.

gobblin-metastore/src/main/java/gobblin/metastore/nameParser/GuidDatasetUrnStateStoreNameParser.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  /**
+   * Merge the given {@link Collection} of datasetUrns with {@link #sanitizedNameToDatasetURNMap}, and write the results
+   * to the {@link Path} of {@link #versionIdentifier}, which is a text file.
+   */
+  @Override
+  public void persistDatasetUrns(Collection<String> datasetUrns)
+      throws IOException {
```
htran1: Is there a case where we might need to merge with the old mapping or are we always guaranteed to have all datasetUrns.

ydai1124: We need to merge with the old mapping because one job execution may not process workunits from all datasets. The parameter `Collection<String> datasetUrns` only contains datasetUrns that are processed in the current job execution.

gobblin-metastore/src/main/java/gobblin/metastore/nameParser/GuidDatasetUrnStateStoreNameParser.java
```diff
@@ -0,0 +1,158 @@
...
+    try (FSDataOutputStream fsout = this.fs.create(tmpMapFile)) {
+      for (String key : this.sanitizedNameToDatasetURNMap.keySet()) {
+        fsout.write(key.getBytes(Charsets.UTF_8));
+        fsout.writeByte('\n');
+        fsout.write(this.sanitizedNameToDatasetURNMap.get(key).getBytes(Charsets.UTF_8));
+        fsout.writeByte('\n');
+      }
+    }
+    // Back up the previous file first, and then rename the new file.
+    if (this.fs.exists(this.versionIdentifier) && !this.fs.rename(this.versionIdentifier,
```
ydai1124: It will be cleaned like state backup files using separate retention jobs. Changed `StateStoreCleaner` for identifying these backups. I also updated javadoc there, since we recommend to use Gobblin-retention for these cleanup jobs.

gobblin-metastore/src/main/java/gobblin/metastore/FsStateStore.java
ydai1124: Reverted

ydai1124:
@htran1 Comments addressed.

PR #1779, opened 2017-04-17
Stress testing classes for throttling service.

ibuenros:
There are two main sets of changes in this PR:

Stres Test for Throttling
=================

* There are two stress tests: `MRStressTest` and `LocalStressTest`.
* A `Stressor` is a class that sends requests to a `Limiter`. Different stressors can simulate different usage patterns.
* The `LocalStressTest` uses and embedded throttling server and multiple threads to run a stressor. The requests do not go through the network, but the stress test can add artificial latency representing the network latency. Every 15 seconds, the stress test logs statistics about the throttling.
* The `MRStressTest` uses a real throttling server. It spawns multiple mappers each one running a stressor. The mappers emit their effective permit use rate every 15 seconds, which then gets collected by the reducer to compute the global effective QPS every 15 seconds.

Throttling algorithm updates
=====================

* Some changes to the batching algorithm. The previous algorithm was keeping the permit requests open for a long time (~7 seconds under high contention) because it allowed up to 7 seconds to acquire even a small number of permits. The new algorithm uses a short timeout instead, generally just short enough to satisfy the minimum number of permits requested. Larger permit allocations only happen if there are permits stored (meaning low contention).

### DIFF
--- a/gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-client/src/main/java/gobblin/util/limiter/stressTest/MRStressTest.java
+++ b/gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-client/src/main/java/gobblin/util/limiter/stressTest/MRStressTest.java
@@ -1,0 +1,363 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-client/src/main/java/gobblin/util/limiter/stressTest/MRStressTest.java
```diff
@@ -0,0 +1,363 @@
...
+
+      super.setup(context);
+    }
+
+    @Override
+    protected void map(Text key, NullWritable value, Context context) throws IOException, InterruptedException {
+      try {
+        Configuration configuration = context.getConfiguration();
+
+        Stressor stressor = context.getConfiguration().getClass(StressTestUtils.STRESSOR_CLASS,
```
ydai1124: Do all mappers get the same type of `STRESSOR_CLASS`? Is it worth to have a list of `STRESSOR_CLASS` and each mapper randomly instantiates one type?

ibuenros: For now yes. I do see the point of having different stressors on each mapper, but I'd want to put that in a separate PR.

ibuenros:
@ydai1124 addressed comments.

PR #1782, opened 2017-04-20
Add pinot based completeness check verifier

yukuai518:
A simple pinot http client to query the tier-to-count map for completeness check.


yukuai518:
@ibuenros

gobblin-compaction/src/main/java/gobblin/compaction/audit/AuditCountClientFactory.java
```diff
@@ -0,0 +1,12 @@
+package gobblin.compaction.audit;
+
+import gobblin.configuration.State;
+
+/**
+ * A factory class responsible for creating {@link AuditCountClient}
+ */
+public interface AuditCountClientFactory {
+  String AUDIT_COUNT_CLIENT_FACTORY = "audit.count.client.factory";
+  String DEFAULT_AUDIT_COUNT_CLIENT_FACTORY = "KafkaAuditCountHttpClientFactory";
```
ibuenros: I don't think there should be a default. If the user doesn't specify a class, then don't run this verification.

PR #1783, opened 2017-04-20
Added Whitelist Blacklist support for Hive view registration in Avro to ORC pipeline

abti:
Added Whitelist Blacklist support for Hive view registration in Avro to ORC pipeline. This is important to allow for a subset of dataset's Hive view registration if config store is not enabled.


abti:
minor changes, self approving and merging in.

PR #1788, opened 2017-04-21
Check the late directory again if prevoius compaction is failed

yukuai518:
@htran1 Can you help review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
@@ -24,6 +24,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.collect.Lists;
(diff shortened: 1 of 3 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactorJobPropCreator.java
```diff
@@ -228,42 +230,72 @@ private void addInputLateFilesForFirstTimeCompaction(State jobProps, Dataset dat
...
 
     dataset.setJobProps(jobProps);
     return Optional.of(dataset);
   }
 
+  private boolean isOutputLateDataExists (Dataset dataset) throws IOException {
+    if (!this.fs.exists(dataset.outputLatePath())) {
+      return false;
+    }
+    List<FileStatus> statuses = FileListUtils.listFilesRecursively(this.fs, dataset.outputLatePath());
```
htran1: Can use fs.listStatus() directly and check whether list is empty. This will avoid checking each file.

yukuai518: Done

PR #1789, opened 2017-04-22
Refator HTTP writers

zxcware:
- Introduce a new framework to construct http writers
- Abstract the way to send request(`HttpClient`), the way to build a request(`RequestBuilder`), and the way to handle a response(`ResponseHandler`) of a http writer
- Provide a new Salesforce http writer solution based on the existing solution(`SalesforceRestWriter`).
- Implement `RecordSyncHttpWriter` which processes record synchronously
- Implement `BatchAndSendSyncHttpWriter` which batches records and then sends the batch synchronously.

@ibuenros @yukuai518 Please review.

### DIFF
--- a/gobblin-core/src/main/java/gobblin/writer/http/BatchAndSendSyncHttpWriter.java
+++ b/gobblin-core/src/main/java/gobblin/writer/http/BatchAndSendSyncHttpWriter.java
@@ -1,0 +1,86 @@
+package gobblin.writer.http;
+
+import java.io.IOException;
+import java.util.Collection;
+
+import com.google.common.collect.Sets;
+
+import gobblin.http.BatchRequestBuilder;
+
+
+/**
+ * Synchronous HTTP writer which buffers and sends a batch at a time
+ *
+ * @param <D> type of record
+ * @param <RQ> type of request
+ * @param <RP> type of response
+ */
+public class BatchAndSendSyncHttpWriter<D, RQ, RP> extends HttpWriterBase<D, RQ, RP> {
+  BatchRequestBuilder<D, RQ> batchRequestBuilder;
+  protected final Collection<D> recordSet;
+  protected final int maxBatchSize;
+  protected long numRecordsWritten = 0L;
+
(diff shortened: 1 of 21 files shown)


gobblin-core/src/main/java/gobblin/writer/http/HttpWriterBaseBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Base builder for http writers
+ *
+ * @param <D> type of record
+ * @param <RQ> type of request
+ * @param <RP> type of response
+ * @param <B> type of builder
+ */
+@Getter
+public abstract class HttpWriterBaseBuilder<D, RQ, RP, B extends HttpWriterBaseBuilder<D, RQ, RP, B>>
```
zxcware: Yes. `BatchHttpWriterBaseBuilder` is a subclass of `HttpWriterBaseBuilder`

gobblin-core/src/main/java/gobblin/writer/http/BatchAndSendSyncHttpWriter.java
```diff
@@ -0,0 +1,86 @@
...
+ * @param <RQ> type of request
+ * @param <RP> type of response
+ */
+public class BatchAndSendSyncHttpWriter<D, RQ, RP> extends HttpWriterBase<D, RQ, RP> {
+  BatchRequestBuilder<D, RQ> batchRequestBuilder;
+  protected final Collection<D> recordSet;
+  protected final int maxBatchSize;
+  protected long numRecordsWritten = 0L;
+
+  public BatchAndSendSyncHttpWriter(BatchHttpWriterBaseBuilder builder) {
```
zxcware: For the builder pattern, it doesn't have a rule that a build shouldn't have a getter. A builder is intended to prepare necessary parts for building a complex object. The contract between a builder and an instance is apparent, so it is normal to expose its detail.

yukuai518:
Drop this PR per Zhixiong's request as a new PR (https://github.com/linkedin/gobblin/pull/1819) is set up.

PR #1793, opened 2017-04-24
Percentage string converter

enjoyear:
Add a converter that converts a string which stands for a percentage number ending with %
to a validate double number.


ibuenros:
Hi Chen,
This seems extremely specific. Should we talk about the correct way of doing this kind of transformations?

PR #1794, opened 2017-04-25
Helix cluster creation ability in GobblinServiceManager and Compiler enhancements

abti:
- Added ability to create Helix cluster if not exist in GobblinServiceManager
- Added flow name and group to job name and group translation in Identity Compiler
- Added exception to error log in Orchestrator
- Added tests for Identity compiler with or without template catalog


abti:
Primarily test additions and minor addition to GobblinServiceManager and Compiler. Self approving.

PR #1797, opened 2017-04-25
Avoiding failure of creation of work units due to other exceptions while finding update time

aditya1105:


aditya1105:
Trivial Merging

PR #1798, opened 2017-04-25
Refactoring and dependencies change in gobblin-compliance

aditya1105:


gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/utils/DatasetUtils.java
```diff
@@ -0,0 +1,72 @@
...
+
+import gobblin.compliance.HivePartitionDataset;
+import gobblin.data.management.copy.hive.HiveDataset;
+
+
+/**
+ * @author adsharma
+ */
+@Slf4j
+public class DatasetUtils {
```
ibuenros: Can you add javadoc here?

gobblin-modules/gobblin-azkaban/build.gradle
```diff
@@ -21,6 +21,7 @@ dependencies {
   compile project(":gobblin-api")
   compile project(":gobblin-compaction")
   compile project(":gobblin-data-management")
+  compile project(":gobblin-modules:gobblin-compliance")
```
ibuenros: Why is this needed?

gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/utils/DatasetUtils.java
aditya1105: Added javadoc for each method

gobblin-modules/gobblin-azkaban/build.gradle
aditya1105: This is for my next pr which will use some classes from compliance.

aditya1105: I have removed this dependency. Will add it with next pr only

PR #1800, opened 2017-04-25
Remove redundant runtime attributes in hive table properties

autumnust:
Problem statement:There's a `runtime.props` contained in hive table properties. The value of this props is all the metadata k-v pair that we obtained in runtime, including kafka topic for example. Keeping this value in final metadata is not necessary. This is a minor fix for removal.

@ibuenros Can you help review? Thanks.

### DIFF
--- a/gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
+++ b/gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
@@ -198,7 +198,6 @@ public class HiveMetaStoreUtils {
       if (!propKey.equals(RUNTIME_PROPS)) {
         parameters.put(propKey, props.getProp(propKey));
       }
-      else continue;
     }
     return parameters;
   }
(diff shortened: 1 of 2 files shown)


gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -119,6 +119,9 @@ public static Table getTable(HiveTable hiveTable) {
             propValue + " is not a valid Hive table/partition property");
         table.getParameters().put(tokens.get(0), tokens.get(1));
       }
+      if (table.getParameters().containsKey(RUNTIME_PROPS)) {
```
ibuenros: Instead of removing it here, can you figure out why the parameter is being added in the first place and remove it there?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -194,8 +185,20 @@ public static HivePartition getHivePartition(Partition partition) {
...
+            propValue + " is not a valid Hive table/partition property");
+        parameters.put(tokens.get(0), tokens.get(1));
+      }
+    }
     for (String propKey : props.getPropertyNames()) {
-      parameters.put(propKey, props.getProp(propKey));
+      if (!propKey.equals(RUNTIME_PROPS)) {
+        parameters.put(propKey, props.getProp(propKey));
+      }
+      else continue;
```
ibuenros: this is not needed

PR #1801, opened 2017-04-25
Fix for #1792

### LINKED ISSUE #1792: JobScheduler failed when "jobconf.fullyQualifiedPath" is not set.
In JobScheduler.java, seems like there is a bug where jobConfigFileDirPath does not get set if "jobconf.fullyQualifiedPath" is not configured. Seems like this should be supported as there is a check in JobScheduler.java.startUp which sets "jobconf.fullyQualifiedPath".

kadaan:
Fixed #1792 by moving the handling of the case where `ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY` exists, but `ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY` does not to the constructor.  In this case, the `ConfigurationKeys.JOB_CONFIG_FILE_GENERAL_PATH_KEY` is added to the property set with the value `"file://" + this.properties.getProperty(ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY)`.  Then during startup, if properties contains `ConfigurationKeys.JOB_CONFIG_FILE_DIR_KEY`, the job file monitor is started.

### DIFF
--- a/gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
+++ b/gobblin-runtime/src/main/java/gobblin/scheduler/JobScheduler.java
@@ -18,6 +18,8 @@
 package gobblin.scheduler;
 
 import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


kadaan:
@autumnust Ping

kadaan:
ty

PR #1808, opened 2017-04-26
Add hive registration for compaction

yukuai518:


gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionHiveRegistrationAction.java
```diff
@@ -1,18 +1,42 @@
 package gobblin.compaction.action;
 
+import gobblin.compaction.parser.CompactionPathParser;
+import gobblin.configuration.ConfigurationKeys;
```
htran1: Gobblin import order convention is java, org, com, gobblin.

PR #1809, opened 2017-04-26
Throttling HA

ibuenros:
This PR is 95% around throttling server HA, with some other changes that were used in previous attempts at HA but are useful either way.

The general idea of HA is as follows:
* Multiple Throttling servers coordinate through Zookeeper to elect a leader.
* Any request to a non-leader returns a 301 redirect to the leader.
* Clients use as input a list of URIs in the cluster (in the future will use D2 uri instead) and chose an initial URI at random.
* Clients transparently process 301 redirects and repeat the request to the new leader up to a maximum number of redirects. The redirect is sticky, i.e. the client remembers the active leader, and all future requests will go to that leader (until a new redirect is received).
* If the client cannnot connect to the stored host, it resets by round-robin'ing over the list of input URIs.

Major changes:
===========
Server side
----------
* Introduction of `LeaderFinder`, an interface allowing a cluster of processes to elect a leader, and allowing each process to get metadata on the leader.
* Introduction of `ZookeeperLeaderElection`, a `LeaderFinder` based on Zookeeper leader election.
* The `ThrottlingGuiceServletConfig`, which is the `ServletContextListener` used to configure the throttling server, was changed to automatically instantiate a `LeaderFinder` if a zk string is provided. Also, it becomes stateful even when not launching through a servlet (for example, when launching through `EmbeddedRestliServer`.

Client side
----------
* Introduction/refining of `RequestSender`, which abstracts the sending of a `PermitRequest` to the throttling server. This also allows easily injecting Mock servers for testing.
* `RestClientRequestSender` uses a `RestClient` to send a permit request.
* `RedirectAwareRestClientRequestSender` is a `RequestSender` than transparently handles redirects, sticky routing, and failures to communicate with the current stored uri.
* `ExponentialBackoff` is a class used to implement generic exponential backoff. Put it in gobblin-utility because it is not the first time I wish we had one of these.


gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-server/src/main/java/gobblin/restli/throttling/ThrottlingGuiceServletConfig.java
```diff
@@ -52,34 +62,56 @@
...
+
   @Override
-  protected Injector getInjector() {
-    return getInjector(this._brokerConfig);
+  public Injector getInjector() {
+    return this._injector;
   }
 
-  public static Injector getInjector(final Config brokerConfig) {
+  private Injector createInjector(final Config config, final Optional<LeaderFinder<URIMetadata>> helixManager) {
```
htran1: Since helix is not used, maybe change all references to Helix to a more generic name.

gobblin-utility/src/main/java/gobblin/util/ExponentialBackoff.java
```diff
@@ -0,0 +1,123 @@
...
+  private final long maxDelay;
+
+  private int retryNumber = 0;
+  private long nextDelay;
+
+  /**
+   * @param alpha the multiplier for each backoff iteration. (def: 2)
+   * @param maxRetries maximum number of retries allowed. (def: infinite)
+   * @param maxDelay maximum delay allowed in millis. (def: infinite)
+   * @param initialDelay initial delay. (def: 20)
```
htran1: Put units here.

PR #1810, opened 2017-04-26
Change check for empty schedule in the IdentityFlowToJobSpecCompilerT…

htran1:
…est test.


htran1:
Trivial change for an assert condition.

PR #1811, opened 2017-04-26
Locks directory clean up for gobblin standalone

enjoyear:
We need to clean up the locks directory if gobblin cannot be stopped gracefully, otherwise, it will hold all following schedules for any job in the locks folder from running the next time we restart the gobblin instance locally.

### DIFF
--- a/bin/gobblin-standalone.sh
+++ b/bin/gobblin-standalone.sh
@@ -213,6 +213,8 @@ stop() {
       if kill -0 $PID_VALUE > /dev/null 2>&1; then
         echo "Gobblin standalone daemon did not stop gracefully, killing with kill -9"
         kill -9 $PID_VALUE
+        echo "Cleaning up lock directory: $GOBBLIN_WORK_DIR/locks"
+        rm -r $GOBBLIN_WORK_DIR/locks
       fi
     else
       echo "Process $PID_VALUE is not running"


bin/gobblin-standalone.sh
```diff
@@ -213,6 +213,8 @@ stop() {
       if kill -0 $PID_VALUE > /dev/null 2>&1; then
         echo "Gobblin standalone daemon did not stop gracefully, killing with kill -9"
         kill -9 $PID_VALUE
+        echo "Cleaning up lock directory: $GOBBLIN_WORK_DIR/locks"
+        rm -r $GOBBLIN_WORK_DIR/locks
```
abti: rm -r $GOBBLIN_WORK_DIR/locks/* ??

enjoyear: locks folder will be auto-generated unless you want to keep an empty locks folder.

ibuenros:
I think this is dangerous. In case of misconfiguration, duplicate jobs could be launched. Please do not merge until we come up with the correct solution.

enjoyear:
@ibuenros , under what conditions will there be misconfigurations? That "remove locks" will only be executed when people call "gobblin-standalone.sh stop".

bin/gobblin-standalone.sh
abti: Thanks for confirming.

PR #1812, opened 2017-04-26
Added flow execution id to the jobSpecs compiled for a flowSpec

abti:
Added flow execution id to the jobSpecs compiled for a flowSpec and related test.


abti:
Pair reviewed with Hung. Merging in.

PR #1813, opened 2017-04-27
Add metadata tags for associating events to a flow and job execution.

htran1:


htran1:
@abti, please take a look.

PR #1816, opened 2017-04-28
changing writer staging and output directories to job level sub direc…

arjun4084346:
changing writer staging and output directories to new job level sub directories, so that two jobs with same ConfigurationKeys.WRITER_STAGING_DIR  .WRITER_OUTPUT_DIR in their job conf do not clean up the common staging and output directories.

@ydai1124 @ibuenros


gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -306,14 +320,18 @@ protected void setTaskStagingAndOutputDirs() {
    * plus {@link #TASK_STAGING_DIR_NAME}.
    */
   private void setTaskStagingDir() {
+    String workingDir;
     if (this.jobState.contains(ConfigurationKeys.WRITER_STAGING_DIR)) {
```
ydai1124: This condition should be a NOOP. This method only takes effect when `WRITER_STAGING_DIR` is not specified.

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -292,6 +292,20 @@ protected void setTaskStagingAndOutputDirs() {
           new Path(this.jobState.getProp(ConfigurationKeys.TASK_DATA_ROOT_DIR_KEY), this.jobId).toString();
       this.jobState.setProp(ConfigurationKeys.TASK_DATA_ROOT_DIR_KEY, taskDataRootDirWithJobId);
 
+      // Add jobId to writer staging dir
```
ydai1124: What if a user only specifies `WRITER_STAGING_DIR` and `WRITER_OUTPUT_DIR`, but not`TASK_DATA_ROOT_DIR_KEY`, will you fix still work?

arjun4084346:
Incorporated review comments. @ydai1124

gobblin-runtime/src/main/java/gobblin/runtime/JobContext.java
```diff
@@ -306,13 +320,12 @@ protected void setTaskStagingAndOutputDirs() {
    * plus {@link #TASK_STAGING_DIR_NAME}.
    */
   private void setTaskStagingDir() {
-    if (this.jobState.contains(ConfigurationKeys.WRITER_STAGING_DIR)) {
-      LOG.warn(String.format("Property %s is deprecated. No need to use it if %s is specified.",
```
ydai1124: Can you put back this warning message?

PR #1818, opened 2017-05-01
Fix maven publishing for gobblin-restli api projects.

ibuenros:


ibuenros:

PR #1819, opened 2017-05-01
Introduce async http writer framework

zxcware:
- Introduce pluggable components for http writers, `HttpClient`,`AsyncWriteRequestBuilder`, `ResponseHandler`
- Implement `AsyncDataDispatcher` for general async data processing
- Implement basic classes for async http writers


gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+    notEmpty = lock.newCondition();
+    isEmpty = lock.newCondition();
+    notFull =  lock.newCondition();
+    buffer = new ArrayDeque<>(capacity);
+    this.capacity = capacity;
+    isClosing = false;
+  }
+
+  /**
+   * Dispatch the record in the buffer. Retries should be done if necessary
```
ibuenros: This should be extremely clear about what the contract is. For example, the assumption is that this method is blocking, every record consumed from the buffer is processed, and for every record consumed the callback is called.

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+          try {
+            notEmpty.await();
+          } catch (InterruptedException e) {
+            LOG.info("Waiting for queue not empty interrupted", e);
+          }
+        }
+
+        // Dispatch records
+        try {
+          if (dispatch(buffer) < 0) {
```
ibuenros: Should this be `<` or `<=`?

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+            throw new IOException("Failed on dispatching");
```
ibuenros: This is not a useful exception unless we know why exactly it failed.

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+          }
+        } catch (Throwable throwable) {
+          LOG.error("RecordProcessor breaks on exception", throwable);
+          throw new RuntimeException("Exception on dispatching", throwable);
```
ibuenros: This will kill the thread, is that the behavior you want?

gobblin-core/build.gradle
```diff
@@ -56,6 +56,7 @@ dependencies {
   compile externalDependency.findBugsAnnotations
   compile externalDependency.oltu
   compile externalDependency.opencsv
+  compile externalDependency.pegasus.d2
```
ibuenros: Let's just create a separate module for this. We need to try not to add unnecessary dependencies to `gobblin-core` as we move forward.

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
```diff
@@ -0,0 +1,84 @@
...
+          LOG.error("Write failed on IOException", e);
+          break;
+        } else {
+          continue;
+        }
+      }
+
+      ResponseStatus status = responseHandler.handleResponse(response);
+      int statusCode = status.getStatusCode();
+      if (statusCode >= 200 && statusCode < 400) {
```
ibuenros: Is a `3xx` considered a success?

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
```diff
@@ -0,0 +1,84 @@
...
+        // Write succeeds
+        asyncWriteRequest.onSuccess(WriteResponse.EMPTY);
+        return asyncWriteRequest.getRecordCount();
+      } else if (statusCode >= 400 && statusCode < 500) {
+        // Client error. Fail!
+        LOG.error("Write failed on invalid request");
```
ibuenros: You should call the failed callback here.

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
```diff
@@ -0,0 +1,84 @@
...
+  protected int dispatch(Queue<BufferedRecord<D>> buffer) throws Throwable {
+    AsyncWriteRequest<D, RQ> asyncWriteRequest = requestBuilder.buildRequest(buffer);
+    if (asyncWriteRequest == null) {
+      return 0;
+    }
+
+    RQ rawRequest = asyncWriteRequest.getRawRequest();
+    RP response;
+
+    int i = 0;
```
ibuenros: rename to something more meaningful (e.g. `retries`)

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
```diff
@@ -0,0 +1,84 @@
...
+        // Write succeeds
+        asyncWriteRequest.onSuccess(WriteResponse.EMPTY);
+        return asyncWriteRequest.getRecordCount();
+      } else if (statusCode >= 400 && statusCode < 500) {
+        // Client error. Fail!
+        LOG.error("Write failed on invalid request");
+        return -1;
+      } else {
+        // Server side error. Retry
+        i++;
```
ibuenros: You should probably check whether you've done max retries? And call failed callback?

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
zxcware: Good catch!

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
zxcware: Good catch!

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+
+  public AbstractAsyncDataWriter(int capacity) {
+    service = Executors.newFixedThreadPool(1);
+    recordProcessorFuture = service.submit(new RecordProcessor());
+
+    lock = new ReentrantLock();
+    notEmpty = lock.newCondition();
+    isEmpty = lock.newCondition();
+    notFull =  lock.newCondition();
+    buffer = new ArrayDeque<>(capacity);
```
zxcware: The reasons are:
1. We want to move the logic of waiting for non empty queue to the `AbstractAsyncDataWriter` layer, for it is shared among all async data writers. The existing blocking queue doesn't provide this functionality without actually taking out an element.
2. Using a blocking queue, every `RequestBuilder` has to maintain some concurrent exception such as `Interrupted Exception`, which introduces too much common overhead and should be taken care of at a higher level as `AbstractAsyncDataWriter`
3. By maintaining a lock, we can introduce an `isEmpty` condition on which a `flush` call can wait. It's a simpler primitive than `CountUpDown` latch

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,179 @@
...
+    isClosing = false;
+  }
+
+  /**
+   * Dispatch the record in the buffer. Retries should be done if necessary
+   *
+   * @param buffer The buffer which contains a collection of records
+   * @return the number of records are successfully dispatched or -1 if dispatch failed
+   */
+  protected abstract int dispatch(Queue<BufferedRecord<D>> buffer) throws Throwable;
```
zxcware: However the actual dispatch consumes the buffer, whether it consumes one or a batch, is totally up to the subclass implementation. The clarification should be made in subclass not in this abstract class.

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
zxcware: You're right. I'll add blocking and callback invocation.

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
zxcware: Yes. Since dispatch may have already retried, all subsequent process would fail for the same reason, meaning the processor is in a bad state and there is no point to keep it alive.

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
yukuai518: We can provide this default behavior, but anyone who derived this class should be able to override.

zxcware: I'll treat `3xx` as client error per https://www.webnots.com/3xx-http-status-codes/. This behavior can be overridden.

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+      LOG.error("Write rejected", e);
+      throw e;
+    }
+
+    while (true) {
+      try {
+        buffer.put(record);
+        break;
+      } catch (InterruptedException e) {
+        LOG.debug("Waiting to put a record interrupted", e);
```
ibuenros: Generally an `InterruptedException` will happen if the thread is being shut down, so it tends to be better to stop processing at that point. (https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html )

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * <p>
+ *   However the records are consumed depends on the actual implementation of {@link #dispatch(Queue)}, which
+ *   may process one record or a batch at a time
+ * </p>
+ *
+ * @param <D> type of record
+ */
+@ThreadSafe
+public abstract class AsyncDataDispatcher<D> implements Closeable {
```
ibuenros: This seems like an `AbstractIdleService` rather than a `Closeable`, which also provides the `ProcessorState` for free.

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+  public void close()
+      throws IOException {
+    flush();
+    state = ProcessorState.CLOSING;
+    try {
+      recordProcessorFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.info("Waiting for RecordProcessor returns an exception", e);
+    }
+    service.shutdown();
```
ibuenros: You might want to block until shutdown finishes? `awaitTermination()`

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  public void flush()
```
ibuenros: Please add javadoc with the expected behavior

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,149 @@
...
+          try {
+            Thread.sleep(300);
+          } catch (InterruptedException e) {
+            LOG.debug("RecordProcessor sleep interrupted", e);
+          }
+        }
+
+        // Dispatch records
+        try {
+          dispatch(buffer);
```
ibuenros: This will call dispatch even if the buffer is empty, is that the expected behavior?

gobblin-core/src/main/java/gobblin/async/DispatchException.java
```diff
@@ -0,0 +1,18 @@
...
+public class DispatchException extends Exception {
+  public DispatchException(String message, Exception e) {
+    super(message, e);
+  }
+
+  public DispatchException(String message) {
+    super(message);
+  }
+
+  public boolean isFatal() {
```
ibuenros: Why not let the constructor specify whether the exception is fatal?

gobblin-core/src/main/java/gobblin/http/ResponseHandler.java
```diff
@@ -0,0 +1,10 @@
+package gobblin.http;
+
+/**
+ * An interface to handle a response
+ *
+ * @param <RP> type of response
+ */
+public interface ResponseHandler<RP> {
```
ibuenros: Should this also be templated on the `ResponseStatus` type?

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
```diff
@@ -0,0 +1,90 @@
...
+        attempt++;
+        if (attempt == maxAttempts) {
+          asyncWriteRequest.onFailure(e);
+          throw new DispatchException("Write failed on IOException", e);
+        } else {
+          continue;
+        }
+      }
+
+      ResponseStatus status = responseHandler.handleResponse(response);
```
ibuenros: It might be better to include an enum response status type in `ResponseStatus`. This lets the responseHandler figure out what the response means (for example, and unexpected 2xx response code could mean we're sending to the wrong endpoint).

gobblin-core/src/main/java/gobblin/writer/http/RecordSyncHttpWriter.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Synchronous HTTP writer which sends one record at a time
+ *
+ * @param <D> type of record
+ * @param <RQ> type of request
+ * @param <RP> type of response
+ */
+@Log4j
+public class RecordSyncHttpWriter<D, RQ, RP> extends HttpWriterBase<D, RQ, RP> {
```
ibuenros: What is the use case for this implementation?

gobblin-core/src/main/java/gobblin/http/ResponseHandler.java
zxcware: That works but it's not necessary. This interface is only used in http domain. Currently, `ResponseStatus` only has the status code. It can be subclassed as any other type, which can do whatever a template does. Returning `ResponseStatus` facilitates add common logic after handling the response in a super class and reduces the number of generics.

gobblin-core/src/main/java/gobblin/writer/http/RecordSyncHttpWriter.java
zxcware: Process a record synchronously. Writing to old version of salesforce needs it as well as writing to http services that don't support batch

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
zxcware: Ah.. I forgot to `while` waits on empty buffer.

gobblin-core/src/main/java/gobblin/writer/http/RecordSyncHttpWriter.java
ibuenros: Writing a single record is different from writing a record synchronously. Old version of salesforce could simply create batches of one record.

zxcware: Old version of salesforce doesn't support batch. `RecordSyncHttpWriter` is to write a record synchronously.

gobblin-core/src/main/java/gobblin/writer/http/AsyncHttpWriter.java
zxcware: A `StatusType` enum is introduced

gobblin-core/src/main/java/gobblin/writer/http/RecordSyncHttpWriter.java
ibuenros: Again, sync vs async is independent of the size of the batch. Single record is a special case of batched writer where the `RequestBuilder` only takes one element. The point here is simply to remove confusion by removing functionally identical interfaces.

gobblin-core/src/main/java/gobblin/async/AsyncDataDispatcher.java
```diff
@@ -0,0 +1,167 @@
...
+      throw e;
+    }
+  }
+
+  protected void waitForABufferEmptyOccurrence() {
+    checkRunning("waitForABufferEmptyOccurrence");
+    try {
+      lock.lock();
+      // Waiting for a buffer empty occurrence
+      while (!isBufferEmptyOccurred) {
```
ibuenros: why do you need this variable? It seems you could just check `buffer.isEmpty`?

gobblin-core/src/test/java/gobblin/async/AsyncDataDispatcherTest.java
```diff
@@ -0,0 +1,237 @@
...
+
+    ExecutorService service = Executors.newFixedThreadPool(2);
+    Writer writer1 = new Writer(dispatcher);
+    Writer writer2 = new Writer(dispatcher);
+    Future<?> future1 = service.submit(writer1);
+    Future<?> future2 = service.submit(writer2);
+
+    // Do some process
+    try {
+      Thread.sleep(200);
```
ibuenros: Try not to use specific sleeps in tests, as they may fail in slower or more contested systems. In general, it is better to await for a condition with a relatively long timeout (couple seconds). You can use for example `ExponentialBackoff.awaitCondition()`

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public AbstractAsyncDataWriter(int capacity) {
+    super(capacity);
+  }
+
+  @Override
+  public Future<WriteResponse> write(D record, @Nullable WriteCallback callback) {
+    BufferedRecord<D> bufferedRecord = new BufferedRecord<>(record, callback);
+    put(bufferedRecord);
+    return null;
```
shirshanka: why are you returning null here?

zxcware: We'll use the callback. It's not necessary to maintain 2 kinds of async semantics at the same time. I'll add some comments.

gobblin-utility/src/main/java/gobblin/util/ExponentialBackoff.java
```diff
@@ -33,22 +33,26 @@
...
+   * @param maxWait maximum wait allowed in millis. (def: infinite)
    * @param maxDelay maximum delay allowed in millis. (def: infinite)
    * @param initialDelay initial delay in millis. (def: 20)
    */
   @Builder
-  private ExponentialBackoff(Double alpha, Integer maxRetries, Long maxDelay, Long initialDelay) {
+  private ExponentialBackoff(Double alpha, Integer maxRetries, Long maxWait, Long maxDelay, Long initialDelay) {
     this.alpha = alpha == null ? 2 : alpha;
     this.maxRetries = maxRetries == null ? Integer.MAX_VALUE : maxRetries;
+    this.maxWait = maxWait == null ? Integer.MAX_VALUE : maxWait;
```
ibuenros: Use `Long.MAX_VALUE`

gobblin-core/src/main/java/gobblin/writer/http/AbstractAsyncDataWriter.java
zxcware: Fixed by using `FutureWrappedWriteCallback`

PR #1825, opened 2017-05-02
Config store based throttling policy.

ibuenros:


gobblin-utility/src/main/java/gobblin/util/limiter/broker/SharedLimiterKey.java
```diff
@@ -55,6 +60,7 @@ public String toString() {
 
   @Override
   public String toConfigurationKey() {
-    return this.resourceLimited;
+    // remove leading "/"
+    return this.resourceLimitedPath.substring(1).replace("/", ".");
```
ydai1124: Doesn't `substring(1)` already remove the leading "/"?

ibuenros: There may be non-leading "/"

PR #1826, opened 2017-05-02
Create workunitstream for CompactionSource

yukuai518:


gobblin-compaction/src/main/java/gobblin/compaction/source/CompactionSource.java
```diff
@@ -84,6 +91,135 @@
...
+          if (last != null) {
+            log.debug ("hasNext() is true because cache is not empty");
+            return true;
+          }
+          if (this.unProcessed > 0) {
+            last = this.workUnits.poll(1, TimeUnit.SECONDS);
+            if (last == null)
+              log.debug ("Waiting for producer to complete...");
+          } else {
+            if (workUnits.size() > 0) {
```
ibuenros: it seems you should add `last = this.workUnits.poll()` here?

gobblin-compaction/src/main/java/gobblin/compaction/source/CompactionSource.java
```diff
@@ -84,6 +91,135 @@
...
+    }
+
+    /**
+     * Check if any {@link WorkUnit} is available. The producer is {@link CompactionDatasetIterator#getDatasetProcessor()}
+     * @return true when a new {@link WorkUnit}  is available
+     *         false when producer exits
+     */
+    public boolean hasNext () {
+      try {
+        while (true) {
```
ibuenros: Would this work?
```java
while(true) {
   if (last != null) return true;
   if (this.unProcessed <= 0 && this.workUnits.isEmpty()) return false;
   this.last = this.workUnits.poll(1, TimeUnit.SECONDS);
}
```

gobblin-compaction/src/main/java/gobblin/compaction/source/CompactionSource.java
```diff
@@ -84,6 +91,135 @@
...
+     * Block the consumer thread until a new {@link WorkUnit} is provided. Otherwise throw an exception
+     */
+    public WorkUnit next () {
+      if (hasNext()) {
+        if (last != null) {
+          WorkUnit tmp = last;
+          last = null;
+          log.debug ("next() pops out a workunit");
+          return tmp;
+        }
```
ibuenros: you should probably have `else { throw new IllegalStateException()} ` to catch the case where `hasNext()` returns true but `last` doesn't get populated.

gobblin-compaction/src/main/java/gobblin/compaction/source/CompactionSource.java
```diff
@@ -84,6 +91,135 @@
...
+      initJobDir(state);
+      copyJarDependencies(state);
+      List<CompactionVerifier> verifiers = suite.getDatasetsFinderVerifiers();
+      DatasetsFinder finder = DatasetUtils.instantiateDatasetFinder(state.getProperties(),
+              getSourceFileSystem(state),
+              DefaultFileSystemGlobFinder.class.getName());
+
+      List<Dataset> datasets = finder.findDatasets();
+      CompactionDatasetIterator iterator = new CompactionDatasetIterator(datasets, verifiers);
+      service.submit(iterator.getDatasetProcessor());
```
ibuenros: You probably want multiple data processors right? Note that will also require to transform the `List` into something like a `Spliterator` (although that is only available on java 8)

gobblin-compaction/src/main/java/gobblin/compaction/source/CompactionSource.java
```diff
@@ -18,32 +18,39 @@
...
 @Slf4j
-public class CompactionSource implements Source<String, String> {
+public class CompactionSource implements WorkUnitStreamSource<String, String> {
   private CompactionSuite suite;
   private Path tmpJobDir;
   private FileSystem fs;
+  private ExecutorService service;
 
+  @Override
   public List<WorkUnit> getWorkunits(SourceState state) {
```
ibuenros: You should remove `getWorkUnits(SourceState)`, it is better to only have to maintain one functionally equivalent method.

yukuai518: Fixed

ibuenros:
@yukuai518 what is the status of this PR?

abti:
Issue: https://issues.apache.org/jira/browse/GOBBLIN-38

Please update your PR title with following prefix: [GOBBLIN-38]

PR #1827, opened 2017-05-03
Preserving partition params during avro2orc conversion

aditya1105:
1) Passing the necessary source partition info and destination partition info from HiveQueryExecutionWriter.java to HiveConvertPublisher.java via WorkUnit
2) Params which need to be preserved should be passed via configuration property "hive.conversion.preserve.partitionParameters"
3) Adding params into the destination partition inside HiveConvertPublisher.java

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.thrift.TException;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
@@ -232,24 +233,7 @@ public class HiveConvertPublisher extends DataPublisher {
       /////////////////////////////////////////
       // Preserving partition params
       /////////////////////////////////////////
-      for (WorkUnitState wus : states) {
-        if (wus.getWorkingState() != WorkingState.COMMITTED) {
-          continue;
-        }
-        if (!wus.contains(COMPLETE_SOURCE_PARTITION_NAME)) {
-          continue;
-        }
-        if (!wus.contains(COMPLETE_DEST_PARTITION_NAME)) {
-          continue;
-        }
-        if (!(wus.contains(PARTITION_PARAMETERS_WHITELIST) || wus.contains(PARTITION_PARAMETERS_BLACKLIST))) {
-          continue;
-        }
-        List<String> whitelist = COMMA_SPLITTER.splitToList(wus.getProp(PARTITION_PARAMETERS_WHITELIST, StringUtils.EMPTY));
-        List<String> blacklist = COMMA_SPLITTER.splitToList(wus.getProp(PARTITION_PARAMETERS_BLACKLIST, StringUtils.EMPTY));
-        copyPartitionParams(wus.getProp(COMPLETE_SOURCE_PARTITION_NAME), wus.getProp(COMPLETE_DEST_PARTITION_NAME),
-            whitelist, blacklist);
-      }
+      preservePartitionParams(states);
 
       /////////////////////////////////////////
       // Post publish cleanup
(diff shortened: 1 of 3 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
```diff
@@ -227,6 +261,57 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+      return;
+    }
+    String sourceDbName = sourcePartitionList.get(0);
+    String sourceTableName = sourcePartitionList.get(1);
+    String sourcePartitionName = sourcePartitionList.get(2);
+    String destDbName = destPartitionList.get(0);
+    String destTableName = destPartitionList.get(1);
+    String destPartitionName = destPartitionList.get(2);
+
+    Optional<HiveMetaStoreClient> clientOptional = Optional.<HiveMetaStoreClient>absent();
```
abti: Recreating a new HiveMetaStoreClient at scale might slow down the publisher significantly. Why not create a HiveMetaStoreClientPool and use / release clients as required from it. Thus reusing the established connection.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
```diff
@@ -227,6 +261,57 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
+      clientOptional = Optional.fromNullable(new HiveMetaStoreClient(new HiveConf()));
+      HiveMetaStoreClient client = clientOptional.get();
+
+      Partition sourcePartition = client.getPartition(sourceDbName, sourceTableName, sourcePartitionName);
+      Partition destPartition = client.getPartition(destDbName, destTableName, destPartitionName);
+
+      Map<String, String> sourceParams = sourcePartition.getParameters();
+      Map<String, String> destParams = destPartition.getParameters();
+
+      for (String param : params) {
```
abti: Can we introduce a mode where all properties are carry forward as-is. So that the operator might not be required to know before hand of all the params to copy over. 
(Think of different heterogenous datasets that we support internally at LinkedIn - metrics, derived, external data, etc. Each might need this feature but might have different params per dataset)

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
aditya1105: Since the code is inside the publisher shouldn't refactoring it out and initializing it only once should suffice.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
aditya1105: Will regex match suffice?? There are properties inside partition parameters which shouldn't be copied like creation time, update time etc

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
```diff
@@ -79,6 +88,15 @@
   private EventSubmitter eventSubmitter;
   private final FileSystem fs;
   private final HiveSourceWatermarker watermarker;
+  private final HiveMetastoreClientPool pool;
+
+  public static final String PARTITION_PARAMETERS_WHITELIST = "hive.conversion.partitionParameters.whitelist";
```
abti: How do you control dataset level whitelist / blacklist for this? This seems to read from WU - which seems to be coming from pull file properties, how about enabling config-store for this? 

nit: keys can go to AvroToORC config key class

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
```diff
@@ -227,6 +269,71 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
     }
   }
 
+  /**
+   * Method to copy partition parameters from source partition to destination partition
+   * @param completeSourcePartitionName dbName@tableName@partitionName
+   * @param completeDestPartitionName dbName@tableName@partitionName
+   */
+  private void copyPartitionParams(String completeSourcePartitionName, String completeDestPartitionName,
```
abti: Can we add couple of tests for these. Might be good to have them to lock in behavior along with increasing coverage / verifying functionality.

gobblin-data-management/src/main/java/gobblin/data/management/conversion/hive/publisher/HiveConvertPublisher.java
aditya1105: After enabling config store I will add properties to workunit so it will still be ok

aditya1105:
@abti I added the test, Can you review again?

aditya1105:
Merging. Will make changes later if needed.

PR #1830, opened 2017-05-03
Small fixes for steaming kafka extractor and converter

ydai1124:
@htran1 address issues for gaas monitoring. can you take a look?

### DIFF
--- a/gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingExtractor.java
+++ b/gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingExtractor.java
@@ -17,48 +17,43 @@
 
 package gobblin.source.extractor.extract.kafka;
 
-import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


gobblin-modules/gobblin-kafka-09/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSimpleStreamingExtractor.java
```diff
@@ -203,15 +210,17 @@ public S getSchema() throws IOException {
...
       throw new IOException("Streaming extractor has not been started.");
     }
     while ((_records == null) || (!_records.hasNext())) {
       synchronized (_consumer) {
-        if (_close.get())
+        if (_close.get()) {
           throw new ClosedChannelException();
-        _records = _consumer.poll(100).iterator();
+        }
+        _records = _consumer.poll(10000).iterator();
```
htran1: This should be configurable. Maybe use a key similar to the one in AbstractBaseKafkaConsumerClient. CONFIG_KAFKA_FETCH_TIMEOUT_VALUE.

ydai1124:
@htran1 Comments addressed.

PR #1831, opened 2017-05-04
Generic RecordAccessor Interface + Avro implementation

eogren:
Add a RecordAccessor that should allow field-level converters to modify an underlying record without directly instantiating AvroConverter,JsonConverter,FooConverter, etc.

This will be used by a FieldLevelEncryption converter which will be coming soon, but has value for other use cases as well (eg: Partitioners that can delegate to the RecordAccessor to actually retrieve the field).


eogren:
@ibuenros @htran1 can you take a look?

gobblin-api/src/main/java/gobblin/recordaccess/RecordAccessor.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * This interface will likely grow over time with more accessors/setters as well as
+ * schema manipulation (for example: rename field or delete field operations).
+ */
+@Alpha
+public interface RecordAccessor {
+  /*
+   * Getters should return null if the field does not exist; may return
+   * ClassCastException if the underlying types do not match. Getters should not
+   * try to do any type coercion -- for example, getAsInt for a value that is the string "1"
```
ibuenros: Why? There are string-only formats (like some `Maps`) so it might be good to allow some implementations to do type coercion.

gobblin-api/src/main/java/gobblin/recordaccess/RecordAccessor.java
```diff
@@ -0,0 +1,59 @@
...
+   * should throw a ClassCastException.
+   */
+  String getAsString(String fieldName);
+  Integer getAsInt(String fieldName);
+  Long getAsLong(String fieldName);
+
+  /*
+   * Set new values for an object. Should throw a FieldDoesNotExist runtime exception if fieldName
+   * is not present in the object's schema.
```
ibuenros: What exception should implementations throw if the field is not of the expected type?

eogren: ah good point, I will add another exception for this.

gobblin-api/src/main/java/gobblin/recordaccess/RecordAccessor.java
eogren: My thought here was to avoid confusion about the underlying schema -- with type coercion getAsInt("stringFieldThatContainsAnInt") could succeed sometimes but fail sometimes depending on record value. Also means that we don't have to worry about type coercion on the set side -- set("stringFieldThatContainsAnInt", 1) may end up failing because Int is the wrong underlying type. What do you think?

PR #1832, opened 2017-05-04
Added getter for Config in GobblinServiceManager

abti:
Added getter for Config in GobblinServiceManager


abti:
Trivial change, merging.

PR #1834, opened 2017-05-05
Added Instrumentation to Gobblin Service

abti:
Added basic Instrumentation to the following Gobblin Service modules: 
- Compiler 
- Orchestrator


abti:
Trivial change, self approving.

PR #1838, opened 2017-05-08
Start the SpecExecutorInstanceConsumer in the StreamingJobConfigurati…

htran1:
…on manager if it is a service.


htran1:
@abti, please review.

PR #1839, opened 2017-05-08
Stream based record processing.

ibuenros:
This is an overhaul of record processing in ETL tasks in Gobblin. The major changes are:

* Make record metadata a first-class citizen of Gobblin. Instead of passing raw records around the pipeline, it now uses `RecordEnvelope`s containing the record, metadata, and relevant callbacks. `RecordEnvelope`s are transparent to all constructs that don't care about them.
* Callbacks are first-class citizens. Each record has a collection of callbacks that get called whenever the record is written, dropped, or failed. Any construct can add callbacks to the record.
* The processing of records is based on java `Stream`s instead of a synchronous record-by-record processing. This will allow, in the future, to do async/out-of-order record processing.


ibuenros:
@shirshanka @eogren @htran1 can you take a look when you have time?

ibuenros:
@eogren @htran1 can I get a review for this? It will get harder to commit as time passes.

PR #1844, opened 2017-05-08
Field-level string encryptor. String -> string is the only supported

eogren:
mode right now so that we don't have to manipulate the output schema.

### DIFF
--- a/gobblin-modules/gobblin-crypto-provider/src/main/java/gobblin/converter/StringFieldEncryptorConverter.java
+++ b/gobblin-modules/gobblin-crypto-provider/src/main/java/gobblin/converter/StringFieldEncryptorConverter.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


gobblin-modules/gobblin-crypto-provider/src/main/java/gobblin/converter/StringFieldEncryptorConverter.java
```diff
@@ -0,0 +1,101 @@
...
+import gobblin.crypto.EncryptionFactory;
+import gobblin.recordaccess.RecordAccessor;
+
+
+/**
+ * Converter that can encrypt a string field in place. Assumes that the encryption algorithm chosen will output
+ * a UTF-8 encoded byte array.
+ */
+public abstract class StringFieldEncryptorConverter<SCHEMA, DATA> extends Converter<SCHEMA, SCHEMA, DATA, DATA> {
+  private static final String FIELDS_TO_ENCRYPT_CONFIG_NAME = "converter.fieldsToEncrypt";
```
shirshanka: public static (in case you want to reference this constant from a different place... e.g. a test?)

gobblin-modules/gobblin-crypto-provider/src/main/java/gobblin/converter/StringFieldEncryptorConverter.java
```diff
@@ -0,0 +1,101 @@
...
+
+    fieldsToEncrypt = Splitter.on(',').splitToList(fieldsToEncryptConfig);
+
+    return this;
+  }
+
+  @Override
+  public SCHEMA convertSchema(SCHEMA inputSchema, WorkUnitState workUnit)
+      throws SchemaConversionException {
+    return inputSchema;
```
shirshanka: This implies we are not supporting changing the field name of the encrypted column in this converter. Are you thinking that it would be handled in a different converter?

eogren: Yes, that's correct, although I could see a case for embedding that converter inside this one as well -- particularly in a use case where you are renaming a field and changing its type at the same time.

PR #1854, opened 2017-05-09
Add multiGet support to the recordAccessor API; update the StringFieldEncryptor to use it.

eogren:
I thought about collapsing the getMultiAsFoo() and getAsFoo() APIs together, but I think keeping the APIs separate makes it much easier for clients.


eogren:
Merging to get this in before vacation, but will appreciate post-commit review!

PR #1858, opened 2017-05-10
Changes to PathAlterationDetector/PathAlterationObserverScheduler

chavdar:
- Rename PathAlterationDetector to PathAlterationObserverScheduler as the current name is confusing. The class does not detect anything.
- Currently, passing an interval <= 0 will throw an exception in start(). Change the logic to not schedule anything instead.
- Trivial simplification in addPathAlterationObserver()


autumnust:
LGTM.

Issue #1598, opened 2017-02-02
Gobblin getting stopped when any container is killed in YARN mode


sidarthakarna:
When any container is killed by YARN, then gobblin is getting a null pointer exception and it is getting stopped. Can be recreated by manually killing any container while running in YARN mode.

Stack trace
```
2017-02-02 17:35:20 UTC ERROR [AMRM Callback Handler Thread] gobblin.yarn.YarnService$AMRMClientCallbackHandler  - Received error: java.lang.NullPointerException
java.lang.NullPointerException
	at gobblin.yarn.YarnService.handleContainerCompletion(YarnService.java:470)
	at gobblin.yarn.YarnService.access$200(YarnService.java:99)
	at gobblin.yarn.YarnService$AMRMClientCallbackHandler.onContainersCompleted(YarnService.java:533)
	at org.apache.hadoop.yarn.client.api.async.impl.AMRMClientAsyncImpl$CallbackHandlerThread.run(AMRMClientAsyncImpl.java:300)
```

Issue is with the following code in YarnService.java handleContainerCompletion method:
```
    int retryCount =
        this.helixInstanceRetryCount.putIfAbsent(completedInstanceName, new AtomicInteger(0)).incrementAndGet();
```

According to the documentation of ConcurrentMap:
```
     * @return the previous value associated with the specified key, or
     *         {@code null} if there was no mapping for the key.
     *         (A {@code null} return can also indicate that the map
     *         previously associated {@code null} with the key,
     *         if the implementation supports null values.)
```

If the key is not present, then it may return null the first time around which causes the null pointer exception

kadaan:
I'll look at fixing this.

shirshanka:
@kadaan:  Thanks for volunteering, is there a PR already for this? Do you need any help?

kadaan:
There isn't a PR yet.  I was testing with our install, but been sidetracked by a prod issue.
--
Joel Baranick
jbaranick@gmail.com
 <mailto:jbaranick@gmail.com>

sidarthakarna:
@kadaan @shirshanka I made the required changes  and sent a PR. Please review.

PR #1864, opened 2017-05-14
Fix for #1598

sidarthakarna:
As explained in the issue, Gobblin is throwing NPE when any container is killed in YARN mode. So, added 0 if the _completedInstanceName_ is not present in collection. Then did the increment.


ibuenros:
@abti can you take a look? It's very short, but you know more about YARN execution I think.

sidarthakarna:
@abti @ibuenros Please take a look and let me know if the code changes are fine. Thanks!

shirshanka:
@sidarthakarna : Thanks. LGTM!

PR #1869, opened 2017-05-15
Rename directory after compaction is complete

yukuai518:
This change is to bridge the gap between old renaming logic for databus-etl, and the new compaction logic in MRCompactionTask. 

In old renaming logic, after the compaction is compete, the directory will be renamed to {DIRNAME}_COMPLETE. Replicate the same logic in the MRCompactionTask now. 

The change will only take effect when compaction.rename.source.dir.enabled=true

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteAction.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteAction.java
@@ -6,5 +6,5 @@ import gobblin.dataset.Dataset;
  * An interface which represents an action that is invoked after a compaction job is finished.
  */
 public interface CompactionCompleteAction<D extends Dataset> {
-  void onCompactionJobComplete(D dataset);
+  void onCompactionJobComplete(D dataset) throws Exception;
 }
--- a/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteAction.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteAction.java
@@ -1,10 +1,12 @@
 package gobblin.compaction.action;
 import gobblin.dataset.Dataset;
 
+import java.io.IOException;
+
 
 /**
  * An interface which represents an action that is invoked after a compaction job is finished.
  */
 public interface CompactionCompleteAction<D extends Dataset> {
-  void onCompactionJobComplete(D dataset) throws Exception;
+  void onCompactionJobComplete(D dataset) throws IOException;
 }
--- a/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
@@ -1,6 +1,9 @@
 package gobblin.compaction.action;
 
+import com.google.common.collect.Lists;
+import gobblin.compaction.dataset.DatasetHelper;
 import gobblin.compaction.mapreduce.CompactionAvroJobConfigurator;
+import gobblin.compaction.mapreduce.MRCompactor;
 import gobblin.compaction.mapreduce.MRCompactorJobRunner;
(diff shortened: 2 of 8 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
```diff
@@ -43,36 +47,60 @@ public CompactionCompleteFileOperationAction (State state, CompactionAvroJobConf
...
-        Counter counter = job.getCounters().findCounter(AvroKeyMapper.EVENT_COUNTER.RECORD_COUNT);
-        long recordCount = counter.getValue();
+      // get record count from map reduce job counter
+      Job job = this.configurator.getConfiguredJob();
+      Counter counter = job.getCounters().findCounter(AvroKeyMapper.EVENT_COUNTER.RECORD_COUNT);
+      long recordCount = counter.getValue();
 
-        // move output from mapreduce to final destination defined by dataset
+
+      boolean renamingRequired = this.state.getPropAsBoolean(MRCompactor.COMPACTION_RENAME_SOURCE_DIR_ENABLED,
```
htran1: The name is a little confusing here. Maybe add a comment that this is append delta mode due to the compaction rename source dir mode being enabled or rename the variable to appendDeltaOutput.

gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
```diff
@@ -43,36 +47,60 @@ public CompactionCompleteFileOperationAction (State state, CompactionAvroJobConf
...
+
+          if (!this.fs.rename(path, outPath)) {
+            throw new IOException(
+                    String.format("Unable to move %s to %s", path.toString(), outPath.toString()));
+          }
+        }
+
+        // update record count (adding previous count)
+        long delta = InputRecordCountHelper.readRecordCount(fs, dstPath);
+        log.info("Will change record count from {} to {} at {}", recordCount, delta, dstPath);
```
htran1: Either change 'to' to 'by' or delta to recordCount + delta.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
```diff
@@ -53,10 +53,20 @@ public void run() {
     }
 
     super.run();
+  }
 
-    List<CompactionCompleteAction> actions = this.suite.getCompactionCompleteActions();
-    for (CompactionCompleteAction action: actions) {
-      action.onCompactionJobComplete(dataset);
+  public void onMRTaskComplete (boolean isSuccess, Throwable throwable) throws RuntimeException {
```
htran1: Don't need to declare throws RuntimeException.

gobblin-compaction/src/main/java/gobblin/compaction/action/CompactionCompleteAction.java
```diff
@@ -6,5 +6,5 @@
  * An interface which represents an action that is invoked after a compaction job is finished.
  */
 public interface CompactionCompleteAction<D extends Dataset> {
-  void onCompactionJobComplete(D dataset);
+  void onCompactionJobComplete(D dataset) throws Exception;
```
htran1: Can this be throws IOException?

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
```diff
@@ -53,10 +53,20 @@ public void run() {
...
+    if (isSuccess) {
+      try {
+        List<CompactionCompleteAction> actions = this.suite.getCompactionCompleteActions();
+        for (CompactionCompleteAction action: actions) {
+          action.onCompactionJobComplete(dataset);
+        }
+      } catch (Exception e) {
```
htran1: catch IOException instead?

PR #1870, opened 2017-05-16
Small fixes to HA for throttling server.

ibuenros:


ibuenros:
Trivial fixes, self-merging.

PR #1878, opened 2017-05-18
Calculating partition date from partition name

aditya1105:


gobblin-modules/gobblin-azkaban/src/main/java/gobblin/data/management/conversion/hive/validation/ValidationJob.java
```diff
@@ -625,11 +632,40 @@ private static FileSystem getSourceFs() throws IOException {
...
+    try {
+      long createTime = getPartitionCreateTime(partition.getName());
+      boolean withinTimeWindow = new DateTime(createTime).isAfter(this.maxLookBackTime) && new DateTime(createTime)
+          .isBefore(this.skipRecentThanTime);
+      if (!withinTimeWindow) {
+        log.info("Skipping partition " + partition.getCompleteName() + " as create time " + new DateTime(createTime)
+            .toString() + " is not within validation time window ");
+      } else {
+        log.info("Validating partition " + partition.getCompleteName());
+        return withinTimeWindow;
```
agepati: It might be better to log all the values and return the boolean directly (without the if, else logic)?

PR #1879, opened 2017-05-18
Fixing minor bug in gobblin-compliance:DatasetUtils

aditya1105:
Int is not sufficient for size of partitions in bytes

### DIFF
--- a/gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/utils/DatasetUtils.java
+++ b/gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/utils/DatasetUtils.java
@@ -44,20 +44,20 @@ public class DatasetUtils {
     return Optional.<HivePartitionDataset>absent();
   }
 
-  public static String getProperty(HivePartitionDataset dataset, String property, int defaultValue) {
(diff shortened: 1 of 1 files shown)


gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/utils/DatasetUtils.java
```diff
@@ -44,20 +44,20 @@
     return Optional.<HivePartitionDataset>absent();
   }
 
-  public static String getProperty(HivePartitionDataset dataset, String property, int defaultValue) {
+  public static String getProperty(HivePartitionDataset dataset, String property, long defaultValue) {
```
ibuenros: What is this used for? Shouldn't the return type of this method be long and the method be `getPropertyAsLong`?

aditya1105: I added this method because it is used at various places in gobblin-compliance to read the value of a property from partition parameters. I made it string because that was the use case.

aditya1105:
Merging

PR #1881, opened 2017-05-18
Default http and r2 rest request builders/response handlers

zxcware:
- Rename `RestliR2Client` to `R2Client`, `buildRequest` to `buildWriteRequest`
- `HttpRequestBuilder` builds a `HttpUriRequest` from a http operation `GenericRecord`
-  `R2RestRequestBuilder` builds a `RestRequest` from a http operation `GenericRecord`

### DIFF
--- a/gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
+++ b/gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
@@ -4,6 +4,7 @@ import java.net.URI;
 import java.util.Map;
 import java.util.Queue;
 
+import org.apache.avro.generic.GenericRecord;
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.methods.RequestBuilder;
@@ -11,6 +12,8 @@ import org.apache.http.entity.ContentType;
 import org.apache.http.entity.StringEntity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
 
 import gobblin.utils.HttpUtils;
 import gobblin.writer.http.AsyncWriteRequest;
@@ -26,7 +29,7 @@ import gobblin.writer.http.BufferedRecord;
  *   a write request from batched records, depending on specific implementation of {@link #buildWriteRequest(Queue)}
  * </p>
  */
-public class HttpRequestBuilder implements AsyncWriteRequestBuilder<HttpOperation, HttpUriRequest> {
+public class HttpRequestBuilder implements AsyncWriteRequestBuilder<GenericRecord, HttpUriRequest> {
   private static final Logger LOG = LoggerFactory.getLogger(HttpRequestBuilder.class);
   private final String urlTemplate;
   private final String verb;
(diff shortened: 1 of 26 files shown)


zxcware:
@abti @ibuenros Please review.

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+      return null;
+    }
+
+    AsyncWriteRequest<GenericRecord, HttpUriRequest> request = new AsyncWriteRequest<>();
+    HttpOperation httpOperation = HttpUtils.toHttpOperation(record.getRecord());
+
+    // Set uri
+    URI uri = HttpUtils.buildURI(urlTemplate, httpOperation.getKeys(), httpOperation.getQueryParams());
+    if (uri == null) {
+      return null;
```
ibuenros: When does this happen?

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+    if (headers != null && headers.size() != 0) {
+      for (Map.Entry<String, String> header : headers.entrySet()) {
+        builder.setHeader(header.getKey(), header.getValue());
+      }
+    }
+
+    // Add payload
+    int bytesWritten = addPayload(builder, httpOperation.getBody());
+    if (bytesWritten == -1) {
+      return null;
```
ibuenros: Should you throw an error here?

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Add payload to request. By default, payload is sent as application/json
+   */
+  protected int addPayload(RequestBuilder builder, String payload) {
+    if (payload == null || payload.length() == 0) {
+      return 0;
+    }
+
+    builder.setHeader(HttpHeaders.CONTENT_TYPE, ContentType.APPLICATION_JSON.getMimeType());
```
ibuenros: The schema says this can be a raw string, should the user specify the mime-type?

gobblin-modules/gobblin-http/src/main/java/gobblin/utils/HttpUtils.java
```diff
@@ -0,0 +1,97 @@
...
+   * Convert the given {@link GenericRecord} to {@link HttpOperation}
+   */
+  public static HttpOperation toHttpOperation(GenericRecord record) {
+    if (record instanceof HttpOperation) {
+      return (HttpOperation) record;
+    }
+
+    HttpOperation httpOperation = new HttpOperation();
+    boolean hasAnyField = false;
+    for (Schema.Field field: HttpOperation.getClassSchema().getFields()) {
```
ibuenros: Note this will not work for nested schemas. That's ok, but it might be important to note that in the schema so future changes don't break the logic.

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
zxcware: When the url syntax is invalid. Looks like we should throw an error.

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
zxcware: Good point.

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpRequestBuilder.java
zxcware: Yes, the user should be able to specify mime-type. I'll make the change.

gobblin-modules/gobblin-http/src/main/java/gobblin/http/HttpResponseHandler.java
```diff
@@ -13,12 +19,21 @@
...
   @Override
   public ResponseStatus handleResponse(CloseableHttpResponse response) {
     ResponseStatus status = new ResponseStatus(StatusType.OK);
-
     int statusCode = response.getStatusLine().getStatusCode();
+    HttpEntity entity = response.getEntity();
+    if (entity != null) {
+      try {
+        String entityStr = EntityUtils.toString(response.getEntity());
+        // TODO process entityStr
```
ibuenros: add an empty method `processEntity(String)`

PR #1882, opened 2017-05-18
Added a CLI for ConfigClient.

ibuenros:


ibuenros:
@autumnust can you take a look?

gobblin-config-management/gobblin-config-client/src/main/java/gobblin/config/client/ConfigClientCLI.java
```diff
@@ -0,0 +1,70 @@
...
+    if (command.resolvedConfig) {
+      Config resolvedConfig = configClient.getConfig(command.uri);
+      System.out.println(resolvedConfig.root().render(ConfigRenderOptions.defaults()));
+    }
+  }
+
+  /**
+   * The parsed user command.
+   */
+  public static class Command {
```
autumnust: Are there any reason to make this class static?

ibuenros: In general it is recommended to make classes static if possible. It helps removing unnecessary hard references that may prevent garbage collection.

PR #1884, opened 2017-05-19
Dropping staging table in ComplianceRetentionJob

aditya1105:


gobblin-modules/gobblin-compliance/src/main/java/gobblin/compliance/retention/ComplianceRetentionJob.java
```diff
@@ -59,28 +66,52 @@
...
     String finderClass = properties.getProperty(GOBBLIN_COMPLIANCE_DATASET_FINDER_CLASS);
     this.finder = GobblinConstructorUtils.invokeConstructor(DatasetsFinder.class, finderClass, new State(properties));
+
+    Iterator<HiveDataset> datasetsIterator =
+        new HiveDatasetFinder(FileSystem.newInstance(new Configuration()), properties).getDatasetsIterator();
+
+    while (datasetsIterator.hasNext()) {
+      // Drop partitions from empty tables if property is set, otherwise skip the table
+      HiveDataset hiveDataset = datasetsIterator.next();
+      List<Partition> partitionsFromDataset = hiveDataset.getPartitionsFromDataset();
```
agepati: The general convention is not to delete during iteration. Could you please keep a list of the tables you'd like to drop and do so separately after the init method, perhaps?

aditya1105: Done. Merging

PR #1886, opened 2017-05-21
Add conversion of the DecodeableKafkaRecord to the extract schema (la…

htran1:
…test schema) since the DecodeableKafkaReocord may have a value with an older schema.


htran1:
@yukuai518, please review.

PR #1888, opened 2017-05-21
adding more information in events for gobblintrackingevent_distcp_ng

arjun4084346:
adding more information, like source cluster, destination cluster, azkaban URL for the job in events for gobblintrackingevent_distcp_ng

@ibuenros Could you please review this.

### DIFF
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -118,7 +118,6 @@ public class ConfigurationKeys {
   public static final String FLOW_GROUP_KEY = "flow.group";
   public static final String FLOW_DESCRIPTION_KEY = "flow.description";
   public static final String FLOW_EXECUTION_ID_KEY = "flow.executionId";
-  public static final String AZKABAN_EXECUTION_URL = "azkaban.link.execution.url";
 
   /**
    * Common topology configuration properties.
--- a/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
@@ -118,6 +118,7 @@ public class ConfigurationKeys {
   public static final String FLOW_GROUP_KEY = "flow.group";
   public static final String FLOW_DESCRIPTION_KEY = "flow.description";
   public static final String FLOW_EXECUTION_ID_KEY = "flow.executionId";
+  public static final String AZKABAN_EXECUTION_URL = "azkaban.link.execution.url";
 
   /**
    * Common topology configuration properties.
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
@@ -17,6 +17,8 @@
 
 package gobblin.data.management.copy.publisher;
 
+
+import gobblin.metrics.event.sla.SlaEventKeys;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
@@ -192,9 +194,6 @@ public class CopyDataPublisher extends DataPublisher implements UnpublishedHandl
     long datasetOriginTimestamp = Long.MAX_VALUE;
     long datasetUpstreamTimestamp = Long.MAX_VALUE;
 
-    final FileSystem sourceFs = getSourceFileSystem(this.state);
-    final FileSystem targetFs = getTargetFileSystem(this.state);
-
     for (WorkUnitState wus : datasetWorkUnitStates) {
       if (wus.getWorkingState() == WorkingState.SUCCESSFUL) {
         wus.setWorkingState(WorkUnitState.WorkingState.COMMITTED);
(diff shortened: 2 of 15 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -190,6 +192,9 @@ private void publishFileSet(CopyEntity.DatasetAndPartition datasetAndPartition,
     long datasetOriginTimestamp = Long.MAX_VALUE;
     long datasetUpstreamTimestamp = Long.MAX_VALUE;
 
+    final FileSystem sourceFs = getSourceFileSystem(this.state);
```
ibuenros: Unfortunately this might not be the source filesystem. DIstcp supports copying data from a third filesystem, so this needs to be read from the path of the origin files. You can look at `gobblin.data.management.copy.writer.FileAwareInputStreamDataWriter` for an example.

Consider having the writer populate this information into the state instead, as the writer is guaranteed to know the source and target filesystems.

gobblin-hive-registration/src/main/java/gobblin/hive/HiveRegister.java
```diff
@@ -212,7 +212,7 @@ public abstract boolean existsPartition(String dbName, String tableName, List<Co
    * @param tableName the table name
    * @throws IOException
    */
-  public abstract void dropTableIfExists(String dbName, String tableName) throws IOException;
+  public abstract void dropTableIfExists(String dbName, String tableName, String azkabanUrl, String metastoreURI) throws IOException;
```
ibuenros: You shouldn't modify public APIs. Furthermore, these parameters shouldn't be part of this method. The HiveRegister implementation must already know the metastoreURI, while the azkabanUri should be injected into the EventSubmitted.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/gobblin/metrics/event/sla/SlaEventSubmitter.java
```diff
@@ -57,6 +57,9 @@
   private String dedupeStatus;
   private String completenessPercentage;
   private String isFirstPublish;
+  private String sourceCluster;
```
ibuenros: I don't think this should be top-level fields of `SlaEventSubmitter`, as they don't make sense for all datasets. Instead put them in additionalMetadata.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -118,6 +118,7 @@
   public static final String FLOW_GROUP_KEY = "flow.group";
   public static final String FLOW_DESCRIPTION_KEY = "flow.description";
   public static final String FLOW_EXECUTION_ID_KEY = "flow.executionId";
+  public static final String AZKABAN_EXECUTION_URL = "azkaban.link.execution.url";
```
ibuenros: Is this a configuration key? Users are never expected to set this by hand.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyEventSubmitterHelper.java
```diff
@@ -41,10 +41,12 @@
   public static final String SIZE_IN_BYTES = "SizeInBytes";
 
   static void submitSuccessfulDatasetPublish(EventSubmitter eventSubmitter,
-      CopyEntity.DatasetAndPartition datasetAndPartition, String originTimestamp, String upstreamTimestamp) {
-    SlaEventSubmitter.builder().eventSubmitter(eventSubmitter).eventName(DATASET_PUBLISHED_EVENT_NAME)
+      CopyEntity.DatasetAndPartition datasetAndPartition, String originTimestamp, String upstreamTimestamp,
+      String source, String destination, String azkabanUrl) {
```
ibuenros: Be a bit more specific with the names: `source` -> `sourceUri`, `destination` -> `destinationUri`. The `azkabanUrl` is only valid for flows running on Azkaban, which is a minority, so it shouldn't be an argument of this method. Instead, you could put a map of additional metadata.

gobblin-data-management/src/main/java/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -230,6 +231,8 @@ public boolean apply(FileStatus input) {
         } else {
           log.info("File {} copied.", copyableFile.getOrigin().getPath());
         }
+        this.state.setProp(SlaEventKeys.SOURCE, copyableFile.getOrigin().getPath().toString());
```
ibuenros: Shouldn't the source be the source cluster as opposed to the fully qualified path?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -258,7 +260,9 @@ public void dropTableIfExists(String dbName, String tableName) throws IOExceptio
     try (AutoReturnableObject<IMetaStoreClient> client = this.clientPool.getClient()) {
       if (client.get().tableExists(dbName, tableName)) {
         client.get().dropTable(dbName, tableName);
-        HiveMetaStoreEventHelper.submitSuccessfulTableDrop(eventSubmitter, dbName, tableName);
+        String metastoreURI = this.clientPool.getHiveConf().get(HiveMetaStoreClientFactory.HIVE_METASTORE_TOKEN_SIGNATURE, "null");
+        String azkabanUrl = this.props.getProp(ConfigurationKeys.AZKABAN_EXECUTION_URL, "null");
```
ibuenros: Please don't include azkaban specific stuff here. Instead, add it at the `RootMetricContext` level, in that case it will be an automatic tag everywhere.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -218,9 +224,11 @@ private void publishFileSet(CopyEntity.DatasetAndPartition datasetAndPartition,
       datasetUpstreamTimestamp = 0;
     }
 
+    additionalMetadata.put(SlaEventKeys.SOURCE_URI, this.state.getProp(SlaEventKeys.SOURCE_URI));
+    additionalMetadata.put(SlaEventKeys.DESTINATION_URI, this.state.getProp(SlaEventKeys.DESTINATION_URI));
     CopyEventSubmitterHelper.submitSuccessfulDatasetPublish(this.eventSubmitter, datasetAndPartition,
-        Long.toString(datasetOriginTimestamp), Long.toString(datasetUpstreamTimestamp));
-  }
+        Long.toString(datasetOriginTimestamp), Long.toString(datasetUpstreamTimestamp), additionalMetadata);
```
ibuenros: I believe we may also want the Hive metastore information here in the case of Hive copy.

arjun4084346: As per the WIMD team request, I added Hive metastore information in PartitionDropSucceed and TableDropSucceed events only.
Isn't hive metastore information deduced from source and target URIs? As per my understanding, it is a one-to-one mapping.

PR #1893, opened 2017-05-23
Adding configStoreSupport for gobblin-kafka and gobblin-compaction

aditya1105:


gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/ConfigStoreUtils.java
```diff
@@ -0,0 +1,178 @@
...
+    try {
+      Collection<URI> importedBy = configClient.getImportedBy(new URI(tagUri.toString()), true, runtimeConfig);
+      return importedBy.stream().filter((URI u) -> u.toString().contains(filterString)).collect(Collectors.toList());
+    } catch (Exception e) {
+      log.info(e.toString());
+    }
+    return new ArrayList();
+  }
+
+  public static Optional<String> getConfigStoreUri(Properties properties) {
```
abti: this seems generic enough to not be in this package / module

aditya1105: Yes will move it with next pr

PR #1902, opened 2017-05-26
update case study to latest version of code

cvaliente:
DatasetUtils.getDatasetSpecificProps is now explicitely looking for the "dataset" object in the JSON so the case study no longer works.


ibuenros:
Thanks @cvaliente. Merging.

cvaliente:
thanks a lot!

Issue #1901, opened 2017-05-26
dataset specific properties are ignored&dropped by KafkaBiLevelWorkUnitPacker


cvaliente:
I failed to get dataset.specific.props to work on our jobs, and I think I found the reason:

in KafkaSource.getWorkUnitForTopicPartition the properties are added correctly to the individual workunits.

The KafkaBiLevelWorkUnitPacker then assigns the WorkUnits to their bins and combines them into one WorkUnit in squeezeMultiWorkUnit() but doesn't copy over the topicSpecificSettings.

Using the KafkaSingleLevelWorkUnitPacker works fine with dataset.specific.props since it doesn't call squeezeMultiWorkUnit on non-empty workUnits.

cvaliente:
fix in #1903

sahilTakiar:
Doesn't `KafkaSource#addTopicSpecificPropsToWorkUnits` handle adding dataset specific configuration? That method is run after the bin-packing is done. So if `dataset.specific.props` isn't working I would guess the bug would be in that method.

cvaliente:
You are right, that wasn't yet implemented in 0.9 and I forgot to check upstream.

PR #1903, opened 2017-05-26
issue 1901 add any properties from the collected workunits to the compacted Workunit

cvaliente:
This used to drop any topic specific settings when using the BI_LEVEL workunitpacker.

also made it more explicit in the documentation about what workunits the squeezeMultiWorkUnit method is supposed to work with.

See #1901


ibuenros:
@cvaliente is this PR still valid? I saw @sahilTakiar commented on the original issue.

cvaliente:
@ibuenros no, this is not needed any more.

PR #1908, opened 2017-05-31
Enabling config store for validation

aditya1105:
Adding config store whitelisting for file format validation


aditya1105:
Self merging

ibuenros:
@aditya1105 why was this self-merged?

PR #1909, opened 2017-06-01
Task push mode using reactivex.

ibuenros:
This PR changes the way Tasks are executed to a push model using Reactivex (http://reactivex.io/).

Main changes:
* Task no longer handles records, instead it handles a record stream containing a Reactivex `Flowable` along with metadata.
* Constructs modified to handle the stream instead of individual records:
  * Extractor: Main point of interaction is `recordStream` method. By default, create a stream that repeatedly calls `readRecord`.
  * Converter / Quality checker: Main point of interaction is `processStream` method. By default, map records using `convertRecord` and `executePolicies`.
  * Writer: minimal changes, mostly around `RecordEnvelope` explained later.
* Forks work a little different. A `Forker` takes a `ForkOperator` and splits the record stream into multiple record streams. Then, we call `Fork.consumeRecordStream` for each such mirrored stream.
* All records are wrapped in `RecordEnvelope` (introduced previously). This is a first step to more closely unify batch and streaming. There's nothing special about `RecordEnvelope`, everything works transparently, except metrics get a little tricky.


ibuenros:
@ydai1124 can you review? @shirshanka you might be interested.

gobblin-api/src/main/java/gobblin/fork/Forker.java
```diff
@@ -0,0 +1,145 @@
...
+    List<Boolean> forkedSchemas = forkOperator.forkSchema(workUnitState, inputStream.getSchema());
+
+    Preconditions.checkState(forkedSchemas.size() == branches, String
+        .format("Number of forked schemas [%d] is not equal to number of branches [%d]", forkedSchemas.size(),
+            branches));
+
+    Flowable<RecordWithForkMap<D>> forkedStream = inputStream.getRecordStream().map(r ->
+        new RecordWithForkMap<>(r, forkOperator.forkDataRecord(workUnitState, r.getRecord())));
+
+    ConnectableFlowable<RecordWithForkMap<D>> connectableFlowable = forkedStream.publish(100);
```
ydai1124: Can you make the buffer size configurable instead of hard coded `100`? A single record size can be very large.

gobblin-api/src/main/java/gobblin/records/RecordStreamConsumer.java
```diff
@@ -0,0 +1,32 @@
...
+ * An object that a {@link RecordStreamWithMetadata}.
+ * @param <S> schema type
+ * @param <D> data type
+ */
+public interface RecordStreamConsumer<S, D> {
+
+  /**
+   * Subscribe to the input {@link RecordStreamWithMetadata}.
+   */
+  void consumeRecordStream(RecordStreamWithMetadata<D, S> stream) throws Exception;
```
ydai1124: Can you create a more specified exception class?

gobblin-api/src/main/java/gobblin/records/RecordStreamProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+ * @param <SO> output schema type
+ * @param <DI> input data type
+ * @param <DO> output data type
+ */
+public interface RecordStreamProcessor<SI, SO, DI, DO> {
+
+  /**
+   * Return a {@link RecordStreamWithMetadata} with the appropriate modifications.
+   */
+  RecordStreamWithMetadata<DO, SO> processStream(RecordStreamWithMetadata<DI, SI> inputStream, WorkUnitState state)
```
ydai1124: ditto.

gobblin-api/src/main/java/gobblin/records/RecordStreamWithMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.Data;
+
+
+/**
+ * A stream of records along with metadata (e.g. Schema).
+ * @param <D> type of record.
+ * @param <S> type of schema.
+ */
+@Data
+public class RecordStreamWithMetadata<D, S> {
```
ydai1124: Are there other metadata besides `schema`? Maybe `RecordStreamWithSchema` is more descriptive?

gobblin-runtime/src/main/java/gobblin/runtime/fork/Fork.java
```diff
@@ -440,17 +356,26 @@ protected void cleanup() {
...
+      }
 
-    DataWriter<Object> writer = new PartitionedDataWriter<>(builder, this.taskContext.getTaskState());
-    logger.info("Wrapping writer " + writer);
-    return new DataWriterWrapperBuilder<>(writer, this.taskState).build();
+      DataWriter<Object> writer = new PartitionedDataWriter<>(builder, this.taskContext.getTaskState());
+      logger.info("Wrapping writer " + writer);
+      return new DataWriterWrapperBuilder<>(writer, this.taskState).build();
+    } catch (Throwable t) {
+      logger.error("blah", t);
```
ydai1124: Can you replace `blah` with some more descriptive blah blah blah? :)

gobblin-runtime/src/main/java/gobblin/runtime/fork/Fork.java
```diff
@@ -210,51 +222,6 @@ public State getFinalState() {
...
-   * it should not expect more incoming data records.
-   *
-   * <p>
-   *   This method is used by the {@link Task} that creates this {@link Fork}.
-   * </p>
-   */
-  public void markParentTaskDone() {
-    this.parentTaskDone = true;
-    try {
-      this.putRecord(SHUTDOWN_RECORD);
```
ydai1124: The definition of constant `SHUTDOWN_RECORD` can be removed?

ydai1124:
There are also build failures.

gobblin-api/src/main/java/gobblin/converter/Converter.java
```diff
@@ -100,4 +105,20 @@ public abstract SO convertSchema(SI inputSchema, WorkUnitState workUnit)
...
+  public RecordStreamWithMetadata<DO, SO> processStream(RecordStreamWithMetadata<DI, SI> inputStream,
+      WorkUnitState workUnitState) throws SchemaConversionException {
+    init(workUnitState);
+    SO outputSchema = convertSchema(inputStream.getSchema(), workUnitState);
+    Flowable<RecordEnvelope<DO>> outputStream =
+        inputStream.getRecordStream()
+            .flatMap(in -> Flowable.fromIterable(convertRecord(outputSchema, in.getRecord(), workUnitState))
+            .map(in::withRecord));
+    outputStream = outputStream.doOnComplete(this::close);
+    return inputStream.withRecordStream(outputStream, outputSchema);
```
ibuenros: No, because it preserver all of the metadata in the stream (if present).

gobblin-api/src/main/java/gobblin/records/RecordStreamWithMetadata.java
ibuenros: Not for now, but later we may want to add more stuff to it.

shirshanka:
@ibuenros : High level question is how are you planning to mitigate risk with this change to the core pipeline?

ibuenros:
@shirshanka the plan was to do it through testing, but I understand even that is risky. An alternative is to leave the old pipeline and put a configuration switch, but I don't know how ugly that may get

shirshanka:
@ibuenros : I'd like to think through how we might do that... maybe a new Task impl ?

gobblin-api/src/main/java/gobblin/fork/CopyHelper.java
```diff
@@ -56,37 +56,26 @@ private static boolean isImmutableType(Object thing) {
...
   /**
    * Copy this object if needed.
-   * @param thing: this object that needs to be copied
-   * @param instance: the number of instances of this object already copied
+   * @param thing : this object that needs to be copied
    * @return: a possibly copied instance
    * @throws CopyNotSupportedException if thing needs to be copied but cannot be
    */
-  public static Object copy(Object thing, int instance) throws CopyNotSupportedException {
-    // First copy is special, we only copy if the thing is an instance of Copyable
```
shirshanka: Why did you move the check out from here to the caller?

gobblin-api/src/main/java/gobblin/source/extractor/Extractor.java
```diff
@@ -74,4 +83,38 @@
...
+  default RecordEnvelope<D> readRecordEnvelope() throws DataRecordException, IOException {
+    D record = readRecord(null);
+    return record == null ? null : new RecordEnvelope<>(record);
+  }
+
+  /**
+   * @param shutdownRequest an {@link AtomicBoolean} that becomes true when a shutdown has been requested.
+   * @return a {@link Flowable} with the records from this source. Note the flowable should honor downstream backpressure.
+   */
+  default RecordStreamWithMetadata<D, S> recordStream(AtomicBoolean shutdownRequest) throws IOException {
```
shirshanka: What is the expected behavior when shutdownRequest is set to true? 

pros and cons of doing it this way versus having an explicit shutdown method?

gobblin-api/src/main/java/gobblin/records/RecordStreamWithMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+  private final Flowable<RecordEnvelope<D>> recordStream;
+  private final S schema;
```
eogren: Hmm - if we are consuming from a Kafka topic in streaming mode and the schema is evolved, would we be able to pick up the change with this approach? Should the Schema actually be part of the envelope?

gobblin-api/src/main/java/gobblin/fork/CopyHelper.java
ibuenros: Because the caller doesn't need to keep track of the instance number, just whether the object needs to be copied or not. For mutable, trivially copyable objects, like byte array, it is incorrect to return `thing` here, as the first fork may mutate the object before it gets copied into other forks (race condition). 

The only instance that is safe to copy is the last one, but then we have an implicit contract that the instances are passed in order to the Copier, which is not specified anywhere and is hard to enforce.

gobblin-api/src/main/java/gobblin/records/RecordStreamWithMetadata.java
ibuenros: This is mostly replicating the current flow where the schema is only passed along once. We could move the schema to be part of the envelope, but then every record would require looking at the schema. At any rate, this design does not prevent us from putting a dummy schema here and the real schema in the envelope.

What I could see as an issue is that on detection of new schema we may want to call pass the schema through the entire flow again. In that case, even in current Gobblin, I believe we would need to re-instantiate the pipeline, as some converters have side effects on `convertSchema`.

gobblin-api/src/main/java/gobblin/source/extractor/Extractor.java
ibuenros: I though about that, but having a `shutdown` method is hard because interfaces cannot hold instance variables, so the shutdown has no way of notifying the stream to stop.

gobblin-api/src/main/java/gobblin/fork/CopyHelper.java
ibuenros: An alternative is to have the method `Iterable<Object> CopyHelper.createCopies(Object obj, int numCopies)`, so the method can guarantee the copies happen before the object is mutated.

gobblin-api/src/main/java/gobblin/fork/Forker.java
ibuenros: Fixed

gobblin-api/src/main/java/gobblin/fork/Forker.java
```diff
@@ -0,0 +1,145 @@
...
+
+    List<RecordStreamWithMetadata<D, S>> forkStreams = Lists.newArrayList();
+
+    for(int i = 0; i < forkedSchemas.size(); i++) {
+      if (forkedSchemas.get(i)) {
+        final int idx = i;
+        Flowable<RecordEnvelope<D>> thisStream =
+            connectableFlowable.filter(new ForkFilter<>(idx)).map(RecordWithForkMap::getRecordCopyIfNecessary);
+        forkStreams.add(inputStream.withRecordStream(thisStream,
+            mustCopy(forkedSchemas) ? (S) CopyHelper.copy(inputStream.getSchema()) : inputStream.getSchema()));
```
ibuenros: Fixed

gobblin-api/src/main/java/gobblin/records/RecordStreamConsumer.java
ibuenros: Fixed

gobblin-api/src/main/java/gobblin/records/RecordStreamProcessor.java
ibuenros: Fixed

gobblin-runtime/src/main/java/gobblin/runtime/fork/Fork.java
ibuenros: Restored old fork execution, so this will stay for now.

gobblin-runtime/src/main/java/gobblin/runtime/fork/Fork.java
ibuenros: Fixed. Thanks for noticing :)

ibuenros:
@shirshanka restored old task execution (albeit refactored into a separate method), which is the default execution method. However, it does still use `RecordEnvelopes` as that is necessary to start unifying stream and batch.

PR #1914, opened 2017-06-02
Added a performance test to compute Gobblin record throughput.

ibuenros:


ibuenros:
Trivial changes. Just added classes to test performance.

PR #1915, opened 2017-06-02
Add a missing line that resulted in files from the old location being…

htran1:
… deleted when a hive table is replaced.


htran1:
@autumnust, please review.

PR #1925, opened 2017-06-06
Add an option to promote an MySQL unsigned int to a bigint to handle …

htran1:
…large unsigned ints.


htran1:
@zxcware, please review.

PR #1928, opened 2017-06-06
Upgrade pegasus dependency

zxcware:


gobblin-rest-service/gobblin-rest-server/build.gradle
```diff
@@ -59,6 +59,7 @@ buildscript {
 
     dependencies {
         classpath group: 'com.linkedin.pegasus', name: 'gradle-plugins', version: pegasusVersion
+        classpath externalDependency.jdkTools
```
ibuenros: Why is this needed?

zxcware: pegasus version 11.0.0 has a bug. It fails to add tools.jar into the classpath. They're working on a clean fix. This is a temporary fix for all the open sources that rely on restli version 7+. I moved the fix to the root `build.script`. Please see the latest diff.

PR #1930, opened 2017-06-06
Check for DecodeableKafkaRecord with null value and trigger undecodab…

htran1:
…le record error handling. A record with a null value is returned when the record cannot be decoded.


htran1:
@ibuenros, @yukuai518, please review.

PR #1931, opened 2017-06-07
re-org http components into gobblin-http module

zxcware:
- Move newly introduced http components from `gobblin-core` to `gobblin-http`
- Decouple AsyncWriteRequestBuilder from write, making one general for both async read and write


zxcware:
@ibuenros @yukuai518 Please review. This will unblock kuai from working on the ambry join converter

gobblin-core-base/src/main/java/gobblin/writer/WriteCallback.java
```diff
@@ -17,15 +17,15 @@
 
 package gobblin.writer;
 
-public interface WriteCallback<T> {
+import gobblin.async.Callback;
 
-  public void onSuccess(WriteResponse<T> writeResponse);
 
-  public void onFailure(Throwable throwable);
+public interface WriteCallback extends Callback<WriteResponse> {
```
ibuenros: Why is this no longer templated?

zxcware: Fixed

PR #1934, opened 2017-06-08
- Add a little bit of hierarchical config for converter encryption - previously converters

eogren:
only looked converter.encrypt.*, but this breaks down if there are
multiple converters in a chain that are doing encryption. This change
allows you to specify config at `converter.<ConverterName>.encrypt.*`

I think this approach is a little cleaner than a positional approach (converter.0.encrypt.foo), but don't know if you have thoughts @htran1 or @ibuenros . Also definitely want to document this somewhere and/or make the config parsing a little more generic. I know Typesafe Config can handle some of the heirarchy stuff as well.


eogren:
going to merge this to unblock work but comments still appreciated.

PR #1936, opened 2017-06-09
Fixing bug for cleaning err dir

aditya1105:


aditya1105:
Very trivial, merging

PR #1937, opened 2017-06-09
Propogating IllegalArgumentException instead of silent failure in HiveDatasetFinder

aditya1105:


ibuenros:
@aditya1105 will this cause immediate job failure? If so, we don't want that behavior.

aditya1105:
@ibuenros gobblin job should fail whenever such a scenario occurs. Illegal argument exception should not be catched and logged.

ibuenros:
I guess I don't exactly know when `IllegalArgumentException` is thrown. Any examples?

aditya1105:
Mostly by com.google.common.base.Preconditions. In the current scenario, Preconditions are throwing IllegalArgumentException. Also IllegalArgumentException extends RuntimeException, so we should let it fail the job.

abti:
Issue: https://issues.apache.org/jira/browse/GOBBLIN-13

Please update your PR title with following prefix: [GOBBLIN-13]

PR #1941, opened 2017-06-12
Added a text file based source and max files to pull to all file base…

ibuenros:
…d sources.


gobblin-core/src/main/java/gobblin/source/extractor/filebased/FileBasedSource.java
```diff
@@ -191,23 +209,31 @@ protected void initLogger(SourceState state) {
...
-    String path = state.getProp(ConfigurationKeys.SOURCE_FILEBASED_DATA_DIRECTORY) + "/*"
-        + state.getProp(ConfigurationKeys.SOURCE_ENTITY) + "*";
+    String path = getLsPattern(state);
 
     try {
       log.info("Running ls command with input " + path);
       results = this.fsHelper.ls(path);
       for (int i = 0; i < results.size(); i++) {
-        String filePath = state.getProp(ConfigurationKeys.SOURCE_FILEBASED_DATA_DIRECTORY) + "/" + results.get(i);
+        File parsedFile = new File(new URI(results.get(i)));
```
zxcware: Would this throw an error if the uri points to a relative path but it can't refer to a file without `ConfigurationKeys.SOURCE_FILEBASED_DATA_DIRECTORY`? Should we use `URI#isAbsolute` instead?

ibuenros: I don't quite understand the your first question.

For your second suggestion, `URI#isAbsolute` checks whether the URI has a scheme, so it would miss "absolute" paths like `/my/path`.

zxcware: For example, if we end up with `File parsedFile = new File("data1.txt")`, would this throw an error as the program can't find "data1.txt" in its current directory?

ibuenros: `File` is equivalent to `Path` in Hadoop world. It does not require that there is a file with that name.

ydai1124:
@ibuenros test `TextFileBasedSourceTest` failed. Can you take a look?

PR #1942, opened 2017-06-12
Add a simple converter that can convert a RecordWithMetadata into its underlying record

eogren:
- Move Metadata types into gobblin-api (this was an older PR comment
from sdas that I'm finally getting around to).
- Add a simple converter that can convert a RecordWithMetadata into its
underlying record


ibuenros:
@eogren this should no longer be needed? all records are wrapped now.

eogren:
closing

PR #1948, opened 2017-06-14
Switch to dataset statestore for KafkaSource

ydai1124:
Add `datasetUrn` when creating workunits.
Use `isDatasetStateEnabled` to ensure backwards compatible when reading `FsStateStore` that is not `datasetAware`.
Tested on holdem.

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -19,8 +19,10 @@ package gobblin.source.extractor.extract.kafka;
 
 import gobblin.source.extractor.limiter.LimiterConfigurationKeys;
 import gobblin.source.workunit.MultiWorkUnit;
+
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -148,18 +159,15 @@ private void setLimiterReportKeyListToWorkUnits(List<WorkUnit> workUnits, List<S
...
-          .resolveClass(
-              state.getProp(GOBBLIN_KAFKA_CONSUMER_CLIENT_FACTORY_CLASS,
-                  DEFAULT_GOBBLIN_KAFKA_CONSUMER_CLIENT_FACTORY_CLASS)).newInstance()
-          .create(ConfigUtils.propertiesToConfig(state.getProperties()));
+      this.kafkaConsumerClient = kafkaConsumerClientResolver.resolveClass(state
+          .getProp(GOBBLIN_KAFKA_CONSUMER_CLIENT_FACTORY_CLASS, DEFAULT_GOBBLIN_KAFKA_CONSUMER_CLIENT_FACTORY_CLASS))
+          .newInstance().create(ConfigUtils.propertiesToConfig(state.getProperties()));
 
       List<KafkaTopic> topics = getFilteredTopics(state);
+      this.topicsToProcess = new HashSet<>(topics.stream().map(KafkaTopic::getName).collect(toList()));
```
ibuenros: you can do `topics.stream().map(KafkaTopic::getName).collect(Collectors.toSet)` instead

PR #1949, opened 2017-06-14
Add createHttpClient implementation to AvroR2JoinConverter.java

yukuai518:
@zxcware Can you review?

### DIFF
--- a/gobblin-modules/gobblin-http/src/main/java/gobblin/converter/AvroR2JoinConverter.java
+++ b/gobblin-modules/gobblin-http/src/main/java/gobblin/converter/AvroR2JoinConverter.java
@@ -13,8 +13,8 @@ import com.typesafe.config.Config;
 import lombok.extern.slf4j.Slf4j;
 
 import gobblin.async.AsyncRequestBuilder;
-import gobblin.config.ConfigBuilder;
-import gobblin.configuration.WorkUnitState;
+import gobblin.broker.gobblin_scopes.GobblinScopeTypes;
+import gobblin.broker.iface.SharedResourcesBroker;
 import gobblin.http.HttpClient;
 import gobblin.http.HttpRequestResponseRecord;
 import gobblin.http.ResponseHandler;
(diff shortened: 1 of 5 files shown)


gobblin-modules/gobblin-http/src/main/java/gobblin/converter/AvroR2JoinConverter.java
```diff
@@ -42,8 +46,19 @@ protected void fillHttpOutputData(Schema schema, GenericRecord outputRecord, Res
 
   @Override
   protected HttpClient<RestRequest, RestResponse> createHttpClient(WorkUnitState workUnitState) {
-    //TODO: should use R2ClientFactory to initilize R2Client
-    return null;
+    Config config = ConfigBuilder.create().loadProps(workUnitState.getProperties(), CONF_PREFIX).build();
+    config = config.withFallback(DEFAULT_FALLBACK);
```
zxcware: We could DRY line 49-50:
1. These `create` methods accept `Config` instead of `WorkUnitState`
2. Apply the `DEFAULT_FALLBACK` in the `init` function
3. On create the HttpClient, pass the broker as the second argument

Since config is namespace driven. This approach helps restrict a subclass to create those constructs based on only the configurations inside that namespace, avoid scoping issues

gobblin-modules/gobblin-http/src/main/java/gobblin/converter/HttpJoinConverter.java
```diff
@@ -48,18 +50,26 @@
   protected ResponseHandler<RP> responseHandler = null;
   protected AsyncRequestBuilder<GenericRecord, RQ> requestBuilder = null;
 
+  public HttpJoinConverter init(WorkUnitState workUnitState) {
+    super.init(workUnitState);
+    Config config = ConfigBuilder.create().loadProps(workUnitState.getProperties(), CONF_PREFIX).build();
+    config.withFallback(DEFAULT_FALLBACK);
```
zxcware: Should it be `config = config.withFallback(DEFAULT_FALLBACK)`?

PR #1927, opened 2017-06-06
Implementation of tag-based retention

autumnust:
- Fix race condition for trash folder accessing
- Add new conf. DATASET_PATH to maintain the generality of ConfigBasedCleanabledDatasetFinder

PR #1876, opened 2017-05-17
Minor fix for /data/management/copy/replication

autumnust:
Also Add dataset existance checking for `HadoopFsEndPoint.java` 

This PR is part of migration to use Config-store-based DistcpNG job.

PR #1951, opened 2017-06-15
Implemented job-level black/whitelist for DistcpNG

autumnust:
- Add job-level black/whitelist in copySource.java
- Enbale logging of version number of configStore
- fix minor bugs in /data/management/copy/replication


shirshanka:
Can you fix the title for the PR to be more descriptive of what the changes are? The current title is only understandable inside LinkedIn.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDataset.java
```diff
@@ -89,9 +89,7 @@ private void calculateDatasetURN(){
       } catch (IOException e1) {
         // ignored
       }
-    }
-
-    this.datasetURN = e.toString();
+    } else this.datasetURN = e.toString();
```
ibuenros: Please wrap all conditional branches with brackets.

autumnust: @ibuenros  
Regarding to the second comments above: 
I agree. The main reason for implementing black/whitelist in this level: 
- There's no base Implementation of `DatasetsFinder`, each kinds of dataset finder is implementing the interface, Which means it is hard to extending black/whitelist into every dataset finder.  For this particular example, I could choose to implement this function in `configBasedDatasetsfinder`.  
- The current implementation is equivalent to put a filter when creating `CopyableDatasetRequestor`, which is on dataset-level but not specific to a type of dataset.

autumnust: Addressed the comments on configuration name and moved the implementation to datasetFinder level. 
@ibuenros Could we first merge #1927 and #1876 since there's some interleaving with this PR? And I will update the PR afterwards.

PR #1955, opened 2017-06-16
changing avro.schema.url for hive partitions before distcp

arjun4084346:
changing avro.schema.url in source partitions to reflect new destination path

@ibuenros Please review.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -17,6 +17,7 @@
 
 package gobblin.data.management.copy.hive;
 
+import gobblin.hive.avro.HiveAvroSerDeManager;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -361,6 +362,11 @@
         if (HiveUtils.isPartitioned(this.dataset.table)) {
           this.sourcePartitions = HiveUtils.getPartitionsMap(multiClient.getClient(source_client), this.dataset.table,
               this.partitionFilter, this.hivePartitionExtendedFilter);
+          for (Map.Entry<List<String>, Partition> partition : this.sourcePartitions.entrySet()) {
```
ibuenros: Please don't put this code here. `HiveCopyEntityHelper` is a generic class that should not have anything avro specific. You can look at `HiveAvroCopyEntityHelper` to see how this kind of modification was done before.

arjun4084346: Moved to HiveAvroCopyEntityHelper.

gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/avro/HiveAvroCopyEntityHelper.java
```diff
@@ -72,6 +74,40 @@ public static void updateTableAttributesIfAvro(Table targetTable, HiveCopyEntity
...
+        Path oldAvroSchemaPath = new Path(oldAvroSchemaURL);
+        URI sourceFileSystemURI = hiveHelper.getDataset().getFs().getUri();
+
+        if (PathUtils.isAbsoluteAndSchemeAuthorityNull(oldAvroSchemaPath) || (oldAvroSchemaPath.toUri().getScheme().equals(sourceFileSystemURI.getScheme())
+            && oldAvroSchemaPath.toUri().getAuthority().equals(sourceFileSystemURI.getAuthority()))) {
+
+          String newAvroSchemaURL = hiveHelper.getTargetPathHelper()
+              .getTargetPath(oldAvroSchemaPath, hiveHelper.getTargetFileSystem(), Optional.<Partition>absent(), true)
+              .toString();
+
```
arjun4084346: instead of relying on the table's newAvroSchemaURL, I am now calculating newAvroSchemaURL for every partition. The reason being is every partition has its own schema file, though they all might be same.

PR #1957, opened 2017-06-19
Added additional timers to kafka source and Hive publisher.

ibuenros:


ibuenros:

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -143,7 +161,9 @@ private boolean createDbIfNotExists(IMetaStoreClient client, String dbName) thro
 
       try {
         client.createDatabase(db);
-        log.info("Created database " + dbName);
+        try (Timer.Context context = this.metricContext.timer(CREATE_HIVE_DATABASE).time()) {
```
htran1: Don't we need the create database call in the try?

gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -145,6 +154,8 @@ private void setLimiterReportKeyListToWorkUnits(List<WorkUnit> workUnits, List<S
 
   @Override
   public List<WorkUnit> getWorkunits(SourceState state) {
+    this.metricContext = Instrumented.getMetricContext(state, KafkaSource.class);
+
```
htran1: What about timing isTopicQualified() to get an idea on the cost of the schema fetch?

gobblin-hive-registration/src/main/java/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ibuenros: Yes, thanks. Fixed.

gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
ibuenros: Done.

PR #1959, opened 2017-06-20
Use a thread local kakfa consumer client for each thread in the Kafka…

htran1:
…Source. Cap the avg fetch time used for binpacking to 100 milliseconds. This is to avoid bursty topics from having a large negative impact on the binpacking.

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -121,7 +121,10 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
   private final AtomicInteger offsetTooEarlyCount = new AtomicInteger(0);
   private final AtomicInteger offsetTooLateCount = new AtomicInteger(0);
 
-  private GobblinKafkaConsumerClient kafkaConsumerClient;
+  // sharing the kafka consumer may result in contention, so support thread local consumers
+  private static final ThreadLocal<GobblinKafkaConsumerClient> kafkaConsumerClient =
+          new ThreadLocal<GobblinKafkaConsumerClient>();
+  private GobblinKafkaConsumerClient sharedKafkaConsumerClient = null;
   private final ClassAliasResolver<GobblinKafkaConsumerClientFactory> kafkaConsumerClientResolver =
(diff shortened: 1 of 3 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -591,11 +602,33 @@ private void startAtLatestOffset() {
     @Override
     public void run() {
       try (Timer.Context context = metricContext.timer(WORK_UNITS_FOR_TOPIC_TIMER).time()) {
+        // use shared client if configure, otherwise create a thread local one
+        if (KafkaSource.this.sharedKafkaConsumerClient != null) {
+          KafkaSource.this.kafkaConsumerClient.set(KafkaSource.this.sharedKafkaConsumerClient);
+        } else {
+          KafkaSource.this.kafkaConsumerClient.set(
```
ibuenros: Isn't this essentially a local variable at this point? You're not reusing it within the thread, every `WorkUnitCreator` creates a new one.

htran1:
@ibuenros, please review.

gobblin-modules/gobblin-kafka-common/src/main/java/gobblin/source/extractor/extract/kafka/KafkaSource.java
htran1: The run() method sets the ThreadLocal using KafkaSource.this.kafkaConsumerClient.set(). This sets one kafkaConsumerClient per thread, so that the references to kafkaConsumerClient elsewhere in KafkaSource, such as getWorkunits() will use the one for allocated for the thread.

ibuenros: But aren't you setting one `KafkaConsumerClient` per topic instead?

htran1: Yes, it is one per topic, but set as a thread local to remove contention. The workunit creation time is more than 10 times faster than without this fix, so the instantiation overhead seems minimal when compared to the reduced contention.
Created workunits for 1706 topics in 20 seconds.

I'll create a pool of clients and set and return them within the run() method to address this concern, but it looks like the overhead is fairly minimal.

PR #1962, opened 2017-06-20
addOptionalWatermarkPredicate function

jenniferzheng:
Moved the code to add the watermark predicate in getExtractQuery to a separate function called addOptionalWatermarkPredicate because SqlQueryUtils.addPredicate does not allow for group by or order by in the inputQuery. So an extractor that extends JDBCExtractor can override addOptionalWatermarkPredicate and allow it to use a query with group by/order by.


ibuenros:
@zxcware can you review?

PR #1964, opened 2017-06-20
adding delimited identifier to column names in sql queries

arjun4084346:
adding back tick to column names so it can execute sql query which has sql keywords as column names.

@htran1 @ibuenros  Please review.


gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -317,11 +317,13 @@ public void extractMetadata(String schema, String entity, WorkUnit workUnit) thr
           JsonObject jsonObject = gson.fromJson(jsonStr, JsonObject.class).getAsJsonObject();
           targetSchema.add(jsonObject);
           headerColumns.add(targetColumnName);
+          sourceColumnName = "`" + sourceColumnName + "`";
```
ibuenros: Different systems might have different escape characters. Please double check this, and probably we don't want to make a global change.

htran1: Yes, backtick is for mysql. Oracle uses double quote and the content within double quotes is case sensitive, so in most cases the column name should be uppercased before enclosing in quotes. SQLServer uses square brackets.

arjun4084346: Done.
SQLServer supports it when quoted_identifier option is set on. MySQL supports it when ANSI mode is on. All other systems support double quotes by default.
Default for SQLServer is [ ] , default for MySQL is backtick.

htran1: What about case-sensitivity? If you put double quotes around lowercase identifiers in Oracle, you might not get the desired results.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -1147,6 +1149,20 @@ private String toCase(String targetColumnName) {
     return columnName;
   }
 
+  /**
+   * Default DelimitedIdentifier should be 'double quotes',
+   * but that would make the column name case sensitive in case of Oracle.
+   * @return leftDelimitedIdentifier
+   */
+
+  public String getLeftDelimitedIdentifier() {
```
htran1: Instead of making making the default delimiter empty, can you leave it as double quote, but make it optional for all databases? Users who need the escaping can set the parameter and we won't break existing configs.  Some databases like SqlServer have server-side config for case sensitivity and escaping lowercase names may not give desired behavior.

gobblin-api/src/main/java/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -459,6 +459,9 @@
       "source.querybased.promoteUnsignedIntToBigInt";
   public static final boolean DEFAULT_SOURCE_QUERYBASED_PROMOTE_UNSIGNED_INT_TO_BIGINT = false;
 
+  public static final String ENABLE_DELIMITED_IDENTIFIER = "enable.delimited.identifier";
+  public static final boolean DEFAULT_DELIMITED_IDENTIFIER = false;
```
htran1: Convention is DEFAULT_<key>, so this should be DEFAULT_ENABLE_DELIMITED_IDENTIFIER.

arjun4084346: done

PR #1965, opened 2017-06-22
Add AsyncHttpJoinConverter

yukuai518:
1. Add ApacheHttpAsyncClient
2. Add AsyncHttpJoinConverter inheriting AsyncConverter1To1
3. Add processStream to InstrumentedConverterDecorator to workaround convertRecord unsupported operation issue.
4. Replace threadSafe annotation to a javax package version
5. Remove doOnComplete from AsyncConverter1To1 because it will duplicate shutdown logic

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/audit/KafkaAuditCountHttpClient.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/audit/KafkaAuditCountHttpClient.java
@@ -8,7 +8,6 @@ import com.google.gson.JsonParser;
 import gobblin.configuration.State;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.http.annotation.ThreadSafe;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpUriRequest;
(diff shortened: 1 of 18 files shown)


yukuai518:
@ibuenros @zxcware I am thinking we should keep both async and sync version. What's your opinion?

gobblin-compaction/src/main/java/gobblin/compaction/audit/KafkaAuditCountHttpClient.java
```diff
@@ -27,7 +26,6 @@
  * to perform audit count query.
  */
 @Slf4j
-@ThreadSafe
```
yukuai518: Because this ThreadSafe is from http-core4.4.4, which seems not forward compatible. Change it to javax.annotation.ThreadSafe

gobblin-core-base/src/main/java/gobblin/instrumented/converter/InstrumentedConverterDecorator.java
```diff
@@ -91,4 +92,9 @@ public State getFinalState() {
   public Object getDecoratedObject() {
     return this.embeddedConverter;
   }
+  @Override
```
yukuai518: Done

ibuenros:
LGTM. I have no preference around keeping an sync version or not.

PR #1972, opened 2017-06-26
Config client thread safe

ibuenros:


ibuenros:
@htran1 can you review?

PR #1974, opened 2017-06-26
Fix issue with job id generation in gobblin cluster when using the in…

htran1:
…ternal scheduler by cloning the properties that get mutated during job execution. This prevents the state in the scheduler from getting affected by the job execution.


htran1:
@abti please review.

PR #1978, opened 2017-06-26
Improve compatibility for avro version 1.7.4

zxcware:
- Use avro 1.7.4 apis in `FlattenNestedKeyConverter`
- Remove avro 1.8.1 checks in `AvroUtils`


yukuai518:
Will it cause "HttpOperation is not a GenericRecord" issue for HttpJoinConverter?

zxcware:
@yukuai518 That issue is caused by compiling `HttpOperation.avsc` with avro version 1.8.1. The fix in this PR is unrelated to that issue. I think it will continue be an error unless we downgrade the avro version from 1.8.1 to 1.7.4 in gobblin.

PR #1981, opened 2017-06-27
JdbcExtractor throws RuntimeException on join operation

zxcware:
## Calcite
Calciate is introduced to assist `JdbcExtractor` with sql query parsing and analyzing. Now, `JdbcExtractor` uses calcite `SqlParser`, which, by default, chooses oracle syntax, to detect join operation in a select query

## `gobblin-sql` module
This extension is intended to contain gobblin components that are working with sql related sub-systems. 
- Move Jdbc, MySql, Oracle, SqlServer, Teradata related components that were in `gobblin-core` to `gobblin-sql`
- Transform `ConverterInitializerFactory` and `WriterInitializerFactory` in `gobblin-core` to get initializer from specific class by reflection


gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -76,6 +76,8 @@
  */
 public abstract class JdbcExtractor extends QueryBasedExtractor<JsonArray, JsonElement>
     implements SourceSpecificLayer<JsonArray, JsonElement>, JdbcSpecificLayer {
+  public static final String JOIN_OPERATION_REGEX = "([a-zA-Z_].*)\\.([a-zA-Z_].*)(\\s*)=(\\s*)([a-zA-Z_].*)\\.([a-zA-Z_].*)";
```
htran1: Identifier names can have numbers. Is the problem the '=' in the where or is it the multiple tables in the from? There are also other join predicates, such as !=.
Would a more appropriate check be for multiple tables in the FROM list?

yukuai518: This regular may overmatch the clause like "a1.site='www.linkedin.com'".

zxcware: Good catch, guys. Checking if there are multiple tables in the `from` clause is a more appropriate.

gobblin-core/src/main/java/gobblin/source/extractor/extract/jdbc/JdbcExtractor.java
```diff
@@ -1125,6 +1129,35 @@ private Schema getCustomColumnSchema(String columnName) {
...
+    final String spaceRegex = " +";
+    String[] items = queryLowercase.substring(tableListStartIndex).trim().split(spaceRegex);
+
+    // "From tableA,tableB"
+    if (items.length == 1) {
+      return items[0].contains(",");
+    }
+
+    // Cases like "FROM tableA, tableB <SQL KEYWORD>"
+    return items[0].contains(",") || items[1].contains(",");
```
htran1: I think this doesn't handle the following case: select a.fromLoc from a aliasA , b aliasB

zxcware: I'll introduce https://github.com/JSQLParser/JSqlParser/wiki/Examples-of-SQL-parsing

zxcware:
@htran1 @ibuenros @yukuai518 Please review. Thanks.

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
```diff
@@ -68,15 +70,27 @@ private static ConverterInitializer newInstance(State state, WorkUnitStream work
...
     for (String converterClass : converterClasses) {
-      if (AvroToJdbcEntryConverter.class.getName().equals(converterClass)) {
-        if (workUnits.isSafeToMaterialize()) {
-          cis.add(new AvroToJdbcEntryConverterInitializer(state, workUnits.getMaterializedWorkUnitCollection(),
-              factory, branches, branchId));
-        } else {
-          throw new RuntimeException(AvroToJdbcEntryConverter.class.getName() + " does not support work unit streams.");
-        }
+      Class<?> klazz;
+      try {
```
ibuenros: I would prefer if this one was less reflection oriented:

```java
((Converter) Class.forName(converterClass).newInstance()).getInitializer(...)
```

gobblin-core/src/main/java/gobblin/writer/initializer/WriterInitializerFactory.java
```diff
@@ -61,14 +63,25 @@ private static WriterInitializer newSingleInstance(State state, WorkUnitStream w
...
     String writerBuilderClass = state.getProp(writerBuilderKey, ConfigurationKeys.DEFAULT_WRITER_BUILDER_CLASS);
 
-    if(JdbcWriterBuilder.class.getName().equals(writerBuilderClass)) {
-      if (workUnits.isSafeToMaterialize()) {
-        return new JdbcWriterInitializer(state, workUnits.getMaterializedWorkUnitCollection(),
-            new JdbcWriterCommandsFactory(), branches, branchId);
-      } else {
-        throw new RuntimeException(JdbcWriterBuilder.class.getName() + " does not support work unit streams.");
-      }
+    Class<?> klazz;
```
ibuenros: Ditto

gobblin-core/src/main/resources/mysql/mysql.properties
```diff
@@ -16,7 +16,7 @@
 #
 
 # Source properties - source class to extract data from Mysql Source
-source.class=gobblin.source.extractor.extract.jdbc.MysqlSource
+source.class=gobblin.source.jdbc.MysqlSource
```
ibuenros: Changing the fully qualified name of the class is dangerous, as all deployed applications will need to be updated. Is this needed?

gobblin-core/src/main/java/gobblin/converter/initializer/ConverterInitializerFactory.java
zxcware: Good point. I'll add the method `getInitializer` to the whole `Converter` or `DataWriterBuilder` type hierarchy.

gobblin-core/src/main/resources/mysql/mysql.properties
zxcware: With calcite introduced, all deployed application will need to be updated, adding a new dependency. 

To me, the `MysqlSource` was given a bad fully qualified name, so are some other sql related components. My plan is to piggyback those name changes on adding the calcite as a new dependency while updating the application.

ibuenros: Yes, but it's significantly harder to change both at the same time. For example, there tend to be a lot of jobs that use the MySqlSource, while adding the dependency requires changing a single file.

It is true that is a bad name, but we have to live with those decisions...

zxcware: Yeah. I'll roll back the name change full source, converter, and writer.

gobblin-api/src/main/java/gobblin/converter/Converter.java
```diff
@@ -121,4 +123,8 @@ public State getFinalState() {
     outputStream = outputStream.doOnComplete(this::close);
     return inputStream.withRecordStream(outputStream, outputSchema);
   }
+
+  public ConverterInitializer getInitializer(State state, WorkUnitStream workUnits, int branches, int branchId) {
+    return null;
```
ibuenros: Maybe return a Noop `ConverterInitializer` instead? That avoids null checks and NPEs.

gobblin-api/src/main/java/gobblin/writer/DataWriterBuilder.java
```diff
@@ -125,6 +129,10 @@
     return this;
   }
 
+  public WriterInitializer getInitializer(State state, WorkUnitStream workUnits, int branches, int branchId) {
+    return null;
```
ibuenros: Ditto

PR #1982, opened 2017-06-27
Change AvroHttpJoinConverter type from HttpJoinConverter to AsyncHttpJoinConverter 

yukuai518:
1) Remove duplicate init() from AsyncConverter1To1
2) Change AvroHttpJoinConverter type from HttpJoinConverter to AsyncHttpJoinConverter 
3) Fix the UnsupportedOperationException issue during prevFsSnapshot addAll() because prevFsSnapshot may be an ImmutableSet type


gobblin-core-base/src/main/java/gobblin/converter/AsyncConverter1to1.java
```diff
@@ -65,7 +65,6 @@
   @Override
   public RecordStreamWithMetadata<DO, SO> processStream(RecordStreamWithMetadata<DI, SI> inputStream,
       WorkUnitState workUnitState) throws SchemaConversionException {
-    init(workUnitState);
```
zxcware: Where else is the `init` function called?

yukuai518: InstrumentedConverterDecorator will invoke the init

PR #1986, opened 2017-06-28
Added partition and table information on HiveWorkUnit

jinhyukchang:
- Added partition and table information on HiveWorkUnit
- Added a capability of taking filesystem uri from job property for HiveSource

Passed unit test and integration test via MR mode.


jinhyukchang:
@abti Could you take a look when you have a chance? Thanks!

PR #1987, opened 2017-06-28
Add configurable EventMetadataGenerator that is used to generate addi…

htran1:
…tional metadata to emit in the timing events. The ClusterEventMetadataGenerator is an implementation of this interface that emits processed record count to the JOB_COMPLETE event and error messages to the JOB_FAILED event.

### DIFF
--- a/gobblin-cluster/src/main/java/gobblin/cluster/ClusterEventMetadataGenerator.java
+++ b/gobblin-cluster/src/main/java/gobblin/cluster/ClusterEventMetadataGenerator.java
@@ -24,7 +24,8 @@ import com.google.common.collect.ImmutableMap;
 
 import gobblin.annotation.Alias;
 import gobblin.configuration.ConfigurationKeys;
-import gobblin.metrics.event.TimingEvent;
+import gobblin.metrics.event.EventName;
+import gobblin.runtime.EventMetadataUtils;
 import gobblin.runtime.JobContext;
 import gobblin.runtime.TaskState;
 import gobblin.runtime.api.EventMetadataGenerator;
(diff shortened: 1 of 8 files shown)


htran1:
@ibuenros, please review.

gobblin-runtime/src/main/java/gobblin/runtime/api/EventMetadataGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * For generating additional event metadata to associate with an event.
+ */
+public interface EventMetadataGenerator {
+  /**
+   * Generate a map of additional metadata for the specified event name.
+   * @param eventName the event name used to determine which additional metadata should be emitted
+   * @return {@link Map} with the additional metadata
+   */
+  Map<String, String> getMetadata(String eventName);
```
ibuenros: What about passing the `JobContext` here and using default constructors instead? Also, it might be better to use an `Enum` instead of `String` as the argument.

gobblin-cluster/src/main/java/gobblin/cluster/ClusterEventMetadataGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+  private JobContext jobContext;
+
+  public ClusterEventMetadataGenerator(JobContext jobContext) {
+    this.jobContext = jobContext;
+  }
+
+  public Map<String, String> getMetadata(String eventName) {
+    switch (eventName) {
+      case TimingEvent.LauncherTimings.JOB_COMPLETE:
+        return ImmutableMap.of("processedCount", Long.toString(getProcessedCount()));
```
ibuenros: Please make strings constants.

gobblin-cluster/src/main/java/gobblin/cluster/ClusterEventMetadataGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return ImmutableMap.of();
+  }
+
+  /**
+   * Get the number of records written by all the writers
+   * @return Sum of the writer records written count across all tasks
+   */
+  private long getProcessedCount() {
```
ibuenros: This is a suggestion that doesn't need to be done now, but might be useful to have eventually:

Various `EventMetadataGenerator`s might be interested in emitting records processed count for example, so it might be useful to refactor this into a separate class that can be called from any generator. Similarly for the task failure exception.

PR #1988, opened 2017-06-29
Sample records for err file in row quality checker.

ibuenros:


ibuenros:
@htran1 can you review?

PR #1993, opened 2017-07-03
[GOBBLIN-6] Support eventual consistent filesystems like S3

treff7es:
In our environment we use Gobblin for shipping logs to s3. Gobblin on it's own works pretty well but at a couple of place it assumes the underlying fs is consistent which is not true in some case like in some S3 region.
To overcome this I added a couple of retry if the created dir/file fails to exist right away after publish.
The following changes were added:
- Refactored RetryWriter to gobblin-core to be able use in WriterUtils and not having circular dependencies.
- Introducing mkdirsWithRecursivePermissionWithRetry to be able set retry if directory does not exists right after creation on eventual consistent fs.
- Adding retry to publisher (data.publisher.retry.enabled=true) like TimestampDataPublisher, TimePartitinedDataPublisher to support eventual consisteny filesystem targets
- Tmp fs can be specified with compaction.tmp.fs in compaction job to be able use hdfs for tmp fs and store result on S3. Earlier it was not possible to use differnet fs for tmp and target.
- Retry can be set for compaction if you don't want to fail right away if directory fails to show up right away which can happen on eventual consistent fs (compaction.retry.enabled=true)
- Adding dataset name for compaction mr job name which makes significantly easier to identify which compaction job belongs to which dataset.
- Some minor modification to support non avro extensions

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/util/RecordCountProvider.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/util/RecordCountProvider.java
@@ -37,7 +37,7 @@ public abstract class RecordCountProvider {
    * Convert a {@link Path} from another {@link RecordCountProvider} so that it can be used
    * in {@link #getRecordCount(Path)} of this {@link RecordCountProvider}.
    */
-  public Path convertPath(Path path, RecordCountProvider src) {
+  public Path convertPath(Path path, String extension, RecordCountProvider src) {
     if (this.getClass().equals(src.getClass())) {
       return path;
     }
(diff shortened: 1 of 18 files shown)


treff7es:
I split up my previous pull request into multiple one (#1686) . This one is about supporting eventual consistent filesystems. I will create a separate pull request for the json specific changes which is a separate module. 
This pr also contains change which will be needed for the following pr as well to be able to support non .avro file extensions.

ibuenros:
@htran1 can you take a look?

htran1:
@abti, can you also take a look since you reviewed the original PR that this was split from?

abti:
Issue: https://issues.apache.org/jira/browse/GOBBLIN-6

Please update your PR title with following prefix: [GOBBLIN-6]

yukuai518:
@treff7es Is it possible to create an ConsistentFsWrapper which converts all inconsistent operations in S3 to a consistent version. This will avoid a lot of tmpFs creation and make it more reusable for other modules.

treff7es:
Sure, will do!
Thanks for the feedback.

abti:
@yukuai518 @treff7es I too like the idea of refactoring and creating ConsistentFsWrapper, however since this PR has been open from 2 months and can potentially be helpful for others in its current shape - I will merge it in now. 

I have created a JIRA to track that: https://issues.apache.org/jira/browse/GOBBLIN-246

PR #1994, opened 2017-07-10
Add compaction complete event

yukuai518:
@htran1 Can you help review?

### DIFF
--- a/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
+++ b/gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
@@ -10,13 +10,13 @@ import gobblin.metrics.event.EventSubmitter;
 import gobblin.runtime.TaskContext;
 import gobblin.runtime.mapreduce.MRTask;
 
-import java.util.HashMap;
 import java.util.List;
 import java.io.IOException;
 import java.util.Map;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.mapreduce.Job;
+import com.google.common.collect.ImmutableMap;
 
 /**
  * Customized task of type {@link MRTask}, which runs MR job to compact dataset.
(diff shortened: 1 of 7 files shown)


gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
```diff
@@ -34,6 +37,8 @@ public MRCompactionTask(TaskContext taskContext) throws IOException {
     this.suite = CompactionSuiteUtils.getCompactionSuiteFactory (taskContext.getTaskState()).
             createSuite(taskContext.getTaskState());
     this.dataset = this.suite.load(taskContext.getTaskState());
+    this.eventSubmitter = new EventSubmitter.Builder(this.metricContext, "gobblin.MRCompactionTask")
```
htran1: "gobblin.MRCompactionTask" is the namespace. Please notify the wimd team that this is the new namespace they should query.

gobblin-compaction/src/main/java/gobblin/compaction/mapreduce/MRCompactionTask.java
```diff
@@ -61,17 +66,27 @@ public void onMRTaskComplete (boolean isSuccess, Throwable throwable) {
...
     } else {
+      submitEvent(CompactionSlaEventHelper.COMPACTION_FAILED_EVENT_NAME);
       super.onMRTaskComplete(false, throwable);
     }
   }
 
+  private void submitEvent(String eventName) {
+    Map<String, String> eventMetadataMap = new HashMap<>();
+    eventMetadataMap.put("datasetUrn", this.dataset.datasetURN());
+    this.eventSubmitter.submit(eventName, eventMetadataMap);
```
htran1: You can use ImmutableMap<String,String>.of("datasetUrn", this.dataset.datasetURN()).
Also, a constant should be created for "datasetUrn" and the other references to "datasetUrn" above should be changed to use the constant.

PR #1995, opened 2017-07-10
submitting event in case of failed R2 request

arjun4084346:
@zxcware  Please review.

### DIFF
--- a/gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
+++ b/gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
@@ -1,5 +1,6 @@
 package gobblin.r2;
 
+
 import java.util.HashSet;
 import java.util.Set;
 
@@ -9,6 +10,8 @@ import lombok.extern.slf4j.Slf4j;
 
 import gobblin.http.ResponseHandler;
 import gobblin.http.StatusType;
+import gobblin.metrics.MetricContext;
+import gobblin.metrics.event.EventSubmitter;
 import gobblin.net.Request;
 import gobblin.utils.HttpUtils;
 
@@ -27,6 +30,7 @@ public class R2RestResponseHandler implements ResponseHandler<RestRequest, RestR
 
   public static final String CONTENT_TYPE_HEADER = "Content-Type";
   private final Set<String> errorCodeWhitelist;
+  MetricContext metricsContext = new MetricContext.Builder("R2ResponseStatus").build();
 
   public R2RestResponseHandler() {
     this(new HashSet<>());
@@ -42,12 +46,15 @@ public class R2RestResponseHandler implements ResponseHandler<RestRequest, RestR
     int statusCode = response.getStatus();
     status.setStatusCode(statusCode);
 
+    EventSubmitter.Builder eventSubmitterBuilder = new EventSubmitter.Builder(metricsContext, this.getClass().getSimpleName());
+
     HttpUtils.updateStatusType(status, statusCode, errorCodeWhitelist);
 
     if (status.getType() == StatusType.OK) {
       status.setContent(response.getEntity());
       status.setContentType(response.getHeader(CONTENT_TYPE_HEADER));
     } else {
+      eventSubmitterBuilder.addMetadata("request", request.toString()).addMetadata("statusCode", String.valueOf(statusCode)).build().submit("R2FailedRequest");
       log.info("Receive an unsuccessful response with status code: " + statusCode);
     }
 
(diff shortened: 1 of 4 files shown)


gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -27,6 +29,7 @@
 
   public static final String CONTENT_TYPE_HEADER = "Content-Type";
   private final Set<String> errorCodeWhitelist;
+  MetricContext metricsContext = new MetricContext.Builder("R2ResponseStatus").build();
```
zxcware: If the context should be shared among a writer and its response handler, it can be passed in on initialization

gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -42,12 +45,15 @@ public R2ResponseStatus handleResponse(Request<RestRequest> request, RestRespons
     int statusCode = response.getStatus();
     status.setStatusCode(statusCode);
 
+    EventSubmitter.Builder eventSubmitterBuilder = new EventSubmitter.Builder(metricsContext, this.getClass().getSimpleName());
```
arjun4084346: I just realized that properties can be overridden here. Bringing eventSubmitter out of handleResponse()...
"gobblin.http" will be a suitable namespace?

gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -26,28 +29,37 @@
 public class R2RestResponseHandler implements ResponseHandler<RestRequest, RestResponse> {
 
   public static final String CONTENT_TYPE_HEADER = "Content-Type";
+  private final String R2_RESPONSE_EVENT_NAMESPACE = "gobblin.http.r2.response";
```
zxcware: Change it to `r2.response`, as we can't assume the parent namespace?

gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -26,28 +29,37 @@
...
+  private final String R2_FAILED_REQUEST = "R2FailedRequest";
+  private final String REQUEST = "request";
```
zxcware: Move `REQUEST` and `STATUS_CODE` to `HttpConstants`. They may be used by other kinds of response handler, say `HttpResponseHandler`

gobblin-modules/gobblin-http/src/main/java/gobblin/writer/AsyncHttpWriterBuilder.java
```diff
@@ -86,6 +87,7 @@
     }
 
     this.state = (WorkUnitState) state;
+    this.metricContext = Instrumented.getMetricContext(this.state, AsyncDataWriter.class);
```
zxcware: Use `AsyncHttpWriter.class`

gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -26,28 +29,37 @@
...
+  private final String STATUS_CODE = "statusCode";
   private final Set<String> errorCodeWhitelist;
+  MetricContext metricsContext;
+  EventSubmitter.Builder eventSubmitterBuilder;
 
   public R2RestResponseHandler() {
-    this(new HashSet<>());
+    this(new HashSet<>(), Instrumented.getMetricContext(new State(), AsyncDataWriter.class));
```
zxcware: Use `R2RestResponseHandler.class`?

gobblin-modules/gobblin-http/src/main/java/gobblin/r2/R2RestResponseHandler.java
```diff
@@ -26,28 +29,37 @@
...
     int statusCode = response.getStatus();
     status.setStatusCode(statusCode);
-
     HttpUtils.updateStatusType(status, statusCode, errorCodeWhitelist);
 
     if (status.getType() == StatusType.OK) {
       status.setContent(response.getEntity());
       status.setContentType(response.getHeader(CONTENT_TYPE_HEADER));
     } else {
+      eventSubmitterBuilder.addMetadata(REQUEST, request.toString()).addMetadata(STATUS_CODE, String.valueOf(statusCode))
```
yukuai518: I think you should bind metadata with your event instead of putting in the builder. It means:

1. eventSubmitter should only be built once. 
2. When submit, you can do eventSubmitter.submit (R2_FAILED_REQUEST, metadataMap...)

PR #1998, opened 2017-07-12
Allows the value in a JSON CredentialStore to be treated as base64 or hex encoded

eogren:
- Add hex and base64 encoder/decoders (only decode is actually needed here but adding encode for completeness sake). The codec pieces are separated out into their own standalone classes
in case custom credential store providers want to use them.

- Add a new encoding_type param to config that defaults to hex (which
was previously the only supported option)


eogren:
@htran1 is it possible for you take a look?

gobblin-modules/gobblin-crypto-provider/src/main/java/gobblin/crypto/GobblinEncryptionProvider.java
```diff
@@ -117,6 +117,23 @@ public CredentialStore buildCredentialStore(Map<String, Object> parameters) {
...
+   */
+  private KeyToStringCodec buildKeyToStringCodec(Map<String, Object> parameters) {
+    String encodingName = EncryptionConfigParser.getKeystoreEncoding(parameters);
+    switch (encodingName) {
+      case HexKeyToStringCodec.TAG:
+        return new HexKeyToStringCodec();
+      case Base64KeyToStringCodec.TAG:
+        return new Base64KeyToStringCodec();
+      default:
+        throw new IllegalArgumentException("Don't know how to build hex to string codec for type " + encodingName);
```
htran1: Change message to remove mention to hex since other formats are supported.

eogren: whoops - thanks!

PR #1999, opened 2017-07-12
Deserialize to the outputSchema specified in the convertRecord() call…

htran1:
… in LiKafkaByteArrayMsgToAvroConverter.
This is required when processing a stream of records with a mix of schema versions. The latest schema needs to be used in this case so that all output records share a schema.


htran1:
@ibuenros, please review.

htran1:
@yukuai518, please review.

PR #2000, opened 2017-07-13
Added control messages to Gobblin stream.

ibuenros:


gobblin-runtime/src/test/java/gobblin/runtime/TestRecordStream.java
```diff
@@ -0,0 +1,237 @@
...
+
+    // Create a TaskExecutor - a real TaskExecutor must be created so a Fork is run in a separate thread
+    TaskExecutor taskExecutor = new TaskExecutor(new Properties());
+
+    // Create the Task
+    Task realTask = new Task(mockTaskContext, mockTaskStateTracker, taskExecutor, Optional.<CountDownLatch> absent());
+    Task task = spy(realTask);
+    doNothing().when(task).submitTaskCommittedEvent();
+
+    // The first run of the Task should fail
```
htran1: Comment doesn't match assert.

PR #2003, opened 2017-07-17
Add lead time functionality to DatePartitionedRetrievers

eogren:
Add lead time functionality to DatePartitionedRetrievers. Directories
that were created after (now() - lead time) will be ignored by the
source.

 - If anyone can think of better naming than 'lead time' please let me know - thought about using maxWatermark but I figured that could be confusing
 - I put the config parsing code in a separate class to be consistent with other pieces of Gobblin, although I know there could be an argument for just putting the method as part of the PartitionAwareFileRetriever interface


eogren:
@htran1 or @ibuenros , time to take a look?

PR #2005, opened 2017-07-18
Acking

ibuenros:
Adds record and control message acking.

Although the changes are large, most of it is adding unit tests, so the review is not actually that big.


ibuenros:
@htran1 can you review?

gobblin-api/src/main/java/gobblin/stream/StreamEntity.java
```diff
@@ -17,29 +17,151 @@
...
 
 /**
  * An entity in the Gobblin ingestion stream.
+ *
+ * Note:
+ * When transforming or cloning a stream entity, it is important to construct the new entity in the correct way to ensure callbacks
+ * are spread correctly:
+ *
+ * 1-to-1 entity transformation constructor:
+ *   super(oldEntity, true);
```
htran1: The name below is upstreamEntity.

gobblin-api/src/main/java/gobblin/stream/StreamEntity.java
```diff
@@ -17,29 +17,151 @@
...
+  /**
+   * @return a clone of this {@link StreamEntity}. Implementations need not worry about the callbacks, they will be set
+   * automatically.
+   */
+  protected abstract StreamEntity<D> buildClone();
+
+  public ForkCloner forkCloner() {
+    return new ForkCloner();
+  }
+
```
htran1: Please add javadoc for the ForkCloner and ForkedEntityBuilder.

PR #2007, opened 2017-07-20
Release 0.11.0

yukuai518:
@ibuenros Can you review?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,96 @@
+GOBBLIN 0.11.0
+-------------
+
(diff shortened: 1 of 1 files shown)


sahilTakiar:
Release notes contain:

```
+* [CredentialStore][Eric Ogren] Added a test credential store and associated provider that can be used for integration testing.
 +* [CredentialStore] [Eric Ogren] Refactored `CredentialStore` factory into its own top-level class.
```

I assume its suppose to say the PR number instead of the author, which would be consist with the rest of the entries.

PR #2008, opened 2017-07-20
[GOBBLIN-174] fix distcp-ng, so it does not remove existing target files

arjun4084346:
fixing fullPathdiff() so it does not remove existing target files in case NO_DELETE delete method is specified.

@abti  @ibuenros  Kindly review.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -619,10 +619,27 @@ public class HiveCopyEntityHelper {
       if (shouldCopy) {
         builder.copyFile(sourcePath);
       } else {
-        // if not copying, we want to keep the file in the target
+        // If not copying, we want to keep the file in the target
         // at the end of this loop, all files in targetExistingPaths will be marked for deletion, so remove this file
         targetExistingPaths.remove(newPath);
         desiredTargetExistingPaths.remove(newPath);
+      }
+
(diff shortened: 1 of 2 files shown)


gobblin-data-management/src/main/java/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -619,11 +619,28 @@ protected static DiffPathSet fullPathDiff(HiveLocationDescriptor sourceLocation,
...
+      if (targetExistingPaths.containsKey(newPath) && shouldCopy) {
+        builder.deleteFile(newPath);
+        targetExistingPaths.remove(newPath);
+        desiredTargetExistingPaths.remove(newPath);
+      }
+    }
+
+    // If deleteMethod is NO_DELETE, clear targetExistingPaths,
+    // so nothing goes to pathsToDelete
+    if (helper.deleteMethod == DeregisterFileDeleteMethod.NO_DELETE) {
```
ibuenros: This logic is incorrect. NO_DELETE should only apply when the location of the table is changing. As it is right now, distcp-ng will not be able to sync directories when NO_DELETE is set.

ibuenros: I believe the easier change is to add `existingTargetTable = Optional.absent();` on line 72, which was probably  missed when adding the `UPDATE_TABLE` option.

arjun4084346: done

PR #2012, opened 2017-07-21
Ensure that the field name is present in the exception text if an error is thrown from the encryptor

eogren:
Also, some of our encryptors throw IllegalArgumentException; explicitly catch those as well as IOExceptions


eogren:
Self-merging as this is trivial

PR #2013, opened 2017-07-21
Add support for a FlushControlMessage. The DataWriter is modified to …

htran1:
…have a flush() method that defaults to doing nothing. Implementation of the flush() method was added to the DataWriters where an implementation was possible.

Changed Extractor.recordStream() to receive records to process from readStreamEntity(). The default implementation of readStreamEntity() returns the result of readRecordEnvelope().


htran1:

htran1:
@abti, please review.

PR #2014, opened 2017-07-24
Add support for publishing aggregated writer metrics in the metadata.json optionally produced by BaseDataPublisher.

eogren:
- Each writer will emit a set of internal writer metric tuples
(writerId, partitionKey, branchId, list of [filename, numRecords in
file]) in its job state. If a PartitionedWriter is being used there may
be multiple keys emitted - it does not try to do any rollup itself

- The BaseDataPublisher will go through each set of writer metrics and
group them by (partitionKey, branchId). An aggregated set of those
metrics will be published to the metadata file in the appropriate output
paths.

- A global Num-Records/Num-Files field is also published as dataset
level metadata for each (partitionKey, branchId) generated by the job

### DIFF
--- a/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
@@ -20,12 +20,13 @@ package gobblin.publisher;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.apache.hadoop.conf.Configuration;
@@ -53,6 +54,9 @@ import gobblin.util.HadoopUtils;
 import gobblin.util.ParallelRunner;
 import gobblin.util.WriterUtils;
 import gobblin.util.reflection.GobblinConstructorUtils;
+import gobblin.writer.FsDataWriter;
+import gobblin.writer.FsWriterMetrics;
+import gobblin.writer.PartitionIdentifier;
 
 
 /**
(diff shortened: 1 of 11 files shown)


eogren:
@abti and @htran1  - can you take a look? Will putting metrics in the state like this cause any long term problems around state store that I'm not aware of?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -481,45 +507,44 @@ public void publishMetadata(WorkUnitState state)
...
-          HadoopUtils.deletePath(fs, metadataOutputPath, false);
-        }
-        LOG.info("Writing metadata for branch " + String.valueOf(branchId) + " to " + metadataOutputPath.toString());
-        try (FSDataOutputStream outputStream = fs.create(metadataOutputPath)) {
-          outputStream.write(metadataValue.getBytes(StandardCharsets.UTF_8));
-        }
-      } catch (IOException e) {
-        LOG.error("Metadata file is not generated: " + e, e);
+      //Delete the file if metadata already exists
+      if (fs.exists(metadataOutputPath)) {
```
htran1: Should the metadata be merged instead of deleted? We have cases where an output directory can have additional data files added to an existing directory by a different job.

gobblin-core/src/main/java/gobblin/writer/FsDataWriter.java
```diff
@@ -146,9 +149,9 @@ public FsDataWriter(FsDataWriterBuilder<?, ?> builder, State properties) throws
       this.defaultMetadata.addTransferEncoding(c.getTag());
     }
 
-    String partitionPath = builder.getPartitionPath(properties);
+    this.partitionKey = builder.getPartitionPath(properties);
     if (builder.getPartitionPath(properties) != null) {
-      properties.setProp(ConfigurationKeys.WRITER_PARTITION_PATH_KEY + builder.getWriterId(), partitionPath);
+      properties.setProp(ConfigurationKeys.WRITER_PARTITION_PATH_KEY + builder.getWriterId(), partitionKey);
```
htran1: Should a separator be added between the base key name and the writer id?

gobblin-core/src/main/java/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -416,56 +406,92 @@ private ParallelRunner getParallelRunner(FileSystem fs) {
...
 
-  private void mergeAndPublishMetadata(Collection<? extends WorkUnitState> states)
-      throws IOException {
-    Set<String> partitions = new HashSet<>();
-
-    // There should be one merged metadata file per branch; first merge all of the pieces together
-    mergeMetadataAndCollectPartitionNames(states, partitions);
+    partitions.removeIf(Objects::isNull);
 
     // Now, pick an arbitrary WorkUnitState and publish its metadata. We assume that publisher config settings
```
htran1: Doesn't the metadata being merged come from all the `WorkUnitState`s?

PR #2015, opened 2017-07-24
[GOBBLIN-173] Add pattern support for job-level blacklist in distcpNG/replication

autumnust:
Updates July 25th:
After discussion with @abti and feature requestor we figured out the desired output is a configStore-based dataset finder plus glob-pattern support for filtering out some unwanted trash files. Only with ConfigStore this is impossible as configstore only have a logical view of dataset configuration, which is not necessary the same as the real dataset. Also configStore is not necessary to be implemented based on file system.

The approach here is to rely on configStore to finish dataset discovery first, and resolve the black/whitelist in the `ConfigBasedDataset.java` level, which in current stage only handles file system based dataset. 

Also here we specified the semantics of two-level black/whitelist. The job-level black/whitelist is mainly for hot-fix in convenience of operation purpose, and it will NOT override config-based black/whitelist. That is, job-level black/whitelist is an enhancement not overriding.

### DIFF
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
@@ -154,9 +154,10 @@ public abstract class ConfigBasedDatasetsFinder implements DatasetsFinder {
   }
 
   /**
-   * Semantic of two-level black/whitelist:
+   * Semantic of black/whitelist:
    * - Whitelist always respect blacklist.
-   * - Job-level black/whitelist is NOT OVERRIDING dataset-level black/whitelist but enhance it.
+   * - Job-level blacklist is reponsible for dataset filtering instead of dataset discovery. i.e.
+   *   There's no implementation of job-level whitelist currently. 
    */
   protected Set<URI> getValidDatasetURIs(Path datasetCommonRoot) {
     Collection<URI> allDatasetURIs;
--- a/gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
+++ b/gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
@@ -87,19 +87,13 @@ public abstract class ConfigBasedDatasetsFinder implements DatasetsFinder {
   public static final String GOBBLIN_CONFIG_STORE_DATASET_COMMON_ROOT =
       ConfigurationKeys.CONFIG_BASED_PREFIX + ".dataset.common.root";
 
-  // In addition to the white/blacklist tags, this configuration let the user to black/whitelist some datasets
+  // In addition to the white/blacklist tags, this configuration let the user to whitelist some datasets
   // in the job-level configuration, which is not specified in configStore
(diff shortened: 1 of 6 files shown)


autumnust:
@htran1 Can you help on review? Thanks.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
```diff
@@ -298,13 +304,32 @@ protected void executeItertorExecutor(Iterator<Callable<Void>> callableIterator)
...
+      FileStatus[] matchedPaths = fs.globStatus(mergedPath);
+      for (FileStatus matchedPath: matchedPaths) {
+        List<FileStatus> fileStatuses = FileListUtils.listFilesRecursively(fs, matchedPath.getPath(), new PathFilter() {
+          @Override
+          public boolean accept(Path path) {
+            return true;
+          }
+        });
+        for (FileStatus fileStatus : fileStatuses) {
+          resultURIs.add(new URI(PathUtils.getPathWithoutSchemeAndAuthority(fileStatus.getPath()).toString()));
```
htran1: Please check whether stripping the scheme and authority is the proper thing to do here.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedDatasetsFinder.java
```diff
@@ -167,14 +172,15 @@ public ConfigBasedDatasetsFinder(FileSystem fs, Properties jobProps) throws IOEx
...
       }
     }
 
     try {
       // get all the URIs which imports {@link #replicationTag} or all from whitelistURNs
       allDatasetURIs = this.whitelistURNs.isPresent()
-          ? this.whitelistURNs.get().stream().map(u -> this.datasetURNtoURI(u)).collect(Collectors.toList()) : configClient.getImportedBy(new URI(whitelistTag.toString()), true);
+          ? this.whitelistURNs.get().stream().flatMap(u -> this.datasetURNtoURI(u).stream()).collect(Collectors.toList())
+          : configClient.getImportedBy(new URI(whitelistTag.toString()), true);
       populateDisabledURIs(disabledURISet);
```
htran1: The configuration options are not symmetric here, which can lead to confusion and does not allow all override possibilities. The code is ignoring the config store whitelist tag when a job-level whitelist is specified, but it is adding the job-level blacklist to the tag-based blacklist.

gobblin-data-management/src/main/java/gobblin/data/management/copy/replication/ConfigBasedMultiDatasets.java
```diff
@@ -138,9 +168,30 @@ private void generateDatasetInPullMode(ReplicationConfiguration rc, URI executio
...
+          if (blacklistFilteringHelper(configBasedDataset, this.blacklist)){
+            this.datasets.add(configBasedDataset);
+          }
+        }
+      }
+    }
+  }
+
+  @VisibleForTesting
+  public boolean blacklistFilteringHelper(ConfigBasedDataset configBasedDataset, Optional<List<Pattern>> patternList){
```
htran1: Make this package private and add javadoc. The name is not vey clear as to the meaning of the return value.

PR #2017, opened 2017-07-25
updating connection url for sqlserver with database name

arjun4084346:
updating connection url for sqlserver with database name
@htran1  Kindly review.

### DIFF
--- a/gobblin-modules/gobblin-sql/src/main/java/gobblin/source/jdbc/SqlServerExtractor.java
+++ b/gobblin-modules/gobblin-sql/src/main/java/gobblin/source/jdbc/SqlServerExtractor.java
@@ -187,7 +187,8 @@ public class SqlServerExtractor extends JdbcExtractor {
   public String getConnectionUrl() {
     String host = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
     String port = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_PORT);
-    String url = "jdbc:sqlserver://" + host.trim() + ":" + port;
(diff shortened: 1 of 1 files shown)


gobblin-modules/gobblin-sql/src/main/java/gobblin/source/jdbc/SqlServerExtractor.java
```diff
@@ -187,7 +187,8 @@ public String getRightDelimitedIdentifier() {
   public String getConnectionUrl() {
     String host = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_HOST_NAME);
     String port = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_PORT);
-    String url = "jdbc:sqlserver://" + host.trim() + ":" + port;
+    String database = this.workUnitState.getProp(ConfigurationKeys.SOURCE_QUERYBASED_SCHEMA);
+    String url = "jdbc:sqlserver://" + host.trim() + ":" + port + ";databaseName=" + database.trim();
```
htran1: Please check why existing code works for our SQLServer use cases and whether this change could break them.

arjun4084346: Existing jobs are using dbname in the query itself, e.g. https://gitli.corp.linkedin.com/ds-platform/gobblin-stand-alone/source/5d3424495b32d5b02a63014a4901066b97d3ab12:conf/job-conf/lynda_esp/lyndageneral_esp_svunsubscribelogs_inc.pull

Nevertheless, making changes to use this key only if it is set. as this is an optional key and not required to make a connection.

PR #2026, opened 2017-07-28
[GOBBLIN-29] Allow GobblinHelixJobScheduler to disable the super class services

kadaan:
gobblin-helix has it's own job-configuration-manager and therefore should not be using the one from `JobScheduler`.  This change allows subclasses of `JobScheduler` to override the default services that `JobScheduler` starts.  This will allow `GobblinHelixJobScheduler` to use it's own method of getting job configuration.  Currently `GobblinHelixJobScheduler` has no test coverage, I didn't add any.


abti:
+1 waiting for Travis

PR #2035, opened 2017-08-01
[GOBBLIN-33] Allow state store locations to be independently configured for intermediate and final state storage

kadaan:
The current implementation of Gobblin Helix requires that both intermediate state storage (task and workunit state) and final state storage (dataset state) be on the same filesystem.  This PR allows the intermediate state store and final state store to be configured independently, while falling back to a common default.


kadaan:
@abti Can you give this a look and tell me if this seems like the right approach?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -144,7 +145,12 @@ public GobblinHelixJobLauncher(Properties jobProps, final HelixManager helixMana
 
     jobConfig = ConfigUtils.propertiesToConfig(jobProps);
 
-    this.stateStores = new StateStores(jobConfig, appWorkDir,
+    Config stateStoreJobConfig = ConfigUtils.propertiesToConfig(jobProps)
+        .withValue(ConfigurationKeys.STATE_STORE_FS_URI_KEY, ConfigValueFactory.fromAnyRef(
```
kadaan: It wasn't passed before and I don't need it, so I'm only making an incremental improvement that fixes my issue.

PR #2037, opened 2017-08-04
[GOBBLIN-159] Allow Helix jobs to be gracefully canceled

kadaan:
The current implementation of GobblinHelixJob does not allow them to be canceled.  This means that when the master node is restarted, jobs are left in a zombie state.  These zombie jobs consume resources and write output, but do not count as a job.  This leads to reduced cluster performance and duplicate data.  This PR adds the ability for jobs to be gracefully cancelled.

@abti Can you look at this and see if this is headed the right direction.  Also, is there an easy way to add helix tests that would cover this scenario?

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJob.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJob.java
@@ -33,9 +33,7 @@ import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
 
 import org.apache.gobblin.annotation.Alpha;
-import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.metrics.Tag;
-import org.apache.gobblin.runtime.JobException;
 import org.apache.gobblin.runtime.JobLauncher;
 import org.apache.gobblin.runtime.listeners.JobListener;
 import org.apache.gobblin.scheduler.BaseGobblinJob;
@@ -74,7 +72,7 @@ public class GobblinHelixJob extends BaseGobblinJob implements InterruptableJob 
               Boolean.toString(GobblinClusterConfigurationKeys.JOB_EXECUTE_IN_SCHEDULING_THREAD_DEFAULT)))) {
         jobScheduler.runJob(jobProps, jobListener, jobLauncher);
       } else {
-        cancellable = jobScheduler.submitJob(jobProps, jobListener, jobLauncher);
+        cancellable = jobScheduler.scheduleJobImmediately(jobProps, jobListener, jobLauncher);
       }
     } catch (Throwable t) {
       throw new JobExecutionException(t);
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/scheduler/JobScheduler.java
```diff
@@ -230,6 +236,57 @@ public void scheduleJob(Properties jobProps, JobListener jobListener)
...
+      public void run() {
+        try {
+          runJob(jobProps, jobListener, jobLauncher);
+        } catch (JobException je) {
+          LOG.error("Failed to run job " + jobProps.getProperty(ConfigurationKeys.JOB_NAME_KEY), je);
+        }
+      }
+    };
+    final Future<?> future = this.jobExecutor.submit(runnable);
+    this.submitRunnableToExecutor(runnable);
```
zxcware: What are you trying to do with line 250 and line 251? Do you need both or line 250 is good enough?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJob.java
```diff
@@ -65,28 +70,32 @@ public void executeImpl(JobExecutionContext context) throws JobExecutionExceptio
...
-        jobScheduler.submitRunnableToExecutor(runnable);
+        cancellable = jobScheduler.submitJob(jobProps, jobListener, jobLauncher);
       }
     } catch (Throwable t) {
       throw new JobExecutionException(t);
     }
   }
+
+  @Override
+  public void interrupt() throws UnableToInterruptJobException {
```
kadaan: As far as it seemed to me, `interrupt` is used like cancel within Gobblin.  Is there another use case?

zxcware: It has other unpredicted use cases. The fact is that this `interrupt` may be invoked unexpectedly by the framework or system, like a thread may be interrupted by os as preemption to schedule another thread. Usually, we can't guarantee it's only invoked by Gobblin.

kadaan: That isn't really my take of how Quartz interrupt is designed.  They specifically call out on the FAQ: 
> ## How do I stop a Job that is currently executing?
> See the org.quartz.InterruptableJob interface, and the Scheduler.interrupt(String, String) method.

Additionally, in the current Gobblin implementation, `interrupt` is called when shutting down `JobScheduler`.  The issue is that the jobs don't implement `interrupt` so nothing happens.  

I can add a boolean that gets set when the JobScheduler starts shutting down.  The Jobs can check for that bool when `interrupt` is called and cancel only if that is set.

abti: I think this is good safety net. LGTM with the flag.

PR #2038, opened 2017-08-04
[GOBBLIN-189] add dataset path to 'dataset published' events

arjun4084346:
Dear Gobblin maintainers,

@ibuenros Kindly review.

### JIRA

### Description
It will add dataset path in gobblintrackingevent_distcp_ng, which can be used to find the , partition type and dataset path of the datasel.

### Tests
Adding one more field to events is trivial. Hence test is not added

### Commits


gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
```diff
@@ -86,10 +89,17 @@
...
 
-  @lombok.Builder(builderClassName = "Builder", builderMethodName = "_hiddenBuilder")
   public CopyableFile(FileStatus origin, Path destination, OwnerAndPermission destinationOwnerAndPermission,
       List<OwnerAndPermission> ancestorsOwnerAndPermission, byte[] checksum, PreserveAttributes preserve,
       String fileSet, long originTimestamp, long upstreamTimestamp, Map<String, String> additionalMetadata) {
+    this(origin, destination, destinationOwnerAndPermission, ancestorsOwnerAndPermission, checksum, preserve, fileSet,
+        originTimestamp, upstreamTimestamp, additionalMetadata, "");
+  }
+
+  @lombok.Builder(builderClassName = "Builder", builderMethodName = "_hiddenBuilder")
```
ibuenros: The point of the builder pattern is to not have to create a new constructor when we add a new parameter (as long as all accesses are through a builder). I think the above constructor can be removed.

gobblin-data-management/src/main/java/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -226,6 +230,7 @@ private void publishFileSet(CopyEntity.DatasetAndPartition datasetAndPartition,
 
     additionalMetadata.put(SlaEventKeys.SOURCE_URI, this.state.getProp(SlaEventKeys.SOURCE_URI));
     additionalMetadata.put(SlaEventKeys.DESTINATION_URI, this.state.getProp(SlaEventKeys.DESTINATION_URI));
+    additionalMetadata.put(SlaEventKeys.DATASET_OUTPUT_PATH, fileSetRoot.or(""));
```
ibuenros: Perhaps it is better to make it clear the output path is unknown instead of leaving it empty.

gobblin-data-management/src/main/java/gobblin/data/management/copy/CopyableFile.java
arjun4084346: Yes, all the test cases are using the old constructor directly. I will change them to use builder.

PR #2039, opened 2017-08-04
[GOBBLIN-202] Added JMX reporting of task execution queue size and time.

kadaan:
Autoscaling the Gobblin AWS worker nodes based on CPU is not very efficient.  It is better to scale based on the amount of time that tasks wait in the queue before running.  This PR added some JMX metrics to TaskExecutor that can be used as part of a custom CloudWatch metric for autoscaling in AWS.  

Pulling these JMX metrics and creating the custom CloudWatch metric are outside the scope of this PR.  We do this with a lambda function that every minute:

1. Looks up all live instances of the worker ASG in AWS
1. Find the private/public DNS for the live instances
1. Communicates with jolokia to pull the count and time metrics
1. Computes a global average and logs it

There is a CloudWatch log filter setup that turns the logged metric into a custom CloudWatch metric.  The ASG has been changed to use the new metric to control scaling.


ibuenros:
@kadaan is there a Jira ticket? Can you link it here?

kadaan:
@ibuenros Here you go: [GOBBLIN-202](https://issues.apache.org/jira/browse/GOBBLIN-202)

PR #2049, opened 2017-08-09
[GOBBLIN-195] Ability to switch Avro schema namespace switch before registering with Kafka Avro Schema registry

abti:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-195

### Description
Ability to switch Avro schema namespace switch before registering with Kafka Avro Schema registry. This is useful when we want to maintain backward compatibility with the schema registered in the registry since registry does not allows for difference in namespace. 
This however has no impact on write / read of actual Avro bytes for the event because they are mapped back to the fields irrespective of namespace. 

Through this, if the following config is set, it will help gobblin-metrics be backwards compatible with the schema already registered in Kafka Avro Schema Registry: 

kafka.schemaRegistry.overrideNamespace = org.apache.gobblin.metrics:gobblin.metrics

### Tests
Added

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
@@ -65,11 +65,9 @@ public class KafkaAvroReporterUtil {
       }
 
       // If no entry found in the config value, mark it absent
-      if (namespaceOverridesMap.size() == 0) {
-        return Optional.<Map<String, String>>absent();
+      if (namespaceOverridesMap.size() != 0) {
+        return Optional.of(namespaceOverridesMap);
       }
-
-      return Optional.of(namespaceOverridesMap);
     }
 
     return Optional.<Map<String, String>>absent();
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gobblin.metrics.reporter.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import org.apache.gobblin.kafka.schemareg.KafkaSchemaRegistryConfigurationKeys;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+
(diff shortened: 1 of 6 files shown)


abti:
@htran1 Please review.

gobblin-utility/src/main/java/org/apache/gobblin/util/AvroUtils.java
```diff
@@ -612,6 +614,127 @@ public static Schema switchName(Schema schema, String newName) {
...
+    Schema newSchema;
+    String newNamespace = StringUtils.EMPTY;
+
+    // Process all Schema Types
+    // (Primitives are simply cloned)
+    switch (schema.getType()) {
+      case ARRAY:
+        newSchema = Schema.createArray(switchNamespace(schema.getElementType(), namespaceOverride));
+        break;
+      case BOOLEAN:
```
htran1: A more compact organization would be to group the primitive types together at the beginning with fall through to avoid repeating newSchema = Schema.create(schema.getType()).

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
```diff
@@ -0,0 +1,77 @@
...
+        List<String> override = Lists.newArrayList(SPLIT_BY_COLON.split(namespaceOverride));
+        if (override.size() != 2) {
+          throw new RuntimeException("Namespace override should be of the format originalNamespace:replacementNamespace,"
+              + " found: " + namespaceOverride);
+        }
+        namespaceOverridesMap.put(override.get(0), override.get(1));
+      }
+
+      // If no entry found in the config value, mark it absent
+      if (namespaceOverridesMap.size() == 0) {
```
htran1: Can reverse condition here and handle this case with fall through to the return at the end of the method.

gobblin-utility/src/main/java/org/apache/gobblin/util/AvroUtils.java
```diff
@@ -612,6 +614,127 @@ public static Schema switchName(Schema schema, String newName) {
...
+        break;
+      default:
+        String exceptionMessage = String.format("Schema namespace replacement failed for \"%s\" ", schema);
+        LOG.error(exceptionMessage);
+
+        throw new AvroRuntimeException(exceptionMessage);
+    }
+
+    // Copy schema metadata
+    copyProperties(schema, newSchema);
```
htran1: How come current switchName() doesn't copy properties?

gobblin-utility/src/test/java/org/apache/gobblin/util/AvroUtilsTest.java
```diff
@@ -196,6 +200,27 @@ public void testSwitchName() {
...
+    map.put(originalNamespace, newNamespace);
+    Schema newSchema = AvroUtils.switchNamespace(schema, map);
+
+    Assert.assertEquals(newSchema.getNamespace(), newNamespace);
+    Assert.assertEquals(newSchema.getName(), originalName);
+    for(Schema.Field field : newSchema.getFields()) {
+      Assert.assertEquals(field, schema.getField(field.name()));
+    }
+
+    Assert.assertEquals(newNamespace, AvroUtils.switchNamespace(schema, map).getNamespace());
```
htran1: Isn't this the same check as line 215?

gobblin-utility/src/main/java/org/apache/gobblin/util/AvroUtils.java
abti: Changed

gobblin-utility/src/main/java/org/apache/gobblin/util/AvroUtils.java
abti: Seems like a miss in case of switchName(), but I would prefer to update it with another PR so that we dont break any behavior that depends on that bug :)

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/util/KafkaAvroReporterUtil.java
abti: Yes, reduces LOC. Changed.

gobblin-utility/src/test/java/org/apache/gobblin/util/AvroUtilsTest.java
abti: Yes, its redundant. Removed.

PR #2061, opened 2017-08-14
[GOBBLIN-209] Support for hocon global configuration files.

kadaan:
Currently, when using Job files, global configuration can only be properties.  Hocon should also be supported for global configuration.  This PR enables support for global configuration files when their extension is `.configuration`.


ibuenros:
+1

PR #2081, opened 2017-08-28
[GOBBLIN-231] Grok to json converter

treff7es:
You can convert text data to json based on a Grok pattern

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-231

### Description
 GrokToJsonConverter accepts already deserialized text row, String, where you can use.

 Converts Text to JSON based on Grok pattern. Schema is represented by the form of JsonArray same interface being used by CsvToJonConverter.
 Each text record is represented by a String.
 The converter only supports Grok patterns where every group is named because it uses the group names as column names.
 
 The following config properties can be set:

-  The grok pattern to use for the conversion:
` converter.grok_to_json.pattern=^%{IPORHOST:clientip} (?:-|%{USER:ident}) (?:-|%{USER:auth}) \[%{HTTPDATE:timestamp}\] \"(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|-)\" %{NUMBER:response} (?:-|%{NUMBER:bytes})
` 

-  Path to the grok patterns (if not set it will use the default ones):
` converter.grok_to_json.patterns=/tmp/grok_patterns` 

-  Treat empty string as null value:

` converter.grok_to_json.empty_as_null=true` 
 

- Specify the null string:

`converter.grok_to_json.null_string=null` 
 
Example of schema:
```
 [
  {
    "columnName": "Day",
    "comment": "",
    "isNullable": "true",
    "dataType": {
      "type": "string"
    }
  },
  {
    "columnName": "Pageviews",
    "comment": "",
    "isNullable": "true",
    "dataType": {
      "type": "long"
    }
  }
]

```

### Tests

### Commits


ibuenros:
@treff7es is there a jira for this PR? Can you create one and update the title and body oft he PR?

treff7es:
@ibuenros Sorry, I was not sure I can create jira issue. Now I created and updated the title and body.

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+      throws DataConversionException {
+    JsonObject outputRecord = new JsonObject();
+
+    Match gm = grok.match(inputRecord);
+    gm.captures();
+    JsonParser parser = new JsonParser();
+    JsonElement capturesJson = parser.parse(gm.toJson());
+
+    for (JsonElement anOutputSchema : outputSchema) {
+      String key = anOutputSchema.getAsJsonObject().get(COLUMN_NAME_KEY).getAsString();
```
zxcware: Can use a variable represent `anOutputSchema.getAsJsonObject()`

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+
+  @VisibleForTesting
+  JsonObject createOutput(JsonArray outputSchema, String inputRecord)
+      throws DataConversionException {
+    JsonObject outputRecord = new JsonObject();
+
+    Match gm = grok.match(inputRecord);
+    gm.captures();
+    JsonParser parser = new JsonParser();
+    JsonElement capturesJson = parser.parse(gm.toJson());
```
zxcware: Based on how it's used below, we're more interested in `parser.parse(gm.toJson()).getAsJsonObject()`

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+          default:
+            outputRecord.addProperty(key, capturesJson.getAsJsonObject().get(key).getAsString());
+        }
+      }
+    }
+    return outputRecord;
+  }
+
+  private boolean isFieldNull(JsonElement capturesJson, String key) {
+    if (!capturesJson.getAsJsonObject().has(key)) {
```
zxcware: With early `return`, you don't need `else if`

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+    JsonParser parser = new JsonParser();
+    JsonElement capturesJson = parser.parse(gm.toJson());
+
+    for (JsonElement anOutputSchema : outputSchema) {
+      String key = anOutputSchema.getAsJsonObject().get(COLUMN_NAME_KEY).getAsString();
+      String type = anOutputSchema.getAsJsonObject().getAsJsonObject(DATA_TYPE).get(TYPE_KEY).getAsString();
+      Boolean nullable = anOutputSchema.getAsJsonObject().get(NULLABLE).getAsBoolean();
+
+      if (isFieldNull(capturesJson, key)){
+        if (!nullable) {
```
zxcware: Can directly use `anOutputSchema.getAsJsonObject().get(NULLABLE).getAsBoolean()`

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+ * Converts Text to JSON based on Grok pattern. Schema is represented by the form of JsonArray same interface being used by CsvToJonConverter.
+ * Each text record is represented by a String.
+ * The converter only supports Grok patterns where every group is named because it uses the group names as column names.
+ *
+ * The following config properties can be set:
+ * The grok pattern to use for the conversion:
+ * converter.grok_to_json.pattern ="^%{IPORHOST:clientip} (?:-|%{USER:ident}) (?:-|%{USER:auth}) \[%{HTTPDATE:timestamp}\] \"(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|-)\" %{NUMBER:response} (?:-|%{NUMBER:bytes})"
+ *
+ * Path to the grok patterns (if not set it will use the default ones):
+ * converter.grok_to_json.patterns=
```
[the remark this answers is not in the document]
treff7es: Patterns is a list of predefined GROK patterns and pattern is the actual pattern you can use. You can specify any GROK expression from your patterns in your pattern like here %{IPORHOST:clientip} to be able to use IPORHOST you need to load that pattern in advance.
I agree the naming is confusing, so I will come up with some better and more descriptive one.

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * Treat empty string as null value:
+ * converter.grok_to_json.empty_as_null=true
+ *
+ * Specify the null string:
+ * converter.grok_to_json.null_string="null"
```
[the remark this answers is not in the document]
treff7es: empty_as_null means if the string is empty then it will be null in the json
null_string is the string which will be null in the converted json like you can use the "None" null string which will be null in the converted json. 
I will add more documentation about it.

treff7es:
@zxcware Thanks for the review, I have updated my pull request.

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * The following config properties can be set:
+ * The grok pattern to use for the conversion:
+ * converter.grok_to_json.pattern ="^%{IPORHOST:clientip} (?:-|%{USER:ident}) (?:-|%{USER:auth}) \[%{HTTPDATE:timestamp}\] \"(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|-)\" %{NUMBER:response} (?:-|%{NUMBER:bytes})"
+ *
+ * Path to the file which contains the base grok patterns which can be used in the converter's GROK pattern (if not set it will use the default ones):
+ * converter.grok_to_json.base_patterns=
+ *
+ * Treat empty string as json null value. If it is true the empty string will be converted to json null:
+ * converter.grok_to_json.empty_as_null=true
```
zxcware: Can we merge the configuration `converter.grok_to_json.empty_as_null` and `converter.grok_to_json.null_string`? We can have something like
`converter.grokToJsonConverter.stringsAsNull=,None`?

Also, the configuration key names in the comments should be updated.

treff7es: Sorry, I missed those but I fixed them now.
I also extended empty as null to treat string which only contains whitespaces as empty.
I still think there is value in having null string and empty as null separate as there can be usecase where people have a specific null string like '-' or 'null' but they would like to treat strings which only contains whitespaces or no character as null. 
The idea is pretty similar like here: http://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-conversion.html#copy-emptyasnull and in our environment are using this option and the null string one as well separately.
If you guys still don't see the value, I'm happy to remove it but I think it is useful and even it could be useful at the csv converter.
What do you think?

zxcware: I do see the value in making certain strings as null, but I don't see the value in using 2 separate configuration keys to achieve that. Only one configuration would suffice. Another idea is to have a configuration like `converter.grokToJsonConverter.nullStringRegexes`, which is a list of regex patterns identifying strings that should be replaced as null. Check out `DatasetFilterUtils` to see how a list of patterns is used.

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
```diff
@@ -0,0 +1,223 @@
...
+    "dataType": {
+      "type": "long"
+    }
+  }
+]
+ */
+public class GrokToJsonConverter extends Converter<String, JsonArray, String, JsonObject> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(GrokToJsonConverter.class);
+  private static final JsonParser JSON_PARSER = new JsonParser();
```
zxcware: Sorry to mislead you earlier. I found a json parser is schema specific and it can't be shared to parse a second schema if the two have some record field type definitions in common.

treff7es: Sorry, I didn't get it fully.

treff7es:
@zxcware Thanks, I really liked your idea about the nullStringRegexes and in that way we can merge the two option what I had earlier and it is more powerful. 
I implemented that and you can check in my latest commit.

gobblin-core/src/main/java/org/apache/gobblin/converter/grok/GrokToJsonConverter.java
zxcware: What I mean is "one schema parser per schema" instead of "a static schema parser for all". Because I found static parser caused some error in my work.

zxcware: Sorry. I forgot that it's a json parser not a schema parser. There is no issue.

treff7es:
@zxcware I fixed the styling and merged all the commits into one. Hopefully this way looks nicer.

treff7es:
@ibuenros @abti @htran1 if you have time and feel so, can you check this and merge, @zxcware checked the pull request already.

abti:
@treff7es done. thanks.

PR #2089, opened 2017-09-01
[GOBBLIN-238] Implement EnvelopePayloadConverter and EnvelopeSchemaDecorator

zxcware:
### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-238

### Description
#### Motivation
The current implementation of `EnvelopeSchemaConverter` has several flaws:
- Assumes top level payload schema field
- Output record is the schema'ed payload but output schema is a String

To address the issues and improve envelope schema conversion, the task implements two types of `EnvelopeSchemaConverter`: `EnvelopePayloadConverter` and `EnvelopeSchemaDecorator`.

#### `EnvelopePayloadConverter`
The previous `EnvelopeSchemaConverter` is transformed into `EnvelopePayloadConverter`. The logic to extract and output the deserialized payload is kept. But Fields projection logic is removed as it's a duplicate effort to `AvroProjectionConverter`.

Given an envelope record, the output schema will be the latest payload schema fetched from a kafka registry. The output record will be the deserialized payload with the latest schema

#### `EnvelopeSchemaDecorator`
Given an envelope record, the output schema will set the payload field to have the latest schema fetched from a kafka registry and set the other fields as they are from the input schema. The output record will set the payload to be the deserialized object with the latest schema and set the other fields as they are from the input record

#### Configurations
One configuration is required to set for any of the converters to work. It has no default value. 
```
// The topic to fetch the latest schema of the payload from a kafka registry
converter.envelopeSchemaConverter.payloadSchemaTopic=
```
The converter supports nested schema id
```
converter.envelopeSchemaConverter.schemaIdField="metadata.payloadSchemaId"
```

### Tests
  1. `EnvelopePayloadConverterTest`
  2. `EnvelopeSchemaDecoratorTest`

### Commits


zxcware:
@jack-moseley @htran1 @ibuenros Please review

abti:
Ignore the Travis failure for Azkaban test, I have disabled the test, and filed a JIRA to fix it: GOBBLIN-239

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -112,8 +113,14 @@ public String convertSchema(Schema inputSchema, WorkUnitState workUnit) throws S
           workUnit.getProp(PAYLOAD_SCHEMA_ID_FIELD) : DEFAULT_PAYLOAD_SCHEMA_ID_FIELD;
       String payloadField = workUnit.contains(PAYLOAD_FIELD) ?
           workUnit.getProp(PAYLOAD_FIELD) : DEFAULT_PAYLOAD_FIELD;
-      String schemaKey = String.valueOf(inputRecord.get(schemaIdField));
+
+      Optional<Object> schemaIdValue = AvroUtils.getFieldValue(inputRecord, schemaIdField);
```
yukuai518: What is the difference between AvroUtils.getFieldValue(inputRecord, schemaIdField) and inputRecord.get(schemaIdField)? Is this only for adding a nullable check?

zxcware: `AvroUtils.getFieldValue` support nested field

zxcware:
@htran1 Please review

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -17,136 +17,141 @@
...
 /**
- * A converter for extracting schema/records from an envelope schema.
- * Input schema: envelope schema - must have fields payloadSchemaId (the schema registry key of the output
- *               schema) and payload (byte data for output record)
- * Input record: record corresponding to input schema
- * Output schema: schema obtained from schema registry using key provided in input record's {@link #PAYLOAD_SCHEMA_ID_FIELD}
- * Output record: record corresponding to output schema obtained from input record's {@link #PAYLOAD_FIELD} as bytes
+ * Base class for an envelope schema converter using {@link KafkaSchemaRegistry}
  */
-public class EnvelopeSchemaConverter extends Converter<Schema, String, GenericRecord, GenericRecord> {
```
yukuai518: Why do we change the signature of this class? This seems to be backward incompatible if anyone is using this class before. What is the class that will represent the old behavior of existing EnvelopeSchemaConverter? It seems like EnvelopePayloadConverter != old EnvelopeSchemaConverter.

Maybe rename your EnvelopeSchemaConverter to BaseEnvelopeSchemaConverter to avoid name conflict?

Do we need a payload converter when upconvert is not needed? This is old EnvelopeSchemaConverter behavior, shall we keep it?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -1,152 +1,145 @@
...
+      Schema payloadSchema = getPayloadSchema(inputRecord);
+      // Set writer schema
+      latestPayloadReader.setSchema(payloadSchema);
+
+      byte[] payloadBytes = getPayloadBytes(inputRecord);
+      Decoder decoder = DecoderFactory.get().binaryDecoder(payloadBytes, null);
+
+      // 'latestPayloadReader.read' will convert the record from 'payloadSchema' to the latest payload schema
+      return latestPayloadReader.read(null, decoder);
+    } catch (Exception e) {
```
htran1: Catch the specific exceptions here if possible.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -1,152 +1,145 @@
...
- * A converter for extracting schema/records from an envelope schema.
- * Input schema: envelope schema - must have fields payloadSchemaId (the schema registry key of the output
- *               schema) and payload (byte data for output record)
- * Input record: record corresponding to input schema
- * Output schema: schema obtained from schema registry using key provided in input record's {@link #PAYLOAD_SCHEMA_ID_FIELD}
- * Output record: record corresponding to output schema obtained from input record's {@link #PAYLOAD_FIELD} as bytes
+ * Base class for an envelope schema converter using {@link KafkaSchemaRegistry}
  */
-public class EnvelopeSchemaConverter extends Converter<Schema, String, GenericRecord, GenericRecord> {
+public abstract class EnvelopeSchemaConverter<P> extends Converter<Schema, Schema, GenericRecord, GenericRecord> {
```
htran1: What about backwards compatibility?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaDecorator.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * <p> Given an envelope record as the input record, the output record will have the payload set
+ * to its deserialized object using the latest schema (see {@code convertPayload(GenericRecord)}).
+ * The converter copies the other fields from the input record to the output record
+ *
+ * <p> If the current payload schema is incompatible with its latest schema, {@code convertPayload(GenericRecord)}
+ * will throw an exception and the job fail
+ */
+
+public class EnvelopeSchemaDecorator extends EnvelopeSchemaConverter<GenericRecord> {
```
yukuai518: Maybe use a better name like EnvelopeSchemaPayloadUpvertConverter?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -17,136 +17,141 @@
...
-    ByteBuffer bb = (ByteBuffer) inputRecord.get(payloadFieldName);
-    byte[] payloadBytes;
+  protected byte[] getPayloadBytes(GenericRecord inputRecord) {
+    ByteBuffer bb = (ByteBuffer) inputRecord.get(payloadField);
     if (bb.hasArray()) {
-      payloadBytes = bb.array();
+      return bb.array();
     } else {
-      payloadBytes = new byte[bb.remaining()];
+      byte[] payloadBytes = new byte[bb.remaining()];
```
yukuai518: Maybe the childclass should overwrite this method instead of putting if/else in the abstract class?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
zxcware: Yes, it's a back incompatible change for 3 reasons
- The old converter returns `String` as the output schema, which makes it rarely used because the majority of our constructs use `Schema` as the schema
- The old converter doesn't have a clear name. It deserializes the payload and ignores the other fields in the envelope schema
- Upconverting to a certain schema is necessary if records of different schemas will be stored into one file, which would fail otherwise. It's like a standardization.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaDecorator.java
zxcware: That's too verbose. Here, a decorator illustrates the payload, adding more details and making it more readable. I agree it's not a very good name and hope the comment would help.

gobblin-modules/gobblin-kafka-common/src/test/java/org/apache/gobblin/converter/EnvelopeSchemaDecoratorTest.java
```diff
@@ -0,0 +1,86 @@
+package org.apache.gobblin.converter;
```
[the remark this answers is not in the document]
zxcware: I just made a new commit with license added.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
zxcware: I guess the converter is rarely used because it returns `String` as the output schema. The cost for this backwards incompatible change would be 0 or negligible.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -1,152 +1,145 @@
...
+          ((Class<? extends KafkaSchemaRegistryFactory>) Class.forName(registryFactoryField)).newInstance();
+      registry = registryFactory.create(workUnit.getProperties());
     } catch (ClassNotFoundException | IllegalAccessException | InstantiationException e) {
-      return null;
+      throw new RuntimeException(e);
     }
-    this.decoderFactory = DecoderFactory.get();
-    this.readers = CacheBuilder.newBuilder().build(new CacheLoader<Schema, GenericDatumReader<GenericRecord>>() {
+
+    schemaCache = CacheBuilder.newBuilder().build(new CacheLoader<String, Schema>() {
```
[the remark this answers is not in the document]
zxcware: Good point. I checked that the existing registry does have cache. I'll remove the schema cache in this converter.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -1,152 +1,145 @@
...
-    byte[] payloadBytes;
+    ByteBuffer bb = (ByteBuffer) payloadValue.get();
     if (bb.hasArray()) {
-      payloadBytes = bb.array();
+      return bb.array();
     } else {
-      payloadBytes = new byte[bb.remaining()];
+      byte[] payloadBytes = new byte[bb.remaining()];
       bb.get(payloadBytes);
+      String hexString = new String(payloadBytes, StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
zxcware: I got the logic in the previous implementation. So far, we don't have a use case for this. It's unclear what the right logic is.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
zxcware: The childclass can always overwrite this method as needed. Here is just a default implementation of the method.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
```diff
@@ -1,152 +1,145 @@
...
+      return DatatypeConverter.parseHexBinary(hexString);
     }
-    String hexString = new String(payloadBytes, StandardCharsets.UTF_8);
-    return DatatypeConverter.parseHexBinary(hexString);
+  }
+
+  protected Schema fetchLatestPayloadSchema() throws Exception {
```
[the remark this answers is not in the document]
zxcware: I don't see the value to throw specific exceptions if we don't do specific handling per exception.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
zxcware: Since we don't do specific exception handling, catching specific exceptions is effectively the same as catching a general `Exception`.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/converter/EnvelopeSchemaConverter.java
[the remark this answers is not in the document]
htran1: Can you check whether the payload is a hex-encoded string? It did look that way in one of the examples.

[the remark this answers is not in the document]
zxcware: In the example, it's not a hex-encoded string. In fact, it can be any format. I'll simply return the payloadBytes as the default implementation and let the subclass overwrite the behavior if needed.

zxcware:
This PR has some back incompatible changes but it has accumulated several commits, which made it not trivial to make amendments. A replacing PR is posted: https://github.com/apache/incubator-gobblin/pull/2099, with the same changes and better class names.

PR #2090, opened 2017-09-05
[GOBBLIN-236] Add a ControlMessage injector as a RecordStreamProcessor

htran1:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-236

### Description
Add a ControlMessage injector as a RecordStreamProcessor.
A `ControlMessageInjector` inspects an incoming record and can inject `ControlMessages` based on the content of the incoming record.

One use case for this is the injection of `MetadataUpdateControlMessages` when the latest schema has been updated to trigger the update of the schema used by other constructs downstream from the `ControlMessageInjector`.

Long running jobs are more likely to encounter issues due to the schema being updated after the start of the job.

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/converter/Converter.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/converter/Converter.java
@@ -25,10 +25,12 @@ import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.converter.initializer.ConverterInitializer;
 import org.apache.gobblin.converter.initializer.NoopConverterInitializer;
+import org.apache.gobblin.metadata.GlobalMetadata;
 import org.apache.gobblin.stream.ControlMessage;
 import org.apache.gobblin.records.ControlMessageHandler;
 import org.apache.gobblin.records.RecordStreamProcessor;
 import org.apache.gobblin.records.RecordStreamWithMetadata;
+import org.apache.gobblin.stream.MetadataUpdateControlMessage;
 import org.apache.gobblin.stream.RecordEnvelope;
 import org.apache.gobblin.source.workunit.WorkUnitStream;
 import org.apache.gobblin.stream.StreamEntity;
@@ -55,6 +57,9 @@ import io.reactivex.Flowable;
  * @param <DO> output data type
  */
 public abstract class Converter<SI, SO, DI, DO> implements Closeable, FinalState, RecordStreamProcessor<SI, SO, DI, DO> {
+  // Metadata containing the output schema. This may be changed when a MetadataUpdateControlMessage is received.
+  private GlobalMetadata<SO> outputGlobalMetadata;
+
   /**
    * Initialize this {@link Converter}.
    *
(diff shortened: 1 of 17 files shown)


htran1:
@ibuenros, please review.

zxcware:
Hi @htran1, could you add more description in the PR or the jira ticket about the motivation and general idea of control message injector? This can help reviewers obtain a direction of what to look for in terms of design and implementation.

gobblin-api/src/main/java/org/apache/gobblin/converter/Converter.java
```diff
@@ -120,16 +125,27 @@ public State getFinalState() {
...
+                ControlMessage out = (ControlMessage) in;
+                // update the output schema with the new input schema from the MetadataUpdateControlMessage
+                if (in instanceof MetadataUpdateControlMessage) {
+                  this.outputGlobalMetadata =
+                      new GlobalMetadata<SO>(convertSchema((SI)((MetadataUpdateControlMessage) in).getGlobalMetadata()
+                                      .getSchema(), workUnitState));
+                  out = new MetadataUpdateControlMessage<SO, DO>(this.outputGlobalMetadata);
+                }
+
                 getMessageHandler().handleMessage((ControlMessage) in);
```
ibuenros: Should this receive `in` or `out`? If the decision is for `in`, then I would prefer this line be before the `MetadataUpdateControlMessage` block.

gobblin-api/src/main/java/org/apache/gobblin/stream/ControlMessageInjector.java
```diff
@@ -0,0 +1,154 @@
...
+  @Getter(AccessLevel.PROTECTED)
+  private GlobalMetadata<SI> inputGlobalMetadata;
+
+  /**
+   * Initialize this {@link ControlMessageInjector}.
+   *
+   * @param workUnitState a {@link WorkUnitState} object carrying configuration properties
+   * @return an initialized {@link ControlMessageInjector} instance
+   */
+  public ControlMessageInjector<SI, DI> init(WorkUnitState workUnitState) {
```
ibuenros: Can this be `protected` instead? Same with all methods except `processStream`.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -173,7 +169,30 @@ public Task(TaskContext context, TaskStateTracker taskStateTracker, TaskExecutor
...
+        this.closer.register((Closeable)r);
+      }
+    }
+
+    List<Converter<?,?,?,?>> converters = this.taskContext.getConverters();
+
+    this.converter = closer.register(new MultiConverter(converters));
+
+    // can't have both record stream processors and converter lists configured
+    if (!this.recordStreamProcessors.isEmpty() && !converters.isEmpty()) {
```
ibuenros: Just to make it a bit clearer, can you do something like `Preconditions.checkState(!this.recordStreamProcessor.isEmpty() ^ !this.converters.isEmpty(), "Mesasge...");`

gobblin-api/src/main/java/org/apache/gobblin/stream/ControlMessageInjector.java
htran1: Done.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
htran1: I need to call the closer here and both empty is valid so ^ won't work.

gobblin-api/src/main/java/org/apache/gobblin/converter/Converter.java
htran1: Okay, moved it before. I think `in` should be the one to be handled so that the handler gets a chance to handle the control message without modifications. The handler can choose to do schema conversion if it needs it.

gobblin-api/src/main/java/org/apache/gobblin/metadata/GlobalMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Global metadata
+ * @param <S> schema type
+ */
+@AllArgsConstructor
+@EqualsAndHashCode
+public class GlobalMetadata<S> implements Copyable<GlobalMetadata<S>> {
+  @Getter
+  private S schema;
+
```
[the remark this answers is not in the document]
htran1: Okay, added a builder.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
htran1: Added precondition check wrapped by a try catch block.

gobblin-api/src/main/java/org/apache/gobblin/converter/Converter.java
```diff
@@ -120,16 +127,29 @@ public State getFinalState() {
...
             .flatMap(in -> {
               if (in instanceof ControlMessage) {
+                ControlMessage out = (ControlMessage) in;
+
                 getMessageHandler().handleMessage((ControlMessage) in);
-                return Flowable.just(((ControlMessage<DO>) in));
+
+                // update the output schema with the new input schema from the MetadataUpdateControlMessage
+                if (in instanceof MetadataUpdateControlMessage) {
+                  this.outputGlobalMetadata = GlobalMetadata.<SI, SO>builderWithInput(inputStream.getGlobalMetadata(),
```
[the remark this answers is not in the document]
htran1: Currently there are no other fields, so it either works. The original idea was to start with the inputStream GlobalMetadata and then overlay from `in`. But I think the suggestion of using `in.getGlobalMetadata()` is good since the copy will be handled in `builderWithInput`.

gobblin-api/src/main/java/org/apache/gobblin/stream/ControlMessageInjector.java
```diff
@@ -0,0 +1,154 @@
...
+
+    Flowable<StreamEntity<DI>> outputStream =
+        inputStream.getRecordStream()
+            .flatMap(in -> {
+              if (in instanceof ControlMessage) {
+                ControlMessage out = (ControlMessage) in;
+                if (in instanceof MetadataUpdateControlMessage) {
+                  setInputGlobalMetadata(((MetadataUpdateControlMessage) in).getGlobalMetadata(),
+                      workUnitState);
+                  out = new MetadataUpdateControlMessage<SI, DI>(this.inputGlobalMetadata);
```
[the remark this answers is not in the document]
htran1: Removed inputGlobalMetadata from this class so that there is no confusion over the propagation of metadata. A `ControlMessageInjector` can store a modified copy of the metadata, but it won't be able to pass that change along to the next construct. Only converters should change schemas.

ibuenros:
+1

PR #2092, opened 2017-09-06
[GOBBLIN-241] Allow multiple datasets send different lineage event for kafka

yukuai518:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-241

### Description
     (1) Add submitLineageEvent() to DataPublisher class. This allows different publisher create its own submission logic. For example, NoopPublisher can simply return success because no submission is not needed.  The default implementation was added inside BaseDataPublisher. The SafeDatasetCommit will simply invoke DataPublisher::submitLineageEvent without changing existing logic.

     (2) Add submitSingleTaskLineageEvent() to SingleTaskDataPublisher class. This allows different task level publisher to submit lineage event. Default implementation is provided inside BaseDataPublisher.

### Tests
    No test needed

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/lineage/LineageInfo.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/lineage/LineageInfo.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.lineage;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -48,7 +49,7 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class LineageInfo {
-
+  public static final String LINEAGE_DATASET_URN = "lineage.dataset.urn";
   public static final String LINEAGE_NAME_SPACE = "gobblin.lineage";
   public static final String BRANCH_ID_METADATA_KEY = "branchId";
   private static final String DATASET_PREFIX =  LINEAGE_NAME_SPACE + ".";
@@ -226,6 +227,17 @@ public class LineageInfo {
 
   public static void setBranchLineageAttribute (State state, int branchId, String key, String value) {
     state.setProp(BRANCH_PREFIX + Joiner.on(".").join(branchId, key), value);
+  }
+
+  public static Map<String, Collection<State>> aggregateByDatasetUrn (Collection<? extends State> states) {
+    Map<String, Collection<State>> datasetStates = new HashMap<>();
+    for (State state: states) {
+      String urn = state.getProp(LINEAGE_DATASET_URN, state.getProp(ConfigurationKeys.DATASET_URN_KEY, ConfigurationKeys.DEFAULT_DATASET_URN));
+      datasetStates.putIfAbsent(urn, new ArrayList<>());
+      Collection<State> datasetState = datasetStates.get(urn);
+      datasetState.add(state);
+    }
+    return datasetStates;
   }
 
   public final String getId() {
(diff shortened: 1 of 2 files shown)


gobblin-api/src/main/java/org/apache/gobblin/publisher/DataPublisher.java
```diff
@@ -125,4 +125,8 @@ public boolean canBeSkipped() {
   protected boolean shouldPublishMetadataFirst() {
     return true;
   }
+
+  public boolean submitLineageEvent (Collection<? extends WorkUnitState> states) {
```
ibuenros: Please name boolean methods with yes/no questions.

gobblin-api/src/main/java/org/apache/gobblin/publisher/SingleTaskDataPublisher.java
```diff
@@ -75,4 +75,8 @@ public static SingleTaskDataPublisher getInstance(Class<? extends DataPublisher>
 
     return (SingleTaskDataPublisher) DataPublisher.getInstance(dataPublisherClass, state);
   }
+
+  public boolean submitSingleTaskLineageEvent (WorkUnitState state) {
```
ibuenros: Ditto

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -292,6 +296,33 @@ private void publishMultiTaskData(WorkUnitState state, int branchId, Set<Path> w
     publishData(state, branchId, false, writerOutputPathsMoved);
   }
 
+  public boolean submitSingleTaskLineageEvent (WorkUnitState state) {
```
ibuenros: Why is this part of `BaseDataPublisher`? Maybe put it in `LineageEvent` instead?

yukuai518:
@ibuenros Can you help review?

ibuenros:
Also, please add Jira number to the PR title.

yukuai518:
@ibuenros We don't have task level lineage event support previously so there is no check about whether events were sent in a task publisher.

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
yukuai518: @ibuenros This will allow individual publisher decide if it needs to submit lineage event. If we move it to LineageEvent. We will still need a hook in the publisher, and that means for all the task publishers we will always send lineage event even if we don't need to.

zxcware: If a `DataPublisher` should have the ability to submit some event, we can add a function `submitEvent(WorkUnitState)` in the `DataPublisher` class.

Another idea is similar to hive registration and audit count, we can have a `DatasetEventPublisher` which sends events at the dataset level but is not be able to send publisher specific event.

gobblin-api/src/main/java/org/apache/gobblin/publisher/DataPublisher.java
zxcware: I guess you can throw an exception as a way to indicate failure instead of returning `false`

gobblin-api/src/main/java/org/apache/gobblin/publisher/SingleTaskDataPublisher.java
yukuai518: @ibuenros Add the aggregation logic in LineageInfo, removed task level lineage event because we are not using this now.

ibuenros:
+1

PR #2101, opened 2017-09-12
[GOBBLIN-248] Converter for Json to Parquet

tilakpatidar:
Dear Gobblin maintainers,

### JIRA

### Description
- [x] This PR adds support for converting JSON records from JSON sources into Parquet format. The user has to define the source.schema which will help in generating Parquet schema and conversion of JSON into Parquet. In future, a ParquetHdfsWriter could be implemented to write Parquet files.
The following data types are supported for now:
STRING
INT
LONG
FLOAT
DOUBLE
BOOLEAN
ARRAY
ENUM
RECORD
MAP
Nesting is not supported for now.
### Tests
org.apache.gobblin.converter.parquet.JsonIntermediateToParquetConverterTest

### Commits


gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
```diff
@@ -0,0 +1,59 @@
...
+ * @author tilakpatidar
+ */
+public class JsonIntermediateToParquetConverter extends Converter<JsonArray, MessageType, JsonObject, ParquetGroup> {
+  private RecordConverter recordConverter;
+
+  @Override
+  public MessageType convertSchema(JsonArray inputSchema, WorkUnitState workUnit)
+      throws SchemaConversionException {
+    String fieldName = workUnit.getExtract().getTable();
+    JsonObject temp = new JsonObject();
```
zxcware: Please provide a more meaningful name, as it's passed to construct a `RecordConverter`

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
```diff
@@ -0,0 +1,248 @@
...
+import parquet.io.api.Binary;
+import parquet.io.api.RecordConsumer;
+import parquet.schema.GroupType;
+import parquet.schema.PrimitiveType;
+
+
+/**
+ * Custom Implementation of {@link Group} to support adding any {@link PrimitiveType} and {@link GroupType}
+ * without their specific type.
+ * Adding a {@link Group} to {@link parquet.example.data.simple.SimpleGroup} using
```
zxcware: Is the comment from line 42 to line 43 related to `ParquetGroup`? Why is it talking about `SimpleGroup`?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
```diff
@@ -0,0 +1,248 @@
...
+ * {@link parquet.example.data.simple.SimpleGroup#addGroup(String)} results in a new
+ * group on which you have to add {@link PrimitiveType} using the API.
+ * This implementation allows adding {@link Group} by passing it to the add method.
+ *
+ * @author tilakpatidar
+ */
+public class ParquetGroup extends Group {
+
+  private final GroupType schema;
+  private final List[] data;
```
zxcware: To facilitate understanding, add comment like `Field data indexed by fieldIndex of the schema`?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
```diff
@@ -0,0 +1,248 @@
...
+import parquet.example.data.simple.Primitive;
+import parquet.io.api.Binary;
+import parquet.io.api.RecordConsumer;
+import parquet.schema.GroupType;
+import parquet.schema.PrimitiveType;
+
+
+/**
+ * Custom Implementation of {@link Group} to support adding any {@link PrimitiveType} and {@link GroupType}
+ * without their specific type.
```
zxcware: Just curious, what happens if a value of a different type is inserted? How many types does Parquet have?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
zxcware: Is this essentially `List<Object>[]`? If so, why not explicitly declare as `List<Object>[]`, making it simpler to understand?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
```diff
@@ -0,0 +1,59 @@
...
+    JsonObject dataType = new JsonObject();
+    dataType.addProperty("type", "record");
+    dataType.add("values", inputSchema);
+    temp.add("dataType", dataType);
+    recordConverter = new RecordConverter(fieldName, false, temp, workUnit, ROOT);
```
zxcware: Where is `RecordConverter`? If it's specific to parquet, it should be put into this converter

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
tilakpatidar: It should only accept PrimitiveTypes and Group and should throw an exception on any other Object type. I will fix this.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
tilakpatidar: SimpleGroup is an implementation of Group provided by the parquet library. But, It does not have an addGroup method which takes a Group. Instead, it returns a group object in which you have to add fields by looping. This makes it difficult to add Groups returned by different converters into the Base Group.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
tilakpatidar: 👍

tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
tilakpatidar: RecordConverter, as well as other Converters, are present in JsonElementConversionFactory.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
tilakpatidar: Parquet only has primitive types. For building complex types such as map or an array, repeatable property and groups can be used on primitive types. The JSON schema is provided by the user in the source.schema property. In this way, the data types supported by the underlying data format is abstracted away from the user. The user provides schema and complex data types are built by the converter. This kind of data types abstraction can be seen in other converters as well. However, if a type defined in source.schema is not supported by the converter it will throw UnsupportedOperationException.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
```diff
@@ -0,0 +1,247 @@
...
+  public String toString() {
+    return this.toString("");
+  }
+
+  public String toString(String indent) {
+    StringBuilder result = new StringBuilder();
+    int i = 0;
+    Iterator<parquet.schema.Type> i$ = this.schema.getFields().iterator();
+
+    while (true) {
```
zxcware: This block from line 71 to 97 has 3 loops. Can we simply the logic as follows?
``` 
StringBuilder result = new StringBuilder();
List<Field> fields = this.schema.getFields();
for (int i = 0; i < fields.length; i++) {
  data[i].forEach(fieldValue -> {
    result.append(indent).append(fields.get(i).getName());
    if (fieldValue == null) {
      result.append(": NULL\n");
    } else if (fieldValue instanceof Group) {
      result.append("\n").append(((ParquetGroup) fieldValue).toString(indent + "  "));
    } else {
      result.append(": ").append(fieldValue.toString()).append("\n");
    }
  });
}
return result.toString();
```

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * Custom Implementation of {@link Group} to support adding {@link Object} of type {@link Primitive} or {@link Group}.
+ * Also provides methods to add {@link Primitive} and {@link Group} with {@link String} key if index is not known.
+ * @author tilakpatidar
+ */
+public class ParquetGroup extends Group {
+
+  private final GroupType schema;
+  //data indexed by fieldIndex of the schema
```
zxcware: The important message here is this each list item represents data of a field, which is indexed by the fieldIndex of the schema. Could you make it clearer?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+package org.apache.gobblin.converter.parquet;
+
+import org.apache.gobblin.configuration.WorkUnitState;
+import org.apache.gobblin.converter.Converter;
+import org.apache.gobblin.converter.DataConversionException;
+import org.apache.gobblin.converter.SchemaConversionException;
+import org.apache.gobblin.converter.SingleRecordIterable;
+import org.apache.gobblin.converter.parquet.JsonElementConversionFactory.RecordConverter;
```
zxcware: I didn't find the `RecordConverter` in the codebase: https://github.com/apache/incubator-gobblin/blob/master/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionFactory.java

tilakpatidar: Sorry, gave you the wrong link. https://github.com/tilakpatidar/incubator-gobblin/blob/f38e4bdd01dbff89245b49134dbef29346d94aa1/gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/ParquetGroup.java
tilakpatidar: 👍

tilakpatidar:
@zxcware suggested changes have been made. Please review.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+     * @param value
+     * @return
+     */
+    abstract Object convertField(JsonElement value);
+
+    /**
+     * Parquet data type after conversion
+     * @return
+     */
+    public abstract PrimitiveTypeName getTargetType();
```
zxcware: With `typeMap`, the default implementation can just return `typeMap.get(sourceType)`. But you need to set the `sourceType` during initialization.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+
+    /**
+     * Returns the source type
+     * @return
+     */
+    public abstract SourceType getSourceType();
```
zxcware: When you pass in the `sourceType` to the constructor, the default implementation can just return it. I don't see the value of this get method, though.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+    @Override
+    public PrimitiveTypeName getTargetType() {
+      throw new UnsupportedOperationException("Complex types does not support PrimitiveTypeName");
+    }
+  }
+
+  /**
+   * For converting complex types which compose same element type.
+   */
+  public static abstract class ComplexConverterForUniformElementTypes extends ComplexConverter {
```
zxcware: Just like java `Object` type is to any java types, is there a super type that unifies all element types? In this case, a single type of `ComplexConverter` can be used for both cases.

It's just an idea. You can drop it if it doesn't satisfy all the requirements.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+      super(fieldName, nullable, repeated);
+    }
+
+    @Override
+    IntegerValue convertField(JsonElement value) {
+      return new IntegerValue(value.getAsInt());
+    }
+
+    @Override
+    public PrimitiveTypeName getTargetType() {
```
zxcware: When the `sourceType` is set during initialization, the `getTargetType` and `getSourceType` methods for the following simple converters can just use the default implementations.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+        throw new UnsupportedOperationException(fieldType + " is unsupported");
+    }
+  }
+
+  /**
+   * Converts a JsonElement into a supported ParquetType
+   * @author tilakpatidar
+   *
+   */
+  public static abstract class JsonElementConverter {
```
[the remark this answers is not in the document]
zxcware: If the targetType isn't necessary to be applied for all cases and it seems not to be used in this PR, should we consider removing it from the base class?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: Realised sourceType is not required however targetType can be passed by specific Converters from constructor to the super. Example, IntConverter can pass INT32 to his super i.e JsonElementConverter

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+
+    /**
+     * Set a {@link JsonElementConverter} for the elements present within Complex data types.
+     * @param elementConverter
+     */
+    protected void setElementConverter(JsonElementConverter elementConverter) {
+      this.elementConverter = elementConverter;
+    }
+
+    public JsonElementConverter getElementConverter() {
```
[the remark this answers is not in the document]
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+     * @param schemaNode
+     * @param itemKey
+     * @return
+     */
+    private SourceType getPrimitiveTypeInSource(JsonObject schemaNode, String itemKey) {
+      String type = schemaNode.get("dataType").getAsJsonObject().get(itemKey).getAsString().toUpperCase();
+      return SourceType.valueOf(type);
+    }
+
+    public PrimitiveTypeName getElementTypeParquet() {
```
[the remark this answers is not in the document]
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+      super(fieldName, nullable, repeated);
+      this.elementPrimitiveName = getPrimitiveTypeInParquet(schemaNode, itemKey);
+      this.elementSourceType = getPrimitiveTypeInSource(schemaNode, itemKey);
+    }
+
+    /**
+     * Set a {@link JsonElementConverter} for the elements present within Complex data types.
+     * @param elementConverter
+     */
+    protected void setElementConverter(JsonElementConverter elementConverter) {
```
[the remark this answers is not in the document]
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,695 @@
...
+    }
+
+    @Override
+    public PrimitiveTypeName getTargetType() {
+      throw new UnsupportedOperationException("Complex types does not support PrimitiveTypeName");
+    }
+  }
+
+  /**
+   * For converting complex types which compose same element type.
```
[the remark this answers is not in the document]
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: 👍

tilakpatidar:
@zxcware Suggested changes have been made. Please review.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final String IS_NULLABLE_KEY = "isNullable";
+  private final DataType type;
+
+  public enum DataType {
+    STRING, INT, LONG, FLOAT, DOUBLE, BOOLEAN, ARRAY, ENUM, RECORD, MAP
+  }
+
+  public JsonSchema(JsonArray jsonArray) {
+    JsonObject temp = buildTempRecord(jsonArray);
+    setSuper(temp);
```
zxcware: Is calling `super.setXXX` is the same as `setXXX`? It can be just `setFromJsonObject(jsonObject)`?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+      return this.getDataType().get(ENUM_SYMBOLS_KEY).getAsJsonArray();
+    }
+    return new JsonArray();
+  }
+
+  /**
+   * Get source type
+   * @return
+   */
+  public DataType getType() {
```
zxcware: I'm confused by `getType` and `getDataType` in the class. What are they?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,505 @@
...
+      typeMap.put(BOOLEAN, PrimitiveTypeName.BOOLEAN);
+      typeMap.put(STRING, BINARY);
+    }
+
+    /**
+     * @param jsonSchema
+     * @param repeated
+     * @param targetType
+     */
+    public JsonElementConverter(JsonSchema jsonSchema, boolean repeated, PrimitiveTypeName targetType) {
```
zxcware: I think `targetType` -> `outputType` is a better name? How do you think of that? It follows the conventions used in gobblin `Converter`s like `inputSchema`, `outputSchema`, `inputRecord`, `outputRecord`, etc.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,505 @@
...
+   */
+  public static abstract class JsonElementConverter {
+    protected final boolean repeated;
+    protected final JsonSchema jsonSchema;
+    private PrimitiveTypeName targetType;
+
+    /**
+     * HashMap to convert {@link DataType} to {@link PrimitiveTypeName}
+     */
+    private final static HashMap<DataType, PrimitiveTypeName> typeMap = new HashMap<>();
```
zxcware: Is this used?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+    return type;
+  }
+
+  /**
+   * Builds a temp element schema based on type
```
zxcware: `temp` gives an impression that the element will not be persisted. However, the `JsonSchema` object returned by this method is actually a base schema which will be customized into what's expected and persisted in the future. 

Can we rename the function to be `buildBaseSchema`?

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+   * @param type
+   * @return
+   */
+  public static JsonSchema buildElementSchema(DataType type) {
+    JsonObject temp = new JsonObject();
```
zxcware: `temp` -> `jsonObject`

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private void setSuper(JsonObject jsonObject) {
+    super.setColumnName(jsonObject.get(COLUMN_NAME_KEY).getAsString());
+    super.setComment(jsonObject.has(COMMENT_KEY) ? jsonObject.get(COMMENT_KEY).getAsString() : "");
+    super.setDataType(jsonObject.get(DATA_TYPE_KEY).getAsJsonObject());
+    super.setDefaultValue(jsonObject.has(DEFAULT_VALUE_KEY) ? jsonObject.get(DEFAULT_VALUE_KEY).getAsString() : "");
+    super.setNullable(jsonObject.has(IS_NULLABLE_KEY) && jsonObject.get(IS_NULLABLE_KEY).getAsBoolean());
+  }
+
+  private JsonObject buildTempRecord(JsonArray jsonArray) {
```
zxcware: I guess I am a bit paranoid about the usage of `temp`. I think a `temp` variable is never supposed to be a return value. It's should be ephemeral and local. Hence, I disagree with the the name here. I would name it as `buildRecord`

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
```diff
@@ -0,0 +1,131 @@
...
+    JsonObject temp = new JsonObject();
```
zxcware: `temp` -> `record`

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
tilakpatidar:
``` js
[
      {
        "columnName": "some_enum",
        "dataType": {
          "type": "enum",
          "symbols": [
            "HELLO",
            "WORLD"
          ]
        }
      }
    ]
```

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
tilakpatidar: I guess, I learned today not to name temp variables if they affect outside the local scope of the method. Thanks 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonSchema.java
tilakpatidar: 👍

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: Not anymore. Forgot to remove it after refactoring.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: Makes sense.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,490 @@
...
+      throw new RuntimeException("Symbol " + value.getAsString() + " does not belong to set " + symbols.toString());
+    }
+
+    @Override
+    public Type schema() {
+      return this.getElementConverter().schema();
+    }
+  }
+
+  public static class RecordConverter extends ComplexConverter {
```
zxcware: After reading more material about Parquet, I understood its terms and concepts more. 

If I understood your goal correctly, the classes hierarchy will be much clearer if it looks like
```
JsonElementConverter
   |- PrimitiveConverter
   |        |- Int, Double, ...
   |- CollectionConverter
   |        |- Array, Map, Enum
   |- RecordConverter
```

The current class structure enforces some subclass to carry what it shouldn't have from the superclass. Like the `PrimitiveTypeName` argument is hardcoded as `INT32` in complex converter. The `elementConverter` defined in `ComplexConverter` is never used in `RecordConverter`.

With the new classes hierarchy I proposed. It will be at ease for you to move shared logic to its right place without breaking the class semantic.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonIntermediateToParquetConverter.java
```diff
@@ -0,0 +1,57 @@
...
+import parquet.schema.MessageType;
+
+import static org.apache.gobblin.converter.parquet.JsonElementConversionFactory.RecordConverter.RecordType.ROOT;
+
+
+/**
+ * A converter to Convert JsonIntermediate to Parquet
+ * @author tilakpatidar
+ */
+public class JsonIntermediateToParquetConverter extends Converter<JsonArray, MessageType, JsonObject, Group> {
```
zxcware: 1. This isn't really a general Parquet converter. Because it only works for `Group` and doesn't work for `Primitive`. How about naming it as `JsonToParquetGroupConverter`?

2. `MessageType` -> `GroupType`?

tilakpatidar: Please see the following links.
MessageType is a root of the schema. [MessageType.java](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/schema/MessageType.java)
GroupType is a list of fields. [GroupType.java](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/schema/GroupType.java)

Data from any source will have RECORD type as the root, thus converting a RECORD type data to MessageType data makes sense. This MessageType will act as a parent for other GroupType which are used to build the MAP, ARRAY etc.

So, MessageType is the root schema. 

We don't need a JsonParquetPrimitiveConverter as the source data will always be a RECORD.
However, I agree that renaming to ```JsonToParquetGroupConverter``` will clarify the usage.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
tilakpatidar: 👍

tilakpatidar:
@zxcware Suggested changes have been made. Please review.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,476 @@
...
+      JsonSchema jsonSchema = JsonSchema.buildBaseSchema(this.elementType);
+      jsonSchema.setColumnName(fieldName);
+      return jsonSchema;
+    }
+
+    /**
+     * Set a {@link JsonElementConverter} for the elements within a collection.
+     * @param elementConverter
+     */
+    protected void setElementConverter(JsonElementConverter elementConverter) {
```
zxcware: If the `elementConverter` is not supposed to change once set, can it be passed as a constructor argument? Then we can remove method `setElementConverter`. It's perfectly find to make it `protected`, then we can get rid of `getElementConverter`.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,476 @@
...
+        ParquetGroup entrySet = (ParquetGroup) mapGroup.addGroup(MAP_KEY);
+        entrySet.add(MAP_KEY_COLUMN_NAME, entry.getKey());
+        entrySet.add(MAP_VALUE_COLUMN_NAME, converter.convert(entry.getValue()));
+      }
+
+      return mapGroup;
+    }
+
+    @Override
+    public GroupType schema() {
```
zxcware: The schema can be computed on initialization and return it here. In this way, every call to `schema()` won't trigger duplicate computation.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,476 @@
...
+        case REQUIRED:
+          return Types.required(BINARY).as(UTF8).named(columnName);
+        default:
+          throw new RuntimeException("Unsupported Repetition type");
+      }
+    }
+  }
+
+  public static class ArrayConverter extends CollectionConverter {
+    private final JsonSchema elementSchema;
```
zxcware: I don't see the necessity to make an extra `elementSchema` field, as the `elementConverter` has inner reference to it. It's applied to `EnumConverter`.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    /**
+     * Convert JsonElement to Parquet type
+     * @param value
+     * @return
+     */
+    abstract Object convertField(JsonElement value);
+
+    /**
```
zxcware: `buildSchema` and `saveSchema` are detail implementations about how the output schema is calcuated. We shouldn't include them in the base class.

Subclasses like `PrimitiveConverter` and `CollectionConverter` could just say `this.schema = buildSchema()` on intialization.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,484 @@
...
+            "Field: " + this.jsonSchema.getColumnName() + " is not nullable and contains a null value");
+      }
+      return convertField(value);
+    }
+
+    /**
+     * Returns a {@link Type} parquet schema
+     * @return
+     */
+    public Type schema() {
```
zxcware: This one is better to be an abstract method since `schema` can only be set in subclasses. Also remove the `schema` property from this class.

gobblin-core/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,484 @@
...
+      super(schema, repeated, BINARY, state);
+    }
+
+    @Override
+    BinaryValue convertField(JsonElement value) {
+      return new BinaryValue(Binary.fromString(value.getAsString()));
+    }
+
+    @Override
+    public Type schema() {
```
zxcware: This can be override `buildSchema()` instead of `schema()`

gobblin-core/build.gradle
```diff
@@ -57,6 +57,7 @@ dependencies {
   compile externalDependency.oltu
   compile externalDependency.opencsv
   compile externalDependency.hadoopHdfs
+  compile externalDependency.parquet
```
zxcware: We plan to decouple external extension from gobblin-core. Please make a new module, say `gobblin-parquet` in `gobblin-modules` and move the new classes there, similar to what we do to `gobblin-sql`.

In this way, an extension can be brought in on-demand instead of bundled with `gobblin-core` and brought in regardless of whether the application needs it or not.

gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,527 @@
...
+      super(schema, repeated, INT32, state);
+    }
+
+    @Override
+    IntegerValue convertField(JsonElement value) {
+      return new IntegerValue(value.getAsInt());
+    }
+
+    @Override
+    public Type schema() {
```
zxcware: Can be moved to `PrimitiveConverter`. The comment applied to the other primitive converters

gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,527 @@
...
+     */
+    public PrimitiveConverter(JsonSchema jsonSchema, boolean repeated, PrimitiveTypeName outputType,
+        WorkUnitState state) {
+      super(jsonSchema, state);
+      this.repeated = repeated;
+      this.outputType = outputType;
+      this.schema = buildSchema();
+    }
+
+    private Type buildSchema() {
```
zxcware: Make this `protected` so that it can be overridden in `StringConverter`?

gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,527 @@
...
+  }
+
+  /**
+   * Converts a JsonElement into a supported ParquetType
+   * @author tilakpatidar
+   *
+   */
+  public static abstract class JsonElementConverter {
+    protected final JsonSchema jsonSchema;
+    protected final WorkUnitState state;
```
zxcware: Is this used? If not, remove?

gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,527 @@
...
+    public Type schema() {
+      return this.schema;
+    }
+  }
+
+  public static class ArrayConverter extends CollectionConverter {
+
+    public ArrayConverter(JsonSchema arraySchema, WorkUnitState state) {
+      super(arraySchema, arraySchema.getElementTypeUsingKey(ARRAY_ITEMS_KEY), state, true);
+      this.schema = buildSchema();
```
zxcware: Can move this line and the function `schema()` to `CollectionConverter`. Subclasses could just override `buildSchema()`

gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java
```diff
@@ -0,0 +1,527 @@
...
+          return new GroupType(this.jsonSchema.optionalOrRequired(), docName, parquetTypes);
+        default:
+          throw new RuntimeException("Unsupported Record type");
+      }
+    }
+  }
+
+  public static class MapConverter extends CollectionConverter {
+
+    private final WorkUnitState state;
```
zxcware: Duplicate or can be removed?

abti:
Thanks @tilakpatidar and @zxcware

@tilakpatidar This seems to have code conflict, can you please resolve them so that I can merge? 
Also, do you want to reduce down your commits to smaller meaningful commits or else I can squash them down to one commit, let me know.

tilakpatidar:
@abti I have fixed the conflict. You can go ahead and squash them into one commit.

tilakpatidar:
@abti can you merge this if it looks okay

PR #2104, opened 2017-09-14
[GOBBLIN-253] Enhance Hive materializer.

ibuenros:
Enhancements:
- No assumption that source tables are Avro.
- Can materialize views and queries.
- Usage simplification.
- Added sample job.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-253

### Description

### Tests

### Commits


ibuenros:
@abti can you review?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * A simple {@link HiveTask} for Hive view materialization.
+ */
+public class HiveMaterializer extends HiveTask {
+
+  protected static final String STAGEABLE_TABLE_METADATA_KEY = "internal.hiveMaterializer.stageableTableMetadata";
+  protected static final String MATERIALIZER_MODE_KEY = "internal.hiveMaterializer.materializerMode";
+  protected static final String STORAGE_FORMAT_KEY = "internal.hiveMaterializer.storageFormat";
+  protected static final String QUERY_TO_MATERIALIZE_KEY = "internal.hiveMaterializer.queryToMaterialize";
```
autumnust: I believe by 'materialization of query' , we are actually meaning the materialization of query result. 
So probably it will be better to name it as 'queryResult_to_materialize_key', and all the other places related to 'materialization of query'

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
```diff
@@ -0,0 +1,155 @@
...
+   */
+  public static WorkUnit queryMaterializationWorkUnit(String query, HiveConverterUtils.StorageFormat storageFormat,
+      StageableTableMetadata destinationTable) {
+    WorkUnit workUnit = new WorkUnit();
+    workUnit.setProp(MATERIALIZER_MODE_KEY, MaterializerMode.QUERY_MATERIALIZATION.name());
+    workUnit.setProp(STORAGE_FORMAT_KEY, storageFormat.name());
+    workUnit.setProp(QUERY_TO_MATERIALIZE_KEY, query);
+    workUnit.setProp(STAGEABLE_TABLE_METADATA_KEY, HiveSource.GENERICS_AWARE_GSON.toJson(destinationTable));
+    TaskUtils.setTaskFactoryClass(workUnit, HiveMaterializerTaskFactory.class);
+    HiveTask.disableHiveWatermarker(workUnit);
```
autumnust: This disabling of watermark seems confusing here. 
In the `HiveMaterializerSource` where this method is actually invoked, there's also disabling there, and that disabling applies to all three kinds of materialization. 

Can you also add more docs explaining the necessity of disabling watermark?

gobblin-example/src/main/java/org/apache/gobblin/example/hivematerializer/HiveMaterializerSource.java
```diff
@@ -0,0 +1,129 @@
...
+      try (AutoReturnableObject<IMetaStoreClient> client = pool.getClient()) {
+        Table sourceTable = new Table(client.get().getTable(sourceDbAndTable.getDb(), sourceDbAndTable.getTable()));
+        return new HiveDataset(fs, pool, sourceTable, ConfigUtils.propertiesToConfig(state.getProperties()));
+      }
+    } catch (TException exc) {
+      throw new RuntimeException(exc);
+    }
+  }
+
+  private HiveConverterUtils.StorageFormat getOutputStorageFormat(State state) {
```
autumnust: This method can be moved to some other places like `HiveMaterializer` as a static method (Just like `HiveMaterializer.viewMaterializationWorkUnit`), as it can be reusable in other places like `LoopingHiveDatasetFinderSource`(Looping coordinator for hive dataset) as well.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
```diff
@@ -0,0 +1,155 @@
...
+   * Create a work unit to copy a source table to a target table using a staging table in between.
+   * @param dataset {@link HiveDataset} for the source table.
+   * @param destinationTable {@link StageableTableMetadata} specifying staging and target tables metadata.
+   */
+  public static HiveWorkUnit tableCopyWorkUnit(HiveDataset dataset, StageableTableMetadata destinationTable,
+      @Nullable String partitionName) {
+    HiveWorkUnit workUnit = new HiveWorkUnit(dataset);
+    workUnit.setProp(MATERIALIZER_MODE_KEY, MaterializerMode.TABLE_COPY.name());
+    workUnit.setProp(STAGEABLE_TABLE_METADATA_KEY, HiveSource.GENERICS_AWARE_GSON.toJson(destinationTable));
+    if (!Strings.isNullOrEmpty(partitionName)) {
```
autumnust: This partition-related branch doesn't seems to be consistent with the way a partition parameter is handle in `HiveWorkUnit(HiveDataset hiveDataset, Partition partition)`, which not only set partition name but other properties. 

The essential question here is, are there use cases when a `HiveProcessingEntity` is representing partitions and they are requested to be materialized ? If so, the construction of `HiveWorkUnit` should be modified.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
abti: +1 would want to understand this better

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/ReplaceTableStageableTableMetadata.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.gobblin.data.management.conversion.hive.entities;
+
+import org.apache.hadoop.hive.ql.metadata.Table;
+
+
+/**
+ * A {@link StageableTableMetadata} intended where the target table is the same as the reference table. Intended to
+ * replace the original table.
+ */
+public class ReplaceTableStageableTableMetadata extends TableLikeStageableTableMetadata {
```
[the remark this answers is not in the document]
ibuenros: This is intended for replacing the original files of a table with new files. So it uses the location and metadata of the reference table as the location of the output table.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
ibuenros: Watermarker needs to be disabled for query result materialization because there is no such thing as a modtime for a query. The other materializations don't disable it.

In the source it is disabled simply because it is an example source, and it simplifies usage (no state store required, views may not have a watermark, etc.)

gobblin-example/src/main/java/org/apache/gobblin/example/hivematerializer/HiveMaterializerSource.java
ibuenros: Not sure about that... the configuration key and default values might be different for different use cases, plus a lot of use cases would just hard code what output format they expect. Ultimately, if refactoring, it would just wrap `HiverConverterUtils.StorageFormat.valueOf`, which is simple enough to use by itself.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
ibuenros: It's an interesting comment. My thought is that this is a bit easier to use, you can do
`tableCopyWorkUnit(dataset, destinationTable, "part=part123")`
or
`tableCopyWorkUnit(dataset, destinationTable, partition.getName())`

The first one does not require doing a lookup in the Hive metastore, while if you have the partition object you can always use the second method.

The main motivation for doing it this way, though, is that it was done this way before the refactoring. I have no problem refactoring even more to make users pass a partition if necessary.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/materializer/HiveMaterializer.java
ibuenros: Fixed

ibuenros:
@autumnust addressed or replied to comments

PR #2107, opened 2017-09-16
[GOBBLIN-234] Add a ControlMessageInjector that generates metadata up…

htran1:
…date control messages

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-234

### Description
* Adds the `LiKafkaByteArrayMsgSchemaChangeInjector` control message injector to detect changes to the latest avro schema registered for the topic.
* This injector maintains a cache of the schema ids it has seen. When a record's schema id is not found in the cache then the latest schema is fetched from the schema registry to check whether it has changed. If the latest schema has changed then a `MetadataUpdateControlMessage` is injected to trigger the downstream constructs to update their schemas.
* A `FlushControlMessage` is also injected since the currently open avro files need to be closed and new files opened to receive records with the new schema.

### Tests

### Commits


htran1:
@ibuenros, please review.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/stream/LiKafkaByteArrayMsgSchemaChangeInjector.java
```diff
@@ -0,0 +1,131 @@
...
+        // latest schema changed, so inject a metadata update control message
+        if (!latestSchema.equals(this.globalMetadata.getSchema())) {
+          // update the metadata in this injector since the control message is only applied downstream
+          this.globalMetadata = GlobalMetadata.<S, S>builderWithInput(this.globalMetadata,
+              Optional.of(latestSchema)).build();
+
+          // inject a metadata update control message followed by a flush control message since the avro data needs to
+          // be flushed and a new file opened to write data in the new schema
+          return Lists.newArrayList(new MetadataUpdateControlMessage<S, byte[]>(this.globalMetadata),
+              FlushControlMessage.<byte[]>builder().flushReason("Metadata change")
```
ibuenros: Why do you have to send a flush? Can't the writer trigger the flush itself when it sees a metadata change?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/stream/LiKafkaByteArrayMsgSchemaChangeInjector.java
```diff
@@ -0,0 +1,131 @@
...
+
+import lombok.AccessLevel;
+import lombok.Getter;
+
+/**
+ * A {@link ControlMessageInjector} that detects changes in the latest schema and notifies downstream constructs by
+ * injecting a {@link MetadataUpdateControlMessage}.
+ * @param <S>
+ */
+public class LiKafkaByteArrayMsgSchemaChangeInjector<S> extends ControlMessageInjector<S, byte[]> {
```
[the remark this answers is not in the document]
htran1: Yes, this is the matching metadata change injector for messages converted by the LiKafkaByteArrayMsgToAvroConverter.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/stream/LiKafkaByteArrayMsgSchemaChangeInjector.java
htran1: Yes, that is another way of doing it. The CloseOnWriteWriterWrapper can handle the message. I thought about both ways and was originally going to do it that way, but then decided to use the existing flush handlers since that is more generic than having the logic branch in the control message handler.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/stream/LiKafkaByteArrayMsgSchemaChangeInjector.java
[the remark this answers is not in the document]
htran1: I'll remove this, but LiKafkaByteArrayMsgToAvroConverter will be left where it is since there might be users depending on it.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/stream/LiKafkaByteArrayMsgSchemaChangeInjector.java
htran1: The Avro case case requires the file to close since the file format does not support evolution, but there may be other formats that support evolution, so I'll move the logic to the writer to allow flexibility for this case.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -341,6 +341,9 @@
   public static final String DEFAULT_WRITER_FILE_PATH_TYPE = "default";
   public static final String SIMPLE_WRITER_DELIMITER = "simple.writer.delimiter";
   public static final String SIMPLE_WRITER_PREPEND_SIZE = "simple.writer.prepend.size";
+  public static final String WRITER_CLOSE_ON_METADATA_UPDATE = WRITER_PREFIX + ".closeOnMetadataUpdate";
```
ibuenros: Why is this in `ConfigurationKeys` if it is only used by one construct?

ibuenros:
+1

PR #2111, opened 2017-09-19
[GOBBLIN-259] Support writing Kafka messages to db/table file path

zxcware:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-259

### Description
- Add a new write file path type `DB_TABLE`, which writes records from an `Extract` to folder '<namespace>/<table>'. 
- A gobblin job which uses the `KafkSource` and `AvroHdfsDataWriter` will write records to 'dbName/tableName' with the following configurations
```
extract.namespace=dbName
extract.table.name=tableName
writer.file.path.type=db_table
```

### Tests
- `WriterUtilsTest#testGetWriterFilePath`

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
@@ -82,7 +82,6 @@ public class WorkUnit extends State {
 
     if (extract != null) {
       this.extract = extract;
-      super.addAll(extract);
     } else {
       this.extract = new Extract(null, null, null, null);
     }
@@ -303,6 +302,20 @@ public class WorkUnit extends State {
   @Deprecated
   public long getLowWaterMark() {
     return getPropAsLong(ConfigurationKeys.WORK_UNIT_LOW_WATER_MARK_KEY);
+  }
+
+  @Override
+  public boolean contains(String key) {
+    return super.contains(key) || this.extract.contains(key);
+  }
+
+  @Override
+  public String getProp(String key) {
+    String value = super.getProp(key);
+    if (value == null) {
+      value = this.extract.getProp(key);
+    }
+    return value;
   }
 
   /**
--- a/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
@@ -82,6 +82,7 @@ public class WorkUnit extends State {
 
     if (extract != null) {
       this.extract = extract;
+      super.addAll(extract);
     } else {
       this.extract = new Extract(null, null, null, null);
     }
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/WriterUtils.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/WriterUtils.java
@@ -57,12 +57,15 @@ public class WriterUtils {
 
   public static final Config NO_RETRY_CONFIG = ConfigFactory.empty();
 
-  /**
-   * TABLENAME should be used for jobs that pull from multiple tables/topics and intend to write the records
-   * in each table/topic to a separate folder. Otherwise, DEFAULT can be used.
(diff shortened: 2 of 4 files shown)


zxcware:
@yukuai518 @htran1 Please review. Thanks.

gobblin-utility/src/test/java/org/apache/gobblin/util/WriterUtilsTest.java
```diff
@@ -93,13 +93,24 @@ public void testGetDataPublisherFinalOutputDir() {
...
     state.setProp(ConfigurationKeys.WRITER_FILE_PATH + ".0", TEST_WRITER_FILE_PATH);
     Assert.assertEquals(WriterUtils.getWriterFilePath(state, 1, 1), TEST_WRITER_FILE_PATH);
+
+    state.removeProp(ConfigurationKeys.WRITER_FILE_PATH);
+
+    state.setProp(ConfigurationKeys.WRITER_FILE_PATH_TYPE, "tablename");
+    Assert.assertEquals(WriterUtils.getWriterFilePath(state, 0, 0), new Path(table));
+
+    state.setProp(ConfigurationKeys.WRITER_FILE_PATH_TYPE, "db_table");
+    Assert.assertEquals(WriterUtils.getWriterFilePath(state, 0, 0), new Path(namespace + "/" + table));
```
htran1: Use `Path.SEPARATOR` here since you are using `Path.SEPARATOR` when constructing the name.

gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
```diff
@@ -82,6 +82,7 @@ public WorkUnit(SourceState state, Extract extract) {
 
     if (extract != null) {
       this.extract = extract;
+      super.addAll(extract);
```
htran1: Why this change instead of leaving the code that propagates only the table name to the workunit? This may bloat up the state which is already large for gobblin-kafka. Also you would have to add this the constructors that don't call this one. Unless there is a reason for this, I would prefer to leave it the way it was to avoid unexpected side effects.

gobblin-utility/src/main/java/org/apache/gobblin/util/WriterUtils.java
```diff
@@ -170,6 +175,19 @@ private static WriterFilePathType getWriterFilePathType(State state) {
...
   /**
+   * Creates {@link Path} for case {@link WriterFilePathType#DB_TABLE} with configurations
+   * {@link ConfigurationKeys#EXTRACT_NAMESPACE_NAME_KEY} and {@link ConfigurationKeys#EXTRACT_TABLE_NAME_KEY}
+   * @param state
+   * @return a path
+   */
+  public static Path getDbTableWriterFilePath(State state) {
+    Preconditions.checkArgument(state.contains(ConfigurationKeys.EXTRACT_NAMESPACE_NAME_KEY));
+    Preconditions.checkArgument(state.contains(ConfigurationKeys.EXTRACT_TABLE_NAME_KEY));
+    return new Path(state.getProp(ConfigurationKeys.EXTRACT_NAMESPACE_NAME_KEY) + Path.SEPARATOR + state
```
yukuai518: What does EXTRACT_NAMESPACE_NAME_KEY mean in reality? Is it a DB name? or the DB name would be included in the topic.name? It's better to use specific name like EXTRACT_DB_NAME_KEY if possible because EXTRACT_NAMESPACE_NAME_KEY is something like "com.linkedin" or "gobblin.copy".

zxcware: EXTRACT_NAMESPACE_NAME_KEY defines a general domain of a table. It can be used as a DB name. A topic name is not necessary a fully qualified table name. A fully qualified namespace can have parents, like "com.linkedin.dbName". So we don't need to create a specific configuration key for DB, using the general one is good enough. 

To avoid confusing, I'll rename the file path type to be `NAMESPACE_TABLE`.

PR #2116, opened 2017-09-22
[GOBBLIN-264] Add a SharedResourceFactory for creating shared DataPub…

htran1:
…lishers

A DataPublisher can indicate whether it has the THREADSAFE capability.
If the DataPublisher is threadsafe then the DataPublisherFactory will return a shareable ResourceInstance.
If the DataPublisher is not threadsafe then an ImmediatelyInvalidResourceEntry is returned to ensure
that the broker won't return the same publisher for a second request.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-264

### Description
Adds the following.
* `ImmediatelyInvalidResourceEntry `- a ResourceInstance that is invalidated when getResource() is called. Immediate invalidation means it won't be returned as a response to a future get call.
* `Capabilities` - a construct can implement the `CapabilityAware` interface to support queries on the capabilities it supports.
* `Threadsafe` - this capability indicates that the `DataPublisher` is can be called concurrently in multiple threads.
* `Reusable` - this capability indicates that the `DataPublisher` publish APIs can be invoked multiple times.
* The `ParallelRunner` is made reusable by adding a method to change the FileSystem and a method to wait for submitted tasks to complete.

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/capability/Capability.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/capability/Capability.java
@@ -18,6 +18,8 @@ package org.apache.gobblin.capability;
 
 import org.apache.gobblin.annotation.Alpha;
 
+import lombok.Data;
+
 /**
  * Represents a set of functionality a job-creator can ask for. Examples could include
  * encryption, compression, partitioning...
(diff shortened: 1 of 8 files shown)


htran1:

gobblin-api/src/main/java/org/apache/gobblin/capability/Capability.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Represents a set of functionality a job-creator can ask for. Examples could include
+ * encryption, compression, partitioning...
+ *
+ * Each Capability has a name and then a set of associated configuration properties. An example is
+ * the encryption algorithm to use.
+ */
+@Alpha
+public class Capability {
```
ibuenros: Any reason not to just use `@Data`?

gobblin-api/src/main/java/org/apache/gobblin/publisher/DataPublisher.java
```diff
@@ -30,7 +33,16 @@
 /**
  * Defines how to publish data and its corresponding metadata. Can be used for either task level or job level publishing.
  */
-public abstract class DataPublisher implements Closeable {
+public abstract class DataPublisher implements Closeable, CapabilityAware {
+  /**
+   * Threadsafe capability.
+   */
+  public static final Capability THREADSAFE = new Capability("THREADSAFE", false);
```
ibuenros: Should this be part of `DataPublisher` or of a common set of capabilities?

gobblin-utility/src/main/java/org/apache/gobblin/broker/ImmediatelyInvalidResourceEntry.java
```diff
@@ -0,0 +1,56 @@
...
+    this.valid = true;
+  }
+
+  @Override
+  public T getResource() {
+    // mark the object as invalid before returning so that a new one will be created on the next
+    // request from the factory
+    this.valid = false;
+
+    return getTheResource();
```
ibuenros: can't you just do `super.getResource()` instead?

gobblin-api/src/main/java/org/apache/gobblin/publisher/DataPublisher.java
htran1: Moved `THREADSAFE` to the `Capability` class. Leaving REUSABLE here since that is specific to `DataPublisher`s.

gobblin-api/src/main/java/org/apache/gobblin/capability/Capability.java
htran1: This was copied from Eric's original PR. Yes, `@Data` generates the desired code.

gobblin-utility/src/main/java/org/apache/gobblin/broker/ImmediatelyInvalidResourceEntry.java
htran1: Yes, `super.getResource()` calls `getTheResource()`. Changed to call  `super.getResource()` in case the implementation changes.

ibuenros:
+1

PR #2119, opened 2017-09-26
[GOBBLIN-267] Changed workunit creation policy to compare update time with maxLookBackDays

aditya1105:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-267

### Description
HiveSource#shouldCreateWorkunit() now compares update time with maxLookBackDays

### Tests
Very trivial change/fix

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
@@ -386,10 +386,13 @@ public class HiveSource implements Source {
 
   /**
    * Check if workunit needs to be created. Returns <code>true</code> If the
-   * <code>updateTime</code> is greater than the <code>lowWatermark</code>.
+   * <code>updateTime</code> is greater than the <code>lowWatermark</code> and <code>maxLookBackTime</code>
    * <code>createTime</code> is not used. It exists for backward compatibility
    */
   protected boolean shouldCreateWorkunit(long createTime, long updateTime, LongWatermark lowWatermark) {
+    if (new DateTime(updateTime).isBefore(this.maxLookBackTime)) {
+      return false;
+    }
     return new DateTime(updateTime).isAfter(lowWatermark.getValue());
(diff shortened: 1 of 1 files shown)


aditya1105:
@abti, can you please review this pr

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -386,10 +386,13 @@ protected boolean shouldCreateWorkunit(Partition sourcePartition, LongWatermark
 
   /**
    * Check if workunit needs to be created. Returns <code>true</code> If the
-   * <code>updateTime</code> is greater than the <code>lowWatermark</code>.
+   * <code>updateTime</code> is greater than the <code>lowWatermark</code> and <code>maxLookBackTime</code>
    * <code>createTime</code> is not used. It exists for backward compatibility
    */
   protected boolean shouldCreateWorkunit(long createTime, long updateTime, LongWatermark lowWatermark) {
+    if (new DateTime(updateTime).isBefore(this.maxLookBackTime)) {
```
erwa: Is `updateTime` here the same as the partition value (i.e. the time corresponding to the partition value YYYY-MM-DD-HH)?  If this is the last time the partition was updated, then I think you might still have a problem. For example, imagine 100 days of partitions are registered today and maxLookBackTime is 5 days ago. The last time all 100 partitions were updated would be > maxLookBackTime and would thus still have workunits created for them.

aditya1105: Update time is controllable via policies.Following are the policies we have:
1) DatePatternUpdateProvider
2) HdfsBasedUpdateProvider
3) HiveMetastoreBasedUpdateProvider

Hence using the 1 and 2 as update policies will make sure workunits won't be created in the scenario mentioned above.
If 3 is used, then both create time and update time should be taken by hive partition property and hence workunits must be created.

PR #2124, opened 2017-10-02
[GOBBLIN-272] Remove hardwired avro extensions from compaction and make it generic

treff7es:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-272

### Description
Avro extension was hardcoded in some place at Compaction and I just made it configurable to be more generic.
You can set it in your compaction config by:
`
compaction.extension=gz
`

It seems like the code style on these file was not the Gobblin style and that's why the change seems to be huge (or am I using the wrong codestyle?).

### Tests

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
@@ -17,8 +17,22 @@
 
 package org.apache.gobblin.compaction.action;
 
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.mapreduce.Counter;
+import org.apache.hadoop.mapreduce.Job;
+
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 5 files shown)


ibuenros:
@treff7es this only handles extensions other than avro right? It seems that it would always use the AvroInputFormat either way?

treff7es:
@ibuenros If you meant that currently compaction only works with avro then you are right or maybe I misunderstood you :). On it's own it does not make much sense generalise if compaction only works with avro.
Why I did this because if somebody want to create compaction job which can work with other formats then he/she should need to take care override all of these implementations.
Like here we are compressing gzip compressed json logs (here gzip is hardwired which I wold like to remove and generalize later) : https://github.com/prezi/gobblin/tree/json_module/gobblin-modules/gobblin-json/src/main/java/org/apache/gobblin/mapreduce .
Did I get right what you were saying? Is it the way I should do it?

treff7es:
@yukuai518 @ibuenros I see that now this is conflicting with the master and I'm happy to fix it if you see the value in it. I know @ibuenros had some comment and I'm not sure if you were ok with my reply. 
What do you think? Should I resolve the conflict?

abti:
@ibuenros ^^

abti:
ping @ibuenros ^^

abti:
@ibuenros ^^

ibuenros:
@treff7es comment makes sense to me. I am ok with the general idea of the changes, although I don't remember the details of the change. If someone reviews it I have no problem with this being committed.

PR #2129, opened 2017-10-04
Change setActive order to prevent flow spec loss

yukuai518:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-276

### Description
1. Originally isActive=true is set after onAddSpec was invoked during the leadership change. However this has a problem because onAddSpec will forward a spec to new leader when isActive==false, but actually current node is already the new leader.
2. Add  a countDownLatch variable to protect the topology being initialized before we load flow specs.
3. Add synchronization for scheduleJob because multiple threads can process the same flowSpec and gobblin based JobScheduler is not threadsafe.
4. Modify the order of topologyCatalog initialization. Move the topologyCatalog addListener after topologyFactory has updated the spec store. This will help prevent the incompatible topology capabilities "any:any" to be load which will break the Multihop compilation because weightedgraph today doesn't support a loop-back node.

### Tests
  Rerun gobblin-service module tests. No new tests being added.

### Commits


yukuai518:
@htran1 Can you review?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -138,7 +138,7 @@ public SpecCompiler getSpecCompiler() {
   /** {@inheritDoc} */
   @Override
   public void onAddSpec(Spec addedSpec) {
-    _log.info("New Spec detected: " + addedSpec);
+    _log.info("New Topology Spec detected: " + addedSpec);
```
htran1: This can be removed. The logging is below in the if block.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -178,6 +178,16 @@ public void onUpdateSpec(Spec updatedSpec) {
   }
 
   public void orchestrate(Spec spec) throws Exception {
+    // Add below waiting because TopologyCatalog and FlowCatalog service can be launched at the same time
```
htran1: Add a note here to remove this wait when sequencing of service startup is introduced.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -103,6 +103,10 @@ public synchronized void setActive(boolean isActive) {
       return;
     }
 
+    // Need to set active first; otherwise in the STANDBY->ACTIVE transition,
+    // the onAddSpec will forward specs to the leader, which is itself.
+    this.isActive = isActive;
```
htran1: What happens in the ACTIVE -> STANDBY transition. If this.isActive is set to false here then won't the onDeleteSpec() below get forwarded instead of resulting in cleanup in the current process?

abti:
Waiting for Travis to finish, will merge after that.

PR #2131, opened 2017-10-06
[GOBBLIN-278] Fix sending lineage event for KafkaSource

zxcware:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-278

### Description
  - Fix lineage event for KafkaSource not send, and void resending the events by removing configurations with key prefix `gobblin.lineage` from the state
  - Fix `KafkaWorkUnitPacker` disregards existing configurations of work units

### Tests
  - `State#testRemovePropsByPrefix`

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
@@ -20,6 +20,7 @@ package org.apache.gobblin.configuration;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -496,6 +497,30 @@ public class State implements WritableShim {
       commonPropsCopy.putAll(this.commonProperties);
       commonPropsCopy.remove(key);
       this.commonProperties = commonPropsCopy;
+    }
+  }
+
+  /**
+   * Remove all properties with a certain keyPrefix
+   *
+   * @param keyPrefix key prefix
+   */
+  public void removeProps(String keyPrefix) {
+    this.specProperties.entrySet().removeIf(entry -> ((String) entry.getKey()).startsWith(keyPrefix));
+
+    Properties newCommonProperties = null;
+    for (Object key: this.commonProperties.keySet()) {
+      if (((String)key).startsWith(keyPrefix)) {
+        if (newCommonProperties == null) {
+          newCommonProperties = new Properties();
+          newCommonProperties.putAll(this.commonProperties);
+        }
+        newCommonProperties.remove(key);
+      }
+    }
+
+    if (newCommonProperties != null) {
+      this.commonProperties = newCommonProperties;
     }
   }
 
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
@@ -503,14 +503,14 @@ public class State implements WritableShim {
   /**
    * Remove all properties with a certain keyPrefix
    *
-   * @param keyPrefix key prefix
+   * @param prefix key prefix
    */
-  public void removeProps(String keyPrefix) {
-    this.specProperties.entrySet().removeIf(entry -> ((String) entry.getKey()).startsWith(keyPrefix));
+  public void removePropsWithPrefix(String prefix) {
+    this.specProperties.entrySet().removeIf(entry -> ((String) entry.getKey()).startsWith(prefix));
 
     Properties newCommonProperties = null;
     for (Object key: this.commonProperties.keySet()) {
-      if (((String)key).startsWith(keyPrefix)) {
+      if (((String)key).startsWith(prefix)) {
         if (newCommonProperties == null) {
           newCommonProperties = new Properties();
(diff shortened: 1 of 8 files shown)


zxcware:
@yukuai518 Please review

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/workunit/packer/KafkaWorkUnitPacker.java
```diff
@@ -81,25 +81,11 @@
...
   protected final SourceState state;
-  protected final Extract.TableType tableType;
-  protected final String extractNameSpace;
-  protected final boolean isFullExtract;
   protected final KafkaWorkUnitSizeEstimator sizeEstimator;
 
   protected KafkaWorkUnitPacker(AbstractSource<?, ?> source, SourceState state) {
     this.source = source;
     this.state = state;
-    if (state.getPropAsBoolean(KafkaSource.GOBBLIN_KAFKA_EXTRACT_ALLOW_TABLE_TYPE_NAMESPACE_CUSTOMIZATION)) {
```
yukuai518: Why this is not needed any more?

gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
```diff
@@ -500,6 +501,30 @@ public void removeProp(String key) {
   }
 
   /**
+   * Remove all properties with a certain keyPrefix
+   *
+   * @param keyPrefix key prefix
+   */
+  public void removeProps(String keyPrefix) {
```
htran1: Maybe rename this to `removePropsWithPrefix`? The current name is too easy to mess up as a typo.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/workunit/packer/KafkaWorkUnitPacker.java
```diff
@@ -227,13 +213,18 @@ protected WorkUnit squeezeMultiWorkUnit(MultiWorkUnit multiWorkUnit) {
     List<KafkaPartition> partitions = getPartitionsFromMultiWorkUnit(multiWorkUnit);
     Preconditions.checkArgument(!partitions.isEmpty(), "There must be at least one partition in the multiWorkUnit");
 
-    Extract extract = this.source.createExtract(tableType, extractNameSpace, partitions.get(0).getTopicName());
-    if (isFullExtract) {
-      extract.setProp(ConfigurationKeys.EXTRACT_IS_FULL_KEY, true);
-    }
-    WorkUnit workUnit = WorkUnit.create(extract, interval);
+    // Squeeze all partitions from the multiWorkUnit into of one the work units, which can be any one
+    WorkUnit workUnit = multiWorkUnit.getWorkUnits().get(0);
```
htran1: Is this to preserve properties from the workunit by using an existing workunit instead of creating a new one? Will removing the properties below affect other places that access this workunit?

Would it be safer and sufficient to create a new workunit and copy the properties over from the first work unit?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -538,17 +540,18 @@ private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, Offsets
     }
 
     WorkUnit workUnit = WorkUnit.create(extract);
-    if (topicSpecificState.isPresent()) {
```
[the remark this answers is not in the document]
zxcware: This is duplicate logic to line 239

gobblin-api/src/main/java/org/apache/gobblin/configuration/State.java
zxcware: Good point.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/workunit/packer/KafkaWorkUnitPacker.java
zxcware: > Is this to preserve properties from the workunit by using an existing workunit instead of creating a new one? 

Yes.

> Will removing the properties below affect other places that access this workunit?

No. Because these properties are either ignored or replaced by the configurations immediately set below.

> Would it be safer and sufficient to create a new workunit and copy the properties over from the first work unit?

No. We do need to remove those properties. Otherwise, the logic won't be the same as before. And all work units of in the same multi work unit have the same properties but those that are removed and extractid.

htran1: I meant copy then remove from the copy. But if it is safe then you can remove directly.

zxcware:
@htran1 @yukuai518 Comments addressed, please review.

PR #2132, opened 2017-10-09
[GOBBLIN-279] pull file unable to reuse the json property.

dallaybatta:
Dear Gobblin maintainers,

### GOBBLIN-279
- [ ] My PR addresses the following [GOBBLIN-279](https://issues.apache.org/jira/browse/GOBBLIN-279) issues and references them in the PR title. 

### Description

### Tests
- [ ] My PR adds the following unit tests org.apache.gobblin.util.PullFileLoaderTest.testJsonPropertyReuseJobLoading

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/PullFileLoader.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/PullFileLoader.java
@@ -72,6 +72,9 @@ public class PullFileLoader {
   public static final Set<String> DEFAULT_JAVA_PROPS_PULL_FILE_EXTENSIONS = Sets.newHashSet("pull", "job");
   public static final Set<String> DEFAULT_HOCON_PULL_FILE_EXTENSIONS = Sets.newHashSet("json", "conf");
 
+  public static final String PROPERTY_DELIMITER_PARSING_ENABLED_KEY = "property.parsing.enablekey";
+  public static final boolean DEFAULT_PROPERTY_DELIMITER_PARSING_ENABLED_KEY = false;
+  
   private final Path rootDirectory;
   private final FileSystem fs;
   private final ExtensionFilter javaPropsPullFileFilter;
@@ -260,6 +263,7 @@ public class PullFileLoader {
     PropertiesConfiguration propertiesConfiguration = new PropertiesConfiguration();
     try (InputStreamReader inputStreamReader = new InputStreamReader(this.fs.open(propertiesPath),
         Charsets.UTF_8)) {
+      propertiesConfiguration.setDelimiterParsingDisabled(ConfigUtils.getBoolean(fallback,PROPERTY_DELIMITER_PARSING_ENABLED_KEY, DEFAULT_PROPERTY_DELIMITER_PARSING_ENABLED_KEY));
       propertiesConfiguration.load(inputStreamReader);
 
       Config configFromProps =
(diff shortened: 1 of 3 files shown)


gobblin-utility/src/test/java/org/apache/gobblin/util/PullFileLoaderTest.java
```diff
@@ -218,7 +219,20 @@ public void testRecursiveJobLoadingWithAndGlobalProps() throws Exception {
...
+  @Test
+  public void testJsonPropertyReuseJobLoading() throws Exception {
+    Path path;
+    Config pullFile;
+    path = new Path(this.basePath, "bjob.pull");
+    Config cfg = ConfigFactory.parseMap(ImmutableMap.<String, Object>builder()
+            .put(PullFileLoader.PROPERTY_DELIMITER_PARSING_ENABLED_KEY, true)
+            .build());
+    pullFile = loader.loadPullFile(path, cfg, false);
+    System.out.println(pullFile.getString("json.property.key"));
```
zxcware: remove line 231 and 232

gobblin-utility/src/main/java/org/apache/gobblin/util/PullFileLoader.java
```diff
@@ -260,6 +263,8 @@ private Config loadJavaPropsWithFallback(Path propertiesPath, Config fallback) t
     PropertiesConfiguration propertiesConfiguration = new PropertiesConfiguration();
     try (InputStreamReader inputStreamReader = new InputStreamReader(this.fs.open(propertiesPath),
         Charsets.UTF_8)) {
+      propertiesConfiguration.setDelimiterParsingDisabled(ConfigUtils.getBoolean(fallback,
```
zxcware: This disables the parsing for all configuration keys. What about a pull file has some key that needs parsing and another key that doesn't need?

gobblin-utility/src/test/java/org/apache/gobblin/util/PullFileLoaderTest.java
dallaybatta: I have seen System.out.printlns in many of the Test cases so though it would be okay to keep, I am going to remove it.

gobblin-utility/src/main/java/org/apache/gobblin/util/PullFileLoader.java
dallaybatta: I don't see an API in Apache Commons to make granulation calls on delimitation. Do you think we should address such use case?

gobblin-utility/src/test/java/org/apache/gobblin/util/PullFileLoaderTest.java
zxcware: Disable or enable, I think one is of equivalent possibility in a use case to the other. I saw in your test case that it can help with json object. If that's the motivation, could you encode a json object into a string as a configuration and parse it when used at runtime?

dallaybatta: I am sorry I could not understand what you meant by the initial statement.
Yes the encoding in configuration and decoding in runtime will work but I wanted to avoid more changes in the runtime code ( i.e decoding of json).

abti:
@zxcware Are your comments addressed? Please let me know when this is ready to be merged in. (I haven't personally reviewed the PR)

gobblin-utility/src/test/java/org/apache/gobblin/util/PullFileLoaderTest.java
zxcware: My concern is, if the delimiter is disabled for all configurations, some configuration, whose value has the delimiter and expects to be parsed as a list, will break. Is it an issue to you?

dallaybatta: No it is not going to be an issue for our use case.

PR #2137, opened 2017-10-11
[GOBBLIN-284] Add retry in SalesforceExtractor to handle transient ne…

htran1:
…twork errors

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-284

### Description
* Add retry to handle transient network errors. The BufferedReader is recreated to reconnect on before reattempting fetch. The Salesforce http connection is sometimes dropped from the server side, so this retry is required to avoid job failure.
* Skip empty result sets since that can cause the job to terminate before pulling all data since returning a null record terminates the pull.

### Tests
Tested with a read pull of the Contact table.

### Commits

### DIFF
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
@@ -594,11 +594,9 @@ public class SalesforceExtractor extends RestApiExtractor {
       // Get data from input stream
       // If bulk load is not finished, get data from the stream
       // Skip empty result sets since they will cause the extractor to terminate early
-      do {
-        if (!this.isBulkJobFinished()) {
-          rs = getBulkData();
-        }
-      } while (rs != null && rs.isEmpty() && !this.isBulkJobFinished());
+      while (!this.isBulkJobFinished() && (rs == null || rs.isEmpty())) {
+        rs = getBulkData();
+      }
 
       // Set bulkApiInitialRun to false after the completion of first run
       this.bulkApiInitialRun = false;
@@ -802,9 +800,12 @@ public class SalesforceExtractor extends RestApiExtractor {
   }
 
   /**
-   * Fetch a batch of records
+   * Fetch records into a {@link RecordSetList} up to the configured batch size {@link #batchSize}. This batch is not
+   * the entire Salesforce result batch. It is an internal batch in the extractor for buffering a subset of the result
+   * stream that comes from a Salesforce batch for more efficient processing.
    * @param rs the record set to fetch into
-   * @param initialRecordCount Initial record count to use. This should correspond to the number of records already in rs
+   * @param initialRecordCount Initial record count to use. This should correspond to the number of records already in rs.
+   *                           This is used to limit the number of records returned in rs to {@link #batchSize}.
    * @throws DataRecordException
    * @throws IOException
    */
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
@@ -108,6 +108,8 @@ public class SalesforceExtractor extends RestApiExtractor {
   private static final int MAX_RETRY_INTERVAL_SECS = 600;
   // avoid using too many bulk API calls by only allowing PK chunking only if max partitions is configured <= this
   private static final int PK_CHUNKING_MAX_PARTITIONS_LIMIT = 3;
+  private static final String FETCH_RETRY_LIMIT_KEY = "salesforce.fetchRetryLimit";
+  private static final int DEFAULT_FETCH_RETRY_LIMIT = 5;
 
   private boolean pullStatus = true;
   private String nextUrl;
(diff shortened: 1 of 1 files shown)


htran1:
@zxcware please review.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -775,6 +790,122 @@ public boolean bulkApiLogin() throws Exception {
...
+
+    // if the result set is partially processed then we need to skip over processed records
+    if (!isNewBulkResultSet()) {
+      List<String> lastCsvRecord = null;
+      InputStreamCSVReader reader = new InputStreamCSVReader(this.bulkBufferedReader);
+
+      // skip header
+      reader.nextRecord();
+
+      int recordsToSkip = this.bulkRecordCount - this.prevBulkRecordCount;
```
zxcware: If a batch is processed one after the other, it's not necessary to skip but clearing the result set?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -581,9 +593,12 @@ public String getTimestampPredicateCondition(String column, long value, String v
 
       // Get data from input stream
       // If bulk load is not finished, get data from the stream
-      if (!this.isBulkJobFinished()) {
-        rs = getBulkData();
-      }
+      // Skip empty result sets since they will cause the extractor to terminate early
```
zxcware: Is it better to do the skip loop in `getBulkData`

htran1: `getBulkData` has a couple places where it returns, so I thought that this was cleaner and less error prone.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -775,6 +790,122 @@ public boolean bulkApiLogin() throws Exception {
...
+    // Get record from CSV reader stream
+    while ((this.csvRecord = reader.nextRecord()) != null) {
+      // Convert CSV record to JsonObject
+      JsonObject jsonObject = Utils.csvToJsonObject(this.bulkRecordHeader, this.csvRecord, this.bulkResultColumCount);
+      rs.add(jsonObject);
+      recordCount++;
+      this.bulkRecordCount++;
+
+      // Insert records in record set until it reaches the batch size
+      if (recordCount >= batchSize) {
```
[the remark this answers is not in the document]
htran1: This is the existing logic moved into a new method. The check here is required to chunk the stream into batches since the while loop is only checking for end of stream. This check could be moved into the while, but I think it is here for the extra logging when the condition is true.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
htran1: No, there are two reasons for this.
1. Clearing and recreating records would be more expensive than skipping.
2. The reset moves to the beginning of the result set. There can be many 2K record batches in a result set and all except the last batch have been processed and given out of the extractor, so we need to skip over at least all batches prior to the last one.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -581,9 +593,12 @@ public String getTimestampPredicateCondition(String column, long value, String v
...
       // If bulk load is not finished, get data from the stream
-      if (!this.isBulkJobFinished()) {
-        rs = getBulkData();
-      }
+      // Skip empty result sets since they will cause the extractor to terminate early
+      do {
+        if (!this.isBulkJobFinished()) {
+          rs = getBulkData();
+        }
+      } while (rs != null && rs.isEmpty() && !this.isBulkJobFinished());
```
zxcware:
```
while ((rs == null || rs.isEmpty()) && !this.isBulkJobFinished()) {
  rs = getBulkData();
}
```

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -775,6 +790,122 @@ public boolean bulkApiLogin() throws Exception {
...
+   * @throws AsyncApiException
+   */
+  private BufferedReader getBulkBufferedReader(int index) throws AsyncApiException {
+    return new BufferedReader(new InputStreamReader(
+        this.bulkConnection.getQueryResultStream(this.bulkJob.getId(), this.bulkResultIdList.get(index).getBatchId(),
+            this.bulkResultIdList.get(index).getResultId()), ConfigurationKeys.DEFAULT_CHARSET_ENCODING));
+  }
+
+  /**
+   * Fetch a batch of records
```
zxcware: It's better to make it clearer that this batch isn't the chunk batch, but the extractor is grouping the records

PR #2140, opened 2017-10-13
[GOBBLIN-288] Add finer-grain dynamic partition generation for Salesf…

htran1:
…orce

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-288

### Description
Added refinement of the daily bucket histogram by splitting large buckets into smaller time ranges until a target size is reached. This is to avoid timeout errors when a partition is too large.

### Tests
Updated partitioning test and ran Salesforce Contact flow.

### Commits


htran1:
@zxcware please review.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -252,23 +462,23 @@ private String getDateString(int year) {
...
-      throw new DataRecordException("Failed to get data from salesforce; REST response has no output");
-    }
-    JsonArray records = GSON.fromJson(output, JsonObject.class).getAsJsonArray("records");
     Iterator<JsonElement> elements = records.iterator();
     JsonObject element;
+
     while (elements.hasNext()) {
       element = elements.next().getAsJsonObject();
-      histogram.add(new HistogramGroup(element.get("time").getAsString(), element.get("cnt").getAsInt()));
+      String time = element.get("time").getAsString() + ZERO_TIME_SUFFIX;
```
zxcware: In terms of the function itself, whether to append the `ZERO_TIME_SUFFIX` really depends on the `element.get("time")`. My concern is that the function won't be correct if `time` already has a proper suffix.

The solution would be:
1. Leverage `Utils.toDateTimeFormat`, given the input time format
2. Remove the function and put the logic to `getHistogramByDayBucketing`, for it's only used there.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -182,7 +208,183 @@ String generateSpecifiedPartitions(Histogram histogram, int maxPartitions, long
...
+
+  /**
+   * Get the row count for a time range
+   */
+  private int getCountForRange(TableCountProbingContext probingContext, StrSubstitutor sub,
+      Map<String, String> subValues, long startTime, long endTime) {
+    String startTimeStr = Utils.dateToString(new Date(startTime), SalesforceExtractor.SALESFORCE_TIMESTAMP_FORMAT);
+    String endTimeStr = Utils.dateToString(new Date(endTime), SalesforceExtractor.SALESFORCE_TIMESTAMP_FORMAT);
+
+    subValues.put("start", startTimeStr);
```
zxcware: It's necessary to consider whether we should include the `start` by comparing it with the global partition, which can be put in the `probingContext`

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -163,6 +182,13 @@ String generateSpecifiedPartitions(Histogram histogram, int maxPartitions, long
       }
     }
 
+    if (partitionPoints.isEmpty()) {
+      throw new RuntimeException("Unexpected empty partition list");
+    }
+
+    // exchange the first partition point with the global low watermark
+    partitionPoints.set(0, Long.toString(lowWatermark));
```
zxcware: Instead of doing this, we can set the first `HistogramGroup` key to be lowWatermak, which also helps with fine probing.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -78,9 +79,23 @@
...
-  private static final String DAY_PARTITION_QUERY_TEMPLATE = "SELECT count(${column}) cnt, DAY_ONLY(${column}) time FROM ${table} "
-      + "WHERE ${column} ${greater} ${start} AND ${column} ${less} ${end} GROUP BY DAY_ONLY(${column}) ORDER BY DAY_ONLY(${column})";
-  private static final String DAY_FORMAT = "yyyy-MM-dd";
+  private static final String ENABLE_DYNAMIC_PROBING = "salesforce.enableDynamicProbing";
+  private static final String DYNAMIC_PROBING_LIMIT = "salesforce.dynamicProbingLimit";
+  private static final int DEFAULT_DYNAMIC_PROBING_LIMIT = 1000;
+  private static final String MIN_TARGET_PARTITION_SIZE = "salesforce.minTargetPartitionSize";
+  private static final int DEFAULT_MIN_TARGET_PARTITION_SIZE = 250000;
+  private static final String PROBE_TARGET_RATIO = "salesforce.probeTargetRatio";
+  private static final double DEFAULT_PROBE_TARGET_RATIO = 0.60;
```
zxcware: This is a great idea!  We should document it so that readers can understand its meaning.

PR #2143, opened 2017-10-19
[GOBBLIN-291] Remove unnecessary spec list and read

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @htran1 @abti Please review.

### JIRA

### Description
This PR address issue of unnecessary listing and reading of flow specs from file system, which can cause huge delay in rest calls response if  number of specs are large.

### Tests
This PR does not add any new functionality. All the flow specs related test cases are already implemented.

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonRecordAvroSchemaToAvroConverter.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonRecordAvroSchemaToAvroConverter.java
@@ -31,7 +31,7 @@ import org.apache.gobblin.converter.ToAvroConverterBase;
 
 import com.google.common.base.Preconditions;
 import com.google.gson.JsonObject;
-import com.google.common.base.Preconditions;
+import com.google.gson.JsonNull;
 import com.google.common.base.Splitter;
 
 
@@ -84,10 +84,6 @@ public class JsonRecordAvroSchemaToAvroConverter<SI> extends ToAvroConverterBase
         continue;
       }
 
-      if (inputRecord.get(field.name()) == null) {
-        throw new DataConversionException("Field missing from record: " + field.name());
-      }
-
       Schema.Type type = field.schema().getType();
       boolean nullable = false;
       Schema schema = field.schema();
@@ -107,6 +103,14 @@ public class JsonRecordAvroSchemaToAvroConverter<SI> extends ToAvroConverterBase
         } else {
           throw new DataConversionException("Unions must be size 2, and contain one null");
         }
+
+        if (inputRecord.get(field.name()) == null) {
+          inputRecord.add(field.name(), JsonNull.INSTANCE);
+        }
+      }
+
+      if (inputRecord.get(field.name()) == null) {
+        throw new DataConversionException("Field missing from record: " + field.name());
       }
 
       if (type.equals(Schema.Type.RECORD)) {


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -98,17 +98,39 @@ public FSSpecStore(Config sysConfig, SpecSerDe specSerDe, Optional<Logger> log)
...
+    Path flowPath = new Path(specUri.getPath());
+    String flowGroup = getFlowGroup(flowPath);
+    String flowName = getFlowName(flowPath);
     FileStatus[] fileStatuses = fs.listStatus(this.fsSpecStoreDirPath);
+
     for (FileStatus fileStatus : fileStatuses) {
-      if (StringUtils.startsWith(fileStatus.getPath().getName(), specUri.toString())) {
+      if (fileStatus.isDirectory() && fileStatus.getPath().getName().equals(flowGroup)) {
+        return exists(fileStatus.getPath(), flowName);
+      } else if(!fileStatus.isDirectory() && fileStatus.getPath().getName().startsWith(flowName)) {
```
htran1: There should always be a flow group, so the else case should never apply.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -98,17 +98,39 @@ public FSSpecStore(Config sysConfig, SpecSerDe specSerDe, Optional<Logger> log)
...
         return true;
       }
     }
+    return false;
+  }
 
+  private boolean exists(Path flowGroupDir, String flowName) throws IOException {
+    FileStatus[] fileStatuses = fs.listStatus(flowGroupDir);
+    for (FileStatus fileStatus : fileStatuses) {
+      if (!fileStatus.isDirectory() && fileStatus.getPath().getName().startsWith(flowName)) {
```
htran1: I think we need to add a delimiter to the name between the name and the version. Today if you have a file abc111 we can't tell if it is version 1 of abc11 or version 11 of abc1.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -98,17 +98,39 @@ public FSSpecStore(Config sysConfig, SpecSerDe specSerDe, Optional<Logger> log)
...
+    return Files.getNameWithoutExtension(specUri.getName());
+  }
+
   @Override
   public boolean exists(URI specUri) throws IOException {
     Preconditions.checkArgument(null != specUri, "Spec URI should not be null");
-
+    Path flowPath = new Path(specUri.getPath());
+    String flowGroup = getFlowGroup(flowPath);
+    String flowName = getFlowName(flowPath);
```
htran1: We should be able to check for the existence of <store_root>/<flowGroup> first then list that directory to avoid listing from the store_root. You can directly call listStatus on <store_root>/<flowGroup> and catch the FileNotFoundException and return false.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -206,23 +228,24 @@ public Spec getSpec(URI specUri, String version) throws IOException, SpecNotFoun
...
-    Preconditions.checkArgument(null != specUri, "Spec URI should not be null");
-
-    Collection<Spec> specs = getSpecs();
-    Collection<Spec> filteredSpecs = Lists.newArrayList();
-    for (Spec spec : specs) {
-      if (spec.getUri().equals(specUri)) {
-        filteredSpecs.add(spec);
+  private Collection<Spec> getAllVersionsOfSpec(String flowGroup, String flowName) throws IOException {
+    Collection<Spec> specs = Lists.newArrayList();
+    FileStatus[] fileStatuses = fs.listStatus(this.fsSpecStoreDirPath);
```
htran1: Same comment as above. We can list directly the flowGroup directory.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
arjun4084346: I started with that only; but realized that TopologyCatalog also uses the same methods.
https://github.com/apache/incubator-gobblin/blob/master/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/TopologyCatalog.java#L257
and topologies do not have group directory.
Maybe we should implement separate method for TopologyCatalog.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
arjun4084346: Will do.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
arjun4084346: Yes, I wanted to discuss this too. But the problem is we have already written 'abc111' in the file system. We need to convert all the materialized specs to new pattern.

PR #2147, opened 2017-10-24
[GOBBLIN-291] Remove unnecessary spec list and read

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @htran1 please review

### JIRA

### Description
This PR address issue of unnecessary listing and reading of flow specs from file system, which can cause huge delay in rest calls response if number of specs are large.

### Tests
This PR does not add any new functionality. All the flow specs related test cases are already implemented.

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/MutableSpecCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/MutableSpecCatalog.java
@@ -34,7 +34,8 @@ public interface MutableSpecCatalog extends SpecCatalog {
   public void put(Spec spec);
 
   /**
-   * Removes an existing {@link Spec} with the given URI. A no-op if such {@link Spec} does not exist.
+   * Removes an existing {@link Spec} with the given URI.
+   * Throws SpecNotFoundException if such {@link Spec} does not exist.
    */
-  void remove(URI uri);
+  void remove(URI uri) throws SpecNotFoundException;
 }
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
@@ -18,6 +18,8 @@
 package org.apache.gobblin.runtime.spec_store;
 
 import com.google.common.io.ByteStreams;
+
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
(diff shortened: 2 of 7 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -206,23 +232,32 @@ public Spec getSpec(URI specUri, String version) throws IOException, SpecNotFoun
...
-    Collection<Spec> specs = getSpecs();
-    Collection<Spec> filteredSpecs = Lists.newArrayList();
-    for (Spec spec : specs) {
-      if (spec.getUri().equals(specUri)) {
-        filteredSpecs.add(spec);
+  private Collection<Spec> getAllVersionsOfSpec(String specGroup, String specName) throws IOException {
+    Collection<Spec> specs = Lists.newArrayList();
+    FileStatus[] fileStatuses;
+    try {
+      if (StringUtils.isEmpty(specGroup)) {
```
htran1: Can you create a private method that returns a path based on the spec group and root dir? Then the listStatus does not have to be duplicated. There is another use like this above.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/FSSpecStore.java
```diff
@@ -98,17 +101,40 @@ public FSSpecStore(Config sysConfig, SpecSerDe specSerDe, Optional<Logger> log)
...
+        fileStatuses = fs.listStatus(this.fsSpecStoreDirPath);
+      } else {
+        fileStatuses = fs.listStatus(new Path(this.fsSpecStoreDirPath, specGroup));
+      }
+    } catch (FileNotFoundException e) {
+      return false;
+    }
     for (FileStatus fileStatus : fileStatuses) {
-      if (StringUtils.startsWith(fileStatus.getPath().getName(), specUri.toString())) {
+      if (!fileStatus.isDirectory() && fileStatus.getPath().getName().startsWith(specName)) {
```
htran1: We need to revisit having a version delimiter. Please add a note here. Currently without a delimiter the prefix check may match other specs that should not be matched.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/MutableSpecCatalog.java
```diff
@@ -36,5 +36,5 @@
   /**
    * Removes an existing {@link Spec} with the given URI. A no-op if such {@link Spec} does not exist.
```
htran1: The comment says that if the spec is not found then it is a no-op. If that is not the behavior then the comment should be changed to say that it raises a `SpecNotFoundException`.

arjun4084346: Corrected the comment. I chose to throw an exception to avoid verifying the existence of spec before calling remove(), in order to avoid duplication of work

PR #2157, opened 2017-11-02
[GOBBLIN-302] Handle stuck Helix workflow

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @htran1 @abti please review

### JIRA

### Description
It stop helix job running beyond a timeout and tries to stop underlying tasks

### Tests
- [] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -108,6 +108,8 @@ public class ConfigurationKeys {
   public static final String SCHEDULER_WAIT_FOR_JOB_COMPLETION_KEY = "scheduler.wait.for.job.completion";
   public static final String DEFAULT_SCHEDULER_WAIT_FOR_JOB_COMPLETION = Boolean.TRUE.toString();
 
+  public static final String HELIX_JOB_TIMEOUT_ENABLED_KEY = "job.timeout.enabled";
+  public static final String DEFAULT_HELIX_JOB_TIMEOUT_ENABLED = "false";
   public static final String HELIX_JOB_TIMEOUT_SECONDS = "job.timeout.seconds";
   public static final String DEFAULT_HELIX_JOB_TIMEOUT_SECONDS = "10800";
 
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -108,6 +108,8 @@ public class ConfigurationKeys {
   public static final String SCHEDULER_WAIT_FOR_JOB_COMPLETION_KEY = "scheduler.wait.for.job.completion";
   public static final String DEFAULT_SCHEDULER_WAIT_FOR_JOB_COMPLETION = Boolean.TRUE.toString();
 
+  public static final String HELIX_JOB_TIMEOUT_SECONDS = "job.timeout.seconds";
+  public static final String DEFAULT_HELIX_JOB_TIMEOUT_SECONDS = "10800";
 
   /**
    * Task executor and state tracker configuration properties.
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
@@ -107,6 +107,7 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
   private final TaskDriver helixTaskDriver;
   private final String helixQueueName;
   private final String jobResourceName;
+  private JobListener jobListener;
 
   private final FileSystem fs;
   private final Path appWorkDir;
@@ -167,6 +168,7 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
 
     this.taskStateCollectorService = new TaskStateCollectorService(jobProps, this.jobContext.getJobState(),
         this.eventBus, this.stateStores.getTaskStateStore(), outputTaskStateDir);
+    startCancellationExecutor();
   }
 
   @Override
@@ -290,6 +292,7 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
 
   public void launchJob(@Nullable JobListener jobListener)
       throws JobException {
+    this.jobListener = jobListener;
     boolean isLaunched = false;
     this.runningMap.putIfAbsent(this.jobContext.getJobName(), false);
     try {
@@ -359,8 +362,13 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
     return workUnitFile.toString();
   }
 
-  private void waitForJobCompletion() throws InterruptedException {
-    while (true) {
+  private void waitForJobCompletion()  throws InterruptedException {
+    LOGGER.info("Waiting for job to complete...");
+    long timeoutInSeconds;
+    timeoutInSeconds = Long.parseLong(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS,
+        ConfigurationKeys.DEFAULT_HELIX_JOB_TIMEOUT_SECONDS));
+    long endTime = System.currentTimeMillis() + timeoutInSeconds*1000;
+    while (System.currentTimeMillis() <= endTime) {
       WorkflowContext workflowContext = TaskDriver.getWorkflowContext(this.helixManager, this.helixQueueName);
       if (workflowContext != null) {
         org.apache.helix.task.TaskState helixJobState = workflowContext.getJobState(this.jobResourceName);
@@ -370,8 +378,34 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
           return;
         }
       }
-
       Thread.sleep(1000);
+    }
+    helixTaskDriverWaitToStop(this.helixQueueName, 10L);
+    try {
+      cancelJob(this.jobListener);
+    } catch (JobException e) {
+      throw new RuntimeException("Unable to cancel job " + jobContext.getJobName() + ": ", e);
+    }
+    this.helixTaskDriver.resume(this.helixQueueName);
+    LOGGER.info("stopped the queue, deleted the job");
+  }
+
+  /**
+   * Because fix https://github.com/apache/helix/commit/ae8e8e2ef37f48d782fc12f85ca97728cf2b70c4
+   * is not available in currently used version 0.6.9
+   */
+  private void helixTaskDriverWaitToStop(String workflow, long timeoutInSeconds) throws InterruptedException {
+    this.helixTaskDriver.stop(workflow);
+    long endTime = System.currentTimeMillis() + timeoutInSeconds*1000;
+    while (System.currentTimeMillis() <= endTime) {
+      WorkflowContext workflowContext = TaskDriver.getWorkflowContext(this.helixManager, this.helixQueueName);
+      if (workflowContext == null || workflowContext.getWorkflowState()
+          .equals(org.apache.helix.task.TaskState.IN_PROGRESS)) {
+        Thread.sleep(1000);
+      } else {
+        LOGGER.info("Successfully stopped the queue");
+        return;
+      }
     }
   }
 
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
@@ -364,11 +364,12 @@ public class GobblinHelixJobLauncher extends AbstractJobLauncher {
 
   private void waitForJobCompletion()  throws InterruptedException {
     LOGGER.info("Waiting for job to complete...");
-    long timeoutInSeconds;
-    timeoutInSeconds = Long.parseLong(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS,
+    boolean timeoutEnabled = Boolean.parseBoolean(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_ENABLED_KEY,
+        ConfigurationKeys.DEFAULT_HELIX_JOB_TIMEOUT_ENABLED));
+    long timeoutInSeconds = Long.parseLong(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS,
         ConfigurationKeys.DEFAULT_HELIX_JOB_TIMEOUT_SECONDS));
     long endTime = System.currentTimeMillis() + timeoutInSeconds*1000;
-    while (System.currentTimeMillis() <= endTime) {
+    while (!timeoutEnabled || System.currentTimeMillis() <= endTime) {
       WorkflowContext workflowContext = TaskDriver.getWorkflowContext(this.helixManager, this.helixQueueName);
       if (workflowContext != null) {
         org.apache.helix.task.TaskState helixJobState = workflowContext.getJobState(this.jobResourceName);
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
@@ -202,8 +202,16 @@ public class GobblinMultiTaskAttempt {
       task.shutdown();
     }
 
-    for (Task task: this.tasks) {
+    for (Task task : this.tasks) {
       task.awaitShutdown(1000);
+    }
+
+    for (Task task : this.tasks) {
+      if (task.cancel()) {
+        log.info("Task {} cancelled.", task.getTaskId());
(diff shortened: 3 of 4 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -363,20 +363,47 @@ public Void call() throws Exception {
...
   }
 
-  private void waitForJobCompletion() throws InterruptedException {
-    while (true) {
+  private void waitForJobCompletion()  throws InterruptedException {
+    LOGGER.info("Waiting for job to complete...");
+    long timeoutInSeconds;
+    timeoutInSeconds = Long.parseLong(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS,
+        ConfigurationKeys.DEFAULT_HELIX_JOB_TIMEOUT_SECONDS));
+    long endTime = System.currentTimeMillis() + timeoutInSeconds;
```
htran1: Mulitply by 1000 to convert to millis.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -363,20 +363,47 @@ public Void call() throws Exception {
...
+        if (helixJobState == org.apache.helix.task.TaskState.COMPLETED
+            || helixJobState == org.apache.helix.task.TaskState.FAILED
+            || helixJobState == org.apache.helix.task.TaskState.STOPPED) {
           return;
         }
       }
-
       Thread.sleep(1000);
     }
+    helixTaskDriverWaitToStop(this.helixQueueName, 10L);
```
htran1: Current time is in milliseconds, so either make it larger here or convert in the method.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -197,12 +199,20 @@ public Void call()
...
     }
 
-    for (Task task: this.tasks) {
-      task.awaitShutdown(1000);
+    for (Map.Entry<Task, Future<?>> task : this.tasks.entrySet()) {
+      task.getKey().awaitShutdown(1000);
+    }
+    for (Map.Entry<Task, Future<?>> task : this.tasks.entrySet()) {
+      if (task.getValue().cancel(true)) {
+        log.info("Task cancelled.");
```
htran1: I think you can add the awaitShutdown() here and remove the onTaskRunCompletion(). The task.run() has a finally block that calls onTaskRunCompletion().

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
arjun4084346: done

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
arjun4084346: done

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -201,9 +202,19 @@ public void shutdownTasks()
...
     }
 
-    for (Task task: this.tasks) {
+    for (Task task : this.tasks) {
       task.awaitShutdown(1000);
     }
+
+    for (Task task : this.tasks) {
+      if (task.getTaskFuture().cancel(true)) {
+        log.info("Task cancelled. Countdown and shutdown latches will be decreased here.");
```
aadarshjajodia: Can you please mention the task name which was cancelled here

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -201,9 +202,19 @@ public void shutdownTasks()
...
+        this.taskStateTracker.onTaskRunCompletion(task);
+        task.completeShutdown();
+      } else {
+        log.info("Task could not be cancelled");
```
aadarshjajodia: Same comment here

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
arjun4084346: done.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -363,20 +363,47 @@ public Void call() throws Exception {
...
+    startCancellationExecutor();
```
htran1: Shouldn't this be started before the stop call? Also, why don't we always start this for this job launcher? cancelJob() may be called in other instances.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -948,4 +953,12 @@ public boolean isSpeculativeExecutionSafe() {
     }
     return true;
   }
+
+  public void setTaskFuture(Future<?> taskFuture) {
+    this.taskFuture = taskFuture;
+  }
+
+  public Future<?> getTaskFuture() {
```
kadaan: Adding this method  and making `completeShutdown` public is not needed if we add a `cancel()` method to `Task` like: 
```java
  public boolean cancel() {
    if (this.taskFuture != null) {
      if (this.taskFuture.cancel(true)) {
        this.completeShutdown();
      }
    }
    return false;
  }
```

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -288,7 +289,7 @@ public boolean awaitShutdown(long timeoutInMillis)
     return this.shutdownLatch.await(timeoutInMillis, TimeUnit.MILLISECONDS);
   }
 
-  private void completeShutdown() {
+  public void completeShutdown() {
```
kadaan: See comment below about `cancel()` method.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -202,9 +203,19 @@ public void shutdownTasks()
       task.shutdown();
     }
 
-    for (Task task: this.tasks) {
+    for (Task task : this.tasks) {
       task.awaitShutdown(1000);
     }
+
+    for (Task task : this.tasks) {
+      if (task.getTaskFuture().cancel(true)) {
```
kadaan: If we call a `cancel()` method on task, then we don't have to expose `getTaskFuture()` or `completeShutdown()`.  See comments below.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -343,8 +354,8 @@ private boolean taskSuccessfulInPriorAttempt(String taskId) {
       // Create a new task from the work unit and submit the task to run
       Task task = createTaskRunnable(workUnitState, countDownLatch);
       this.taskStateTracker.registerNewTask(task);
+      task.setTaskFuture(this.taskExecutor.submit(task));
```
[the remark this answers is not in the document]
kadaan: Also, if we are going to "require" a future, we should probably remove the `TaskExecutor#execute()` methods and figure out how to ensure that a submitted task always has the future set on it without requiring the caller to do it explicitly.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -365,20 +369,51 @@ public Void call() throws Exception {
...
+    }
+    this.helixTaskDriver.resume(this.helixQueueName);
+    LOGGER.info("stopped the queue, deleted the job");
+  }
+
+  /**
+   * Because waitToStop is not available 0.6.7
+   */
+  private void helixTaskDriverWaitToStop(String workflow, long timeoutInSeconds) throws InterruptedException {
+    this.helixTaskDriver.stop(workflow);
```
kadaan: @htran1 Helix operations can fail if there is ZK connectivity issues.  Do we retry these at a lower level, or should then be retried by the caller?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -105,6 +105,9 @@
   public static final String SCHEDULER_WAIT_FOR_JOB_COMPLETION_KEY = "scheduler.wait.for.job.completion";
   public static final String DEFAULT_SCHEDULER_WAIT_FOR_JOB_COMPLETION = Boolean.TRUE.toString();
 
+  public static final String HELIX_JOB_TIMEOUT_SECONDS = "job.timeout.seconds";
+  public static final String DEFAULT_HELIX_JOB_TIMEOUT_SECONDS = "3600";
```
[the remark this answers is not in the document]
arjun4084346: Yes, but this 1 hour timeout applies to all Gobblin Helix jobs running on gobblin cluster only.
We will talk to the fliptop and other relevant teams and will adjust it accordingly before pushing in the changes.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -365,20 +369,51 @@ public Void call() throws Exception {
...
+    long endTime = System.currentTimeMillis() + timeoutInSeconds*1000;
+    while (System.currentTimeMillis() <= endTime) {
+      WorkflowContext workflowContext = TaskDriver.getWorkflowContext(this.helixManager, this.helixQueueName);
+      if (workflowContext == null || !workflowContext.getWorkflowState()
+          .equals(org.apache.helix.task.TaskState.STOPPED)) {
```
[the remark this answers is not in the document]
arjun4084346: When trying to stop, STOPPED is the intuitive state we are expecting it to be. I wrote this method to imitate waitToStop() method available in higher version of Helix
https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/task/TaskDriver.java#L519

But I agree that state could become FAILED or COMPLETED in between, I will talk to Helix team about this suggestion.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -365,20 +369,51 @@ public Void call() throws Exception {
     return workUnitFile.toString();
   }
 
-  private void waitForJobCompletion() throws InterruptedException {
-    while (true) {
+  private void waitForJobCompletion()  throws InterruptedException {
+    LOGGER.info("Waiting for job to complete...");
+    long timeoutInSeconds;
+    timeoutInSeconds = Long.parseLong(this.jobProps.getProperty(ConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS,
```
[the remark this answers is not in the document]
arjun4084346: Yes, the requirement of this ticket is to handle stuck helix tasks. Assuming any task can stuck, we will be cancelling them after some timeout. jobs can set the timeout value, otherwise the default timeout value will be used, which is 3 hours - same as azkaban's default SLA time

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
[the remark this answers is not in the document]
arjun4084346: Will change this. Helix team will also be making changes in their code. Thank you for the suggestion.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
arjun4084346: done

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
arjun4084346: done

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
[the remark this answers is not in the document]
arjun4084346: Why do we not want TaskExecutor#execute() along submit() ?
Both methods were present since the beginning.
taskExecutor#execute() is not being used in gobblin anywhere, and can be removed safely. But there is a possibility some forked branch is using it.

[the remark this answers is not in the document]
kadaan: Right, but seems like that method is not work correctly going forward.  Unless there is a compelling reason to keep it, it seems like we shouldn't have a method there that "might" be used.  Then we have to test it and maintain it.  That is one of the main challenges with Gobblin at the moment.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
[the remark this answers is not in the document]
kadaan: Hmmm.  I'm worried about having a naive SLA breaking existing uses.  Is there no way to have a heartbeat style check to ensure that helix tasks are not stuck?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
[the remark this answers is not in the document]
arjun4084346: No no, execute() does work correctly, but execute() does not return a Future, and we need a handle to the Future in this case. In one of the Gobblin based internal project of LinkedIn, we used execute()

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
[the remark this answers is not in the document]
arjun4084346: I will give it a thought and talk to the Helix team.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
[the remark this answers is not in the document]
kadaan: It only works correctly if not used in the scope of Helix.  Either way, but we need to work to a very clear interface that both internal LinkedIn projects and external projects use.  This will ensure clarity and reduce the test surface.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -948,4 +953,23 @@ public boolean isSpeculativeExecutionSafe() {
...
+  public void setTaskFuture(Future<?> taskFuture) {
+    this.taskFuture = taskFuture;
+  }
+
+  /**
+   * return true if the task is successfully cancelled.
+   * @param taskStateTracker TaskStateTracker tracking this task
+   * @return
+   */
+  public boolean cancel(TaskStateTracker taskStateTracker) {
```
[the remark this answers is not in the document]
arjun4084346: my bad. corrected.

kadaan:
I really don't know much about Helix, but it seems like retries for ZK operations were added in commit [bada911](https://github.com/apache/helix/commit/bada911c7f246cf685c30323e118402cca89111d) from Mar 6, 2017.  Unfortunately, this fix is only in master.  Also, this fix seems to retry without any backoff so it may not gracefully handle intermittent network issues.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -364,8 +365,12 @@ public void run() {
     } catch (Throwable t) {
       failTask(t);
     } finally {
-      this.taskStateTracker.onTaskRunCompletion(this);
-      completeShutdown();
+      if (this.taskFuture == null || !this.taskFuture.isCancelled()) {
```
htran1: This block needs to synchronize access to the taskFuture with the cancel() call below. It should also modify the state of the taskFuture to avoid double cancellation when the cancel() block below gets called after this block executes.

arjun4084346: I will add synchronized access to the future, but I think changing future's state is not required. The state should be changed to 'completed' when it actually completes.

PR #2161, opened 2017-11-07
[GOBBLIN-307] Implement lineage event as LineageEventBuilder in gobblin

zxcware:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-307

### Description
  - Previous implementation of lineage event is not clearly defined. This PR provides a `LineageEventBuilder` which captures the definition of a lineage event
  - The way lineage info is collected in a job is framed by `LineageInfo`
  - Establish the pattern of defining a specific event in Gobblin

### Tests
  - `LineageEventTest`

### Commits


gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageEventBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+  public static boolean isLineageEvent(GobblinTrackingEvent event) {
+    String eventType = event.getMetadata().get(EVENT_TYPE);
+    return StringUtils.isNotEmpty(eventType) && eventType.equals(LINEAGE_EVENT_TYPE);
+  }
+
+  /**
+   * Create a {@link LineageEventBuilder} from a {@link GobblinEventBuilder}. An inverse function
+   * to {@link LineageEventBuilder#build()}
+   */
+  public static LineageEventBuilder fromEvent(GobblinTrackingEvent event) {
```
yukuai518: When will this be used?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageEventBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+
+    return destination != null ? destination.equals(event.destination) : event.destination == null;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = name.hashCode();
+    result = 31 * result + namespace.hashCode();
+    result = 31 * result + metadata.hashCode();
+    result = 31 * result + (source != null ? source.hashCode() : 0);
```
yukuai518: Should we have different calculation for source and destination? It seems like (source=A, dest=B) has the same hash code with (source=B, dest=A)?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageInfo.java
```diff
@@ -0,0 +1,200 @@
...
+   *
+   * <p>
+   *   Only the {@link org.apache.gobblin.writer.DataWriter} or {@link org.apache.gobblin.publisher.DataPublisher}
+   *   is supposed to put the destination dataset information. Since different branches may concurrently put,
+   *   the method is implemented to be threadsafe
+   * </p>
+   */
+  public static void putDestination(DatasetDescriptor destination, int branchId, State state) {
+    if (!hasLineageInfo(state)) {
+      return;
```
yukuai518: Should we throw an exception here? In which case we will putDestination without having lineageInfo setup in the state?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageInfo.java
```diff
@@ -0,0 +1,200 @@
...
+ *   <ol>
+ *     <li> source sets its {@link DatasetDescriptor} to each work unit </li>
+ *     <li> destination puts its {@link DatasetDescriptor} to the work unit </li>
+ *     <li> load and send all lineage events from all states </li>
+ *     <li> purge lineage info from all states </li>
+ *   </ol>
+ * </p>
+ */
+public final class LineageInfo {
+  static final String BRANCH = "branch";
```
yukuai518: Do you think these names like "branch or names" should have a namespace or scope? Otherwise user can overwrite these attribute in gobblin constructs, without noticing these attributes are important for sending lineage event. We need to document this behavior somewhere. Or we should have a scoped name.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopyableDatasetBase.java
```diff
@@ -18,11 +18,15 @@
...
 import org.apache.gobblin.dataset.Dataset;
+import org.apache.gobblin.dataset.DatasetDescriptor;
 
 
 /**
  * A common superinterface for {@link Dataset}s that can be operated on by distcp.
  * Concrete classes must implement a subinterface of this interface ({@link CopyableDataset} or {@link IterableCopyableDataset}).
  */
 public interface CopyableDatasetBase extends Dataset {
+  default DatasetDescriptor getDatasetDescriptor() {
```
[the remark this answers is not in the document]
zxcware: It's a descriptor associated with this dataset. The dataset can be a source dataset or destination dataset.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageEventBuilder.java
zxcware: For now, it will be used internally to transform a gobblin traking event to wherehows event

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageEventBuilder.java
zxcware: source and destination will have different hash code because they will be different objects.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageInfo.java
zxcware: Not really. This is really a constant. It's not a name. And it's final, won't be changed.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageInfo.java
zxcware: Lineage info shouldn't fail a job. I think a better option is to add a log warn message

gobblin-core/src/main/java/org/apache/gobblin/publisher/TimePartitionedDataPublisher.java
```diff
@@ -68,15 +65,4 @@ protected void addWriterOutputToExistingDir(Path writerOutput, Path publisherOut
       movePath(parallelRunner, workUnitState, status.getPath(), outputPath, branchId);
     }
   }
-
-  @Override
-  protected void publishData(WorkUnitState state, int branchId, boolean publishSingleTaskData, Set<Path> writerOutputPathsMoved) throws IOException {
```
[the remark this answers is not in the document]
zxcware: Because the logic is moved to `BaseDataPublisher`

gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
```diff
@@ -0,0 +1,114 @@
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    DatasetDescriptor that = (DatasetDescriptor) o;
+    return !(!platform.equals(that.platform) || !name.equals(that.name) || !metadata.equals(that.metadata));
```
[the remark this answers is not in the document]
zxcware: +1

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/lineage/LineageInfo.java
zxcware: Actually. for job configurations, we do want to provide the flexibility where namespace can vary but keys are the same.

yukuai518: What would happen if some components set "name=XYZ" in gobblin State object for other purpose instead of lineage event? Will it break the lineage event logic?

PR #2163, opened 2017-11-08
[GOBBLIN-309] Disabled rewrite and enabled retry for adding jar file on HDFS

autumnust:
This is a rough PR for initial review. 

The approach is to keep the record of unsuccessful copy of jar files, and have busy spin loop to ensure they are added successfully, or drop the attempt if any other job instances have added jar file. 

@htran1  Can you help on this rough review? Just want to make sure it is on the right track. Thanks.

### JIRA
- [ ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-309] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-309

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -21,8 +21,10 @@ import java.io.DataOutputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.net.URI;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
@@ -56,9 +58,9 @@ import com.google.common.io.Closer;
 import com.google.common.util.concurrent.ServiceManager;
 import com.typesafe.config.ConfigFactory;
 
+import org.apache.gobblin.broker.SharedResourcesBrokerFactory;
 import org.apache.gobblin.broker.gobblin_scopes.GobblinScopeTypes;
 import org.apache.gobblin.broker.gobblin_scopes.JobScopeInstance;
-import org.apache.gobblin.broker.SharedResourcesBrokerFactory;
 import org.apache.gobblin.broker.iface.SharedResourcesBroker;
 import org.apache.gobblin.commit.CommitStep;
 import org.apache.gobblin.configuration.ConfigurationKeys;
@@ -113,6 +115,11 @@ public class MRJobLauncher extends AbstractJobLauncher {
   static final String INPUT_DIR_NAME = "input";
   private static final String OUTPUT_DIR_NAME = "output";
   private static final String WORK_UNIT_LIST_FILE_EXTENSION = ".wulist";
+
+  // Configuration that make uploading of jar files more reliable,
+  // since multiple Gobblin Jobs are sharing the same jar directory.
+  private static final int MAXIMUM_JAR_COPY_RETRY_TIMES = 5;
+  private static final int WAITING_TIME_ON_IMCOMPLETE_UPLOAD = 3000;
 
   private static final Splitter SPLITTER = Splitter.on(',').omitEmptyStrings().trimResults();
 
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -86,6 +68,25 @@
 import org.apache.gobblin.util.JobLauncherUtils;
 import org.apache.gobblin.util.ParallelRunner;
 import org.apache.gobblin.util.SerializationUtils;
+import org.apache.hadoop.conf.Configuration;
```
htran1: The org.* imports should be moved back to being after the java.* imports. Only the org.apache.gobblin ones should be here.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -389,23 +391,61 @@ static void serializeJobState(FileSystem fs, Path mrJobDir, Configuration conf,
...
+        // Adding destJarFile into HDFS until it exists
         if (!this.fs.exists(destJarFile)) {
           // Copy the jar file from local file system to HDFS
-          this.fs.copyFromLocalFile(status.getPath(), destJarFile);
+          // Set the first parameter as false for not deleting sourceFile
+          // Set the second parameter as false for not overwriting existing file on the target, by default it is true.
+            // If the file is preExisted but overwrite flag set to false, then an IOException if thrown.
+          try {
+            this.fs.copyFromLocalFile(false, false, status.getPath(), destJarFile);
+          }catch (IOException ioe){
```
htran1: Add spaces after and before braces.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -389,23 +391,61 @@ static void serializeJobState(FileSystem fs, Path mrJobDir, Configuration conf,
...
-        Path destJarFile = new Path(this.fs.makeQualified(baseDir), status.getPath().getName());
+        Path destJarFile = calculateDestJarFile(status, jarFileDir);
+        // Adding destJarFile into HDFS until it exists
         if (!this.fs.exists(destJarFile)) {
           // Copy the jar file from local file system to HDFS
-          this.fs.copyFromLocalFile(status.getPath(), destJarFile);
+          // Set the first parameter as false for not deleting sourceFile
+          // Set the second parameter as false for not overwriting existing file on the target, by default it is true.
+            // If the file is preExisted but overwrite flag set to false, then an IOException if thrown.
+          try {
```
htran1: You can avoid duplicating the code by adding the retry here. Any reason the retry is done later?
We should add a check for file size to make sure that the file is complete before we continue.

autumnust: I am trying to increase the window time between a failure of copying and a retry for a specific file. I can refactor it if it is not necessary in this case. 

By "add a check for file size", does that mean we are not only checking the file existence but compare the file size when we decided that copy is not necessary for a file?  Does this check mainly to avoid jar file being corrupted ? It doesn't seem to be necessary since we have disable the overwrite flag.

htran1: The failure is most likely due to the file existing, so I think an immediate retry should succeed. The file size check is to make sure that the file copied by the other process is complete before proceeding to use it in this process. The file can exist, but still be in the process of getting filled in.

autumnust: Got it.  Thanks for explanation.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -392,14 +395,30 @@ private void addJars(Path jarFileDir, String jarFileList, Configuration conf)
...
+            // Set the first parameter as false for not deleting sourceFile
+            // Set the second parameter as false for not overwriting existing file on the target, by default it is true.
+            // If the file is preExisted but overwrite flag set to false, then an IOException if thrown.
+            this.fs.copyFromLocalFile(false, false, status.getPath(), destJarFile);
+          } catch (IOException ioe) {
+            LOG.warn("Path:" + destJarFile + " is not copied successfully. Will require retry.");
+            retryCount += 1 ;
+            if (retryCount >= MAXIMUM_JAR_COPY_RETRY_TIMES) {
+              LOG.error("The jar file:" + destJarFile + " has been successfully copied into hdfs");
```
htran1: This message doesn't look right since this is the error case.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -392,14 +395,30 @@ private void addJars(Path jarFileDir, String jarFileList, Configuration conf)
...
+              // If retry reaches upper limit, skip copying this file. 
+              continue jarFileLoop;
```
htran1: Should break here to avoid a need for the label.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -392,14 +395,30 @@ private void addJars(Path jarFileDir, String jarFileList, Configuration conf)
...
-        if (!this.fs.exists(destJarFile)) {
-          // Copy the jar file from local file system to HDFS
-          this.fs.copyFromLocalFile(status.getPath(), destJarFile);
+        // For each FileStatus there are chances it could fail in copying at the first attempt, due to file-existence
+        // or file-copy is ongoing by other job instance since all Gobblin jobs share the same jar file directory.
+        // the retryCount is to avoid cases (if any) where retry is going too far and causes job hanging.
+        int retryCount = 0 ;
+        Path destJarFile = calculateDestJarFile(status, jarFileDir);
+          // Adding destJarFile into HDFS until it exists and the size of file on targetPath matches the one on local path.
+        while(!this.fs.exists(destJarFile) || fs.listStatus(destJarFile).length != status.getLen()) {
```
htran1: In the file exists, but size does not match case we need to give time for the other process to copy the file.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
autumnust: If break, line 423 and 424 will still be executed. Is that OK in this case ?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -392,23 +399,56 @@ private void addJars(Path jarFileDir, String jarFileList, Configuration conf)
...
+        // For each FileStatus there are chances it could fail in copying at the first attempt, due to file-existence
+        // or file-copy is ongoing by other job instance since all Gobblin jobs share the same jar file directory.
+        // the retryCount is to avoid cases (if any) where retry is going too far and causes job hanging.
+        int retryCount = 0 ;
+        boolean shouldFileBeAddedIntoDC = true;
+        Path destJarFile = calculateDestJarFile(status, jarFileDir);
+          // Adding destJarFile into HDFS until it exists and the size of file on targetPath matches the one on local path.
+        while(!this.fs.exists(destJarFile) || fs.getFileStatus(destJarFile).getLen() != status.getLen()) {
+          try {
+            if (fs.getFileStatus(destJarFile).getLen() != status.getLen()){
```
htran1: This check should only be done if `fs.exists(destJarFile)` otherwise and exception is raised.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -392,23 +399,56 @@ private void addJars(Path jarFileDir, String jarFileList, Configuration conf)
...
+              Thread.sleep(WAITING_TIME_ON_IMCOMPLETE_UPLOAD);
```
htran1: 15 seconds might not be sufficient time for a large jar. Can you make the retry parameter configurable so it can be bumped up if a job requires more time to copy jars?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -386,34 +394,65 @@ static void serializeJobState(FileSystem fs, Path mrJobDir, Configuration conf,
...
+        // Adding destJarFile into HDFS until it exists and the size of file on targetPath matches the one on local path.
+        while (!this.fs.exists(destJarFile) || fs.getFileStatus(destJarFile).getLen() != status.getLen()) {
+          try {
+            if (this.fs.exists(destJarFile) && fs.getFileStatus(destJarFile).getLen() != status.getLen()) {
+              Thread.sleep(WAITING_TIME_ON_IMCOMPLETE_UPLOAD);
+            }
+            // Set the first parameter as false for not deleting sourceFile
+            // Set the second parameter as false for not overwriting existing file on the target, by default it is true.
+            // If the file is preExisted but overwrite flag set to false, then an IOException if thrown.
+            this.fs.copyFromLocalFile(false, false, status.getPath(), destJarFile);
```
htran1: This should only be done when !this.fs.exists(destJarFile).

autumnust: Adding logic for throwing exception above should avoid copying jar file which is being uploaded by other Gobblin Job blindly.

PR #2181, opened 2017-12-02
[GOBBLIN-329]Add a new basic cluster integration test

HappyRay:
Run the hello world job and verify that the output files are generated.

It takes about 12 seconds to run locally by itself within IDEA and
4 seconds when running with others.

The test is modeled after the existing GobblinClusterKillTest test.

Also fixed a typo: GobblinClusterConfigurationKeys.CLUSTRER_WORK_DIR should be
GobblinClusterConfigurationKeys.CLUSTER_WORK_DIR

### DIFF
--- a/gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
+++ b/gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
@@ -34,11 +34,6 @@
 
 package org.apache.gobblin.cluster;
 
-
-import com.google.common.base.Optional;
-import com.google.common.io.Resources;
-import com.typesafe.config.Config;
-import com.typesafe.config.ConfigFactory;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -49,13 +44,20 @@ import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.curator.test.TestingServer;
-import org.apache.gobblin.testing.AssertWithBackoff;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.Test;
+
+import com.google.common.base.Optional;
+import com.google.common.io.Resources;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+import org.apache.gobblin.testing.AssertWithBackoff;
 
 
 public class ClusterIntegrationTest {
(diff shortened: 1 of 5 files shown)


gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -0,0 +1,224 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.cluster;
+
+
+import com.google.common.base.Optional;
```
htran1: Our import order convention is java, org, com, org.apache.gobblin.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -0,0 +1,224 @@
...
+    Config configFromResource = getConfigFromResource();
+    Config configOverride = getConfigOverride();
+    _config = configOverride.withFallback(configFromResource).resolve();
+  }
+
+  private Config getConfigOverride() {
+    Map<String, String> configMap = new HashMap<>();
+    String zkConnectionString = _testingZKServer.getConnectString();
+    configMap.put(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY, zkConnectionString);
+    configMap.put(GobblinClusterConfigurationKeys.CLUSTRER_WORK_DIR, _workPath.toString());
```
htran1: Existing typo. It is only referenced in one other file, so can you fix it?

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
HappyRay: How is it enforced? 
I used intellij's import organization feature. It's inefficient to enforce this convention without automation.

htran1: There is an intellij code style template file at https://gobblin.readthedocs.io/en/latest/developer-guide/CodingStyle/. It needs to be updated for the org.apache.gobblin namespace.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
HappyRay: Fixed. Thanks

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
HappyRay: Fixed.

PR #2204, opened 2017-12-14
[GOBBLIN-336] Set unused Helix task result info field to an empty string

HappyRay:
This field is not used.
This change removes the need for the child task process to send the
workunit count back to the control process.


yukuai518:
+1 LGTM

PR #2208, opened 2017-12-14
[GOBBLIN-336] move single task runner option constants to the option class

HappyRay:
This will improve class coherency.

Also removed an unused constant.


HappyRay:
Thanks

PR #2209, opened 2017-12-15
[GOBBLIN-336] Create a class to start the task java child process.

HappyRay:
It's not used yet.

Testing:

Add a new unit test to verify the correct parameters are passed to the java ProcessBuilder.

Manually tested the class. Verified that the child process is started and console output is redirected to the parent process's console.


HappyRay:
depend on #2208

gobblin-utility/src/main/java/org/apache/gobblin/util/GobblinProcessBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
abti: double license

abti: multiple license actually :)

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/TaskInstance.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+class TaskInstance {
+  Process taskProcess;
+
+  public TaskInstance(Process taskProcess) {
+    this.taskProcess = taskProcess;
+  }
+
+  void Cancel() {
+    this.taskProcess.destroyForcibly();
```
abti: why not also let TaskInstance attempt to gracefully exit?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/TaskInstance.java
```diff
@@ -0,0 +1,33 @@
...
+ * A reference to an instance of a single task that runs in a separate process.
+ */
+class TaskInstance {
+  Process taskProcess;
+
+  public TaskInstance(Process taskProcess) {
+    this.taskProcess = taskProcess;
+  }
+
+  void Cancel() {
```
abti: should the return type be exit code?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private void addOptions() {
+      addClusterConfigPath();
+      addJobId();
+      addWorkUnitPath();
+    }
+
+    private void addClusterConfigPath() {
+      this.cmd.add(formatParam(CLUSTER_CONFIG_FILE_PATH));
+      this.cmd.add(SingleTaskLauncher.this.clusterConfigFilePath.toString());
```
abti: in all addX() methods, shouldn't we check for availability of param values they are adding? else we might have empty or null strings being attached to param keys

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
```diff
@@ -0,0 +1,121 @@
...
+
+import org.apache.gobblin.util.GobblinProcessBuilder;
+import org.apache.gobblin.util.SystemPropertiesWrapper;
+
+import static org.apache.gobblin.cluster.SingleTaskRunnerMainOptions.CLUSTER_CONFIG_FILE_PATH;
+import static org.apache.gobblin.cluster.SingleTaskRunnerMainOptions.JOB_ID;
+import static org.apache.gobblin.cluster.SingleTaskRunnerMainOptions.WORK_UNIT_FILE_PATH;
+
+
+class SingleTaskLauncher {
```
abti: why not let the access modifier be public? it might be reused in AWS mode or Azure mode which can be different package.

HappyRay: We can change it easily when such a need comes up. I would prefer keeping the scope as small as possible.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
HappyRay: The assumption is that these values are valid when they get here. It expects the validation to be done at the place where these values are created.

What additional values would validation here bring?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/TaskInstance.java
HappyRay: I deleted this file. You may be reviewing an older version.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/TaskInstance.java
HappyRay: I am not sure if it is needed yet. 
We can discuss in person. 
We can add it later if needed.

gobblin-utility/src/main/java/org/apache/gobblin/util/GobblinProcessBuilder.java
HappyRay: This is annoying. I hope to find out what goes wrong with my intellij settings. It's supposed to detect and update it automatically.

Fixed.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/TaskInstance.java
HappyRay: Anyway, this change is no longer in this PR.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
```diff
@@ -0,0 +1,121 @@
...
+      return this.cmd;
+    }
+
+    private void addClassName() {
+      final String runnerClassName = SingleTaskRunnerMain.class.getCanonicalName();
+      this.cmd.add(runnerClassName);
+    }
+
+    private void addJavaBin() {
+      final String javaHomeDir = SingleTaskLauncher.this.propertiesWrapper.getJavaHome();
```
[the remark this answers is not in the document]
HappyRay: This is a problem. I will work on a fix in a later PR.

Thanks.

HappyRay:
@htran1 It's ready to merge. Thanks

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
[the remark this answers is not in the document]
HappyRay: This is a problem. I will work on a fix in a later PR.

Thanks.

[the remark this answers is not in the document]
HappyRay: Turns out this is not a problem.
I verified 
that it returns the path to the JRE that is used to run the program which does the lookup. It is not related to JAVA_HOME.

Added a new change to clarify.

HappyRay:
@htran1 Ready to merge?

HappyRay:
Thanks

PR #2216, opened 2017-12-18
[GOBBLIN-336] Add a method to save a config to a file

HappyRay:
Added unit tests.

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/FileUtils.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/FileUtils.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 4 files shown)


gobblin-utility/src/main/java/org/apache/gobblin/util/FileUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+
+public class FileUtils {
+  public void saveToFile(final String text, final Path destPath)
```
abti: You can use this: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#writeStringToFile(java.io.File,%20java.lang.String,%20java.nio.charset.Charset) 
We use this already in our code, so its available as a dependency.

HappyRay: Yes, I am aware of this lib/method.

This method will flush the buffer. This is important for the child process to reliably get the file. This apache API doc doesn't allow you to control that behavior.

Secondly, the method is a static method which makes it harder to mock out in a unit test.

HappyRay:
Thanks

PR #2224, opened 2017-12-20
[GOBBLIN-336] Run tasks in a separate process if enabled

HappyRay:
Default is disabled.

Testing:
Added a unit test.

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
@@ -19,6 +19,7 @@ package org.apache.gobblin.cluster;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -137,11 +142,15 @@ public GobblinTaskRunner(String applicationName, String helixInstanceName, Strin
...
     this.taskRunnerId = taskRunnerId;
 
     Configuration conf = HadoopUtils.newConfiguration();
-    this.fs = buildFileSystem(this.config, conf);
+    this.fs = buildFileSystem(config, conf);
+    Path appWorkDir = appWorkDirOptional.isPresent() ? appWorkDirOptional.get()
+        : GobblinClusterUtils
+            .getAppWorkDirPathFromConfig(config, this.fs, applicationName, applicationId);
+
+    this.config = saveConfigToFile(config, appWorkDir);
```
abti: we should only need this if we are running task in separate process. so should this be within a check?

HappyRay: Correct.
However, there is no harm in generating this config. In fact, it is useful as a debugging aid.

abti: Also seems like all tasks will save to the same location? unless I am missing something.

HappyRay: No. The config is cluster level config. It will only be saved once per worker process when it starts up.

HappyRay: The child processes will read from the same location. 
Also, the assumption is that if there are multiple worker processes running on the same host, they will start from different work directory.

abti: Thanks, just saw that this gets initialized once. Its a fair assumption that each worker process will have its own work directory.

HappyRay:
Thanks

PR #2230, opened 2018-01-05
[GOBBLIN-357] Fix logging during Zookeeper connection loss

kadaan:
@abti Feedback?

Fix log messages when the zookeeper connection is lost which were not being formatted correctly.
Add start & end log messages to job cancellation.


abti:
LGTM +1

PR #2235, opened 2018-01-09
Log child process cmd line at the debug level

HappyRay:
Due to the long class path, the log can be long and makes the debug
log less readable.

Testing:
Verified with the integration test with a custom log4j config.


HappyRay:
Thanks

PR #2236, opened 2018-01-09
[GOBBLIN-366] Adding retry into LoopingDatasetSource

autumnust:
Dear Gobblin maintainers,

### JIRA
- [x] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-366] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-366

### Description

### Tests

### Commits


abti:
Please add Gobblin Jira for it

autumnust:
@ibuenros Can you review it? Thanks.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -85,31 +86,52 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
...
-          int wu1Ordinal = wu1.getPropAsInt(WORK_UNIT_ORDINAL);
-          int wu2Ordinal = wu2.getPropAsInt(WORK_UNIT_ORDINAL);
-          return wu1Ordinal > wu2Ordinal ? wu1 : wu2;
-        });
-      } catch (NumberFormatException nfe) {
-        throw new RuntimeException("Work units in state store are corrupted! Missing or malformed " + WORK_UNIT_ORDINAL);
+      int tempWorkUnitOrdinal = 0;
+      Optional<WorkUnitState> optionalMaxWorkUnit = Optional.empty();
+      List<WorkUnit> failedPreviousWorkUnits = new ArrayList<>();
+      int maxRetryTimes = state.contains(ConfigurationKeys.MAX_TASK_RETRIES_KEY) ? state
```
ibuenros: I don't think you should use task_max_retries, as that configuration is intended for retries within a job. Create a new one instead.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -85,31 +86,52 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
...
+          .getPropAsInt(ConfigurationKeys.MAX_TASK_RETRIES_KEY) : ConfigurationKeys.DEFAULT_MAX_TASK_RETRIES;
+
+      // Searching for workunit with maximum ordinal meanwhile collect all failed workunit in previous run.
+      // These two processes should not influent each other.
+
+      // Workunits that exceed max retry limit will not be added into failedPreviousWorkUnits
+      // and corresponding event/notification should be sent in the task level.
```
ibuenros: You are not sending events right now, are you? Also, instead of just discarding it and sending an event, we should execute the work unit on the next run.

erwa:
General comment: can you please avoid reformatting changes in your PRs? All the reformatting changes create a lot of visual noise and make the change harder to review.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -85,53 +88,82 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
...
+      // and rollovered into next quota-available execution.
+      for (WorkUnitState workUnitState : previousWorkUnitStates) {
+        if ((workUnitState.getWorkingState() == WorkUnitState.WorkingState.FAILED
+            || workUnitState.getWorkingState() == WorkUnitState.WorkingState.ROLLOVERED) && (
+            !workUnitState.contains(ConfigurationKeys.TASK_RETRIES_KEY)
+                || workUnitState.getPropAsInt(ConfigurationKeys.TASK_RETRIES_KEY) < maxRetryTimes + 1)) {
+          log.info("Dataset " + workUnitState.getProp(DATASET_URN) + " is failed previously, retrying...");
+          WorkUnit retryWorkUnit = new WorkUnit(workUnitState.getWorkunit());
+          failedPreviousWorkUnits.add(retryWorkUnit);
+        }
```
erwa: And if the max retries exceeds the maxRetryTimes, we just silently drop the workunit and never try processing it again? A few questions:

* Before your change here, would failed workunits just be retried forever in all subsequent executions?
* Does Gobblin have a concept of "permanently failed" workunits that are written to a separate location and never retried in subsequent executions?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -169,6 +204,22 @@ public DeepIterator(Iterator<Dataset> baseIterator, String previousDatasetUrnWat
 
     @Override
     protected WorkUnit computeNext() {
+      if (retryWorkUnits.hasNext()) {
```
erwa: Won't partitions that appear in the retryWorkUnits be double-processed? I think they'll be processed once when retryWorkUnits is iterated over and then processed a second time when the baseIterator is iterated over. In order to avoid double processing, I think you would need to filter out the partitions that have exceeded the max cross-execution retries limit from the dataset stream returned by the dataset finder.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
autumnust: max retries exceeds the limit, a event will be sent and it will result in a failed workunit therefore never being retried automatically. But users/admins should be aware of this by our notification mechanism which I believe has not been fully discussed yet @ibuenros . 

For your questions: 
- No. Current retry model will retry the failed workunit in current execution until reaching the maximum retry limit. 
- No. I think we should discuss the notification mechanism about such kind of failure.

erwa: Here, I think we should add least `log.warn()` about the workunits we are skipping because they've exceeded the max retries.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -85,53 +88,82 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
       int maxWorkUnits = state.getPropAsInt(MAX_WORK_UNITS_PER_RUN_KEY, MAX_WORK_UNITS_PER_RUN);
```
erwa: This method is getting a little hard to read. I suggest the following refactorings:

```
// (I suggest renaming maxRetryTimes to just maxRetries)
int maxRetries = getMaxRetries(state);

List<WorkUnit> failedPreviousWorkUnits = getFailedPreviousWorkUnits(previousWorkUnitStates);
WorkUnitState maxWorkUnitState = getMaxWorkUnitState(previousWorkUnitStates);

if (maxWorkUnitState != null && !maxWorkUnitState.getPropAsBoolean(END_OF_DATASETS_KEY, false)) {
  // set watermarks
}
```

I think it's worth iterating over previousWorkUnitStates multiple times for the sake of readability.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
     return Lists.newArrayList(getWorkunitStream(state).getMaterializedWorkUnitCollection());
   }
 
+  /**
+   * Collect all failed workunit in previous run into a set.
```
erwa: Javadoc says "set" but method actually returns a "list".

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+   *
+   * Workunits that exceed max retry limit will not be added into failedPreviousWorkUnits
+   * and rollovered into next quota-available execution.
```
erwa: rollovered -> rolled over

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+   *
+   * @param retriedUrns Track all dataset/partition's urns that is included in the retried workunits, so that in the
+   *                    case where the number of remaining dataset/partitions are less than the maximum number of
+   *                    workunits can be processed by single execution, a retried workunit won't be processed twice
+   */
```
erwa: If you have @param for one param, please add @param for all params (previousWorkUnitStates and maxRetries, too).

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+  private List<WorkUnit> getPreviousFailedWorkUnits(List<WorkUnitState> previousWorkUnitStates, int maxRetries,
+      Set<String> retriedUrns) {
+    List<WorkUnit> failedPreviousWorkUnits = new ArrayList<>();
+    for (WorkUnitState workUnitState : previousWorkUnitStates) {
+      if ((workUnitState.getWorkingState() == WorkUnitState.WorkingState.FAILED
+          || workUnitState.getWorkingState() == WorkUnitState.WorkingState.ROLLOVER)) {
```
erwa: Please create a local variable `WorkingState workingState = workUnitState.getWorkingState();` and using `workingState` to make this more readable. Also, you have an extra set of parentheses around the if check.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+        if (!workUnitState.contains(ConfigurationKeys.TASK_RETRIES_KEY)
+            || workUnitState.getPropAsInt(ConfigurationKeys.TASK_RETRIES_KEY) < maxRetries + 1) {
```
erwa: Why is there a `+ 1` here? I would expect the `maxRetries` passed in to this method to be used unchanged.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+          log.info("Dataset " + workUnitState.getProp(DATASET_URN) + " is failed previously, retrying...");
+          WorkUnit retryWorkUnit = new WorkUnit(workUnitState.getWorkunit());
```
erwa: Why is creating a new WorkUnit necessary? Why can't we just reuse `workUnitState.getWorkunit()`? Please add a comment explaining this.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+   *
+   * Workunits that exceed max retry limit will not be added into failedPreviousWorkUnits
+   * and rollovered into next quota-available execution.
+   *
+   * @param retriedUrns Track all dataset/partition's urns that is included in the retried workunits, so that in the
+   *                    case where the number of remaining dataset/partitions are less than the maximum number of
+   *                    workunits can be processed by single execution, a retried workunit won't be processed twice
+   */
+  private List<WorkUnit> getPreviousFailedWorkUnits(List<WorkUnitState> previousWorkUnitStates, int maxRetries,
+      Set<String> retriedUrns) {
```
erwa: There is no need to pass in a `Set` to this method that gets modified. The `Set` can be populated outside this method (better yet, in a separate method) from the `List<WorkUnit>` returned by this method.

Generally, it's desirable for methods not to have side effects. It's best if a method takes inputs and returns an output based on the inputs without modifying any of the inputs.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+            || workUnitState.getPropAsInt(ConfigurationKeys.TASK_RETRIES_KEY) < maxRetries + 1) {
+          log.info("Dataset " + workUnitState.getProp(DATASET_URN) + " is failed previously, retrying...");
+          WorkUnit retryWorkUnit = new WorkUnit(workUnitState.getWorkunit());
+          failedPreviousWorkUnits.add(retryWorkUnit);
+          String retriedUrn = workUnitState.getWorkunit().contains(PARTITION_URN) ? retryWorkUnit.getProp(PARTITION_URN)
+              : retryWorkUnit.getProp(DATASET_URN);
+          log.info("The workunit: " + retriedUrn + " is added into retry bucket.");
+          retriedUrns.add(retriedUrn);
+        } else {
+          log.warn("Dataset " + workUnitState.getProp(DATASET_URN) + " exceeds retry limit : " + maxRetries);
```
erwa: Please add "skipping" in this message, too.

Also, can we emit an event to track this? (I'm ok adding a TODO for now and doing this in a separate patchset.)

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+        }
+      }
+    }
+    return failedPreviousWorkUnits;
+  }
+
+  /**
+   * Searching for workunit with maximum ordinal.
+   * @param previousWorkUnitStates
+   * @return
```
erwa: Please add explanations for @return and @param or omit them.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+   */
+  private Optional<WorkUnitState> getMaxWorkUnitState(List<WorkUnitState> previousWorkUnitStates) {
+    int tempWorkUnitOrdinal = 0;
+    Optional<WorkUnitState> optionalMaxWorkUnit = Optional.empty();
```
erwa: optionalMaxWorkUnit -> optionalMaxWorkUnitState

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+    for (WorkUnitState workUnitState : previousWorkUnitStates) {
+      try {
+        if (workUnitState.getPropAsInt(WORK_UNIT_ORDINAL) > tempWorkUnitOrdinal) {
+          optionalMaxWorkUnit = Optional.of(workUnitState);
+          tempWorkUnitOrdinal = workUnitState.getPropAsInt(WORK_UNIT_ORDINAL);
+        }
+      } catch (NumberFormatException nfe) {
+        throw new RuntimeException(
+            "Work units in state store are corrupted! Missing or malformed " + WORK_UNIT_ORDINAL);
```
erwa: Please include the `nfe` in the `RuntimeException`. Whenever possible, please always include nested exceptions.

`throw new RuntimeException("...message...", nfe)`

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,122 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+    return failedPreviousWorkUnits;
+  }
+
+  /**
+   * Searching for workunit with maximum ordinal.
+   * @param previousWorkUnitStates
+   * @return
+   */
+  private Optional<WorkUnitState> getMaxWorkUnitState(List<WorkUnitState> previousWorkUnitStates) {
+    int tempWorkUnitOrdinal = 0;
```
erwa: Can we call this `maxWorkUnitOrdinalSeen`?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -170,19 +245,44 @@ public DeepIterator(Iterator<Dataset> baseIterator, String previousDatasetUrnWat
     @Override
     protected WorkUnit computeNext() {
       if (this.generatedWorkUnits >= this.maxWorkUnits) {
-        return endOfData();
+        if (retryWorkUnits.hasNext()) {
+          //For the case where retryWorkUnits.size() > maxWorkUnits, we need to keep track of retryWorkunits that aren't
+          //able to fit in a single execution, by committing their state thereby being recognizable in next execution.
```
erwa: Nice comment.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -170,19 +245,44 @@ public DeepIterator(Iterator<Dataset> baseIterator, String previousDatasetUrnWat
...
+          WorkUnit addtionWorkUnit = retryWorkUnits.next();
```
erwa: addtion -> additional

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -170,19 +245,44 @@ public DeepIterator(Iterator<Dataset> baseIterator, String previousDatasetUrnWat
...
+        generatedWorkUnits++;
+        return workUnit;
       }
 
       while (this.baseIterator.hasNext() || this.currentPartitionIterator.hasNext()) {
         if (this.currentPartitionIterator != null && this.currentPartitionIterator.hasNext()) {
           PartitionableDataset.DatasetPartition partition = this.currentPartitionIterator.next();
+          // For partition appeared in retryUrns, just skip.
+          if (retryUrns.contains(partition.getUrn())) {
+            continue;
```
erwa: I think we should add a `log.info()` message here for informational purposes and to help us debug in case we encounter issues.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -170,19 +245,44 @@ public DeepIterator(Iterator<Dataset> baseIterator, String previousDatasetUrnWat
...
           WorkUnit workUnit = workUnitForDatasetPartition(partition);
           addDatasetInfoToWorkUnit(workUnit, partition.getDataset(), this.generatedWorkUnits++);
           addPartitionInfoToWorkUnit(workUnit, partition);
           return workUnit;
         }
 
         Dataset dataset = this.baseIterator.next();
+        // For dataset appeared in retryUrns, just skip.
+        if (retryUrns.contains(dataset.getUrn())) {
+          continue;
```
erwa: Same.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/task/RolloverTask.java
```diff
@@ -0,0 +1,75 @@
...
+   * @return A {@link WorkUnit} that will run a {@link RolloverTask}.
+   */
+  public static WorkUnit rolloverWorkunit() {
+    WorkUnit workUnit = new WorkUnit();
+    TaskUtils.setTaskFactoryClass(workUnit, RolloverTask.Factory.class);
+    return workUnit;
+  }
+
+  /**
+   * ROLLOVERED represents that current task won't be executed while it will be resumed in the next execution if quota
```
erwa: ROLLOVERED -> ROLLOVER

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
autumnust: Is side effect in java always considered as bad practice?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
autumnust: Then it should be <= here.

autumnust: No actually, my mistake. I should add a `currentRetryCount` to ensure that is retry limit is set to 0, no retry will be conducted.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
autumnust: I would like to send this event in our internal MP.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
erwa: Almost all rules have exceptions. I just think as a general rule of thumb, you should avoid writing methods with side effects. This is the main idea behind functional programming, where outputs of a function only depend on inputs and functions have no side effects.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -248,9 +260,9 @@ protected WorkUnit computeNext() {
         if (retryWorkUnits.hasNext()) {
           //For the case where retryWorkUnits.size() > maxWorkUnits, we need to keep track of retryWorkunits that aren't
           //able to fit in a single execution, by committing their state thereby being recognizable in next execution.
-          WorkUnit addtionWorkUnit = retryWorkUnits.next();
-          TaskUtils.setTaskFactoryClass(addtionWorkUnit, RolloverTask.Factory.class);
-          return addtionWorkUnit;
+          WorkUnit addtionalWorkUnit = retryWorkUnits.next();
```
erwa: Missing 'i' in additional
addtional -> additional

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -89,58 +89,70 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
-          log.info("Dataset " + workUnitState.getProp(DATASET_URN) + " is failed previously, retrying...");
+      WorkUnitState.WorkingState workingState = workUnitState.getWorkingState();
+      if (workingState == WorkUnitState.WorkingState.FAILED || workingState == WorkUnitState.WorkingState.ROLLOVER) {
+        int currentRetryCount = 0;
+        if (workUnitState.contains(ConfigurationKeys.TASK_RETRIES_KEY)) {
+          currentRetryCount = workUnitState.getPropAsInt(ConfigurationKeys.TASK_RETRIES_KEY);
+        }
+        if (currentRetryCount < maxRetries) {
+          log.info("Dataset " + workUnitState.getProp(DATASET_URN) + " is failed previously, retrying for the "
+              + currentRetryCount + " times ");
```
erwa: This log message doesn't make much sense if `currentRetryCount` is `0`, which it might be if `TASK_RETRIES_KEY` is not set in the `workUnitState`.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -79,59 +88,133 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+  private List<WorkUnit> getPreviousFailedWorkUnits(List<WorkUnitState> previousWorkUnitStates, int maxRetries) {
+    List<WorkUnit> failedPreviousWorkUnits = new ArrayList<>();
+    for (WorkUnitState workUnitState : previousWorkUnitStates) {
+      WorkUnitState.WorkingState workingState = workUnitState.getWorkingState();
+      if (workingState == WorkUnitState.WorkingState.FAILED || workingState == WorkUnitState.WorkingState.ROLLOVER) {
+        int currentRetryCount = 0;
+        if (workUnitState.contains(ConfigurationKeys.TASK_RETRIES_KEY)) {
+          currentRetryCount = workUnitState.getPropAsInt(ConfigurationKeys.TASK_RETRIES_KEY);
+        }
+        if (currentRetryCount < maxRetries && currentRetryCount > 0 ) {
```
erwa: If TASK_RETRIES_KEY is not set, currentRetryCount will be 0 and the work unit will not be retried. We should remove the `currentRetryCount > 0` part and just update the log.info() message on the next line to not include the `currentRetryCount`.

erwa:
+1

gobblin-api/src/main/java/org/apache/gobblin/configuration/WorkUnitState.java
```diff
@@ -77,6 +77,7 @@
     SUCCESSFUL,
     COMMITTED,
     FAILED,
+    ROLLOVER,
```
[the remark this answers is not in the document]
ibuenros: Actually, do we really need a new `WorkUnitState` for this? Rollover is not a work unit state, it is an identity of the work unit. Why not have it as part of the taskState instead? i.e. use a `NoopWorkUnit` with a key-value `isRollOver = true`, or we could check whether the task factory was rollover.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -216,11 +347,58 @@ private void addPartitionInfoToWorkUnit(WorkUnit workUnit, PartitionableDataset.
...
+
+  /**
+   * Determine if a workunit is rolled over from previous execution or not.
+   *
+   * @param state The workUnit from previous execution or to be inspected.
+   *
+   * @return
+   */
+  private boolean isRollOverWU(State state) {
+    return state.getPropAsBoolean(IS_ROLLOVER_WU);
```
ibuenros: Do you need a default here?

PR #2238, opened 2018-01-10
GOBBLIN-365 : Add lookback days config property for CopyableGlobDatasetFinder

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-365

### Description

This feature adds a lookback days config property for CopyableGlobDatasetFinder to control the number of days to go back for distcp-ing time based data sets.

### Tests
     - My PR adds DateRangeIteratorTest to the package org.apache.gobblin.data.management.copy.

### Commits


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/TimeAwareRecursiveCopyableDataset.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public static class DateRangeIterator implements Iterator {
+    private LocalDateTime startDate;
+    private LocalDateTime endDate;
+    private boolean isDatePatternHourly;
+
+    public DateRangeIterator(LocalDateTime startDate, LocalDateTime endDate, String datePattern) {
+      this.startDate = startDate;
+      this.endDate = endDate;
+      this.isDatePatternHourly = datePattern.contains("HH")? true: false;
```
ibuenros: This is not fully correct, consider the pattern:
```
`hourlyHH`/yyy/MM/dd
```
where `hourlyHH` is a string literal in the pattern (we allow string literals, as sometimes there may be prefixes, infixes, or suffixes in the pattern). This will resolve to true, even though the pattern is not actually hourly.

A different way of achieving this is using `DateTimeFormatter#getResolverFields`, then checking whether any of the resolver fields is an hourly field.

sv2000: Thanks for the feedback! There are 2 points to be mentioned here:
1. Any invalid dateTime pattern will result in an IllegalArgumentException to be thrown.
2. To check if input dateTime pattern has hour granularity, DateTimeFormatter#getResolverFields does not help. I created a helper method that passes a reference dateTime (with non zero hour) through the input pattern and then passing the same dateTime (with zeroed hour) through the same pattern and checking if the two output strings are the same. If not, the input dateTime pattern has hourly granularity.

ibuenros:
+1

PR #2241, opened 2018-01-11
[Gobblin-359]  fix task executor logging

kadaan:
This PR fixes the logging from TaskExecutor.  The issue was that the thread 
pools were not ensuring that the MDCs were being properly cleaned up.  
This resulted in the logs indicating that a specific job was running on the 
thread, when in reality it was a different job.


kadaan:
@abti Can you review?

kadaan:
@htran1 Thanks for merging this!

PR #2246, opened 2018-01-13
[Gobblin-371]  eagerly test for jira-python package

kadaan:
When running the gobblin_pr dev tool, if the jira-python package is not installed, the script will break when you try to resolve a jira issue.  At that point, there is no way to resume the script to resolve the issue.  A simple way to solve this is to move the loading of the jira-python library to the beginning (where click and keyring are loaded).  This ensures that all necessary library are installed before running the script and will keep the script from failing in mid run.


kadaan:
@abti What do you think?

abti:
@kadaan Can you please resolve the conflicts and merge this? It could have saved me some trouble on another Mac .. hehe

PR #2252, opened 2018-01-17
Log job child process exit status

HappyRay:
Also

- log the cmd line in one line for the child process.

- ignore temp directory under the project root.

Testing:

Manually checked the log messages in the log of an integration test.

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleHelixTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleHelixTask.java
@@ -61,8 +61,13 @@ public class SingleHelixTask implements Task {
               this.jobName, this.jobId));
       int exitCode = this.taskProcess.waitFor();
       if (exitCode == 0) {
+        logger.info(String
+            .format("Task process finished. job name: %s. job id: %s", this.jobName, this.jobId));
         return new TaskResult(TaskResult.Status.COMPLETED, "");
       } else {
+        logger.warn(String
+            .format("Task process failed with exitcode (%d). job name: %s. job id: %s", exitCode,
+                this.jobName, this.jobId));
         return new TaskResult(TaskResult.Status.FAILED, "Exit code: " + exitCode);
       }
     } catch (final Throwable t) {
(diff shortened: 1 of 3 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleHelixTask.java
```diff
@@ -61,8 +61,13 @@ public TaskResult run() {
               this.jobName, this.jobId));
       int exitCode = this.taskProcess.waitFor();
       if (exitCode == 0) {
+        logger.info(String
```
htran1: Instead of format, you can use {} like logger.info("Task process finished. job name: {}. job id: {}", this.jobName, this.jobId). This is more efficient since it defers the formatting.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
```diff
@@ -56,6 +56,9 @@ Process launch(final String jobId, final Path workUnitFilePath)
 
     // The -cp parameter list can be very long.
     logger.debug("cmd: " + command);
+    final String completeCmdLine = String.join(" ", command);
+    logger.debug("cmd line: \n" + completeCmdLine);
```
htran1: Same here. Use of {} is more efficient especially since debug usually won't be emitted.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleHelixTask.java
htran1: Instead of format, you can use {} like logger.info("Task process finished. job name: {}. job id: {}", this.jobName, this.jobId). This is more efficient since it defers the formatting.

Same for the warn and debug messages below.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTaskLauncher.java
HappyRay: Done. Thanks

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleHelixTask.java
HappyRay: Done. Thanks

PR #2258, opened 2018-01-23
Don't let Helix retry when a child process fails

HappyRay:


HappyRay:
Thanks. Could you merge?

PR #2280, opened 2018-02-07
[GOBBLIN-405] Fix race condition with access to immediately invalidat…

htran1:
…ed resources

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-405

### Description
There are race conditions in the retrieval of immediately invalidate resources from the SharedResourcesBroker. This can result in an object being returned multiple times even though it should only be returned once. It can also result in valid objects being invalidated before a thread picks it up and it can also result in stack overflow due to a thread starving and retrying many times.

The fix introduces ResourceEntry#getResourceIfValid() to atomically get the resource if it is valid. A null is returned for invalid objects. The recursion in DefaultBrokerCache#getScoped() was replaced with iteration to avoid stack overflow and a lock was added to minimize retries.

### Tests

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/broker/ImmediatelyInvalidResourceEntry.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/broker/ImmediatelyInvalidResourceEntry.java
@@ -36,7 +36,7 @@ public class ImmediatelyInvalidResourceEntry<T> extends ResourceInstance<T> {
   }
 
   @Override
-  public T getResource() {
+  public synchronized T getResource() {
     // mark the object as invalid before returning so that a new one will be created on the next
     // request from the factory
     this.valid = false;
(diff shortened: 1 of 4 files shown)


gobblin-utility/src/main/java/org/apache/gobblin/broker/ImmediatelyInvalidResourceEntry.java
```diff
@@ -36,7 +36,7 @@ public ImmediatelyInvalidResourceEntry(T resource) {
   }
 
   @Override
-  public T getResource() {
+  public synchronized T getResource() {
```
ibuenros: Is this method used anymore? If not, can we deprecate it?

ibuenros:
+1

PR #2284, opened 2018-02-12
[GOBBLIN-407] fix job output path for full snapshot

arjun4084346:
fix job output path for full snapshot

Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @htran1 please review.

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-407

### Description
If the property 'extract.is.full' is set to true, job output should be written to _full directories. This stopped happening since changes in https://github.com/apache/incubator-gobblin/commit/df75b13e1f7ee5776ed18b7aade9014fae8deeea were pushed in, because it made it to ignore this property by not including it while creating 'Extract'.

In this PR, I moved the logic of finding output directory name from Extract to WorkUnitState, as Extract:getIsFull() is marked as deprecated.

### Tests
test case already present

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
@@ -311,11 +311,7 @@ public class WorkUnit extends State {
 
   @Override
   public String getProp(String key) {
-    String value = super.getProp(key);
-    if (value == null) {
-      value = this.extract.getProp(key);
-    }
-    return value;
+    return getProp(key, null);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


gobblin-api/src/main/java/org/apache/gobblin/source/workunit/WorkUnit.java
```diff
@@ -318,6 +318,15 @@ public String getProp(String key) {
     return value;
   }
 
+  @Override
+  public String getProp(String key, String def) {
```
htran1: getProp(String key) can call this one with null for default.

PR #2291, opened 2018-02-21
GOBBLIN-415: Check for the value of configuration key flow.runImmediately in Job config.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-415

### Description
GobblinServiceJobScheduler checks for the existence of the Config Key "flow.runImmediately" instead of its boolean value in deciding whether to schedule a flow for execution. We need to modify this behavior to check for the actual value of the config key provided in the job config.

### Tests
  Change is trivial and ran clean build with unit tests. Also tested the change in local set up by restarting gobblin-service.

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
@@ -28,6 +28,7 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.gobblin.util.PropertiesUtils;
 import org.apache.helix.HelixManager;
 import org.apache.helix.InstanceType;
 import org.quartz.DisallowConcurrentExecution;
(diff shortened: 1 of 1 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -223,8 +223,10 @@ public void onAddSpec(Spec addedSpec) {
           _log.info("Scheduling flow spec: " + addedSpec);
           scheduleJob(jobConfig, null);
           if (jobConfig.containsKey(ConfigurationKeys.FLOW_RUN_IMMEDIATELY)) {
```
htran1: This will be cleaner with fewer indentation levels as `if (PropertiesUtils.getPropAsBoolean(jobConfig, ConfigurationKeys.FLOW_RUN_IMMEDIATELY, "false"))`

sv2000: Thanks for the suggestion! Incorporated this change.

PR #2294, opened 2018-02-23
[GOBBLIN-412] Set compaction's compression params

sushantpande:
Make compaction's compression params configurable through job config

### JIRA
  - https://issues.apache.org/jira/browse/GOBBLIN-412

### Description
 Parameters to control compression-

1. **mapreduce.output.fileoutputformat.compress**
2. **mapreduce.output.fileoutputformat.compress.codec**
3. **mapreduce.output.fileoutputformat.compress.type**

are not passed on to Hadoop from compaction job configuration file. In effect, these parameter's value are always picked up from mapred-site.xml.

Following Three new parameters are introduced as part of the fix to control compression behavior in compaction job

1. **compaction.job.mapreduce.output.fileoutputformat.compress**
2. **compaction.job.mapreduce.output.fileoutputformat.compress.codec**
3. **compaction.job.mapreduce.output.fileoutputformat.compress.type**

### Tests

1. Set _compaction.job.mapreduce.output.fileoutputformat.compress_  and verify output of compaction job. Output should be compressed with default codec. 
2. Reset _compaction.job.mapreduce.output.fileoutputformat.compress_  and verify output of compaction job. Output shouldn't be compressed.
3. Set _compaction.job.mapreduce.output.fileoutputformat.compress_ and set  _compaction.job.mapreduce.output.fileoutputformat.compress_ to _org.apache.hadoop.io.compress.SnappyCodec_ and verify output of compaction job. Output should be compressed with Snappy codec.
4. Set _compaction.job.mapreduce.output.fileoutputformat.compress_ and set  _compaction.job.mapreduce.output.fileoutputformat.compress_ to _org.apache.hadoop.io.compress.DefaultCodec_ and verify output of compaction job. Output should be compressed with Deflate codec.
5. Do not set introduced parameters and remove _compaction.job.mapreduce.output.fileoutputformat.compress_ and _mapred.output.compress_ parameters from mapred-site.xml. Output should be compressed with Deflate codec.
6. Run test case (4) with  _compaction.job.mapreduce.output.fileoutputformat.compress.type_  set to RECORD and verify output of compaction job. Output should be compressed with Deflate codec. Output should be compressed with Snappy codec and record level.
7. Run test case (4) with  _compaction.job.mapreduce.output.fileoutputformat.compress.type_  set to BLOCK and verify output of compaction job. Output should be compressed with Deflate codec. Output should be compressed with Snappy codec and block level.

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
@@ -127,6 +127,11 @@ public abstract class MRCompactorJobRunner implements Runnable, Comparable<MRCom
   public static final String COMPACTION_JOB_USE_PRIME_REDUCERS = COMPACTION_JOB_PREFIX + "use.prime.reducers";
   public static final boolean DEFAULT_COMPACTION_JOB_USE_PRIME_REDUCERS = true;
 
+  // Compression enable, codec and type parameters
+  public static final String MAPREDUCE_OUTPUT_FILEOUTPUTFORMAT_COMPRESS = COMPACTION_JOB_PREFIX + "mapreduce.output.fileoutputformat.compress";
+  public static final String MAPREDUCE_OUTPUT_FILEOUTPUTFORMAT_COMPRESS_CODEC = COMPACTION_JOB_PREFIX + "mapreduce.output.fileoutputformat.compress.codec";
+  public static final String MAPREDUCE_OUTPUT_FILEOUTPUTFORMAT_COMPRESS_TYPE = COMPACTION_JOB_PREFIX + "mapreduce.output.fileoutputformat.compress.type";
+
   public static final String HADOOP_JOB_NAME = "Gobblin MR Compaction";
   private static final long MR_JOB_CHECK_COMPLETE_INTERVAL_MS = 5000;
   private final boolean isRetryEnabled;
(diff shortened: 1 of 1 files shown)


sushantpande:
Please review.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -275,17 +280,30 @@ protected MRCompactorJobRunner(Dataset dataset, FileSystem fs) {
   public void run() {
     Configuration conf = HadoopUtils.getConfFromState(this.dataset.jobProps());
 
-    // Turn on mapreduce output compression by default
-    if (conf.get("mapreduce.output.fileoutputformat.compress") == null && conf.get("mapred.output.compress") == null) {
-      conf.setBoolean("mapreduce.output.fileoutputformat.compress", true);
-    }
+    try {
```
htran1: Why not set the mapreduce.output.fileoutputformat.compress* config directly in the job properties which should be propagated to the Configuration object above?

sushantpande: If property name is same and is defined at both the places, job config and hadoop config, I have observed that property's value from the latter over rides the former. For this reason I have defined new properties by prefixing "compaction.job" to mapreduce.output.fileoutputformat.compress*.

htran1: How are you running the compaction job? Are you using MRCompactionRunner or CompactionRunner?

sushantpande: I use**gobblin-compaction.sh**. with type as "mr".  I pass following class parameters as part of job config file
**compaction.jobprops.creator.class=gobblin.compaction.mapreduce.MRCompactorTimeBasedJobPropCreator
compaction.job.runner.class=gobblin.compaction.mapreduce.avro.MRCompactorAvroKeyDedupJobRunner
compaction.datasets.finder=gobblin.compaction.dataset.TimeBasedSubDirDatasetsFinder**

htran1: That will use the MRCompactionRunner. The main of that class will call CliOptions.parseArgs(). This is where the job config is being overwritten with the hadoop configuration. I think this is wrong. The proper fix is to change parseArgs to only add the hadoop config into the job config if it is not already set.

abti:
@sushantpande @htran1 ping to get traction again on this :)

abti:
@sushantpande @htran1 ^^

SivaMaplelabs:
Resubmitted the PR as per the suggestions.
https://github.com/apache/incubator-gobblin/pull/2386

abti:
@htran1 ^^

PR #2300, opened 2018-03-08
[GOBBLIN-423] Add record count limit to salesforce source

yukuai518:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-423

### Description
   Set fixed upper bound in bootstrapping mode. Don't move the upper bound in the future run.

### Tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/partition/Partitioner.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/partition/Partitioner.java
@@ -53,6 +53,7 @@ public class Partitioner {
   public static final String WATERMARKTIMEFORMAT = "yyyyMMddHHmmss";
   public static final String HAS_USER_SPECIFIED_PARTITIONS = "partitioner.hasUserSpecifiedPartitions";
   public static final String USER_SPECIFIED_PARTITIONS = "partitioner.userSpecifiedPartitions";
+  public static final String EARLY_STOP = "partitioner.earlyStop";
 
   public static final Comparator<Partition> ascendingComparator = new Comparator<Partition>() {
     @Override
@@ -109,7 +110,9 @@ public class Partitioner {
    * Get partitions with low and high water marks
    *
    * @param previousWatermark previous water mark from metadata
-   * @return map of partition intervals
+   * @return map of partition intervals.
+   *         map's key is interval begin time (in format {@link Partitioner#WATERMARKTIMEFORMAT})
+   *         map's value is interval end time (in format {@link Partitioner#WATERMARKTIMEFORMAT})
    */
   @Deprecated
   public HashMap<Long, Long> getPartitions(long previousWatermark) {
@@ -199,6 +202,7 @@ public class Partitioner {
     List<Partition> partitions = new ArrayList<>();
 
     List<String> watermarkPoints = state.getPropAsList(USER_SPECIFIED_PARTITIONS);
+    boolean isEarlyStop = state.getPropAsBoolean(EARLY_STOP);
     if (watermarkPoints == null || watermarkPoints.size() == 0 ) {
       LOG.info("There should be some partition points");
       long defaultWatermark = ConfigurationKeys.DEFAULT_WATERMARK_VALUE;
@@ -234,13 +238,16 @@ public class Partitioner {
     highWatermark = adjustWatermark(watermarkPoints.get(i), watermarkType);
     ExtractType extractType =
         ExtractType.valueOf(this.state.getProp(ConfigurationKeys.SOURCE_QUERYBASED_EXTRACT_TYPE).toUpperCase());
-    if (isFullDump() || isSnapshot(extractType)) {
+
+    // If it is early stop, we should not remove upper bounds
+    if ((isFullDump() || isSnapshot(extractType)) && !(isEarlyStop && isRetriggeringEnabled())) {
       // The upper bounds can be removed for last work unit
       partitions.add(new Partition(lowWatermark, highWatermark, true, false));
     } else {
       // The upper bounds can not be removed for last work unit
       partitions.add(new Partition(lowWatermark, highWatermark, true, true));
     }
+
 
     return partitions;
   }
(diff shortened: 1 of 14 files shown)


yukuai518:
@zxcware @htran1 Please review

gobblin-core/src/main/java/org/apache/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -322,6 +333,17 @@ protected long getLowWatermark(ExtractType extractType, WatermarkType watermarkT
         lowWatermark = this.getAppendLowWatermark(watermarkType, previousWatermark, deltaForNextWatermark);
       }
     }
+
+    if (isBootStrapWithFixedHighWatermark()) {
```
htran1: This looks the same as the else block above, so it can be merged by changing the if to ` if (!isBootStrapWithFixedHighWatermark() && (this.isFullDump() || this.isWatermarkOverride())) {`

gobblin-core/src/main/java/org/apache/gobblin/source/extractor/partition/Partitioner.java
```diff
@@ -407,9 +429,30 @@ protected long getHighWatermark(ExtractType extractType, WatermarkType watermark
...
     }
+
+    // Use fixed high watermark in bootstrapping mode
+    if (isBootStrapWithFixedHighWatermark()) {
+      hasUserSpecifiedHighWatermark = true;
+      if (this.state.contains(ConfigurationKeys.SOURCE_QUERYBASED_END_VALUE)) {
+        highWatermark = this.state.getPropAsLong(ConfigurationKeys.SOURCE_QUERYBASED_END_VALUE);
+        LOG.info("Bootstrapping mode: persisted high watermark " + highWatermark);
+      } else {
+        this.state.setProp(ConfigurationKeys.SOURCE_QUERYBASED_END_VALUE, highWatermark);
```
htran1: Is this state getting persisted and propagated to the next run? If it does then what happens when user changes the configuration? Will the config change override the state?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -79,6 +79,11 @@
   public static final boolean DEFAULT_USE_ALL_OBJECTS = false;
 
   private static final String ENABLE_DYNAMIC_PARTITIONING = "salesforce.enableDynamicPartitioning";
+  private static final String DYNAMIC_PROBING_TOTAL_RECORDS_LIMIT = "salesforce.dynamicProbing.total.records.limit";
```
htran1: Camel case names should be used here.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -131,7 +136,19 @@
...
 
-    String specifiedPartitions = generateSpecifiedPartitions(histogram, minTargetPartitionSize, maxPartitions,
+    // we should look if the count is too big, cut off early if count exceeds the limit, or bucket size is too large
+    Histogram histogramActual = new Histogram();
+    for (HistogramGroup group: histogram.getGroups()) {
+      if (histogramActual.getTotalRecordCount() + group.count > state.getPropAsLong(DYNAMIC_PROBING_TOTAL_RECORDS_LIMIT, DYNAMIC_PROBING_DEFAULT_TOTAL_RECORDS_LIMIT)) {
+        break;
+      }
+      histogramActual.add(group);
+      if (histogramActual.getGroups().size() >= state.getPropAsLong(DYNAMIC_PROBING_TOTAL_BUCKETS_LIMIT, DYNAMIC_PROBING_DEFAULT_TOTAL_BUCKETS_LIMIT)) {
```
htran1: Why have the bucket limit here?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -131,7 +136,19 @@
     Partition partition = new Partitioner(state).getGlobalPartition(previousWatermark);
     Histogram histogram = getHistogram(sourceEntity.getSourceEntityName(), watermarkColumn, state, partition);
 
-    String specifiedPartitions = generateSpecifiedPartitions(histogram, minTargetPartitionSize, maxPartitions,
+    // we should look if the count is too big, cut off early if count exceeds the limit, or bucket size is too large
+    Histogram histogramActual = new Histogram();
```
htran1: We should consider pushing this logic into getHistogram() for early termination of the search or add a note here indicating possible future work.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
yukuai518: @htran1 do we have total query limits? I think each bucket is a single query, if we have total limits for querying count, we can set this bucket number.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
yukuai518: @htran1 do you mean you want to change other attributes as well? this is constant, why need Camel case?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
htran1: The query limit is enforced by `salesforce.dynamicProbingLimit`. The loop here doesn't execute any queries, so the limit is not required here.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -504,6 +534,14 @@ void add(Histogram histogram) {
       totalRecordCount += histogram.totalRecordCount;
     }
 
+    HistogramGroup get(int idx) {
```
htran1: Why is this required? I don't see anywhere where the null result is checked.

yukuai518: I use this to get the latest group.

PR #2302, opened 2018-03-12
Partitionaware Hive Registration Policy

treff7es:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-425

### Description
This is the partition aware implementation of {@link HiveRegistrationPolicyBase}.
You can specify hive.partition.regex where the first match is the table location and the rest match are the partitions.
The partition names can be specified with the hive.table.partition.keys property.
The order of the partition names should match the order of regexp matches in the hive.partition.regexp expression.

 For example in the case of path: s3://testbucket/myawesomlogs/compacted/dt=20170101/hr=22/
 The hive.partition.regexp would look like: hive.partition.regex=(s3://testbucket/myawesomelogs/compacted/)dt=(.*)/hr=(.*) and the hive.table.partition.keys=dt,hr

### Tests

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/policy/PartitionAwareHiveRegistrationPolicy.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/policy/PartitionAwareHiveRegistrationPolicy.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.hive.policy;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
+
+import org.apache.gobblin.configuration.State;
+import org.apache.gobblin.hive.HivePartition;
+import org.apache.gobblin.hive.HiveRegistrationUnit;
+import org.apache.gobblin.hive.HiveTable;
+
+
+/**
+ * This is the partition aware implementation of {@link HiveRegistrationPolicyBase}.
+ * You can specify hive.partition.regex where the first match is the table location and the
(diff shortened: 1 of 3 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/policy/PartitionAwareHiveRegistrationPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+  protected Path getTableLocation(Path path) {
+    Optional<String> locationRegex = Optional.fromNullable(this.props.getProp(HIVE_PARTITION_REGEX));
+    if (locationRegex.isPresent()) {
+      Pattern pattern = Pattern.compile(locationRegex.get());
+      Matcher matcher = pattern.matcher(path.toString());
+      if (matcher.matches()) {
+        String location = matcher.group(1);
+        return new Path(location);
+      }else {
+        return path;
```
autumnust: Why if the pattern is not matched, we should return the original path here?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/policy/PartitionAwareHiveRegistrationPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * This is the partition aware implementation of {@link HiveRegistrationPolicyBase}.
+ * You can specify hive.partition.regex where the first match is the table location and the
+ * rest match are the partitions.
+ * The partition names can be specified with the hive.table.partition.keys property.
+ * The order of the partition names should match the order of regexp matches in the hive.partition.regexp expression.
+ *
+ * For example in the case of path: s3://testbucket/myawesomlogs/compacted/dt=20170101/hr=22/
+ * The hive.partition.regexp would look like: hive.partition.regex=(s3://testbucket/myawesomelogs/compacted/)dt=(.*)/hr=(.*)
```
autumnust: I wonder if there are no recognizable key in the partition identifier (Like what you have here `dt`, `hr`), what should we do about it ?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/policy/PartitionAwareHiveRegistrationPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+
+    if (this.props.contains(HIVE_PARTITION_REGEX)) {
+      Pattern pattern = Pattern.compile(this.props.getProp(HIVE_PARTITION_REGEX));
+      List<String> partitionValues = Lists.newArrayList();
+
+      Matcher matcher = pattern.matcher(path.toString());
+      if (matcher.matches() && matcher.groupCount() >=2){
+        for (int i = 2; i <= matcher.groupCount(); i++) {
+          partitionValues.add(matcher.group(i));
+        }
```
deepak-batra: Hey Tamas, 
As we're specifying partition keys, shouldn't we iterate on them to make sure all the partition key values are filled. For ex.: 
say 
`partition.keys` = dt, hr, min
and the regex is = (some-path)/dt=*/hr=* (something like this). `
Although the table will be registered with different partition keys, we would pass lesser number of values to register a partition which would lead an exception from HiveMetastore's thrift API. However, if we do it the other way, then the HivePartitionBuilder would fail quite early as compared to the other fail scenario.

treff7es: Make sense I will fix this and fail early.

abti:
@treff7es @autumnust ^^

treff7es:
@abti Fyi, I have updated this pull request, sorry for the delay.

PR #2313, opened 2018-03-23
Always  mark custom tasks as complete even if they throw exception in run()

eogren:
If a custom task throws an exception in the run() method, it was not getting marked as complete which would result in the job hanging. Change the task wrapper to mark tasks complete in a finally block.


eogren:
@ibuenros Can you take a look?

PR #2336, opened 2018-04-12
GOBBLIN-464: Enhance LoopingDatasetFinderSource to support global watermark and per-dataset watermark.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-464

### Description
We need the ability to support a global watermark that spans all datasets as well as a per-dataset watermark in LoopingDatasetFinderSource. In particular we need to keep track of:

The last dataset processed in the previous run of a gobblin job (global watermark) so that the subsequent run can process the "next" dataset (based on some ordering between datasets), and
The last time a particular dataset is processed (per-dataset watermark).
The actual use case is to limit the time range of queries to an external database based on the last processed time of a dataset. 

### Tests

### Commits


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -83,7 +97,8 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
   public WorkUnitStream getWorkunitStream(SourceState state) {
     try {
       int maxWorkUnits = state.getPropAsInt(MAX_WORK_UNITS_PER_RUN_KEY, MAX_WORK_UNITS_PER_RUN);
-
+      Preconditions.checkArgument(maxWorkUnits > 0);
```
ibuenros: Can you put a meaningful message in case this check is not true?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -83,7 +97,8 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
...
+      String globalWatermarkDatasetUrn = state.getProp(ConfigurationKeys.GLOBAL_WATERMARK_DATASET_URN, "");
```
ibuenros: Why is this user settable?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/JobContext.java
```diff
@@ -146,12 +152,23 @@ public JobContext(Properties jobProps, Logger logger, SharedResourcesBroker<Gobb
 
     this.datasetStateStore = createStateStore(ConfigUtils.propertiesToConfig(jobProps));
     this.jobHistoryStoreOptional = createJobHistoryStore(jobProps);
+    this.globalWatermarkDatasetUrn = jobProps.containsKey(ConfigurationKeys.GLOBAL_WATERMARK_DATASET_URN) ? Optional
```
ibuenros: Why is this part of `JobContext`, what is the intended behavior as a framework level flag?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -188,39 +232,111 @@ protected WorkUnit computeNext() {
...
-    if (spliterator.hasCharacteristics(Spliterator.SORTED) &&
-        spliterator.getComparator().equals(this.lexicographicalComparator)) {
+    if (spliterator.hasCharacteristics(Spliterator.SORTED) && spliterator.getComparator()
+        .equals(this.lexicographicalComparator)) {
       return StreamSupport.stream(spliterator, false);
     }
     return StreamSupport.stream(spliterator, false).sorted(this.lexicographicalComparator);
   }
 
+  protected DatasetStateStore createStateStore(Config jobConfig)
```
ibuenros: Can't you use `DatasetStateStore.buildDatasetStateStore(Config)` instead?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -188,39 +232,111 @@ protected WorkUnit computeNext() {
...
       workUnit.setProp(PARTITION_URN, partition.getUrn());
     }
+
+    private void addWatermarkIntervalToWorkUnit(WorkUnit workUnit, Dataset dataset) {
+      if (!Strings.isNullOrEmpty(globalWatermarkDatasetUrn)) {
+        addWatermarkIntervalToWorkUnit(workUnit, dataset, null);
+      }
+    }
+
+    private void addWatermarkIntervalToWorkUnit(WorkUnit workUnit, Dataset dataset,
```
ibuenros: This only works for long watermarks. Would it be better to pass the previously run `WorkUnit` to `workUnitForDataset`/`workUnitForDatasetPartition` for the subclass to decide what to do with it?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -125,13 +145,24 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
 
     private Iterator<PartitionableDataset.DatasetPartition> currentPartitionIterator;
     private int generatedWorkUnits = 0;
+    private String globalWatermarkDatasetUrn;
```
ibuenros: It seems that this flag is a proxy for "should we use dataset state store". Why not use a `useDatasetStateStore` flag instead? The `globalWatermarkDatasetUrn` in general need not be modified by the user.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -105,11 +120,16 @@ public WorkUnitStream getWorkunitStream(SourceState state) {
...
+          datasetsFinder.getDatasetsStream(Spliterator.SORTED, this.lexicographicalComparator);
       datasetStream = sortStreamLexicographically(datasetStream);
 
-      return new BasicWorkUnitStream.Builder(new DeepIterator(datasetStream.iterator(), previousDatasetUrnWatermark,
-          previousPartitionUrnWatermark, maxWorkUnits)).setFiniteStream(true).build();
+      Config config = ConfigUtils.propertiesToConfig(state.getProperties());
+      String storeName = JobState.getJobNameFromState(state);
+
+      return new BasicWorkUnitStream.Builder(
+          new DeepIterator(datasetStream.iterator(), previousDatasetUrnWatermark, previousPartitionUrnWatermark,
```
ibuenros: `previousDatasetUrnWatermark` is still determined from the largest ordinal in the previous run? Doesn't this need to be updated?

sv2000: Just wanted to leave a quick response here. When datasetstatestore is enabled, the ordinal is effectively useless, since there will be only one workunit passed to the source (i.e. the globalwatermark dataset) in the getPreviourWorkUnitStates() call, which will end up as the maxWorkUnit. The intent was to make minimal changes to allow enabling/disabling dataset state store.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/CombinedWorkUnitAndDatasetStateGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+
+import org.apache.gobblin.configuration.CombinedWorkUnitAndDatasetState;
+import org.apache.gobblin.configuration.CombinedWorkUnitAndDatasetStateCallable;
+import org.apache.gobblin.configuration.WorkUnitState;
+import org.apache.gobblin.metastore.DatasetStateStore;
+
+public class CombinedWorkUnitAndDatasetStateGenerator implements CombinedWorkUnitAndDatasetStateCallable<CombinedWorkUnitAndDatasetState> {
```
ibuenros: Please add javadoc.

gobblin-api/src/main/java/org/apache/gobblin/configuration/SourceState.java
```diff
@@ -153,19 +218,35 @@ public SourceState getPreviousDatasetState(String datasetUrn) {
...
-        previousWorkUnitStatesByDatasetUrns.put(datasetUrn, Lists.<WorkUnitState> newArrayList());
+        previousWorkUnitStatesByDatasetUrns.put(datasetUrn, Lists.newArrayList());
       }
       ((List<WorkUnitState>) previousWorkUnitStatesByDatasetUrns.get(datasetUrn)).add(workUnitState);
     }
 
     return ImmutableMap.copyOf(previousWorkUnitStatesByDatasetUrns);
   }
 
+  private void materializeWorkUnitAndDatasetStates(String datasetUrn) {
```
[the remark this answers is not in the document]
sv2000: Yes that is correct! In fact, I had it as synchronized. I think I may have forgotten to do a git push after the change.

gobblin-api/src/main/java/org/apache/gobblin/configuration/CombinedWorkUnitAndDatasetStateCallable.java
```diff
@@ -0,0 +1,27 @@
...
+package org.apache.gobblin.configuration;
+
+import java.util.concurrent.Callable;
+
+/**
+ * An interface that extends {@link java.util.concurrent.Callable}. Used to pass datasetUrn as an argument to
+ * the Callable.
+ */
+public interface CombinedWorkUnitAndDatasetStateCallable<T extends CombinedWorkUnitAndDatasetState> extends Callable {
+  public CombinedWorkUnitAndDatasetState getCombinedWorkUnitAndDatasetState(String datasetUrn) throws Exception;
```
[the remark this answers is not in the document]
sv2000: Added comments to specify the behavior.

gobblin-api/src/main/java/org/apache/gobblin/configuration/SourceState.java
```diff
@@ -64,19 +65,34 @@
...
+  private List<WorkUnitState> previousWorkUnitStates = Lists.newArrayList();
 
-  @Getter @Setter
+  @Getter
+  @Setter
   private SharedResourcesBroker<GobblinScopeTypes> broker;
 
+  @Getter
+  @Setter
+  private CombinedWorkUnitAndDatasetStateCallable<CombinedWorkUnitAndDatasetState> callable;
```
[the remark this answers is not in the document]
sv2000: Done.

gobblin-api/src/main/java/org/apache/gobblin/configuration/SourceState.java
```diff
@@ -153,19 +218,35 @@ public SourceState getPreviousDatasetState(String datasetUrn) {
...
+        previousWorkUnitStatesByDatasetUrns.put(datasetUrn, Lists.newArrayList());
       }
       ((List<WorkUnitState>) previousWorkUnitStatesByDatasetUrns.get(datasetUrn)).add(workUnitState);
     }
 
     return ImmutableMap.copyOf(previousWorkUnitStatesByDatasetUrns);
   }
 
+  private void materializeWorkUnitAndDatasetStates(String datasetUrn) {
+    if (this.previousWorkUnitStates.isEmpty()) {
```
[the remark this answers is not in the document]
sv2000: Thanks! Added a flag to check if workunit states have already been materialized.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/CombinedWorkUnitAndDatasetStateGenerator.java
sv2000: Done.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/source/LoopingDatasetFinderSource.java
```diff
@@ -81,19 +87,26 @@ public LoopingDatasetFinderSource(boolean drilldownIntoPartitions) {
 
   @Override
   public WorkUnitStream getWorkunitStream(SourceState state) {
+    return this.getWorkunitStream(state,false);
+  }
+
+  public WorkUnitStream getWorkunitStream(SourceState state, boolean isDatasetStateStoreEnabled) {
```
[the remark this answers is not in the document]
sv2000: The Source implementation will call this method with datasetStateStoreEnabled to true/false.

ibuenros:
+1

PR #2342, opened 2018-04-20
[GOBBLIN-469] Add Task for running the DatasetCleaner

htran1:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-469

### Description
The DatasetCleaner currently can only be invoked as a CLI app or Azkaban job. Make a task to wrap it so it can be run as a Gobblin job. This allows dataset state store retention jobs to be scheduled gobblin standalone and standalone cluster.

### Tests

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/source/DatasetCleanerSource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/source/DatasetCleanerSource.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


htran1:
@ibuenros, please review.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/source/DatasetCleanerSource.java
```diff
@@ -0,0 +1,106 @@
...
+public class DatasetCleanerSource implements Source<Object, Object> {
+  public static final String DATASET_CLEANER_SOURCE_PREFIX = "datasetCleanerSource";
+
+  /**
+   * This config holds a list of configuration names. Each configuration name defines a job that can have scoped config.
+   *
+   * So the list "config1, config2" means that two jobs are configured.
+   *
+   * Then the config can be scoped like:
+   * datasetCleanerSource.config1.state.store.db.table=state_table1
```
zxcware: Can we use a `DatasetFinder` to generalize the process of finding datasets to clean? For this case, we can create a `ConfigBasedDatasetFinder`. It makes the `DatasetCleanerSource` more general. @jack-moseley has done similar work for limited retention.

htran1: This is not at the dataset level, but instead is configuration the connection to the state store to clean. A state store stores states for many datasets. For the current use cases of this a config store is not used. Will revisit this later if required.

PR #2361, opened 2018-05-11
GOBBLIN-491: Create a FlowGraph representation for multi-hop support in Gobblin-as-a-Service.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-491

### Description
To enable multi-hop routing in GaaS, we need an underlying graph data structure over which paths for data movement get computed. We define basic interfaces and classes to represent the data flow graph.

### Tests
Added a unit test for DAG representation. More unit tests to follow.

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/BaseHdfsDatasetDescriptor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/BaseHdfsDatasetDescriptor.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.dataset;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.typesafe.config.Config;
+
+import org.apache.gobblin.annotation.Alpha;
+import org.apache.gobblin.service.ServiceConfigKeys;
+import org.apache.gobblin.service.modules.flowgraph.DatasetDescriptorConfigKeys;
+import org.apache.gobblin.util.ConfigUtils;
+
+import lombok.Getter;
+
+
+/**
+ * An implementation of {@link HdfsDatasetDescriptor}.
+ */
+@Alpha
+public class BaseHdfsDatasetDescriptor implements HdfsDatasetDescriptor {
+  public static final String HDFS_PLATFORM_NAME = "hdfs";
+
+  @Getter
+  private final String path;
+  @Getter
+  private final String format;
+  @Getter
+  private final String description;
+  @Getter
+  private final String platform = HDFS_PLATFORM_NAME;
+
+  public BaseHdfsDatasetDescriptor(Config config) {
+    Preconditions.checkArgument(config.hasPath(DatasetDescriptorConfigKeys.PATH_KEY), String.format("Missing required property %s", DatasetDescriptorConfigKeys.PATH_KEY));
+    Preconditions.checkArgument(config.hasPath(DatasetDescriptorConfigKeys.FORMAT_KEY), String.format("Missing required property %s", DatasetDescriptorConfigKeys.FORMAT_KEY));
+
+    this.path = ConfigUtils.getString(config, DatasetDescriptorConfigKeys.PATH_KEY, null);
+    this.format = ConfigUtils.getString(config, DatasetDescriptorConfigKeys.FORMAT_KEY, null);
+    this.description = ConfigUtils.getString(config, DatasetDescriptorConfigKeys.DESCRIPTION_KEY, "");
+  }
+
+  /**
+   * A {@link HdfsDatasetDescriptor} is compatible with another {@link DatasetDescriptor} iff they have identical
+   * platform, type, path, and format.
+   * TODO: Currently isCompatibleWith() only checks if HDFS paths described by the two {@link DatasetDescriptor}s
+   * being compared are identical. Need to enhance this for the case of where paths can contain glob patterns.
+   * e.g. paths described by the pattern /data/input/* are a subset of paths described by /data/* and hence, the
+   * two descriptors should be compatible.
+   * @return true if this {@link HdfsDatasetDescriptor} is compatibe with another {@link DatasetDescriptor}.
+   */
+  @Override
+  public boolean isCompatibleWith(DatasetDescriptor o) {
+    return this.equals(o);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    HdfsDatasetDescriptor other = (HdfsDatasetDescriptor) o;
+    if(this.getPlatform() == null || other.getPlatform() == null) {
+      return false;
+    }
+    if(!this.getPlatform().equalsIgnoreCase(other.getPlatform()) || !(o instanceof HdfsDatasetDescriptor)) {
+      return false;
+    }
+
+    return this.getPath().equals(other.getPath()) && this.getFormat().equalsIgnoreCase(other.getFormat());
+  }
+
+  @Override
+  public String toString() {
+     return "(" + Joiner.on(",").join(this.getPlatform(),this.getPath(),this.getFormat()) + ")";
+  }
+
+  @Override
+  public int hashCode() {
+    return this.toString().hashCode();
+  }
+}
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.dataset;
+
+import org.apache.gobblin.annotation.Alpha;
+
+
+/**
+ * The interface for dataset descriptors.
+ */
+@Alpha
+public interface DatasetDescriptor {
+  /**
+   * @return the dataset platform i.e. the storage backing the dataset (e.g. HDFS, JDBC, Espresso etc.)
+   */
+  public String getPlatform();
+
+  /**
+   * @return a human-readable description of the dataset.
+   */
+  public String getDescription();
+
+  /**
+   * @return true if this {@link DatasetDescriptor} is compatible with the other {@link DatasetDescriptor} i.e. the
+   * datasets described by this {@link DatasetDescriptor} is a subset of the datasets described by the other {@link DatasetDescriptor}.
+   * This check is non-commutative.
+   */
+  public boolean isCompatibleWith(DatasetDescriptor other);
+}
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/HdfsDatasetDescriptor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/HdfsDatasetDescriptor.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.dataset;
+
+import org.apache.gobblin.annotation.Alpha;
+
+
+/**
+ * A descriptor interface for HDFS datasets
+ */
+@Alpha
+public interface HdfsDatasetDescriptor extends DatasetDescriptor {
+  /**
+   *
+   * @return dataset path.
+   */
+  public String getPath();
+
+  /**
+   *
+   * @return storage format of the dataset.
+   */
+  public String getFormat();
+
+}
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowEdge.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowEdge.java
@@ -1,0 +1,191 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.flowgraph;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.hadoop.security.UserGroupInformation;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+import com.typesafe.config.Config;
+
+import org.apache.gobblin.annotation.Alpha;
+import org.apache.gobblin.runtime.api.SpecExecutor;
+import org.apache.gobblin.service.modules.template_catalog.FSFlowCatalog;
+import org.apache.gobblin.service.modules.template.FlowTemplate;
+import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
+import org.apache.gobblin.util.ConfigUtils;
+
+import joptsimple.internal.Strings;
+import lombok.Getter;
+
+
+/**
+ * An implementation of {@link FlowEdge}.
+ */
+@Alpha
+public class BaseFlowEdge implements FlowEdge {
+  public static final String FLOW_EDGE_LABEL_JOINER_CHAR = ":";
+
+  @Getter
+  protected List<String> endPoints;
+
+  @Getter
+  protected FlowTemplate flowTemplate;
+
+  @Getter
+  private List<SpecExecutor> executors;
+
+  @Getter
+  private Config props;
+
+  @Getter
+  private String id;
+
+  @Getter
+  private boolean active;
+
+  //Constructor
+  public BaseFlowEdge(List<String> endPoints, String edgeName, FlowTemplate flowTemplate, List<SpecExecutor> executors, Config properties, boolean active) {
+    this.endPoints = endPoints;
+    this.flowTemplate = flowTemplate;
+    this.executors = executors;
+    this.active = active;
+    this.props = properties;
+    this.id = generateEdgeId(endPoints, edgeName);
+  }
+
+  @Override
+  public boolean isAccessible(UserGroupInformation user) {
+    return true;
+  }
+
+  @VisibleForTesting
+  protected static String generateEdgeId(List<String> endPoints, String edgeName) {
+    return Joiner.on(FLOW_EDGE_LABEL_JOINER_CHAR).join(endPoints.get(0), endPoints.get(1), edgeName);
+  }
+  /**
+   *   The {@link FlowEdge}s are the same if they have the same endpoints and both refer to the same {@FlowTemplate} i.e.
+   *   the {@link FlowTemplate} uris are the same
+   */
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    FlowEdge that = (FlowEdge) o;
+
+    if(!(this.getEndPoints().get(0).equals(that.getEndPoints().get(0))) && ((this.getEndPoints().get(1)).equals(that.getEndPoints().get(1)))) {
+      return false;
+    }
+
+    if(!this.getFlowTemplate().getUri().equals(that.getFlowTemplate().getUri())) {
+      return false;
+    }
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
+    return this.id.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    return this.id;
+  }
+
+  /**
+   * A {@link FlowEdgeFactory} for creating {@link BaseFlowEdge}.
+   */
+  public static class Factory implements FlowEdgeFactory {
+
+    /**
+     * A method to return an instance of {@link BaseFlowEdge}. The method performs all the validation checks
+     * and returns
+     * @param edgeProps Properties of edge
+     * @param flowCatalog Flow Catalog used to retrieve {@link FlowTemplate}s.
+     * @return a {@link BaseFlowEdge}
+     */
+    @Override
+    public FlowEdge createFlowEdge(Config edgeProps, FSFlowCatalog flowCatalog) throws FlowEdgeCreationException {
+      try {
+        String source = ConfigUtils.getString(edgeProps, FlowGraphConfigurationKeys.FLOW_EDGE_SOURCE_KEY,"");
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(source), "A FlowEdge must have a non-null or empty source");
+        String destination = ConfigUtils.getString(edgeProps, FlowGraphConfigurationKeys.FLOW_EDGE_DESTINATION_KEY,"");
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(destination), "A FlowEdge must have a non-null or empty destination");
+        List<String> endPoints = Lists.newArrayList(source, destination);
+        String edgeName = ConfigUtils.getString(edgeProps, FlowGraphConfigurationKeys.FLOW_EDGE_NAME_KEY,"");
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(edgeName), "A FlowEdge must have a non-null or empty name");
+
+        List<Config> specExecutorConfigList = new ArrayList<>();
+        boolean flag;
+        for(int i = 0; (flag = edgeProps.hasPath(FlowGraphConfigurationKeys.FLOW_EDGE_SPEC_EXECUTORS_KEY + "." + i)) != false; i++) {
+          specExecutorConfigList.add(edgeProps.getConfig(FlowGraphConfigurationKeys.FLOW_EDGE_SPEC_EXECUTORS_KEY + "." + i));
+        }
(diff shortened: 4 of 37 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * @return the {@link FlowTemplate} that performs the data movement along the edge.
+   */
+  FlowTemplate getFlowTemplate();
+
+  /**
+   *
+   * @return a list of {@link SpecExecutor}s that can execute the {@link FlowTemplate} corresponding to this edge.
+   */
+  List<SpecExecutor> getExecutors();
```
zxcware: Similar reasoning as in `DataNode`, I would suggest move this out of `FlowEdge`. The association could be moved to `FlowGaph`.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
```diff
@@ -0,0 +1,74 @@
...
+   * @return true if the {@link FlowEdge} is active.
+   */
+  boolean isActive();
+
+  /**
+   *
+   * @param user
+   * @return true if the user has ACL permissions to access the {@link FlowEdge},
+   */
+  boolean isAccessible(UserGroupInformation user);
```
zxcware: We should make it `boolean isAccessible()`? Using `UserGroupInformation` is an implementation detail and is only one of the possible implementations.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.gobblin.service.modules.flowgraph;
+
+import java.util.Collection;
+
+
+public interface FlowGraph {
+  public Collection<DataNode> getNodes();
+
+  public void addNode(DataNode node);
```
zxcware: `removeNode`, `removeEdge` are also necessary

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/Dag.java
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Concatenate two dags together. Join the "other" dag to "this" dag and return "this" dag.
+   * @return the concatenated dag
+   */
+  public Dag<T> concatenate(Dag<T> other) throws IOException {
+    if (other == null) {
+      return this;
+    }
+    for (DagNode node : this.endNodes) {
```
abti: why not a similar logic as line 64? why proactively put all endNodes as parent?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraphTest.java
```diff
@@ -0,0 +1,163 @@
...
+    Assert.assertEquals(((BaseDataNode)graph.getNode("node1")).getFlowEdges().size(), 1);
+    Assert.assertEquals(((BaseDataNode)graph.getNode("node2")).getFlowEdges().size(), 1);
+    Assert.assertEquals(((BaseDataNode)graph.getNode("node3")).getFlowEdges().size(), 1);
+
+    Assert.assertTrue(((BaseDataNode)graph.getNode("node1")).getFlowEdges().contains(edge1));
+    Assert.assertTrue(((BaseDataNode)graph.getNode("node2")).getFlowEdges().contains(edge2));
+    Assert.assertTrue(((BaseDataNode)graph.getNode("node3")).getFlowEdges().contains(edge3));
+  }
+
+  @Test
```
abti: Add test dependency on testAddFlowEdge(), else this can execute before the testAddFlowEdge when parallelism is enabled, leading to flaky tests.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraphTest.java
```diff
@@ -0,0 +1,163 @@
...
+  public void testDeleteDataNode() throws Exception {
+    //Delete node1 from graph
+    graph.deleteDataNode(node1);
```
abti: what happens if we delete all datanodes corresponding to a flowedge, should flowedge be cleaned via some algorithm like mark and sweep?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
```diff
@@ -0,0 +1,74 @@
...
+ *   <li> a list of {@link SpecExecutor}s where the {@link FlowTemplate} can be executed.
+ * </ul></p> and
+ *
+ */
+public interface FlowEdge {
+  /**
+   *
+   * @return the {@link DataNode} ids that are the end points of the edge.
+   */
+  List<String> getEndPoints();
```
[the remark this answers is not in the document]
sv2000: Each DataNode is uniquely identified by a string id. The direction of the flow is implicit. All edges are unidirectional. So the first element in the list is the source and second the destination.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
sv2000: I agree. Right now it is a placeholder method. The signature will likely change in the future.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DataNode.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * @return true if the {@link DataNode} is active
+   */
+  boolean isActive();
+
+  /**
+   * @return a list of {@link FlowEdge}s incident on the {@link DataNode}.
+   */
+  Collection<FlowEdge> getFlowEdges();
```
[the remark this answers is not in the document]
sv2000: Fair point. I also think graph mutations such as addition/deletion of edges have cleaner implementations when FlowEdges are part of FlowGraph as opposed to DataNode. I made this change.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DataNode.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Add a {@link FlowEdge} to a node.
+   */
+  void addFlowEdge(FlowEdge edge);
```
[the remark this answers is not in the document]
sv2000: Same comment as above.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
sv2000: Yes, I have the implementations now.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
sv2000: Actually, SpecExecutors are associated with FlowEdge i.e. it represents executor instance where the flow corresponding to the edge can be executed. Moving it to FlowGraph will need yet another map to maintain association between FlowEdge and SpecExecutor, which needs to be updated with every graph mutation.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowEdge.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  public static class Builder {
+    private List<String> endPoints;
+    private FlowTemplate template;
+    private List<SpecExecutor> executors;
+    private boolean active;
+    private Config properties;
+
+    public Builder withEndPoints(String sourceNode, String destNode) {
```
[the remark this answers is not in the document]
sv2000: Got rid of the Builder. Replaced it with a Factory class which provides a method for creating FlowEdge from a Properties object.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/HdfsDatasetDescriptor.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.dataset;
+
+import org.apache.gobblin.annotation.Alpha;
+
+
+/**
+ * A descriptor interface for HDFS datasets
```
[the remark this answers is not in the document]
sv2000: Good point. Its a "to-do" item.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+public interface DatasetDescriptor {
+  /**
+   * @return the dataset platform i.e. the storage backing the dataset (e.g. HDFS, JDBC, Espresso etc.)
+   */
+  public String getPlatform();
+
+  /**
+   * @return the type of dataset descriptor.
+   */
+  public String getType();
```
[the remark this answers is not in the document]
sv2000: Typically, the type would be the class name itself. However, individual implementations may override this e.g. in the future we may have a TimeStampBasedHdfsDatasetDescriptor implementation which declares its type to be a HdfsDatasetDescriptor.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * The interface for dataset descriptors.
+ */
+@Alpha
+public interface DatasetDescriptor {
+  /**
+   * @return the dataset platform i.e. the storage backing the dataset (e.g. HDFS, JDBC, Espresso etc.)
+   */
+  public String getPlatform();
```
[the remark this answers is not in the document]
sv2000: Yes, datasets on different platforms have different urns, which is consistent with TMS.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraphTest.java
sv2000: Done.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraphTest.java
sv2000: Deletion of a datanode deletes all outgoing edges out of that node.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/BaseHdfsDatasetDescriptor.java
```diff
@@ -0,0 +1,101 @@
...
+   */
+  @Override
+  public boolean isCompatibleWith(DatasetDescriptor o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    if(!this.getPlatform().equalsIgnoreCase(o.getPlatform()) || !this.getType().equalsIgnoreCase(o.getType())) {
```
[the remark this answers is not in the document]
sv2000: Thanks for the comment! Added null checks.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/Dag.java
sv2000: The Dag concatenation is ensuring that all the jobs of the previous edge (which may have multiple end nodes) are completed before starting any job of this edge. The logic is adding the endNodes of previous edge as dependencies of the startNodes of the next edge.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+   * @param edgeId label of the edge to be removed
+   * @return true if edge is removed from the {@link FlowGraph} successfully.
+   */
+  public boolean deleteFlowEdge(String edgeId);
+
+  /**
+   * Get the factory class for creating {@link DataNode}s to be added to the {@link FlowGraph}.
+   * @return a {@link DataNodeFactory} implementation.
+   */
+  public DataNodeFactory getDataNodeFactory();
```
zxcware: `DataNode`s are quite heterogeneous in a graph. Different `DataNode` will be created in different ways. We should consider removing this behavior.

Managing the creation of different DataNodes beyonds the responsiblily of a graph, which only manages the in-and-out of a node and discovers path

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Get the factory class for creating {@link FlowEdge}s to be added to the {@link FlowGraph}.
+   * @return a {@link FlowEdgeFactory} implementation.
+   */
+  public FlowEdgeFactory getFlowEdgeFactory();
```
zxcware: Same as above

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Get a collection of edges adjacent to a {@link DataNode}.
+   * @param nodeId identifier of the {@link DataNode}
+   * @return a collection of edges adjacent to the {@link DataNode}
+   */
+  public Collection<FlowEdge> getEdges(String nodeId);
```
zxcware: What's its use case?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Get a collection of edges adjacent to a {@link DataNode}.
+   * @param node {@link DataNode}
+   * @return a collection of edges adjacent to the {@link DataNode}
+   */
+  public Collection<FlowEdge> getEdges(DataNode node);
```
zxcware: What's its use case? Are we exposing too many details of a graph?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.gobblin.runtime.api.Spec;
+import org.apache.gobblin.service.modules.dataset.DatasetDescriptor;
+import org.apache.gobblin.service.modules.flowgraph.Dag;
+
+/**
+ * An interface primarily for representing a flow of {@link JobTemplate}s. It also has
+ * method for retrieving required configs for every {@link JobTemplate} in the flow.
+ */
+@Alpha
+public interface FlowTemplate extends Spec {
```
zxcware: Can we have a `FlowTemplate` exposes one  `JobTemplate`, which works with only one pair with `Dataset`s. We may end up with lots of `FlowEdges` in this way, but the mappings among JobTemplate and Dataset pairs are naturally captured.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * @return the {@link Collection} of {@link JobTemplate}s that belong to this {@link FlowTemplate}.
+   */
+  List<JobTemplate> getJobTemplates();
+
+  /**
+   *
+   * @return the {@link Dag<JobTemplate>} that backs the {@link FlowTemplate}.
+   */
+  Dag<JobTemplate> getDag() throws IOException;
```
zxcware: It looks like an implementation detail of a `FlowTemplate`, which resolves the final job template from a dag of job templates

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdgeFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+public interface FlowEdgeFactory {
+  /**
+   * Construct a {@link FlowEdge} from the edge properties
+   * @param edgeProps properties of the {@link FlowEdge}
+   * @param catalog an instance of {@link FSFlowCatalog}
+   * @return an instance of {@link FlowEdge}
+   * @throws FlowEdgeCreationException
+   */
+  public FlowEdge createFlowEdge(Properties edgeProps, FSFlowCatalog catalog) throws FlowEdgeCreationException;
```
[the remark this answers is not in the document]
sv2000: FSFlowCatalog provides the FlowTemplate needed for constructing a FlowEdge.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * An interface for {@link FlowGraph}. A {@link FlowGraph} consists of {@link DataNode}s and {@link FlowEdge}s.
+ * The interface provides methods for adding and removing {@link DataNode}s and {@link FlowEdge}s to the {@link FlowGraph}.
+ * In addition the interface provides methods to return factory classes for creation of {@link DataNode}s and {@link FlowEdge}s.
+ */
+
+@Alpha
+public interface FlowGraph {
+  public Collection<DataNode> getNodes();
```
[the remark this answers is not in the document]
sv2000: Fair point! Could be useful in the future for graph traversal algorithms. Removed it from the interface definition for now.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
sv2000: DataNodes and FlowEdges are homogeneous. Each DataNode/FlowEdge in the graph is created in a common way. DataNodes and FlowEdges are distinguishable by their properties. 

The idea of providing node/edge creation factories is to hide the details of the creation process to the external classes.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
sv2000: See comment above.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
sv2000: Useful for path-finding algorithms such as Djikstra, BFS etc. Added more details to the comment to explain the use of the API.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdgeFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Get an edge label from the edge properties
+   * @param edgeProps properties of the edge
+   * @return a string label identifying the edge
+   */
+  public String getEdgeLabel(Properties edgeProps) throws IOException;
```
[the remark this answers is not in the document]
sv2000: Thats a good point! The idea is that the construction of edge label can be different for different implementations of FlowEdge. The intent here was to expose a common method that hides the specific implementation of the FlowEdge to the outside classes. Suggestions on cleaner way to do this are welcome.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
sv2000: Sorry - I don't quite understand this comment. Can you explain?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
```diff
@@ -0,0 +1,84 @@
...
+   * Get properties of this edge.
+   * @return the properties of this edge as a {@link Config} object.
+   */
+  Config getProps();
+
+  /**
+   * Get a string label identifying the edge.
+   * @return the label of the {@link FlowEdge}.
+   */
+  String getLabel();
```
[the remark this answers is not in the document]
sv2000: Done.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdge.java
```diff
@@ -0,0 +1,84 @@
...
+  FlowTemplate getFlowTemplate();
+
+  /**
+   *
+   * @return a list of {@link SpecExecutor}s that can execute the {@link FlowTemplate} corresponding to this edge.
+   */
+  List<SpecExecutor> getExecutors();
+
+  /**
+   * Get properties of this edge.
```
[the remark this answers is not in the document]
sv2000: Ideally, the properties should encapsulate all the information (e.g. its end points, FlowTemplate URI,  SpecExecutors) about a FlowEdge from which a FlowEdge can be instantiated.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
sv2000: Each JobTemplate itself could have multiple pairs of Input/Output datasets. Think Distcp. As you correctly point out, it will result in too many FlowEdges and requires generation of too many JobTemplates (likely, manually).

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   *
+   * @return all configuration inside pre-written template.
+   */
+  Config getRawTemplateConfig();
+
+  /**
+   * @return list of input/output {@link DatasetDescriptor}s for the {@link FlowTemplate}.
+   */
+  List<Pair<DatasetDescriptor, DatasetDescriptor>> getInputOutputDatasetDescriptors();
```
zxcware: We've talked about this in meeting. Most of the interface defined are exposing its properties. For this one, the actual use case is check if a give pair of input and ouput `DatasetDescriptor`s is acceptable in the flow template. We could just expose `isAcceptable(Pair<DatasetDescriptor, DatasetDescriptor>)` to encapsulate the object more.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
zxcware: +1

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
zxcware: This is a pending point about whether we could just expose a resolved `Dag` instead of a series of properties for some external object to resolve the `Dag`. 

In my opinion, the goal is to provide a resolved `Dag` of a flow edge and getting the job templates as a collection or dag is not a complete use case of this object. At the moment, abstracting or extracting the logic to resolve job templates is unnecessary and sounds a bit over-engineered. 

I looked at the implementation of `JobTemplate`, which has the following method or behavior:
` Config getResolvedConfig(Config userProps) throws SpecNotFoundException, TemplateException;`
It also has a method `getRawTemplateConfig` for the purpose of inheritance not for resolution.

Yeah, others may have different opinions.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/Dag.java
```diff
@@ -0,0 +1,153 @@
...
+ * An implementation of Dag. Assumes that nodes have unique values. Nodes with duplicate values will produce
+ * unpredictable behavior.
+ */
+@Alpha
+@Getter
+public class Dag<T> {
+  private List<DagNode<T>> startNodes;
+  private List<DagNode<T>> endNodes;
+  // Map to maintain parent to children mapping.
+  private Map<DagNode, List<DagNode<T>>> parentChildMap;
```
zxcware: Why the mapping is necessary with parent pointers in the `DagNode`?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/Dag.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * An implementation of Dag. Assumes that nodes have unique values. Nodes with duplicate values will produce
+ * unpredictable behavior.
+ */
+@Alpha
+@Getter
+public class Dag<T> {
+  private List<DagNode<T>> startNodes;
+  private List<DagNode<T>> endNodes;
```
zxcware: Either start nodes or end nodes are sufficient to get all paths of a dag. We don't need both

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
sv2000: Not quite. A FlowEdge may accept a set of I/O dataset descriptors. During path finding, we need to potentially look at each pair in this set to determine if there is at least one I/O descriptor that is compatible with the target. Hence the need for this API.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/Dag.java
sv2000: startNodes and endNodes help:
1. to determine which jobs to start execution with and when a dag execution is complete
2. Useful for concatenating two dags. Just set the parent/child pointers of the endNodes and startNodes.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @return a human-readable description of the dataset.
+   */
+  public String getDescription();
+
+  /**
+   * @return the dataset urn.
+   */
+  public String getDatasetUrn();
```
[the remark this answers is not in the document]
sv2000: Removing this. The path/format/platform properties of a DatasetDescriptor seem to be sufficient for now.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowEdgeFactory.java
```diff
@@ -0,0 +1,53 @@
...
+   * @throws FlowEdgeCreationException
+   */
+  public FlowEdge createFlowEdge(Config edgeProps, FSFlowCatalog catalog) throws FlowEdgeCreationException;
+
+  /**
+   * Get an edge label from the edge properties
+   * @param edgeProps properties of the edge
+   * @return a string label identifying the edge
+   */
+  public String getEdgeId(Config edgeProps) throws IOException;
```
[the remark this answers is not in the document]
sv2000: You are correct that it is not related to creating a flow edge. It is meant be used for edge deletion. We want to delete the edge from a graph using the edgeId, which is constructed from edge properties. How the edgeId is constructed from edge properties is an implementation detail of the specific FlowGraph implementation. The method is intended to hide the details of constructing the edgeId, and hence, it was put into the Factory class.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+   * @return true if {@link FlowEdge} is added to the {@link FlowGraph} successfully.
+   */
+  public boolean addFlowEdge(FlowEdge edge);
+
+  /**
+   * Remove a {@link DataNode} and all its incident edges from the {@link FlowGraph}
+   * @param node {@link DataNode} to be removed
+   * @return true if {@link DataNode} is removed from the {@link FlowGraph} successfully.
+   */
+  public boolean deleteDataNode(DataNode node);
```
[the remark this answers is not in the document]
sv2000: Removed from the interface.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
```diff
@@ -0,0 +1,102 @@
...
+   * @return true if {@link DataNode} is removed from the {@link FlowGraph} successfully.
+   */
+  public boolean deleteDataNode(String nodeId);
+
+  /**
+   * Remove a {@link FlowEdge} from the {@link FlowGraph}
+   * @param edge to be removed
+   * @return true if edge is removed from the {@link FlowGraph} successfully.
+   */
+  public boolean deleteFlowEdge(FlowEdge edge);
```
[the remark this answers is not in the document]
sv2000: Removed from the interface.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
```diff
@@ -0,0 +1,212 @@
...
+ *    <p>nodes - the set of {@link DataNode}s in the graph</p>
+ *    <p>nodesToEdges - the mapping from each {@link DataNode} to its outgoing {@link FlowEdge}s</p>
+ *    <p>dataNodeMap - the mapping from a node identifier to the {@link DataNode} instance</p>
+ *    <p>flowEdgeMap - the mapping from a edge label to the {@link FlowEdge} instance</p>
+ */
+@Alpha
+@Slf4j
+public class BaseFlowGraph implements FlowGraph {
+  @Getter
+  private Set<DataNode> nodes = new HashSet<>();
```
zxcware: Is `dataNodeMap` sufficient to do what the set does?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/FlowGraph.java
zxcware: Is it better to decouple the factory from flow graph?  Like how we treat the data node factory, edge factory class is specified by configuration.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
sv2000: Yes good point! Fixed it.

PR #2393, opened 2018-07-09
GOBBLIN-528: Multihop Flow Compiler for Gobblin-as-a-Service (GaaS).

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-528

### Description
Create a new flow compiler that can compile flows into a DAG of JobSpecs. 

### Tests
Among other unit tests, added unit test FlowGraphPathFinderTest for testing flow compilation.

### Commits


autumnust:
Hi @sv2000,
Can you point a design doc for this PR to start with?  That might accelerate the reviewing process. Thanks !

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+                .atPath(DatasetDescriptorConfigKeys.FLOW_EDGE_INPUT_DATASET_DESCRIPTOR_PREFIX);
+            Config outputDatasetDescriptorConfig = outputDatasetDescriptor.getRawConfig()
+                .atPath(DatasetDescriptorConfigKeys.FLOW_EDGE_OUTPUT_DATASET_DESCRIPTOR_PREFIX);
+            userConfig = userConfig.withFallback(inputDatasetDescriptorConfig).withFallback(outputDatasetDescriptorConfig);
+            if (!isFlowTemplateResolvable(flowEdge, userConfig)) {
+              continue;
+            }
+            FlowEdgeContext flowEdgeContext;
+            if (currentDatasetDescriptor.isCompatibleWith(outputDatasetDescriptor)) {
+              //If datasets described by the currentDatasetDescriptor is a subset of the datasets described
```
yukuai518: Should we rename the API to isSubset(descriptor) or doesBelongsTo(descriptor)? I don't think isCompatibleWith can always infer a subset relationship, and more importantly it doesn't imply who is more specific to be preserved. What if currentDatasetDescriptor is not a subset of outputDatasetDescriptor and vice versa, but they are compatible -- then we lost outputDatasetDescriptor info?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -41,4 +65,15 @@
    * This check is non-commutative.
    */
   public boolean isCompatibleWith(DatasetDescriptor other);
+
+  /**
+   * @return true if the platform-independent properties of this {@link DatasetDescriptor} are compatible with the
+   * platform-independent properties of the other {@link DatasetDescriptor}.
+   */
+  public boolean isPropertyCompatibleWith(DatasetDescriptor other);
```
zxcware: What are `platform-independent` properties? Can we group them similar to `EncryptionConfig`? One can specifically get the group and check its compatibility instead of the general `DatasetDescriptor` provides a partial and unclear compatibility check besides a full compatibility check.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+
+
+@Alpha
+@Slf4j
+public class FlowGraphPathFinder {
+  private FlowGraph flowGraph;
+  private FlowSpec flowSpec;
+  private Config flowConfig;
+
+  private String srcNodeId;
```
zxcware: minor: Can we just use `srcNode`?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
yukuai518: We need to clarify what is the difference between property compatibility and normal compatibility.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+   * @param currentDatasetDescriptor Output {@link DatasetDescriptor} of the current edge.
+   * @param destDatasetDescriptor Target {@link DatasetDescriptor}.
+   * @return prioritized list of {@link FlowEdge}s to be added to the edge queue for expansion.
+   */
+  private List<FlowEdgeContext> getNextEdges(DataNode dataNode, DatasetDescriptor currentDatasetDescriptor,
+      DatasetDescriptor destDatasetDescriptor, Config userConfig) {
+    List<FlowEdgeContext> prioritizedEdgeList = new LinkedList<>();
+    for (FlowEdge flowEdge : this.flowGraph.getEdges(dataNode)) {
+      try {
+        DataNode edgeDestination = this.flowGraph.getNode(flowEdge.getEndPoints().get(1));
```
zxcware: `flowEdge.getEndPoints().get(1)` looks like a hacky way to represent src and dest data nodes of a flow edge. We should just define `String getSrcNodeId()` and `String getDestNodeId()`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DataNode.java
```diff
@@ -37,7 +36,17 @@
    * @return the attributes of a {@link DataNode}. It also includes properties for resolving a {@link org.apache.gobblin.runtime.api.JobTemplate}
    * e.g. "source.fs.uri" for an HDFS node, "jdbc.publisher.url" for JDBC node.
    */
-  Config getProps();
+  Config getRawConfig();
+
+  /**
+   * @return the attributes of a {@link DataNode} as a source of a {@link FlowEdge}.
+   */
+  Config getSrcConfig();
```
yukuai518: I am not sure why we need to have both Src and Dst for a single DataNode, should it be either of them and relationship should be determined by the edge?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+        //Should we skip this edge from any further consideration?
+        if (!edgeDestination.isActive() || !flowEdge.isActive()) {
+          continue;
+        }
+        for (Pair<DatasetDescriptor, DatasetDescriptor> datasetDescriptorPair : flowEdge.getFlowTemplate()
```
zxcware: If we moved this logic to `FlowTemplate` and expose it as a function, we could simplify `FlowGraphPathFinder` to some degree.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof AdlsDatasetDescriptor)) {
+      return false;
+    }
+    AdlsDatasetDescriptor other = (AdlsDatasetDescriptor) o;
+    if (this.getPlatform() == null || other.getPlatform() == null || !this.getPlatform().equalsIgnoreCase(other.getPlatform())) {
```
zxcware: Can we leverage `super.equals()`? This applicable to the other implementation of `DatasetDescriptor`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+  @Override
+  public boolean isCompatibleWith(DatasetDescriptor o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof AdlsDatasetDescriptor)) {
+      return false;
+    }
+    AdlsDatasetDescriptor other = (AdlsDatasetDescriptor) o;
+    if (this.getPlatform() == null || other.getPlatform() == null || !this.getPlatform().equalsIgnoreCase(other.getPlatform())) {
```
zxcware: can we leverage `super.isCompatibleWith()`? This applicable to the other implementation of `DatasetDescriptor`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+      return false;
+    }
+    return this.getPath().equals(other.getPath()) && this.getFormat().equalsIgnoreCase(other.getFormat())
+        && this.getCodecType().equalsIgnoreCase(other.getCodecType()) && this.getEncryptionConfig().equals(other.getEncryptionConfig());
+  }
+
+  @Override
+  public int hashCode() {
+    return this.toString().hashCode();
```
zxcware: This is wrong. We must make sure that if 2 objects, which could be 2 different references, are equal, they have the same hash code. This applicable to the other implementation of `DatasetDescriptor`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+
+        //Base condition 2: Check if we are already at the target. If so, return an empty dag.
+        if ((srcNode.equals(destNode)) && srcDatasetDescriptor.isCompatibleWith(destDatasetDescriptor)) {
+          return new Dag<>(new ArrayList<>());
+        }
+
+        LinkedList<FlowEdgeContext> edgeQueue = new LinkedList<>();
+        edgeQueue.addAll(getNextEdges(srcNode, srcDatasetDescriptor, destDatasetDescriptor, flowConfig));
+        for (FlowEdgeContext flowEdgeContext : edgeQueue) {
+          this.pathMap.put(flowEdgeContext, flowEdgeContext);
```
zxcware: Instead of using a map to keep track of all links, can we just use a list or queue to keep track of the path we're interested?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+        // If the edge E' satisfies 1 and 2, add it to the edge queue for further consideration.
+        while (!edgeQueue.isEmpty()) {
+          FlowEdgeContext flowEdgeContext = edgeQueue.pop();
+
+          DataNode currentNode = this.flowGraph.getNode(flowEdgeContext.getEdge().getEndPoints().get(1));
+          DatasetDescriptor currentOutputDatasetDescriptor = flowEdgeContext.getOutputDatasetDescriptor();
+
+          //Are we done?
+          if (isPathFound(currentNode, destNode, currentOutputDatasetDescriptor, destDatasetDescriptor)) {
+            return buildDag(flowEdgeContext);
```
zxcware: Why are we putting this in the path finder? Can we move the logic to resolve job specs into a dag from the path into a different class? Path finding and Dag resolution are 2 different responsibilities and should be decoupled. It also reduces the size of this class.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
yukuai518: @zxcware I think this is fine because the map is used to recover the original path.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
yukuai518: Should we consider the multicast case, or like a diamond shape path? For example, a flow might look like A->B->D and A->C->D (or A->B and A->C and A->D). Currently there is no way to specify multiple dst. But is there a way in the future to expand this algorithm to handle multiple cases? Otherwise we will have to create another PathFinder which might have a lot of duplicates. If you need different algorithm, please make this PathFinder as a base class so that more strategy can be applied.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/StaticFlowTemplate.java
```diff
@@ -67,25 +67,21 @@
   private transient Dag<JobTemplate> dag;
 
   private transient Config rawConfig;
-  private boolean isTemplateMaterialized;
 
-  public StaticFlowTemplate(URI uri, String version, String description, Config config,
+  public StaticFlowTemplate(URI flowTemplateDirUri, String version, String description, Config config,
```
yukuai518: Is it possible to have Builder so that we can avoid so many parameters. It would be much earlier in the future to expand the functionality.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/spec/JobSpecWithExecutor.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.gobblin.runtime.api.JobSpec;
+import org.apache.gobblin.runtime.api.SpecExecutor;
+
+
+/**
+ * A wrapper around {@link JobSpec} to encapsulate the {@link SpecExecutor}.
+ */
+@Data
+@AllArgsConstructor
+public class JobSpecWithExecutor {
```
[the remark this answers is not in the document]
sv2000: Good suggestion! I changed it to JobExecutionPlan.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+      synchronized (this.flowGraph) {
+        //Base condition 1: Source Node or Dest Node is inactive; return null
+        if (!srcNode.isActive() || !destNode.isActive()) {
+          log.warn("Either source node {} or destination node {} is inactive; skipping path computation.", srcNodeId,
+              destNodeId);
+          return null;
+        }
+
+        //Base condition 2: Check if we are already at the target. If so, return an empty dag.
+        if ((srcNode.equals(destNode)) && srcDatasetDescriptor.isCompatibleWith(destDatasetDescriptor)) {
```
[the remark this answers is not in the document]
sv2000: This is a base condition check. If they are not compatible, we need to find a path that would make the src dataset compatible with the destination and land the dataset at the destination.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,428 @@
...
+          .invokeLongestConstructor(destDatasetDescriptorClass, destDatasetDescriptorConfig);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * A helper class used to maintain additional context associated with each {@link FlowEdge} during path
+   * computation while the edge is explored for its eligibility. The additional context includes the parent
+   * {@link FlowEdge}, the input {@link DatasetDescriptor} of this edge which is compatible with the parent's output
```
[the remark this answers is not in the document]
sv2000: Changed the comment to make it more clear. Removed the reference to "parent".

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: Agreed. Changed it to getSrc() and getDest().

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: @yukuai518 belongsTo() is a good suggestion!

1. If currentDatasetDescriptor is not a subset of outputDatasetDescriptor (e.g. when the edge transforms input dataset from one format to another), then we choose the outputDatasetDescriptor to carry forward. That is what is being checked on line 262. Note that compatibility definition (or belongsTo()) includes both path as well as dataset format check.

2. I think the dataset descriptor should be independent of node specific information. Your suggestion to drop scheme and authority is a good one.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: Sure. Done.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
sv2000: By platform-independent properties, I really mean the dataset format which includes: 1. the data representation (e.g. avro/json/csv), 2. the encoding (e.g. gzip/bzip2/deflate), and 3. encryption config. Definitely possible to group them like EncryptionConfig.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: @yukuai518 I was planning to have a separate PR to handle the case of multiple destinations. In the first version, single source-multiple destinations will be treated as a sequence of single source single destination, and we will leverage the findPath() for this case.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
sv2000: Yes, we can simplify this and reduce the number of dataset descriptor classes. We can have a single FSDatasetDescriptor class and make specific implementations of FS (e.g. HDFS, LocalFS, ADLS) as a platform property.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
sv2000: Same as comment above.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/AdlsDatasetDescriptor.java
sv2000: Yes, the current implementation is wrong, since we ignore case in equals() but not in toString(). Made the change.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseDataNode.java
```diff
@@ -38,7 +38,11 @@
   @Getter
   private String id;
   @Getter
-  private Config props;
+  private Config rawConfig;
```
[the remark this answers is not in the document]
sv2000: Similar to @yukuai518 's comment. Original intent was to perform substitution of a template property (e.g. source.fs.uri/target.fs.uri) based on whether the node is a source or destination. 

I changed it to do the src/dest node specific substitution in the check for FlowGraphPathFinder#isFlowTemplateResolvable().

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DataNode.java
sv2000: See my response to @zxcware's similar comment. We do not need this distinction and have made the change.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: Changed belongsTo() to contains().

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: @zxcware Moved the logic of building the job spec out of the class. The rest of the logic is tied to path finding.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: I think it would be a bit too cumbersome to move this logic to FlowTemplate. Further, it would need the FlowTemplate to become aware of graph constructs such as FlowEdge, which does not seem ideal either.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -17,28 +17,54 @@
...
+   * when the dataset is backed by a FileSystem. In the case of a database table, it is dbName.tableName.
+   * @return dataset path.
+   */
+  public String getPath();
+
+  /**
+   *
+   * @return storage format of the dataset.
+   */
+  public FormatDescriptor getFormatDescriptor();
```
zxcware: For code readability, name it as `FormatConfig`, consistent with `EncryptionConfig`? In my experience, a descriptor usually describes a kind of resource. Format is not really a resource.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -17,28 +17,54 @@
...
 
   /**
-   * @return true if this {@link DatasetDescriptor} is compatible with the other {@link DatasetDescriptor} i.e. the
-   * datasets described by this {@link DatasetDescriptor} is a subset of the datasets described by the other {@link DatasetDescriptor}.
-   * This check is non-commutative.
+   * @return true if this {@link DatasetDescriptor} contains the other {@link DatasetDescriptor} i.e. the
+   * datasets described by this {@link DatasetDescriptor} is a subset of the datasets described by the other
+   * {@link DatasetDescriptor}. This operation is non-commutative.
+   */
+  public boolean contains(DatasetDescriptor other);
```
zxcware: Did you mean this dataset includes the other, for example, this is `/a`, other is `/a/b`?  In `EncryptionConfig`, there is also a method named `contains`, which somewhat means a regex pattern. 

I guess what we want is to check if `other` could point to the same thing as this. It is tough to give a good name.  It's kind of a generalize relationship. A general representation can "contain" a specific one, but not vice versa.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
zxcware: FlowEdge does not need to, but its merged config, be exposed to FlowTemplate

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,380 @@
...
+        // If the edge E' satisfies 1 and 2, add it to the edge queue for further consideration.
+        while (!edgeQueue.isEmpty()) {
+          FlowEdgeContext flowEdgeContext = edgeQueue.pop();
+
+          DataNode currentNode = this.flowGraph.getNode(flowEdgeContext.getEdge().getDest());
+          DatasetDescriptor currentOutputDatasetDescriptor = flowEdgeContext.getOutputDatasetDescriptor();
+
+          //Are we done?
+          if (isPathFound(currentNode, destNode, currentOutputDatasetDescriptor, destDatasetDescriptor)) {
+            return buildDag(flowEdgeContext);
```
zxcware: My idea actually meant:
- `FlowGraphPathFinder` doesn't not build the Dag, but return the path
- A new class, possibly named `DagBuilder` builds the Dag from the path. 

In this way, the responsibility is separated so that the way to build the `Dag` can be replaced without touching the logic that focuses on find a path.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
zxcware: The search is a depth first search, aggressively going deeper to find the first working path. We can push an candidate edge to a queue, popping it out if it can't match further. In the end, the queue describes all hops of a result path.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
sv2000: Introduced a new class FlowGraphPath with method asDag() that returns a Dag from a List<FlowEdgeContext>.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/DatasetDescriptor.java
```diff
@@ -17,28 +17,54 @@
...
```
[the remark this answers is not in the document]
sv2000: I prefer to leave it as contains(), since it captures the set relationship. At any rate, isApplicableTo() is the same as isCompatibleWith() that I had earlier.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/dataset/EncryptionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+  public EncryptionConfig(Config encryptionConfig) {
+    this.encryptionAlgorithm = ConfigUtils.getString(encryptionConfig, DatasetDescriptorConfigKeys.ENCRYPTION_ALGORITHM_KEY,
+        DatasetDescriptorConfigKeys.DATASET_DESCRIPTOR_CONFIG_ANY);
+    this.keystoreType = ConfigUtils.getString(encryptionConfig, DatasetDescriptorConfigKeys.ENCRYPTION_KEYSTORE_TYPE_KEY,
+        DatasetDescriptorConfigKeys.DATASET_DESCRIPTOR_CONFIG_ANY);
+    this.keystoreEncoding = ConfigUtils.getString(encryptionConfig, DatasetDescriptorConfigKeys.ENCRYPTION_KEYSTORE_ENCODING_KEY,
+        DatasetDescriptorConfigKeys.DATASET_DESCRIPTOR_CONFIG_ANY);
+  }
+
+  public boolean contains(EncryptionConfig other) {
```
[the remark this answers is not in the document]
sv2000: EncryptionConfig is part of DatasetConfig. It is not an implementation of DatasetConfig.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/FlowGraphPathFinder.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+  }
+
+  /**
+   * A simple path finding algorithm based on Breadth-First Search. At every step the algorithm adds the adjacent {@link FlowEdge}s
+   * to a queue. The {@link FlowEdge}s whose output {@link DatasetDescriptor} matches the destDatasetDescriptor are
+   * added first to the queue. This ensures that dataset transformations are always performed closest to the source.
+   * @return a list of {@link FlowEdge}s starting at the srcNode and ending at the destNode.
+   */
+  public Dag<JobExecutionPlan> findPath() throws PathFinderException {
```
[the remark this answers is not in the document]
sv2000: Sure, I will make this change.

PR #2397, opened 2018-07-13
[GOBBLIN-534] switched from helix job queue to work flow

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @htran1 @yukuai518  please review

### JIRA

### Description
Gobblin Helix Jobs are more suitable for Helix Work Flow instead of JobQueue, because work flow automatic cleanup after job completion and we submit only one job per work flow. More details and differences between the two can be found at http://helix.apache.org/0.8.0-docs/tutorial_task_framework.html section 'Creating a Queue'.

### Tests
tests already present

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
@@ -18,7 +18,6 @@
 package org.apache.gobblin.cluster;
 
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.helix.HelixManager;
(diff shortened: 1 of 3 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -89,34 +90,43 @@ public static String getHelixInstanceName(
...
       workflowConfig = null;
     }
 
-    // Create one queue for each job with the job name being the queue name
+    // Create a work flow for each job with the name being the queue name
     if (workflowConfig == null) {
-      JobQueue jobQueue = new JobQueue.Builder(queueName).build();
-      helixTaskDriver.createQueue(jobQueue);
-      log.info("Created job queue {}", queueName);
+      // Create a workflow, set expiry to 5 minutes and add the job
```
yukuai518: Can you put some documentation for this method to explain :
1) Why we want to deleteAndWaitForCompletion?
2) How the expiration works from Helix point of view?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -89,34 +90,43 @@ public static String getHelixInstanceName(
...
+      workflowConfig = new WorkflowConfig.Builder().setExpiry(5L, TimeUnit.MINUTES).build();
```
yukuai518: Can we move this expiration to next PR so that this PR only contains the migration from jobQueue to workflow?

If I remember correctly, Helix team want us to use the stop method and then fail the job on purpose via some specific API so that the delete will be handled by Helix properly.

arjun4084346: Even if we fail the job with some new Helix API, there will always be race condition between job stop->fail->cleanup and job status checking in HelixUtils::waitJobCompletion().
if waitJobCompletion could not check status between fail and cleanup, it will stuck there; and this is very likely.
I think we should go with adding a synchronization primitive as you suggested offline.

yukuai518: Yes, let's handle this in our next PR: 
   1) Auto deletion
      a) Add expiration attributes.
      b) stop->fail->cleanup
   2) An additional status flag to avoid race condition
Once we have auto deletion ready, we can
   3) Remove workflow deletion in the job submission (but you have to use jobId as the workflow name so a non-deleted workflow won't block the new jobs.)

yukuai518:
+1

PR #2405, opened 2018-07-24
[GOBBLIN-542] Change AsynRequest instance variable into protected 

autumnust:
To make this class is extensible. 
Gobblin Salesforce Writer implementation is going to extend this class to have some customized behavior in the `Request`.

Dear Gobblin maintainers,

### JIRA
- [x] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-542] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-542

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/async/AsyncRequest.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/async/AsyncRequest.java
@@ -37,8 +37,8 @@ import org.apache.gobblin.net.Request;
  */
 public class AsyncRequest<D, RQ> implements Request<RQ> {
   @Getter @Setter
-  private RQ rawRequest;
-  private final List<Thunk<D>> thunks = new ArrayList<>();
+  protected RQ rawRequest;
+  protected final List<Thunk<D>> thunks = new ArrayList<>();
 
   /**
    * Get the total number of records processed in the request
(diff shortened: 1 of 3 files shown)


gobblin-core/src/main/java/org/apache/gobblin/async/AsyncRequest.java
```diff
@@ -61,7 +61,7 @@ public long getBytesWritten() {
   /**
    * Get all records information in the request
    */
-  public List<Thunk<D>> getThunks() {
+  public List<Thunk<D>> getThunksCopy() {
```
zxcware: Avoid back incompatible change. There are 3rd party codes already using this method.

autumnust: Sure. One follow up question here is why do we have a copy instead of returning a reference to thunk container. Any specific use case for that?

autumnust: Talked offline, copy was meant to protect `AsyncRequest` from unexpected modification.

abti:
+1 Thanks @zxcware

PR #2422, opened 2018-08-13
GOBBLIN-561: Handle data completeness checks for data partitions with no records.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-561

### Description
Compaction jobs fail on ETL-ed data partitions with no input records. We need to handle handle this case gracefully and output the empty partitions.

### Tests
Tested the flow in Dev environment.

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/CompactionAvroJobConfigurator.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/CompactionAvroJobConfigurator.java
@@ -64,6 +64,7 @@ import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.dataset.Dataset;
 import org.apache.gobblin.dataset.FileSystemDataset;
+import org.apache.gobblin.hive.policy.HiveRegistrationPolicy;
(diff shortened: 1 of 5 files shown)


sv2000:
@ibuenros Please review this PR.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/CompactionAvroJobConfigurator.java
```diff
@@ -270,7 +278,10 @@ public Job createJob(FileSystemDataset dataset) throws IOException {
     addJars(conf);
     Job job = Job.getInstance(conf);
     job.setJobName(MRCompactorJobRunner.HADOOP_JOB_NAME);
-    this.configureInputAndOutputPaths(job, dataset);
+    if (!this.configureInputAndOutputPaths(job, dataset)) {
+      log.info("No valid input paths found for dataset: " + dataset.getUrn());
+      return null;
```
ibuenros: It's worth considering whether we want to create an empty output directory in this case, since that is technically the correct compacted partition, and it will not impact consumers expecting the partition to be present.

sv2000: Good point! Addressed it. Thanks!

PR #2423, opened 2018-08-13
GOBBLIN-559: Implement FlowGraph as a concurrent data structure.

sv2000:
GOBBLIN-559: Implement FlowGraph as a concurrent data structure.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-559.

### Description
Current implementation of FlowGraph only allows a single-threaded access to the data structure for reads and writes. We improve the implementation by allowing the readers to not block on other readers, and only block on a write.

### Tests
Existing unit tests in MultiHopFlowCompilerTest class.

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
@@ -21,8 +21,15 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.gobblin.annotation.Alpha;
+import org.apache.gobblin.runtime.api.FlowSpec;
+import org.apache.gobblin.service.modules.flow.FlowGraphPath;
+import org.apache.gobblin.service.modules.flowgraph.pathfinder.PathFinder;
+import org.apache.gobblin.util.ConfigUtils;
+import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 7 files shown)


sv2000:
@zxcware Please review this PR.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
```diff
@@ -74,25 +90,30 @@ public synchronized boolean addDataNode(DataNode node) {
...
+        return false;
       }
-      this.nodesToEdges.put(dataNode, adjacentEdges);
-      String edgeId = edge.getId();
-      this.flowEdgeMap.put(edgeId, edge);
-      return true;
-    } else {
-      return false;
+      DataNode dataNode = getNode(srcNode);
+      if (dataNode != null) {
```
zxcware: Nitpick: I would the following first
```
if (dataNode == null) {
 return false
}
```

This makes the codes more stream-lined, especially in situation where we have so many bracket.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
```diff
@@ -115,20 +137,25 @@ public synchronized boolean deleteDataNode(String nodeId) {
...
+        dataNodeMap.remove(node.getId());
+
+        //Delete all the edges adjacent to the node. First, delete edges from flowEdgeMap and next, remove the edges
+        // from nodesToEdges
+        for (FlowEdge edge : nodesToEdges.get(node)) {
+          flowEdgeMap.remove(edge.getId());
+        }
+        nodesToEdges.remove(node);
+        return true;
+      } else {
```
zxcware: Same here, the stream-line nitpick

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
sv2000: Fixed it. Thanks!

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseFlowGraph.java
sv2000: Fixed it.

PR #2447, opened 2018-09-11
[GOBBLIN-583] Add support for writing the schema file using a staging…

htran1:
… file when there is an existing target file

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-583

### Description
When a partition with a schema file is reregistered there is a window where the file does not exist and a new file is created, permissions are set, then the schema is written.

Reduce this window by adding support to stage the file creation, permission setting, and schema write in a temporary file. Then delete the target file and replace with the temp file.

### Tests

### Commits


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/avro/HiveAvroSerDeManager.java
```diff
@@ -56,6 +56,10 @@
   public static final boolean DEFAULT_USE_SCHEMA_FILE = false;
   public static final String SCHEMA_FILE_NAME = "schema.file.name";
   public static final String DEFAULT_SCHEMA_FILE_NAME = "_schema.avsc";
+  public static final String SCHEMA_TEMP_FILE_NAME = "schema.temp.file.name";
```
abti: dots between words .. but I see you tried to follow naming convention of other config keys. may be we can break convention to do the right thing?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/avro/HiveAvroSerDeManager.java
```diff
@@ -79,7 +85,9 @@ public HiveAvroSerDeManager(State props) throws IOException {
     }
 
     this.useSchemaFile = props.getPropAsBoolean(USE_SCHEMA_FILE, DEFAULT_USE_SCHEMA_FILE);
+    this.useSchemaTempFile = props.getPropAsBoolean(USE_SCHEMA_TEMP_FILE, DEFAULT_USE_SCHEMA_TEMP_FILE);
     this.schemaFileName = props.getProp(SCHEMA_FILE_NAME, DEFAULT_SCHEMA_FILE_NAME);
+    this.schemaTempFileName = props.getProp(SCHEMA_TEMP_FILE_NAME, DEFAULT_SCHEMA_TEMP_FILE_NAME);
```
abti: what is the parent directory for this temp schema file? can it be same over or between executions, if so should we add a UUID to the name to avoid it being overwritten?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/avro/HiveAvroSerDeManager.java
htran1: This is very similar to the existing `schema.file.name`, so I thought it would be confusing to have `schema.tempFileName`, but leave the other one as `schema.file.name`.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/avro/HiveAvroSerDeManager.java
htran1: The temp file is placed in the same directory as the schema file. It is only temporary and won't exist after being renamed to the final schema file name.

PR #2462, opened 2018-09-27
[GOBBLIN-571] Add optional state property to KafkaTopic

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-571

### Description

Add optional state property to `KafkaTopic` which gets added to topic specific state. This allows consumer clients to add topic specific configuration while getting filtered topics.

### Tests

### Commits


jack-moseley:
@zxcware please review

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -221,6 +221,16 @@ public String apply(KafkaTopic topic) {
             }
           }), state);
 
+      for (KafkaTopic topic : topics) {
+        if (topic.getTopicSpecificState().isPresent()) {
+          if (topicSpecificStateMap.containsKey(topic.getName())) {
+            topicSpecificStateMap.get(topic.getName()).addAll(topic.getTopicSpecificState().get());
```
zxcware: Use `addAllIfNotExist`? Because we don't want to override configurations specified by the user.

jack-moseley: Good point, changed it.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -221,6 +221,16 @@ public String apply(KafkaTopic topic) {
             }
           }), state);
 
+      for (KafkaTopic topic : topics) {
+        if (topic.getTopicSpecificState().isPresent()) {
+          if (topicSpecificStateMap.containsKey(topic.getName())) {
+            topicSpecificStateMap.get(topic.getName()).addAllIfNotExist(topic.getTopicSpecificState().get());
+          } else {
+            topicSpecificStateMap.put(topic.getName(), topic.getTopicSpecificState().get());
```
htran1: Is there any chance this state gets mutated later? Is there a need for a deep copy?
If a deep copy is required then you can replace the if/else block with 
`topicSpecificStateMap.computeIfAbsent(topic.getName(), k -> new State()).addAllIfNotExist(.getTopicSpecificState().get());`

jack-moseley: Good point, the way I'm using it I don't think it would be mutated, but it's probably better to do that to be safe.

PR #2492, opened 2018-10-29
[GOBBLIN-623] Add getLocations functionality for distcp block locality

ca-shen98:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-623

### Description
To make use of data locality within the mapreduce framework, the InputSplits that represent the work unit task needs to provide node names as hints in the getLocations method. This PR implements this by retrieving and storing these block locations as a property in the work unit in DistcpFileSplitter.splitFiles, and then retrieving the list by deserializing the work unit file when creating the GobblinWorkUnitsInputFormat.GobblinSplit.

### Tests
The actual code functionality should remain the same, the only difference that would be seen is MR jobs should be able to be run with some data locality. Modified some of the GobblinSplit unit tests since the deserialization of the work unit file needed to be mocked when getSplits is tested.

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -534,6 +534,8 @@ public class ConfigurationKeys {
    * Configuration properties used by the CopySource.
    */
   public static final String COPY_SOURCE_FILESET_WU_GENERATOR_CLASS = "copy.source.fileset.wu.generator.class";
+  public static final String GOBBLIN_COPY_PREFIX = "gobblin.copy";
+  public static final String GOBBLIN_COPY_WORK_UNIT_WEIGHT = GOBBLIN_COPY_PREFIX + ".workUnitWeight";
 
   /**
    * Configuration properties used by the FileBasedExtractor
@@ -546,6 +548,7 @@ public class ConfigurationKeys {
   public static final String SOURCE_FILEBASED_PRESERVE_FILE_NAME = "source.filebased.preserve.file.name";
   public static final String SOURCE_FILEBASED_OPTIONAL_DOWNLOADER_CLASS = "source.filebased.downloader.class";
   public static final String SOURCE_FILEBASED_ENCRYPTED_CONFIG_PATH = "source.filebased.encrypted";
+  public static final String SOURCE_FILEBASED_BLOCK_LOCATIONS = "source.filebased.block.locations";
 
   public static final String SOURCE_FILEBASED_FS_PRIOR_SNAPSHOT_REQUIRED = "source.filebased.fs.prior.snapshot.required";
   public static final boolean DEFAULT_SOURCE_FILEBASED_FS_PRIOR_SNAPSHOT_REQUIRED = false;
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -531,14 +531,6 @@ public class ConfigurationKeys {
   public static final String SQL_SERVER_CONNECTION_PARAMETERS = "source.querybased.sqlserver.connectionParameters";
 
   /**
-   * Configuration properties used internally by Gobblin Splits to determine block locations
-   */
-  public static final String GOBBLIN_SPLIT_PREFIX = "gobblin.split";
-  public static final String GOBBLIN_SPLIT_FILE_PATH = GOBBLIN_SPLIT_PREFIX + ".file.path";
-  public static final String GOBBLIN_SPLIT_FILE_LOW_POSITION = GOBBLIN_SPLIT_PREFIX + ".file.low.position";
-  public static final String GOBBLIN_SPLIT_FILE_HIGH_POSITION = GOBBLIN_SPLIT_PREFIX + ".file.high.position";
-
-  /**
    * Configuration properties used by the CopySource.
    */
   public static final String COPY_SOURCE_FILESET_WU_GENERATOR_CLASS = "copy.source.fileset.wu.generator.class";
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -531,11 +531,17 @@ public class ConfigurationKeys {
   public static final String SQL_SERVER_CONNECTION_PARAMETERS = "source.querybased.sqlserver.connectionParameters";
 
   /**
+   * Configuration properties used internally by Gobblin Splits to determine block locations
+   */
+  public static final String GOBBLIN_SPLIT_PREFIX = "gobblin.split";
+  public static final String GOBBLIN_SPLIT_FILE_PATH = GOBBLIN_SPLIT_PREFIX + ".file.path";
+  public static final String GOBBLIN_SPLIT_FILE_LOW_POSITION = GOBBLIN_SPLIT_PREFIX + ".file.low.position";
+  public static final String GOBBLIN_SPLIT_FILE_HIGH_POSITION = GOBBLIN_SPLIT_PREFIX + ".file.high.position";
+
+  /**
    * Configuration properties used by the CopySource.
    */
   public static final String COPY_SOURCE_FILESET_WU_GENERATOR_CLASS = "copy.source.fileset.wu.generator.class";
-  public static final String GOBBLIN_COPY_PREFIX = "gobblin.copy";
-  public static final String GOBBLIN_COPY_WORK_UNIT_WEIGHT = GOBBLIN_COPY_PREFIX + ".workUnitWeight";
 
   /**
    * Configuration properties used by the FileBasedExtractor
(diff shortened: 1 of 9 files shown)


cshen98:
@sv2000 @htran1 please review when you have the chance

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -531,6 +531,8 @@
    * Configuration properties used by the CopySource.
    */
   public static final String COPY_SOURCE_FILESET_WU_GENERATOR_CLASS = "copy.source.fileset.wu.generator.class";
+  public static final String GOBBLIN_COPY_PREFIX = "gobblin.copy";
+  public static final String GOBBLIN_COPY_WORK_UNIT_WEIGHT = GOBBLIN_COPY_PREFIX + ".workUnitWeight";
```
sv2000: Why should this config be in ConfigurationKeys as opposed to CopySource/CopyConfiguration?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -468,7 +479,17 @@ private static void setWorkUnitWeight(WorkUnit workUnit, CopyEntity copyEntity,
...
+    workUnit.setProp(ConfigurationKeys.GOBBLIN_COPY_WORK_UNIT_WEIGHT, Long.toString(weight));
+  }
+
+  public static void setWorkUnitBlockLocations(WorkUnit workUnit, FileSystem sourceFs, CopyableFile file,
+      long startPos, long len) throws IOException {
+    Set<String> blockLocationsNames = Sets.newHashSet();
+    for (BlockLocation bl : sourceFs.getFileBlockLocations(file.getFileStatus(), startPos, len)) {
+      blockLocationsNames.addAll(Arrays.asList(bl.getHosts()));
+    }
+    workUnit.setProp(ConfigurationKeys.SOURCE_FILEBASED_BLOCK_LOCATIONS,
```
sv2000: We should think about where the correct location is for adding block locations. The approach you have would work, but is very Distcp-specific. Can we have the logic of fetching block locations inside the GobblinWorkUnitsInputFormat, so that other sources can benefit from it ?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -131,23 +183,37 @@ public static int getMaxMapper(Configuration conf) {
     @Singular
     private List<String> paths;
 
+    /**
+     * Locations (names of nodes) of blocks of files specified in the split's work unit files defined by paths
+     */
+    private List<String> blockLocationNames;
```
sv2000: Can this just be an Array?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -89,9 +100,13 @@ public static int getMaxMapper(Configuration conf) {
       if (inputs == null) {
         throw new IOException(String.format("Path %s does not exist.", path));
       }
+
       log.info(String.format("Found %d input files at %s: %s", inputs.length, path, Arrays.toString(inputs)));
+
       for (FileStatus input : inputs) {
         allPaths.add(input.getPath().toString());
+        blockLocationNamesForWorkUnitPaths.put(input.getPath().toString(),
```
sv2000: Can we not just get the block locations from the FS API here instead of from the de-serialized workunit file?

cshen98: So I was considering doing it here, but the first thing is that we want the block locations for the files specified by the work unit, not the work unit file itself. Because of this, we still have to deserialize the work unit file, and then the issue was that I don't think the actual file path is listed in the work unit, only the dataset pattern - the actual file paths for work units are only retrieved in the CopySource ...

cshen98: Although now that I think of it, I think it must be in the WorkUnit somewhere so that the extractor is able to open an inputstream to it ... I will look into it further to see if it can be done here, since I agree that it would be a better, more general solution

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
cshen98: Oh, I think this was a remnant from an attempt to implement the GobblinSplit.getLength method  before I realized we already do work unit bin packing so this wouldn't be something that would be very useful - I guess I forget to change this back to the way it was before...

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
cshen98: Ok, so we have to be able to deserialize the CopyableFile/CopyEntity from the WorkUnit in the GobblinWorkUnitsInputFormat class, but the problem is that the CopyableFile/CopyEntity classes are in the gobblin-data-management module which I don't think should be a dependency of gobblin-runtime ... Also, the split offset info is json serialized from a DistcpFileSplitter.Split object, which is also in the gobblin-data-management module.

cshen98:
I moved the retrieval of the block locations from DistcpFileSplitter to within GobblinWorkUnitsInputFormat by adding three GobblinSplit specific properties to ConfigurationKeys - file path, file low position (offset), & file high position. This information can be set by sources (in the distcp case, the CopySource/DistcpFileSplitter), and will then be used to get the block locations. The properties do have some overlap/duplication with properties which may already be there (i.e for distcp, the serialized copyable file stored in the work unit should already contain the file path, and the serialized split if it exits would already contain the low and high position), but these three properties provide something more general that can be used by the GobblinWorkUnitsInputFormat.getSplits method without requiring any dependencies from gobblin-runtime on for example the DistcpFileSplitter.Split class in the gobblin-data-management module.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -152,14 +157,46 @@ public static int getMaxMapper(Configuration conf) {
...
+    TreeMultimap<Long, String> blockLocationNamesOrderedByLength =
+        TreeMultimap.create(Collections.reverseOrder(), Comparator.naturalOrder());
+    HashMap<String, Long> lengthsForBlockLocationNames = Maps.newHashMap();
 
-    List<String> blockLocationNames = Lists.newArrayList();
     for (WorkUnit workUnit : workUnits) {
-      if (workUnit.contains(ConfigurationKeys.SOURCE_FILEBASED_BLOCK_LOCATIONS)) {
-        blockLocationNames.addAll(workUnit.getPropAsList(ConfigurationKeys.SOURCE_FILEBASED_BLOCK_LOCATIONS));
+      if (!workUnit.contains(ConfigurationKeys.GOBBLIN_SPLIT_FILE_PATH)) {
+        continue;
```
sv2000: Maybe add a log message warning about missing split file path.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -152,14 +157,46 @@ public static int getMaxMapper(Configuration conf) {
     } else {
       workUnits = Lists.newArrayList(tmpWorkUnit);
     }
+    return workUnits;
+  }
+
+  @VisibleForTesting
+  Collection<String> getBlockLocationsForWorkUnits(FileSystem fs, List<WorkUnit> workUnits)
+      throws IOException {
+    TreeMultimap<Long, String> blockLocationNamesOrderedByLength =
```
sv2000: Can you add javadoc explaining what the purpose of this data structure is?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -152,14 +157,46 @@ public static int getMaxMapper(Configuration conf) {
...
+        TreeMultimap.create(Collections.reverseOrder(), Comparator.naturalOrder());
+    HashMap<String, Long> lengthsForBlockLocationNames = Maps.newHashMap();
```
sv2000: Add javadoc for this data structure too.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -136,7 +141,7 @@ public static int getMaxMapper(Configuration conf) {
   }
 
   @VisibleForTesting
-  List<String> getBlockLocationNamesFromDeserializeWorkUnitFile(FileSystem fs, Path workUnitFile)
+  List<WorkUnit> getWorkUnitListFromDeserializeWorkUnitFile(FileSystem fs, Path workUnitFile)
```
sv2000: How about renaming the method to just getWorkUnitList()?

sv2000: Add javadoc explaining what the method does?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/GobblinWorkUnitsInputFormat.java
```diff
@@ -152,14 +157,46 @@ public static int getMaxMapper(Configuration conf) {
     } else {
       workUnits = Lists.newArrayList(tmpWorkUnit);
     }
+    return workUnits;
+  }
+
+  @VisibleForTesting
+  Collection<String> getBlockLocationsForWorkUnits(FileSystem fs, List<WorkUnit> workUnits)
```
sv2000: Can the method be called getBlockLocations?

cshen98:
@sv2000 (and @htran1), I implemented a `RackLocalGobblinWorkUnitsInputFormat` that is similar to the `CombineFileInputFormat`, on top of some refactoring of the previous locality stuff that I did within the `GobblinWorkUnitsInputFormat` itself. Would appreciate if you could take a look ...

PR #2497, opened 2018-11-06
GOBBLIN-627: Allow files in task output directory to be overwritten when renaming files with record count.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-627

### Description
In FsDataWriter#close(), we rename files in task output directory with overwrite option enabled. This prevents failures in the close() method of the writer, in case previous attempts of the tasks already have already written files in the task output directory. 

### Tests
Tested the changes in dev environment.

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
@@ -134,6 +134,8 @@ public class Task implements TaskIFace {
   private final String jobId;
   private final String taskId;
   private final String taskKey;
+  private final boolean isIgnoreCloseFailures;
+
   private final TaskContext taskContext;
   private final TaskState taskState;
   private final TaskStateTracker taskStateTracker;
@@ -180,6 +182,7 @@ public class Task implements TaskIFace {
     this.jobId = this.taskState.getJobId();
     this.taskId = this.taskState.getTaskId();
     this.taskKey = this.taskState.getTaskKey();
+    this.isIgnoreCloseFailures = this.taskState.isIgnoreCloseFailures();
     this.taskStateTracker = taskStateTracker;
     this.taskExecutor = taskExecutor;
     this.countDownLatch = countDownLatch;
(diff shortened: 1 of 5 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -180,6 +182,7 @@ public Task(TaskContext context, TaskStateTracker taskStateTracker, TaskExecutor
     this.jobId = this.taskState.getJobId();
     this.taskId = this.taskState.getTaskId();
     this.taskKey = this.taskState.getTaskKey();
+    this.isIgnoreCloseFailures = this.taskState.isIgnoreCloseFailures();
```
htran1: Instead of setting this in the `TaskState` where it is set in two places, why not set it here like `this.isIgnoreCloseFailures = this.taskState.getJobState().getProp(...)`?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskState.java
```diff
@@ -103,6 +106,7 @@ public TaskState(WorkUnitState workUnitState) {
     this.jobId = workUnitState.getProp(ConfigurationKeys.JOB_ID_KEY);
     this.taskId = workUnitState.getProp(ConfigurationKeys.TASK_ID_KEY);
     this.taskKey = workUnitState.getProp(ConfigurationKeys.TASK_KEY_KEY, "unknown_task_key");
+    this.ignoreCloseFailures = workUnitState.getPropAsBoolean(ConfigurationKeys.TASK_IGNORE_CLOSE_FAILURES, true);
```
htran1: I think it is safer to default to `false`. The existing behavior can result in silent data loss where the state moves forward on error. Let's consider that behavior a bug and conservatively raise an error. If users encounter an error and decide that ignoring close errors is appropriate then they can update their jobs with this config set to true.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
sv2000: Thanks for the comment! Addressed it.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskState.java
sv2000: Agreed ! Changed it.

PR #2504, opened 2018-11-20
[GOBBLIN-634] Add requester information to flowconfigs

kyuamazon:
Dear Gobblin maintainers,

### JIRA

### Description
  1. This PR adds an abstract class naming 'RequesterService'. An instance of this class is responsible for getting information who sends a request to gobblin-service. The implementation are encapsulated via abstract method findRequesters(). 
  2. To describe who sends a request, this PR creates a class named 'ServiceRequester'.
  3. The RequesterService provides a seriale/deserialize method. We are thinking to use Json string to present 'ServiceRequester'. Due to the template resolving logic will first convert properties to a config type and then convert it back, a simple json string which might contain '[],{}', etl. will break the resolution logic. The workaround is to encode this json string to Base64, then encode with URLEncoder to get rid of '+,=,/'.
   4. We fixed the injection annotation issue by using Javax package because javax package is compatible for both javax injection and guice injection, but guice annotation only works for guice.

### Tests
   tests in gobblin-service package is run successfully.

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/RequesterService.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/RequesterService.java
@@ -50,7 +50,7 @@ public abstract class RequesterService {
    */
   public String serialize(List<ServiceRequester> requestersList) throws IOException {
     String arrayToJson = objectMapper.writeValueAsString(requestersList);
-    String encodedString = Base64.getEncoder().encodeToString(arrayToJson.getBytes());
+    String encodedString = Base64.getEncoder().encodeToString(arrayToJson.getBytes("UTF-8"));
     return URLEncoder.encode(encodedString, "UTF-8");
   }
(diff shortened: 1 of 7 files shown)


gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/RequesterService.java
```diff
@@ -0,0 +1,73 @@
...
+public abstract class RequesterService {
+
+  public static final String REQUESTER_LIST = "gobblin.service.requester.list";
+
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  // user should provide customized logic to get real requester
+  @SuppressWarnings(value = "MS_SHOULD_BE_FINAL",
+      justification = "This is not final variable because derived service can override it.")
+  public static RequesterService rs = new NoopRequesterService();
```
htran1: Please try injection for this so that it is not tied to a single global instance. Currently we don't have a use for multiple instances, but I think the injection will be cleaner.

PR #2514, opened 2018-12-05
GOBBLIN-644: Add metrics reporting config dynamically to compiled flows in MultiHopFlowCompiler.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-644

### Description
Add metrics reporting config dynamically to the compiled flows to avoid having to specify the config for each job template. 

### Tests
Tested in local environment with end-to-end execution.

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/spec/JobExecutionPlan.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/spec/JobExecutionPlan.java
@@ -38,6 +38,7 @@ import org.apache.gobblin.runtime.api.FlowSpec;
 import org.apache.gobblin.runtime.api.JobSpec;
 import org.apache.gobblin.runtime.api.SpecExecutor;
 import org.apache.gobblin.service.ExecutionStatus;
+import org.apache.gobblin.service.ServiceConfigKeys;
 import org.apache.gobblin.service.modules.flowgraph.FlowGraphConfigurationKeys;
 import org.apache.gobblin.service.modules.orchestration.DagManager;
 import org.apache.gobblin.service.modules.template_catalog.FSFlowCatalog;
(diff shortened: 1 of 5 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/spec/JobExecutionPlan.java
```diff
@@ -54,11 +57,11 @@
 
   public static class Factory {
     public static final String JOB_NAME_COMPONENT_SEPARATION_CHAR = "_";
-    private static final Random random = new Random();
+    public static final String METRICS_REPORTING_CONFIG_PREFIX = "metrics.reporting";
```
kyuamazon: Is this the scope name? Can we have a more scoped name like "gobblin.service.metrics.reporting"?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/spec/JobExecutionPlan.java
```diff
@@ -115,12 +118,42 @@ private static JobSpec buildJobSpec(FlowSpec flowSpec, Config jobConfig, Long fl
...
+     * A method to add tracking event configurations to a JobSpec.
+     * This enables {@link org.apache.gobblin.metrics.GobblinTrackingEvent}s
+     * to be emitted from each Gobblin job orchestrated by Gobblin-as-a-Service, which will then be used for tracking the
+     * execution status of the job.
+     * @param jobSpec representing a fully resolved {@link JobSpec}.
+     */
+    private static void addTrackingEventConfig(JobSpec jobSpec, Config sysConfig) {
+      Config reportingConfig = ConfigUtils.getConfig(sysConfig, METRICS_REPORTING_CONFIG_PREFIX, ConfigFactory.empty());
+      if (!reportingConfig.isEmpty()) {
+        Config jobConfig = jobSpec.getConfig().withFallback(reportingConfig.atPath(METRICS_REPORTING_CONFIG_PREFIX));
```
kyuamazon: Why the fallback is required here? If user doesn't specify the metrics reporter, is it okay to reuse the gaas metrics reporter? I have concern because one is reporting from PROD network, but one is from Azkaban. They might have different settings.

PR #2525, opened 2018-12-17
[GOBBLIN-656] Add filtering of table format in HiveDatasetFinder

autumnust:
Dear Gobblin maintainers,

### JIRA
- [x] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-656] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-656

### Description
- [x] In some of pipelines using `HiveDatasetFinder` we are asked to have functionalities to reject specific types of tables. The rejection criteria could include the table format which is reflected from SerDe library it uses.  

### Tests
- [x] Unit Tested by the added case. 

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDatasetFinder.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDatasetFinder.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.data.management.copy.hive;
 
+import com.google.common.base.Enums;
 import com.google.common.base.Throwables;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


autumnust:
@htran1  Can you help review? Thanks.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDatasetFinder.java
```diff
@@ -181,6 +190,11 @@ protected HiveDatasetFinder(FileSystem fs, Properties properties, HiveMetastoreC
       throw new RuntimeException(e);
     }
     this.jobConfig = ConfigUtils.propertiesToConfig(properties);
+
+    String rejectedFormatSerDe = properties.getProperty(REJECTED_DATA_FORMAT_KEY, StringUtils.EMPTY);
+    Optional<HiveSerDeWrapper.BuiltInHiveSerDe> serDe =
```
htran1: Why not support a list of rejected data formats?

Also, how about extracting out the SerDe class name as a string when an enum match is found, but if one is not found then use the string as provoided? This will allow SerDes not in BuiltInHiveSerDe to be supported.

PR #2533, opened 2019-01-10
GOBBLIN-662: Enhance SSH-based access to Git to enable/disable host key checking.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-662

### Description
We expose configurations to allow users to enable or disable host key checking and provide known hosts file or string. We also provide configs to supply public/private key as base64 encoded strings.  

### Tests
Tested in local environment. 

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
@@ -90,7 +90,6 @@ public abstract class GitMonitoringService extends AbstractIdleService {
 
   private String privateKeyPath;
   private byte[] privateKey;
-  private byte[] publicKey;
   private String passphrase;
   private boolean isJschLoggerEnabled;
   private boolean strictHostKeyCheckingEnabled;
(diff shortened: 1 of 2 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
```diff
@@ -426,7 +450,18 @@ protected JSch createDefaultJSch(FS fs) throws JSchException {
           JSch.setLogger(new JschLogger());
         }
         JSch defaultJSch = super.createDefaultJSch(fs);
-        defaultJSch.addIdentity(GitMonitoringService.this.privateKeyPath, GitMonitoringService.this.passphrase);
+        defaultJSch.getIdentityRepository().removeAll();
+        if (GitMonitoringService.this.privateKeyPath != null) {
+          defaultJSch.addIdentity(GitMonitoringService.this.privateKeyPath, GitMonitoringService.this.passphrase);
```
htran1: Why is the path case only passing the private key, but the other case is passing both public and private? Shouldn't both cases be the same or if only private key is required then the second case should only pass the private key?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
```diff
@@ -114,16 +120,32 @@
...
-        throw new RuntimeException("Path to private key must be provided");
+      this.privateKeyPath = ConfigUtils.getString(config, ConfigurationKeys.GIT_MONITOR_SSH_PRIVATE_KEY_PATH, null);
+      String privateKeyBase64Encoded = ConfigUtils.getString(config, ConfigurationKeys.GIT_MONITOR_SSH_PRIVATE_KEY_BASE64_ENCODED, null);
+      String publicKeyBase64Encoded = ConfigUtils.getString(config, ConfigurationKeys.GIT_MONITOR_SSH_PUBLIC_KEY_BASE64_ENCODED, null);
+
+      if ((Strings.isNullOrEmpty(this.privateKeyPath)) && ((Strings.isNullOrEmpty(privateKeyBase64Encoded)) || (Strings.isNullOrEmpty(publicKeyBase64Encoded)))) {
+        throw new RuntimeException("Path to private key or private/public key strings must be provided");
+      }
+
+      if (!Strings.isNullOrEmpty(privateKeyBase64Encoded)) {
```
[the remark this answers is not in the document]
sv2000: Removed the public key check, since the public key string is unnecessary.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
sv2000: With private key path, Jsch by default looks for the corresponding .pub file in the same dir as the private key and hence it is not needed in one of the APIs.

PR #2542, opened 2019-01-28
[GOBBLIN-672] add EventHub source

fabi54:


gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubExtractor.java
```diff
@@ -0,0 +1,210 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Iterator;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Created by Jan on 23-Jan-19.
```
shirshanka: Could you add more descriptive javadoc?

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubExtractor.java
```diff
@@ -0,0 +1,210 @@
...
+            final ConnectionStringBuilder connStr = new ConnectionStringBuilder()
+                    .setEndpoint(endPoint)
+                    .setEventHubName(eventHubName)
+                    .setSasKeyName(sasKeyName)
+                    .setSasKey(sasKey);
+
+            executorService = Executors.newSingleThreadScheduledExecutor();
+            ehClient = EventHubClient.createSync(connStr.toString(), executorService);
+            final EventHubRuntimeInformation eventHubInfo = ehClient.getRuntimeInformation().get();
+            final String partitionId = eventHubInfo.getPartitionIds()[0]; // get first partition's id
```
shirshanka: why just the first?

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubSource.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.gobblin.source.workunit.WorkUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.*;
+
+
+/**
+ * Created by Jan on 23-Jan-19.
```
shirshanka: More descriptive javadoc please!

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubSource.java
```diff
@@ -0,0 +1,152 @@
...
+        for (int i = 0; i < numOfMultiWorkunits; i++) {
+            result.add(MultiWorkUnit.createEmpty());
+        }
+        for (int i = 0; i < workUnits.size(); i++) {
+            ((MultiWorkUnit) result.get(i % numOfMultiWorkunits)).addWorkUnit(workUnits.get(i));
+        }
+        return result;
+    }
+
+    protected List<WorkUnit> generateWorkUnits(SourceState state, long previousWatermark) {
```
yukuai518: I think we should not have a global previousWatermark. If you only support one partition, this will work. But if you have multiple partitions, each partition should have its own previousWatermark. I guess right now your source is much like a QueryBasedSource instead of a partition like source such as KafkaSource. You need to check if you can model it in the same way as KafkaSource, which maintains previousOffsets list for each workunit.

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubSource.java
```diff
@@ -0,0 +1,152 @@
...
+        List<WorkUnit> workUnits = new ArrayList<>();
+
+        state.setProp(ConfigurationKeys.EXTRACT_DELTA_FIELDS_KEY, "offset");
+        state.setProp(ConfigurationKeys.SOURCE_QUERYBASED_WATERMARK_TYPE, "simple");
+        state.setProp(ConfigurationKeys.SOURCE_QUERYBASED_EXTRACT_TYPE, "snapshot");
+        List<Partition> partitions = new Partitioner(state).getPartitionList(previousWatermark);
```
yukuai518: Like I mentioned, there shouldn't be a global previous watermark. Right now if you have one partition, it will cut one partition range into multiple pieces, which doesn't have any ordering guarantees. This is not what we want.

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubExtractor.java
```diff
@@ -0,0 +1,210 @@
...
+            logStatus();
+            iterator = null;
+            if (endTime != null && LocalDateTime.now(ZoneId.of("UTC")).isBefore(endTime)) {
+                LOG.info("Attempting to receive records from source.");
+                Iterable<EventData> received = receiver.receiveSync(eventsInBatch);
+                if (received != null) {
+                    iterator = received.iterator();
+                    LOG.info(String.format("Received batch size: %s", received.spliterator().getExactSizeIfKnown()));
+                } else {
+                    LOG.info("No records received from source.");
```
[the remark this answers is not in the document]
yukuai518: Also, if no records were found, the old iterator object will be returned here, the readRecord() will fail in the .next() method call. You may want to reset the iterator to null, or have another flag to indicate the reading can be stopped.

jhsenjaliya:
@fabi54, are you going to address the comments?

fabi54:
@jhsenjaliya Yes, I've seen the comments and I'm working on improvements.

gobblin-modules/gobblin-eventhub/src/main/java/org/apache/gobblin/eventhub/source/EventhubExtractor.java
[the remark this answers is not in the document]
fabi54: Iterator is reset to null at the beginning of the method, so null iterator will be returned (not the old iterator). This is then handled in readRecord(), so it won't fail.

PR #2548, opened 2019-02-07
[GOBBLIN-677] - Allow early termination of Gobblin jobs based on a predicate on the job progress

ibuenros:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-677

### Description

This PR adds the possibility of terminating a job early once a predicate (based on the job progress is satisfied). 

There are four main changes:
- `ReflectivePredicateEvaluator` is an evaluator for SQL queries which takes as an input a list of objects implementing a reference type. Each object is considered a row, and each public method in the reference type is a column. The evaluator runs the expression on the list of objects and expects a result with a single boolean column, which is then `AND`'d to obtain the result of the predicate. This is powered by Calcite, and can be used for expressive policies in a number of use cases.
- `FiniteStateMachine` is a thread-safe finite state machine implementation that gates certain actions on the current state of the machine and whether a transition is legal. It is used, for example, to manage the state of the MR job launcher and make sure that interruptions and cancels work correctly regardless of whether the MR job is pending, running, or already finished.
- Actual early termination logic. `JobInterruptionPredicate` used a `ReflectivePredicateEvaluator` with base type `JobProgress` to evaluate a predicate on the job progress every 30 seconds (for example, `SELECT elapsedTime > (30 * 60 * 1000) AND completedTasks >= totalTasks * 0.75 FROM jobProgress`). Once the predicate is satisfied, it calls a shutdown trigger specific to the current job launcher.
- For `MRJobLauncher`, the early termination is driven via creating a specific file in the `mrJobDir`. The Gobblin mappers, once they find this file, indicate to the `Extractor` that they should terminate. If the `Extractor` understands this command, it is supposed to do a graceful early termination of extraction, committing watermarks, etc. If it does not recognize it, an exception is thrown triggering a task failure. An implementation for graceful early termination of `KafkaExtractor` is provided.

### Tests

There are a bunch of unit tests for the individual components. Additionally, tests where run on actual Gobblin executions.

### Commits


ibuenros:
@htran1 @sv2000 can you take a look?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -170,6 +181,12 @@ public MRJobLauncher(Properties jobProps, Configuration conf, SharedResourcesBro
       throws Exception {
     super(jobProps, metadataTags);
 
+    this.fsm = new FiniteStateMachine.Builder<MRJobLauncherState>()
```
sv2000: This FSM instance (and the associated states) looks generic enough that it can be moved out of this class. Can we have a FSM factory class that builds out specific FSM types and hands it back to the caller?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -286,17 +316,57 @@ protected void runWorkUnits(List<WorkUnit> workUnits) throws Exception {
...
+          transition.changeEndState(MRJobLauncherState.FAILED);
+        }
+      }
+    } catch (FiniteStateMachine.UnallowedTransitionException | InterruptedException exc) {
+      LOG.error("Failed to cancel job " + this.jobContext.getJobId(), exc);
+    }
+  }
+
+  /**
+   * Attempt a gracious interruptiong of the running job
```
sv2000: Spelling error - interruption instead of interruptiong

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -252,26 +270,38 @@ protected void runWorkUnits(List<WorkUnit> workUnits) throws Exception {
...
+      } catch (FiniteStateMachine.UnallowedTransitionException unallowed) {
+        LOG.info("Cannot start MR job.", unallowed);
       }
 
-      TimingEvent mrJobRunTimer = this.eventSubmitter.getTimingEvent(TimingEvent.RunJobTimings.MR_JOB_RUN);
-      LOG.info(String.format("Waiting for Hadoop MR job %s to complete", this.job.getJobID()));
-      this.job.waitForCompletion(true);
-      mrJobRunTimer.stop(ImmutableMap.of("hadoopMRJobId", this.job.getJobID().toString()));
+      if (this.fsm.getCurrentState().equals(MRJobLauncherState.RUNNING)) {
+        JobInterruptionPredicate jobInterruptionPredicate = new JobInterruptionPredicate(jobState, this::interruptJob, true);
```
sv2000: Can the JobInterruptionPredicate be made part of the FSM definition itself? I am thinking in general, each state of the FSM could potentially have a Predicate service of its own. The service is started/stopped as states change.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -679,6 +751,12 @@ protected void setup(Context context) {
     @Override
     public void run(Context context) throws IOException, InterruptedException {
       this.setup(context);
+
+      Path interruptPath = new Path(context.getConfiguration().get(GOBBLIN_JOB_INTERRUPT_PATH_KEY));
+      if (this.fs.exists(interruptPath)) {
```
sv2000: A log message here will be useful for debugging purposes.

gobblin-runtime/src/test/java/org/apache/gobblin/util/ReflectivePredicateEvaluatorTest.java
```diff
@@ -0,0 +1,108 @@
...
+					new ReflectivePredicateEvaluator("SELECT anInt = 1 FROM myInterface WHERE aString = 'foo'",
+							MyInterface.class);
+			evaluator.evaluate(new MyImplementation(1, "bar"));
+			Assert.fail();
+		} catch (IllegalArgumentException exc) {
+			// Expected
+		}
+	}
+
+	public interface MyInterface {
```
sv2000: Can this be private?

gobblin-utility/src/main/java/org/apache/gobblin/fsm/FiniteStateMachine.java
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	/**
+	 * Get the current state. This method will wait until the FSM is in a non-transitioning state (although a transition
+	 * may start immediately after).
+	 * @throws InterruptedException if the thread got interrupted while waiting for a non-transitioning state.
+	 */
+	public T getCurrentState() throws InterruptedException {
+		try {
+			this.lock.lock();
```
sv2000: Why do we need a lock for getting current state? If we do indeed need a lock, can we have a ReadWriteLock that will allow multiple concurrent readers?

gobblin-utility/src/main/java/org/apache/gobblin/fsm/FiniteStateMachine.java
```diff
@@ -0,0 +1,357 @@
...
+			return this.currentState;
+		} finally {
+			this.lock.unlock();
+		}
+	}
+
+	@VisibleForTesting
+	T getCurrentStateEvenIfTransitioning() {
+		try {
+			this.lock.lock();
```
sv2000: Same comment as above. Maybe add javadoc, why a lock is needed to read state.

ibuenros:
@sv2000 addressed comments, please take another look?

sv2000:
@ibuenros can you ensure that the travis build is successful before @htran1 can merge this PR?

PR #2555, opened 2019-02-14
[GOBBLIN-683] Add azkaban client retry logic.

yukuai518:
Dear Gobblin maintainers,

### JIRA

### Description
   Retry azkaban client if session expires.
   Move all the client logic to AzkabanMultiCallables. Each Callable can be wrapped into a Retryer for better retry support.

### Tests
   Use Azkaban headless challenge authentication to verify the retry succeeded after a manual created invalid session exception gets caught.

### Commits

### DIFF
--- a/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
+++ b/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
@@ -113,7 +113,7 @@ public class AzkabanClient implements Closeable {
         .retryIfExceptionOfType(InvalidSessionException.class)
         .withAttemptTimeLimiter(AttemptTimeLimiters.fixedTimeLimit(10, TimeUnit.SECONDS, this.executorService))
         .withWaitStrategy(WaitStrategies.exponentialWait(60, TimeUnit.SECONDS))
-        .withStopStrategy(StopStrategies.stopAfterAttempt(5))
+        .withStopStrategy(StopStrategies.stopAfterAttempt(3))
         .build();
     this.sessionId = this.sessionManager.fetchSession();
     this.sessionCreationTime = System.nanoTime();
@@ -137,7 +137,7 @@ public class AzkabanClient implements Closeable {
 
   private void intializeExecutorService() {
     if (this.executorService == null) {
-      this.executorService = Executors.newFixedThreadPool(300);
+      this.executorService = Executors.newFixedThreadPool(30);
     }
   }
 
--- a/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
+++ b/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
@@ -25,6 +25,8 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.IOUtils;
@@ -44,18 +46,24 @@ import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.rholder.retry.AttemptTimeLimiters;
 import com.github.rholder.retry.RetryException;
 import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.StopStrategies;
 import com.github.rholder.retry.WaitStrategies;
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
+import com.google.common.io.Closer;
+import com.google.common.util.concurrent.SimpleTimeLimiter;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 
 import lombok.Builder;
+
+import org.apache.gobblin.util.ExecutorsUtils;
 
 
 /**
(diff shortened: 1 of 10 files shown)


yukuai518:
@sv2000 Please review

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -402,97 +329,71 @@ public AzkabanExecuteFlowStatus executeFlowWithOptions(
...
   }
 
-  private void addFlowOptions(List<NameValuePair> nvps, Map<String, String> flowOptions) {
-    if (flowOptions != null) {
-      for (Map.Entry<String, String> option : flowOptions.entrySet()) {
-        log.debug("New flow option added:" + option .getKey()+ "-->" + option.getValue());
-        nvps.add(new BasicNameValuePair(option.getKey(), option.getValue()));
+  private <T> T runWithRetry(Callable callable, Class<T> cls) throws AzkabanClientException {
+    try {
+      AzkabanClientStatus status = this.retryer.call(callable);
```
sv2000: How will retry help for the situation when Azkaban server reports that a session is expired? The problem is that our refresh interval could be much longer e.g. 12h, but Az server reports that session is invalid after 10 mts. In this case, retry will not be much useful. We will need to catch the exception and forcibly refresh the session.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -88,19 +96,23 @@ protected AzkabanClient(String username,
...
+    this.executorService = executorService;
     this.initializeClient();
     this.initializeSessionManager();
+    this.intializeExecutorService();
     this.retryer = RetryerBuilder.<AzkabanClientStatus>newBuilder()
         .retryIfExceptionOfType(InvalidSessionException.class)
-        .withWaitStrategy(WaitStrategies.exponentialWait(10, TimeUnit.SECONDS))
+        .withAttemptTimeLimiter(AttemptTimeLimiters.fixedTimeLimit(10, TimeUnit.SECONDS, this.executorService))
+        .withWaitStrategy(WaitStrategies.exponentialWait(60, TimeUnit.SECONDS))
         .withStopStrategy(StopStrategies.stopAfterAttempt(5))
```
sv2000: 3 attempts should be sufficient I think.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -123,6 +135,12 @@ private void initializeSessionManager() throws AzkabanClientException {
     }
   }
 
+  private void intializeExecutorService() {
+    if (this.executorService == null) {
+      this.executorService = Executors.newFixedThreadPool(300);
```
sv2000: 300 threads may be too much. Is 30 too low? Also, do you think some of the retryer settings be made configurable?

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -88,19 +96,23 @@ protected AzkabanClient(String username,
...
     this.httpClient = httpClient;
     this.sessionManager = sessionManager;
+    this.executorService = executorService;
     this.initializeClient();
     this.initializeSessionManager();
+    this.intializeExecutorService();
     this.retryer = RetryerBuilder.<AzkabanClientStatus>newBuilder()
         .retryIfExceptionOfType(InvalidSessionException.class)
-        .withWaitStrategy(WaitStrategies.exponentialWait(10, TimeUnit.SECONDS))
+        .withAttemptTimeLimiter(AttemptTimeLimiters.fixedTimeLimit(10, TimeUnit.SECONDS, this.executorService))
```
sv2000: Is this really needed? I am a little concerned about creating executor service with a fixed thread pool of 300 threads per client. With multiple Azkaban executor instances, this can lead to a lot of threads.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
yukuai518: I force a retry now. Please review the latest

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
yukuai518: Sure, I can make it configurable

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
yukuai518: Will change it

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
yukuai518: The decision would be based on if you want to share the Azkaban producer or not, if you want to share the producer, then 300 is okay, but if you don't want to share, each job request will have its own AzkabanClient, then you could use SingleThreadPool. It depends on the implementation, that's why I give user option to setExecutorService() by themselves if they want to change the default behavior.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
yukuai518: @sv2000 I change it back to 30 now, the whole idea is to let user customize their own executorService which meets their requirement. So here is just a default fallback.

PR #2556, opened 2019-02-15
GOBBLIN-684: Ensure buffered messages are flushed before close() in K…

sv2000:
…afkaProducerPusher.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-684

### Description
Currently, when KafkaProducerPusher is closed, it invokes KafkaProducer#close(). However,close() only guarantees delivery of in-flight messages, not the messages in the producer buffer waiting to be sent out. This results in data loss.

The fix ensures that we call flush() before close(). As a result, any buffered messages are immediately pushed out and we block until the messages are acked. 

### Tests
Tested in real deployment.

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
+++ b/gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
@@ -51,8 +51,15 @@ public class KafkaKeyValueProducerPusher<K, V> implements Pusher<Pair<K, V>> {
   private final String topic;
   private final KafkaProducer<K, V> producer;
   private final Closer closer;
+  //Queue to keep track of the futures returned by the Kafka asynchronous send() call. The futures queue is used
+  // to mimic the functionality of flush() call (available in Kafka 09 and later). Currently, there are no
+  // capacity limits on the size of the futures queue. In general, if queue capacity is enforced, a safe lower bound for queue
+  // capacity is MAX_NUM_FUTURES_TO_BUFFER + (numThreads * maxNumMessagesPerInterval), where numThreads equals the number of
+  // threads sharing the producer instance and maxNumMessagesPerInterval is the estimated maximum number of messages
+  // emitted by a thread per reporting interval.
   private final Queue<Future<RecordMetadata>> futures = new LinkedBlockingDeque<>();
 
+  //Low watermark for the size of the futures queue, to trigger flushing of messages.
   private static final long MAX_NUM_FUTURES_TO_BUFFER = 1000L;
 
   public KafkaKeyValueProducerPusher(String brokers, String topic, Optional<Config> kafkaConfig) {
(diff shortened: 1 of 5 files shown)


gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
```diff
@@ -80,17 +88,46 @@ public KafkaKeyValueProducerPusher(String brokers, String topic) {
...
         }
-      });
+      }));
+    }
+
+    //Accumulate futures returned from send() into a buffer; will be used to simulate flush by calling get() on
+    // each of the accumulated futures.
+    if (this.futures.size() >= MAX_NUM_FUTURES_TO_BUFFER) {
+      flush(MAX_NUM_FUTURES_TO_BUFFER);
+      this.futures.clear();
```
htran1: This won't be thread safe since if there are concurrent calls to pushMessages() some unacknowledged futures may be cleared. `flush` is already removing with `poll`, so I think this can be removed.

Also, this forces a synchronous wait every 1000 messages. Should there be different max buffer and flush amounts to avoid waiting for the newest messages?

gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
```diff
@@ -47,6 +51,9 @@
   private final String topic;
   private final KafkaProducer<K, V> producer;
   private final Closer closer;
+  private final Queue<Future<RecordMetadata>> futures = new LinkedBlockingDeque<>();
```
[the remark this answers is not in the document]
sv2000: Thanks! Added more comments to explain how to set the capacity for the buffer.

gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
sv2000: Good catch! Yes, the clear() is not needed. Also, changed the pushMessages() method to exclude the newest added messages.

gobblin-modules/gobblin-kafka-08/src/main/java/org/apache/gobblin/metrics/kafka/KafkaKeyValueProducerPusher.java
```diff
@@ -80,17 +98,48 @@ public KafkaKeyValueProducerPusher(String brokers, String topic) {
...
+
+  /**
+   * Flush any records that may be present in the producer buffer upto a maximum of <code>numRecordsToFlush</code>.
+   * This method is needed since Kafka 0.8 producer does not have a flush() API. In the absence of the flush()
+   * implementation, records which are present in the buffer but not in-flight may not be delivered at all when close()
+   * is called, leading to data loss.
+   * @param numRecordsToFlush
+   */
+  private void flush(long numRecordsToFlush) {
+    log.info("Flushing records from producer buffer");
```
htran1: Can you make this log and the one at the end debug level or print it only when `numRecordsToFlush == Long.MAX_VALUE`? I think this may flood the log since flush() is called whenever the futures queue has hit the limit.

PR #2559, opened 2019-02-20
GOBBLIN-687: Pass TopologySpec map to DagManager to allow reuse of Sp…

sv2000:
…ecExecutors during DAG deserialization.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-687

### Description
DagManager maintains state of all currently executing DAGs, by serializing each DAG on compilation and persisting it to a durable store. The serialized DAG includes Job config as well as the SpecExecutor config for each job in the DAG. This is done to correctly resume execution of DAGs in case of service restarts or leadership change. 

Currently, on service restart/leadership change, the new master de-serializes SpecExecutor config and creates a SpecExecutor instance for each job in the DAG. If the number of DAGs is large, this can result in many connections to the underlying executor instance. The proposed fix allows the DagManager to re-use the SpecExecutor instances created by the TopologySpecFactory when it deserializes a DAG.

### Tests
Enhanced unit tests in FSDagStateStoreTest and tested in local environment.

### Commits


sv2000:
@autumnust please review this PR.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -107,6 +104,10 @@ public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Op
         | ClassNotFoundException e) {
       throw new RuntimeException(e);
     }
+    //Pass the topologySpecMap to the DagManager
```
autumnust: Shall we move this block under line 93?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -218,6 +211,12 @@ public GobblinServiceManager(String serviceName, String serviceId, Config config
       this.helixManager = Optional.absent();
     }
 
+    // Initialize DagManager
```
autumnust: Initially `DagManager` is only initialized when `isFlowCatalogEnabled` set to true, but moving it out means this condition doesn't apply anymore.  Is that on purpose ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -354,7 +354,10 @@ private void handleLeadershipChange(NotificationContext changeContext) {
       }
 
       if (this.isDagManagerEnabled) {
-        this.dagManager.setActive(true);
+        //Activate DagManager only if TopologyCatalog is initialized. If not; skip activation.
+        if (this.topologyCatalog.getInitComplete().getCount() <= 0) {
```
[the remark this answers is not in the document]
sv2000: Changed it to == 0.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
sv2000: We need this after the spec compiler initialization, to make sure that the topologySpecMap has been instantiated. Added comment to make it explicit.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
sv2000: Thanks for noticing this! The DagManager initialization does not depend on whether FlowCatalog is enabled or not. We have a separate config to enable/disable DagManager. I can't recall why it was inside if (isFlowCatalogEnabled). Yes, the move outside the if block is intentional.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -232,6 +231,7 @@ public GobblinServiceManager(String serviceName, String serviceId, Config config
       this.serviceLauncher.addService(this.scheduler);
     }
 
+
```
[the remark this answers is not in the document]
sv2000: Done.

sv2000:
@autumnust addressed your comments. Can you please take another look?

autumnust:
@sv2000 LGTM.

PR #2568, opened 2019-03-05
[GOBBLIN-697] Implementation of data file versioning and preservation in distcp.

ibuenros:
Dear Gobblin maintainers,

### JIRA

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
@@ -88,8 +88,8 @@ public class CopyDataPublisher extends DataPublisher implements UnpublishedHandl
   protected final EventSubmitter eventSubmitter;
   protected final RecoveryHelper recoveryHelper;
   protected final Optional<LineageInfo> lineageInfo;
-  protected final DataFileVersionStrategy srcDataFileVersion;
-  protected final DataFileVersionStrategy dstDataFileVersion;
+  protected final DataFileVersionStrategy srcDataFileVersionStrategy;
+  protected final DataFileVersionStrategy dstDataFileVersionStrategy;
 
   /**
    * Build a new {@link CopyDataPublisher} from {@link State}. The constructor expects the following to be set in the
@@ -128,9 +128,9 @@ public class CopyDataPublisher extends DataPublisher implements UnpublishedHandl
 
     Config config = ConfigUtils.propertiesToConfig(state.getProperties());
 
-    this.srcDataFileVersion = DataFileVersionStrategy
-        .instantiateDataFileVersion(HadoopUtils.getSourceFileSystem(state), config);
-    this.dstDataFileVersion = DataFileVersionStrategy.instantiateDataFileVersion(this.fs, config);
+    this.srcDataFileVersionStrategy = DataFileVersionStrategy
+        .instantiateDataFileVersionStrategy(HadoopUtils.getSourceFileSystem(state), config);
+    this.dstDataFileVersionStrategy = DataFileVersionStrategy.instantiateDataFileVersionStrategy(this.fs, config);
   }
 
   @Override
@@ -233,9 +233,9 @@ public class CopyDataPublisher extends DataPublisher implements UnpublishedHandl
         CopyableFile copyableFile = (CopyableFile) copyEntity;
 
         if (copyableFile.getPreserve().preserve(PreserveAttributes.Option.VERSION)
-            && this.dstDataFileVersion.hasCharacteristic(DataFileVersionStrategy.Characteristic.SETTABLE)) {
-          this.dstDataFileVersion.setVersion(copyableFile.getDestination(),
-              this.srcDataFileVersion.getVersion(copyableFile.getOrigin().getPath()));
+            && this.dstDataFileVersionStrategy.hasCharacteristic(DataFileVersionStrategy.Characteristic.SETTABLE)) {
+          this.dstDataFileVersionStrategy.setVersion(copyableFile.getDestination(),
+              this.srcDataFileVersionStrategy.getVersion(copyableFile.getOrigin().getPath()));
         }
 
         if (wus.getWorkingState() == WorkingState.COMMITTED) {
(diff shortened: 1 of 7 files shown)


ibuenros:
@yukuai518 @htran1 can you review? Thanks!

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/DataFileVersion.java
```diff
@@ -0,0 +1,127 @@
...
+ * at time 1 data is copied to location B, sync mechanism might incorrectly believe that since mod time of location B
+ * is higher, it should be synced back to location A, etc.
+ *
+ * Required properties:
+ * - REPLICABLE: Two calls to `getVersion` on a file that has clearly not been modified must return the same version.
+ * - MONOTONOUS: The default version of a file is an increasing function of modification time.
+ * - CONSERVATIVE: If file f had its version last set to v, but the versioning implementation determines the file MIGHT
+ *                 have been modified and it chooses to return a version, it will return a value strictly larger than v.
+ *
+ * A common pattern to achieve monotonicity and conservativeness will be to invalidate the version of a data file
```
yukuai518: What does 'invalidate' imply here? Does that infer the getVersion will return a new version? In this case if we have STRICT will it return exception?

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/ModTimeDataFileVersion.java
```diff
@@ -0,0 +1,66 @@
...
+import com.typesafe.config.Config;
+
+import lombok.Data;
+
+
+/**
+ * An implementation of {@link DataFileVersion} that uses modtime as the file version.
+ */
+@Data
+public class ModTimeDataFileVersion implements DataFileVersion<Long> {
```
yukuai518: When do we use this ModTimeDataFileVersion? I remember in the design document (https://docs.google.com/document/d/1VhCHE2xfmwf4WTfle_wFhz-1U3TttIWDMU_YL0jkUbE/edit), we will have 'version string' written by Purger, but at the same time we will also have a 'version timestamp' that is equivalent to modTime in order to detect the situation where a file is modified without updating the version. For both cases, we need both versions to be set in the XAttr, but here the setVersion returns false, did I miss something?

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/DataFileVersion.java
```diff
@@ -0,0 +1,127 @@
...
+   */
+  boolean setVersion(Path path, T version) throws IOException;
+
+  /**
+   * Set the version of a path to a value automatically set by the versioning system. Note this call must respect the
+   * monotonicity requirement.
+   *
+   * @return whether the version was set successfully
+   */
+  boolean setDefaultVersion(Path path) throws IOException;
```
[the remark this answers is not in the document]
ibuenros: In some way there are two operations you would want to do with versioning: set a version to the "current" or "default" value, and propagate the version from one file to another. 

For setting the version, the caller might not know which strategy the user chose, but the version might have a way to determine what the version of a file might be (e.g. current timestamp, or look up in a registry, etc.), so the caller would just call `setDefaultVersion`.

For propagating a version, you can just get the version of one file, and apply it to the new file, without knowledge about what strategy the user chose.

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/DataFileVersion.java
ibuenros: It means to fallback to default version if anything external has modified the file I guess? I can rephrase this.

STRICT does mean that if a modification was made, `getVersion` will return an error to indicate that.

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/ModTimeDataFileVersion.java
ibuenros: Modtime is a sensible default for versioning strategy, and lets distcp work on versions comparison instead of modtime comparison. If no alternate strategy is specified, distcp just falls back to what it is doing now (because the default version is just the modtime), but if an alternate is specified, it will respect it.

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/DataFileVersion.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * A Factory for {@link DataFileVersion}s.
+   */
+  interface DataFileVersionFactory<T extends Comparable<T> & Serializable> {
+    /**
+     * Build a {@link DataFileVersion} with the input configuration.
+     */
+    DataFileVersion<T> createDataFileVersionStrategy(FileSystem fs, Config config);
```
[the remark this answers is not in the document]
ibuenros: That makes sense. Fixed.

gobblin-utility/src/main/java/org/apache/gobblin/util/filesystem/DataFileVersion.java
```diff
@@ -0,0 +1,127 @@
...
+   * Get the version of a path.
+   */
+  T getVersion(Path path) throws IOException;
+
+  /**
+   * Set the version of a path to a specific version (generally replicated from another path).
+   *
+   * @return whether the version was set successfully
+   */
+  boolean setVersion(Path path, T version) throws IOException;
```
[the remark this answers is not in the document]
ibuenros: I see the point. I added info to the comment clarifying when to return `false` vs throwing exception.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -119,6 +125,12 @@ public CopyDataPublisher(State state) throws IOException {
 
     this.recoveryHelper = new RecoveryHelper(this.fs, state);
     this.recoveryHelper.purgeOldPersistedFile();
+
+    Config config = ConfigUtils.propertiesToConfig(state.getProperties());
+
+    this.srcDataFileVersion = DataFileVersionStrategy
```
yukuai518: this should be this.srcDataFileVersionStrategy?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -119,6 +125,12 @@ public CopyDataPublisher(State state) throws IOException {
...
+        .instantiateDataFileVersion(HadoopUtils.getSourceFileSystem(state), config);
+    this.dstDataFileVersion = DataFileVersionStrategy.instantiateDataFileVersion(this.fs, config);
```
yukuai518: same dstDataFileVersion => dstDataFileVersionStrategy?

ibuenros:
@yukuai518 addressed comments. @htran1 can you take a look and merge when you get a chance?

PR #2576, opened 2019-03-21
[GOBBLIN-706]enable dynamic mappers

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-706

### Description
Today the number of mappers is hard coded and almost always reached, however, many mappers do very little work. This tends to cause small files and short jobs that are overhead dominated. The improvement here is to enable user add a job property named target.mapper.size to set the target mapper size and Gobblin will dynamically scale the number of mappers up and down depending on the total load. If the property is not set, the number of mappers will still be the vale of mr.job.max.mappers.

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -617,6 +617,7 @@ public class ConfigurationKeys {
   /** Specifies a static location in HDFS to upload jars to. Useful for sharing jars across different Gobblin runs.*/
   public static final String MR_JARS_DIR = "mr.jars.dir";
   public static final String MR_JOB_MAX_MAPPERS_KEY = "mr.job.max.mappers";
+  public static final String TARGET_MAPPER_SIZE = "target.mapper.size";
   public static final String MR_REPORT_METRICS_AS_COUNTERS_KEY = "mr.report.metrics.as.counters";
   public static final boolean DEFAULT_MR_REPORT_METRICS_AS_COUNTERS = false;
   public static final int DEFAULT_MR_JOB_MAX_MAPPERS = 100;
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -617,7 +617,7 @@ public class ConfigurationKeys {
   /** Specifies a static location in HDFS to upload jars to. Useful for sharing jars across different Gobblin runs.*/
   public static final String MR_JARS_DIR = "mr.jars.dir";
   public static final String MR_JOB_MAX_MAPPERS_KEY = "mr.job.max.mappers";
-  public static final String TARGET_MAPPER_SIZE = "target.mapper.size";
+  public static final String MR_TARGET_MAPPER_SIZE = "mr.target.mapper.size";
   public static final String MR_REPORT_METRICS_AS_COUNTERS_KEY = "mr.report.metrics.as.counters";
   public static final boolean DEFAULT_MR_REPORT_METRICS_AS_COUNTERS = false;
   public static final int DEFAULT_MR_JOB_MAX_MAPPERS = 100;
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -264,12 +264,11 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
       KafkaWorkUnitPacker kafkaWorkUnitPacker = KafkaWorkUnitPacker.getInstance(this, state);
       int numOfMultiWorkunits = maxMapperNum;
       if(state.contains(ConfigurationKeys.MR_TARGET_MAPPER_SIZE)) {
-        double totalEstDataSize = kafkaWorkUnitPacker.getWorkUnitEstSizes(workUnits);
+        double totalEstDataSize = kafkaWorkUnitPacker.setWorkUnitEstSizes(workUnits);
         LOG.info(String.format("The total estimated data size is %.2f", totalEstDataSize));
         double targetMapperSize = state.getPropAsDouble(ConfigurationKeys.MR_TARGET_MAPPER_SIZE);
-        numOfMultiWorkunits = (int) (totalEstDataSize / targetMapperSize);
-        numOfMultiWorkunits = numOfMultiWorkunits>maxMapperNum? maxMapperNum:numOfMultiWorkunits;
-        numOfMultiWorkunits = numOfMultiWorkunits==0? 1: numOfMultiWorkunits;
+        numOfMultiWorkunits = (int) (totalEstDataSize / targetMapperSize) + 1;
+        numOfMultiWorkunits = Math.min(numOfMultiWorkunits, maxMapperNum);
       }
       List<WorkUnit> workUnitList = kafkaWorkUnitPacker.pack(workUnits, numOfMultiWorkunits);
       addTopicSpecificPropsToWorkUnits(workUnitList, topicSpecificStateMap);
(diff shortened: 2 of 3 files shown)


ZihanLi58:
@ibuenros @autumnust @htran1

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -617,6 +617,7 @@
   /** Specifies a static location in HDFS to upload jars to. Useful for sharing jars across different Gobblin runs.*/
   public static final String MR_JARS_DIR = "mr.jars.dir";
   public static final String MR_JOB_MAX_MAPPERS_KEY = "mr.job.max.mappers";
+  public static final String TARGET_MAPPER_SIZE = "target.mapper.size";
```
ibuenros: Can you add the `mr` prefix to the key similar to the other configuration keys? e.g. `mr.target.mapper.size`

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -258,9 +258,15 @@ public String apply(KafkaTopic topic) {
...
       createEmptyWorkUnitsForSkippedPartitions(workUnits, topicSpecificStateMap, state);
-
+      //determine the number of mappers
       int numOfMultiWorkunits =
           state.getPropAsInt(ConfigurationKeys.MR_JOB_MAX_MAPPERS_KEY, ConfigurationKeys.DEFAULT_MR_JOB_MAX_MAPPERS);
+      if(state.contains(ConfigurationKeys.TARGET_MAPPER_SIZE)) {
+        double totalEstDataSize = KafkaWorkUnitPacker.getInstance(this, state).getWorkUnitEstSizes(workUnits);
+        LOG.info(String.format("The total estimated data size is %.2f", totalEstDataSize));
+        double targetMapperSize = state.getPropAsDouble(ConfigurationKeys.TARGET_MAPPER_SIZE);
+        numOfMultiWorkunits = (int) (totalEstDataSize / targetMapperSize);
```
ibuenros: You probably need to add 1 to this. If `totalEstDataSize < targetMapperSize` you would end up with 0 mappers :)

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -258,9 +258,15 @@ public String apply(KafkaTopic topic) {
       // Create empty WorkUnits for skipped partitions (i.e., partitions that have previous offsets,
       // but aren't processed).
       createEmptyWorkUnitsForSkippedPartitions(workUnits, topicSpecificStateMap, state);
-
+      //determine the number of mappers
       int numOfMultiWorkunits =
           state.getPropAsInt(ConfigurationKeys.MR_JOB_MAX_MAPPERS_KEY, ConfigurationKeys.DEFAULT_MR_JOB_MAX_MAPPERS);
+      if(state.contains(ConfigurationKeys.TARGET_MAPPER_SIZE)) {
+        double totalEstDataSize = KafkaWorkUnitPacker.getInstance(this, state).getWorkUnitEstSizes(workUnits);
```
ibuenros: It be better to instantiate `KafkaWorkUnitPacker` once in line 264 then use the same object both here and in line 270.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/workunit/packer/KafkaWorkUnitPacker.java
```diff
@@ -362,6 +362,15 @@ public static KafkaWorkUnitPacker getInstance(PackerType packerType, AbstractSou
         throw new IllegalArgumentException("WorkUnit packer type " + packerType + " not found");
     }
   }
+  public double getWorkUnitEstSizes(Map<String, List<WorkUnit>> workUnitsByTopic){
```
ibuenros: Is there a reason we can't just use `setWorkUnitEstSizes`? the logic in here is almost identical.

ZihanLi58: The method setWorkUnitEsiSizes is a protected method. So I just create a public method, and this new method just calculating the size instead of setting the property of workUnits

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -258,10 +258,20 @@ public String apply(KafkaTopic topic) {
...
-      List<WorkUnit> workUnitList = KafkaWorkUnitPacker.getInstance(this, state).pack(workUnits, numOfMultiWorkunits);
+      KafkaWorkUnitPacker kafkaWorkUnitPacker = KafkaWorkUnitPacker.getInstance(this, state);
+      int numOfMultiWorkunits = maxMapperNum;
+      if(state.contains(ConfigurationKeys.MR_TARGET_MAPPER_SIZE)) {
+        double totalEstDataSize = kafkaWorkUnitPacker.getWorkUnitEstSizes(workUnits);
+        LOG.info(String.format("The total estimated data size is %.2f", totalEstDataSize));
+        double targetMapperSize = state.getPropAsDouble(ConfigurationKeys.MR_TARGET_MAPPER_SIZE);
+        numOfMultiWorkunits = (int) (totalEstDataSize / targetMapperSize);
+        numOfMultiWorkunits = numOfMultiWorkunits>maxMapperNum? maxMapperNum:numOfMultiWorkunits;
+        numOfMultiWorkunits = numOfMultiWorkunits==0? 1: numOfMultiWorkunits;
```
ibuenros: You actually need to add 1 in general. Since integer division truncates down, you always want to add 1. (e.g. total size is 1.99, but targetMapperSize is 1, the current algorithm would use 1 mapper).

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -258,10 +258,20 @@ public String apply(KafkaTopic topic) {
...
           state.getPropAsInt(ConfigurationKeys.MR_JOB_MAX_MAPPERS_KEY, ConfigurationKeys.DEFAULT_MR_JOB_MAX_MAPPERS);
-      List<WorkUnit> workUnitList = KafkaWorkUnitPacker.getInstance(this, state).pack(workUnits, numOfMultiWorkunits);
+      KafkaWorkUnitPacker kafkaWorkUnitPacker = KafkaWorkUnitPacker.getInstance(this, state);
+      int numOfMultiWorkunits = maxMapperNum;
+      if(state.contains(ConfigurationKeys.MR_TARGET_MAPPER_SIZE)) {
+        double totalEstDataSize = kafkaWorkUnitPacker.getWorkUnitEstSizes(workUnits);
+        LOG.info(String.format("The total estimated data size is %.2f", totalEstDataSize));
+        double targetMapperSize = state.getPropAsDouble(ConfigurationKeys.MR_TARGET_MAPPER_SIZE);
+        numOfMultiWorkunits = (int) (totalEstDataSize / targetMapperSize);
+        numOfMultiWorkunits = numOfMultiWorkunits>maxMapperNum? maxMapperNum:numOfMultiWorkunits;
```
ibuenros: nit: You can use `Math.min(maxMapperNum, numOfMultiWorkunits)`

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/workunit/packer/KafkaWorkUnitPacker.java
ibuenros: You can make the other method public. In general, it is a bad idea to have the same logic duplicated in different locations.

Also, please add javadoc to any public methods.

PR #2589, opened 2019-04-04
[GOBBLIN-722] Unschedule gaas flow

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @sv2000  please review

### JIRA

### Description
it adds an option to let user unschedule a gaas flow.
it also made it possible to schedule a job even if it is already scheduled

### Tests
added a unit test in FlowConfigTest

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/scheduler/JobScheduler.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/scheduler/JobScheduler.java
@@ -351,8 +351,8 @@ public class JobScheduler extends AbstractIdleService {
     String jobName = jobProps.getProperty(ConfigurationKeys.JOB_NAME_KEY);
 
     if (this.scheduledJobs.containsKey(jobName)) {
-      LOG.warn("Job " + jobName + " has already been scheduled");
-      return;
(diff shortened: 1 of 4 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/scheduler/JobScheduler.java
```diff
@@ -351,8 +351,8 @@ public void scheduleJob(Properties jobProps, JobListener jobListener, Map<String
     String jobName = jobProps.getProperty(ConfigurationKeys.JOB_NAME_KEY);
 
     if (this.scheduledJobs.containsKey(jobName)) {
-      LOG.warn("Job " + jobName + " has already been scheduled");
-      return;
+      LOG.info("Job " + jobName + " was already scheduled, un-scheduling it now.");
+      unscheduleJob(jobName);
```
sv2000: Why are we calling uscheduleJob here?

arjun4084346: because cron-scheduler APIs do not support rescheduling, we have to call cron-scheduler.unschedule first

PR #2593, opened 2019-04-08
enable schema check for ticket ETL-8753, GOBBLIN-726

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-726

### Description
During the process of derived data deployment, we need to verify that data schema matches deployed schema. Each distcp mapper will first read the schema of the file to be copied, and abort if the file schema does not match the expected schema. 

### Tests
My PR adds the checkSchemaTest in EmbeddedGobblinDistcpTest to test the function when schema is right or wrong,.

### Commits


ZihanLi58:
@ibuenros @autumnust can you take a look at this?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,79 @@
...
+      DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
+      DataFileReader<GenericRecord> dataFileReader = new DataFileReader(new FsInput(this.file.getFileStatus().getPath(),fsFromFile), datumReader);
+      Schema schema = dataFileReader.getSchema();
+      if(!schema.toString().equals(this.state.getProp(EXPECTED_SCHEMA)))
+      {
+        try {
+          throw new DataRecordException("Schema does match the expected schema");
+        }
+        finally{
+          return null;
```
autumnust: Why this returned value is necessary?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public FileAwareInputStream readRecord(@Deprecated FileAwareInputStream reuse)
+      throws DataRecordException, IOException {
+    if (!this.recordRead) {
+      Configuration conf =
+          this.state == null ? HadoopUtils.newConfiguration() : HadoopUtils.getConfFromState(this.state);
+      FileSystem fsFromFile = this.file.getOrigin().getPath().getFileSystem(conf);
+      DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
```
autumnust: This schema validation will only work for Avro, which is good for now. For extensibility, we could encapsulate it into a method like `schemaChecking` or a better name you could think of, so that in the future we can just extend this class and override this method for different format.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,79 @@
...
+  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file)
+  {
+    this(fs, file, null);
+  }
+
+  @Override
+  public FileAwareInputStream readRecord(@Deprecated FileAwareInputStream reuse)
+      throws DataRecordException, IOException {
+    if (!this.recordRead) {
+      Configuration conf =
```
autumnust: Is that better to make this `Configuration` object reusable instead of creating one each time ? `HadoopUtils.getConfFromState(this.state)` is an expensive operation. I might overlook something here since base class did the same.

ZihanLi58: From my opinion, I think it's because the configuration only be used in this method. And temporary, this extractor will only handle one record so this readRecord method will be called only once. If we enable this extractor to handle multi records, then each time the state will change and the configuration need to be recomputed.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/SchemaCheckedCopySource.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.IOException;
+
+import org.apache.hadoop.fs.FileSystem;
+
+import org.apache.gobblin.configuration.WorkUnitState;
+import org.apache.gobblin.data.management.copy.extractor.FileAwareInputStreamExtractorWithCheckSchema;
+import org.apache.gobblin.source.extractor.Extractor;
+
+
+public class SchemaCheckedCopySource extends CopySource {
```
ibuenros: Please add javadoc for all classes and public methods.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.apache.gobblin.configuration.WorkUnitState;
+import org.apache.gobblin.data.management.copy.CopyableFile;
+import org.apache.gobblin.data.management.copy.FileAwareInputStream;
+import org.apache.gobblin.source.extractor.DataRecordException;
+import org.apache.gobblin.util.HadoopUtils;
+
+
+
+public class FileAwareInputStreamExtractorWithCheckSchema extends FileAwareInputStreamExtractor{
```
ibuenros: Please add javadoc.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,74 @@
...
+  public final static String EXPECTED_SCHEMA = "expected.schema";
```
ibuenros: It is good practice to namespace the configuration keys. You could do `gobblin.copy.expectedSchema`

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public FileAwareInputStream readRecord(@Deprecated FileAwareInputStream reuse)
+      throws DataRecordException, IOException {
+    if (!this.recordRead) {
+      Configuration conf =
+          this.state == null ? HadoopUtils.newConfiguration() : HadoopUtils.getConfFromState(this.state);
+      FileSystem fsFromFile = this.file.getOrigin().getPath().getFileSystem(conf);
+      if(!schemaChecking(fsFromFile))
+      {
+        throw new DataRecordException("Schema does match the expected schema");
```
ibuenros: I'm guessing your missing a *not* in the message?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,74 @@
...
+  {
+    super(fs, file, state);
+  }
+  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file)
+  {
+    this(fs, file, null);
+  }
+
+  @Override
+  public FileAwareInputStream readRecord(@Deprecated FileAwareInputStream reuse)
```
ibuenros: A lot of the logic in here is the same as `FileAwareInputStreamExtractor`. A different way of doing things with less repetition is to override `buildStream()`:

```java
@Override
public void FileAwareInputStream buildStream(FileSystem fs) {
  if (!schemaChecking(fs)) {
     throw new Exception(...);
  }
  return super.buildStream(fs);
}
```

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/SchemaCheckedCopySource.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.hadoop.fs.FileSystem;
+
+import org.apache.gobblin.configuration.WorkUnitState;
+import org.apache.gobblin.data.management.copy.extractor.FileAwareInputStreamExtractorWithCheckSchema;
+import org.apache.gobblin.source.extractor.Extractor;
+
+
+/**
+ * Used instead of {@link CopySource} for {@link FileSystem}s that need to check the schema
+ * during teh process of data deployment.
```
ibuenros: nit: `teh` -> `the`

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return super.buildStream(fsFromFile);
+  }
+
+  protected boolean schemaChecking(FileSystem fsFromFile)
+      throws IOException {
+    DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
+    DataFileReader<GenericRecord> dataFileReader = new DataFileReader(new FsInput(this.file.getFileStatus().getPath(),fsFromFile), datumReader);
+    Schema schema = dataFileReader.getSchema();
+    return schema.toString().equals(this.state.getProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA));
```
ibuenros: There are a lot of attributes of the schema that we may not want to compare (for example, if the doc for a field has changed, should we abort the copy?). It's fine to leave it as-is for now, but the next step will probably be to implement different schema comparison strategies.

gobblin-data-management/src/test/java/org/apache/gobblin/runtime/embedded/EmbeddedGobblinDistcpTest.java
```diff
@@ -71,6 +82,56 @@ public void test() throws Exception {
...
+    Assert.assertTrue(new File(tmpSource, fileName).exists());
+    Assert.assertFalse(new File(tmpTarget, fileName).exists());
+
+    EmbeddedGobblinDistcp embedded = new EmbeddedGobblinDistcp(new Path(tmpSource.getAbsolutePath()),
+        new Path(tmpTarget.getAbsolutePath()));
+    embedded.setLaunchTimeout(30, TimeUnit.SECONDS);
+    embedded.setConfiguration(ConfigurationKeys.SOURCE_CLASS_KEY, SchemaCheckedCopySource.class.getName());
+    //test when schema is not the expected one, the job will be aborted.
+    embedded.setConfiguration(ConfigurationKeys.COPY_EXPECTED_SCHEMA, "{\"type\":\"record\",\"name\":\"baseRecord\",\"fields\":[{\"name\":\"foo1\",\"type\":[\"null\",\"int\"],\"default\":null}]}");
+    embedded.run();
```
ibuenros: Can you also check that the run is marked as failed?

PR #2594, opened 2019-04-08
GOBBLIN-727: Skip commit in CloseOnFlushWriterWrapper if a commit has…

sv2000:
… already been invoked on the underlying writer.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-727

### Description
We skip commit() on the underlying writer if a commit has been previously invoked. Currently, duplicate commits on the underlying writer can result in Exceptions due to non-existing data in task staging location (as in the case of an FsDataWriter). 

### Tests
Tested with a Gobblin job running locally.

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/CloseOnFlushWriterWrapper.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/CloseOnFlushWriterWrapper.java
@@ -57,6 +57,8 @@ public class CloseOnFlushWriterWrapper<D> extends WriterWrapper<D> implements De
   private DataWriter<D> writer;
   private final Supplier<DataWriter<D>> writerSupplier;
   private boolean closed;
+  private boolean committed;
+
   // is the close functionality enabled?
   private final boolean closeOnFlush;
   private final ControlMessageHandler controlMessageHandler;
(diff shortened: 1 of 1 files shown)


sv2000:
@htran1 please review.

gobblin-core/src/main/java/org/apache/gobblin/writer/CloseOnFlushWriterWrapper.java
```diff
@@ -104,13 +107,15 @@ public void close() throws IOException {
 
   @Override
   public void commit() throws IOException {
-    writer.commit();
+    if (!this.committed) {
```
shirshanka: Is it expected behavior or unexpected behavior for commit() to be called multiple times? 
If unexpected, you should at least log a message (warn?) that you're skipping the commit.

sv2000: It can happen - if we end up with back-to-back control messages which is expected. (e.g. MetaDataUpdateControlMessage followed by FlushControlMessage). Currently, each control message will cause the writer to invoke commit(). But regardless, adding a log message is a good idea.

htran1:
+1

PR #2609, opened 2019-04-20
GOBBLIN-744: Support cancellation of a Helix workflow via a DELETE Spec.

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-744

### Description
This task supports the ability to interrupt and cancel a running job on a Gobblin Helix cluster via a DELETE Spec submitted to the JobConfigurationManager. The DELETE Spec should have "gobblin.cluster.shouldCancelRunningJobOnDelete" set to true for cancelling a running job. The default behavior is to simply delete the corresponding JobSpec from the JobCatalog. 

### Tests
IntegrationJobCancelViaSpecSuite and ClusterIntegrationTest.

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
@@ -57,10 +57,15 @@ public class FsScheduledJobConfigurationManager extends ScheduledJobConfiguratio
     for (Pair<SpecExecutor.Verb, JobSpec> entry : jobSpecs) {
       JobSpec jobSpec = entry.getValue();
       SpecExecutor.Verb verb = entry.getKey();
-      if (verb.equals(SpecExecutor.Verb.ADD) || verb.equals(SpecExecutor.Verb.UPDATE)) {
+      if (verb.equals(SpecExecutor.Verb.ADD)) {
         // Handle addition
         this._jobCatalog.put(jobSpec);
         postNewJobConfigArrival(jobSpec.getUri().toString(), jobSpec.getConfigAsProperties());
+      } else if (verb.equals(SpecExecutor.Verb.UPDATE)) {
+        //Handle update.
+        //Overwrite the jobSpec in the jobCatalog and post an UpdateJobConfigArrivalEvent.
+        this._jobCatalog.put(jobSpec);
+        postUpdateJobConfigArrival(jobSpec.getUri().toString(), jobSpec.getConfigAsProperties());
       } else if (verb.equals(SpecExecutor.Verb.DELETE)) {
         // Handle delete
         this._jobCatalog.remove(jobSpec.getUri());
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
@@ -18,7 +18,6 @@ package org.apache.gobblin.cluster;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 
 import org.apache.commons.lang3.tuple.Pair;
@@ -31,7 +30,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.gobblin.runtime.api.FsSpecConsumer;
 import org.apache.gobblin.runtime.api.JobSpec;
 import org.apache.gobblin.runtime.api.MutableJobCatalog;
-import org.apache.gobblin.runtime.api.Spec;
 import org.apache.gobblin.runtime.api.SpecExecutor;
 
 
@@ -53,27 +51,26 @@ public class FsScheduledJobConfigurationManager extends ScheduledJobConfiguratio
 
   @Override
   protected void fetchJobSpecs() throws ExecutionException, InterruptedException {
-    List<Pair<SpecExecutor.Verb, Spec>> jobSpecs =
-        (List<Pair<SpecExecutor.Verb, Spec>>) this._specConsumer.changedSpecs().get();
+    List<Pair<SpecExecutor.Verb, JobSpec>> jobSpecs =
+        (List<Pair<SpecExecutor.Verb, JobSpec>>) this._specConsumer.changedSpecs().get();
 
-    for (Pair<SpecExecutor.Verb, Spec> entry : jobSpecs) {
-      Spec spec = entry.getValue();
+    for (Pair<SpecExecutor.Verb, JobSpec> entry : jobSpecs) {
+      JobSpec jobSpec = entry.getValue();
       SpecExecutor.Verb verb = entry.getKey();
       if (verb.equals(SpecExecutor.Verb.ADD) || verb.equals(SpecExecutor.Verb.UPDATE)) {
         // Handle addition
-        JobSpec jobSpec = (JobSpec) spec;
         this._jobCatalog.put(jobSpec);
         postNewJobConfigArrival(jobSpec.getUri().toString(), jobSpec.getConfigAsProperties());
       } else if (verb.equals(SpecExecutor.Verb.DELETE)) {
         // Handle delete
-        this._jobCatalog.remove(spec.getUri());
-        postDeleteJobConfigArrival(spec.getUri().toString(), new Properties());
+        this._jobCatalog.remove(jobSpec.getUri());
+        postDeleteJobConfigArrival(jobSpec.getUri().toString(), jobSpec.getConfigAsProperties());
       }
 
       try {
         //Acknowledge the successful consumption of the JobSpec back to the SpecConsumer, so that the
         //SpecConsumer can delete the JobSpec.
-        this._specConsumer.commit(spec);
+        this._specConsumer.commit(jobSpec);
       } catch (IOException e) {
         log.error("Error when committing to FsSpecConsumer: ", e);
       }
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
@@ -159,4 +159,7 @@ public class GobblinClusterConfigurationKeys {
   public static final String KILL_DUPLICATE_PLANNING_JOB = GOBBLIN_CLUSTER_PREFIX + "kill.duplicate.planningJob";
   public static final boolean DEFAULT_KILL_DUPLICATE_PLANNING_JOB = true;
 
+  public static final String SHOULD_CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "shouldCancelRunningJobOnDelete";
+
+
 }
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
@@ -159,7 +159,6 @@ public class GobblinClusterConfigurationKeys {
   public static final String KILL_DUPLICATE_PLANNING_JOB = GOBBLIN_CLUSTER_PREFIX + "kill.duplicate.planningJob";
   public static final boolean DEFAULT_KILL_DUPLICATE_PLANNING_JOB = true;
 
-  public static final String SHOULD_CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "shouldCancelRunningJobOnDelete";
-
-
+  public static final String CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "job.cancelRunningJobOnDelete";
+  public static final String DEFAULT_CANCEL_RUNNING_JOB_ON_DELETE = "false";
 }
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
@@ -39,6 +39,7 @@ import com.google.common.collect.Maps;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.google.common.util.concurrent.Striped;
+import com.typesafe.config.Config;
 
 import org.apache.gobblin.annotation.Alpha;
 import org.apache.gobblin.cluster.event.DeleteJobConfigArrivalEvent;
@@ -118,7 +119,8 @@ public class GobblinHelixJobScheduler extends JobScheduler implements StandardMe
     this.jobCatalog = jobCatalog;
     this.metricContext = Instrumented.getMetricContext(new org.apache.gobblin.configuration.State(properties), this.getClass());
 
-    int metricsWindowSizeInMin = ConfigUtils.getInt(ConfigUtils.propertiesToConfig(this.properties),
+    Config jobConfig = ConfigUtils.propertiesToConfig(this.properties);
+    int metricsWindowSizeInMin = ConfigUtils.getInt(jobConfig,
                                                     ConfigurationKeys.METRIC_TIMER_WINDOW_SIZE_IN_MINUTES,
                                                     ConfigurationKeys.DEFAULT_METRIC_TIMER_WINDOW_SIZE_IN_MINUTES);
 
@@ -144,7 +146,7 @@ public class GobblinHelixJobScheduler extends JobScheduler implements StandardMe
 
     this.startServicesCompleted = false;
 
-    this.helixJobStopTimeoutSeconds = ConfigUtils.getLong(ConfigUtils.propertiesToConfig(properties),
+    this.helixJobStopTimeoutSeconds = ConfigUtils.getLong(jobConfig,
         GobblinClusterConfigurationKeys.HELIX_JOB_STOP_TIMEOUT_SECONDS,
         GobblinClusterConfigurationKeys.DEFAULT_HELIX_JOB_STOP_TIMEOUT_SECONDS);
   }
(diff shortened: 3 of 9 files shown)


sv2000:
@htran1 @yukuai518 @arjun4084346 @shirshanka Please review.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -82,6 +97,72 @@ public void testJobShouldComplete()
...
+
+    helixManager.connect();
+
+    while (TaskDriver.getWorkflowContext(helixManager, IntegrationJobCancelViaSpecSuite.JOB_ID) == null) {
+      log.warn("Waiting for the job to start...");
+      Thread.sleep(1000L);
+    }
+
+    // Give the job some time to reach writer, where it sleeps
+    Thread.sleep(2000L);
```
shirshanka: any way to avoid sleeping, but instead use something else to detect if the job has reached the writer?

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -82,6 +97,72 @@ public void testJobShouldComplete()
...
+    ZNRecord record = zkClient.readData(zNodePath);
+    String targetState = record.getSimpleField("TargetState");
+    Assert.assertEquals(targetState, TargetState.START.name());
+
+    //Add a JobSpec with DELETE verb signalling the Helix cluster to cancel the workflow
+    cancelViaSpecSuite.addJobSpec(IntegrationJobCancelViaSpecSuite.JOB_ID, SpecExecutor.Verb.DELETE.name());
+
+    //Give some time for the FsScheduledJobConfigurationManager to pick up the DELETE spec and send
+    // DeleteJobConfigArrivalEvent.
+    Thread.sleep(3000L);
```
shirshanka: Any way to avoid sleeping but waiting on some other observable event to figure out if delete has been processed?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
```diff
@@ -159,4 +159,7 @@
   public static final String KILL_DUPLICATE_PLANNING_JOB = GOBBLIN_CLUSTER_PREFIX + "kill.duplicate.planningJob";
   public static final boolean DEFAULT_KILL_DUPLICATE_PLANNING_JOB = true;
 
+  public static final String SHOULD_CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "shouldCancelRunningJobOnDelete";
```
shirshanka: suggestion: drop the prefix "should" from this config. So config could be: cancelRunningJobOnDelete

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
```diff
@@ -159,4 +159,7 @@
...
+
```
shirshanka: Add the default value for this config here (following the standard gobblin pattern)

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -324,10 +330,17 @@ public void handleUpdateJobConfigArrival(UpdateJobConfigArrivalEvent updateJobAr
   }
 
   @Subscribe
-  public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobArrival) {
+  public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobArrival) throws InterruptedException {
     LOGGER.info("Received delete for job configuration of job " + deleteJobArrival.getJobName());
     try {
       unscheduleJob(deleteJobArrival.getJobName());
+      Properties jobConfig = deleteJobArrival.getJobConfig();
+      if (PropertiesUtils.getPropAsBoolean(jobConfig, GobblinClusterConfigurationKeys.SHOULD_CANCEL_RUNNING_JOB_ON_DELETE, "false")) {
```
shirshanka: Should the default be false or true? 
It could be argued that the expected behavior on a job spec deletion is that running jobs should be cancelled? 

Are you just trying to maintain backward compatibility of this with current jobs?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
jhsenjaliya: +1, same with variable name :) 
btw, should this be `job.cancelR unningOnDelete`, since it's a job property just like `job.schedule` ?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -141,6 +143,10 @@ public GobblinHelixJobScheduler(Properties properties,
                                                   metricsWindowSizeInMin);
 
     this.startServicesCompleted = false;
+
+    this.helixJobStopTimeoutSeconds = ConfigUtils.getLong(ConfigUtils.propertiesToConfig(properties),
```
jhsenjaliya: probably not required to change at this point but i see this usage of ConfigUtils.propertiesToConfig over an over as if its O(1). may be it needs separate PR for reusing the created config.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
jhsenjaliya: I think it has to be `False` since to update the job config `handleUpdateJobConfigArrival` method deletes and insert as new job config. Or if we want it to be `True` then the method `handleUpdateJobConfigArrival` has to override this config while updating the jobConfig. #my2Cents.

sv2000: Yes, it is to maintain backward compatibility.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
sv2000: Its an easy change.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
sv2000: @jhsenjaliya +1 for noticing that. If the default were true, it will impact the job config updates as well.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
sv2000: Since this config is specific to cluster jobs, I added the GOBBLIN_CLUSTER_PREFIX to the config. Maybe, gobblin.cluster.job.cancelRunningOnDelete?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
sv2000: Ack.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
sv2000: Modified SleepingTask to write an empty file declaring that it is RUNNING. However, the test case will wait only a bounded amount of time for the existence of this file, to avoid the pathological condition where the file cannot be created and the test case keeps polling for the existence of this file forever.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
sv2000: Similar to the comment above. Keep polling the znode for a bounded amount of time before giving up and declaring failure.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
```diff
@@ -53,27 +51,26 @@ public FsScheduledJobConfigurationManager(EventBus eventBus, Config config, Muta
...
-        (List<Pair<SpecExecutor.Verb, Spec>>) this._specConsumer.changedSpecs().get();
+    List<Pair<SpecExecutor.Verb, JobSpec>> jobSpecs =
+        (List<Pair<SpecExecutor.Verb, JobSpec>>) this._specConsumer.changedSpecs().get();
 
-    for (Pair<SpecExecutor.Verb, Spec> entry : jobSpecs) {
-      Spec spec = entry.getValue();
+    for (Pair<SpecExecutor.Verb, JobSpec> entry : jobSpecs) {
+      JobSpec jobSpec = entry.getValue();
       SpecExecutor.Verb verb = entry.getKey();
       if (verb.equals(SpecExecutor.Verb.ADD) || verb.equals(SpecExecutor.Verb.UPDATE)) {
```
htran1: Should call `postUpdateJobConfigArrival` for the update case.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
```diff
@@ -53,27 +51,26 @@ public FsScheduledJobConfigurationManager(EventBus eventBus, Config config, Muta
 
   @Override
   protected void fetchJobSpecs() throws ExecutionException, InterruptedException {
-    List<Pair<SpecExecutor.Verb, Spec>> jobSpecs =
-        (List<Pair<SpecExecutor.Verb, Spec>>) this._specConsumer.changedSpecs().get();
+    List<Pair<SpecExecutor.Verb, JobSpec>> jobSpecs =
```
htran1: Should make sure that the job configuration manager receives the specs in modification time order.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -324,10 +332,18 @@ public void handleUpdateJobConfigArrival(UpdateJobConfigArrivalEvent updateJobAr
   }
 
   @Subscribe
-  public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobArrival) {
+  public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobArrival) throws InterruptedException {
     LOGGER.info("Received delete for job configuration of job " + deleteJobArrival.getJobName());
     try {
       unscheduleJob(deleteJobArrival.getJobName());
+      Properties jobConfig = deleteJobArrival.getJobConfig();
```
arjun4084346: This should go to a private method cancelJobIfRequired() coupled with a public util method cancelHelixJob

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -116,7 +119,8 @@ public GobblinHelixJobScheduler(Properties properties,
     this.jobCatalog = jobCatalog;
     this.metricContext = Instrumented.getMetricContext(new org.apache.gobblin.configuration.State(properties), this.getClass());
 
-    int metricsWindowSizeInMin = ConfigUtils.getInt(ConfigUtils.propertiesToConfig(this.properties),
+    Config jobConfig = ConfigUtils.propertiesToConfig(this.properties);
```
arjun4084346: Conversion of all the properties to Config looks unnecessary to me.

shirshanka: This should be the recommended pattern, so we can use ConfigUtils for the rest of the config gets, without incurring repeated txformations between properties and Config.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
```diff
@@ -17,24 +17,51 @@
...
 import org.apache.gobblin.runtime.TaskContext;
+import org.apache.gobblin.runtime.TaskState;
 import org.apache.gobblin.runtime.task.BaseAbstractTask;
 
 @Slf4j
 public class SleepingTask extends BaseAbstractTask {
+  public static final String TASK_STATE_FILE_KEY = "task.state.file.path";
+
   private final long sleepTime;
+  private FileSystem fs;
```
shirshanka: Do you need Hadoop FileSystem here? If you use java filesystem, you can set it to delete on process shutdown to ensure cleanup happens.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -82,6 +99,95 @@ public void testJobShouldComplete()
...
+
+    //Add a new JobSpec to the path monitored by the SpecConsumer
+    cancelViaSpecSuite.addJobSpec(IntegrationJobCancelViaSpecSuite.JOB_ID, SpecExecutor.Verb.ADD.name());
+
+    //Start the cluster
+    cancelViaSpecSuite.startCluster();
+
+    helixManager.connect();
+
+    while (TaskDriver.getWorkflowContext(helixManager, IntegrationJobCancelViaSpecSuite.JOB_ID) == null) {
```
shirshanka: maybe you can use AssertWithBackoff class from org.apache.gobblin.testing

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -82,6 +99,95 @@ public void testJobShouldComplete()
...
+    ZNRecord record = zkClient.readData(zNodePath);
+    String targetState = record.getSimpleField("TargetState");
+    Assert.assertEquals(targetState, TargetState.START.name());
+
+    //Add a JobSpec with DELETE verb signalling the Helix cluster to cancel the workflow
+    cancelViaSpecSuite.addJobSpec(IntegrationJobCancelViaSpecSuite.JOB_ID, SpecExecutor.Verb.DELETE.name());
+
+    int j = 0;
+    boolean successFlag = false;
+    while (true) {
```
shirshanka: AssertWithBackoff can be used here as well

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -26,19 +26,21 @@
...
+  public static final String TASK_STATE_FILE = "/tmp/IntegrationJobCancelSuite/taskState/_RUNNING";
 
   @Override
   protected Map<String, Config> overrideJobConfigs(Config rawJobConfig) {
     Config newConfig = ConfigFactory.parseMap(ImmutableMap.of(
         ConfigurationKeys.SOURCE_CLASS_KEY, "org.apache.gobblin.cluster.SleepingCustomTaskSource",
         ConfigurationKeys.JOB_ID_KEY, JOB_ID,
         GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_ENABLED_KEY, Boolean.TRUE,
-        GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS, 10L))
+        GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS, 10L, SleepingTask.TASK_STATE_FILE_KEY, TASK_STATE_FILE))
```
shirshanka: formatting seems off

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
sv2000: Added a method.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsScheduledJobConfigurationManager.java
sv2000: Done.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
sv2000: Used Java File API.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
sv2000: Done.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
sv2000: Done.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
```diff
@@ -17,24 +17,52 @@
 
 package org.apache.gobblin.cluster;
 
+import java.io.File;
+import java.io.IOException;
+
+import com.google.common.io.Files;
+
 import avro.shaded.com.google.common.base.Throwables;
```
shirshanka: spotted a faulty import from before

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -51,28 +68,30 @@ public void testJobShouldComplete()
...
     helixManager.connect();
 
     TaskDriver taskDriver = new TaskDriver(helixManager);
 
-    while (TaskDriver.getWorkflowContext(helixManager, IntegrationJobCancelSuite.JOB_ID) == null) {
-      log.warn("Waiting for the job to start...");
-      Thread.sleep(1000L);
-    }
+    AssertWithBackoff asserter1 = AssertWithBackoff.create().maxSleepMs(1000).backoffFactor(1);
+    asserter1.assertTrue(isTaskStarted(helixManager, IntegrationJobCancelSuite.JOB_ID),
```
shirshanka: you could chain the entire call without needing the local variable asserter1

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -82,6 +101,105 @@ public void testJobShouldComplete()
...
+  }
+
+  private Predicate<Void> isTaskStarted(HelixManager helixManager, String jobId) {
+    return input -> TaskDriver.getWorkflowContext(helixManager, jobId) != null;
+  }
+
+  private Predicate<Void> isTaskRunning(String taskStateFileName) {
+    return input -> {
+      try {
+        FileSystem fs = FileSystem.getLocal(new Configuration());
```
shirshanka: can avoid using hadoop fs here?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
```diff
@@ -74,6 +78,8 @@ public FsSpecConsumer(Config config) {
       return null;
     }
 
+    Arrays.sort(fileStatuses, Comparator.comparingLong(FileStatus::getModificationTime));
```
shirshanka: add a comment for why you're doing it and what you're expecting (ascending sort?)

htran1:
+1

PR #2620, opened 2019-04-24
[GOBBLIN-756] Add flow catalog that updates when filesystem is modified

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-756

### Description
Add `UpdatingFSFlowTemplateCatalog` that listens on the file system for changes and updates the catalog with changes. Also added `DynamicFlowEdge` which doesn't store `FlowTemplate` in memory so that changes to the flow catalog can be picked up.

### Tests
Added `UpdatingFSFlowTemplateCatalogTest`.

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
@@ -88,15 +88,16 @@ public class GitFlowGraphMonitor extends GitMonitoringService {
           .put(SHOULD_CHECKPOINT_HASHES, false)
           .build());
 
-  private Optional<FSFlowTemplateCatalog> flowCatalog;
+  private Optional<? extends FSFlowTemplateCatalog> flowTemplateCatalog;
   private FlowGraph flowGraph;
   private final Map<URI, TopologySpec> topologySpecMap;
   private final Config emptyConfig = ConfigFactory.empty();
   private final CountDownLatch initComplete;
 
-  public GitFlowGraphMonitor(Config config, Optional<FSFlowTemplateCatalog> flowCatalog, FlowGraph graph, Map<URI, TopologySpec> topologySpecMap, CountDownLatch initComplete) {
+  public GitFlowGraphMonitor(Config config, Optional<? extends FSFlowTemplateCatalog> flowTemplateCatalog,
+      FlowGraph graph, Map<URI, TopologySpec> topologySpecMap, CountDownLatch initComplete) {
     super(config.getConfig(GIT_FLOWGRAPH_MONITOR_PREFIX).withFallback(DEFAULT_FALLBACK));
-    this.flowCatalog = flowCatalog;
+    this.flowTemplateCatalog = flowTemplateCatalog;
     this.flowGraph = graph;
     this.topologySpecMap = topologySpecMap;
     this.initComplete = initComplete;
@@ -224,15 +225,15 @@ public class GitFlowGraphMonitor extends GitMonitoringService {
         Class flowEdgeFactoryClass = Class.forName(ConfigUtils.getString(edgeConfig, FlowGraphConfigurationKeys.FLOW_EDGE_FACTORY_CLASS,
             FlowGraphConfigurationKeys.DEFAULT_FLOW_EDGE_FACTORY_CLASS));
         FlowEdgeFactory flowEdgeFactory = (FlowEdgeFactory) GobblinConstructorUtils.invokeLongestConstructor(flowEdgeFactoryClass, edgeConfig);
-        if (flowCatalog.isPresent()) {
-          FlowEdge edge = flowEdgeFactory.createFlowEdge(edgeConfig, flowCatalog.get(), specExecutors);
+        if (flowTemplateCatalog.isPresent()) {
+          FlowEdge edge = flowEdgeFactory.createFlowEdge(edgeConfig, flowTemplateCatalog.get(), specExecutors);
           if (!this.flowGraph.addFlowEdge(edge)) {
             log.warn("Could not add edge {} to FlowGraph; skipping", edge.getId());
           } else {
             log.info("Added edge {} to FlowGraph", edge.getId());
           }
         } else {
-          log.warn("Could not add edge defined in {} to FlowGraph as FlowCatalog is absent", change.getNewPath());
+          log.warn("Could not add edge defined in {} to FlowGraph as FlowTemplateCatalog is absent", change.getNewPath());
         }
       } catch (Exception e) {
         log.warn("Could not add edge defined in {} due to exception {}", change.getNewPath(), e.getMessage());
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
@@ -75,7 +75,7 @@ public class MultiHopFlowCompiler extends BaseFlowToJobSpecCompiler {
 
   private GitFlowGraphMonitor gitFlowGraphMonitor;
 
-  private ReadWriteLock rwLock;
+  private ReadWriteLock rwLock = new ReentrantReadWriteLock(true);
 
   public MultiHopFlowCompiler(Config config) {
     this(config, true);
@@ -91,7 +91,6 @@ public class MultiHopFlowCompiler extends BaseFlowToJobSpecCompiler {
 
   public MultiHopFlowCompiler(Config config, Optional<Logger> log, boolean instrumentationEnabled) {
     super(config, log, instrumentationEnabled);
-    this.rwLock = new ReentrantReadWriteLock(true);
     this.flowGraph = new BaseFlowGraph();
     Optional<ObservingFSFlowEdgeTemplateCatalog> flowTemplateCatalog = Optional.absent();
     if (config.hasPath(ServiceConfigKeys.TEMPLATE_CATALOGS_FULLY_QUALIFIED_PATH_KEY)
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
@@ -75,6 +75,8 @@ public class MultiHopFlowCompiler extends BaseFlowToJobSpecCompiler {
 
   private GitFlowGraphMonitor gitFlowGraphMonitor;
 
+  private ReadWriteLock rwLock;
+
   public MultiHopFlowCompiler(Config config) {
     this(config, true);
   }
@@ -89,8 +91,8 @@ public class MultiHopFlowCompiler extends BaseFlowToJobSpecCompiler {
 
   public MultiHopFlowCompiler(Config config, Optional<Logger> log, boolean instrumentationEnabled) {
     super(config, log, instrumentationEnabled);
-    ReadWriteLock rwLock = new ReentrantReadWriteLock(true);
-    this.flowGraph = new BaseFlowGraph(rwLock);
+    this.rwLock = new ReentrantReadWriteLock(true);
+    this.flowGraph = new BaseFlowGraph();
     Optional<ObservingFSFlowEdgeTemplateCatalog> flowTemplateCatalog = Optional.absent();
     if (config.hasPath(ServiceConfigKeys.TEMPLATE_CATALOGS_FULLY_QUALIFIED_PATH_KEY)
         && StringUtils.isNotBlank(config.getString(ServiceConfigKeys.TEMPLATE_CATALOGS_FULLY_QUALIFIED_PATH_KEY))) {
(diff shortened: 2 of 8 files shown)


jack-moseley:
@sv2000 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -88,13 +88,13 @@
           .put(SHOULD_CHECKPOINT_HASHES, false)
           .build());
 
-  private Optional<FSFlowTemplateCatalog> flowCatalog;
+  private Optional<? extends FSFlowTemplateCatalog> flowCatalog;
```
autumnust: Shall we rename the varaible to `flowTemplateCatalog` as well?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.gobblin.util.filesystem.PathAlterationListenerAdaptor;
+
+
+/**
+ * {@link FSFlowTemplateCatalog} that keeps a cache of flow and job templates. It has a
+ * {@link org.apache.gobblin.util.filesystem.PathAlterationListener} on the root path, and clears the cache when a change
+ * is detected so that the next call to {@link #getFlowTemplate(URI)} will use the updated files.
+ */
+@Slf4j
+public class UpdatingFSFlowTemplateCatalog extends FSFlowTemplateCatalog {
```
autumnust: How does `DynamicFSFlowTemplateCatalog` sound since you used "dynamic" in the edge ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
```diff
@@ -88,11 +88,11 @@ public MultiHopFlowCompiler(Config config, Optional<Logger> log) {
   public MultiHopFlowCompiler(Config config, Optional<Logger> log, boolean instrumentationEnabled) {
     super(config, log, instrumentationEnabled);
     this.flowGraph = new BaseFlowGraph();
-    Optional<FSFlowTemplateCatalog> flowCatalog = Optional.absent();
+    Optional<UpdatingFSFlowTemplateCatalog> flowCatalog = Optional.absent();
```
autumnust: This is `flowTemplateCatalog`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DynamicFlowEdge.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.apache.gobblin.runtime.api.JobTemplate;
+import org.apache.gobblin.runtime.api.SpecExecutor;
+import org.apache.gobblin.runtime.api.SpecNotFoundException;
+import org.apache.gobblin.service.modules.template.FlowTemplate;
+import org.apache.gobblin.service.modules.template_catalog.FSFlowTemplateCatalog;
+
+
+/**
+ * Implementation of {@link FlowEdge} that loads a {@link FlowTemplate} from the {@link org.apache.gobblin.runtime.spec_catalog.FlowCatalog}
```
autumnust: The documentation here is not accurate. `FlowCatalog` is not handling templates, I guess you are referring to the template catalog?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DynamicFlowEdge.java
```diff
@@ -0,0 +1,72 @@
...
+ * when {@link #getFlowTemplate()} is called rather than storing it in memory.
+ */
+@Slf4j
+public class DynamicFlowEdge extends BaseFlowEdge {
+  private final FSFlowTemplateCatalog flowCatalog;
```
autumnust: Template Catalog

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
```diff
@@ -0,0 +1,108 @@
...
+  private static Map<URI, FlowTemplate> flowTemplateMap = new ConcurrentHashMap<>();
```
autumnust: Given the size of flow templates is limited, loading them all in memory should be fine. Or we may need to change it to something like guava loading cache to cap the total number of templates could be held in memory. @sv2000 what was usual size of the number of flow templates ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * {@link org.apache.gobblin.util.filesystem.PathAlterationListener} that clears job template cache if a file is
+   * created or updated.
+   */
+  private class FlowCatalogPathAlterationListener extends PathAlterationListenerAdaptor {
+    @Override
+    public void onFileCreate(Path path) {
+      clearTemplates();
```
autumnust: Correct me if I am wrong: The folder listener is listening to any file changes and template catalog needs to return them dynamically in runtime. 
Based on that, why do we need to clear templates when new template file is added ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public void onFileChange(Path path) {
+      clearTemplates();
```
autumnust: Should it be a invalidation of an entry in cache?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
```diff
@@ -0,0 +1,108 @@
...
+public class UpdatingFSFlowTemplateCatalog extends FSFlowTemplateCatalog {
+  private static Map<URI, FlowTemplate> flowTemplateMap = new ConcurrentHashMap<>();
+  private static Map<URI, List<JobTemplate>> jobTemplateMap = new ConcurrentHashMap<>();
+
+  public UpdatingFSFlowTemplateCatalog(Config sysConfig) throws IOException {
+    super(sysConfig);
+  }
+
+  @Override
+  public FlowTemplate getFlowTemplate(URI flowTemplateDirURI)
```
autumnust: This method is potentially having race condition with file listener. Should we have sort of lock to make sure `getFlowTemplate` won't return while changes on cache triggered by file listener is in progress?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalogTest.java
```diff
@@ -0,0 +1,83 @@
...
+    Path flowConfPath = new File(new File(templateDir, FSFlowTemplateCatalogTest.TEST_TEMPLATE_NAME), "flow.conf").toPath();
+    List<String> lines = java.nio.file.Files.readAllLines(flowConfPath);
+    for (int i = 0; i < lines.size(); i++) {
+      if (lines.get(i).equals("gobblin.flow.edge.input.dataset.descriptor.0.format=avro")) {
+        lines.set(i, "gobblin.flow.edge.input.dataset.descriptor.0.format=any");
+        break;
+      }
+    }
+    java.nio.file.Files.write(flowConfPath, lines);
+    Thread.sleep(10000);
```
autumnust: Usually it is not a good practice to sleep in unit tests. @sv2000 pointed a useful tool called `AssertWithBackoff` which would be useful in this situation

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DynamicFlowEdge.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.gobblin.service.modules.template.FlowTemplate;
+import org.apache.gobblin.service.modules.template_catalog.FSFlowTemplateCatalog;
+
+
+/**
+ * Implementation of {@link FlowEdge} that loads a {@link FlowTemplate} from the {@link org.apache.gobblin.runtime.spec_catalog.FlowCatalog}
+ * when {@link #getFlowTemplate()} is called rather than storing it in memory.
+ */
+@Slf4j
+public class DynamicFlowEdge extends BaseFlowEdge {
```
sv2000: We should move the logic of DynamicFlowEdge to BaseFlowEdge i.e. I don't think we need to extend BaseFlowEdge.

sv2000:
@jack-moseley Please check why the travis build is failing.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
sv2000: @autumnust yeah it should be fine to load the templates into memory. Each template is a few KBs in size.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
sv2000: Can we call this class ObservingFSFlowEdgeTemplateCatalog?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
jack-moseley: There are cases where, for example, a job template could get updated and multiple flow templates are pointing to that job template, so we would have to figure out which flows used that job template and invalidate those as well. Due to this I think it's easier to just clear entire cache.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
jack-moseley: In most cases we wouldn't need to clear when file is added, but I think there could be edge cases where it would be problem. Let's say we have a flow template which points to a job template that does not exist. Now if we create that job template (without modifying anything else, we need to reload the flow template. Another case would be deleting and recreating a changed template file instead of modifying it.

Pretty small edge cases but it is not very expensive to reload all templates so I think it's better to leave it like this.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/UpdatingFSFlowTemplateCatalog.java
jack-moseley: Updated to use a thread-safe way of getting flow template from map.

jack-moseley:
@sv2000 I fixed the possible concurrency issue by just sharing the read-write lock with `BaseFlowGraph`, since that lock is already acquired when finding a path.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
```diff
@@ -87,12 +89,13 @@ public MultiHopFlowCompiler(Config config, Optional<Logger> log) {
 
   public MultiHopFlowCompiler(Config config, Optional<Logger> log, boolean instrumentationEnabled) {
     super(config, log, instrumentationEnabled);
-    this.flowGraph = new BaseFlowGraph();
-    Optional<FSFlowTemplateCatalog> flowCatalog = Optional.absent();
+    ReadWriteLock rwLock = new ReentrantReadWriteLock(true);
+    this.flowGraph = new BaseFlowGraph(rwLock);
```
sv2000: Using the RW lock of the FlowGraph works, but may not be desirable. The semantics of the RW lock of the BaseFlowGraph is to control read-write access to the FlowGraph. It seems that it might be better to have a different RW lock for controlling access to the FlowEdgeTemplateCatalog. The compiler can acquire this lock at the beginning of compileFlow() and release it once the compilation is done.

sv2000:
@autumnust can you give a +1, if the changes look good to you?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
```diff
@@ -186,6 +192,9 @@ public void awaitHealthy() throws InterruptedException {
     }
     Instrumented.markMeter(flowCompilationSuccessFulMeter);
     Instrumented.updateTimer(flowCompilationTimer, System.nanoTime() - startTime, TimeUnit.NANOSECONDS);
+
+    this.rwLock.readLock().unlock();
```
autumnust: Shall we put it in `finally` block ?

sv2000: Good catch @autumnust !

PR #2626, opened 2019-05-02
[GOBBLIN-762] Add automatic scaling for Gobblin on YARN

htran1:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-762

### Description
Added `YarnAutoScalingManager` which can be started by the `GobblinApplicationMaster` by setting the `gobblin.yarn.app.master.serviceClasses` configuration. This class runs a scheduled task with a default interval of 60 seconds to detect the number of required partitions for the workflows submitted to Helix. It will request the `YarnService` to scale to a computed number of containers. If the requested number of containers is higher than the YarnService has previously requested then it will request more containers. If the requested count is less than the current number of allocated containers then it will free any unused containers.

### Tests
YarnAutoScalingManagerTest and YarnServiceTest

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinApplicationMaster.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinApplicationMaster.java
@@ -44,15 +44,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Optional;
+import com.google.common.util.concurrent.Service;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+
+import lombok.Getter;
 
 import org.apache.gobblin.annotation.Alpha;
 import org.apache.gobblin.cluster.GobblinClusterConfigurationKeys;
 import org.apache.gobblin.cluster.GobblinClusterManager;
 import org.apache.gobblin.cluster.GobblinClusterUtils;
+import org.apache.gobblin.util.ConfigUtils;
 import org.apache.gobblin.util.JvmUtils;
 import org.apache.gobblin.util.logs.Log4jConfigurationHelper;
+import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
 import org.apache.gobblin.yarn.event.DelegationTokenUpdatedEvent;
 
 
@@ -68,8 +73,10 @@ import org.apache.gobblin.yarn.event.DelegationTokenUpdatedEvent;
  */
 @Alpha
 public class GobblinApplicationMaster extends GobblinClusterManager {
-
   private static final Logger LOGGER = LoggerFactory.getLogger(GobblinApplicationMaster.class);
+
+  @Getter
+  private final YarnService yarnService;
 
   public GobblinApplicationMaster(String applicationName, ContainerId containerId, Config config,
       YarnConfiguration yarnConfiguration) throws Exception {
@@ -82,12 +89,21 @@ public class GobblinApplicationMaster extends GobblinClusterManager {
           .addService(gobblinYarnLogSource.buildLogCopier(this.config, containerId, this.fs, this.appWorkDir));
     }
 
-    this.applicationLauncher
-        .addService(buildYarnService(this.config, applicationName, this.applicationId, yarnConfiguration, this.fs));
+    this.yarnService = buildYarnService(this.config, applicationName, this.applicationId, yarnConfiguration, this.fs);
+    this.applicationLauncher.addService(this.yarnService);
 
     if (UserGroupInformation.isSecurityEnabled()) {
       LOGGER.info("Adding YarnContainerSecurityManager since security is enabled");
       this.applicationLauncher.addService(buildYarnContainerSecurityManager(this.config, this.fs));
+    }
+
+    // Add additional services
+    List<String> serviceClassNames = ConfigUtils.getStringList(this.config,
+        GobblinYarnConfigurationKeys.APP_MASTER_SERVICE_CLASSES);
+
+    for (String serviceClassName : serviceClassNames) {
+      Class<?> serviceClass = Class.forName(serviceClassName);
+      this.applicationLauncher.addService((Service) GobblinConstructorUtils.invokeLongestConstructor(serviceClass, this));
     }
   }
 
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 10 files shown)


gobblin-yarn/src/test/java/org/apache/gobblin/yarn/YarnAutoScalingManagerTest.java
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Unit tests for {@link YarnAutoScalingManager}
+ */
+@Test(groups = { "gobblin.yarn" })
+public class YarnAutoScalingManagerTest {
+  /**
+   * Test for one workflow with one job
+   */
+  @Test
+  public void testOneJob() throws IOException {
```
jhsenjaliya: test looks good.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -0,0 +1,158 @@
...
+     */
+    @Override
+    public void run() {
+      Set<String> inUseInstances = new HashSet<>();
+
+      int numPartitions = 0;
+      for (Map.Entry<String, WorkflowConfig> workFlowEntry : taskDriver.getWorkflows().entrySet()) {
+        WorkflowContext workflowContext = taskDriver.getWorkflowContext(workFlowEntry.getKey());
+
+        // Only allocate for active workflows
```
jhsenjaliya: Is the idea to allocate containers for active workflow asynchronously ? ( every 60 seconds )
shouldn't it be inline with WorkFlow execution, i.e. when workflow gets picked up for execution, it should figure out how many container it needs and allocate accordingly?
OR, is it that Gobblin will keep certain number of containers alive to reuse ?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -82,13 +89,22 @@ public GobblinApplicationMaster(String applicationName, ContainerId containerId,
...
       this.applicationLauncher.addService(buildYarnContainerSecurityManager(this.config, this.fs));
     }
+
+    // Add additional services
+    List<String> serviceClassNames = ConfigUtils.getStringList(this.config,
+        GobblinYarnConfigurationKeys.APP_MASTER_SERVICE_CLASSES);
+
+    for (String serviceClassName : serviceClassNames) {
+      Class<?> serviceClass = Class.forName(serviceClassName);
+      this.applicationLauncher.addService((Service) GobblinConstructorUtils.invokeLongestConstructor(serviceClass, this));
```
shirshanka: why longest constructor?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -0,0 +1,158 @@
...
+        Set<String> jobs = jobDag.getAllNodes();
+
+        // sum up the number of partitions
+        for (String jobName : jobs) {
+          JobContext jobContext = taskDriver.getJobContext(jobName);
+
+          if (jobContext != null) {
+            log.debug("JobContext {} num partitions {}", jobContext, jobContext.getPartitionSet().size());
+
+            inUseInstances.addAll(jobContext.getPartitionSet().stream().map(jobContext::getAssignedParticipant)
```
shirshanka: indent seems off on this, might want to multi-line this?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -0,0 +1,158 @@
...
+                .filter(e -> e != null).collect(Collectors.toSet()));
+
+            numPartitions += jobContext.getPartitionSet().size();
+          }
+        }
+      }
+
+      int numTargetContainers = (numPartitions + (this.partitionsPerContainer - 1)) / this.partitionsPerContainer;
```
shirshanka: why is (this.partitionsPerContainer - 1) within parenthesis.

shirshanka: Might want to document that you're basically doing ceiling(numPartitions/this.partitionsPerContainer) efficiently :)

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -229,6 +248,21 @@ public void handleContainerShutdownRequest(ContainerShutdownRequest containerShu
...
+   * Request the Resource Manager to release the container
+   * @param containerReleaseRequest containers to release
+   */
+  @Subscribe
+  public void handleContainerReleaseRequest(ContainerReleaseRequest containerReleaseRequest) {
+    for (Container container : containerReleaseRequest.getContainers()) {
+      LOGGER.info(String.format("Releasing container %s running on %s", container.getId(), container.getNodeId()));
+
+      // record that this container was explicitly released so that a new one is not spawned to replace it
+      this.releasedContainerSet.add(container.getId());
```
shirshanka: If the ordering of inserting into the "blacklist" before firing the "releaseAssignedContainer" call matters, please document why.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -159,6 +173,11 @@
 
   private volatile boolean shutdownInProgress = false;
 
+  // The number of containers requested
```
shirshanka: Need a bit more documentation on this. Is this the current outstanding number of containers requested based on the last call to requestContainers .. or the total lifetime number..

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -468,6 +562,13 @@ private void handleContainerCompletion(ContainerStatus containerStatus) {
           containerStatus.getContainerId(), containerStatus.getDiagnostics()));
     }
 
+    if (this.releasedContainerSet.contains(containerStatus.getContainerId())) {
+      LOGGER.info("Container release requested, so not spawning a replacement for containerId {}",
+          containerStatus.getContainerId());
+      this.releasedContainerSet.remove(containerStatus.getContainerId());
```
shirshanka: Why is this remove safe to perform. Are we assured that spawning a replacement for this containerId will not happen again?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
yukuai518: Are you always assume the handleContainerCompletion() logic would happen after releasedContainerSet.add() ? What if they happened in parallel? For example, a container encountered an issue so it triggers the handleContainerCompletion logic, which does nothing to releasedContainerSet, but suddenly the YarnAutoScaling service also thinks it needs to recycle the container, and it happened to choose the container that is in the middle of handleContainerCompletion processing? Then you will add the problematic container to the releasedContainerSet and it will never be cleaned up?

yukuai518: I think we may also need to consider to update containerMap in this case? Because the containerMap was used for your container target number shrinking logic. We don't want to have a race condition here.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
jhsenjaliya: i see, thanks for the explanation around this.

PR #2632, opened 2019-05-14
[GOBBLIN-769] Support string record timestamp in TimeBasedAvroWriterPartitioner

zxcware:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-769

### Description
- Currently, if a record timestamp is a string, `TimeBasedAvroWriterPartitioner` will not be able to recognize it and will use current time. This change will try to convert a string to Long before using current time

### Tests
- `TimeBasedAvroWriterPartitioner.testGetRecordTimestamp` covers various cases for reading string record timestamp

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
@@ -18,6 +18,7 @@
 package org.apache.gobblin.writer.partitioner;
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.avro.generic.GenericRecord;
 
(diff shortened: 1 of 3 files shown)


gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
```diff
@@ -73,9 +74,25 @@ public long getRecordTimestamp(GenericRecord record) {
   /**
    *  Check if the partition column value is present and is a Long object. Otherwise, use current system time.
    */
-  private static long getRecordTimestamp(Optional<Object> writerPartitionColumnValue) {
-    return writerPartitionColumnValue.orNull() instanceof Long ? (Long) writerPartitionColumnValue.get()
-        : System.currentTimeMillis();
+  private long getRecordTimestamp(Optional<Object> writerPartitionColumnValue) {
+
+    // Default to current time
+    Long recordTimestamp = timeUnit.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
```
htran1: Should move this conversion after the if block and only do it when no value has been set to avoid the cost of get time and conversion.

gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
```diff
@@ -73,9 +74,25 @@ public long getRecordTimestamp(GenericRecord record) {
...
+
+    if (writerPartitionColumnValue.isPresent()) {
+      Object val = writerPartitionColumnValue.get();
+      if (val instanceof Long) {
+        recordTimestamp = (Long) val;
+      } else {
+        try {
+          recordTimestamp = Long.parseLong(val.toString());
+        } catch (NumberFormatException e) {
+          // ignore
```
htran1: This may mask errors and the behavior of parseable values being used and other cases getting the latest time may not be desired.

I'm not sure why the old code allows an invalid partition column to be specified, but this new code will add conversion overhead and may change the behavior.

So I think a config should be added to maintain the old behavior by default. And maybe there should also be a config to raise an error immediately on parsing errors to help detect bad data.

gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
zxcware: +1

gobblin-core/src/main/java/org/apache/gobblin/writer/partitioner/TimeBasedAvroWriterPartitioner.java
zxcware: I'll add a config to enable parsing string as long

PR #2636, opened 2019-05-15
[GOBBLIN-766]  Emit Workunits Created event

krishraman:
Dear Gobblin maintainers,

### JIRA

### Description

### Tests
 Unit test cases added.
 tested new created event log messages in Yugioh cluster.

### Commits

### DIFF
--- a/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
+++ b/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
@@ -31,6 +31,7 @@ public class JobEvent {
   public static final String LOCK_IN_USE = "LockInUse";
   public static final String WORK_UNITS_MISSING = "WorkUnitsMissing";
   public static final String WORK_UNITS_EMPTY = "WorkUnitsEmpty";
+  public  static final String WORK_UNITS_CREATED = "WorkUnitsCreated";
   public static final String TASKS_SUBMITTED = "TasksSubmitted";
 
   public static final String METADATA_JOB_ID = "jobId";
--- a/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
+++ b/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
@@ -31,7 +31,7 @@ public class JobEvent {
   public static final String LOCK_IN_USE = "LockInUse";
   public static final String WORK_UNITS_MISSING = "WorkUnitsMissing";
   public static final String WORK_UNITS_EMPTY = "WorkUnitsEmpty";
-  public  static final String WORK_UNITS_CREATED = "WorkUnitsCreated";
+  public static final String WORK_UNITS_CREATED = "WorkUnitsCreated";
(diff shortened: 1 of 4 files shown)


krishh85:
@htran1 @yukuai518 Can you please review this PL. Thanks

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
```diff
@@ -31,6 +31,7 @@
   public static final String LOCK_IN_USE = "LockInUse";
   public static final String WORK_UNITS_MISSING = "WorkUnitsMissing";
   public static final String WORK_UNITS_EMPTY = "WorkUnitsEmpty";
+  public static final String WORK_UNITS_CREATED = "WorkUnitsCreated";
```
jhsenjaliya: seems like these event name constants are scattered between `JobEvents` class and `EventNames` enum.  should we combine them?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -377,6 +378,13 @@ public void apply(JobListener jobListener, JobContext jobContext)
           jobState.setState(JobState.RunningState.COMMITTED);
           isWorkUnitsEmpty = true;
           return;
+        } else {
```
jhsenjaliya: how about reporting `workUnitsCreated=0` when there is no workUnits instead of reporting `WorkUnitsEmpty` event?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
krishh85: EventName enum seems to be defined specifically for Job Timers. I am not sure if this is by design or it just happened to be that way..

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
krishh85: This depends on the consumer of the event. In our case we consume these events in time-series and the workUnitsCreated event is set to 0 every minute irrespective of this event emitted or otherwise and hence workUnitsEmpty helps to distinguish in this scenario.
We could potentially emit both by removing the "else" clause.

yukuai518:
+1 LGTM

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -377,6 +378,13 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
+          if (workUnitStream.isSafeToMaterialize()) {
+            this.eventSubmitter.submit(JobEvent.WORK_UNITS_CREATED, "workUnitsCreated",
+                Long.toString(workUnitStream.getMaterializedWorkUnitCollection().size()));
```
ibuenros: This call will force work units to be materialized and prevent all optimizations from streaming them (in terms of memory and speed). Ideally, the correct way of doing this is to count the work units as we work on them and emit the event at the end of the run.

PR #2637, opened 2019-05-16
[GOBBLIN-772]Implement Schema Comparison Strategy during Disctp

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-772

### Description
We need a schema comparison strategy to make sure the real schema and the expected schema have compatibility.
- Add an interface of schema check strategy.
- Add configuration using which user can specify the schema check strategy
- Add default schema check strategy. The default schema check strategy will check if the two schemas have matching field names and types.
- The file org.apache.gobblin.restli.throttling.permits.snapshot.json changed because the file gobblin-github/gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-api/src/main/pegasus/org/apache/gobblin/restli/throttling/PermitAllocation.pdsc changed in ticket[GOBBLIN-724]. And after a full build, this change will automatically shows up.

### Tests
- Unit test: Make the real schema and expected schema have different name or type and make sure the method return false. And make they have matching name and type but not the same doc and make sure the method return true.
- Integration test: Run end to end test on Hadoop to make sure the SchemaCheckedCopySource works well with the strategy.

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -936,4 +936,10 @@ public class ConfigurationKeys {
   public static final String GIT_MONITOR_SSH_KNOWN_HOSTS = "knownHosts";
   public static final String GIT_MONITOR_SSH_KNOWN_HOSTS_FILE = "knownHostsFile";
   public static final String GIT_MONITOR_JSCH_LOGGER_ENABLED = "isJschLoggerEnabled";
+
+  /**
+   * Configuration related to avro schema check strategy
+   */
+  public static final String AVRO_SCHEMA_CHECK_STRATEGY = "avro.schema.check.strategy";
+  public static final String AVRO_SCHEMA_CHECK_STRATEGY_DEFAULT = "org.apache.gobblin.util.schema_check.AvroSchemaCheckDefaultStrategy";
 }
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
@@ -26,6 +26,8 @@ import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.DatumReader;
 import org.apache.avro.mapred.FsInput;
+import org.apache.gobblin.util.schema_check.AvroSchemaCheckStrategy;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 
 import org.apache.gobblin.configuration.ConfigurationKeys;
@@ -39,32 +41,42 @@ import org.apache.gobblin.source.extractor.DataRecordException;
  * check if the schema matches the expected schema. If not it will abort the job.
  */
 
-public class FileAwareInputStreamExtractorWithCheckSchema extends FileAwareInputStreamExtractor{
+public class FileAwareInputStreamExtractorWithCheckSchema extends FileAwareInputStreamExtractor {
 
-  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file, WorkUnitState state)
-  {
+  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file, WorkUnitState state) {
     super(fs, file, state);
   }
-  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file)
-  {
+
+  public FileAwareInputStreamExtractorWithCheckSchema(FileSystem fs, CopyableFile file) {
     this(fs, file, null);
   }
 
   @Override
-  protected FileAwareInputStream buildStream(FileSystem fsFromFile)
-      throws DataRecordException, IOException{
-    if(!schemaChecking(fsFromFile))
-    {
+  protected FileAwareInputStream buildStream(FileSystem fsFromFile) throws DataRecordException, IOException {
+    if (!schemaChecking(fsFromFile)) {
       throw new DataRecordException("Schema does not match the expected schema");
     }
     return super.buildStream(fsFromFile);
   }
 
-  protected boolean schemaChecking(FileSystem fsFromFile)
-      throws IOException {
+  /**
+   * Use {@link AvroSchemaCheckStrategy} to make sure the real schema and the expected schema have matching field names and types
+   * @param fsFromFile
+   * @return
+   * @throws IOException
+   */
+  protected boolean schemaChecking(FileSystem fsFromFile) throws IOException {
     DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
-    DataFileReader<GenericRecord> dataFileReader = new DataFileReader(new FsInput(this.file.getFileStatus().getPath(),fsFromFile), datumReader);
+    DataFileReader<GenericRecord> dataFileReader =
+        new DataFileReader(new FsInput(this.file.getFileStatus().getPath(), new Configuration()), datumReader);
     Schema schema = dataFileReader.getSchema();
-    return schema.toString().equals(this.state.getProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA));
+    Schema expectedSchema = new Schema.Parser().parse(this.state.getProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA));
+    AvroSchemaCheckStrategy strategy = AvroSchemaCheckStrategy.AvroSchemaCheckStrategyFactory.create(this.state);
+    if(strategy == null)
+    {
+      throw new IOException("schema check strategy cannot be initialized");
+    }
+    return strategy.compare(expectedSchema,schema);
   }
+
 }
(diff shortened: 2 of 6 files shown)


ZihanLi58:
@ibuenros @autumnust Can you take a look at this code?

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-api/src/main/snapshot/org.apache.gobblin.restli.throttling.permits.snapshot.json
```diff
@@ -17,6 +17,18 @@
       "type" : "long",
       "doc" : "Client should not try to acquire permits before this delay has passed.",
       "optional" : true
+    }, {
```
autumnust: Is this changes from other PR? Can you rebase the newest changes so that it won't be included?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -39,32 +42,109 @@
...
+  protected FileAwareInputStream buildStream(FileSystem fsFromFile) throws DataRecordException, IOException {
+    if (!schemaChecking(fsFromFile)) {
       throw new DataRecordException("Schema does not match the expected schema");
     }
     return super.buildStream(fsFromFile);
   }
 
-  protected boolean schemaChecking(FileSystem fsFromFile)
-      throws IOException {
+  protected boolean schemaChecking(FileSystem fsFromFile) throws IOException {
```
autumnust: What is the scope of this schema checking ? Is this verifying all fields inside `toValidateSchema` is appearing in expected schema ? Please add javadoc on this.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -39,32 +42,109 @@
...
+              return false;
+            } else {
+              if (!compare(toValidateField.schema(), expectedFiled.schema())) {
+                return false;
+              }
+            }
+          }
+          return true;
+        }
+        case UNION: {
```
autumnust: Just want to confirm, do you care about the ordering of fields ?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -39,32 +42,109 @@
...
+    DataFileReader<GenericRecord> dataFileReader =
+        new DataFileReader(new FsInput(this.file.getFileStatus().getPath(), fsFromFile), datumReader);
     Schema schema = dataFileReader.getSchema();
-    return schema.toString().equals(this.state.getProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA));
+    Schema expectedSchema = new Schema.Parser().parse(this.state.getProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA));
+
+    return compare(schema, expectedSchema);
+  }
+
+  private boolean compare(Schema toValidate, Schema expected) {
```
autumnust: Depending on the scope of your schema checking, this method can be reused for other purposes. May be make it static method or put in some utilities classes?

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-api/src/main/snapshot/org.apache.gobblin.restli.throttling.permits.snapshot.json
ZihanLi58: Every time I build the project, this change will be automatically made.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
ZihanLi58: No, I just want to make sure they have matching field.

autumnust:
@ZihanLi58 Travis is failing

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
```diff
@@ -39,32 +43,38 @@
  * check if the schema matches the expected schema. If not it will abort the job.
  */
 
-public class FileAwareInputStreamExtractorWithCheckSchema extends FileAwareInputStreamExtractor{
+public class FileAwareInputStreamExtractorWithCheckSchema extends FileAwareInputStreamExtractor {
```
jhsenjaliya: just a comment on this class: i understand that this class already exists and its adding proper field level schema check, but can schemaCheck not be part of `FileAwareInputStreamExtractor` itself instead of introducing another class for schemaCheck functionality, since that is the only difference with this class which probably can be controlled by a schemaCheck flag. may be its a separate PR.

gobblin-data-management/src/main/java/org/apache/gobblin/util/schema_check/AvroSchemaCheckStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+     * Use the configuration to create a schema check strategy. If it's not found, return null.
+     * @param state
+     * @return
+     */
+    public static AvroSchemaCheckStrategy create(WorkUnitState state)
+    {
+      try {
+        return (AvroSchemaCheckStrategy) Class.forName(state.getProp(ConfigurationKeys.AVRO_SCHEMA_CHECK_STRATEGY, ConfigurationKeys.AVRO_SCHEMA_CHECK_STRATEGY_DEFAULT)).newInstance();
+      } catch (Exception e)
+      {
```
autumnust: nitpick: usually we would have `{` in the same line of `catch`

gobblin-data-management/src/main/java/org/apache/gobblin/util/schema_check/AvroSchemaCheckStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+        return null;
```
autumnust: Exception caught but not propagated/logged will be a silent failure which is usually not what we wanted. If there's nothing (recovery type of thing) we could do after catching the exception, let's just log the exception as a starting point

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-api/src/main/snapshot/org.apache.gobblin.restli.throttling.permits.snapshot.json
autumnust: Got the point. Just mention in the description that why this changes (which is unrelated to the PR itself) shows up as part of code change.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -936,4 +936,10 @@
   public static final String GIT_MONITOR_SSH_KNOWN_HOSTS = "knownHosts";
   public static final String GIT_MONITOR_SSH_KNOWN_HOSTS_FILE = "knownHostsFile";
   public static final String GIT_MONITOR_JSCH_LOGGER_ENABLED = "isJschLoggerEnabled";
+
+  /**
+   * Configuration related to data derived
```
ibuenros: Please adjust the comment so it makes sense in Gobblin itself. This is leaking LinkedIn implementation details.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
ibuenros: @jhsenjaliya the reason we did this in a different class is because it is leaking Avro details. CopySource and the associated extractor are able to copy arbitrary files simply as byte streams. This particular extension requires the copied files to be Avro.

In general, I think that in a byte copier any functionality that assumes a schema associated with those bytes is pollution of the logic. What do you think?

gobblin-data-management/src/test/java/org/apache/gobblin/data/management/util/AvroSchemaCheckStrategyForTest.java
```diff
@@ -0,0 +1,112 @@
...
+package org.apache.gobblin.data.management.util;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+import org.apache.gobblin.util.schema_check.AvroSchemaCheckStrategy;
+
+
+public class AvroSchemaCheckStrategyForTest implements AvroSchemaCheckStrategy {
```
ibuenros: Is there a reason this needs to be only for testing? Is it robust enough to be made the default implementation in Gobblin?

ibuenros:
+1

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/extractor/FileAwareInputStreamExtractorWithCheckSchema.java
jhsenjaliya: ok, sure. Thanks

PR #2640, opened 2019-05-21
[GOBBLIN-775] Add job level retries for gobblin service

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-775

### Description

- Add option `job.maxRetries` to retry a job in gobblin service.
- Refactor `DagManager` to use a string identifier as the key in `jobToDag` map, since having a `dagNode` as the key causes problems if the underlying object is modified

### Tests

Added tests in `DagManagerTest`

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -391,6 +391,12 @@ public class DagManager extends AbstractIdleService {
             jobExecutionPlan.setExecutionStatus(RUNNING);
             break;
         }
+
+        if (jobStatus.isShouldRetry()) {
+          log.info("Retrying job: {}, current attempts: {}, max attempts: {}", DagManagerUtils.getFullyQualifiedJobName(node),
+              jobStatus.getCurrentAttempts(), jobStatus.getMaxAttempts());
+          submitJob(node);
+        }
       }
 
       for (Map.Entry<String, Set<DagNode<JobExecutionPlan>>> entry: nextSubmitted.entrySet()) {
@@ -452,6 +458,7 @@ public class DagManager extends AbstractIdleService {
      * Submits a {@link JobSpec} to a {@link org.apache.gobblin.runtime.api.SpecExecutor}.
      */
     private void submitJob(DagNode<JobExecutionPlan> dagNode) {
+      DagManagerUtils.incrementJobAttempt(dagNode);
       JobExecutionPlan jobExecutionPlan = DagManagerUtils.getJobExecutionPlan(dagNode);
       jobExecutionPlan.setExecutionStatus(RUNNING);
       JobSpec jobSpec = DagManagerUtils.getJobSpec(dagNode);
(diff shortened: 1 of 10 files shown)


jack-moseley:
@sv2000 please review

jack-moseley:
- Changed `JobExecutionPlan` equals and hashCode instead of changing the key of the `dagNode` maps
- Changed orchestration events to include attempt counter, so that when there's a failure event we can avoid updating the `JobStatus` to failed if there will be a retry

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManagerUtils.java
```diff
@@ -75,8 +75,10 @@ static String getJobName(DagNode<JobExecutionPlan> dagNode) {
    * @return a fully qualified name of the underlying job.
    */
   static String getFullyQualifiedJobName(DagNode<JobExecutionPlan> dagNode) {
-    Config jobConfig = dagNode.getValue().getJobSpec().getConfig();
+    return getFullyQualifiedJobName(dagNode.getValue().getJobSpec().getConfig());
+  }
 
+  public static String getFullyQualifiedJobName(Config jobConfig) {
```
autumnust: Let's give it an another name instead.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -457,6 +462,15 @@ private void submitJob(DagNode<JobExecutionPlan> dagNode) {
       JobSpec jobSpec = DagManagerUtils.getJobSpec(dagNode);
       Map<String, String> jobMetadata = TimingEventUtils.getJobMetadata(Maps.newHashMap(), jobExecutionPlan);
 
+      // Increment submission attempt
```
autumnust: Should this blocked be part of `TimingEventUtils.getJobMetadata` ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -391,6 +392,10 @@ private void pollAndAdvanceDag()
             jobExecutionPlan.setExecutionStatus(RUNNING);
             break;
         }
+
+        if (jobStatus.isShouldRetry()) {
```
autumnust: Only failed job's status will have this flag turned on?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: I don't think the incrementing logic should be in there because it could be called at a time other than job submission. But I moved the logic for updating the map to there.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: Yes, and they also have their status updated to running at the same time.

jack-moseley:
@sv2000 can you merge? Thanks

PR #2647, opened 2019-05-24
[GOBBLIN-781] Skeleton for GaaS DR mode clean transition

autumnust:
Dear Gobblin maintainers,

A little bit background on this PR: 
In Gobblin-as-a-Service deployment, there are extreme cases that the whole data center where a GaaS cluster is deployed went down. In that case we will launch another GaaS cluster to handle request-serving and orchestration. To guarantee a valid working state for DR-Nominated GaaS Cluster, we will need to clean certain left-over states.  Also, not all `FlowSpec`s would be eligible for running in alternative environment due to data affinity or so. So we provide `SpecStore` an ability to tag certain `FlowSpec`s and only those will be picked up by DR-Nominated cluster to work on. 

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-781

### Description

This PR is mainly solving problems mentioned in background sections by: 
- Adding "tag" into `MysqlSpecStore`. 
- Adding `isNominatedDRHanlder ` as a configuration for newly-launched GaaS master node.
- Adding logic inside `scheduleSpecsFromCatalog`  to pick up tagged `FlowSpec`s in DR-nominated node. This is just a skeleton code for now since the real cancellation features is not yet there. 
- Fixing some of API in `SpecStore` and deprecated the old method load all `Spec`s into memory. 
- Some changes in `GobblinServiceManager` to get rid of left-over during prototyping phase.

### Tests
   - Unit test for loading `spec`s with tag. 

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
@@ -93,7 +93,7 @@ public class GobblinServiceJobScheduler extends JobScheduler implements SpecCata
    * e.g. There are multi-datacenter deployment of GaaS Cluster. Intra-datacenter fail-over could be handled by
    * leadership change mechanism, while inter-datacenter fail-over would be handled by DR handling mechanism.
    */
-  private boolean isNominatedDRHanlder;
+  private boolean isNominatedDRHandler;
 
   /**
    * Use this to tag all DR-applicable FlowSpec entries in {@link org.apache.gobblin.runtime.api.SpecStore}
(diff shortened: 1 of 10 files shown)


autumnust:
@sv2000  @arjun4084346  Can you take a look ? Thanks.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -170,6 +195,19 @@ private void scheduleSpecsFromCatalog() {
     }
   }
 
+  /**
+   * In DR-mode, the running {@link FlowSpec} will all be cancelled and rescheduled.
+   * We will need to make sure that running {@link FlowSpec}s' state are cleared, and corresponding running jobs are
+   * killed before rescheduling them.
+   * @param drUris The uris that applicable for DR discovered from FlowCatalog.
+   */
+  private void clearRunningFlowState(Iterator<URI> drUris) {
```
sv2000: Should the DR related failover handling be moved to a separate class? For now, you are triggering it based on config. In the future, we might have a REST end-point or some other mechanism for triggering a fail over, without needing a service restart.

gobblin-restli/gobblin-throttling-service/gobblin-throttling-service-api/src/main/snapshot/org.apache.gobblin.restli.throttling.permits.snapshot.json
```diff
@@ -17,6 +17,18 @@
       "type" : "long",
```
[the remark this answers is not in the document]
autumnust: Addressed

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
autumnust: Actually the structure is pretty isolated since all additional thing added is calling this method. It shouldn't be very hard to migrate to a RESTful design.

PR #2653, opened 2019-05-28
[GOBBLIN-787] Add an option to include the task start time in the out…

htran1:
…put file name

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-787

### Description
In some cases a task may be scheduled to run on multiple workers. One case where this happens is when running with the Helix task execution framework. Helix may reschedule a task on a different worker if it loses contact with a worker. That worker may continue executing for some time before the task is terminated. During this period if the output file names collide then there may be an error during data publish.

Add an option "writer.addTaskTimestamp" that can be used to reduce the chance of name collisions by appending a task startup timestamp to the file name.

### Tests
LocalJobLauncherTest.testLaunchJobWithTaskTimestamps()

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
@@ -529,9 +529,18 @@ public class Fork<S, D> implements Closeable, FinalState, RecordStreamConsumer<S
    */
   private DataWriter<Object> buildWriter()
       throws IOException {
+    String writerId = this.taskId;
+
+    // Add the task starting time if configured.
+    // This is used to reduce file name collisions which can happen due to the scheduling of a task on multiple workers.
+    // One case where this occurs is when a worker gets disconnected from a Helix cluster.
+    if (this.taskState.getPropAsBoolean(ConfigurationKeys.WRITER_ADD_TASK_TIMESTAMP, false)) {
+        writerId = this.taskId + "_" + this.taskState.getProp(ConfigurationKeys.TASK_START_TIME_MILLIS_KEY, "0");
+    }
+
     DataWriterBuilder<Object, Object> builder = this.taskContext.getDataWriterBuilder(this.branches, this.index)
         .writeTo(Destination.of(this.taskContext.getDestinationType(this.branches, this.index), this.taskState))
-        .writeInFormat(this.taskContext.getWriterOutputFormat(this.branches, this.index)).withWriterId(this.taskId)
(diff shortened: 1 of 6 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -529,9 +529,18 @@ protected void cleanup() {
    */
   private DataWriter<Object> buildWriter()
       throws IOException {
+    String writerId = this.taskId;
+
+    // Add the task starting time if configured.
+    // This is used to reduce file name collisions which can happen due to the scheduling of a task on multiple workers.
+    // One case where this occurs is when a worker gets disconnected from a Helix cluster.
+    if (this.taskState.getPropAsBoolean(ConfigurationKeys.WRITER_ADD_TASK_TIMESTAMP, false)) {
+        writerId = this.taskId + "_" + this.taskState.getProp(ConfigurationKeys.TASK_START_TIME_MILLIS_KEY, "0");
```
jhsenjaliya: minor : not sure if all task execution will always come from `GobblinMultiTaskAttempt.java`, so can it use `Long.toString(System.currentTimeMillis())` instead of "0" as default? Also do we have to put `TASK_START_TIME_MILLIS_KEY` while beginning the task? can it not be just any current time when it comes here if its configured? ( start time would be helpful for other metrics though)

jhsenjaliya:
+1, LGTM

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -529,9 +529,18 @@ protected void cleanup() {
    */
   private DataWriter<Object> buildWriter()
       throws IOException {
+    String writerId = this.taskId;
+
+    // Add the task starting time if configured.
+    // This is used to reduce file name collisions which can happen due to the scheduling of a task on multiple workers.
+    // One case where this occurs is when a worker gets disconnected from a Helix cluster.
```
shirshanka: Doesn't it also happen if the task fails on a worker and is subsequently dispatched to another worker? 
File name collisions due to concurrent executions of the same task, or file name collisions due to serial executions of the same task.. both are equally bad.. right?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
shirshanka: If it is a bug to not find a set value for TASK_START_TIME_MILLIS_KEY, choose from:
  a) throw an exception and die
  b) log a warning and continue with a sane default (System.currentTime)

PR #2656, opened 2019-05-31
[GOBBLIN-790]DagStateStore MySQL

autumnust:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-790

### Description
This is the implementation of `DagStateStore` backed by MySQL. 
Instead of implementing its own MySQLl-related logic, it instantiate a `MySQLStateStore` inside the DagStateStore and delegate all interactions between MySQL to that. To archive this, also added: 
- Conversion from `Dag` to `State` by having `Dag` serialized first and put it as a value in State's k-v pair. 
- `getAll` method in `MysqlStateStore` where it used to strictly require `storeName` to show up. 
- Model the schema of `MysqlStateStore`  in the `Dag'gy way: `FlowId` <--> `storeName`,  `FlowExecutionId` <--> `tableName`. 
- Refactor some of test methods to reduce code redundancy. 

### Tests
- Relevant unit tests are added. 

### Commits

### DIFF
--- a/gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlStateStore.java
+++ b/gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlStateStore.java
@@ -125,7 +125,7 @@ public class MysqlStateStore<T extends State> implements StateStore<T> {
 
   private final String UPSERT_JOB_STATE_SQL;
   private final String SELECT_JOB_STATE_SQL;
-  private final String SELECT_ALL_JOB_STATE_SQL;
+  private final String SELECT_ALL_JOBS_STATE_SQL;
   private final String SELECT_JOB_STATE_WITH_LIKE_SQL;
   private final String SELECT_JOB_STATE_EXISTS_SQL;
   private final String SELECT_JOB_STATE_NAMES_SQL;
@@ -152,7 +152,7 @@ public class MysqlStateStore<T extends State> implements StateStore<T> {
     UPSERT_JOB_STATE_SQL = UPSERT_JOB_STATE_TEMPLATE.replace("$TABLE$", stateStoreTableName);
     SELECT_JOB_STATE_SQL = SELECT_JOB_STATE_TEMPLATE.replace("$TABLE$", stateStoreTableName);
     SELECT_JOB_STATE_WITH_LIKE_SQL = SELECT_JOB_STATE_WITH_LIKE_TEMPLATE.replace("$TABLE$", stateStoreTableName);
-    SELECT_ALL_JOB_STATE_SQL = SELECT_ALL_JOBS_STATE.replace("$TABLE$", stateStoreTableName);
+    SELECT_ALL_JOBS_STATE_SQL = SELECT_ALL_JOBS_STATE.replace("$TABLE$", stateStoreTableName);
     SELECT_JOB_STATE_EXISTS_SQL = SELECT_JOB_STATE_EXISTS_TEMPLATE.replace("$TABLE$", stateStoreTableName);
     SELECT_JOB_STATE_NAMES_SQL = SELECT_JOB_STATE_NAMES_TEMPLATE.replace("$TABLE$", stateStoreTableName);
     DELETE_JOB_STORE_SQL = DELETE_JOB_STORE_TEMPLATE.replace("$TABLE$", stateStoreTableName);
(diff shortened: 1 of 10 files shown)


autumnust:
@htran1 @sv2000  Please help take a look, thanks.

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlStateStore.java
```diff
@@ -123,6 +125,7 @@
 
   private final String UPSERT_JOB_STATE_SQL;
   private final String SELECT_JOB_STATE_SQL;
+  private final String SELECT_ALL_JOB_STATE_SQL;
```
sv2000: rename variable to SELECT_ALL_JOBS_STATE_SQL?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -582,11 +581,18 @@ private void cleanUp() {
...
-      this.dagStateStore.cleanUp(dag);
+    /**
+     * Cleaning of all relevant states need to be
+     * @param dagId
+     */
+    private synchronized void cleanUpDag(String dagId) {
+       try {
+         this.dagStateStore.cleanUp(dags.get(dagId));
+       } catch (IOException ioe) {
+         log.error(String.format("Failed to clean %s from backStore due to:", dagId), ioe);
```
sv2000: log.error("Failed to clean {} from backStore due to: ", dagId, ioe);?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -582,11 +581,18 @@ private void cleanUp() {
...
+       }
       this.dags.remove(dagId);
+      this.dagToJobs.remove(dagId);
```
sv2000: Any reason why cleanUp(dag) and remove(dagId) are re-ordered? If the ordering is important, make it clear in the comments.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/MysqlDagStateStore.java
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * The schema of {@link MysqlStateStore} is fixed but the columns are semantically projected into Dag's context:
+   * - The 'storeName' is FlowId.
+   * - The 'tableName' is FlowExecutionId.
+   */
+  private MysqlStateStore<State> mysqlStateStore;
+  private final GsonSerDe<List<JobExecutionPlan>> serDe;
+  private JobExecutionPlanDagFactory jobExecPlanDagFactory;
+
+  public MysqlDagStateStore(Config config, Map<URI, TopologySpec> topologySpecMap) {
```
sv2000: Any chance we can just have a DagStateStoreImpl class that instantiates the appropriate StateStore instance based on the config? Most of what you have implemented is generic and can be used for any StateStore implementation. In other words, we can replace FSDagStateStore and MysqlDagStateStore classes with a single DagStateStoreImpl class.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
autumnust: I believe `log.error (String, Exception)` will format the line better. What is the reason to have them cancatted as single string ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
autumnust: The reason is I get rid of `this.dags.get(dagId);` and there's no reference to a `dag` associated with this `dagId`. So clean up has to happen before `remove`, or `dag` object won't be fetched. 
Will add comment to make it clear

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/MysqlDagStateStore.java
autumnust: Agree. Since this PR is large enough I would like to separate this out:  https://issues.apache.org/jira/browse/GOBBLIN-801

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
sv2000: Not necessarily better. I have seen parametrized logging being used more frequently in Gobblin code.

PR #2663, opened 2019-06-05
[GOBBLIN-796] Add support partial updates for flowConfig

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-796

### Description
Add option to trigger flow through update by setting `runImmediately=true` if `flow.trigger=true` in the request.

### Tests
Updated `FlowConfigTest`

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
@@ -21,7 +21,7 @@ import java.io.File;
 import java.util.Map;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.gobblin.runtime.spec_store.FSSpecStore;
+import org.apache.commons.io.IOUtils;
 import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
@@ -35,8 +35,12 @@ import com.google.inject.Guice;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.name.Names;
+import com.linkedin.data.DataMap;
 import com.linkedin.data.template.StringMap;
+import com.linkedin.restli.common.PatchRequest;
+import com.linkedin.restli.internal.server.util.DataMapUtils;
 import com.linkedin.restli.server.resources.BaseResource;
+import com.linkedin.restli.server.util.PatchApplier;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 
@@ -44,6 +48,7 @@ import org.apache.gobblin.config.ConfigBuilder;
 import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.restli.EmbeddedRestliServer;
 import org.apache.gobblin.runtime.spec_catalog.FlowCatalog;
+import org.apache.gobblin.runtime.spec_store.FSSpecStore;
 
 
 @Test(groups = { "gobblin.service" })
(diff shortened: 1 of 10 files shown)


jack-moseley:
@arjun4084346 please review

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -139,6 +139,7 @@
   public static final String FLOW_ALLOW_CONCURRENT_EXECUTION = "flow.allowConcurrentExecution";
   public static final String FLOW_EXPLAIN_KEY = "flow.explain";
   public static final String FLOW_UNSCHEDULE_KEY = "flow.unschedule";
+  public static final String FLOW_TRIGGER_KEY = "flow.trigger";
```
sv2000: Do we need a separate config? Can't we just use the runImmediately flag which is already available?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigResourceLocalHandler.java
```diff
@@ -171,6 +171,15 @@ public UpdateResponse updateFlowConfig(FlowId flowId, FlowConfig flowConfig, boo
       FlowConfig originalFlowConfig = getFlowConfig(flowId);
       originalFlowConfig.setSchedule(NEVER_RUN_CRON_SCHEDULE);
       flowConfig = originalFlowConfig;
+    } else if (isTriggerRequest(flowConfig)) {
```
sv2000: We do not need special handling to detect a "trigger" request, if we treat all Update FlowConfig operations as upserts i.e. merge the old FlowConfig with the config provided in the update request.

sv2000: Adding a few more thoughts:
For Collections resource type (as is the case with FlowConfigs), the recommended way to support updates to resources is via Partial Updates. At any rate, all full updates can be converted into partial updates.  We should deprecate the use of full updates of FlowConfigs and use partial updates as the only means for supporting updates.

jack-moseley:
@sv2000 as suggested, repurposed this PR to add partial updates for flow config and then triggering a flow can be done by partial update with `runImmediately=true`.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowConfigClient.java
```diff
@@ -121,6 +123,25 @@ public void updateFlowConfig(FlowConfig flowConfig)
     response.getResponse();
   }
 
+  /**
+   * Partially update a flow configuration
+   * @param flowId flow ID to update
+   * @param flowConfigPatch {@link PatchRequest} containing changes to the flowConfig
+   * @throws RemoteInvocationException
+   */
+  public void partialUpdateFlowConfig(FlowId flowId, PatchRequest<FlowConfig> flowConfigPatch) throws RemoteInvocationException {
```
sv2000: Why is the partialUpdateFlowConfig() method duplicated in 3 different classes? Maybe move this to a Utils class?

jack-moseley:
@sv2000 Updated to support in V2.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -122,6 +127,39 @@ public void testCheckFlowExecutionId() throws Exception {
     Assert.assertEquals(_client.createFlowConfig(flowConfig).getFlowExecutionId().longValue(), -1L);
   }
 
+  @Test
+  public void testPartialUpdate() throws Exception {
```
sv2000: Can we test for an exception being thrown in case of a bad partial update?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -119,6 +121,20 @@ public UpdateResponse update(ComplexResourceKey<FlowId, FlowStatusId> key, FlowC
...
 
+  /**
+   * Partial update the flowConfig specified
+   * @param key composite key containing group name and flow name that identifies the flow to update
+   * @param flowConfigPatch patch describing what fields to change
+   * @return {@link UpdateResponse}
+   */
+  @Override
+  public UpdateResponse update(ComplexResourceKey<FlowId, FlowStatusId> key, PatchRequest<FlowConfig> flowConfigPatch) {
+    String flowGroup = key.getKey().getFlowGroup();
```
arjun4084346: Do we not want partialUpdate in V1 of flowConfig ?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigResourceLocalHandler.java
```diff
@@ -188,6 +192,11 @@ public UpdateResponse updateFlowConfig(FlowId flowId, FlowConfig flowConfig) thr
     return updateFlowConfig(flowId, flowConfig, true);
   }
 
+  @Override
+  public UpdateResponse partialUpdateFlowConfig(FlowId flowId, PatchRequest<FlowConfig> flowConfigPatch) throws FlowConfigLoggedException {
+    throw new UnsupportedOperationException("Partial update only supported by GobblinServiceFlowConfigResourceHandler");
```
arjun4084346: shouldn't partial update be supported in localhandler and not in gobblinhandler?

sv2000:
@arjun4084346 - please approve if the changes look good to you.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
jack-moseley: I think we can just add features to V2 from now on, there isn't really any reason to add every feature to both versions.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigResourceLocalHandler.java
jack-moseley: As discussed, GaaS will call it in gobblin service handler and we want to reuse the logic for update in the gobblin service handler class. Since partial update should never be getting called from local handler we are just throwing exception to be clear.

arjun4084346:
+1. LGTM.

PR #2665, opened 2019-06-06
[GOBBLIN-798] Clean up workflows from Helix when the Gobblin applicat…

htran1:
…ion master starts

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-798

### Description
If the application master aborts a new one may be spawned by YARN. The second application master will resubmit the jobs. This results in duplicate jobs in Helix and multiple instances of the job may run, resulting in duplicate data.

The Gobblin application master should clean up all workflows on startup to avoid executing multiple instances of a job.

### Tests
GobblinYarnAppLauncherTest.testJobCleanup()

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
@@ -161,8 +161,4 @@ public class GobblinClusterConfigurationKeys {
 
   public static final String CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "job.cancelRunningJobOnDelete";
   public static final String DEFAULT_CANCEL_RUNNING_JOB_ON_DELETE = "false";
-
-  // for cleaning up jobs on cluster manager startup
-  public static final String CLEAN_UP_JOBS_ON_MANAGER_START = GOBBLIN_CLUSTER_PREFIX + "cleanUpJobsOnManagerStart";
-  public static final boolean DEFAULT_CLEAN_UP_JOBS_ON_MANAGER_START = false;
 }
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
@@ -161,4 +161,8 @@ public class GobblinClusterConfigurationKeys {
 
   public static final String CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "job.cancelRunningJobOnDelete";
   public static final String DEFAULT_CANCEL_RUNNING_JOB_ON_DELETE = "false";
+
+  // for cleaning up jobs on cluster manager startup
+  public static final String CLEAN_UP_JOBS_ON_MANAGER_START = GOBBLIN_CLUSTER_PREFIX + "cleanUpJobsOnManagerStart";
+  public static final boolean DEFAULT_CLEAN_UP_JOBS_ON_MANAGER_START = false;
 }
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterManager.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterManager.java
@@ -126,6 +126,8 @@ public class GobblinClusterManager implements ApplicationLauncher, StandardMetri
 
   private final boolean isStandaloneMode;
 
+  private final boolean cleanUpJobsOnStartup;
+
   @Getter
   protected GobblinHelixMultiManager multiManager;
   @Getter
@@ -145,6 +147,8 @@ public class GobblinClusterManager implements ApplicationLauncher, StandardMetri
     this.config = config;
     this.isStandaloneMode = ConfigUtils.getBoolean(config, GobblinClusterConfigurationKeys.STANDALONE_CLUSTER_MODE_KEY,
         GobblinClusterConfigurationKeys.DEFAULT_STANDALONE_CLUSTER_MODE);
+    this.cleanUpJobsOnStartup = ConfigUtils.getBoolean(config, GobblinClusterConfigurationKeys.CLEAN_UP_JOBS_ON_MANAGER_START,
+        GobblinClusterConfigurationKeys.DEFAULT_CLEAN_UP_JOBS_ON_MANAGER_START);
 
     this.applicationId = applicationId;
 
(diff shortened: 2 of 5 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
```diff
@@ -161,4 +161,8 @@
 
   public static final String CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "job.cancelRunningJobOnDelete";
   public static final String DEFAULT_CANCEL_RUNNING_JOB_ON_DELETE = "false";
+
+  // for cleaning up jobs on cluster manager startup
+  public static final String CLEAN_UP_JOBS_ON_MANAGER_START = GOBBLIN_CLUSTER_PREFIX + "cleanUpJobsOnManagerStart";
+  public static final boolean DEFAULT_CLEAN_UP_JOBS_ON_MANAGER_START = false;
```
sv2000: Is the default "false" to preserve the current behavior with Gobblin cluster in non-Yarn mode?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
```diff
@@ -161,4 +161,8 @@
 
   public static final String CANCEL_RUNNING_JOB_ON_DELETE = GOBBLIN_CLUSTER_PREFIX + "job.cancelRunningJobOnDelete";
   public static final String DEFAULT_CANCEL_RUNNING_JOB_ON_DELETE = "false";
+
+  // for cleaning up jobs on cluster manager startup
+  public static final String CLEAN_UP_JOBS_ON_MANAGER_START = GOBBLIN_CLUSTER_PREFIX + "cleanUpJobsOnManagerStart";
```
sv2000: Shouldn't we always clean up on restart?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinApplicationMaster.java
```diff
@@ -81,7 +83,10 @@
   public GobblinApplicationMaster(String applicationName, ContainerId containerId, Config config,
       YarnConfiguration yarnConfiguration) throws Exception {
     super(applicationName, containerId.getApplicationAttemptId().getApplicationId().toString(),
-        GobblinClusterUtils.addDynamicConfig(config), Optional.<Path>absent());
+        GobblinClusterUtils.addDynamicConfig(config)
+        .withFallback(ConfigFactory.parseMap(
+            ImmutableMap.of(GobblinClusterConfigurationKeys.CLEAN_UP_JOBS_ON_MANAGER_START, "true"))),
```
sv2000: Should it be ImmutableMap.of(GobblinClusterConfigurationKeys.CLEAN_UP_JOBS_ON_MANAGER_START, DEFAULT_CLEAN_UP_JOBS_ON_MANAGER_START) ?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixMultiManager.java
```diff
@@ -363,12 +359,26 @@ void handleLeadershipChange(NotificationContext changeContext) {
...
+    }
+  }
+
   private void cleanUpJobs(HelixManager helixManager) {
     // Clean up existing jobs
     TaskDriver taskDriver = new TaskDriver(helixManager);
 
     Map<String, WorkflowConfig> workflows = taskDriver.getWorkflows();
 
+    log.debug("cleanUpJobs workflow count {} workflows {}", workflows.size(), workflows);
```
sv2000: Maybe just dump workflows.keySet() instead of the entire map?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterManager.java
```diff
@@ -264,6 +268,12 @@ public synchronized void start() {
     this.eventBus.register(this);
     this.multiManager.connect();
 
+    // Standalone mode registers a handler to clean up on leadership change, so don't do the cleanup
+    // now even if the option to clean up on startup is set.
+    if (this.cleanUpJobsOnStartup && !this.isStandaloneMode) {
```
sv2000: Is this check needed for correctness or to avoid duplicate clean up calls? If it is the latter, shouldn't the 2nd call be handled as a No-op?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
htran1: I have this default to true for YARN mode. For standalone mode we clean up on leader ship change. For all other modes the existing behavior is maintained. I wanted to avoid changing behavior as much as possible. Initial startup already blows away the Helix cluster. This is only for yarn restart of the Gobblin application master without a restart of the application launcher.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinApplicationMaster.java
htran1: No, this is explicitly setting it to "true" for the Gobblin app master. All other gobblin cluster modes use the default of "false".

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixMultiManager.java
htran1: Sure.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterManager.java
htran1: It is to avoid duplicate clean up calls. The code would be correct without this check. A standalone mode manager can transition through multiple leadership change events and there is only one clean up call per transition, so there is no need to dedupe/no-op in that code path.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinClusterConfigurationKeys.java
htran1: Yes, this is to keep behavior the same.

htran1: Removed this option since the existing behavior of not cleaning up on startup lead to the issue being fixed in this PR, so it is reasonable to not have an option to preserve the buggy behavior.

PR #2667, opened 2019-06-07
[GOBBLIN-800] Remove the metric context cache from GobblinMetricsRegistry

yukuai518:
Remove the metric context cache from GobblinMetricsRegistry

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-800

### Description
   1) Create the ForkMetrics that can be used by Fork constrct when creating the metric context.
   2) Add the clean up logic in the GobblinMultiTaskAttempt to remove the metrics context hold by GobblinMetricsRegistry.
   3) Fix the Nullpointer exception due to an uninitialized taskAttemptId field inside of the TaskState deserialization.
   4) Some string replacement (refactoring) for GobblinMetrics.
   5) Add the metrics cleanup validation in ClusterIntegrationTest
   6) Add the creatorTag inside JobMetrics. The removal can first check if the JobMetircs was created by the real creator; otherwise it just ignore the removal operation.

### Tests
   Rerun ClusterIntegrationTest
   Rerun MRJobLauncher test

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
@@ -47,7 +47,7 @@ import org.apache.gobblin.commit.CommitStep;
 import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.metastore.StateStore;
-import org.apache.gobblin.metrics.GobblinMetrics;
+import org.apache.gobblin.metrics.Tag;
 import org.apache.gobblin.metrics.event.EventSubmitter;
 import org.apache.gobblin.metrics.event.JobEvent;
 import org.apache.gobblin.runtime.task.TaskFactory;
@@ -89,6 +89,7 @@ public class GobblinMultiTaskAttempt {
   private final Logger log;
   private final Iterator<WorkUnit> workUnits;
   private final String jobId;
+  private final String attemptId;
   private final JobState jobState;
   private final TaskStateTracker taskStateTracker;
   private final TaskExecutor taskExecutor;
@@ -116,6 +117,7 @@ public class GobblinMultiTaskAttempt {
     super();
     this.workUnits = workUnits;
     this.jobId = jobId;
+    this.attemptId = this.getClass().getName() + "." + this.jobId;
     this.jobState = jobState;
     this.taskStateTracker = taskStateTracker;
     this.taskExecutor = taskExecutor;
@@ -415,7 +417,8 @@ public class GobblinMultiTaskAttempt {
       }
     }
 
-    new EventSubmitter.Builder(JobMetrics.get(this.jobId).getMetricContext(), "gobblin.runtime").build()
+    new EventSubmitter.Builder(JobMetrics.get(this.jobId, new JobMetrics.CreatorTag(this.attemptId)).getMetricContext(), "gobblin.runtime")
+        .build()
         .submit(JobEvent.TASKS_SUBMITTED, "tasksCount", Long.toString(countDownLatch.getRegisteredParties()));
 
     return tasks;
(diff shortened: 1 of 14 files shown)


yukuai518:
@sv2000 can you help review?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/TaskMetrics.java
```diff
@@ -61,15 +62,35 @@ public GobblinMetrics call() throws Exception {
...
+    remove(name(task));
+  }
+
   private static String name(TaskState taskState) {
-    return "gobblin.metrics." + taskState.getJobId() + "." + taskState.getTaskId();
+    return METRICS_ID_PREFIX + taskState.getJobId() + "." + taskState.getTaskId();
+  }
+
+  private static String name(Task task) {
+    return METRICS_ID_PREFIX + task.getJobId() + "." + task.getTaskId();
```
sv2000: return name(task.getTaskState())?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ForkMetrics.java
```diff
@@ -0,0 +1,80 @@
...
+    return state.getProp(ConfigurationKeys.FORK_BRANCH_NAME_KEY + "." + index,
+        ConfigurationKeys.DEFAULT_FORK_BRANCH_NAME + index);
+  }
+
+  /**
+   * Creates a {@link List} of {@link Tag}s for a {@link Fork} instance. The {@link Tag}s are purely based on the
+   * index and the branch name.
+   */
+  private static List<Tag<?>> getForkMetricsTags(State state, int index) {
+    return ImmutableList.<Tag<?>>of(new Tag<>(FORK_METRICS_BRANCH_NAME_KEY, getForkMetricsId(state, index)));
```
sv2000: Maybe add a null check here i.e. if (getForkMetricsId(state,index) != null)? Else, ImmutableList.of() may throw an exception.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -451,13 +452,26 @@ private Task createTaskRunnable(WorkUnitState workUnitState, CountDownLatch coun
...
+      // During the task execution, the fork/task instances will create metric contexts (fork, task, job, container)
+      // along the hierarchy up to the root metric context. Although root metric context has a weak reference to
+      // those metric contexts, they are meanwhile cached by GobblinMetricsRegistry. Here we will remove all those
+      // strong reference from the cache to make sure it can be reclaimed by Java GC when JVM has run out of memory.
+
+      this.tasks.forEach(task-> {
+        TaskMetrics.remove(task);
+      });
+
+      JobMetrics.remove(GobblinMetrics.METRICS_ID_PREFIX + jobState.getJobId());
```
ibuenros: I don't think this should be done in here, the job will still run after the tasks are done. If this is needed for distributed mode, then let's check that we are running on a worker before removing the job metrics.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ForkMetrics.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.gobblin.metrics.GobblinMetrics;
+import org.apache.gobblin.metrics.MetricContext;
+import org.apache.gobblin.metrics.Tag;
+import org.apache.gobblin.runtime.TaskState;
+import org.apache.gobblin.runtime.fork.Fork;
+
+/**
+ * An extension to {@link GobblinMetrics} specifically for {@link Fork}.
+ */
+public class ForkMetrics extends GobblinMetrics {
```
ibuenros: Is this just refactoring or is different logic being added?

yukuai518: This is just refactoring

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
yukuai518: Most of our Gobblin code runs in distributed mode, like MRJobLauncher, HelixJobLauncher, etc. I will add the defensive check to make sure that this GobblinMultiTaskAttempt was actually in the separate process.

But even for LocalJobLauncher, GobblinMetricsRegistry is actually a cache, there is no guarantee that JobMetrics cannot be evicted if TaskMetrics is still alive.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ForkMetrics.java
yukuai518: The getForkMetricsId never returns null, because it has a default value.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
yukuai518: @ibuenros Please check my latest change where a creatorTag was added to make sure the JobMetrics can be removed by the correct owner.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ForkMetrics.java
yukuai518: > Minor comments. LGTM. One question: can we enhance the ClusterIntegrationTest to verify that metrics objects are getting removed from the metrics registry?

@sv2000 I have added the metrics cleanup validation in ClusterIntegrationTest. Please review.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/JobMetrics.java
```diff
@@ -36,67 +40,112 @@
...
       public GobblinMetrics call() throws Exception {
-        return new JobMetrics(jobState, parentContext);
+        return new JobMetrics(jobState, parentContext, creatorTag);
       }
     });
   }
 
   /**
    * Get a {@link JobMetrics} instance for the job with the given {@link JobState} instance.
+   * This method has been deprecated. Please consider to use {@link JobMetrics#get(JobState, CreatorTag)}.
```
ibuenros: Use `@deprecated` instead.

ibuenros:
+1

PR #2669, opened 2019-06-11
[GOBBLIN-803] HivePartition record count

autumnust:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-803

### Description

- Collection of record count metadata: Each WorkUnit contains the high watermark and low watermark for a kafka partition and that is equivalent to the number of events handled by that workunit. Given each workunit will output only single HDFS file under a path to be registered, we maintained a map(`pathToRecordCount`) from to-register path to the number of records it contains.  
- Whenever a `HiveSpec` is logically representing a partition (`spec.getPartition.isPresent` equals to `true`), it will be updated with record count information collected in the map `pathToRecordCount` before handed over to `hiveRegister`. 
- Add a unit test to examine correctness of record counting. 

### Tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/publisher/HiveRegistrationPublisher.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/publisher/HiveRegistrationPublisher.java
@@ -19,21 +19,20 @@ package org.apache.gobblin.publisher;
 
 import java.io.IOException;
 import java.util.Collection;
-import java.util.Set;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 
 import org.apache.hadoop.fs.Path;
 
 import com.codahale.metrics.Timer;
 import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
-import com.google.common.collect.Sets;
+import com.google.common.collect.Maps;
 import com.google.common.io.Closer;
 
 import org.apache.gobblin.annotation.Alias;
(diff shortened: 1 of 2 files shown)


autumnust:
@ibuenros Can you help review ? Thanks

gobblin-core/src/main/java/org/apache/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -189,9 +200,27 @@ public void publishMetadata(Collection<? extends WorkUnitState> states) throws I
...
   }
 
-  private static void addRuntimeHiveRegistrationProperties(State state) {
-    // Use seconds instead of milliseconds to be consistent with other times stored in hive
-    state.appendToListProp(HiveRegProps.HIVE_TABLE_PARTITION_PROPS,
-        String.format("%s:%d", DATA_PUBLISH_TIME, TimeUnit.SECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS)));
+  /**
+   * A wrapper for updating partition record count to {@link #pathToRecordCount}.
+   */
+  private void countUpForPath(String path, State state) {
```
jhsenjaliya: this function name says count, but its just getting the record count from the state, is plan to actually count the records in future? otherwise it should be renamed as getRecordCountsFromState or getRecordCounts or something.

gobblin-core/src/main/java/org/apache/gobblin/publisher/HiveRegistrationPublisher.java
```diff
@@ -189,9 +200,27 @@ public void publishMetadata(Collection<? extends WorkUnitState> states) throws I
...
+    long recordCountInWorkUnit = state.getPropAsLong(ConfigurationKeys.WORK_UNIT_HIGH_WATER_MARK_KEY)
```
jhsenjaliya: so record count is basically the diff on watermark ? what if the watermark is date? is that acceptable?

PR #2672, opened 2019-06-17
[GOBBLIN-806]Enable metrics reporter during dataset discovery for retention job

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-806

### Description
 1. Change datasetCleaner to pass the eventSubmitter to dataset finder
 2. Change MultiDatasetFinder to enable datasetFinder with/without EventSubmitter

### Tests
Test on faro for both ConfigBasedCleanabledDatasetFinder and TMSBasedCleanableDatasetsFinder for retention job.

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/profile/MultiCleanableDatasetFinder.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/profile/MultiCleanableDatasetFinder.java
@@ -16,6 +16,7 @@
  */
 package org.apache.gobblin.data.management.retention.profile;
 
+import com.google.common.base.Optional;
 import org.apache.gobblin.data.management.retention.DatasetCleaner;
 
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


ZihanLi58:
@sv2000 Can you take a look at this?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/retention/profile/MultiCleanableDatasetFinder.java
```diff
@@ -52,6 +53,10 @@ public MultiCleanableDatasetFinder(FileSystem fs, Properties jobProps) {
     super(fs, jobProps);
```
sv2000: Should we create a default eventSubmitter here and call MultiCleanableDatasetFinder() which accepts eventSubmitter as an argument?

ZihanLi58: In the super class MultiDatasetFinder, the construction MultiDatasetFinder(fs, jobProps) will create a default eventSubmitter and call MultiDatasetFinder() which accepts eventSubmitter as an argument. So I think there is no need to create a default eventSubmitter here.

sv2000: hmm..ok. Can't we change the super class constructor to MultiDatasetFinder(fs, jobProps, eventSubmitter) and create the submitter in the MultiCleanableDatasetFinder() constructor and pass it to the parent class?

PR #2677, opened 2019-06-24
[GOBBLIN-813] Make SFDC connector support encrypted Salesforce client id and client secret 

haojiwu:
When reading client id and client secret from state properties,
SalesforceConnector will use PasswordManager.readPassword() to decrypt
it.
readPassword() will check value to see if it is in encrypted pattern.
If not, the original value will be returned.
Therefore this changes can be backward compatible.

Dear Gobblin maintainers,

### JIRA

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConnector.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConnector.java
@@ -66,6 +66,11 @@ public class SalesforceConnector extends RestApiConnector {
     log.debug("Authenticating salesforce");
     String clientId = this.state.getProp(ConfigurationKeys.SOURCE_CONN_CLIENT_ID);
     String clientSecret = this.state.getProp(ConfigurationKeys.SOURCE_CONN_CLIENT_SECRET);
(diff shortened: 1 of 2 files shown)


gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConnector.java
```diff
@@ -64,8 +64,10 @@ public SalesforceConnector(State state) {
   @Override
   public HttpEntity getAuthentication() throws RestApiConnectionException {
     log.debug("Authenticating salesforce");
-    String clientId = this.state.getProp(ConfigurationKeys.SOURCE_CONN_CLIENT_ID);
-    String clientSecret = this.state.getProp(ConfigurationKeys.SOURCE_CONN_CLIENT_SECRET);
+    String clientId = PasswordManager.getInstance(this.state)
```
yukuai518: Can we have this decryption logic configurable? So that not all the case we need this decyption. Today BDE team is also relying on this code base.

yukuai518: Considering to reuse the same returned variable from PasswordManager.getInstance(this.state)

yukuai518:
+1 LGTM

PR #2686, opened 2019-07-15
[GOBBLIN-825] Initialize message schema at object construction rather than creating a new instance for every message

vikrambohra:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-825

### Description
An issue with a schema registry client caused this issue. This issue fixes the creation of new instances of the same schema for every message sent. Instead we now create an instance of the schema during object construction and use the same schema instance for all the messages.

### Tests
- overrideSchemaNameAndNamespaceTest in AvroUtilsTest.java
- KeyValueEventObjectReporterTest.java
- KeyValueMetricObjectReporterTest.java

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/KeyValueEventObjectReporter.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/KeyValueEventObjectReporter.java
@@ -58,15 +58,10 @@ public class KeyValueEventObjectReporter extends EventReporter {
   protected List<String> keys;
   protected final String randomKey;
   protected KeyValuePusher pusher;
-  protected Optional<Map<String, String>> namespaceOverride;
-  protected final String topic;
   protected final Schema schema;
 
   public KeyValueEventObjectReporter(Builder builder) {
     super(builder);
-
-    this.topic = builder.topic;
-    this.namespaceOverride = builder.namespaceOverride;
 
     Config config = builder.config.get();
     Config pusherConfig = ConfigUtils.getConfigOrEmpty(config, PUSHER_CONFIG).withFallback(config);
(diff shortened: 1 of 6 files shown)


vikrambohra:
@zxcware

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/KeyValueEventObjectReporter.java
```diff
@@ -58,6 +60,7 @@
   protected KeyValuePusher pusher;
   protected Optional<Map<String, String>> namespaceOverride;
   protected final String topic;
```
zxcware: remove this

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/KeyValueEventObjectReporter.java
```diff
@@ -58,6 +60,7 @@
   protected KeyValuePusher pusher;
   protected Optional<Map<String, String>> namespaceOverride;
```
zxcware: remove this

vikrambohra: done

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/metrics/reporter/KeyValueEventObjectReporter.java
```diff
@@ -83,17 +81,24 @@ public KeyValueEventObjectReporter(Builder builder) {
...
 
     List<Pair<String, GenericRecord>> events = Lists.newArrayList();
     GobblinTrackingEvent event;
 
     while (null != (event = queue.poll())) {
-      GenericRecord record = AvroUtils.overrideNameAndNamespace(event, this.topic, this.namespaceOverride);
+
+      GenericRecord record=event;
+      try {
+        record = AvroUtils.convertRecordSchema(event, schema);
```
htran1: This conversion is expensive to do on every event. This is the same behavior as today, so it is not making things worse, but we should consider supporting pluggable compiled schemas to avoid the need to override the namespace.

vikrambohra: Its not on every event but every new instance of the reporter. So it's probably the order of mappers than number of messages sent out which is probably in the order of thousands.

PR #2688, opened 2019-07-15
[GOBBLIN-827] Add more events

zxcware:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-827

### Description
Add the following events
- `JobStateEventBuilder` to report gobblin job state or MR job state
- `EntityMissingEventBuilder` to report a missing instance of a certain entity

### Tests
- `GobblinEventTest.testJobStateEvent` covers build and parse `JobStateEventBuilder`
- `GobblinEventTest.testEntityMissingEvent` covers build and parse `EntityMissingEventBuilder`

### Commits

### DIFF
--- a/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/EntityMissingEventBuilder.java
+++ b/gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/EntityMissingEventBuilder.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gobblin.metrics.event;
+
+import java.util.Map;
+
+import org.apache.commons.lang.StringUtils;
+
+import lombok.Getter;
+
+import org.apache.gobblin.metrics.GobblinTrackingEvent;
+
+
+/**
+ * An event reporting missing any kind of entity, e.g. WorkUnit, Record, kafka topic
+ */
+public class EntityMissingEventBuilder extends GobblinEventBuilder {
+
+  private static final String ENTITY_MISSING_EVENT_TYPE = "EntityMissingEvent";
+  private static final String INSTANCE_KEY = "entityInstance";
+
+  /**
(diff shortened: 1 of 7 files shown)


gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
```diff
@@ -27,7 +27,7 @@
  */
 public class JobEvent {
 
-  public static final String JOB_STATE = "JobStateEvent";
+  public static final String JOB_STATE = "JobStateEvent"; // TODO: Migrate to JobStateEventBuilder
```
autumnust: Since `JobStateEventBuilder ` is newly created, why not use that now ?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobStateEventBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import org.apache.gobblin.metrics.GobblinTrackingEvent;
+
+
+/**
+ * An event reporting job state. It can be used to report the overall Gobblin job state or
+ * the state of an internal job, e.g MR job
+ */
+public class JobStateEventBuilder extends GobblinEventBuilder {
```
autumnust: what's special about this JobStateEvent so you need to invent a new class for doing this things? What are the alternative events achieving similar functionalities before ?

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobEvent.java
zxcware: It is used to emit MRJob status in this pr.  Migrating gobblin job status to use it is a separate task.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/JobStateEventBuilder.java
zxcware: The idea of all these specific event builders is to implement typed gobblin event. It is described in the class comment of `GobblinEventBuilder`. The motivation is that monitoring system built on gobblin events can inspect event attributes as java object fields, instead of relying on keys of metadata, which could be changed without control.

Currently, we are using the constants in `JobEvent` and `EventSubmitter` to send gobblin job state event.

gobblin-metrics-libs/gobblin-metrics-base/src/main/java/org/apache/gobblin/metrics/event/EntityMissingEventBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+
+import lombok.Getter;
+
+import org.apache.gobblin.metrics.GobblinTrackingEvent;
+
+
+/**
+ * An event reporting missing any kind of entity, e.g. WorkUnit, Record, kafka topic
+ */
+public class EntityMissingEventBuilder extends GobblinEventBuilder {
```
htran1: Builder is a confusing name here since the object is directly used to host the values and not just for building the event. There is a `fromEvent`. Instead of `build`, maybe `toEvent` makes more sense. This can be considered in a future refactoring.

PR #2691, opened 2019-07-17
[GOBBLIN-830] launcher.type to job.launcher.type

jhsenjaliya:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-830

### Description
config key JOB_LAUNCHER_TYPE_KEY = "launcher.type" should be set to "job.launcher.type"

### Tests
- [x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason: config change only/ updates the test cases accordingly

### Commits


jhsenjaliya:
the reason for this change is standardization or making it clear, when i came across this property, I had to look up the code to make sure whether it really means job launcher of something else.
there are some configuration keys that are not standardized or having misleading names, that would need to be updated eventually. 
thought this would be an easy change while upgrading, may be i can just put both and mark this for deprecation, would that work for you guys?

arjun4084346:
Yes, let's use both the configs and mark one as deprecated.

jhsenjaliya:
@arjun4084346 , tried to handle both properties but its getting complex
- {JobLauncherType} is nested, need to taken out, since most of the jobLauncher is defined in gobblin-runtime, it needs to be placed under that module.
- instead of handling if/else at multiple place, I should add  {JobLauncherUtils.getJobLauncherType} method, but that would create circular dependency between gobblin-utility and gobblin-runtime.
This is minor change compare to amount of refactoring that is required to support both properties which is basically touching the complexity of the project that requires proper refactoring ( [GIP2](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=119547565) ) 

Any thoughts/comments ?

jhsenjaliya:
@arjun4084346 , i made whatever changes i had to make to facilitate the backward compatibility, please take a look. Thanks

jhsenjaliya:
@arjun4084346, @sv2000 can u pls take a look? Thanks

PR #2700, opened 2019-08-01
[GOBBLIN-839] Catch all exceptions when getting flow template at runtime

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-839

### Description

If a template throws RuntimeException during resolution when GaaS is started, the exception will be caught and the edge will just not be added to the graph. However if GaaS is already started then a template is changed to something that won't resolve, then every job will fail to start due to not handling the exception.

### Tests

Reproduced issue locally

### Commits


jack-moseley:
@sv2000 please review

arjun4084346:
I would rather catch all the exceptions than catching a broader `Exception` to detect any new unforeseen issue. LGTM otherwise.

jack-moseley:
I think it's better to catch Exception because if there is an unforeseen error it causes all jobs to fail (even if it was only an issue with one job). We also catch Exception when loading the flow edges for the first time.

PR #2701, opened 2019-08-08
GOBBLIN-846: Enhance LogCopier service to handle continuous YARN log …

sv2000:
…aggregation

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-846

### Description
This task enhances LogCopier service to perform continuous log aggregation for long-running Gobblin-on-Yarn applications. It periodically copies rolled container logs to a specified location on HDFS. The "current" log file is copied in the end as part of the service shutdown sequence. 

### Tests
Added unit tests in LogCopierTest

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
@@ -40,7 +40,6 @@ import org.apache.commons.cli.ParseException;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
@@ -146,7 +145,6 @@ public class GobblinTaskRunner implements StandardMetricsBridge {
   private final List<Service> services = Lists.newArrayList();
   protected final String applicationName;
   protected final String applicationId;
-  protected final Optional<ContainerId> containerId;
   private final Path appWorkPath;
   private boolean isTaskDriver;
   private boolean dedicatedTaskDriverCluster;
@@ -159,22 +157,11 @@ public class GobblinTaskRunner implements StandardMetricsBridge {
       String taskRunnerId,
       Config config,
       Optional<Path> appWorkDirOptional) throws Exception {
-    this(applicationName, helixInstanceName, applicationId, taskRunnerId, Optional.absent(), config, appWorkDirOptional);
-  }
-
-  public GobblinTaskRunner(String applicationName,
-      String helixInstanceName,
-      String applicationId,
-      String taskRunnerId,
-      Optional<ContainerId> containerId,
-      Config config,
-      Optional<Path> appWorkDirOptional) throws Exception {
     this.isTaskDriver = ConfigUtils.getBoolean(config, GobblinClusterConfigurationKeys.TASK_DRIVER_ENABLED,false);
     this.helixInstanceName = helixInstanceName;
     this.taskRunnerId = taskRunnerId;
     this.applicationName = applicationName;
     this.applicationId = applicationId;
-    this.containerId = containerId;
     this.dedicatedTaskDriverCluster = ConfigUtils.getBoolean(config,
         GobblinClusterConfigurationKeys.DEDICATED_TASK_DRIVER_CLUSTER_ENABLED, false);
     Configuration conf = HadoopUtils.newConfiguration();
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
@@ -40,6 +40,7 @@ import org.apache.commons.cli.ParseException;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
@@ -143,8 +144,9 @@ public class GobblinTaskRunner implements StandardMetricsBridge {
 
   protected final FileSystem fs;
   private final List<Service> services = Lists.newArrayList();
-  private final String applicationName;
-  private final String applicationId;
+  protected final String applicationName;
+  protected final String applicationId;
+  protected final Optional<ContainerId> containerId;
   private final Path appWorkPath;
   private boolean isTaskDriver;
   private boolean dedicatedTaskDriverCluster;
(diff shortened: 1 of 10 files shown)


sv2000:
@autumnust @htran1 Please review the PR

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -143,8 +144,9 @@
 
   protected final FileSystem fs;
   private final List<Service> services = Lists.newArrayList();
-  private final String applicationName;
-  private final String applicationId;
+  protected final String applicationName;
+  protected final String applicationId;
+  protected final Optional<ContainerId> containerId;
```
autumnust: Are we spilling implementation details (e.g. yarn for here) into a generic-purpose class (`GobblinTaskRunner`) by doing this?

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
```diff
@@ -85,10 +86,7 @@
  *           .useDestFileSystem(FileSystem.get(URI.create(destFsUri), new Configuration()))
  *           .readFrom(new Path(srcLogDir))
  *           .writeTo(new Path(destLogDir))
- *           .useInitialDelay(60)
```
autumnust: I found several initial-delay related comments in the class, please clean as well.

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
```diff
@@ -536,14 +521,15 @@ private void copyChangesOfLogFile(Path srcFile, Path destFile) throws IOExceptio
...
-        // Seek to the the most recent position if it is available
-        LOGGER.debug(String.format("Reading log file %s from position %d", srcFile, this.currentPos));
-        fsDataInputStream.seek(this.currentPos);
         BufferedReader srcLogFileReader = closer.register(
             new BufferedReader(new InputStreamReader(fsDataInputStream, ConfigurationKeys.DEFAULT_CHARSET_ENCODING)));
 
-        FSDataOutputStream outputStream = LogCopier.this.destFs.exists(destFile)
-            ? LogCopier.this.destFs.append(destFile) : LogCopier.this.destFs.create(destFile);
+        //If file already exists, delete and re-copy the file.
+        if (LogCopier.this.destFs.exists(destFile)) {
```
autumnust: Why do we need to re-copy the file ?

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
```diff
@@ -178,53 +188,91 @@ protected Scheduler scheduler() {
     return Scheduler.newFixedRateSchedule(0, this.sourceLogFileMonitorInterval, this.timeUnit);
   }
 
+  private boolean shouldIncludeLogFile(FileStatus logFile) {
```
autumnust: What happen if one thread pass the branch `if (copiedFileNames.contains(logFilePath.getName())) {`, and the other thread just add a file in `copiedFileNames `?

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
```diff
@@ -536,14 +521,15 @@ private void copyChangesOfLogFile(Path srcFile, Path destFile) throws IOExceptio
...
             new BufferedReader(new InputStreamReader(fsDataInputStream, ConfigurationKeys.DEFAULT_CHARSET_ENCODING)));
 
-        FSDataOutputStream outputStream = LogCopier.this.destFs.exists(destFile)
-            ? LogCopier.this.destFs.append(destFile) : LogCopier.this.destFs.create(destFile);
+        //If file already exists, delete and re-copy the file.
+        if (LogCopier.this.destFs.exists(destFile)) {
+          LogCopier.this.destFs.delete(destFile, false);
+        }
+
+        FSDataOutputStream outputStream = LogCopier.this.destFs.create(destFile);
```
autumnust: This call seems creating an entry in namenode directly. Based my previous comment on race condition, if that happens, two threads are copying bytes into the same `destFile`. Then is that a problem if we see existed file and then delete-followed-by-recopy ?

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
sv2000: This should be fine, since only the LogCopier thread will prune copiedFileNames while LogCopyTask will add copiedFileNames. The main LogCopier thread blocks until LogCopyTasks finish and then deletes files from the copiedFileNames.

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
sv2000: Thanks for pointing it out. Will clean it up.

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
sv2000: Agreed. This is un-necessary. This condition will never be satisfied since we filter out files which have already been copied.

gobblin-utility/src/main/java/org/apache/gobblin/util/logs/LogCopier.java
sv2000: This should not happen. The set of files to copy in each iteration does not have duplicates. Further, each iteration finishes only when all copy tasks complete. So we should not have a situation where two or more threads to copy the same file.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
sv2000: Agreed.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnLogSource.java
```diff
@@ -69,21 +78,31 @@ protected boolean isLogSourcePresent() {
...
     return builder.build();
   }
 
+  /**
+   * Return a new (non-cached) {@link FileSystem} instance. The {@link FileSystem} instance
+   * returned by the method has automatic closing disabled. The user of the instance needs to handle closing of the
+   * instance, typically as part of its shutdown sequence.
+   */
+  private FileSystem buildFileSystem(Config config, boolean isLocal) throws IOException {
+    return isLocal ? FileSystem.newInstanceLocal(AUTO_CLOSE_CONFIG)
```
autumnust: This config is new to me and educated me a lot after reading : http://johnjianfang.blogspot.com/2015/03/hadoop-filesystem-internal-cache.html 

Just curious, are you seeing any issues from previous implementation or just want to ensure the lifecycle of `fs` object is fully controlled ?  If the former, what kind of symptom is that, an exception while doing fs operation or silent failure of copy ?

sv2000: Yeah. What happens is that each FileSystem instance is configured with a Shutdown hook that invokes close() when a SIGTERM is issued. If the application has a shutdown sequence that depends on the instance being available (as in our case), we run into FileSystem object already Closed exception.

PR #2707, opened 2019-08-12
[GOBBLIN-851] Provide capability to disable Hive partition schema registration.

yukuai518:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-851

### Description
We had problems when table level schema and partition level schema diverges. Think about the case when user register two partitions : 2019/08/10, 2019/08/11, but schema changes in between(S1->S2). Now the table level has schema S2, but 2019/08/10 will have schema S1. 

Query on the latest schema will cause the old partition failure.

### Tests
   Unit tests in gobblin-hive-registration module.

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnit.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnit.java
@@ -50,6 +50,7 @@ public class HiveRegistrationUnit {
 
   protected final String dbName;
   protected final String tableName;
+  protected final boolean registerSchema;
   protected final List<Column> columns = Lists.newArrayList();
   protected final State props = new State();
   protected final State storageProps = new State();
@@ -85,6 +86,7 @@ public class HiveRegistrationUnit {
 
     this.dbName = builder.dbName;
     this.tableName = builder.tableName;
+    this.registerSchema = builder.registerSchema;
     this.columns.addAll(builder.columns);
     this.props.addAll(builder.props);
     this.storageProps.addAll(builder.storageProps);
@@ -400,6 +402,7 @@ public class HiveRegistrationUnit {
   static abstract class Builder<T extends Builder<?>> {
     private String dbName;
     private String tableName;
+    private boolean registerSchema = true;
     private List<Column> columns = Lists.newArrayList();
     private State props = new State();
     private State storageProps = new State();
@@ -415,6 +418,11 @@ public class HiveRegistrationUnit {
     @SuppressWarnings("unchecked")
     public T withTableName(String tableName) {
       this.tableName = tableName;
+      return (T) this;
+    }
+
+    public T withRegisterSchema(boolean registerSchema) {
+      this.registerSchema = registerSchema;
       return (T) this;
     }
 
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
@@ -211,7 +211,7 @@ public class HiveMetaStoreUtils {
     State props = unit.getStorageProps();
     StorageDescriptor sd = new StorageDescriptor();
     sd.setParameters(getParameters(props));
-    if (unit instanceof HiveTable) {
+    if (unit.isRegisterSchema()) {
       sd.setCols(getFieldSchemas(unit));
     }
     if (unit.getLocation().isPresent()) {
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
@@ -211,7 +211,9 @@ public class HiveMetaStoreUtils {
     State props = unit.getStorageProps();
     StorageDescriptor sd = new StorageDescriptor();
     sd.setParameters(getParameters(props));
-    sd.setCols(getFieldSchemas(unit));
+    if (unit instanceof HiveTable) {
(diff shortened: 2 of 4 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
```diff
@@ -211,7 +211,9 @@ private static StorageDescriptor getStorageDescriptor(HiveRegistrationUnit unit)
     State props = unit.getStorageProps();
     StorageDescriptor sd = new StorageDescriptor();
     sd.setParameters(getParameters(props));
-    sd.setCols(getFieldSchemas(unit));
+    if (unit instanceof HiveTable) {
```
autumnust: I believe this should happen inside `addSchemaProperties` method

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnit.java
```diff
@@ -418,6 +421,11 @@ public T withTableName(String tableName) {
       return (T) this;
     }
 
+    public T withRegisterSchema(boolean registerSchema) {
```
autumnust: What's the usage of this field?

yukuai518: This is the builder method, so that our gobblin-hive-registration can set weather the partition needs to register schema.

yukuai518: Put it this way, the getHivePartition's actual implementation is in gobblin-hive-registration MP, which is out of the open source code base.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
yukuai518: Due to we skip the schema registration in partition, the conversion from HivePartition to Partition will fail due to below exception. The addSchemaProperties is for HivePartition creation, not the HivePartition to Partition conversion.

org.apache.hadoop.hive.serde2.avro.AvroSerdeException: Neither avro.schema.literal nor avro.schema.url specified, can't determine table schema
        at org.apache.hadoop.hive.serde2.avro.AvroSerdeUtils.determineSchemaOrThrowException(AvroSerdeUtils.java:119)
        at org.apache.hadoop.hive.serde2.avro.AvroSerDe.determineSchemaOrReturnErrorSchema(AvroSerDe.java:177)
        at org.apache.hadoop.hive.serde2.avro.AvroSerDe.initialize(AvroSerDe.java:103)
        at org.apache.hadoop.hive.serde2.avro.AvroSerDe.initialize(AvroSerDe.java:80)
        at org.apache.hadoop.hive.serde2.SerDeUtils.initializeSerDe(SerDeUtils.java:520)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreUtils.getDeserializer(HiveMetaStoreUtils.java:399)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreUtils.getFieldSchemas(HiveMetaStoreUtils.java:356)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreUtils.getStorageDescriptor(HiveMetaStoreUtils.java:214)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreUtils.getPartition(HiveMetaStoreUtils.java:164)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreBasedRegister.addOrAlterPartition(HiveMetaStoreBasedRegister.java:458)
        at org.apache.gobblin.hive.metastore.HiveMetaStoreBasedRegister.registerPath(HiveMetaStoreBasedRegister.java:159)
        at org.apache.gobblin.hive.HiveRegister$1.call(HiveRegister.java:109)
        at org.apache.gobblin.hive.HiveRegister$1.call(HiveRegister.java:93)
        at org.apache.gobblin.util.executors.MDCPropagatingCallable.call(MDCPropagatingCallable.java:42)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)

autumnust: What I meant was: The determination of HiveRegistrationUnit is an instance of HiveTable or HivePartition (so that we decide to add schema into storage properties or not)should not happen here but in `addSchemaPropertiesIfRequired` only.  The exception you saw if because `getDeserializer `, as the conversion from Gobblin's Hive object to the real Hive object, requires schema related info to be there for SerDe process, which should not be there totally if a converted object is an HivePartition.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnit.java
autumnust: Fair enough.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
yukuai518: @autumnust , we already had differentiation in addSchemaPropertiesIfRequired method, where we use `isRegisterSchema` flag to check if a schema needs to be added. The logic here is to convert Gobblin hive obj to native hive obj, which should follow the same check, otherwise we will see the exception due to a missing schema. Please check my latest code change.

autumnust:
LGTM + 1

PR #2718, opened 2019-08-20
Security token encryption support in SFDC connector

mvachhani:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-862

### Description
Security token encryption support in SFDC connector so as not to have security token as plain text. This change will decrypt the security token of the encrypt flag is set.

### Tests

### Commits


gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConnector.java
```diff
@@ -83,6 +83,10 @@ public HttpEntity getAuthentication() throws RestApiConnectionException {
       String password = PasswordManager.getInstance(this.state)
           .readPassword(this.state.getProp(ConfigurationKeys.SOURCE_CONN_PASSWORD));
       String securityToken = this.state.getProp(ConfigurationKeys.SOURCE_CONN_SECURITY_TOKEN);
+      if (this.state.getPropAsBoolean(ConfigurationKeys.SOURCE_CONN_DECRYPT_SECURITY_TOKEN, false)) {
```
htran1: Why have the option? The password manager will pass the password through if it is not encoded.

mvachhani: Agreed it's not required. Fixed it.

PR #2721, opened 2019-08-24
[GOBBLIN-873] Add offset look-back option in Kafka consumer

MeghaUpadhyay:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-873

### Description
   Current kafka consumer supports only latest and easiest offset reset options if previous state is not available. This PR is to add a new option , offset look-back , through which offset can be reset to specific amount from the latest offset for each topic partition.

### Tests
  

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -461,8 +461,9 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
         offsets.startAtEarliestOffset();
       } else if (offsetOption.equals(OFFSET_LOOKBACK)) {
         long lookbackOffsetRange = state.getPropAsLong(KAFKA_OFFSET_LOOKBACK , 0L);
-        long offset = offsets.getLatestOffset() - lookbackOffsetRange;
-        LOG.warn(offsetNotFoundMsg + "This partition will start from latest-lookback [ " + offsets.getLatestOffset() + " - " + lookbackOffsetRange + " ]  start offset: " + offset);
+        long latestOffset = offsets.getLatestOffset();
+        long offset = latestOffset - lookbackOffsetRange;
+        LOG.warn(offsetNotFoundMsg + "This partition will start from latest-lookback [ " + latestOffset + " - " + lookbackOffsetRange + " ]  start offset: " + offset);
         try {
           offsets.startAt(offset);
         } catch (StartOffsetOutOfRangeException e) {
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
@@ -90,6 +90,8 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
   public static final String LATEST_OFFSET = "latest";
   public static final String EARLIEST_OFFSET = "earliest";
   public static final String NEAREST_OFFSET = "nearest";
+  public static final String OFFSET_LOOKBACK = "offset_lookback";
+  public static final String TIMESTAMP_LOOKBACK = "timestamp_lookback";
   public static final String BOOTSTRAP_WITH_OFFSET = "bootstrap.with.offset";
   public static final String DEFAULT_BOOTSTRAP_WITH_OFFSET = LATEST_OFFSET;
   public static final String TOPICS_MOVE_TO_LATEST_OFFSET = "topics.move.to.latest.offset";
@@ -440,8 +442,14 @@ public abstract class KafkaSource<S, D> extends EventBasedSource<S, D> {
       offsets.startAtLatestOffset();
     } else if (previousOffsetNotFound) {
 
-      // When previous offset cannot be found, either start at earliest offset or latest offset, or skip the partition
-      // (no need to create an empty workunit in this case since there's no offset to persist).
+      /**
+       * When previous offset cannot be found, either start at earliest offset, latest offset, go back with (latest - lookback)
+       * (long value to be deducted from latest offset in order to avoid data loss) or skip the partition
+       * (no need to create an empty workunit in this case since there's no offset to persist).
+       * In case of no previous state OFFSET_LOOKBACK will make sure to avoid consuming huge amount of data (earlist) and data loss (latest offset)
+       * lookback can be set to any long value where (latest-lookback) is nearest offset for each partition. If computed offset is out of range then
+       * partition will be consumed from latest offset
+       **/
       String offsetNotFoundMsg = String.format("Previous offset for partition %s does not exist. ", partition);
       String offsetOption = state.getProp(BOOTSTRAP_WITH_OFFSET, DEFAULT_BOOTSTRAP_WITH_OFFSET).toLowerCase();
       if (offsetOption.equals(LATEST_OFFSET)) {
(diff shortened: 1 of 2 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -451,7 +459,21 @@ private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, SourceSt
         LOG.warn(
             offsetNotFoundMsg + "This partition will start from the earliest offset: " + offsets.getEarliestOffset());
         offsets.startAtEarliestOffset();
-      } else {
+      } else if (offsetOption.equals(OFFSET_LOOKBACK)) {
+        long lookbackOffsetRange = state.getPropAsLong("kafka.offset.lookback", 0L);
```
shirshanka: refactor to static constant  "kafka.offset.lookback"

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -451,7 +459,21 @@ private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, SourceSt
...
+        long offset = offsets.getLatestOffset() - lookbackOffsetRange;
+        LOG.warn(offsetNotFoundMsg + "This partition will start from latest-lookback [ " + offsets.getLatestOffset() + " - " + lookbackOffsetRange + " ]  start offset: " + offset);
+        try {
+          offsets.startAt(offset);
+        } catch (StartOffsetOutOfRangeException e) {
+          String offsetOutOfRangeMsg = String.format(
+                  "Start offset for partition %s is out of range. Start offset = %d, earliest offset = %d, latest offset = %d.",
+                  partition, offsets.getStartOffset(), offsets.getEarliestOffset(), offsets.getLatestOffset());
+          LOG.warn(offsetOutOfRangeMsg + "This partition will start from the latest offset: " + offsets.getLatestOffset());
```
shirshanka: Can have race condition (latest offset can change) between this log message and the next startAtLatestOffsets call. 

So get latest offset first. 
Then log it and start consuming from it as a provided offset?

gobblin-runtime/src/main/resources/templates/kafka-hdfs.template
```diff
@@ -43,7 +43,10 @@ source.kafka.json.schema=${kafka.schema.json}
 
 topic.whitelist=${kafka.topics}
 org.apache.gobblin.kafka.consumerClient.class="org.apache.gobblin.kafka.client.Kafka09ConsumerClient$Factory"
+# Accepted values - latest, earliest and with offset lookback (latest offset - lookback)
 bootstrap.with.offset=earliest
+#bootstrap.with.offset=earliest
```
shirshanka: remove this line

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -90,6 +90,8 @@
   public static final String LATEST_OFFSET = "latest";
   public static final String EARLIEST_OFFSET = "earliest";
   public static final String NEAREST_OFFSET = "nearest";
+  public static final String OFFSET_LOOKBACK = "offset_lookback";
+  public static final String TIMESTAMP_LOOKBACK = "timestamp_lookback";
```
shirshanka: seems like timestamp lookback is not being used. Do you plan to modify this PR with support for timestamp lookback as well?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
autumnust: In `KafkaSource` there's a configuration `RESET_ON_OFFSET_OUT_OF_RANGE` which configures a specific behavior when offset is out of bound.  Let's use that configuration here to determine the behavior for consistency.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
MeghaUpadhyay: Kafka 0.10.1.0 consumer supports offsetsForTimes. I am exploring on how can I have same functionality for lower kafka versions. I will remove this from this PR and create a new PR with time based lookup. Could you please suggest on how should we incorporate support for new kafka client versions (kafka09+) ?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
MeghaUpadhyay: sure , I have made the changes to use same functionality.

MeghaUpadhyay: I have made changes as per the below comment from autumnust. I am reusing the functionality to check for "RESET_ON_OFFSET_OUT_OF_RANGE" and make decision accordingly

MeghaUpadhyay: We are using this functionality (with lookback) in our production pipeline since many days. It is going to be very useful where data retention is couple of days in kafka and wants to avoid using earliest. Lookback help us in going back to certain offset and start consuming from that point (when we lost previous state information).

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -451,7 +459,43 @@ private WorkUnit getWorkUnitForTopicPartition(KafkaPartition partition, SourceSt
         LOG.warn(
             offsetNotFoundMsg + "This partition will start from the earliest offset: " + offsets.getEarliestOffset());
         offsets.startAtEarliestOffset();
-      } else {
+      } else if (offsetOption.equals(OFFSET_LOOKBACK)) {
+        long lookbackOffsetRange = state.getPropAsLong(KAFKA_OFFSET_LOOKBACK , 0L);
+        long offset = offsets.getLatestOffset() - lookbackOffsetRange;
+        LOG.warn(offsetNotFoundMsg + "This partition will start from latest-lookback [ " + offsets.getLatestOffset() + " - " + lookbackOffsetRange + " ]  start offset: " + offset);
```
shirshanka: You could have an incorrect log line due to a race condition: offsets.getLatestOffset() here could be different from the result on line 464. I'd suggest caching the value in a variable (on line 464) and using it in subsequent code.

MeghaUpadhyay: Sure Thanks .. I have updated this line to store offsets.getLatestOffset() in a variable  and use that in subsequent code. But I am finding same issue in multiple places like here https://github.com/apache/incubator-gobblin/pull/2721/files#diff-1f786615c11702cd74b076ab157da963R479 ... Which might lead to incorrect logging and using different offset in "offsets.startAtLatestOffset();". Please suggest something here.

MeghaUpadhyay:
@shirshanka @autumnust Please review latest changes.

MeghaUpadhyay:
@autumnust Can you please review latest changes. Thanks !

shirshanka:
@MeghaUpadhyay : could you update this PR in line with the Gobblin PR requirements? 

Please go through the PR checklist and make sure you check them all off (jira, description, testing, subject line)
I have created a JIRA for you to link to: https://issues.apache.org/jira/browse/GOBBLIN-873

Check other PR-s for examples: Here is a good example of what we want in the Pull request body: 
https://github.com/apache/incubator-gobblin/pull/2701

I'll merge this in after you've taken care of these!

MeghaUpadhyay:
> @MeghaUpadhyay : could you update this PR in line with the Gobblin PR requirements?
> 
> Please go through the PR checklist and make sure you check them all off (jira, description, testing, subject line)
> I have created a JIRA for you to link to: https://issues.apache.org/jira/browse/GOBBLIN-873
> 
> Check other PR-s for examples: Here is a good example of what we want in the Pull request body:
> #2701
> 
> I'll merge this in after you've taken care of these!

Thanks @shirshanka

shirshanka:
Congrats on your first contribution! Looking forward to more 👍

PR #2722, opened 2019-08-29
GOBBLIN-865: Add feature that enables PK-chunking in partition

arekusuri:
Dear Gobblin maintainers,

### JIRA
  - https://issues.apache.org/jira/browse/GOBBLIN-865

### Description
In SFDC(salesforce) connector, we have partitioning mechanisms to split a giant query to multiple sub queries. There are 3 mechanisms:

simple partition (equally split by time)
dynamic pre-partition (generate histogram and split by row numbers)
user specified partition (set up time range in job file)
However there are tables like Task and Contract are failing time to time to fetch full data.

We may want to utilize PK-chunking to partition the query.

The pk-chunking doc from SFDC - https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/async_api_headers_enable_pk_chunking.htm

### Solution

- In Source, execute the giant query with PK-Chunking and partition to workUnits
  - Create a state with all PK-chunking arguments 
  - Instantiate an extractor with the state
  - Call extractor function to execute the giant query
  - we would get a list of result IDs.
  - generate a list of workUnit( one workUnit contains a slice of the result ID list)
- In extractor, execute the workUnit to fetch resultset
  - identify the workUnit whether it is for pk chunking
  - if not, keep run as before (using the existing 3 partition)
  - if yes, just fetch resultset files by the result id list.

### DIFF
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultIterator.java
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gobblin.salesforce;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.sforce.async.BulkConnection;
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.gobblin.configuration.ConfigurationKeys;
+import org.apache.gobblin.source.extractor.utils.InputStreamCSVReader;
+import org.apache.gobblin.source.extractor.utils.Utils;
+
+
+/**
+ * Result Iterator.
+ * Take jobId and 'batchId:resultId,batchId2:resultId2' as input build a result record iterator.
+ */
+@Slf4j
+public class ResultIterator implements Iterator {
+  private Iterator<ResultStruct> batchIdResultIdIterator;
+  private BulkConnection bulkConnection;
+  private InputStreamCSVReader csvReader;
+  private List<String> csvHeader;
+  private int columnSize;
+  private int urlLoadRetryLimit;
+  private ResultStruct resultStruct;
+  private List<String> currentRecord = null;
+  private Boolean isLoadedCurrentRecord = false;
+  private int currentFileRowCount = 0;
+  private int totalRowCount = 0;
+
+  /**
+   *  constructor
+   *  need to initiate the reader and currentRecord
+   */
+  public ResultIterator(BulkConnection bulkConnection, String jobId, String batchIdResultIdString, int urlLoadRetryLimit) {
+    this.urlLoadRetryLimit = urlLoadRetryLimit;
+    this.bulkConnection = bulkConnection;
+    this.batchIdResultIdIterator = this.parsebatchIdResultIdString(jobId, batchIdResultIdString);
+    if (this.batchIdResultIdIterator.hasNext()) {
+      this.resultStruct = this.batchIdResultIdIterator.next();
+      this.csvReader = this.fetchResultsetAsCsvReader(this.resultStruct); // first file reader
+    } else {
+      throw new RuntimeException("No batch-result id found.");
+    }
+    this.fulfilCurrentRecord();
+    this.csvHeader = this.currentRecord;
+    this.columnSize = this.csvHeader.size();
+    // after fetching cvs header, clean up status
+    this.resetCurrentRecordStatus();
+  }
+
+  /**
+   * call reader.next and set up currentRecord
+   */
+  private void fulfilCurrentRecord() {
+    if (this.isLoadedCurrentRecord) {
+      return; // skip, since CurrentRecord was loaded.
+    }
+    try {
+      this.currentRecord = this.csvReader.nextRecord();
+      if (this.currentRecord == null) { // according InputStreamCSVReader, it returns null at the end of the reader.
+        log.info("Fetched {} - rows: {}", this.resultStruct, this.currentFileRowCount); // print out log before switch result file
+        this.currentFileRowCount = 0; // clean up
+        if (this.batchIdResultIdIterator.hasNext()) { // if there is next file, load next file.
+          this.resultStruct = this.batchIdResultIdIterator.next();
+          this.csvReader = this.fetchResultsetAsCsvReader(resultStruct);
+          this.csvReader.nextRecord(); // read and ignore the csv header.
+          this.currentRecord = this.csvReader.nextRecord();
+        } else {
+          log.info("---- Fetched {} rows -----", this.totalRowCount); // print out log when all records were fetched.
+        }
+      }
+      this.isLoadedCurrentRecord = true;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private void resetCurrentRecordStatus() {
+    this.currentRecord = null;
+    this.isLoadedCurrentRecord = false;
+  }
+
+  @Override
+  public boolean hasNext() {
+    this.fulfilCurrentRecord();
+    return this.currentRecord != null;
+  }
+
+  @Override
+  public JsonElement next() {
+    this.fulfilCurrentRecord();
+    List<String> csvRecord = this.currentRecord;
+    this.resetCurrentRecordStatus();
+    if (csvRecord == null) {
+      throw new NoSuchElementException();
+    }
+    this.currentFileRowCount++;
+    this.totalRowCount++;
+    JsonObject jsonObject = Utils.csvToJsonObject(this.csvHeader, csvRecord, this.columnSize);
+    return jsonObject;
+  }
+
+  /**
+   * resultStruct has all data which identify a result set file
+   * fetch it and convert to a csvReader
+   */
+  private InputStreamCSVReader fetchResultsetAsCsvReader(ResultStruct resultStruct) {
+    String jobId = resultStruct.jobId;
+    String batchId = resultStruct.batchId;
+    String resultId = resultStruct.resultId;
+    log.info("PK-Chunking workunit: fetching [jobId={}, batchId={}, resultId={}]", jobId, batchId, resultId);
+    for (int i = 0; i < this.urlLoadRetryLimit; i++) { // retries
+      try {
+        InputStream is = this.bulkConnection.getQueryResultStream(jobId, batchId, resultId);
+        BufferedReader br = new BufferedReader(new InputStreamReader(is, ConfigurationKeys.DEFAULT_CHARSET_ENCODING));
+        return new InputStreamCSVReader(br);
+      } catch (Exception e) { // skip, for retry
+      }
+    }
+    // tried fetchRetryLimit times, always getting exception
+    throw new RuntimeException("Tried " + this.urlLoadRetryLimit + " times, but couldn't fetch data.");
+  }
+
+  /**
+   * input string format is "batchId:resultId,batchId2:resultId2"
+   * parse it to iterator
+   */
+  private Iterator<ResultStruct> parsebatchIdResultIdString(String jobId, String batchIdResultIdString) {
+    return Arrays.stream(batchIdResultIdString.split(",")).map(x -> x.split(":")).map(x -> new ResultStruct(jobId, x[0], x[1])).iterator();
+  }
+
+  @Data
+  class ResultStruct {
+    private final String jobId;
+    private final String batchId;
+    private final String resultId;
+    public String toString() {
+      return String.format("[jobId=%s, batchId=%s, resultId=%s]", jobId, batchId, resultId);
+    }
+  }
+
+}
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultIterator.java
@@ -65,7 +65,7 @@ public class ResultIterator implements Iterator {
     } else {
       throw new RuntimeException("No batch-result id found.");
     }
-    this.fulfilCurrentRecord();
+    this.fulfillCurrentRecord();
     this.csvHeader = this.currentRecord;
     this.columnSize = this.csvHeader.size();
     // after fetching cvs header, clean up status
@@ -75,7 +75,7 @@ public class ResultIterator implements Iterator {
   /**
    * call reader.next and set up currentRecord
    */
-  private void fulfilCurrentRecord() {
+  private void fulfillCurrentRecord() {
     if (this.isLoadedCurrentRecord) {
       return; // skip, since CurrentRecord was loaded.
     }
@@ -106,13 +106,13 @@ public class ResultIterator implements Iterator {
 
   @Override
   public boolean hasNext() {
-    this.fulfilCurrentRecord();
+    this.fulfillCurrentRecord();
     return this.currentRecord != null;
   }
 
   @Override
   public JsonElement next() {
-    this.fulfilCurrentRecord();
+    this.fulfillCurrentRecord();
     List<String> csvRecord = this.currentRecord;
     this.resetCurrentRecordStatus();
     if (csvRecord == null) {
@@ -154,7 +154,7 @@ public class ResultIterator implements Iterator {
   }
 
   @Data
-  class ResultStruct {
+  static class ResultStruct {
     private final String jobId;
     private final String batchId;
     private final String resultId;
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
@@ -17,14 +17,24 @@
 
 package org.apache.gobblin.salesforce;
 
-public class SalesforceConfigurationKeys {
+public final class SalesforceConfigurationKeys {
+  private SalesforceConfigurationKeys() {
+  }
   public static final String SOURCE_QUERYBASED_SALESFORCE_IS_SOFT_DELETES_PULL_DISABLED =
       "source.querybased.salesforce.is.soft.deletes.pull.disabled";
-  public static final int DEFAULT_SALESFORCE_MAX_CHARS_IN_FILE = 200000000;
-  public static final int DEFAULT_SALESFORCE_MAX_ROWS_IN_FILE = 1000000;
+  public static final int DEFAULT_FETCH_RETRY_LIMIT = 5;
+  public static final String BULK_API_USE_QUERY_ALL = "salesforce.bulkApiUseQueryAll";
 
   // pk-chunking
+  public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkchunking.test.batch.id.list";
+  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkchunking.test.job.id";
+  public static final String SALESFORCE_PARTITION_TYPE = "salesforce.partitionType";
   public static final String PARTITION_PK_CHUNKING_SIZE = "salesforce.partition.PkChunkingSize";
   public static final String PK_CHUNKING_JOB_ID = "_salesforce.job.id";
   public static final String PK_CHUNKING_BATCH_RESULT_IDS = "_salesforce.result.ids";
+  public static final int MAX_PK_CHUNKING_SIZE = 250_000; // this number is from SFDC's doc - https://tinyurl.com/ycjvgwv2
+  public static final int MIN_PK_CHUNKING_SIZE = 20_000;
+  public static final int DEFAULT_PK_CHUNKING_SIZE = 250_000; // default to max for saving request quota
 }
+
+
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
@@ -22,4 +22,9 @@ public class SalesforceConfigurationKeys {
       "source.querybased.salesforce.is.soft.deletes.pull.disabled";
   public static final int DEFAULT_SALESFORCE_MAX_CHARS_IN_FILE = 200000000;
   public static final int DEFAULT_SALESFORCE_MAX_ROWS_IN_FILE = 1000000;
+
+  // pk-chunking
+  public static final String PARTITION_PK_CHUNKING_SIZE = "salesforce.partition.PkChunkingSize";
+  public static final String PK_CHUNKING_JOB_ID = "_salesforce.job.id";
+  public static final String PK_CHUNKING_BATCH_RESULT_IDS = "_salesforce.result.ids";
 }
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
@@ -26,10 +26,10 @@ public final class SalesforceConfigurationKeys {
   public static final String BULK_API_USE_QUERY_ALL = "salesforce.bulkApiUseQueryAll";
 
   // pk-chunking
-  public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkchunking.test.batch.id.list";
-  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkchunking.test.job.id";
+  public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkChunking.testBatchIdList";
+  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkChunking.testJobId";
   public static final String SALESFORCE_PARTITION_TYPE = "salesforce.partitionType";
-  public static final String PARTITION_PK_CHUNKING_SIZE = "salesforce.partition.PkChunkingSize";
+  public static final String PARTITION_PK_CHUNKING_SIZE = "salesforce.partition.pkChunkingSize";
   public static final String PK_CHUNKING_JOB_ID = "_salesforce.job.id";
   public static final String PK_CHUNKING_BATCH_RESULT_IDS = "_salesforce.result.ids";
   public static final int MAX_PK_CHUNKING_SIZE = 250_000; // this number is from SFDC's doc - https://tinyurl.com/ycjvgwv2
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
@@ -21,6 +21,7 @@ import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.FileNotFoundException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URI;
 import java.text.ParseException;
@@ -29,11 +30,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 
-import org.apache.gobblin.salesforce.SalesforceConfigurationKeys;
 import org.apache.http.HttpEntity;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.methods.HttpGet;
@@ -89,6 +90,8 @@ import org.apache.gobblin.source.workunit.WorkUnit;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
+import static org.apache.gobblin.salesforce.SalesforceConfigurationKeys.*;
+
 
 /**
  * An implementation of salesforce extractor for extracting data from SFDC
@@ -105,17 +108,11 @@ public class SalesforceExtractor extends RestApiExtractor {
   private static final int MAX_PK_CHUNKING_SIZE = 250000;
   private static final int MIN_PK_CHUNKING_SIZE = 100000;
   private static final int DEFAULT_PK_CHUNKING_SIZE = 200000;
-  private static final String ENABLE_PK_CHUNKING_KEY = "salesforce.enablePkChunking";
-  private static final String PK_CHUNKING_SIZE_KEY = "salesforce.pkChunkingSize";
   private static final int MAX_RETRY_INTERVAL_SECS = 600;
-  // avoid using too many bulk API calls by only allowing PK chunking only if max partitions is configured <= this
-  private static final int PK_CHUNKING_MAX_PARTITIONS_LIMIT = 3;
   private static final String FETCH_RETRY_LIMIT_KEY = "salesforce.fetchRetryLimit";
   private static final int DEFAULT_FETCH_RETRY_LIMIT = 5;
   private static final String BULK_API_USE_QUERY_ALL = "salesforce.bulkApiUseQueryAll";
   private static final boolean DEFAULT_BULK_API_USE_QUERY_ALL = false;
-  private static final String PK_CHUNKING_SKIP_COUNT_CHECK = "salesforce.pkChunkingSkipCountCheck";
-  private static final boolean DEFAULT_PK_CHUNKING_SKIP_COUNT_CHECK = false;
 
 
   private boolean pullStatus = true;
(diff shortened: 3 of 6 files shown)


mvachhani:
+ @htran1 for review.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceWorkUnit.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.gobblin.salesforce;
+
+import java.util.Iterator;
+import java.util.List;
+import org.apache.gobblin.source.workunit.Extract;
+import org.apache.gobblin.source.workunit.WorkUnit;
+
+
+public class SalesforceWorkUnit extends WorkUnit {
```
htran1: This class won't serialize all the state. Instead of adding this class, you should add a config key for the batch and result id list.
Something like "salesforceSource.batchAndResultList" with a value like "xxx1:zzz1,xxx2:zzz2".
Also, will you retry fetching from this state if they result is still available?

arekusuri: Thanks! Hung. Will remove this class. Will also do test with `launcher.type= MAPREDUCE`. I only tested `type=LOCAL`.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -896,6 +928,43 @@ private void reinitializeBufferedReader() throws IOException, AsyncApiException
     }
   }
 
+  private List<JsonElement> fetchWithRetry(
```
zxcware: bulk api may have retry internally?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,6 +156,95 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
...
+      List<SalesforceExtractor.BatchIdAndResultId> batchIdAndResultIds = executeQueryWithPkChunking(state, previousWatermark);
+      List<WorkUnit> ret = createWorkUnits(sourceEntity, state, batchIdAndResultIds);
+      return ret;
+  }
+
+  private List<SalesforceExtractor.BatchIdAndResultId> executeQueryWithPkChunking(
+      SourceState sourceState,
+      long previousWatermark
+  ) throws RuntimeException {
+    Properties commonProperties = sourceState.getCommonProperties();
```
zxcware: What's the motivation from line 180 - 183? If you want to copy, we can simply do `State state = new State(sourceState)`

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,6 +156,95 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
...
+    Properties specProperties = sourceState.getSpecProperties();
+    State state = new State();
+    state.setProps(commonProperties, specProperties);
+    WorkUnit workUnit = WorkUnit.createEmpty();
+    try {
+      WorkUnitState workUnitState = new WorkUnitState(workUnit, state);
+      workUnitState.setId("test" + new Random().nextInt());
+      workUnitState.setProp(ENABLE_PK_CHUNKING_KEY, true); // set extractor enable pk chunking
+      int chunkSize = workUnitState.getPropAsInt(PARTITION_PK_CHUNKING_SIZE, DEFAULT_PK_CHUNKING_SIZE);
```
zxcware: Is it necessary to introduce another config key to chunk size?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,6 +156,95 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
...
+      workUnitState.setProp(PK_CHUNKING_SIZE_KEY, chunkSize); // set extractor pk chunking size
+      workUnitState.setProp(PK_CHUNKING_SKIP_COUNT_CHECK, true); // don't use count check for we couldn't get count
+      SalesforceExtractor salesforceExtractor = (SalesforceExtractor) this.getExtractor(workUnitState);
+      String entity = sourceState.getProp(ConfigurationKeys.SOURCE_ENTITY);
+      Partitioner partitioner = new Partitioner(sourceState);
+      if (isEarlyStopEnabled(state) && partitioner.isFullDump()) {
+        throw new UnsupportedOperationException("Early stop mode cannot work with full dump mode.");
+      }
+      Partition partition = partitioner.getGlobalPartition(previousWatermark);
+      String condition = "";
```
zxcware: For 199 - 207, Setting the query predicate has additional considerations other than the values. Check out `QueryBasedExtractor.setRangePredicates` for more details.  You probably need to refactor `QueryBasedExtractor.build` method for reusability.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,6 +156,95 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
...
+    String nameSpaceName = state.getProp(ConfigurationKeys.EXTRACT_NAMESPACE_NAME_KEY);
+    Extract.TableType tableType = Extract.TableType.valueOf(state.getProp(ConfigurationKeys.EXTRACT_TABLE_TYPE_KEY).toUpperCase());
+    String outputTableName = sourceEntity.getDestTableName();
+    Extract extract = createExtract(tableType, nameSpaceName, outputTableName);
+
+    List<WorkUnit> workUnits = Lists.newArrayList();
+    int partitionNumber = state.getPropAsInt(SOURCE_SOURCE_MAX_NUMBER_OF_PARTITIONS, 1);
+    int maxPartition = (batchResultIds.size() + partitionNumber - 1)/partitionNumber;
+    List<List<SalesforceExtractor.BatchIdAndResultId>> partitionedResultIds = Lists.partition(batchResultIds, maxPartition);
+    String bulkJobId = batchResultIds.get(0).getBulkJobId();
```
zxcware: We could define some object that includes the jobId and BatchIdAndResultId, given that it's the same for all batches of a job, which helps reduces duplicate data. for example:
```
SalesforceBulkJob {
  String jobId;
  BatchIdAndResultId;
}
```

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: Bulk API has retry. It means if the result set is too big, it would retry 15 times creating/caching the result set
This retry is our logic. When we fetch the result set file, if there is any failure, we want to retry specified times. It won't hurt. if the result file can be fetch at first time, the function returns the result set.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: good to know. Will fix them.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: I was thinking we may want to keep 2nd level PK-chunking and better have different property for them.
As we discussed, we don't 2nd level PK-chunking doesn't make sense. Will remove this property.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
zxcware: If bulk api already does retry, can we remove the retry on our side? Application layer retry after api layer retries mostly won't change the error result if any.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: Made it 
```
class SalesforceBulkJob {
  String jobId;
  List<BatchIdAndResultId> batchIdAndResultIdList;
}
```

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: sfdc retrying is `to prepare and create result set files from query data`. 
After sfdc, create the result set files, we request to fetch/download the files.
Our retrying is for fetch/download files.

arekusuri: Closing the thread. Please reopen it, if necessary.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: How about we leave this refactoring later?
It doesn't look like I can use the function directly. In PK-chunking case, we only have start-time, we don't have end-time. end-time is always the current-time.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -896,6 +927,43 @@ private void reinitializeBufferedReader() throws IOException, AsyncApiException
     }
   }
 
+  private List<JsonElement> fetchPkChunkingResultSetWithRetry(
```
zxcware: So, bulkConnection.getQueryResultStream does not have Retry? 

If we need retry, we should some backoff to avoid immediate attempts as consecutive retries without waiting might be fruitless? In this case, we can leverage `Retryer`, checkout `WriterUtils.mkdirsWithRecursivePermissionWithRetry` for an example. It's straightforward to make the change.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
zxcware: What about inclusiveness? Is the start-time  or end-time inclusive or exclusive?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: You are right! There is no wait! I copied the retry logic from **fetchResultBatchWithRetry**  and didn't look into it.
I have another thought - Since the gobblin would re-execute the workUnit again anyway when the workUnit fails, we can just remove the retry. 
how do you think of?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: We are using **>=**. It is start time. And it should be inclusive. we don’t need end time in our case.
We usually fetch the whole table. In case we want to start from a specific time, users can set up **source.querybased.start.value**.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
zxcware: I see. It's an existing logic. Then that's fine.

htran1:
@arekusuri, can you look at the Travis build failures?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -588,11 +576,41 @@ public String getTimestampPredicateCondition(String column, long value, String v
...
+  }
+
   @Override
   public Iterator<JsonElement> getRecordSetFromSourceApi(String schema, String entity, WorkUnit workUnit,
       List<Predicate> predicateList) throws IOException {
     log.debug("Getting salesforce data using bulk api");
-    RecordSet<JsonElement> rs = null;
+
+    // new version of extractor: bulk api with pk-chunking in pre-partitioning of SalesforceSource
+    if (!workUnit.getProp(PK_CHUNKING_JOB_ID, "").equals("")) {
```
zxcware: Are you checking if `PK_CHUNKING_JOB_ID` exists in workUnit? If so, we can use `workUnit.contains(PK_CHUNKING_JOB_ID)`

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,12 +156,98 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
 
   @Override
   protected List<WorkUnit> generateWorkUnits(SourceEntity sourceEntity, SourceState state, long previousWatermark) {
+    String partitionType = state.getProp(PARTITION_TYPE, "");
+    if (partitionType.equals("PK_CHUNKING")) {
+      return generateWorkUnitsPkChunking(sourceEntity, state, previousWatermark);
```
zxcware: This is where to comment the right use cases to enable `PK_CHUNKING`, as it's the entry point
```
      // pk-chunking only supports start-time by source.querybased.start.value, and does not support end-time.
      // always ingest data later than or equal source.querybased.start.value.
      // we should only pk chunking based work units only in case of snapshot/full ingestion
```

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: Thanks! will do.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: Thanks, will add it!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -779,7 +880,7 @@ public boolean bulkApiLogin() throws Exception {
       BatchInfoList batchInfoList = this.bulkConnection.getBatchInfoList(this.bulkJob.getId());
```
htran1: Remove this?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -979,39 +1118,33 @@ private void fetchResultBatchWithRetry(RecordSetList<JsonElement> rs)
...
-    for (int i = 1; i < batchInfos.length; i++) {
-      BatchInfo bi = batchInfos[i];
-
-      // get refreshed job status
-      bi = this.bulkConnection.getBatchInfo(this.bulkJob.getId(), bi.getId());
-
-      while ((bi.getState() != BatchStateEnum.Completed)
-          && (bi.getState() != BatchStateEnum.Failed)) {
-        Thread.sleep(retryInterval * 1000);
+    for (BatchInfo bi: batchInfos) {
```
htran1: Is it no longer true that the first BatchInfo is for the parent and not the pk chunking batches or are you considering it safe to apply the same check since the state should be completed?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -979,39 +1118,33 @@ private void fetchResultBatchWithRetry(RecordSetList<JsonElement> rs)
...
+      BatchStateEnum state = bi.getState();
+      while (state != BatchStateEnum.Completed && state != BatchStateEnum.Failed && state != BatchStateEnum.NotProcessed) {
+        Thread.sleep(waitInterval * 1000);
         bi = this.bulkConnection.getBatchInfo(this.bulkJob.getId(), bi.getId());
```
htran1: You are assigning to the outside loop variable. Not sure if this is intentional. Also, I don't see state being updated, so won't this loop forever?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -144,28 +142,18 @@
...
   private final boolean bulkApiUseQueryAll;
 
+  private WorkUnitState workUnitState;
+
   public SalesforceExtractor(WorkUnitState state) {
     super(state);
-    this.sfConnector = (SalesforceConnector) this.connector;
-
-    // don't allow pk chunking if max partitions too high or have user specified partitions
-    if (state.getPropAsBoolean(Partitioner.HAS_USER_SPECIFIED_PARTITIONS, false)
```
htran1: How come this check was removed? If the user has specified partitions then that should override PK chunking.

arekusuri: This part code was only for **PK chunking**. We don't need 2nd level PK-chunking any more. (if we have normal pre-partition, we should not use pk chunking. It burns out request quota)

I create separate function - getQueryResultIdsPkChunking for pkchunking. and made it not to depends on class member this.pkChunking, therefore removed pkChunking.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: >>       BatchInfoList batchInfoList = this.bulkConnection.getBatchInfoList(this.bulkJob.getId());
I try not to change the existing code in **getQueryResultIds**. And this line is necessary. **getBatchInfoList** function's return value is **BatchIdAndResultId** which means a list of resultIds with corresponded batch Ids.
One bulkJobId could have multiple results even though it is not a pk-chunking.

I am not use **getQueryResultIds** for pkchunking. If we want to do further code refactoring, let's do it in another ticket.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -779,7 +880,7 @@ public boolean bulkApiLogin() throws Exception {
       BatchInfoList batchInfoList = this.bulkConnection.getBatchInfoList(this.bulkJob.getId());
 
       if (usingPkChunking && bulkBatchInfo.getState() == BatchStateEnum.NotProcessed) {
```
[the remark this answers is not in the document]
arekusuri: Thanks! removed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: I had a ticket for this - https://jira01.corp.linkedin.com:8443/browse/DSS-22221
pkchunking is using this function.
Sometimes, the parent may not be the first element in the list. (see screenshot in ticket)

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: Good catch! Thanks!
I should not use the state variable. My test worked, because I was using break point, the time was enough to let sfdc execute.

I did more refactoring for this part. Pushing code advance. Will do test during today. and update you. Thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -144,28 +142,18 @@
...
+    this.workUnitState = state;
 
+    this.sfConnector = (SalesforceConnector) this.connector;
     this.pkChunkingSize =
         Math.max(MIN_PK_CHUNKING_SIZE,
-            Math.min(MAX_PK_CHUNKING_SIZE, state.getPropAsInt(PK_CHUNKING_SIZE_KEY, DEFAULT_PK_CHUNKING_SIZE)));
+            Math.min(MAX_PK_CHUNKING_SIZE, state.getPropAsInt(PARTITION_PK_CHUNKING_SIZE, DEFAULT_PK_CHUNKING_SIZE)));
 
-    this.pkChunkingSkipCountCheck = state.getPropAsBoolean(PK_CHUNKING_SKIP_COUNT_CHECK, DEFAULT_PK_CHUNKING_SKIP_COUNT_CHECK);
+    this.pkChunkingSkipCountCheck = true;// won't be able to get count
```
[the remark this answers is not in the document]
arekusuri: removed. thanks!

arekusuri:
Test result:
https://ltx1-holdemaz05.grid.linkedin.com:8443/hdfs/jobs/exttest/lumos/uif-avro/uif/Salesforce_Core/User/20190919062550_full/part.task_Salesforce_User_full_1568874183318_0_0.avro

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -598,6 +595,8 @@ public String getTimestampPredicateCondition(String column, long value, String v
     String[] batchIdResultIdArray = partitionPkChunkingBatchIdResultIterator.next().split(":");
     String batchId = batchIdResultIdArray[0];
     String resultId = batchIdResultIdArray[1];
+    log.info(String.format("PK-Chunking work unit: fetching file for (jobId=%s, batchId=%s, resultId=%s) ",
```
htran1: Please use {} logger format instead of String.format.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -775,17 +775,15 @@ public SalesforceBulkJobId getQueryResultIdsPkChunking(String entity, List<Predi
       while (batchState != BatchStateEnum.Completed && batchState != BatchStateEnum.Failed && batchState != BatchStateEnum.NotProcessed) {
         Thread.sleep(retryInterval * 1000);
         executedBatchResponse = connection.getBatchInfo(createdJob.getId(), executeQueryBatch.getId());
-        log.debug("Bulk Api Batch Info:" + executedBatchResponse);
-        log.info("Waiting for bulk resultSetIds");
+        log.info(String.format("Waiting for first batch (jobId=%s, pkChunkingBatchId=%s)", jobId, pkChunkingBatchId));
```
htran1: Same comment to use {} for formatting.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -1130,35 +1120,28 @@ public void closeConnection() throws Exception {
...
-   * @throws AsyncApiException
    */
-  private BatchInfo waitForPkBatches(BatchInfoList batchInfoList, int waitInterval)
-      throws InterruptedException, AsyncApiException {
-    BatchInfo batchInfo = null;
+  private void waitForPkBatches(String jobId, BatchInfoList batchInfoList, int waitInterval)  {
+    long toWait = (long)waitInterval * 1000;
     BatchInfo[] batchInfos = batchInfoList.getBatchInfo();
-
+    log.info(String.format("Waiting for bulk (jobId=%s)", jobId));
```
htran1: Logger {} formatting.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -869,7 +862,7 @@ public SalesforceBulkJobId getQueryResultIdsPkChunking(String entity, List<Predi
       // wait for completion, failure, or formation of PK chunking batches
       while ((bulkBatchInfo.getState() != BatchStateEnum.Completed)
           && (bulkBatchInfo.getState() != BatchStateEnum.Failed)
-          && (!usingPkChunking || bulkBatchInfo.getState() != BatchStateEnum.NotProcessed)) {
+          && (bulkBatchInfo.getState() != BatchStateEnum.NotProcessed)) {
```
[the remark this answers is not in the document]
arekusuri: Let's remove this line - 
`&& (!usingPkChunking || bulkBatchInfo.getState() != BatchStateEnum.NotProcessed)`
This line was for pkchunking logic. And now we won't have any pkchuking comes here. it is safe to remove.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -857,7 +850,7 @@ public SalesforceBulkJobId getQueryResultIdsPkChunking(String entity, List<Predi
 
       BatchInfo bulkBatchInfo = this.bulkConnection.createBatchFromStream(this.bulkJob, bout);
 
-      long expectedSizePerBatch = usingPkChunking ? this.pkChunkingSize : this.getExpectedRecordCount();
+      long expectedSizePerBatch = this.getExpectedRecordCount();
```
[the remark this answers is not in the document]
arekusuri: My original approach was - change this function as less as possible. But you asked to remove a lot useless code, my understanding was to clean up the pk-chunking logic from this function.
I removed the field `this.usingPkChunking` so we have to remove this part.
It won't break anything.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: BTW, we were using this way a lot though. I copied the code :)
I did some searching in our code, I didn't find good example. Can you pint me out a sample code? Are you talking about `MessageFormat`?

arekusuri: thanks for online talk, fixed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: fixed. thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: fixed. thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: thanks for offline talk, fixed. :)

arekusuri:
Hi @htran1 and @zxcware Please hold on this pull request a little bit.
We will try another approach to ensure the code is enough mature before we merge into opensource. 
the doc is here - https://docs.google.com/document/d/1fJ7Gju9tXR8WBbwxct0_l21Ijhb4hZJykYFDhDcRLp4/edit#heading=h.37qi9whhekol

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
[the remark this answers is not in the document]
arekusuri: added code to regard NotProcessed as failed for non-pk-chunking bulk request.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -871,8 +871,10 @@ public SalesforceBulkJobId getQueryResultIdsPkChunking(String entity, List<Predi
 
       // Wait for pk chunking batches
       BatchInfoList batchInfoList = this.bulkConnection.getBatchInfoList(this.bulkJob.getId());
-
-      if (bulkBatchInfo.getState() == BatchStateEnum.Failed) {
+      BatchStateEnum state = bulkBatchInfo.getState();
+      // not sure if the state can be "NotProcessed" in non-pk-chunking bulk request
+      // SFDC doc says - This state is assigned when a job is aborted while the batch is queued
+      if (state == BatchStateEnum.Failed || state == BatchStateEnum.InProgress) {
```
htran1: Do you mean `BatchStateEnum.NotProcessed` instead of `InProgress` here? Should never reach here with `InProgress` state because of the loop above.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
```diff
@@ -17,9 +17,24 @@
...
+  }
   public static final String SOURCE_QUERYBASED_SALESFORCE_IS_SOFT_DELETES_PULL_DISABLED =
       "source.querybased.salesforce.is.soft.deletes.pull.disabled";
-  public static final int DEFAULT_SALESFORCE_MAX_CHARS_IN_FILE = 200000000;
-  public static final int DEFAULT_SALESFORCE_MAX_ROWS_IN_FILE = 1000000;
+  public static final int DEFAULT_FETCH_RETRY_LIMIT = 5;
+  public static final String BULK_API_USE_QUERY_ALL = "salesforce.bulkApiUseQueryAll";
+
+  // pk-chunking
+  public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkchunking.test.batch.id.list";
```
htran1: Please follow the new naming convention for new config, like "salesforce.pkChunking.testBatchIdList" and "salesforce.pkChunking.testJobId" where the dots are for namespacing.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
```diff
@@ -17,9 +17,24 @@
...
+  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkchunking.test.job.id";
+  public static final String SALESFORCE_PARTITION_TYPE = "salesforce.partitionType";
+  public static final String PARTITION_PK_CHUNKING_SIZE = "salesforce.partition.PkChunkingSize";
```
htran1: Should be "salesforce.partition.pkChunkingSize"

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -588,16 +566,41 @@ public String getTimestampPredicateCondition(String column, long value, String v
     return dataTypeMap;
   }
 
+
+  private Boolean isPkChunkingFetchDone = false;
+
+  private Iterator<JsonElement> getRecordSetPkchunking(WorkUnit workUnit) throws RuntimeException {
```
htran1: getRecordSetPkchunking -> getRecordSetPkChunking

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -588,16 +566,41 @@ public String getTimestampPredicateCondition(String column, long value, String v
...
+    if (isPkChunkingFetchDone) {
+      return null; // must return null to represent no more data.
+    }
+    isPkChunkingFetchDone = true; // set to true, never come here twice.
```
htran1: Please add new lines between blocks, so before and after the if and try blocks.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -704,36 +708,114 @@ public boolean bulkApiLogin() throws Exception {
     return success;
   }
 
+  public JobIdAndBatchIdResultIdList getQueryResultIdsPkChunkingFetchOnly(String jobId, String batchIdListStr) {
```
htran1: Please add javadoc for this public method. If this is only for testing then mention that in the javadoc.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,12 +153,120 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
 
   @Override
   protected List<WorkUnit> generateWorkUnits(SourceEntity sourceEntity, SourceState state, long previousWatermark) {
+    log.info("====sfdc connector with pkchk===="); // TODO: remove this after merge back to OS.
```
htran1: Should be removed according to TODO?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -146,12 +153,120 @@ protected void addLineageSourceInfo(SourceState sourceState, SourceEntity entity
...
+
+    List<WorkUnit> workUnits = Lists.newArrayList();
+    int partitionNumber = state.getPropAsInt(SOURCE_MAX_NUMBER_OF_PARTITIONS, 1);
+    List<BatchIdAndResultId> batchResultIds = jobIdAndBatchIdResultIdList.getBatchIdAndResultIdList();
+    int total = batchResultIds.size();
+    // size of every partition should be: math.ceil(total/partitionNumber), use simpler way: (total+partitionNumber-1)/partitionNumber
+    int sizeOfPartition = (total + partitionNumber - 1) / partitionNumber;
+    List<List<BatchIdAndResultId>> partitionedResultIds = Lists.partition(batchResultIds, sizeOfPartition);
+    log.info("----partition strategy: max-parti={}, size={}, actual-parti={}, total={}", partitionNumber, sizeOfPartition, partitionedResultIds.size(), total);
+    for (List<BatchIdAndResultId> resultIds : partitionedResultIds) {
```
htran1: Can you add new lines between the blocks?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -588,16 +566,41 @@ public String getTimestampPredicateCondition(String column, long value, String v
...
   @Override
   public Iterator<JsonElement> getRecordSetFromSourceApi(String schema, String entity, WorkUnit workUnit,
       List<Predicate> predicateList) throws IOException {
     log.debug("Getting salesforce data using bulk api");
-    RecordSet<JsonElement> rs = null;
 
+    // new version of extractor: bulk api with pk-chunking in pre-partitioning of SalesforceSource
+    if (workUnit.contains(PK_CHUNKING_JOB_ID)) {
+      log.info("----pk-chunking get record set----" + workUnit.getProp(PK_CHUNKING_JOB_ID));
+      return getRecordSetPkchunking(workUnit);
```
htran1: What would it take to unify the code? I see that the new `ResultIterator` class is not used for the non-pk chunking bulk fetch.

arekusuri: It won't be an easy change.
We need to change the dynamic-mode to execute all the sub-queries in one batch right after we spilt the giant query, and then we can use the fetching module which PK-Chunking is using and the `ResultIterator` is involved.

However we don't have to do it.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: good catch! removed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: added new lines.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: added javadoc
----
   * same as getQueryResultIdsPkChunking but the arguments are different.
   * this function can take existing batch ids to return JobIdAndBatchIdResultIdList
   * It is for test/debug. developers may want to skip execute query on SFDC, use a list of existing batch ids

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: added. thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: fixed. thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
arekusuri: thanks! fixed. good catch!

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
arekusuri: thanks! fixed

>   public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkChunking.testBatchIdList";
>  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkChunking.testJobId";

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultIterator.java
```diff
@@ -0,0 +1,166 @@
...
+    this.csvHeader = this.currentRecord;
+    this.columnSize = this.csvHeader.size();
+    // after fetching cvs header, clean up status
+    this.resetCurrentRecordStatus();
+  }
+
+  /**
+   * call reader.next and set up currentRecord
+   */
+  private void fulfilCurrentRecord() {
```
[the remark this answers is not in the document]
arekusuri: fixed. thanks!

PR #2730, opened 2019-09-09
GOBBLIN-876: Expose metrics() API in GobblinKafkaConsumerClient to al…

sv2000:
…low consume metrics to be reported.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-876

### Description
Newer Kafka consumer expose metrics() API that report a number of consumer metrics such as lag, latency, etc. which are very useful for monitoring and debugging. We expose a metrics() API in GobblinKafkaConsumerClient to allow consumer metrics to be reported. 

### Tests
NA.

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
+++ b/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
@@ -171,7 +171,7 @@ public class Kafka09ConsumerClient<K, V> extends AbstractBaseKafkaConsumerClient
   }
 
   @Override
-  public Map<String, Metric> metrics() {
+  public Map<String, Metric> getMetrics() {
     Map<MetricName, KafkaMetric> kafkaMetrics = (Map<MetricName, KafkaMetric>) this.consumer.metrics();
     Map<String, Metric> codaHaleMetricMap = new HashMap<>();
 
--- a/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
+++ b/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
@@ -176,7 +176,7 @@ public class Kafka09ConsumerClient<K, V> extends AbstractBaseKafkaConsumerClient
     Map<String, Metric> codaHaleMetricMap = new HashMap<>();
 
     kafkaMetrics
-        .forEach((key, value) -> codaHaleMetricMap.put(canonicalMerticName(value), kafkaToCodaHaleMetric(value)));
+        .forEach((key, value) -> codaHaleMetricMap.put(canonicalMetricName(value), kafkaToCodaHaleMetric(value)));
     return codaHaleMetricMap;
   }
 
@@ -193,7 +193,7 @@ public class Kafka09ConsumerClient<K, V> extends AbstractBaseKafkaConsumerClient
     return gauge;
   }
 
-  private String canonicalMerticName(KafkaMetric kafkaMetric) {
+  private String canonicalMetricName(KafkaMetric kafkaMetric) {
     MetricName name = kafkaMetric.metricName();
     return canonicalMetricName(name.group(), name.tags().values(), name.name());
   }
--- a/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
+++ b/gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
@@ -18,8 +18,10 @@ package org.apache.gobblin.kafka.client;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 
@@ -27,9 +29,13 @@ import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.metrics.KafkaMetric;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
@@ -43,6 +49,7 @@ import com.typesafe.config.ConfigFactory;
 import javax.annotation.Nonnull;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
 
 import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.source.extractor.extract.kafka.KafkaOffsetRetrievalFailureException;
@@ -58,6 +65,7 @@ import org.apache.gobblin.util.ConfigUtils;
  * @param <K> Message key type
  * @param <V> Message value type
  */
+@Slf4j
 public class Kafka09ConsumerClient<K, V> extends AbstractBaseKafkaConsumerClient {
 
   private static final String KAFKA_09_CLIENT_BOOTSTRAP_SERVERS_KEY = "bootstrap.servers";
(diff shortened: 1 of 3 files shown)


gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
```diff
@@ -162,6 +170,35 @@ public KafkaConsumerRecord apply(ConsumerRecord<K, V> input) {
...
+        .forEach((key, value) -> codaHaleMetricMap.put(canonicalMerticName(value), kafkaToCodaHaleMetric(value)));
+    return codaHaleMetricMap;
+  }
+
+  /**
+   * Convert a {@link KafkaMetric} instance to a {@link Metric}.
+   * @param kafkaMetric
+   * @return
+   */
+  private Metric kafkaToCodaHaleMetric(final KafkaMetric kafkaMetric) {
```
autumnust: Isn't this likely to be re-used in other implementation under the same package? I am suggesting to make it package private.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/kafka/client/AbstractBaseKafkaConsumerClient.java
```diff
@@ -84,6 +83,33 @@ public boolean apply(@Nonnull KafkaTopic kafkaTopic) {
...
 
+  /**
+   * A helper method that returns the canonical metric name for a kafka metric. A typical canonicalized metric name would
+   * be of the following format: "{metric-group}_{client-id}_{metric-name}".
+   * @param metricGroup the type of the Kafka metric e.g."consumer-fetch-manager-metrics", "consumer-coordinator-metrics" etc.
+   * @param metricTags any tags associated with the Kafka metric, typically include the kafka client id, topic name, partition number etc.
+   * @param metricName the name of the Kafka metric e.g. "records-lag-max", "fetch-throttle-time-max" etc.
+   * @return the canonicalized metric name.
+   */
+  public String canonicalMetricName(String metricGroup, Collection<String> metricTags, String metricName) {
```
autumnust: Shall we make it `protected` or `package private`?

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
sv2000: Good point. Unfortunately, the KafkaMetric class is specific to a given Kafka version, making code duplication necessary. So it seems that each consumer client version will have its own implementation of the method.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/kafka/client/AbstractBaseKafkaConsumerClient.java
sv2000: Will make the change.

shirshanka:
@sv2000 : why is codecov so red? Something wrong with the codecov plugin?

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
```diff
@@ -162,6 +170,35 @@ public KafkaConsumerRecord apply(ConsumerRecord<K, V> input) {
...
+    if (log.isDebugEnabled()) {
+      log.debug("Processing a metric change for {}", kafkaMetric.metricName().toString());
+    }
+    Gauge<Double> gauge = () -> kafkaMetric.value();
+    return gauge;
+  }
+
+  private String canonicalMerticName(KafkaMetric kafkaMetric) {
```
shirshanka: typo on *Mertic*

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/kafka/client/AbstractBaseKafkaConsumerClient.java
```diff
@@ -84,6 +83,33 @@ public boolean apply(@Nonnull KafkaTopic kafkaTopic) {
...
 
+  /**
+   * A helper method that returns the canonical metric name for a kafka metric. A typical canonicalized metric name would
+   * be of the following format: "{metric-group}_{client-id}_{metric-name}".
+   * @param metricGroup the type of the Kafka metric e.g."consumer-fetch-manager-metrics", "consumer-coordinator-metrics" etc.
+   * @param metricTags any tags associated with the Kafka metric, typically include the kafka client id, topic name, partition number etc.
+   * @param metricName the name of the Kafka metric e.g. "records-lag-max", "fetch-throttle-time-max" etc.
+   * @return the canonicalized metric name.
+   */
+  String canonicalMetricName(String metricGroup, Collection<String> metricTags, String metricName) {
```
shirshanka: who should call it and when? is it a good idea to call it many times with the same input?

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
```diff
@@ -162,6 +170,35 @@ public KafkaConsumerRecord apply(ConsumerRecord<K, V> input) {
...
+    MetricName name = kafkaMetric.metricName();
+    return canonicalMetricName(name.group(), name.tags().values(), name.name());
```
shirshanka: since this method will be called repeatedly with the same input values (the Kafka Metric Names) over the lifetime of the process, does it make sense to cache the mapping of KafkaMetricName -> canonicalMetricName and re-use it?

sv2000:
> @sv2000 : why is codecov so red? Something wrong with the codecov plugin?

Not sure how code coverage is computed. It shows my changes impacting classes that seem totally unrelated.

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
sv2000: Yeah - thought about it. Even if we cache the metric name mappings, we need to compute the key for the map every time. It seems like the key needs to be derived off of metric group, metric name and tags to make the metric unique, which is essentially the canonical representation of the kafka metric.

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
sv2000: Fixed. Thanks!

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/kafka/client/AbstractBaseKafkaConsumerClient.java
sv2000: Added a clarifying comment.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/kafka/client/GobblinKafkaConsumerClient.java
```diff
@@ -86,6 +89,15 @@
    */
   public Iterator<KafkaConsumerRecord> consume(KafkaPartition partition, long nextOffset, long maxOffset);
 
+  /**
+   * API to return underlying Kafka consumer metrics. The individual implementations must translate
+   * org.apache.kafka.common.Metric to Coda Hale Metrics.
+   * @return
+   */
+  public default Map<String, Metric> metrics() {
```
shirshanka: since metrics is not a verb... maybe this method should be called getMetrics()?

gobblin-modules/gobblin-kafka-09/src/main/java/org/apache/gobblin/kafka/client/Kafka09ConsumerClient.java
```diff
@@ -162,6 +170,35 @@ public KafkaConsumerRecord apply(ConsumerRecord<K, V> input) {
     });
   }
 
+  @Override
+  public Map<String, Metric> metrics() {
+    Map<MetricName, KafkaMetric> kafkaMetrics = (Map<MetricName, KafkaMetric>) this.consumer.metrics();
+    Map<String, Metric> codaHaleMetricMap = new HashMap<>();
```
shirshanka: do we need a new instance every time? 
This diff doesn't show when and how this method is being called, so its hard to say if this new instance and copy on each call to metrics is a good idea or not.

sv2000:
> I think the PR can be improved a bit by
> 
> 1. Changing the name of the metrics method to getMetrics
> 2. Not creating extra instances of CodaHale metrics
>    2 is less of a show-stopper than 1

Thanks @shirshanka ! Changed metrics method to getMetrics per your suggestion. For 2, the possibility of TopicPartition reassignments to the underlying Kafka consumer makes it difficult to retain the same map, unless we compute a diff to obtain new metrics on each call to KafkaConsumer#metrics(). For simplicity and due to the fact that this method will be called inside a scheduled thread (typically, once every 30 s or more), the relative overhead of instantiating a new map should be insignificant. As a result, I am keeping the implementation as is.

PR #2731, opened 2019-09-09
[GOBBLIN-877]Add column metadata for partition for inline hive registration

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-877

### Description
Previously, we remove the schema.literal for partition.  Because Avro schemas should only be defined at the table level. Hive overrides table properties if the same property is defined on the partition. Defining them at the partition level may lead to partitions with inconsistent schemas. And because column metadata is calculated from schema.literal, so we remove the column metadata as well.

Then we encounter a problem that presto cannot read data from orc file. Because ORC (and other Hive serdes) need metadata in the partitions so that coercion can be done between a partition schema and the table schema.

So we need to treat Avro and other formate separately to make sure hive registration works well so that user can read right data from Presto.

### Tests
Test on dataset written in orc format and I can read data from presto 

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreUtils.java
@@ -212,7 +212,7 @@ public class HiveMetaStoreUtils {
     State props = unit.getStorageProps();
     StorageDescriptor sd = new StorageDescriptor();
     sd.setParameters(getParameters(props));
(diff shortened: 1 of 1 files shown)


ZihanLi58:
@sv2000 @autumnust Can you take a look at this minor change? Thank you

sv2000:
@ZihanLi58 can you check why Travis failed?

ZihanLi58:
> @ZihanLi58 can you check why Travis failed?
I think there is something wrong in the sever side because I cannot reproduce it locally. I'll try to force it run build again?

PR #2751, opened 2019-10-03
[GOBBLIN-895] Fixes Gobblin Standalone configs and scripts so that user guide is accurate

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-895

### Description

[#2578] Introduced inconsistencies with our user guide and the startup scripts for gobblin standalone. I modified the configs and logging output so that the user guide is now accurate again and updated the docs. Also fixed some spelling typos.

### Tests

### Commits


Will-Lo:
@sv2000 Could you take a look? Thanks!

bin/gobblin.sh
```diff
@@ -238,6 +238,11 @@ if [[ "$GOBBLIN_MODE_TYPE" == "$CLI" ]]; then
     fi
 fi
 
+CHECK_ENV_VARS=false
+if [ $ACTION == "start" ] || [ $ACTION == "restart" ]; then
```
jhsenjaliya: actually, only start and restart goes through the function that does this check. Also the setting of ENV variables should be optional.

bin/gobblin.sh
```diff
@@ -417,7 +427,15 @@ function start() {
             CLASS_N_ARGS=''
             if [[ "$GOBBLIN_MODE" = "$STANDALONE_MODE" ]]; then
                 CLASS_N_ARGS="$STANDALONE_CLASS $GOBBLIN_CONF/application.conf"
+                ADDITIONAL_ARGS="-Dgobblin.logs.dir=${GOBBLIN_LOGS}"
+
+                if [ -z "$GOBBLIN_WORK_DIR" ] && [ "$CHECK_ENV_VARS" == true ]; then
```
jhsenjaliya: I originally removed the ENV variables deliberately to make one way of controlling everything. but i can see ENV variable can be additional way of configuring too. but should not be the only way. may be we can merge it or make it overridable by user config for GOBBLIN_JOB_CONFIG_DIR and GOBBLIN_WORK_DIR.

jhsenjaliya:
@Will-Lo , I am trying to see how can we make this consistent, I think with addition of GOBBLIN_WORK_DIR and GOBBLIN_JOB_CONFIG_DIR envrionment varialble support with ability to override from command line would perfectly work. what do you think? I will make the changes accordingly. Thanks

jhsenjaliya:
not it looks much better with #2788, pls check it out.

Will-Lo:
@jhsenjaliya sure we can go with that, I think the main thing is that the documentation should be updated to reflect the change in workflows that we introduce. I think following an approach with #2788 is good, we just need to test it to ensure we don't introduce regressions

jhsenjaliya:
Sure, let me updates the docs accordingly.

bin/gobblin.sh
jhsenjaliya: btw, there is a syntax error here as per bash 4, did you encounter this ?

jhsenjaliya: @Will-Lo , this line has syntax error as per bash 4.0, can u pls take a look.
since `CHECK_ENV_VARS` check happens only in start function anyway, do you want to remove this conditional set and set it by default to TRUE always. Also if you set the variable at  line 34, it can be overridable by `gobblin-env.sh`.

PR #2763, opened 2019-10-14
[GOBBLIN-909] Return error message for unresolved substitutions in explain query

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-909

### Description

- If an explain query does not return a compiled flow, instead return a list of error messages from the config that was not successfully resolved
- To make this more useful, added a whitelisted edges property that specifies which edges are allowed (otherwise there may be many unresolved substitutions and it's not clear which one was the issue)

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
@@ -20,6 +20,7 @@ package org.apache.gobblin.runtime.api;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
@@ -73,6 +74,9 @@ public class FlowSpec implements Configurable, Spec {
   /** Child {@link Spec}s to this {@link FlowSpec} **/
   // Note that a FlowSpec can be materialized into multiple FlowSpec or JobSpec hierarchy
   final Optional<List<Spec>> childSpecs;
+
+  /** List of exceptions that occurred during compilation of this FlowSpec **/
+  final Set<String> compilationErrors = new HashSet<>();
 
   public static FlowSpec.Builder builder(URI flowSpecUri) {
     return new FlowSpec.Builder(flowSpecUri);
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
@@ -39,6 +39,7 @@ import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.util.ConfigUtils;
 
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 
 
 /**
@@ -50,6 +51,7 @@ import lombok.Data;
  */
 @Alpha
 @Data
+@EqualsAndHashCode(exclude={"compilationErrors"})
 public class FlowSpec implements Configurable, Spec {
   /** An URI identifying the flow. */
   final URI uri;
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
@@ -51,10 +51,10 @@ public interface FlowTemplate extends Spec {
 
   /**
    * @param userConfig a list of user customized attributes.
-   * @return list of input/output {@link DatasetDescriptor}s that fully resolve the {@link FlowTemplate} using the
-   * provided userConfig.
+   * @param resolvable if true, only return descriptors that resolve the {@link FlowTemplate}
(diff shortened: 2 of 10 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
```diff
@@ -65,7 +65,7 @@
    * @param outputDescriptor output {@link DatasetDescriptor}
    * @return true if the {@link FlowTemplate} is resolvable
```
arjun4084346: you forgot to update the javadoc

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/StaticFlowTemplate.java
```diff
@@ -103,11 +103,12 @@ public StaticFlowTemplate(URI uri, String version, String description, Config co
   /**
    * Generate the input/output dataset descriptors for the {@link FlowTemplate}.
    * @param userConfig
-   * @return a List of Input/Output DatasetDescriptors that resolve this {@link FlowTemplate}.
+   * @param resolvable
```
arjun4084346: what does resolvable mean?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompilerTest.java
```diff
@@ -640,7 +640,18 @@ public void testCompileCombinedDatasetFlow() throws Exception {
...
+    FlowSpec spec = createFlowSpec("flow/flow5.conf", "HDFS-1", "HDFS-3", false, false);
+
+    Dag<JobExecutionPlan> dag = specCompiler.compileFlow(spec);
+
+    Assert.assertTrue(dag.isEmpty());
+    Assert.assertEquals(spec.getCompilationErrors().size(), 1);
+    Assert.assertTrue(spec.getCompilationErrors().iterator().next().contains("user.to.proxy"));
+  }
+
+  @Test (dependsOnMethods = "testUnresovedFlow")
```
arjun4084346: does it depend on testUnresovedFlow ?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/StaticFlowTemplate.java
```diff
@@ -165,7 +167,7 @@ private Config getResolvedFlowConfig(Config userConfig) {
    * @return true if the {@link FlowTemplate} is resolvable
```
arjun4084346: please update return types and what they mean

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
```diff
@@ -74,6 +75,9 @@
   // Note that a FlowSpec can be materialized into multiple FlowSpec or JobSpec hierarchy
   final Optional<List<Spec>> childSpecs;
 
+  /** List of exceptions that occurred during compilation of this FlowSpec **/
```
arjun4084346: should we exclude compilationErrors from @EqualsAndHashCode ?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompilerTest.java
jack-moseley: Just following the pattern for the rest of the test, I think the reason is that the setup only needs to be done once then.

arjun4084346:
+1 LGTM

jack-moseley:
@sv2000 please merge

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/StaticFlowTemplate.java
```diff
@@ -176,10 +178,10 @@ public boolean isResolvable(Config userConfig, DatasetDescriptor inputDescriptor
       try {
         this.jobSpecResolver.resolveJobSpec(jobSpecBuilder.withTemplate(template).build());
       } catch (JobTemplate.TemplateException | ConfigException | SpecNotFoundException exc) {
-        return false;
+        return exc;
```
sv2000: See comment earlier. We probably can throw the Exception instead of catching here, and let the caller handle the exception.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template/FlowTemplate.java
```diff
@@ -63,9 +63,9 @@
    * @param userConfig User supplied Config
    * @param inputDescriptor input {@link DatasetDescriptor}
    * @param outputDescriptor output {@link DatasetDescriptor}
-   * @return true if the {@link FlowTemplate} is resolvable
+   * @return any exception that occurred when resolving or null if the {@link FlowTemplate} is resolvable
```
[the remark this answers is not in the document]
jack-moseley: Changed to be void function and handle the exception outside (there is not really any point in it being boolean since it would always return true or throw exception).

PR #2770, opened 2019-10-17
[GOBBLIN-916] Make ContainerLaunchContext instantiation in YarnService more efficient

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-916

### Description
Handle NPE in containerCompleteAction and change the excutor of contextLaunch to be scaling thread pool excutor

### Tests
Test on test cluster to make sure the wait time for initialize contextLaunch will not accumulate.

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
@@ -32,10 +32,10 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.gobblin.util.executors.ScalingThreadPoolExecutor;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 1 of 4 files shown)


ZihanLi58:
@sv2000 can you take a look at this minor change?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -610,6 +610,9 @@ private boolean shouldStickToTheSameNode(int containerExitStatus) {
    */
   private void handleContainerCompletion(ContainerStatus containerStatus) {
     Map.Entry<Container, String> completedContainerEntry = this.containerMap.remove(containerStatus.getContainerId());
+    if(completedContainerEntry == null) {
+      LOGGER.warn(String.format("Container %s already been removed from containerMap before completed, there may be something wrong when starting the container", containerStatus.getContainerId()));
+    }
     String completedInstanceName = completedContainerEntry.getValue();
```
sv2000: We may still get a NPE on line 616 if completedContainerEntry is null. I think the fix probably should be to not remove the instance from the containerMap inside NMClientCallbackHandler#onStartContainerError(), since handleContainerCompletion() should be called even if there is a container launch error. Also, the double removal is preventing a replacement container being launched from handleContainerCompletion() method.

ZihanLi58:
> Any chance we can actually verify that onStartContainerError() and handleContainerCompletion() are both invoked? Is it possible to subscribe to the eventbus defined in YarnService and handle the events emitted in each of these method calls?

We have already emitted "HelixInstanceCompletion" event in handleContainerCompletion() and "ContainerStartError" event in onStartContainerError(). And the eventbus is used to handle container-related requests, which seems not related to this.

PR #2777, opened 2019-10-22
[GOBBLIN-921]Make pull/push mode when registering partition to be configurable

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
- [ ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-921] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-921

### Description
In pull mode, register will first try to check if the partition has already existed to reduce the call to add_partition. In push mode, register will try to call add_partition directly and relying on the exception to determine whether existed which mode should be used when most of the partition the HiveRegister try to register is not existed to reduce the call to HiveMetaStore.

### Tests
Test on cluster

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -103,6 +103,9 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
   public static final String PATH_REGISTER_TIMER = HIVE_REGISTER_METRICS_PREFIX + "pathRegisterTimer";
   public static final String SKIP_PARTITION_DIFF_COMPUTATION = HIVE_REGISTER_METRICS_PREFIX + "skip.partition.diff.computation";
   public static final String FETCH_LATEST_SCHEMA = HIVE_REGISTER_METRICS_PREFIX + "fetchLatestSchemaFromSchemaRegistry";
+  //A config which when enabled checks for the existence of a partition in Hive before adding the partition.
+  // This is done to minimize the add_partition calls sent to Hive.
+  public static final String REGISTER_PARTITION_WITH_PULL_MODE = HIVE_REGISTER_METRICS_PREFIX + "registerPartitionWithPullMode";
   /**
    * To reduce lock aquisition and RPC to metaStoreClient, we cache the result of query regarding to
    * the existence of databases and tables in {@link #tableAndDbExistenceCache},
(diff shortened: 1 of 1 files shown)


ZihanLi58:
@sv2000 Here is the separate PR for hive registration, can you take a look at it?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -492,7 +499,7 @@ public void addOrAlterPartition(HiveTable table, HivePartition partition) throws
     }
   }
 
-  private void addOrAlterPartition(IMetaStoreClient client, Table table, HivePartition partition)
+  private void addOrAlterPartitionWithPushMode(IMetaStoreClient client, Table table, HivePartition partition)
```
sv2000: Not a blocker for this PR. But if "pull" mode is the way to go, maybe we should remove the "push" mode in future. Same for skipping diff computations on partitions. Not sure if it is needed.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -128,7 +132,7 @@
    * when the first time a table/database is loaded into the cache, whether they existed on the remote hiveMetaStore side.
    */
   CacheLoader<String, Boolean> cacheLoader = new CacheLoader<String, Boolean>() {
-  @Override
+    @Override
```
sv2000: Minor nit: Extra white space here. Likely unintentional.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: Since when we call add_partition, they do a check before really adding to the meta store, so the mainly cost here is about handling exception. In batch mode, most of the partitions it tries to register are not existed. In that case, I think push mode is a better way to go.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: Will address

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
sv2000: Fair point!

PR #2781, opened 2019-10-23
[GOBBLIN-913] Add MySQL and configurations to cluster

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-913

### Description
Adds SQL to the Kubernetes cluster so that flowspecs are stored in sql instead of the local file system

### Tests
Modifying runtime environment

### Commits


bin/gobblin.sh
```diff
@@ -105,7 +107,7 @@ function print_gobblin_cli_usage() {
                 job-store-schema-manager    Database job history store schema manager
                 gobblin-classpath           shows the constructed gobblin classpath"
     echo ""
-    echo "    --conf-dir <gobblin-conf-dir-path>    Gobblon config path. default is '\$GOBBLIN_HOME/conf/<exe-mode-name>'."
+    echo "    --conf-dir <gobblin-conf-dir-path>    Gobblin config path. default is '\$GOBBLIN_HOME/conf/<exe-mode-name>'."
```
jhsenjaliya: lol, Thanks for taking care of such spelling mistakes..

bin/gobblin.sh
```diff
@@ -135,6 +137,7 @@ function print_gobblin_service_usage() {
     echo "    --job-conf-file <job-conf-file-path>  Only for mapreduce mode: configuration file for the job to run"
     echo "    --help                                Display this help."
     echo "    --verbose                             Display full command used to start the process."
+    echo "    --args                                Additional -D args"
```
jhsenjaliya: "args" can be confusing since its basically a java arguments
btw isnt this covered as jvmopts ? since any strings passed as jvmopts will be placed in java command.

gobblin-docker/gobblin-service/alpine-gaas-latest/entrypoint.sh
```diff
@@ -15,8 +15,24 @@
...
+            shift
+        ;;
+    esac
+    shift
+done
 
-./bin/gobblin.sh service gobblin-as-service start alpine-gaas-latest_gobblin-standalone
-busybox tail -F $GOBBLIN_HOME/logs/gobblin-as-service.out
+./bin/gobblin.sh service gobblin-as-service start ${ARG_FLAG} "${CONFIG_ARGS}" $@
+busybox tail -F $GOBBLIN_HOME/logs/gobblin-as-service.out $GOBBLIN_HOME/logs/gobblin-as-service.err
```
jhsenjaliya: does busybox becomes requirements/dependencies for GAAS ?

bin/gobblin.sh
Will-Lo: Haha no probs

bin/gobblin.sh
Will-Lo: Ah true I missed that, thanks for pointing it out!

gobblin-docker/gobblin-service/alpine-gaas-latest/entrypoint.sh
Will-Lo: It was to workaround a bug with tailing files in kubernetes, but busybox becomes unneeded in https://github.com/apache/incubator-gobblin/pulls
Regardless, this file is for executing GAAS in a dockerized context so the dependencies should be all within the container

bin/gobblin.sh
```diff
@@ -400,7 +401,7 @@ function start() {
             fi
 
             # execute the command
-            $JAVA_HOME/bin/java $GC_OPTS $JVM_OPTS -cp $GOBBLIN_CLASSPATH $CLI_CLASS $CMD_PARAMS
+            $JAVA_HOME/bin/java $CONFIG_ARGS $GC_OPTS $JVM_OPTS -cp $GOBBLIN_CLASSPATH $CLI_CLASS $CMD_PARAMS
```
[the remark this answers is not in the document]
Will-Lo: Oh I forgot to remove this, good catch

gobblin-kubernetes/gobblin-service/dev-cluster/application.yaml
```diff
@@ -0,0 +1,110 @@
...
+          volumeMounts:
+            - name: shared-jobs
+              mountPath: /tmp/gobblin-as-service/jobs
+            - name: shared-template-catalogs
+              mountPath: /tmp/templateCatalog
+            - name: gaas-config
+              mountPath: /home/gobblin/conf/gobblin-as-service/application.conf
+              subPath: gaas-application.conf
+      # dependency on mysql to be initialized before gaas can be initialized
+      initContainers:
```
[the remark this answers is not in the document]
Will-Lo: Other features of K8s to help with dependencies between pods that I'm aware of besides `initContainers` is to define a custom `readiness probe` and `readiness probe` to help with applications that take longer to boot up, I think we could incorporate that in the future as well to have a more robust startup

gobblin-kubernetes/gobblin-service/dev-cluster/application.yaml
```diff
@@ -0,0 +1,110 @@
...
+                  key: username
+            - name: MYSQL_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql-credentials
+                  key: password
+          volumeMounts:
+            - name: shared-jobs
+              mountPath: /tmp/gobblin-as-service/jobs
+            - name: shared-template-catalogs
```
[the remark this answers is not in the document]
Will-Lo: `I'll remove share-template-catalog`, I was hoping to use it to share default templates among pods but I determined a better way of doing that in a future PR using configMaps
`gaas-config` is for the `.conf` file for GaaS, I use this to replace the one defined in the `conf` folder due a change in the configs to support MySQL

bin/gobblin.sh
Will-Lo: @jhsenjaliya I looked through the scripts again, you mean `jvmflags` not `jvmopts` right? Because the argument parsing seems to only accept `jvmflags`

gobblin-docker/gobblin-service/alpine-gaas-latest/entrypoint.sh
```diff
@@ -15,7 +15,23 @@
...
+
 GOBBLIN_HOME="$(cd `dirname $0`/..; pwd)"
+JVM_OPT_FLAG=""
+# Treat the config env args differently as they need to be combined together in 1 arg
+CONFIG_ARGS=""
+shopt -s nocasematch
+for i in "$@"
+do
+    case "$1" in
+        --args)
```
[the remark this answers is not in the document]
Will-Lo: I realized there was a way to remove all of these hacks, should've looked into more detailed bash...

bin/gobblin.sh
jhsenjaliya: actually thats the typo in the help text `--jvmopts`, and I meant to keep the jvmopts since that is more intutive, let me fix that since its some more variable changes and send PR, in meantime, you can assume to pass it as `jvmopts`  Thanks

gobblin-kubernetes/gobblin-service/dev-cluster/gaas-application.conf
```diff
@@ -0,0 +1,73 @@
...
+# DagManager
+gobblin.service.dagManager.enabled=true
+gobblin.service.dagManager.jobStatusRetriever.class="org.apache.gobblin.service.monitoring.FsJobStatusRetriever"
+gobblin.service.dagManager.dagStateStoreClass="org.apache.gobblin.service.modules.orchestration.FSDagStateStore"
+gobblin.service.dagManager.dagStateStoreDir=${gobblin.service.work.dir}/dagStateStoreDir
+
+# RestLI
+gobblin.service.port=6956
+
+# MySQL State Store
```
zxcware: if `dev` is built around file systems, do we need `mysql` environment here? A user can pick either `dev` suite or `mysql` suite.

gobblin-kubernetes/gobblin-service/dev-cluster/mysql-deployment.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: v1
```
zxcware: How do we organize the content between `fs` based and `mysql` based? Using the same parent folder `dev-cluster` is confusing. 

Currently, we're using `kustomization` file to share and organize k8s application configs. I guess we'll have a mysql folder with its own `kustomization` file?

gobblin-kubernetes/gobblin-service/dev-cluster/gaas-application.conf
Will-Lo: Ah I should rename it to be better then, `dev` was originally intended for `mysql` plus whatever job communication mechanism we are relying on, which is `fs`. The only difference between `dev` and `base` would be the mysql store, so I guess it would be better to rename it to mysql-cluster

PR #2782, opened 2019-10-24
[GOBBLIN-925] Create option to log outputs to console, fix docker-com…

Will-Lo:
…pose

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-925

### Description

Currently, Gobblin containers `tail -F` a file to output logs.
While this works, this means that the container will not stop execution if the java process is terminated. This becomes problematic when Kubernetes uses container status to restart pods.

This patch adds a flag variable to the startup script so that containers can read console directly by running the gobblin executable in the foreground 

### Tests
scripts change

### Commits

### DIFF
--- a/conf/standalone/log4j.xml
+++ b/conf/standalone/log4j.xml
@@ -16,6 +16,13 @@
     </layout>
   </appender>
 
+
+  <appender name="console" class="org.apache.log4j.ConsoleAppender">
+    <layout class="org.apache.log4j.PatternLayout">
+      <param name="ConversionPattern" value="%d{yyyy-MM-dd HH:mm:ss z} %-5p [%t] %C %X{tableName} %L - %m%n"/>
+    </layout>
+  </appender>
+
   <logger name="org.apache.commons.httpclient">
     <level value="DEBUG"/>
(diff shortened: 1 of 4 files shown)


conf/standalone/log4j.xml
```diff
@@ -3,7 +3,7 @@
 <log4j:configuration>
 
   <appender name="FileRoll" class="org.apache.log4j.rolling.RollingFileAppender">
-    <param name="file" value="${gobblin.logs.dir}/standalone.out" />
+    <param name="file" value="${gobblin.logs.dir}/gobblin-current.log" />
```
jhsenjaliya: file name here is mode name, any reason to change it to gobblin-current ?

jhsenjaliya:
LGTM, other than a file name comment.

conf/standalone/log4j.xml
Will-Lo: Ah good point I'll just change it back, I think I wanted it to match prior to the script change but I realize now there's no real point of breaking naming convention

PR #2792, opened 2019-11-01
[GOBBLIN-933] add support for array of unions in json schema

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @zxcware  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-933

### Description
It handles unions inside an array.

### Tests
enhanced a test case

### Commits


gobblin-core/src/test/resources/converter/jsonToAvroRecord.json
```diff
@@ -1,4 +1,6 @@
-{"nullableField": null,
+{
+  "arrayFieldWithUnion": ["arrU1", "arrU2", "arrU3"],
```
ahmedahamid: This would probably need a `null` value to make sure they're handled properly.

ahmedahamid: Just following up. I tried adding a `null` value to this array but the `JsonRecordAvroSchemaToAvroConverterTest.testConverter` failed:
```java
Could not convert field arrayFieldWithUnion
org.apache.gobblin.converter.DataConversionException: Could not convert field arrayFieldWithUnion
	at org.apache.gobblin.converter.avro.JsonRecordAvroSchemaToAvroConverter.convertNestedRecord(JsonRecordAvroSchemaToAvroConverter.java:127)
	at org.apache.gobblin.converter.avro.JsonRecordAvroSchemaToAvroConverter.convertRecord(JsonRecordAvroSchemaToAvroConverter.java:71)
	at org.apache.gobblin.converter.avro.JsonRecordAvroSchemaToAvroConverterTest.testConverter(JsonRecordAvroSchemaToAvroConverterTest.java:69)
...
Caused by: java.lang.RuntimeException: Field: arrayFieldWithUnion is not nullable and contains a null value
	at org.apache.gobblin.converter.avro.JsonElementConversionFactory$JsonElementConverter.convert(JsonElementConversionFactory.java:278)
	at org.apache.gobblin.converter.avro.JsonElementConversionWithAvroSchemaFactory$ArrayConverter.convertField(JsonElementConversionWithAvroSchemaFactory.java:98)
	at org.apache.gobblin.converter.avro.JsonElementConversionFactory$JsonElementConverter.convert(JsonElementConversionFactory.java:280)
	at org.apache.gobblin.converter.avro.JsonRecordAvroSchemaToAvroConverter.convertNestedRecord(JsonRecordAvroSchemaToAvroConverter.java:125)
	... 51 more
```

I think the key issue here is that `UnionConverter`'s does not determine its nullability correctly; it uses the nullability it receives from `JsonElementConversionWithAvroSchemaFactory.getConvertor()` to decide its own nullability. This is incorrect because `arrayFieldWithUnion` is a non-nullable array. However, the union itself (i.e. `UnionConverter`) would return `false` if you invoke `isNullable()` on it because it is construing `arrayFieldWithUnion`'s non-nullability as  its own (see the values `UnionConverter` provides to `super`).

One possible fix could be to have `UnionConverter` override `isNullable()` so it can decide its own nullability based on its schemas, e.g.

```java
    @Override
    public boolean isNullable() {
      return this.firstSchema.getType() == Schema.Type.NULL ||
          this.secondSchema.getType() == Schema.Type.NULL;
    }
```

autumnust: +1 

BTW we should try to stay aligned with Avro schema guide to always place null in the first to indicate nullable schema.

gobblin-core/src/test/java/org/apache/gobblin/converter/avro/JsonRecordAvroSchemaToAvroConverterTest.java
```diff
@@ -85,5 +85,17 @@ public void testConverter()
     Assert.assertEquals(recordInArray.get("field1").toString(), "test1");
 
     Assert.assertEquals((record.get("enumField")).toString(), "ENUM2");
+
+    Assert.assertTrue(record.get("arrayFieldWithUnion") instanceof GenericArray);
```
autumnust: Please add more cases for this converter which is pretty unit-testable.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +215,49 @@ public Schema schema() {
...
+    private final Schema secondSchema;
+    private final JsonElementConverter firstConverter;
+    private final JsonElementConverter secondConverter;
+
+    public UnionConverter(String fieldName, boolean nullable, String sourceType, Schema schemaNode,
+        WorkUnitState state, List<String> ignoreFields) throws UnsupportedDateTypeException {
+      super(fieldName, nullable, sourceType);
+      List<Schema> types = schemaNode.getTypes();
+      if (types.size() != 2) {
+        throw new RuntimeException("Unions must be size 2.");
```
autumnust: This is not a valid assumption in the context of Avro schema. What's the reason to enforce `size == 2` for union here?

arjun4084346: the existing code only supports two types.
https://github.com/apache/incubator-gobblin/blob/master/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonRecordAvroSchemaToAvroConverter.java#L93

PR #2798, opened 2019-11-07
[GOBBLIN-950] Avoid persisting dag right after loading it on startup

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-950

### Description

The dag only needs to be persisted when being added by `Orchestrator`, but right now it also is persisted when being loaded from store on GaaS restart (even though it won't have changed). This PR removes the unnecessary persist there.

### Tests
Tested locally

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -225,13 +225,17 @@ public class DagManager extends AbstractIdleService {
   }
 
   /**
-   * Method to submit a {@link Dag} to the {@link DagManager}. The {@link DagManager} first persists the
+   * Method to submit a {@link Dag} to the {@link DagManager}. The {@link DagManager} optionally persists the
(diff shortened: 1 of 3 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -243,6 +247,10 @@ synchronized void addDag(Dag<JobExecutionPlan> dag) throws IOException {
     submitEventsAndSetStatus(dag);
   }
 
+  synchronized void addDag(Dag<JobExecutionPlan> dag) throws IOException {
```
haojiliu: Is this for backwards compatibility? I'm curious why we need an extra function for implicit persistence.

jack-moseley: Mainly just to provide a default value since some unit tests call this, but yeah maybe it's better to update all the callers.

jack-moseley:
@sv2000 please merge

PR #2799, opened 2019-11-07
[GOBBLIN-932] Create deployment for Azure, clean up existing deployments

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-932

### Description

- Adds ingress specification for running GaaS K8s on Azure
- Adds documentation about running Gaas on K8s in different environments
- Refactors kustomization files so that they can leverage one another for build purposes

### Tests
Environment changes

### Commits


gobblin-kubernetes/gobblin-service/azure-cluster/ingress.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: extensions/v1beta1
```
zxcware: why do we need nginx? For load balancing? Let's add some comments.

gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
```diff
@@ -22,18 +22,17 @@ spec:
         - name: 'shared-jobs'
           persistentVolumeClaim:
             claimName: shared-jobs-claim
-        - name: 'shared-template-catalogs'
-          persistentVolumeClaim:
-            claimName: shared-template-catalogs-claim
+        - name: flowconfig-templates
```
zxcware: Are we introducing `shared-template-catalog` persistentVolumeClaim with configMap?

gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
```diff
@@ -61,19 +60,4 @@ spec:
           image: will97/gobblin-standalone:latest
           volumeMounts:
             - name: shared-jobs
-              mountPath: /tmp/gobblin-standalone/jobs
----
-apiVersion: v1
```
zxcware: We don't need it?

gobblin-kubernetes/gobblin-service/azure-cluster/ingress.yaml
Will-Lo: Ah we need nginx as the ingress, because Azure AKS doesn't have a built in ingress like minikube does. I'll add some comments.

gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
Will-Lo: I'm going to deprecate `shared-template-catalog` because the `flowconfig-templates` volume is applied for every pod, and it loads the templates from the `base` folder

gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
Will-Lo: Huh I'm not sure why git says I removed the line, I only removed the newline after it because it shows up on the other side of the diff. Probably the lack of a newline? The volume is still there.
I removed the service underneath it to move it to a new file to better separate the yamls

PR #2800, opened 2019-11-08
[GOBBLIN-933] add support for array of unions in json schema_new

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-933

### Description
It handles unions inside an array.
But we have a known bug here, if the type in union contains String and other types like int, etc, then it's possible that all the value even int value will be parse as String.

### Tests
Unit test

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
@@ -26,6 +26,7 @@ import java.util.Map;
 
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
+import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.converter.DataConversionException;
 
@@ -217,6 +218,13 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
     }
   }
 
+  /**
+   * A converter to convert Union type to avro
+   * Here it will try all the possible converters for one type, for example, to convert an int value, it will try all Number converters
+   * until meet the first workable one.
+   * So a known bug here is if there are long and double inside the union type, it's possible that all the value will be parse as long
+   * We're doing this since there is no way to determine what exact type it is for a JasonElement
+   */
   public static class UnionConverter extends ComplexConverter {
     private final List<Schema> schemas;
     private final List<JsonElementConverter> converters;
@@ -238,8 +246,55 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
        for(JsonElementConverter converter: converters)
        {
          try {
-           Object o = converter.convert(value);
-           return o;
+           switch (converter.getTargetType()) {
+             case STRING: {
+               if (value.isJsonPrimitive() && value.getAsJsonPrimitive().isString()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case FIXED:
+             case BYTES:
+             case INT:
+             case LONG:
+             case FLOAT:
+             case DOUBLE: {
+               if (value.isJsonPrimitive() && value.getAsJsonPrimitive().isNumber()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case BOOLEAN:{
+               if (value.isJsonPrimitive() && value.getAsJsonPrimitive().isBoolean()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case ARRAY:{
+               if (value.isJsonArray()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case MAP:
+             case ENUM:
+             case RECORD:{
+               if (value.isJsonObject()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case NULL:{
+               if(value.isJsonNull()) {
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case UNION:
+               return new UnsupportedDateTypeException("does not support union type in union");
+             default:
+               return converter.convert(value);
+           }
          } catch (Exception e){}
        }
        throw new RuntimeException(StringFormatter.format("Cannot convert %s to avro using schema %s", value.getAsString(), schemaNode.toString()).toString());
@@ -247,15 +302,7 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
 
     @Override
     public Schema.Type getTargetType() {
-      if(schemas.size() == 2 && isNullable())
-      {
-        if(schemas.get(0).getType() == Schema.Type.NULL) {
-          return schemas.get(1).getType();
-        } else {
-          return schemas.get(0).getType();
-        }
-      }
-      return Schema.Type.UNION;
+      return schema().getType();
     }
 
     @Override
--- a/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
@@ -307,8 +307,7 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
 
     @Override
     public Schema schema() {
-      if(schemas.size() == 2 && isNullable())
-      {
+      if(schemas.size() == 2 && isNullable()) {
         if(schemas.get(0).getType() == Schema.Type.NULL) {
           return schemas.get(1);
         } else {
--- a/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
@@ -17,7 +17,9 @@
 
 package org.apache.gobblin.converter.avro;
 
+import com.sun.javafx.binding.StringFormatter;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -68,6 +70,10 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
 
       case RECORD:
         return new JsonElementConversionWithAvroSchemaFactory.RecordConverter(fieldName, nullable, type.toString(),
+            schemaNode, state, ignoreFields);
+
+      case UNION:
+        return new JsonElementConversionWithAvroSchemaFactory.UnionConverter(fieldName, nullable, type.toString(),
             schemaNode, state, ignoreFields);
 
       default:
@@ -183,7 +189,7 @@ public class JsonElementConversionWithAvroSchemaFactory extends JsonElementConve
     WorkUnitState state;
 
     public RecordConverter(String fieldName, boolean nullable, String sourceType, Schema schemaNode,
-        WorkUnitState state, List<String> ignoreFields) throws UnsupportedDateTypeException {
+        WorkUnitState state, List<String> ignoreFields) {
       super(fieldName, nullable, sourceType);
       this.schema = schemaNode;
       this.state = state;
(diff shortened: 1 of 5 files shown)


gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
...
+      converters = new ArrayList<>();
+      for(Schema schema: schemas) {
+        converters.add(getConvertor(fieldName, schema.getType().getName(), schemaNode, state, isNullable(), ignoreFields));
+      }
+      this.schemaNode = schemaNode;
+    }
+
+    @Override
+    Object convertField(JsonElement value) {
+       for(JsonElementConverter converter: converters)
```
autumnust: I wasn't aware it is using Gson library for JSON element representation. You could use `value.isXXXType` to determine the type instead of hitting the first workable one.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
       return this.schema;
     }
   }
-}
+
+  public static class UnionConverter extends ComplexConverter {
```
autumnust: Add javadoc to state the assumption made.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
...
+    private final List<JsonElementConverter> converters;
+    private final Schema schemaNode;
+
+    public UnionConverter(String fieldName, boolean nullable, String sourceType, Schema schemaNode,
+        WorkUnitState state, List<String> ignoreFields) throws UnsupportedDateTypeException {
+      super(fieldName, nullable, sourceType);
+      this.schemas = schemaNode.getTypes();
+      converters = new ArrayList<>();
+      for(Schema schema: schemas) {
+        converters.add(getConvertor(fieldName, schema.getType().getName(), schemaNode, state, isNullable(), ignoreFields));
```
haojiliu: Is this a typo?

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
haojiliu: this logic doesn't seem quite intuitive to me, you are trying each converter and see which can convert the given value? The performance impact is probably negligible, just curious.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
...
+           Object o = converter.convert(value);
+           return o;
+         } catch (Exception e){}
+       }
+       throw new RuntimeException(StringFormatter.format("Cannot convert %s to avro using schema %s", value.getAsString(), schemaNode.toString()).toString());
+    }
+
+    @Override
+    public Schema.Type getTargetType() {
+      if(schemas.size() == 2 && isNullable())
```
haojiliu: you can extract line263-270 out and reuse it here. It looks like custom logic and therefore should comply with DRY

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
...
+      {
+        if(schemas.get(0).getType() == Schema.Type.NULL) {
+          return schemas.get(1).getType();
+        } else {
```
haojiliu: what if both are NULL?

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -183,7 +189,7 @@ public Schema schema() {
     WorkUnitState state;
 
     public RecordConverter(String fieldName, boolean nullable, String sourceType, Schema schemaNode,
-        WorkUnitState state, List<String> ignoreFields) throws UnsupportedDateTypeException {
```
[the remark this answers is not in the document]
ZihanLi58: No. Just because inside this method, nothing will throw that exception. Seems like useless code so I remove it

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Will address

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Sorry I didn't see the typo. Can you help me point it out? Thanks!

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Then it will return type NULL.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
arjun4084346: If you are going to add support for multiple types inside a `union` in this PR, you need to make changes https://github.com/apache/incubator-gobblin/blob/master/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonRecordAvroSchemaToAvroConverter.java#L93 also. This is the place where it tries to process a union when it is at the first level in a `Record`, i.e. not inside a map or array.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +216,70 @@ public Schema schema() {
...
+        } else {
+          return schemas.get(0).getType();
+        }
+      }
+      return Schema.Type.UNION;
+    }
+
+    @Override
+    public Schema schema() {
+      if(schemas.size() == 2 && isNullable())
```
arjun4084346: Sorry, I did not understand why this check is required.
Also, start braces should be in this line, not in the next line

ZihanLi58: This is a requirement for avro, if it's a union of null and XXX type, it's a nullable XXX type in avro.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: I see that. Since it's a oncall issue, I just want to unblock them first ^^, and we may add ticket for fixing this in the future.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Addressed

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Addressed

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
ZihanLi58: Addressed

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
arjun4084346: What should be the schema, if it is a union of null and XXX type and YYY type?

ZihanLi58: And then it's still a union

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +217,116 @@ public Schema schema() {
...
+                 return converter.convert(value);
+               }
+               break;
+             }
+             case UNION:
+               return new UnsupportedDateTypeException("does not support union type in union");
+             default:
+               return converter.convert(value);
+           }
+         } catch (Exception e){}
```
autumnust: Why empty ? If not sure what to, just propagate it.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +217,116 @@ public Schema schema() {
...
     }
   }
-}
+
+  /**
+   * A converter to convert Union type to avro
+   * Here it will try all the possible converters for one type, for example, to convert an int value, it will try all Number converters
+   * until meet the first workable one.
+   * So a known bug here is if there are long and double inside the union type, it's possible that all the value will be parse as long
+   * We're doing this since there is no way to determine what exact type it is for a JasonElement
```
autumnust: JasonElement -> JsonElement

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +217,116 @@ public Schema schema() {
       return this.schema;
     }
   }
-}
+
+  /**
+   * A converter to convert Union type to avro
+   * Here it will try all the possible converters for one type, for example, to convert an int value, it will try all Number converters
```
autumnust: Long story short: "there's no guarantee on preserving precision from Json to Avro type as the exact type information is clear from JsonElement"

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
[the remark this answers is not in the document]
haojiliu: are you sure super() won't throw such exception at all?

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
haojiliu: getConvertor() vs. getConverter(). You can ignore this, not your fault.

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
haojiliu: second this

ZihanLi58: Since this still can not determine what exact type it is like RECORD and MAP, so I try to rely on the exception to determine it. If there is an exception it will continue to try next possible converter instead of just throw out the exception

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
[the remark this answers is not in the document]
ZihanLi58: Yes, super() will not throw this exception. The other constructers throw out this exception is because getConvertor() method may throw this exception

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -210,4 +215,115 @@ public Schema schema() {
...
+   * until meet the first workable one.
+   * So a known bug here is there's no guarantee on preserving precision from Json to Avro type as the exact type information is clear from JsonElement
+   * We're doing this since there is no way to determine what exact type it is for a JsonElement
```
arjun4084346: +1

gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
arjun4084346: Ok

arjun4084346:
+1. Please create an internal tracking ticket to remove the discrepancy about the support of multiple types in union

PR #2803, opened 2019-11-09
[GOBBLIN-954] Added support to swap different HiveRegistrationPublishers

vikrambohra:
Dear Gobblin maintainers,

### JIRA
- [x ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-954] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-954

### Description
- [x ] Here are some details about my PR, including screenshots (if applicable):
1. Added logic to support swapping of different implementations of HiveRegistrationPublisher in HiveRegTaskStateCollectorServiceHandlerImpl
2. Added "data.publisher.latest.file.arrival.timestamp" as current time in millis whenever a task finishes

### Tests
- [ x] My PR adds the following unit tests __OR__ does not need testing for this extremely good reason:
Code already covered by test cases TaskStateCollectorServiceTest

### Commits
- [ x] My commits all reference JIRA issues in their subject lines, and I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)":


gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -874,6 +890,9 @@
    * Hive registration properties
    */
   public static final String HIVE_REGISTRATION_POLICY = "hive.registration.policy";
+  public static final String HIVE_REG_PUBLISHER_CLASS = "hive.reg.publisher.class";
```
autumnust: I actually believe we should stop adding configurations into this `ConfigurationKey` class but put all these configurations into most relevant class. The reason is if you put everything here and make them public static final, any constructs within Gobblin that carrying state can modify it without any restriction, and also you lost context on what does this configuration really mean.

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -290,6 +290,7 @@ public void close()
       for (Path path : this.publisherOutputDirs) {
         this.state.appendToSetProp(ConfigurationKeys.PUBLISHER_DIRS, path.toString());
       }
+      this.state.setProp(ConfigurationKeys.PUBLISHER_LATEST_FILE_ARRIVAL_TIMESTAMP, System.currentTimeMillis());
```
autumnust: I would suggest make `System.currentTimeMillis()` into a protected method as the default implementation(with a state object as parameter input) so that it can be easily extended in the future.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegister.java
```diff
@@ -329,22 +349,28 @@ protected boolean needToUpdatePartition(HivePartition existingPartition, HivePar
...
-        try {
-          entry.getValue().get();
-        } catch (ExecutionException ee) {
-          throw new IOException("Failed to finish registration for " + entry.getKey(), ee.getCause());
-        }
-      }
-    } catch (InterruptedException e) {
-      throw new IOException(e);
+      waitOnFuturesToFinish();
+    } catch (IOException e) {
```
autumnust: What's the purpose for this catch block if everything it does is to rethrow?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/HiveRegTaskStateCollectorServiceHandlerImpl.java
```diff
@@ -39,16 +41,25 @@
   private HiveRegistrationPublisher hiveRegHandler;
 
   public HiveRegTaskStateCollectorServiceHandlerImpl(JobState jobState) {
-    hiveRegHandler = new HiveRegistrationPublisher(jobState);
+    String className = jobState
+        .getProp(ConfigurationKeys.HIVE_REG_PUBLISHER_CLASS, ConfigurationKeys.DEFAULT_HIVE_REG_PUBLISHER_CLASS);
+    try {
+      hiveRegHandler =
+          (HiveRegistrationPublisher) Class.forName(className).getConstructor(State.class).newInstance(jobState);
```
autumnust: use `GobblinConstructorUtils` instead.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/HiveRegTaskStateCollectorServiceHandlerImpl.java
```diff
@@ -39,16 +41,25 @@
...
+    } catch (Exception e) {
```
autumnust: What's the specific type of exception here?

autumnust:
Please check Travis error also.

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
vikrambohra: Requirement changed to add a flag called update_operation=new_files_added. should that be added here or in gobblin-kafka?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
vikrambohra: agreed

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/HiveRegTaskStateCollectorServiceHandlerImpl.java
```diff
@@ -38,17 +37,19 @@
 @Slf4j
 public class HiveRegTaskStateCollectorServiceHandlerImpl implements TaskStateCollectorServiceHandler {
 
+  private static final String HIVE_REG_PUBLISHER_CLASS = "hive.reg.publisher.class";
```
autumnust: This naming is ambiguous, better to equip it with some information about task-collection-service

PR #2809, opened 2019-11-13
[GOBBLIN-960]Resolving multiple templates in top-level 

autumnust:
Dear Gobblin maintainers,

### JIRA
-  https://issues.apache.org/jira/browse/GOBBLIN-960

### Description
It contains some fixes on access-modifier of Avro2ORC while reading through code, but the major part is to enabled `gobblin.templates.uri` to be able to accept multiple templates.

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
@@ -75,6 +75,16 @@ public abstract class InheritingJobTemplate implements JobTemplate {
     resolveTemplates(loadedTemplates);
   }
 
+  /**
+   * Resolve all superTemplates being field variables within the class.
+   * There are two types of resolution being involved in this method:
+   * 1) When all templates are being represented as {@link #superTemplateUris}, the actual template will be loaded from
+   * catalog first and enter resolution process. The physical {@link #superTemplates} are being materialized after that.
+   * 2) org.apache.gobblin.runtime.template.InheritingJobTemplate#InheritingJobTemplate(java.util.List, boolean) provides
+   * interface to directly provide physical {@link #superTemplates}. This case is determined by non-null containers of
+   * {@link #superTemplates}.
+   *
+   */
   private void resolveTemplates(Map<URI, JobTemplate> loadedTemplates) throws SpecNotFoundException, TemplateException {
     if (this.resolved) {
       return;
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
@@ -51,11 +51,15 @@ public abstract class InheritingJobTemplate implements JobTemplate {
     this.resolved = false;
   }
 
-  public InheritingJobTemplate(List<JobTemplate> superTemplates) {
+  public InheritingJobTemplate(List<JobTemplate> superTemplates, boolean skipResolve) {
     this.superTemplateUris = Lists.newArrayList();
     this.catalog = null;
     this.superTemplates = superTemplates;
-    this.resolved = true;
+    this.resolved = skipResolve;
+  }
+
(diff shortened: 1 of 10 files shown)


autumnust:
@jack-moseley  @sv2000  Can you take a look once you have time? Thanks

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
```diff
@@ -75,23 +79,48 @@ private void resolveTemplates(Map<URI, JobTemplate> loadedTemplates) throws Spec
...
     }
-    this.superTemplates = Lists.newArrayList();
-    for (URI uri : this.superTemplateUris) {
-      if (!loadedTemplates.containsKey(uri)) {
-        JobTemplate newTemplate = this.catalog.getTemplate(uri);
-        loadedTemplates.put(uri, newTemplate);
-        if (newTemplate instanceof InheritingJobTemplate) {
-          ((InheritingJobTemplate) newTemplate).resolveTemplates(loadedTemplates);
+
+    if (this.superTemplateUris.size() > 0) {
```
jack-moseley: Maybe add some javadoc for `resolveTemplates` method since this logic is a bit confusing

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
```diff
@@ -103,13 +132,14 @@ public Config getRawTemplateConfig() throws SpecNotFoundException, TemplateExcep
...
       throws SpecNotFoundException, TemplateException {
     Config rawTemplate = getLocalRawTemplate();
-    for (JobTemplate template : Lists.reverse(this.superTemplates)) {
-      if (!alreadyInheritedTemplates.contains(template)) {
-        alreadyInheritedTemplates.add(template);
-        Config thisFallback = template instanceof InheritingJobTemplate
-            ? ((InheritingJobTemplate) template).getRawTemplateConfigHelper(alreadyInheritedTemplates)
-            : template.getRawTemplateConfig();
-        rawTemplate = rawTemplate.withFallback(thisFallback);
+    if (this.superTemplates != null) {
```
jack-moseley: Why did all these `superTemplates` null checks have to be added? What is the case where it will be null?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/StaticJobTemplate.java
```diff
@@ -64,13 +64,23 @@
   private Collection<String> dependencies;
 
   public StaticJobTemplate(URI uri, String version, String description, Config config, JobCatalogWithTemplates catalog)
-      throws SpecNotFoundException, TemplateException {
+      throws TemplateException {
     this(uri, version, description, config, getSuperTemplateUris(config), catalog);
   }
 
+  /** An constructor that materialize multiple templates into a single static template
+   * The constructor provided multiple existing templates as the input instead of templateURIs
```
jack-moseley: `multiple in-memory templates` instead of `multiple existing templates`?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
autumnust: Added

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/InheritingJobTemplate.java
autumnust: It needs to be added because this change essentially gives resolution process an interface to provide loaded templates and ask for resolution, instead of providing template URI and ask for resolution as it was. This is used for determining if the physical templates are provided.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/template/StaticJobTemplate.java
autumnust: Addressed.

PR #2810, opened 2019-11-13
[GOBBLIN-961]Bypass locked directories when calculating src watermark

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-961

### Description
Background:
 Found the issue where there are locking directory in place, which gobblin failed to scan through list of files (source) to copy, and the watermark retrieve as NAN.
Therefore, data is not able to copy successfully and possibly causing the delay
Solution:
Simple change to bypass locked partitions when calculating src watermark

### Tests
Test on azkaban to make sure when there is locked partition, it can bypass that partition and do copy

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/replication/SourceHadoopFsEndPoint.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/replication/SourceHadoopFsEndPoint.java
@@ -17,7 +17,6 @@
 
 package org.apache.gobblin.data.management.copy.replication;
 
-import com.sun.javafx.binding.StringFormatter;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/replication/SourceHadoopFsEndPoint.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.data.management.copy.replication;
 
+import com.sun.javafx.binding.StringFormatter;
```
autumnust: Is this dependency intentional ?

ZihanLi58: Yes, used to formate the error message

PR #2818, opened 2019-11-19
[GOBBLIN-963] Remove duplicated copies of TaskContext/TaskState when constructing TaskIFaceWrapper

enjoyear:
Dear Gobblin maintainers,

### JIRA
- [X] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-963] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-963

### Description
Currently in GobblinMultiTaskAttempt, when we call createTaskRunnable, we rely on a TaskFactory to create a task and return a wrapper for that task. However, when we create the wrapper, we initialized the TaskContext twice. 

For the task creation
For the task wrapper creation
These two TaskContexts, which generate two copies of TaskState, will cause inconsistency in the following operation. The task state update within the task won't be reflected in TaskWrapper.

 

Specifically, this is the buggy line of code
```java
return new TaskIFaceWrapper(taskFactoryOpt.get().createTask(new TaskContext(workUnitState)),
 new TaskContext(workUnitState), countDownLatch, this.taskStateTracker);
```
Ideally, we should create the TaskContext once, and reuse the same TaskContext.

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
@@ -38,6 +38,9 @@ import com.google.common.base.Optional;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 
+import javax.annotation.Nullable;
+import lombok.Setter;
+
 import org.apache.gobblin.annotation.Alpha;
 import org.apache.gobblin.broker.gobblin_scopes.GobblinScopeTypes;
 import org.apache.gobblin.broker.gobblin_scopes.TaskScopeInstance;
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -78,8 +78,7 @@
     /**
      * Commit {@link GobblinMultiTaskAttempt} immediately after running is done.
      */
-    IMMEDIATE,
-    /**
+    IMMEDIATE, /**
```
sv2000: Is this intentional? Shouldn't the "/**" start on a new line?

enjoyear: I guess it's enforced based on the GobblinCode style done through "Reformat Code" action in  IntelliJ

enjoyear: Just realized that adding one more empty line will fix it. Will make the change.

PR #2819, opened 2019-11-19
[GOBBLIN-971]Enable speculative execution awareness for RowQualityChecker

autumnust:
Dear Gobblin maintainers,

### JIRA
- [x] https://issues.apache.org/jira/browse/GOBBLIN-971

### Description
- [x] 
Problem: 
When speculative execution is enabled with ERR_FILE set as the policy when encountering records that failed rowPolicy, we will run into HDFS file lease contention since the file name are only depends on `state.getId()`. 

Solution: To maintain backward-compatibility we allow speculative-execution to happen by default, but whenever the policy is set to `ERR_FILE` , we created different file name for speculative execution. This won't guarantee contention-free as different execution environment could coincidentally give identical timestamp, but the chances are pretty low. 

### Tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.apache.gobblin.commit.SpeculativeAttemptAwareConstruct;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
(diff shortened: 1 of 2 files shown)


autumnust:
@ZihanLi58 Please take a look, thanks.

gobblin-core/src/main/java/org/apache/gobblin/qualitychecker/row/RowLevelPolicyChecker.java
```diff
@@ -45,7 +46,17 @@
...
+public class RowLevelPolicyChecker<S, D> implements Closeable, FinalState, RecordStreamProcessor<S, S, D, D>,
+                                                    SpeculativeAttemptAwareConstruct {
+
+  /**
+   * Given the existence of writer object when the policy is set to {@link RowLevelPolicy.Type#ERR_FILE}, objects of
+   * this class needs to be speculative-attempt-aware.
+   */
+  @Override
+  public boolean isSpeculativeAttemptSafe() {
+    return !this.list.stream().allMatch(x -> x.getType().equals(RowLevelPolicy.Type.ERR_FILE)) || this.allowSpeculativeExecWhenWriteErrFile;
```
ZihanLi58: Why this one is allMatch?

autumnust: oops, bad mistake, thanks for catching

PR #2820, opened 2019-11-20
[GOBBLIN-974] Avoid updating job/flow status if messages arrive out of order

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-974

### Description

Define an order that `ExecutionStatus` are allowed to be updated for a job/flow so that if messages arrive out of order the job/flow status is not set to an incorrect value.

### Tests
Updated unit test

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
@@ -27,6 +27,7 @@ import java.util.concurrent.TimeUnit;
 import com.google.common.base.Charsets;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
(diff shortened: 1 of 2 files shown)


jack-moseley:
@sv2000 please review

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
```diff
@@ -146,6 +151,19 @@ static void addJobStatusToStateStore(org.apache.gobblin.configuration.State jobS
     String storeName = jobStatusStoreName(flowGroup, flowName);
     String tableName = jobStatusTableName(flowExecutionId, jobGroup, jobName);
 
+    List<org.apache.gobblin.configuration.State> states = stateStore.getAll(storeName, tableName);
```
sv2000: Can we avoid duplicate stateStore.getAll() call, as we are doing one in mergedProperties() as well?

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
```diff
@@ -146,6 +151,19 @@ static void addJobStatusToStateStore(org.apache.gobblin.configuration.State jobS
...
+    if (states.size() > 0) {
+      String previousStatus = states.get(states.size() - 1).getProp(JobStatusRetriever.EVENT_NAME_FIELD);
+      String currentStatus = jobStatus.getProp(JobStatusRetriever.EVENT_NAME_FIELD);
```
sv2000: Not sure if getProp() returns null in case of missing key or throws an Exception?

jack-moseley: It returns null.

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
jack-moseley: Good point, addressed.

PR #2823, opened 2019-11-22
GOBBLIN-976: Add dynamic config to the state before instantiating met…

sv2000:
…rics reporter in MRJobLauncher

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-976

### Description
To properly instantiate metric reporter, the dynamic config generated via dynamicConfigGenerator.classes needs to be added into the state object passed to the reporter. 

### Tests
Tested with an end-to-end pipeline.

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
@@ -27,6 +27,41 @@ import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocalFileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.io.Text;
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -766,6 +767,7 @@ protected void setup(Context context) {
       for (Map.Entry<String, ConfigValue> entry : dynamicConfig.entrySet()) {
         this.jobState.setProp(entry.getKey(), entry.getValue().unwrapped().toString());
         configuration.set(entry.getKey(), entry.getValue().unwrapped().toString());
+        gobblinJobState.setProp(entry.getKey(), entry.getValue().unwrapped().toString());
```
autumnust: Is there anyway to only keep `gobblinJobState` here, instead of have `configuration` object and `gobblinJobState`  ?

sv2000: I did consider it. I think this needs a significant clean up. In addition to configuration and gobblinJobState, we also have jobProps (which is a java.util.Properties). It seems that there is overlap in configs across these types, and it is unclear which property should go where. Arguably, since Hadoop APIs use Configuration, we should set all properties in the configuration object. But I am open to suggestions. I will create a tech-debt ticket and use that for addressing this.

autumnust: +1 Sounds good.

PR #2827, opened 2019-11-26
[COBBLIN-981]Handle backward compatibility issue in HiveSource

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-981

### Description
Handle backward compatibility issue in HiveSource

### Tests
Local test

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
@@ -222,7 +222,10 @@ public class HiveSource implements Source {
     silenceHiveLoggers();
   }
 
-
+  @Deprecated
+  protected void createWorkunitForNonPartitionedTable(HiveDataset hiveDataset) throws IOException {
+    this.createWorkunitForNonPartitionedTable(hiveDataset, false);
+  }
   protected void createWorkunitForNonPartitionedTable(HiveDataset hiveDataset, boolean disableAvroCheck) throws IOException {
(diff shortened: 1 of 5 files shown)


ZihanLi58:
@sv2000 Can you help take a look at this change? Thanks!

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/source/HiveSource.java
```diff
@@ -276,6 +279,11 @@ protected void createWorkunitForNonPartitionedTable(HiveDataset hiveDataset, boo
     }
   }
 
+  @Deprecated
+  protected HiveWorkUnit workUnitForTable(HiveDataset hiveDataset) throws IOException {
```
sv2000: Can we just keep the older APIs and get rid of the methods where we pass disableAvroCheck as an argument? Instead, we can just access this property via the source state object.

ZihanLi58: I did not do this because state is not a public variable. And this change dose not do this is because some of the downstream user has already change their code to enable the new API. And I also want to make this backward compatible with the latest version

sv2000: Do you mean to say that state is not a member variable? If so, we can create one. But I agree that reverting the newer APIs will be backward incompatible. Approving the RB.

PR #162, opened 2015-06-09
Added a StateStoreCleaner for cleaning up old state store files based on some retention

liyinan926:
Signed-off-by: Yinan Li <liyinan926@gmail.com>

PR #2833, opened 2019-11-28
[GOBBLIN-987] Reject unrecognized Enum symbols in JsonRecordAvroSchemaToAvroConverter

ahmedahamid:
This change fixes JsonRecordAvroSchemaToAvroConverter by rejecting Enum
symbols not defined in the specified Avro schema. Allowing such values
causes runtime errors when the resulting objects are encoded.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-987

### Description
`JsonRecordAvroSchemaToAvroConverter` allows unrecognized Enum symbols not declared in the specified Avro schema. Attempting to write/encode the resulting objects cause runtime errors that are difficult to diagnose or root-cause.

### Tests
`org.apache.gobblin.converter.avro.JsonRecordAvroSchemaToAvroConverterTest.testConverterThrowsOnUnrecognizedEnumSymbols()`

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.converter.avro;
 
+import com.google.common.base.Preconditions;
 import com.sun.javafx.binding.StringFormatter;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -25,7 +26,6 @@ import java.util.Map;
 
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
-import org.apache.commons.lang3.Validate;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.converter.DataConversionException;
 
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionWithAvroSchemaFactory.java
```diff
@@ -165,7 +166,10 @@ public EnumConverter(String fieldName, boolean nullable, String sourceType, Sche
 
     @Override
     Object convertField(JsonElement value) {
-      return new GenericData.EnumSymbol(this.schema, value.getAsString());
+      String valueString = value.getAsString();
+      Validate.isTrue(this.enumSet.contains(valueString),
```
zxcware: most of our check is done with guava `Preconditions`. Replace `Validate.isTrue` with `Preconditions` to be consistent and avoid introducing a new dependency?

ahmedahamid: No problem. The code already has a dependency on commons lang3 (I checked the `build.gradle` of the `gobblin-core` module) but I do agree regarding consistency.

One more thing: I had to make a tiny change to line #162 in `JsonElementConversionWithAvroSchemaFactory.java` because the code wasn't picking the enum name correctly. It was retrieving the type name instead, which is always `enum`. So, now the exception message reads as follows
```
PURPLE is not one of the valid symbols for the org.apache.gobblin.test.Colors enum: [RED, GREEN, BLUE]
```

PR #2851, opened 2019-12-17
[GOBBLIN-1006] Enable configurable case-preserving and schema source-of-truth in table level properties

autumnust:
Dear Gobblin maintainers,

Two changes: 
- By setting `<PREFIX>.destination.tableProperties.schema.sourceOfTruth,avro.schema.literal` for example, we could have conversion output table have the ultimate Avro-type schema for offline-system to leverage, since ORC schema tends to be lossy. 
- Make the case-preservation configurable since it is not applicable for all use-cases and we will need to maintain backward-compatibility for legacy datasets. 

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1006

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
@@ -100,7 +100,7 @@ public class StageableTableMetadata {
 
   /**
    * Attributes like "avro.schema.literal" are usually used in offline system as the source-of-truth of schema.
-   * This configuration's value should the key name that users expects to preserve to schema string if necessary.
+   * This configuration's value should be the key name that users expects to preserve to schema string if necessary.
    */
   public static final String SCHEMA_SOURCE_OF_TRUTH = "schema.sourceOfTruth";
 
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
@@ -53,6 +53,12 @@ public class StageableTableMetadata {
   public static final String ROW_LIMIT_KEY = "rowLimit";
   public static final String HIVE_VERSION_KEY = "hiveVersion";
   public static final String HIVE_RUNTIME_PROPERTIES_LIST_KEY = "hiveRuntimeProperties";
+
+  /**
+   * The output of Hive-based conversion can preserve casing in ORC-writer if "columns" and "columns.types" are being set
+   * in table-level properties. Turning this off by-default as Hive engine itself doesn't preserve the case.
+   */
+  public static final String OUTPUT_FILE_CASE_PRESERVED = "casePreserved";
   /***
    * Comma separated list of string that should be used as a prefix for destination partition directory name
    * ... (if present in the location path string of source partition)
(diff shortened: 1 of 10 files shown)


autumnust:
@ZihanLi58  Please take a look, thanks.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
```diff
@@ -92,6 +98,12 @@
    */
   public static final String SOURCE_DATA_PATH_IDENTIFIER_KEY = "source.dataPathIdentifier";
 
+  /**
+   * Attributes like "avro.schema.literal" are usually used in offline system as the source-of-truth of schema.
+   * This configuration's value should the key name that users expects to preserve to schema string if necessary.
```
ZihanLi58: "should the key name" -> "should be the key name" ?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/query/HiveAvroORCQueryGenerator.java
```diff
@@ -160,21 +163,25 @@ public static String generateCreateTableDDL(Schema schema,
     //    (evolution does not matter if its new destination table)
     // 4. If evolution is disabled, and destination table does exists
     //    .. use columns from destination schema
+    // Make sure the schema attribute will be updated in source-of-truth attribute.
+    // Or fall back to default attribute-pair used in Hive for ORC format.
+    if (tableProperties.containsKey(SCHEMA_SOURCE_OF_TRUTH)) {
+      tableProperties.setProperty(tableProperties.getProperty(SCHEMA_SOURCE_OF_TRUTH), schema.toString());
+      tableProperties.remove(SCHEMA_SOURCE_OF_TRUTH);
```
ZihanLi58: Just curious about what is this for?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
autumnust: Fixed.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/query/HiveAvroORCQueryGenerator.java
autumnust: So, usually this should be `avro.schema.literal`, which is what we did internally for most of tables. 

While there are cases where we cannot set `avro.schema.literal`: When there were older files  written in lower-case ORC but newly added `avro.schema.literal` is case-preserved.  Any mismatch between `avro.schema.literal` and file-level schema will cause problem, if `avro.schema.literal` is being set which gives offline system like Dali an illusion that this is the correct schema. 
But we still need something to inform pipelines like "view-creation" the Avro schema of a table, since interfering it from Hive schema is lossy.  When this case happens, we will set this `SCHEMA_SOURCE_OF_TRUTH ` to be something else (different from `avro.schema.literal`) as a workaround to provide Avro schema but not messing with Dali.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/entities/StageableTableMetadata.java
```diff
@@ -92,6 +98,12 @@
    */
   public static final String SOURCE_DATA_PATH_IDENTIFIER_KEY = "source.dataPathIdentifier";
 
+  /**
+   * Attributes like "avro.schema.literal" are usually used in offline system as the source-of-truth of schema.
+   * This configuration's value should be the key name that users expects to preserve to schema string if necessary.
+   */
+  public static final String SCHEMA_SOURCE_OF_TRUTH = "schema.sourceOfTruth";
```
sv2000: Call it "schema.original"? schema.sourceOfTruth refers to the source where you obtained the schema from, which does not look like how you are setting this property.

sv2000:
@autumnust can you check the travis failure?

autumnust:
> @autumnust can you check the travis failure?

Seems to be irrelevant to us according to: https://github.com/travis-ci/travis-ci/issues/7987. I will trigger another build

PR #2868, opened 2020-01-14
GOBBLIN-1025: Add retry for PK-Chuking iterator

arekusuri:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1025

### Description
In SFDC connector, there is a class called `ResultIterator` (refactored to BulkResultIterator and ResultChainingIterator).
It was using by only PK-Chunking currently. It encapsulated fetching a list of result files to a record iterator.

However, the csvReader.nextRecord() may throw out network IO exception. We should do retry in this case.

When a result file is fetched partly and one network IO exception happens, we are in a special situation - first half of the file is already fetched to our local, but another half of the file is still on datasource. 
We need to
1. reopen the file stream
2. skip all the records that we already fetched, seek the cursor to the record which we haven't fetched yet.
Also refactor normal bulk batch fetching result file to use the Iterators.

### Tests
pk-chunking with kraken test:
https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=22265515&job=salesforce_task_full&attempt=0
bulk batch with gaap-proxy test (map/reduce)
https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=22265502&job=Salesforce_Email_Logger_full&attempt=0
bulk batch with gaap-proxy test (azkaban local)
https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=22264750&job=Salesforce_Email_Logger_full&attempt=0


gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
```diff
@@ -26,12 +26,12 @@ private SalesforceConfigurationKeys() {
   public static final String BULK_API_USE_QUERY_ALL = "salesforce.bulkApiUseQueryAll";
 
   // pk-chunking
-  public static final String PK_CHUNKING_TEST_BATCH_ID_LIST = "salesforce.pkChunking.testBatchIdList";
-  public static final String PK_CHUNKING_TEST_JOB_ID = "salesforce.pkChunking.testJobId";
+  public static final String BULK_TEST_JOB_ID = "salesforce.bulk.test.job.id";
```
htran1: Please follow the config naming convention of using '.' as namespace separators, so this should be "salesforce.bulk.testJobId". Change the other new keys to conform.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceRecordIterator.java
```diff
@@ -99,7 +103,40 @@ private void fulfillCurrentRecord() {
     }
   }
 
-  private void resetCurrentRecordStatus() {
+  private InputStreamCSVReader reopenStreamWithRetry(ResultStruct resultStruct, int seekLineNumber, int retryNumber, int count) {
```
htran1: Can you refactor the logic in `SalesforceExtractor.reinitializeBufferedReader` and use it here? That has an additional validation to make sure that after skipping we arrived at the expected location based on the last record read before the failure.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
arekusuri: Thanks! Changed to 
```
public static final String BULK_TEST_JOB_ID = "salesforce.bulk.testJobId";
public static final String BULK_TEST_BATCH_ID_LIST = "salesforce.bulk.testBatchIds";
```

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceRecordIterator.java
arekusuri: Thanks! I did a refactor to make both pk-chunking and normal bulk batch using Iterator.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
```diff
@@ -563,92 +547,95 @@ public String getTimestampPredicateCondition(String column, long value, String v
...
+    log.info("----Get records for bulk batch job----");
     try {
-      if (!bulkApiLogin()) {
-        throw new IllegalArgumentException("Invalid Login");
-      }
+      // set finish status to false before starting the bulk job
+      this.setBulkJobFinished(false);
+      this.bulkResultIdList = getQueryResultIds(entity, predicateList);
+      log.info("Number of bulk api resultSet Ids:" + this.bulkResultIdList.size());
+      List<FileIdVO> fileIdVoList = this.bulkResultIdList.stream()
```
htran1: Is the `o` in fileIdVoList intentional or is it supposed to be a zero?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -0,0 +1,115 @@
...
+    closeCsvReader();
+    try {
+      InputStream is = conn.getQueryResultStream(jobId, batchId, resultId);
+      BufferedReader br = new BufferedReader(new InputStreamReader(is, ConfigurationKeys.DEFAULT_CHARSET_ENCODING));
+      csvReader = new InputStreamCSVReader(br);
+      for (int j = 0; j < lineCount; j++) {
+        csvReader.nextRecord(); // skip these record
+      }
+      // failed to skip seekLineNumber lines, try to open the file again. same time count+1
+      return csvReader;
```
htran1: I don't see the sanity check that after skipping the position matches the prior position. This was the old logic.

      for (int i = 0; i < recordsToSkip; i++) {	
        lastCsvRecord = reader.nextRecord();	
      }	

      // make sure the last record processed before the error was the last record skipped so that the next	
      // unprocessed record is processed in the next call to fetchResultBatch()	
      if (recordsToSkip > 0) {	
        if (!this.csvRecord.equals(lastCsvRecord)) {	
          throw new RuntimeException("Repositioning after reconnecting did not point to the expected record");	
        }	
      }	
    }

arekusuri: Added it back. test - https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=22307199&job=salesforce_task_full&attempt=0

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceExtractor.java
arekusuri: FileIdVO: I was trying to say "Value Object". it is a plain object for data transfer.
fileIdVoList: since fileIdVOList doesn't look smart, I made the O lower case.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
```diff
@@ -0,0 +1,92 @@
...
+    if (!queryResultIter.hasNext()) {
+      // no more data, print out total
+      log.info("Soft delete records total:{}", recordCount);
+    }
+    return queryResultIter.hasNext();
+  }
+
+  private void initQueryResultIter() {
+    try {
+      log.info("Pull soft delete records");
```
htran1: Is this the right log message?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public JsonElement next() {
+    if (queryResultIter == null) {
+      initQueryResultIter();
+    }
+    recordCount ++;
+    if (!queryResultIter.hasNext()) {
+      // no more data, print out total
+      log.info("Soft delete records total:{}", recordCount);
```
htran1: Same here. How come this class mentions soft delete?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.gobblin.source.extractor.DataRecordException;
+import org.apache.gobblin.source.extractor.watermark.Predicate;
+import org.apache.gobblin.source.workunit.WorkUnit;
+
+
+/**
+ * Iterator for rest api query
+ */
+@Slf4j
+public class QueryResultIterator implements Iterator<JsonElement> {
```
htran1: The name of the class and comments seem to indicate this is for fetching rest api query results in general, but the code below has logging specific to fetching "soft delete records".  If this class is specific to soft delete records, then that should be specified in the class name and comment.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultChainingIterator.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.common.collect.Iterators;
+import com.google.gson.JsonElement;
+import com.sforce.async.BulkConnection;
+import java.util.Iterator;
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+
+
+/**
+ * Result Iterator.
```
htran1: Please have a less generic comment.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
arekusuri: good catch! Fixed it. It is general.
In our case we only use it for soft delete.
QueryResultIterator is a iterate wrap for function RestApiExtractor.getRecordSet()
getRecordSet was called 2 places:
1. soft delete
2. QueryBasedExtractor.getIterator function. (using in simple mode)

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
arekusuri: fixed. thanks

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/QueryResultIterator.java
arekusuri: I changed the java doc

* Iterator for rest api query
 * It is a wrapper of
 * RestApiExtractor.getRecordSet(schema, entity, workUnit, predicateList)
 * the reason why we want to a wrapper for the function is -
 * We want to delay the execution of the function. Only when the next get called, we fetch the data.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/ResultChainingIterator.java
arekusuri: fixed, thanks!

>  * The Iterator to chain all result iterators together.
>  * It is to create only one iterator for a list of result files of BulkAPI.
>  * Same time it can also be able to add other iterator with function `add` to combine to 1 iterator

PR #2872, opened 2020-01-16
Refactor AbstractYarnSecurityManager to expose method for sending token file update message

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1030

### Description
Current implementation has this method in YarnAppSecurityManagerWithKeytabs. The refactoring will allow other implementations to use this method.

### Tests
Existing unit tests.

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/AbstractYarnAppSecurityManager.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/AbstractYarnAppSecurityManager.java
@@ -45,6 +45,7 @@ import com.google.common.util.concurrent.AbstractIdleService;
 import com.typesafe.config.Config;
 
 import org.apache.gobblin.cluster.GobblinClusterConfigurationKeys;
+import org.apache.gobblin.cluster.GobblinClusterManager;
 import org.apache.gobblin.cluster.GobblinHelixMessagingService;
 import org.apache.gobblin.util.ConfigUtils;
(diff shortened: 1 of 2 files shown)


gobblin-yarn/src/main/java/org/apache/gobblin/yarn/AbstractYarnAppSecurityManager.java
```diff
@@ -166,6 +175,15 @@ public void loginAndScheduleTokenRenewal() {
     }
   }
 
+  protected void sendTokenFileUpdatedMessage() {
+    // Send a message to the controller (when the cluster is not managed)
+    // and all the participants if this is not the first login
+    if (!this.isHelixClusterManaged) {
```
ZihanLi58: Why do we only send token file update message to controller when it's not managed? Will helix maintain the token refresh for us?

sv2000: In the managed mode, the GobblinClusterManager joins as a PARTICIPANT and hence, will get the token file updated message and handle the refresh. In the managed mode, sending this message to the controller will only queue messages on the CONTROLLER, since the CONTROLLER will likely not have a handler. In the unmanaged cluster mode, we maintain the same behavior as before, sending the updated message to both CONTROLLER and PARTICIPANT.

PR #2876, opened 2020-01-28
GOBBLIN-1034: Ensure underlying writers are expired from the Partitio…

sv2000:
…nedDataWriter cache to avoid accumulation of writers for long running Gobblin jobs

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1034

### Description
Currently, the underlying writers are never evicted from the PartitionedDataWriter cache. For long running Gobblin jobs (e.g. streaming), this will cause a memory leak particularly if the underlying writers maintain state. 

### Tests
Added unit test.

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
@@ -76,7 +76,7 @@ public class PartitionedDataWriter<S, D> extends WriterWrapper<D> implements Fin
   // NOTE: this config must be set only in streaming mode. For batch mode, setting this config will result
   // in incorrect behavior.
   public static final String PARTITIONED_WRITER_CACHE_TTL_SECONDS = "partitionedDataWriter.cache.ttl.seconds";
-  public static final Long DEFAULT_PARTITIONED_WRITER_CACHE_TTL_SECONDS = 3600L;
+  public static final Long DEFAULT_PARTITIONED_WRITER_CACHE_TTL_SECONDS = Long.MAX_VALUE;
 
   private static final GenericRecord NON_PARTITIONED_WRITER_KEY =
       new GenericData.Record(SchemaBuilder.record("Dummy").fields().endRecord());
@@ -136,6 +136,8 @@ public class PartitionedDataWriter<S, D> extends WriterWrapper<D> implements Fin
               writer.close();
             } catch (IOException e) {
               log.error("Exception {} encountered when closing data writer on cache eviction", e);
+              //Should propagate the exception to avoid committing/publishing corrupt files.
+              throw new RuntimeException(e);
             }
           }
         }
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -99,13 +118,32 @@ public PartitionedDataWriter(DataWriterBuilder<S, D> builder, final State state)
     if(builder.schema != null) {
       this.state.setProp(WRITER_LATEST_SCHEMA, builder.getSchema());
     }
-    this.partitionWriters = CacheBuilder.newBuilder().build(new CacheLoader<GenericRecord, DataWriter<D>>() {
+    Long cacheExpiryInterval = this.state.getPropAsLong(PARTITIONED_WRITER_CACHE_TTL_SECONDS, DEFAULT_PARTITIONED_WRITER_CACHE_TTL_SECONDS);
```
htran1: Can be `long` instead of `Long`.

gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -99,13 +118,32 @@ public PartitionedDataWriter(DataWriterBuilder<S, D> builder, final State state)
...
+      public void onRemoval(RemovalNotification<GenericRecord, DataWriter<D>> notification) {
+        synchronized (PartitionedDataWriter.this) {
+          if (notification.getValue() != null) {
+            try {
+              DataWriter<D> writer = notification.getValue();
+              totalRecordsFromEvictedWriters += writer.recordsWritten();
+              totalBytesFromEvictedWriters += writer.bytesWritten();
+              writer.close();
+            } catch (IOException e) {
+              log.error("Exception {} encountered when closing data writer on cache eviction", e);
```
htran1: Does failure need to be propagated to fail the task, like if a close failed?

sv2000: Hmm. I was not sure if we should make close exceptions fatal. I am propagating the exception per the suggestion. Just curious if there are strong reasons we want to make close exceptions fatal.

htran1: The existing close that happens from the closer propagates errors. There are some writers like the `HiveWritableHdfsDataWriter` that finalizes files on close. If that fails then there may be an incomplete file and the task should fail instead of publish.

gobblin-core/src/test/java/org/apache/gobblin/writer/PartitionedWriterTest.java
```diff
@@ -176,4 +165,31 @@ public void testControlMessageHandler() throws IOException {
...
+    PartitionedDataWriter writer = new PartitionedDataWriter<String, String>(builder, state);
+
+    String record1 = "abc";
+    writer.writeEnvelope(new RecordEnvelope(record1));
+
+    String record2 = "123";
+    writer.writeEnvelope(new RecordEnvelope(record2));
+
+    //Sleep for more than cache expiration interval
+    Thread.sleep(1500);
```
aplex: Is there a way to test this without Thread.sleep, e.g. with some virtual time or by explicitly triggering expiration? 
Using sleeps can slow down the test suite, and also make it flakey. For example, if the tests are running on overloaded Tracis CI machine, the cache timer that normally runs every second, can run every 3rd second, and the test will randomly fail.

sv2000: Guava does provide Ticker interface for testing cache evictions without having to put sleeps. Unfortunately, the cache has to be built with the ticker, which is not what we want for our test. In our test, since we explicitly call cache cleanup(), we should not encounter the issue you are describing.

PR #2880, opened 2020-01-31
[GOBBLIN-1038] Set default dataset descriptor configs based on the DataNode

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1038

### Description

Set default dataset descriptor config for class/platform by including them in each DataNode. For ones like `BaseDataNode` where it doesn't make sense, they are null so there is no default.

Also added a check that all dest nodes are the same type of datanode since it wouldn't work anyway for there to be multiple types.

## Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/pathfinder/AbstractPathFinder.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/pathfinder/AbstractPathFinder.java
@@ -24,6 +24,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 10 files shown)


jack-moseley:
@arjun4084346 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/pathfinder/AbstractPathFinder.java
```diff
@@ -103,6 +105,11 @@
       this.destNodes.add(destNode);
     }
 
+    // All dest nodes should be the same class
+    if (this.destNodes != null && this.destNodes.stream().map(Object::getClass).collect(Collectors.toSet()).size() > 1) {
```
arjun4084346: Can our path finding algorithm return dest nodes with different types? just curious.

arjun4084346:
+1 LGTM.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/pathfinder/AbstractPathFinder.java
jack-moseley: This is the destination nodes specified by the user in the `destinationIdentifiers` property, it allows more than one. This makes it explicitly fail if they specified an HDFS and mysql cluster for example (which would probably fail somewhere else in compilation anyway).

PR #2882, opened 2020-02-06
[GOBBLIN-1042] Fix ForkMetric incorrect return type of parent metric object and relevant unit tests 

autumnust:
Along with some documentation enhancement, typo fixing, renaming of methods, etc. 

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1042

### Description

### Tests

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/InMemoryWuSingleTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/InMemoryWuSingleTask.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.cluster;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.gobblin.configuration.ConfigurationKeys;
+import org.apache.gobblin.runtime.JobState;
+import org.apache.gobblin.runtime.util.StateStores;
+import org.apache.gobblin.source.workunit.WorkUnit;
+import org.apache.gobblin.writer.DataWriter;
+import org.apache.gobblin.writer.DataWriterBuilder;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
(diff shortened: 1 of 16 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -400,15 +403,20 @@ private boolean taskSuccessfulInPriorAttempt(String taskId) {
...
+            this.taskStateTracker.onTaskRunCompletion(task);
+            log.error("Could not submit task for workunit {}", workUnit, e);
+          } else {
+            // task was created and submitted, but failed later, so cancel the task to decrement the CountDownLatch
+            task.cancel();
+            log.error("Failure after task submitted for workunit {}", workUnit, e);
+          }
+
+          // Remove entry in the metrics registry when task failed to be executed .
+          this.cleanMetrics();
```
sv2000: Should we be cleaning metrics when task execution fails? Currently, cleanMetrics() will clean up metrics for all tasks associated with a Multi-task attempt. Won't we delete metrics for tasks which may be currently running?

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/TestSingleTaskRerun.java
```diff
@@ -0,0 +1,56 @@
...
+package org.apache.gobblin.cluster;
+
+import org.junit.Assert;
+import org.testng.annotations.Test;
+
+
+/**
+ * Notes & Usage:
+ * 0. This test could be used to reproduce task-execution issue in Gobblin-Cluster, within each container.
+ * 1. The workunit is being drafted in {@link InMemoryWuFailedSingleTask}.
```
sv2000: drafted -> created?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/InMemoryWuSingleTask.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.collect.Lists;
+import com.typesafe.config.Config;
+
+/**
+ * Instead of deserializing {@link JobState} and {@link WorkUnit} from filesystem, create them in memory.
+ * Uses {@link DummyDataWriter} so that the execution of a Task goes through.
+ */
+public class InMemoryWuSingleTask extends InMemoryWuFailedSingleTask {
```
sv2000: Perhaps better to not have a successful task extend a "meant-to-fail" task. If avoiding code duplication is the intention here, we can perhaps refactor the getJobState and getWorkUnits method of InMemoryWuFailedSingleTask to an abstract InMemorySingleTask class and have both failed/successful tasks extend the abstract class?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
autumnust: Looks like I shouldn't clean the metric registry since it is a class-static variable shared by all instances. Adjust the logic.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/InMemoryWuSingleTask.java
autumnust: SGTM. Addressed.

PR #2892, opened 2020-02-14
GOBBLIN-1052: Create a spec consumer path if it does not exist in FS …

sv2000:
…SpecConsumer

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1052

### Description
Create a spec consumer path if it does not exist in FS SpecConsumer. This PR also makes the following improvements:
1. FsSpecProducer writes JobSpecs to a tmp location followed by a rename, instead of directly overwriting specs in the final location.
2. Change constructors of FsSpecProducer/FsSpecConsumer to allow passing in a FileSystem object.

### Tests
Modified existing unit test cases.

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
@@ -278,7 +278,7 @@ public class HelixUtils {
   }
 
   /**
-   * Returns the Helix Workflow Ids given {@link Iterable} of Gobblin job names. The method returns a
+   * Returns the currently running Helix Workflow Ids given an {@link Iterable} of Gobblin job names. The method returns a
    * {@link java.util.Map} from Gobblin job name to the corresponding Helix Workflow Id. This method iterates
    * over all Helix workflows, and obtains the jobs of each workflow from its jobDag.
    *
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
@@ -31,6 +31,7 @@ import org.apache.helix.HelixManager;
 import org.apache.helix.manager.zk.ZKHelixManager;
 import org.apache.helix.model.HelixConfigScope;
 import org.apache.helix.task.JobConfig;
+import org.apache.helix.task.TargetState;
 import org.apache.helix.task.TaskConfig;
 import org.apache.helix.task.TaskDriver;
 import org.apache.helix.task.TaskState;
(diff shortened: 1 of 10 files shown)


sv2000:
@autumnust please review.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -293,6 +294,10 @@ private static void deleteStoppedHelixJob(HelixManager helixManager, String work
     Map<String, WorkflowConfig> workflowConfigMap = taskDriver.getWorkflows();
     for (String workflow : workflowConfigMap.keySet()) {
       WorkflowConfig workflowConfig = taskDriver.getWorkflowConfig(workflow);
+      //Filter out any stale Helix workflows which are not running.
+      if (workflowConfig.getTargetState() != TargetState.START) {
```
autumnust: Why filter out stopped job if the method call is to get a list of workFlowId from Gobblin job name ?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
```diff
@@ -55,10 +56,13 @@
   private Map<URI, Path> specToPathMap = new HashMap<>();
 
 
-  public FsSpecConsumer(Config config) {
+  public FsSpecConsumer(FileSystem fs, Config config) {
```
autumnust: Since it is a public API, should we leave an option of `FsSpecConsumer(Config config)`  (by calling this constructor)?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecProducer.java
```diff
@@ -52,11 +52,13 @@
   protected static final String VERB_KEY = "Verb";
 
   private Path specConsumerPath;
+  private FileSystem fs;
 
-  public FsSpecProducer(Config config) {
+  public FsSpecProducer(FileSystem fs, Config config) {
```
autumnust: Same as above

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
sv2000: Thanks for this comment. The purpose of this method is two fold: 1. The caller may want to cancel a Helix workflow given a job name and 2. The caller may want to get the workflow id and use that to query other information such as the workunits processed by the job. For both 1 and 2, the caller is only interested in the active workflows. In fact, the stale workflows are auto-pruged by Helix based on an expiry interval.

I have modified the javadoc for this method to be more explicit about which workflows are returned.

sv2000:
@autumnust Addressed your comments.

PR #2907, opened 2020-03-04
[GOBBLIN-1068]Clean up cyclic logic in task cancellation

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1068

### Description
The original `cancel` method in `Task.java` has several problems: 
--`Task` is an implementation of `Runable` while itself contains a reference of `Future<?>` which is the handler of the `Runnable` after being submitted. The life-cycle of the `Runnable` shouldn't be controlled by itself as it looks like right now, but should be the `TaskExecutor` or something externally. 
-- Because of the existence of a `Future<?>`  object within Task, there's a bug in the implementation: The extension of `Task`, which is `TaskIFaceWrapper` doesn't have the constructor to initialize this variable, while one of the public method(`completeShutdown` within `cancel`) is exposing the accessing of this variable. This is a bug and the original fix in this PR:https://github.com/apache/incubator-gobblin/pull/2157/files is doing wrong thing to copy the base class's method and remove the calling of `completeShutdown`. In fact, the future object doesn't have to be existed in the first place. 
-- Fixing some other typos, and unit test problems as well. 

### Tests

### Commits


autumnust:
@sv2000  Can you please take a look whenever you have time? Thanks.

autumnust:
> @sv2000 Can you please take a look whenever you have time? Thanks.

Looks like there's a code-change-irrelevant travis failure happening which couldn't be reproduced locally. Will trigger another build after addressing comments if any.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
```diff
@@ -64,19 +74,27 @@ public void run() {
...
-      while (System.currentTimeMillis() <= endTime) {
-        Thread.sleep(1000L);
-        log.warn("Sleeping for {} seconds", sleepTime);
-      }
+
+      AssertWithBackoff.create().maxSleepMs(1000).timeoutMs(Integer.MAX_VALUE).backoffFactor(1).
+          assertTrue(new Predicate<Void>() {
+            @Override
+            public boolean apply(@Nullable Void input) {
+              return System.currentTimeMillis() > endTime && !infiniteSleep;
```
sv2000: Nit: change the order of evaluation to: !infiniteSleep && System.currentTimeMillis() > endTIme

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -40,15 +42,23 @@
         ConfigurationKeys.SOURCE_CLASS_KEY, "org.apache.gobblin.cluster.SleepingCustomTaskSource",
         ConfigurationKeys.JOB_ID_KEY, JOB_ID,
         GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_ENABLED_KEY, Boolean.TRUE,
-        GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS, 10L, SleepingTask.TASK_STATE_FILE_KEY, TASK_STATE_FILE))
-        .withFallback(rawJobConfig);
+        GobblinClusterConfigurationKeys.HELIX_JOB_TIMEOUT_SECONDS, 10L,
+
+        // Put SleepingTask in infinite sleep so that cancellation thereby ensuring cancellation to happen.
```
sv2000: Change comment to: 
// Put SleepingTask in infinite sleep allowing cancellation to happen.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -40,15 +42,23 @@
...
+        SleepingTask.TASK_STATE_FILE_KEY, TASK_STATE_FILE))
+        .withFallback(rawJobConfig.withValue(SLEEPING_TASK_SLEEP_TIME, ConfigValueFactory.fromAnyRef(-1)));
     return ImmutableMap.of(JOB_NAME, newConfig);
   }
 
+  /**
+   * Note This verification now ties to {@link SleepingTask} which is not ideal.
```
sv2000: Note: This verification is now tied to....

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -40,15 +42,23 @@
...
+   */
   @Override
   public void waitForAndVerifyOutputFiles() throws Exception {
+    // Sleeing task is in infinite sleeping, so unless an cancelling is hitting task execution, this line won't be printed.
```
sv2000: Change comment to: "SleepingTask is in an infinite sleep. The log line is printed only when a cancellation in invoked."

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -160,25 +163,40 @@ public void run()
...
-      interruptTaskExecution(countDownLatch);
+      interruptTaskExecution(Optional.of(countDownLatch));
     }
     log.info("All assigned tasks of job {} have completed in container {}", jobId, containerIdOptional.or(""));
   }
 
-  private void interruptTaskExecution(CountDownLatch countDownLatch)
-      throws InterruptedException {
+  /**
+   * A helper function that shutting down all outstanding tasks and destroy taskExecutor if timeout on waiting for
```
sv2000: Change to "...that shuts down all outstanding tasks and shuts down the taskExecutor if it times out on a task termination"

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -160,25 +163,40 @@ public void run()
...
-    this.tasks.forEach(Task::shutdown);
-    if (!countDownLatch.await(5, TimeUnit.SECONDS)) {
-      log.warn("Graceful shutdown of job timed out. Killing all outstanding tasks.");
-      try {
+    this.shutdownTasks();
+    try {
+      if (countDownLatch.isPresent()) {
+        if (!countDownLatch.get().await(5, TimeUnit.SECONDS)) {
+          log.warn("Graceful shutdown of job timed out. Killing all outstanding tasks.");
+          this.taskExecutor.shutDown();
```
sv2000: Can this line be refactored outside the else {..} block to avoid it being called in both if and the else?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
```diff
@@ -64,19 +70,27 @@ public void run() {
...
+            public boolean apply(@Nullable Void input) {
+              return System.currentTimeMillis() > endTime;
+            }
+          }, "Waiting for the job to start...");
+
       log.info("Hello World!");
       super.run();
     } catch (InterruptedException e) {
-      log.error("Sleep interrupted.");
+      log.info("Sleep interrupted.");
```
sv2000, line 85: Nit: Prefer to keep the log level here to error, given that this is an exception we are propagating.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -121,7 +124,8 @@ void testJobShouldGetCancelled() throws Exception {
    *   We confirm the execution by again inspecting the zNode and ensuring its TargetState is START. </li>
    * </ul>
    */
-  @Test (dependsOnMethods = { "testJobShouldGetCancelled" }, groups = {"disabledOnTravis"})
+//  @Test (dependsOnMethods = { "testJobShouldGetCancelled" }, groups = {"disabledOnTravis"})
```
sv2000, line 127: Can we remove the commented line?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -152,13 +155,15 @@ public void run()
...
+    Pair<List<Task>, List<Future<?>>> taskExecutionResult = runWorkUnits(countDownLatch);
+    this.tasks = taskExecutionResult.getLeft();
+    this.taskFutures = taskExecutionResult.getRight();
     log.info("Waiting for submitted tasks of job {} to complete in container {}...", jobId, containerIdOptional.or(""));
     try {
       while (countDownLatch.getCount() > 0) {
         if (this.interruptionPredicate.test(this)) {
           log.info("Interrupting task execution due to satisfied predicate.");
-          interruptTaskExecution(countDownLatch);
+          interruptTaskExecution(Optional.of(countDownLatch));
```
sv2000, line 166: Optional is not intended to be used as a function parameter and should be used only as a return value. [Here](https://www.quora.com/Why-is-using-java-util-Optional-as-a-method-parameter-not-recommended) is one of several references describing the problems with using Optional as a function parameter.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -169,25 +174,35 @@ public void run()
...
       }
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to shutdown task executor.", e);
     }
   }
 
+  /**
+   * Shutting down the whole {@link GobblinMultiTaskAttempt} by shutting down all its outstanding tasks and taskExecutor.
+   */
+  public void cancel() throws InterruptedException{
```
sv2000, line 202: Nit: Add a whitespace after InterruptedException.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -95,6 +78,21 @@
 import org.apache.gobblin.writer.WatermarkAwareWriter;
 import org.apache.gobblin.writer.WatermarkManager;
 import org.apache.gobblin.writer.WatermarkStorage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+import com.google.common.base.Optional;
```
sv2000, line 85: Is the ordering of imports correct? I thought org.apache.gobblin should come after com.google.*.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -169,25 +174,35 @@ public void run()
...
-      try {
+    this.shutdownTasks();
+    try {
+      if (!countDownLatch.isPresent() || !countDownLatch.get().await(5, TimeUnit.SECONDS)) {
+        log.warn("Shutting down TaskExecutor. Killing all outstanding tasks.");
         this.taskExecutor.shutDown();
-      } catch (Throwable t) {
-        throw new RuntimeException("Failed to shutdown task executor.", t);
       }
+    } catch (Exception e) {
```
sv2000, line 194: Curious why we changed from Throwable to Exception here.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
sv2000, line 202: Can we merge interruptTaskExecution and cancel() methods into one, particularly since cancel () is not doing anything more than calling the interruptTaskExecution method?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/Task.java
```diff
@@ -394,15 +391,8 @@ public void run() {
...
-      synchronized (this) {
-        if (this.taskFuture == null || !this.taskFuture.isCancelled()) {
-          this.taskStateTracker.onTaskRunCompletion(this);
-          completeShutdown();
-          this.taskFuture = null;
-        } else {
-          LOG.info("will not decrease count down latch as this task is cancelled");
-        }
-      }
+      this.taskStateTracker.onTaskRunCompletion(this);
```
sv2000, line 394: Is the removing of the synchronized block intentional? Is there a chance that Task#cancel() is invoked from multiple threads?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/task/TaskIFaceWrapper.java
```diff
@@ -31,6 +28,9 @@
 import org.apache.gobblin.runtime.TaskStateTracker;
 import org.apache.gobblin.runtime.fork.Fork;
 
+import com.google.common.base.Optional;
```
sv2000, line 31: Same comment about the import order as earlier.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
sv2000: Also: should we just call cancel here?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
arjun4084346: I think I wrongly wrote it error. warn seems more appropriate to me. If it comes here, it does not mean that the SleepingTask did anything **erroneous**, but got interrupted. could be info too, but looks like an important message, especially in SleepingTask

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SleepingTask.java
```diff
@@ -64,19 +70,27 @@ public void run() {
         throw new IOException("File creation error: " + taskStateFile.getName());
       }
       long endTime = System.currentTimeMillis() + sleepTime * 1000;
-      while (System.currentTimeMillis() <= endTime) {
-        Thread.sleep(1000L);
-        log.warn("Sleeping for {} seconds", sleepTime);
-      }
+
+      AssertWithBackoff.create().maxSleepMs(1000).timeoutMs(Integer.MAX_VALUE).backoffFactor(1).
```
arjun4084346, line 74: What's the point of giving timeoutMs(Integer.MAX_VALUE) ?

I think AssertWithBackoff should be used in place of Thread.Sleep when sleeping time is not known, e.g. when we are expecting for something to be true (or something to happen), but do not know when that will happen, and at the same time we do not want to sleep for a fixed amount of time which could be 1) too big, which would delay build, or 2) too small , that what we were expecting did not happen in that time.

I know AssertWithBackoff is recommended in Gobblin team, but in this case, sleep time is fixed. AssertWithBackoff makes no sense here.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -167,9 +171,20 @@ public void testJobRestartViaSpec() throws Exception {
...
     }, "Waiting for Workflow TargetState to be STOP");
 
-    //Ensure that the SleepingTask did not terminate normally i.e. it was interrupted. We check this by ensuring
-    // that the line "Hello World!" is not present in the logged output.
-    suite.waitForAndVerifyOutputFiles();
+    // Ensure that the SleepingTask did not terminate normally i.e. it was interrupted. We check this by ensuring
+    // that the line "Hello World!" is not present in the logged output but only "Sleeping interrupted"
+    // It is important to have back-off here since ZNode STOP state doesn't mean the actual sleeping task has been terminated
+    // since shutdown of task is async.
+    AssertWithBackoff.create().maxSleepMs(1000).timeoutMs(120000).backoffFactor(1).assertTrue(input -> {
```
arjun4084346, line 178: should be `maxSleepMs(1000L).timeoutMs(120000L)`

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -20,20 +20,26 @@
...
+    super(jobConfigOverrides.withValue(SLEEPING_TASK_SLEEP_TIME, ConfigValueFactory.fromAnyRef(100)));
   }
 
   @Override
   public void waitForAndVerifyOutputFiles() throws Exception {
+    // SleepingTask is in an infinite sleep. The log line is printed only when a cancellation in invoked.
+    Assert.assertTrue(verifyFileForMessage(this.jobLogOutputFile, "Sleep interrupted"));
+
     // If the job is cancelled, it should not have been able to write 'Hello World!'
     Assert.assertFalse(verifyFileForMessage(this.jobLogOutputFile, "Hello World!"));
```
arjun4084346, line 43: Can we keep the deleted line. A bug was fixed and that line ensure that NPE does not happen. Yea, it can probably be replaced by Assert.assertFalse(verifyFileForMessage(this.jobLogOutputFile, "Exception"));

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -372,9 +379,9 @@ private boolean taskSuccessfulInPriorAttempt(String taskId) {
    * @param countDownLatch a {@link java.util.concurrent.CountDownLatch} waited on for job completion
    * @return a list of {@link Task}s from the {@link WorkUnit}s
    */
-  private List<Task> runWorkUnits(CountUpAndDownLatch countDownLatch) {
-
+  private Pair<List<Task>, List<Future<?>>> runWorkUnits(CountUpAndDownLatch countDownLatch) {
```
arjun4084346, line 382: How about returning List<Pair<Task, Future<?>> ?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -416,14 +425,17 @@ private boolean taskSuccessfulInPriorAttempt(String taskId) {
...
           this.taskStateTracker.onTaskRunCompletion(task);
           log.error("Could not submit task for workunit {}", workUnit, e);
         } else {
-          // task was created and submitted, but failed later, so cancel the task to decrement the CountDownLatch
-          task.cancel();
+          // task was created and submitted, but failed later(and still pop back here), so cancel the task to decrement the CountDownLatch
+          if (!taskFuture.cancel(true)) {
+            log.error("Failure in terminating submitted task");
+          }
+          this.taskStateTracker.onTaskRunCompletion(task);
```
arjun4084346, line 438: Do we want to call `this.taskStateTracker.onTaskRunCompletion(task)` if `taskFuture.cancel(true)` returns false?

PR #2908, opened 2020-03-04
[GOBBLIN-1069] Add NPE check in handleContainerCompletion method

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1069

### Description
Add NPE check in handleContainerCompletion method to make sure call the method twice for the same container will not fail the job

### Tests
Unit tests

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
@@ -620,7 +620,10 @@ public class YarnService extends AbstractIdleService {
    */
   protected void handleContainerCompletion(ContainerStatus containerStatus) {
     Map.Entry<Container, String> completedContainerEntry = this.containerMap.remove(containerStatus.getContainerId());
-
+    if(completedContainerEntry == null) {
+      //No map for this container means we don't maintain this container, directly return
+      return;
+    }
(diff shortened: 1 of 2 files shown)


gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -620,7 +620,10 @@ private boolean shouldStickToTheSameNode(int containerExitStatus) {
    */
   protected void handleContainerCompletion(ContainerStatus containerStatus) {
     Map.Entry<Container, String> completedContainerEntry = this.containerMap.remove(containerStatus.getContainerId());
-
+    if (completedContainerEntry == null) {
```
autumnust: In what kind of scenarios will it happen, when you say " we don't maintain this container " ?

ZihanLi58: We have met some NPE when the container completion happens. And we alway add the containerId to the map when one container is allocated and remove the id only inside this method. So we assume that there may be chances that for one completed container, this method will be called more than once. And in that scenario (the method has already be called once to handle this completion), we can say we no longer maintain this container.

PR #2909, opened 2020-03-05
[GOBBLIN-1071] Retry task initialization

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1071

### Description
The major part of this PR are: 
- In `GobblinMultiTaskAttempt.java`, created a method `createTaskWithRetry`  to wrap around `createTaskRunnable`, so that the task-initialization as part of `createTaskRunnable`  could be retried. The retry-component is built using `Retryer` library which essentially uses guava's retry library. To make that easier, also added some features in `RetryerFactory.java`, especially attempt-based retry to better fit our requirement here. This part is tested within `TaskErrorIntegrationTest.java ` where sits a method `extractorCreationError`. We created a test case for `extractorCreationErrorWithRetry` to demonstrate retry case. 
- In `GobblinHelixTask` also introduced retry of calling `SingleTask`'s constructor in the same approach. The relevant testing is in `GobblinHelixTaskTest.java` where we fail the constructor call on purpose and since we cannot easily heal the failure, it results in a retry-failure at the end. 

This PR also contains many renaming for refactoring as I went through the code base. Most of the rename is because the original name is far from being accurate. 

### Tests

### Commits


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -50,49 +54,74 @@
...
+      TaskAttemptBuilder taskAttemptBuilder, StateStores stateStores, Config dynamicConfig) throws IOException {
     _jobId = jobId;
     _workUnitFilePath = workUnitFilePath;
     _jobStateFilePath = jobStateFilePath;
     _fs = fs;
     _taskAttemptBuilder = taskAttemptBuilder;
     _stateStores = stateStores;
     _dynamicConfig = dynamicConfig;
+
+    _workUnits = getWorkUnits();
```
sv2000: Why do we have to eagerly initialize workunits?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -50,49 +54,74 @@
...
 
     try (SharedResourcesBroker<GobblinScopeTypes> globalBroker = SharedResourcesBrokerFactory
         .createDefaultTopLevelBroker(jobConfig, GobblinScopeTypes.GLOBAL.defaultScopeInstance())) {
-      SharedResourcesBroker<GobblinScopeTypes> jobBroker = getJobBroker(jobState, globalBroker);
+      SharedResourcesBroker<GobblinScopeTypes> jobBroker = getJobBroker(_jobState, globalBroker);
+
+      // Secure atomicity of taskAttempt's execution.
+      _lock.lock();
+      try {
+        _taskAttempt = _taskAttemptBuilder.build(_workUnits.iterator(), _jobId, _jobState, jobBroker);
```
sv2000: Looks like the attempt creation can be moved outside the critical section.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -50,49 +54,74 @@
...
+        _taskAttempt.runAndOptionallyCommitTaskAttempt(GobblinMultiTaskAttempt.CommitPolicy.IMMEDIATE);
+        _taskAttemptBuilt.signal();
```
sv2000: How do we handle missed signals? Are we sure the thread waiting on the signal is waiting?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
autumnust: Originally was thinking to have all heavy-lifting work in constructor ( and also make retry more meaning ful there) I don't have strong opinion on this though..

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
```diff
@@ -44,6 +49,9 @@
 import org.apache.gobblin.source.workunit.WorkUnit;
 import org.apache.gobblin.util.Id;
 
+import static org.apache.gobblin.util.retry.RetryerFactory.RETRY_INTERVAL_MS;
```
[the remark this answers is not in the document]
autumnust: Addressed, thanks for catching.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
```diff
@@ -97,28 +105,35 @@ public GobblinHelixTask(TaskRunnerSuiteBase.Builder builder,
...
-                               dynamicConfig);
+    final Config taskLevelConfig = builder.getConfig()
+        .withValue(GobblinClusterConfigurationKeys.TASK_RUNNER_HOST_NAME_KEY, ConfigValueFactory.fromAnyRef(builder.getHostName()))
+        .withValue(GobblinClusterConfigurationKeys.CONTAINER_ID_KEY, ConfigValueFactory.fromAnyRef(builder.getContainerId()))
+        .withValue(GobblinClusterConfigurationKeys.HELIX_INSTANCE_NAME_KEY, ConfigValueFactory.fromAnyRef(builder.getInstanceName()))
+        .withValue(GobblinClusterConfigurationKeys.HELIX_JOB_ID_KEY, ConfigValueFactory.fromAnyRef(this.helixJobId))
+        .withValue(GobblinClusterConfigurationKeys.HELIX_TASK_ID_KEY, ConfigValueFactory.fromAnyRef(this.helixTaskId))
+        .withValue(GobblinClusterConfigurationKeys.HELIX_PARTITION_ID_KEY, ConfigValueFactory.fromAnyRef(partitionNum));
+
+    Retryer<SingleTask> retryer = RetryerFactory.newInstance(taskLevelConfig);
```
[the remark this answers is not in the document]
autumnust: Regarding to the default setting, the timeout is 5 minutes ( it uses TimeUnit.MINUTES.toMillis to convert 5 minutes to ms )

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -456,6 +464,33 @@ private Task createTaskRunnable(WorkUnitState workUnitState, CountDownLatch coun
...
+    // An "effectively final" variable for counting how many retried has been done, mostly for logging purpose.
+    final AtomicInteger counter = new AtomicInteger(0);
+
+    try {
+      return retryer.call(new Callable<Task>() {
+        @Override
+        public Task call()
+            throws Exception {
+          counter.incrementAndGet();
+          log.info(String.format("The %s time that trying create Task object", counter.get()));
```
sv2000: "Task creation Attempt# %s"?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -456,6 +464,33 @@ private Task createTaskRunnable(WorkUnitState workUnitState, CountDownLatch coun
...
+          return createTaskRunnable(workUnitState, countDownLatch);
+        }
+      });
+    } catch (Exception e) {
+      throw new RuntimeException("Execution in creating a Task-with-retry failed", e);
```
sv2000: "Exception creating Task after k retries"?

sv2000:
@autumnust LGTM. Once travis passes, will merge.

PR #2910, opened 2020-03-05
[GOBBLIN-1179] Add typed config in salesforce

arekusuri:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1179

### Description
replace `properties.get()` with typed class field `config.bulkApiUseQueryAll`.
It would be neater/readable and any typo can be find out at compiling time.

>-workUnitState.getPropAsBoolean(BULK_API_USE_QUERY_ALL, DEFAULT_BULK_API_USE_QUERY_ALL);
>+this.bulkApiUseQueryAll = conf.bulkApiUseQueryAll;

### Tests
https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=23983645&job=salesforce_task_full&attempt=0


autumnust:
Please go through https://gobblin.readthedocs.io/en/latest/developer-guide/CodingStyle/ for style guidance, also all license header for newly created files are missing. 

Also, an Linkedin-specific test link is not useful for other folks. You may consider unit tests for `ConstraintUtil` to maintain code-coverage.

zxcware:
Alex. Thanks for taking the action to improve the config key story. Is there a grand plan for the gobblin configuration system? Or you're just trying to improve life of gobblin dev to work with salesforce configurations, for gobblin users, they have to deal with strings.

arekusuri:
> Please go through https://gobblin.readthedocs.io/en/latest/developer-guide/CodingStyle/ for style guidance, also all license header for newly created files are missing.
> 
> Also, an Linkedin-specific test link is not useful for other folks. You may consider unit tests for `ConstraintUtil` to maintain code-coverage.

Thanks for your comment. This is a very early version, I am not sure how far it could go.  My original idea was to create a separate lib project and simply add a library into gobblin project.

jhsenjaliya:
@autumnust, @arekusuri , If this style make sense and dont see any possible problem,  I think we should put some of the core classes in gobblin.config package and start using this style everywhere. looking at config improvement myself, I like this style.
@arekusuri , btw, this could be very nice feature set on top of [typesafe](https://github.com/lightbend/config) or [tscfg](https://github.com/carueda/tscfg) itself. Thanks for the improvements. let me know if you need any help to extend this to all gobblin config.

arekusuri:
> @autumnust, @arekusuri , If this style make sense and dont see any possible problem, I think we should put some of the core classes in gobblin.config package and start using this style everywhere. looking at config improvement myself, I like this style.
> @arekusuri , btw, this could be very nice feature set on top of [typesafe](https://github.com/lightbend/config) or [tscfg](https://github.com/carueda/tscfg) itself. Thanks for the improvements. let me know if you need any help to extend this to all gobblin config.

Thanks for your reviewing. Here is the [design doc](https://docs.google.com/document/d/1ZqRN4PHq2LXFMfMwpbe9UHfgOO-xIRUKdwzZ24b-F00/edit#)

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
```diff
@@ -0,0 +1,80 @@
...
+public class ConstraintUtil {
+  private ConstraintUtil() {
+  }
+
+  public static Object constraint(Field field, Object value, Object defaultValue) {
+    if (value == null) {
+      return defaultValue;
+    }
+    Class type = field.getType();
+    if (type == int.class) {
```
zxcware, line 22: too many `if...else`, use switch and `FieldValidator` subclasses?

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/Alias.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+@Target({ElementType.FIELD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Alias {
```
zxcware, line 11: How many aliases can a key have?

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
```diff
@@ -0,0 +1,126 @@
...
+import lombok.SneakyThrows;
+import org.apache.commons.beanutils.BeanUtilsBean;
+import org.apache.commons.beanutils.ConvertUtilsBean;
+
+
+public class TypedConfig {
+
+  public TypedConfig(Properties prop) {
+    if (prop != null) {
+      fulFillFields(this.getClass(), prop);
```
zxcware, line 16: typo `fulFill` -> `fulfill`

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
```diff
@@ -0,0 +1,126 @@
...
+      afterFulfill();
+    }
+  }
+
+  /**
+   * subclass could do something after fulfill by override afterFulfill
+   */
+  protected void afterFulfill() {
```
zxcware, line 24: remove premature abstractions

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+      }
+    });
+    return beanUtilsBean.getConvertUtils().convert(value, targetClazz);
+  }
+
+  /**
+   * subclass could do something before toProp by overriding
+   */
+  protected void beforeToProp() {
```
zxcware, line 102: remove premature abstractions

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/Alias.java
arekusuri: One. Do we want multiple?

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
arekusuri: Better keep using if-else. I think `switch` is like for the case there is very little code between  switch-cases. https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
arekusuri: good catch!

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import org.apache.gobblin.typedconfig.compiletime.EnumOptions;
+import org.apache.gobblin.typedconfig.compiletime.IntRange;
+import org.apache.gobblin.typedconfig.compiletime.LongRange;
+import org.apache.gobblin.typedconfig.compiletime.StringRegex;
+
+
+public class ConstraintUtil {
```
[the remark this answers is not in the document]
arekusuri: Added. Thanks!

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
arekusuri: removed.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/TypedConfig.java
arekusuri: removed.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
zxcware: I mean something like this https://github.com/apache/incubator-gobblin/blob/master/gobblin-modules/gobblin-parquet/src/main/java/org/apache/gobblin/converter/parquet/JsonElementConversionFactory.java.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/Alias.java
arekusuri: changed to multiple.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
arekusuri: create separate function for every type of data.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+  static private Object getLongValue(Field field, Object value, Object defaultValue) {
+    LongRange longRange = field.getAnnotation(LongRange.class);
+    long[] range = longRange.value();
+    if (range == null) {
+      return value;
+    }
+    long longValue = Long.parseLong(value.toString());
+    if (range.length != 2 || (range.length == 2 && longValue > range[0] && longValue < range[1])) {
```
htran1, line 84: We should raise an error instead of using the default if an out of range value is provided. This is to avoid confusion where the user thinks that the execution used the config they provide.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.gobblin.typedconfig.compiletime.EnumOptions;
+import org.apache.gobblin.typedconfig.compiletime.IntRange;
+import org.apache.gobblin.typedconfig.compiletime.LongRange;
+import org.apache.gobblin.typedconfig.compiletime.StringRegex;
+
+
+/**
+ * Util class for handling constrain annotation
```
htran1, line 32: constrain -> constraint

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
arekusuri: Thanks! it makes sense.

gobblin-salesforce/src/main/java/org/apache/gobblin/typedconfig/ConstraintUtil.java
arekusuri: good catch! thanks!

PR #2912, opened 2020-03-06
[GOBBLIN-1072] Being more conservative on releasing containers

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1072

### Description
- An additional data structure `MaxValueEvictingQueue` to wrap multiple instances of `numTargetContainers` and behaved like an sliding-window where we could fetch the maxValue in the sliding-window in constant time. 
- The advantage of this data structure is, if there's fluctuation in the number of active Helix partitions, a single fluctuation wont't impact the number of containers we requested in `YarnService`. 
- The sliding-window's size is defined by the number of attempts that we query Helix for getting partitions number. 
- Use a different way to tag an instance as "unused" : Currently whenever we found an instance with no Helix partition assigned to it when inspecting, we tag it as "unused" and tell YarnService it could be a candidate for scaling down. To make it less aggressive, we maintain a map with Helix instance as the Key and the "lastIdleTime" as the value, when currentTime - lastIdleTime is beyond tolerance, we then tag this instance as unused. 

### Tests

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
@@ -50,6 +50,8 @@ import com.typesafe.config.Config;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+import static org.apache.gobblin.yarn.GobblinYarnTaskRunner.HELIX_YARN_INSTANCE_NAME_PREFIX;
+
 
 /**
  * The autoscaling manager is responsible for figuring out how many containers are required for the workload and
@@ -68,13 +70,13 @@ public class YarnAutoScalingManager extends AbstractIdleService {
   private final int DEFAULT_AUTO_SCALING_MIN_CONTAINERS = 1;
   private final String AUTO_SCALING_MAX_CONTAINERS = AUTO_SCALING_PREFIX + "maxContainers";
   // A rough value of how much containers should be an intolerable number.
-  private final int DEFAULT_AUTO_SCALING_MAX_CONTAINERS = 5000;
+  private final int DEFAULT_AUTO_SCALING_MAX_CONTAINERS = Integer.MAX_VALUE;
   private final String AUTO_SCALING_INITIAL_DELAY = AUTO_SCALING_PREFIX + "initialDelay";
   private final int DEFAULT_AUTO_SCALING_INITIAL_DELAY_SECS = 60;
 
-  private final String WINDOW_SIZE_OBSERVING_CONTAINER_REQUEST = AUTO_SCALING_PREFIX + "windowSize";
+  private final String AUTO_SCALING_WINDOW_SIZE = AUTO_SCALING_PREFIX + "windowSize";
 
-  private final static int DEFAULT_MAX_IDLE_TIME_BEFORE_SCALING_DOWN_MIN = 10;
+  private final static int DEFAULT_MAX_IDLE_TIME_BEFORE_SCALING_DOWN_MINUTES = 10;
 
   private final Config config;
   private final HelixManager helixManager;
@@ -83,8 +85,8 @@ public class YarnAutoScalingManager extends AbstractIdleService {
   private final int partitionsPerContainer;
   private final int minContainers;
   private final int maxContainers;
-  private final MaxValueEvictingQueue slidingFixedSizeWindow;
-  private static int maxIdleTimeInMinBeforeScalingDown = DEFAULT_MAX_IDLE_TIME_BEFORE_SCALING_DOWN_MIN;
+  private final SlidingWindowReservoir slidingFixedSizeWindow;
+  private static int maxIdleTimeInMinutesBeforeScalingDown = DEFAULT_MAX_IDLE_TIME_BEFORE_SCALING_DOWN_MINUTES;
 
   public YarnAutoScalingManager(GobblinApplicationMaster appMaster) {
     this.config = appMaster.getConfig();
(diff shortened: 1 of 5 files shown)


gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -196,9 +207,84 @@ void runInternal() {
       // adjust the number of target containers based on the configured min and max container values.
       numTargetContainers = Math.max(this.minContainers, Math.min(this.maxContainers, numTargetContainers));
 
+      slidingFixedWindow.add(numTargetContainers);
```
sv2000: A problem with your proposal is that it is based on numTargetContainers (and indirectly, numPartitions) which may remain unchanged over a window of time, even though the partitions with no assigned participant may be changing from one observation interval to the next. 

E.g. Let us assume we have 4 Helix partitions. In one observation interval, suppose that [1,2,3] have an assigned participant but not 4. In the next observation interval, suppose partitions [2,3,4] have an assigned participant but not 1. If your sliding window is of size 2, then you will infer that numTargetContainers is 3, while what you want is for numTargetContainers to be 4. 

I think what you want is to track which Helix instances are unused. The current implementation is based on a point-in-time check against Helix for the set of partitions which have no assigned participant. One possible improvement here could be marking a Helix instance as "unused" if it has no assigned participant for the last 10 minutes.

autumnust: Discussed offline. Agreed on new way to tag an instance as "unused" and be more conservative on container release.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -189,16 +224,123 @@ void runInternal() {
...
+     */
+    boolean absenceUnderTolerance(String participant){
+      return System.currentTimeMillis() - instanceIdleSinceWhen.get(participant) <
+          TimeUnit.MINUTES.toMillis(maxIdleTimeInMinBeforeScalingDown);
+    }
+  }
+
+  /**
+   * A FIFO queue with fixed size and returns maxValue among all elements within the queue in constant time.
+   * This data structure prevent temporary fluctuation in the number of active helix partitions as the size of queue
```
sv2000: prevent -> prevents

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -189,16 +224,123 @@ void runInternal() {
...
+   * grows and will be less sensitive when scaling down is actually required.
+   *
+   * The interface for this lass is implemented in a minimal-necessity manner to serve only as a sliding-sized-window
```
sv2000: Typo: drop "lass"

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -189,16 +224,123 @@ void runInternal() {
...
+   * which captures max value. It is NOT built for general purpose.
+   */
+  static class MaxValueEvictingQueue {
```
sv2000: Can we call this a "SlidingWindowReservoir" instead of MaxValueEvictingQueue, as it seems to be tracking the last K elements with a method getMax() to return the maximum value from the reservoir.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -189,16 +224,123 @@ void runInternal() {
...
+    private ArrayDeque<Integer> evictQueue;
```
sv2000: evictQueue -> fifoQueue?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -62,17 +67,24 @@
   private final String AUTO_SCALING_MIN_CONTAINERS = AUTO_SCALING_PREFIX + "minContainers";
   private final int DEFAULT_AUTO_SCALING_MIN_CONTAINERS = 1;
   private final String AUTO_SCALING_MAX_CONTAINERS = AUTO_SCALING_PREFIX + "maxContainers";
-  private final int DEFAULT_AUTO_SCALING_MAX_CONTAINERS = Integer.MAX_VALUE;
+  // A rough value of how much containers should be an intolerable number.
+  private final int DEFAULT_AUTO_SCALING_MAX_CONTAINERS = 5000;
```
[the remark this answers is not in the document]
autumnust: If we really request that much, does Yarn have any protection against us ?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
```diff
@@ -189,16 +224,123 @@ void runInternal() {
         }
       }
 
+      // Find all participants appearing in this cluster. Note that Helix instances can contain cluster-manager
+      // and potentially replanner-instance.
+      Set<String> allParticipants = getParticipants(GobblinYarnTaskRunner.class.getSimpleName());
```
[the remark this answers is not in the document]
autumnust: Good catch, I prefer to put that in `GobblinYarnTaskRunner` itself.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnAutoScalingManager.java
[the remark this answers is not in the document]
sv2000: In practice, the number of containers we request is limited by number of Helix partitions. So we would never get close to this limit. On the other hand, setting the default to a smaller value may result in hard-to-debug scenarios, where some partitions are stuck due to resource limits imposed by this setting.

PR #2915, opened 2020-03-09
[GOBBLIN-1075] Add option to return latest failed flows

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1075

### Description

Add query boolean option that will only return failed flow statuses/executions.

### Tests

Updated test

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowexecutions.restspec.json
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowexecutions.restspec.json
@@ -31,6 +31,10 @@
         "name" : "tag",
         "type" : "string",
         "optional" : true
+      }, {
+        "name" : "executionStatus",
+        "type" : "string",
+        "optional" : true
       } ]
     } ],
     "entity" : {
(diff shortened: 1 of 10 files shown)


jack-moseley:
@arjun4084346 please review

arjun4084346:
I think instead of giving an option to return "failed" flows, we should give option to return flows with ANY status. e.g. how many flows are in COMPILED state, how many are RUNNING.... and so on

arjun4084346:
We should also code it in an flexible way, so that now the filter supports a particular flow status(es) , and in future maybe a filter to support all GET flowConfigs with a particular config.
The second point can be really helpful, when users(SREs) have to add/change one config to multiple flows. This has happened in past. cc @Pawan Reddy

jack-moseley:
> I think instead of giving an option to return "failed" flows, we should give option to return flows with ANY status. e.g. how many flows are in COMPILED state, how many are RUNNING.... and so on

This is a good idea, I modified it to instead take a query like `&executionStatus=RUNNING` which will return the last running flow.

jack-moseley:
> We should also code it in an flexible way, so that now the filter supports a particular flow status(es) , and in future maybe a filter to support all GET flowConfigs with a particular config.
> The second point can be really helpful, when users(SREs) have to add/change one config to multiple flows. This has happened in past. cc @pawan Reddy

I'm not sure if that should be relevant to this PR, since the code paths for getting flowstatus/flowconfig are completely separate, and I don't think there's much else in flowstatus that would be useful to query by.

Maybe that is a change to getting flowconfig that we could do in the future though (being able to query which flowconfig have a particular key set, etc.)

arjun4084346:
+1 LGTM.

jack-moseley:
@sv2000 please merge

PR #2919, opened 2020-03-10
[GOBBLIN-1078] Coordination between task cancel and initialization in Helix Task

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1078

### Description
The task cancel and task.run is called in different thread from helix side, which introduced problems when: 
- Task get cancelled for some reasons from Helix side, therefore calling `SingleTask#cancel `
- However, TaskAttempt object within SingleTask has not been initialized, therefore it didn't pass the non-null checking and only prints a log. After that, the run method in different thread start to run and never get cancelled. 

This sequence could result in violation of helix quota in each participant: While helix believe there's no partition assigned to a participant and assign a new partition, the previous not-yet-cancelled partition is still running. 

We added a barrier in cancel method to ensure cancel is blocked if taskAttempt object is not yet being initialized. Added unit test to simulate the sequence as well. 

### Tests

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
@@ -20,12 +20,17 @@ package org.apache.gobblin.cluster;
 import java.io.IOException;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
@@ -40,6 +45,7 @@ import org.apache.gobblin.runtime.JobState;
 import org.apache.gobblin.runtime.util.StateStores;
 import org.apache.gobblin.source.workunit.MultiWorkUnit;
 import org.apache.gobblin.source.workunit.WorkUnit;
+import org.apache.gobblin.util.ConfigUtils;
 import org.apache.gobblin.util.JobLauncherUtils;
 import org.apache.gobblin.util.SerializationUtils;
 
@@ -51,7 +57,11 @@ public class SingleTask {
 
   private static final Logger _logger = LoggerFactory.getLogger(SingleTask.class);
 
-  private GobblinMultiTaskAttempt _taskAttempt;
+  public static final String MAX_RETRY_WAITING_FOR_INIT_KEY = "maxRetryBlockedOnTaskAttemptInit";
+  public static final int DEFAULT_MAX_RETRY_WAITING_FOR_INIT = 2;
+
+  @VisibleForTesting
+  GobblinMultiTaskAttempt _taskAttempt;
   private String _jobId;
   private Path _workUnitFilePath;
   private Path _jobStateFilePath;
@@ -60,6 +70,12 @@ public class SingleTask {
   private StateStores _stateStores;
   private Config _dynamicConfig;
   private JobState _jobState;
+
+  // Preventing Helix calling cancel before taskAttempt is created
+  // Checking if taskAttempt is empty is not enough, since canceller runs in different thread as runner, the case to
+  // to avoid here is taskAttempt being created and start to run after cancel has been called.
+  private Condition _taskAttemptBuilt;
+  private Lock _lock;
 
   /**
    * Do all heavy-lifting of initialization in constructor which could be retried if failed,
@@ -75,11 +91,12 @@ public class SingleTask {
     _stateStores = stateStores;
     _dynamicConfig = dynamicConfig;
     _jobState = getJobState();
+    _lock = new ReentrantLock();
+    _taskAttemptBuilt = _lock.newCondition();
   }
 
   public void run()
       throws IOException, InterruptedException {
-    List<WorkUnit> workUnits = getWorkUnits();
 
     // Add dynamic configuration to the job state
     _dynamicConfig.entrySet().forEach(e -> _jobState.setProp(e.getKey(), e.getValue().unwrapped().toString()));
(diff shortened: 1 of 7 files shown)


autumnust:
@ZihanLi58  Please take a look whenever you have time, thanks !

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -160,16 +181,32 @@ protected JobState getJobState()
...
+    int maxRetry = this._dynamicConfig.hasPath(MAX_RETRY_WAITING_FOR_INIT_KEY)
+        ? this._dynamicConfig.getInt(MAX_RETRY_WAITING_FOR_INIT_KEY) : DEFAULT_MAX_RETRY_WAITING_FOR_INIT;
+
+    try {
+      _lock.lock();
       try {
+        while (_taskAttempt == null) {
+          // await return false if timeout on this around
+          if (!_taskAttemptBuilt.await(5, TimeUnit.SECONDS) && ++retryCount > maxRetry) {
+            throw new IllegalStateException("Failed to initialize taskAttempt object before cancel");
```
ZihanLi58: Will this fail the job? Or we just want to log the error and ignore.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -63,6 +69,12 @@
   private Config _dynamicConfig;
   private JobState _jobState;
 
+  // Preventing Helix calling cancel before taskAttempt is created
+  // Checking if taskAttempt is empty is not enough, since canceller runs in different thread as runner, the case to
+  // to avoided here is taskAttempt being created and start to run after cancel has been called.
```
ZihanLi58: nit: the case to be avoided?

autumnust: Addressed.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
autumnust: It will fail the cancel method and propagate exception back to Helix. 

The reason of doing so, is because it is potentially dangerous to proceed in this case where the cancel was issue but the `taskAttempt` has not been initialized. For example it could result in two task running in the same helix instance in our streaming-ingestion pipeline where both task not falling behind due to the lack of enough resources

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -95,7 +108,16 @@ public void run()
...
+      // Signaling blocking threads if any whenever taskAttempt is nonNull.
+      _lock.lock();
+      try {
+        _taskAttempt = _taskAttemptBuilder.build(getWorkUnits().iterator(), _jobId, _jobState, jobBroker);
+        _taskAttemptBuilt.signal();
+      } finally {
+        _lock.unlock();
+      }
+
       _taskAttempt.runAndOptionallyCommitTaskAttempt(GobblinMultiTaskAttempt.CommitPolicy.IMMEDIATE);
```
ZihanLi58: why do we move this out of the lock? What will happen if _taskAttemp is initiated but not run, and cancel happen and then call _taskAttempt.runAndOptionallyCommitTaskAttempt()

autumnust: I think it is fine, the way that cancel works is, it call `shutdownTasks()` which actually sets a flag for each task, indicating the task has been terminated externally. This signal will be checked in extractor as it pull data from upstream, whenever it is being detected, it accomplish the ongoing loop of pulling data.  So the `run()` method within `runAndOptionallyCommitTaskAttempt` should still return. 

But I don't have test case to verify this though, I might be too aggressive for pruning critical section since it is not performance sensitive at all. Will put it back.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
```diff
@@ -93,8 +109,17 @@ public void run()
...
-      _taskAttempt = _taskAttemptBuilder.build(workUnits.iterator(), _jobId, _jobState, jobBroker);
-      _taskAttempt.runAndOptionallyCommitTaskAttempt(GobblinMultiTaskAttempt.CommitPolicy.IMMEDIATE);
+      // Secure atomicity of taskAttempt's execution.
+      // Signaling blocking threads if any whenever taskAttempt is nonNull.
+      _taskAttempt = _taskAttemptBuilder.build(getWorkUnits().iterator(), _jobId, _jobState, jobBroker);
+
+      _lock.lock();
+      try {
+        _taskAttemptBuilt.signal();
+        _taskAttempt.runAndOptionallyCommitTaskAttempt(GobblinMultiTaskAttempt.CommitPolicy.IMMEDIATE);
```
sv2000: Maybe I am missing something obvious - won't this block until the taskAttempt is finished? If so, is it right to hold the lock() until the attempt is finished? For instance, if a cancel() is invoked, won't the cancel() be waiting forever for the lock()? Alternately, if the cancel() acquires the lock() before the taskAttempt is built, then taskAttempt may be created after the cancel() fails.

sv2000:
@autumnust - On thinking more about the synchronization problem between GobblinMultiTaskAttempt#run() and cancel() methods, I think the following strategy might work. You can create a countdownLatch object that is shared between the run() and cancel() threads. When a cancel() is invoked, the countDownLatch is incremented. The cancel() may proceed to perform other actions but before it returns, it blocks until this latch is down to 0. The run() thread on the other hand, detects that the latch has been incremented, cancels the tasks submitted to the task executor, decrements the latch and returns. This should solve the synchronization issues between the threads without needing a lock/condition variable.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/SingleTask.java
ZihanLi58: I was mistakenly thinking runAndOptionallyCommitTaskAttempt will return immediately. Then how about move this method out and make the build of taskAttemp inside the lock block?

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -80,28 +83,49 @@ private HelixManager getHelixManager() {
 
   @Test
   void testJobShouldGetCancelled() throws Exception {
+    // Cancellation usually needs long time to successfully be executed, therefore seeing the sleeping time to 100.
```
sv2000, line 86: Typo: seeing -> setting

PR #2921, opened 2020-03-11
compile a flow before storing it in spec catalog

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @jack-moseley  @sv2000  please review

### JIRA
- [x] My PR addresses the following [GOBBLIN-1082](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-XXX] My Gobblin PR"

### Description
compile a flow , and if compilation is successful then only store it in spec catalog

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
@@ -34,7 +34,6 @@ import org.apache.gobblin.runtime.spec_catalog.AddSpecResponse;
 import org.apache.gobblin.runtime.spec_catalog.FlowCatalog;
 @Slf4j
 public class FlowConfigV2ResourceLocalHandler extends FlowConfigResourceLocalHandler implements FlowConfigsResourceHandler {
-  public static final String GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS = "org.apache.gobblin.service.modules.scheduler.GobblinServiceJobScheduler";
 
   public FlowConfigV2ResourceLocalHandler(FlowCatalog flowCatalog) {
     super(flowCatalog);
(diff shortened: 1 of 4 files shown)


gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
```diff
@@ -68,19 +67,27 @@ public CreateKVResponse createFlowConfig(FlowConfig flowConfig, boolean triggerL
...
       //Enrich original FlowConfig entity by adding the compiledFlow to the properties map.
       StringMap props = flowConfig.getProperties();
-      AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, null);
+      AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, null);
       props.put("gobblin.flow.compiled",
           addSpecResponse != null && addSpecResponse.getValue() != null ? StringEscapeUtils.escapeJson(addSpecResponse.getValue()) : "");
       flowConfig.setProperties(props);
+    }
+
+    if (flowConfig.hasExplain() && flowConfig.isExplain()) {
```
jack-moseley: I think you're ending an if block and then starting an identical one (`if (flowConfig.hasExplain() && flowConfig.isExplain())`) right after instead of just leaving it open here?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -296,27 +297,44 @@ public Spec getSpecWrapper(URI uri) {
...
+      responseMap.put(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("true"));
+    } else {
+      responseMap.put(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("false"));
     }
+
     return responseMap;
   }
 
+  private boolean isCompileSuccessful(Map<String, AddSpecResponse> responseMap) {
+    AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, new AddSpecResponse<>(""));
```
jack-moseley: Should this be `responseMap.getOrDefault(ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, null);` so that `isCompileSuccessful` returns false if it's not set (seems like it would return true now).

Actually where does the key `ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS` get set anyway, I couldn't figure it out by looking at the code.

arjun4084346: FlowCatalog::put inserts this
responseMap.put(entry.getKey().getName(), entry.getValue().getResult());
entry.getKey().getName() is 
default String getName() {
    return getClass().getName();
  }

arjun4084346: I had that null in the first commit, but then tests in FlowCatalogTest started failing because tests do not set any listeners.
So I changed the logic to categorized it as "compiledSuccessfully" if GobblinServiceJobScheduler listener (it is a SpecCatalogListener) does not present.
What do you think?

jack-moseley: Hmm I guess that is fine. But is it possible for that key to be missing in actual GaaS?

arjun4084346: I think that is not possible, because GobblinServiceJobScheduler will always be on listeners list, otherwise jobs will not be scheduled (         "this.scheduledFlowSpecs.put(addedSpec.getUri().toString(), addedSpec);")

And if it is on listeners, it will add its entry in responseMap.

PR #2924, opened 2020-03-13
[GOBBLIN-1084] Refresh flowgraph when templates are modified

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1084

### Description

Currently, if edges fail to be added on startup, they will not be tried again even if the issue is fixed through modifying the templates. This change causes the entire flowgraph to be refreshed if any template is modified.

### Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
@@ -25,8 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReadWriteLock;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.fs.Path;
@@ -52,7 +51,6 @@ import org.apache.gobblin.service.modules.flowgraph.FlowEdgeFactory;
 import org.apache.gobblin.service.modules.flowgraph.FlowGraph;
 import org.apache.gobblin.service.modules.flowgraph.FlowGraphConfigurationKeys;
 import org.apache.gobblin.service.modules.template_catalog.FSFlowTemplateCatalog;
-import org.apache.gobblin.service.modules.template_catalog.ObservingFSFlowEdgeTemplateCatalog;
 import org.apache.gobblin.util.ConfigUtils;
 import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
 
@@ -96,15 +94,16 @@ public class GitFlowGraphMonitor extends GitMonitoringService {
   private final Map<URI, TopologySpec> topologySpecMap;
   private final Config emptyConfig = ConfigFactory.empty();
   private final CountDownLatch initComplete;
-  private final Lock lock = new ReentrantLock();
+  private final ReadWriteLock rwLock;
 
   public GitFlowGraphMonitor(Config config, Optional<? extends FSFlowTemplateCatalog> flowTemplateCatalog,
-      FlowGraph graph, Map<URI, TopologySpec> topologySpecMap, CountDownLatch initComplete) {
+      FlowGraph graph, Map<URI, TopologySpec> topologySpecMap, CountDownLatch initComplete, ReadWriteLock rwLock) {
     super(config.getConfig(GIT_FLOWGRAPH_MONITOR_PREFIX).withFallback(DEFAULT_FALLBACK));
     this.flowTemplateCatalog = flowTemplateCatalog;
     this.flowGraph = graph;
     this.topologySpecMap = topologySpecMap;
     this.initComplete = initComplete;
+    this.rwLock = rwLock;
   }
 
   /**
@@ -131,16 +130,15 @@ public class GitFlowGraphMonitor extends GitMonitoringService {
    */
   @Override
   void processGitConfigChanges() throws GitAPIException, IOException {
-    if (lock.tryLock()) {
-      try {
-        if (flowTemplateCatalog.isPresent() && flowTemplateCatalog.get().isShouldRefreshFlowGraph()) {
+    try {
+      this.rwLock.writeLock().lock();
+      if (flowTemplateCatalog.isPresent() && flowTemplateCatalog.get().isShouldRefreshFlowGraph()) {
           log.info("Change to template catalog detected, refreshing FlowGraph");
           this.gitRepo.initRepository();
           flowTemplateCatalog.get().setShouldRefreshFlowGraph(false);
-        }
-      } finally {
-        lock.unlock();
       }
+    } finally {
+        this.rwLock.writeLock().unlock();
     }
 
     List<DiffEntry> changes = this.gitRepo.getChanges();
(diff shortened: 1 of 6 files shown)


jack-moseley:
@sv2000 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -127,6 +128,14 @@ public boolean shouldPollGit() {
    */
   @Override
   void processGitConfigChanges() throws GitAPIException, IOException {
+    if (flowTemplateCatalog.isPresent() && (flowTemplateCatalog.get() instanceof ObservingFSFlowEdgeTemplateCatalog)) {
+      ObservingFSFlowEdgeTemplateCatalog catalog = (ObservingFSFlowEdgeTemplateCatalog) flowTemplateCatalog.get();
+      if (catalog.isShouldRefreshFlowGraph()) {
+        this.gitRepo.initRepository();
+        catalog.setShouldRefreshFlowGraph(false);
```
sv2000: shouldn't resetting this boolean variable be inside a lock?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -127,6 +128,14 @@ public boolean shouldPollGit() {
    */
   @Override
   void processGitConfigChanges() throws GitAPIException, IOException {
+    if (flowTemplateCatalog.isPresent() && (flowTemplateCatalog.get() instanceof ObservingFSFlowEdgeTemplateCatalog)) {
+      ObservingFSFlowEdgeTemplateCatalog catalog = (ObservingFSFlowEdgeTemplateCatalog) flowTemplateCatalog.get();
+      if (catalog.isShouldRefreshFlowGraph()) {
+        this.gitRepo.initRepository();
```
sv2000: Add a log line here indicating that we have detected updates in the template catalog and that we need to re-bootstrap the FlowGraph.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -127,6 +128,14 @@ public boolean shouldPollGit() {
    */
   @Override
   void processGitConfigChanges() throws GitAPIException, IOException {
+    if (flowTemplateCatalog.isPresent() && (flowTemplateCatalog.get() instanceof ObservingFSFlowEdgeTemplateCatalog)) {
```
sv2000: We shouldn't be doing this check on every processGitCOnfigChanges() call.  Can't we just have a method isShouldRefreshFlowGraph() in the parent Catalog class that returns false and have this method overriden in the observing catalog child class?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitMonitoringService.java
```diff
@@ -304,7 +304,7 @@ protected void shutDown() throws Exception {
      * @throws GitAPIException
      * @throws IOException
      */
-    private void initRepository() throws GitAPIException, IOException {
+    public void initRepository() throws GitAPIException, IOException {
```
sv2000: package private instead of public?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/ObservingFSFlowEdgeTemplateCatalog.java
```diff
@@ -49,6 +51,10 @@
   private Map<URI, List<JobTemplate>> jobTemplateMap = new ConcurrentHashMap<>();
   private ReadWriteLock rwLock;
 
+  @Getter
+  @Setter
+  private boolean shouldRefreshFlowGraph = false;
```
sv2000: See comment earlier about making access of this variable thread safe.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
jack-moseley: I didn't do this since refresh flow graph only makes sense for observing flow catalog, but yeah I guess doing instanceof is not great either so I'll change it.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/template_catalog/ObservingFSFlowEdgeTemplateCatalog.java
jack-moseley: Is it necessary to add another lock here? I've added a lock in the flowgraphmonitor (which is where it gets read/set to false), and there is already a lock in this class on when it's set to true.

sv2000: Sorry, I did not mean to add a lock at this assignment. But my comment was more to ensure that access to the variable is guarded by a lock.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -128,11 +131,15 @@ public boolean shouldPollGit() {
    */
   @Override
   void processGitConfigChanges() throws GitAPIException, IOException {
-    if (flowTemplateCatalog.isPresent() && (flowTemplateCatalog.get() instanceof ObservingFSFlowEdgeTemplateCatalog)) {
-      ObservingFSFlowEdgeTemplateCatalog catalog = (ObservingFSFlowEdgeTemplateCatalog) flowTemplateCatalog.get();
-      if (catalog.isShouldRefreshFlowGraph()) {
-        this.gitRepo.initRepository();
-        catalog.setShouldRefreshFlowGraph(false);
+    if (lock.tryLock()) {
```
sv2000: Shouldn't we be using the lock defined in the ObservingFlowCatalog here? Defining a new lock here is not useful since the GitFlowGraphMonitor is single threaded.

jack-moseley: Hmm okay that's what I was thinking but I'm a bit confused what the concurrency issue is then. I guess the problem is that this variable could be modified from both git monitor and catalog at the same time?

In that case it should work to modify the `setShouldRefreshFlowGraph` in catalog to acquire the same rwlock right?

sv2000: So here is the race condition I was thinking of: 
1. FlowCatalog has been refreshed so shouldRefreshFlowGraph is marked true. 
2. GitMonitoringService calls processGitConfigChanges() and at some point before this method is done, FlowCatalog observes another refresh and sets shouldRefreshFlowGraph flag to true. 
3. This flag is reset by GitMonitoringService to false at the end of processGitConfigChanges(). 

So the GitMonitoringService can miss the 2nd catalog refresh and fail to add edges. LMK, if this is still not clear.

jack-moseley: Makes sense, updated to use the same lock as the catalog.

PR #2925, opened 2020-03-13
[GOBBLIN-1080] Add configuration to add schema creation time in converter

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1080

### Description
Add configuration to add schema creation time in converter

### Tests
unit test 

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -309,6 +309,8 @@ public class ConfigurationKeys {
   /**
    * Converter configuration properties.
    */
+  public static final String CONVERTER_PRESERVE_SCHEMA_CREATION_TIME = "converter.preserve.schema.creation.time";
+  public static final Boolean DEFAULT_CONVERTER_PRESERVE_SCHEMA_CREATION_TIME = true;
   public static final String CONVERTER_AVRO_DATE_FORMAT = "converter.avro.date.format";
   public static final String CONVERTER_AVRO_DATE_TIMEZONE = "converter.avro.date.timezone";
   public static final String CONVERTER_AVRO_TIME_FORMAT = "converter.avro.time.format";
@@ -833,6 +835,7 @@ public class ConfigurationKeys {
   public static final String KAFKA_SCHEMA_REGISTRY_RETRY_TIMES = "kafka.schema.registry.retry.times";
   public static final String KAFKA_SCHEMA_REGISTRY_RETRY_INTERVAL_IN_MILLIS =
       "kafka.schema.registry.retry.interval.inMillis";
+  public static final String SCHEMA_CREATION_TIME_KEY = "CreatedOn";
 
   /**
    * Job execution info server and history store configuration properties.
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -309,8 +309,8 @@ public class ConfigurationKeys {
   /**
    * Converter configuration properties.
    */
-  public static final String CONVERTER_PRESERVE_SCHEMA_CREATION_TIME = "converter.preserve.schema.creation.time";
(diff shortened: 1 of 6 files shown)


gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -309,6 +309,8 @@
   /**
    * Converter configuration properties.
    */
+  public static final String CONVERTER_AVRO_INCLUDE_SCHEMA_CREATION_TIME = "converter.avro.include.schema.creation.time";
```
autumnust: Do we need to make it configurable ?

gobblin-core-base/src/main/java/org/apache/gobblin/converter/filter/AvroProjectionConverter.java
```diff
@@ -74,10 +74,15 @@ public AvroProjectionConverter init(WorkUnitState workUnit) {
...
   @Override
   public Schema convertSchema(Schema inputSchema, WorkUnitState workUnit) throws SchemaConversionException {
+    Schema outputSchema = inputSchema;
     if (this.fieldRemover.isPresent()) {
-      return this.fieldRemover.get().removeFields(inputSchema);
+      outputSchema = this.fieldRemover.get().removeFields(inputSchema);
     }
-    return inputSchema;
+    if(workUnit.getPropAsBoolean(ConfigurationKeys.CONVERTER_AVRO_INCLUDE_SCHEMA_CREATION_TIME,
+        ConfigurationKeys.DEFAULT_CONVERTER_AVRO_INCLUDE_SCHEMA_CREATION_TIME)) {
```
autumnust: what if a pipeline needs this but not using this converter?  shall we consider to add it into higher level ?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -833,6 +835,7 @@
   public static final String KAFKA_SCHEMA_REGISTRY_RETRY_TIMES = "kafka.schema.registry.retry.times";
   public static final String KAFKA_SCHEMA_REGISTRY_RETRY_INTERVAL_IN_MILLIS =
       "kafka.schema.registry.retry.interval.inMillis";
+  public static final String SCHEMA_CREATION_TIME_KEY = "CreatedOn";
```
autumnust: If this is supposed to be an internal attribute, you won't need it as public static but should be as close as possible to wherever you are using it. ( I am a bit opposed to adding anything more into ConfigurationKeys but looks like it is common practice that everyone is following.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
ZihanLi58: this is only a key name for the config value. Will rename it to avoid confuse.

gobblin-core-base/src/main/java/org/apache/gobblin/converter/filter/AvroProjectionConverter.java
ZihanLi58: my concern is that will make the change not backward compatible, and most converter will not erase the schema prop, how about your thoughts?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
ZihanLi58: Will address

gobblin-core-base/src/main/java/org/apache/gobblin/converter/filter/AvroProjectionConverter.java
autumnust: Given the timestamp is set in the properties of an Avro schema object, does that matter for other writers that are not aware of the existence of this property ? We can try with a simple unit test to see if setting properties will impact writer. 

I mostly concern on the operability problem since there's no guarantee that this converter is and will used everywhere.

PR #2928, opened 2020-03-17
GOBBLIN-1087: Track and report histogram of observed lag from Gobblin…

sv2000:
… Kafka pipeline

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1087

### Description
In this PR, we instrument the KafkaExtractor to track the observed latency of Kafka consumer records processed by the pipeline. Here, observed latency is measured as the time difference between processing time of the record and the original creation time. The latency distribution is tracked in an HdrHistogram, which is serialized into a string when emitted as part of a GobblinTrackingEvent.

### Tests
Added unit tests and a JMH benchmarking test.

Sample results from running micro-benchmarks on a Macbook with a 3.5GHz Dual core Intel Core i7:
Estimated memory footprint of histogram is: 147968

Benchmark                                                             Mode  Cnt  Score    Error  Units
HdrHistogramPerformanceBenchmark.trackBuildHistogram                  avgt   10  0.007 ±  0.001  ms/op
HdrHistogramPerformanceBenchmark.trackHistogram10MToStringConversion  avgt   10  1.000 ±  0.054  ms/op
HdrHistogramPerformanceBenchmark.trackHistogram1MToStringConversion   avgt   10  0.930 ±  0.525  ms/op
HdrHistogramPerformanceBenchmark.trackHistogram2MToStringConversion   avgt   10  0.823 ±  0.065  ms/op
HdrHistogramPerformanceBenchmark.trackHistogram4MToStringConversion   avgt   10  0.850 ±  0.037  ms/op
HdrHistogramPerformanceBenchmark.trackMergeHistogram                  avgt   10  0.181 ±  0.028  ms/op
HdrHistogramPerformanceBenchmark.trackResetHistogram                  avgt   10  0.002 ±  0.001  ms/op

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/build.gradle
+++ b/gobblin-modules/gobblin-kafka-common/build.gradle
@@ -47,8 +47,6 @@ dependencies {
   testCompile externalDependency.mockito
   testCompile externalDependency.testng
   testCompile externalDependency.jmh
-
-  //jmh group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.17.3'
 }
 
 configurations {
--- a/gobblin-modules/gobblin-kafka-common/build.gradle
+++ b/gobblin-modules/gobblin-kafka-common/build.gradle
@@ -46,9 +46,9 @@ dependencies {
   testCompile project(":gobblin-test-utils")
   testCompile externalDependency.mockito
   testCompile externalDependency.testng
+  testCompile externalDependency.jmh
 
-  jmh 'org.openjdk.jmh:jmh-core:1.17.3'
-  jmh group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.17.3'
+  //jmh group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.17.3'
 }
 
 configurations {
--- a/gobblin-modules/gobblin-kafka-common/build.gradle
+++ b/gobblin-modules/gobblin-kafka-common/build.gradle
@@ -16,6 +16,7 @@
  */
 
 apply plugin: 'java'
+apply plugin: 'me.champeau.gradle.jmh'
 
 dependencies {
   compile project(":gobblin-api")
@@ -33,6 +34,7 @@ dependencies {
   compile externalDependency.commonsPool
   compile externalDependency.guava
   compile externalDependency.gson
+  compile externalDependency.hdrHistogram
   compile externalDependency.jacksonCore
   compile externalDependency.jacksonMapper
   compile externalDependency.slf4j
@@ -44,6 +46,9 @@ dependencies {
   testCompile project(":gobblin-test-utils")
   testCompile externalDependency.mockito
   testCompile externalDependency.testng
+
+  jmh 'org.openjdk.jmh:jmh-core:1.17.3'
+  jmh group: 'org.openjdk.jmh', name: 'jmh-generator-annprocess', version: '1.17.3'
 }
 
 configurations {
@@ -57,6 +62,12 @@ configurations {
 
 test {
   workingDir rootProject.rootDir
+}
+
+jmh {
+  include = ""
+  zip64 = true
+  duplicateClassesStrategy = "EXCLUDE"
 }
 
 ext.classification="library"
--- a/gobblin-modules/gobblin-kafka-common/src/jmh/java/org/apache/gobblin/source/extractor/extract/kafka/HdrHistogramPerformanceBenchmark.java
+++ b/gobblin-modules/gobblin-kafka-common/src/jmh/java/org/apache/gobblin/source/extractor/extract/kafka/HdrHistogramPerformanceBenchmark.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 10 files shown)


sv2000:
@autumnust @ZihanLi58 please review.

gobblin-modules/gobblin-kafka-common/src/jmh/java/org/apache/gobblin/source/extractor/extract/kafka/HdrHistogramPerformanceBenchmark.java
```diff
@@ -0,0 +1,130 @@
...
+    private Histogram histogram4;
+
+    private final RandomDataGenerator random = new RandomDataGenerator();
+
+    @Setup (value = Level.Iteration)
+    public void setUp() {
+      this.histogram1 = buildHistogram(1000000);
+      this.histogram2 = buildHistogram(2000000);
+      this.histogram3 = buildHistogram(4000000);
+      this.histogram4 = buildHistogram(10000000);
```
autumnust: Within our flush time-window, what's the magnitude of #events? Curious on the number ensuring that we are indeed doing stress test here.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractorStatsTracker.java
```diff
@@ -398,5 +480,8 @@ public void reset() {
     for (int partitionIdx = 0; partitionIdx < this.partitions.size(); partitionIdx++) {
       resetStartFetchEpochTime(partitionIdx);
     }
+    if (this.observedLagHistogram != null) {
+      this.observedLagHistogram.reset();
```
autumnust: Is the `reset` method reclaim memory ? Would it be safer if destroy the object and re-create next time if reclaim is not happening ?

gobblin-modules/gobblin-kafka-common/src/jmh/java/org/apache/gobblin/source/extractor/extract/kafka/HdrHistogramPerformanceBenchmark.java
sv2000: There are 2 ways to address this question:
1. The Histogram performance itself (e.g. adding a value or memory size) are independent of number of samples. They are only dependent on the range of values and the precision needed to report a value.  The benchmark numbers indeed confirm that the size is indeed constant across different counts and is around 150 KB. The benchmark also confirms that the serialization constants are also more or less independent of number of samples observed. 
2. Assuming a typical container processing rates in the order of 10-20 MB/s. Assuming records of size 1K, which corresponds to 10-20K records/second, a histogram should hold counts for 3-6M records within a flush interval of 5 minutes. 

I have updated the description of this PR to include results from running the benchmark on a Macbook.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractorStatsTracker.java
sv2000: Added benchmark to compare reset vs a new Histogram creation. While both reset and new are cheap, reset is 3x cheaper than new object creation. reset simply keeps the allocated count array as is and zeroes out the array. In general, it would be better to avoid new object creation to avoid GCs/memory fragmentation that can occur over time.

gobblin-modules/gobblin-kafka-common/build.gradle
```diff
@@ -44,6 +46,9 @@ dependencies {
   testCompile project(":gobblin-test-utils")
   testCompile externalDependency.mockito
   testCompile externalDependency.testng
+
+  jmh 'org.openjdk.jmh:jmh-core:1.17.3'
```
shirshanka: Now that we seem to have multiple modules using jmh, can you move out the version definition to common place like definitions.gradle and refer to it like we refer to them normally?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractorStatsTracker.java
autumnust: +1

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractorStatsTracker.java
```diff
@@ -161,14 +208,24 @@ public void resetStartFetchEpochTime(int partitionIdx) {
...
-  public void onDecodeableRecord(int partitionIdx, long readStartTime, long decodeStartTime, long recordSizeInBytes, long logAppendTimestamp) {
+  public void onDecodeableRecord(int partitionIdx, long readStartTime, long decodeStartTime, long recordSizeInBytes, long logAppendTimestamp, long recordCreationTimestamp) {
     this.statsMap.computeIfPresent(this.partitions.get(partitionIdx), (k, v) -> {
       long currentTime = System.nanoTime();
       v.processedRecordCount++;
       v.partitionTotalSize += recordSizeInBytes;
       v.decodeRecordTime += currentTime - decodeStartTime;
       v.readRecordTime += currentTime - readStartTime;
+      if (this.observedLagHistogram != null && recordCreationTimestamp > 0) {
+        long observedLag = System.currentTimeMillis() - recordCreationTimestamp;
```
autumnust: Shall we use "latency" instead since "lag" internally refers to number of events behind the head of stream ?  Maybe you can come up with better words, it is just my intuition of "lag" refers to something else.

PR #2940, opened 2020-03-26
GOBBLIN-1099: Handle orphaned Yarn containers in Gobblin-on-Yarn clus…

sv2000:
…ters

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1099

### Description
A Yarn application may leave behind orphaned containers, which can happen due to lost node managers. The orphaned containers however can continue to run (potentially forever) as participants in the Helix cluster. This can cause the following problems for a Gobblin-on-Yarn application:

Double publish of data and commit of state
Task failures and partition starvation during application restarts, as Helix may assign tasks to the orphaned containers which have a stale state and configuration
Container failures on application restarts due to Helix instance name collisions with orphaned containers

This PR incorporates the following changes to handle orphaned containers:
1. Disables live instances during Yarn application start up and shutdown in GobblinYarnAppLauncher. This ensures that any orphaned Helix instances are disabled from joining the cluster on restart and any tasks running on these instances are cancelled.
2. The GobblinApplicationMaster (inside YarnService) ensures that Helix instance name assigned to a newly allocated container is not a currently live instance to avoid instance name collisions. We also handle NM failures/restarts that result in Yarn RM "aborting" the container (i.e. container is deemed dead from Yarn's point of view, even though the container may physically be alive). In this case, we disable the instance to ensure it is fenced off from the Helix cluster.
3. Gobblin workers (i.e. GobblinYarnTaskRunner) retry in case of failure in joining a Helix cluster. The retry logic disconnects from the Helix cluster and drops the instance before reattempting to join the cluster. 

### Tests
Added unit test in GobblinTaskRunnerTest

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -42,12 +44,15 @@ import org.apache.commons.cli.ParseException;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.HelixProperty;
 import org.apache.helix.InstanceType;
 import org.apache.helix.NotificationContext;
+import org.apache.helix.manager.zk.ZKHelixAdmin;
 import org.apache.helix.messaging.handling.HelixTaskResult;
 import org.apache.helix.messaging.handling.MessageHandler;
 import org.apache.helix.messaging.handling.MessageHandlerFactory;
@@ -57,6 +62,10 @@ import org.apache.helix.task.TaskStateModelFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.Retryer;
+import com.github.rholder.retry.RetryerBuilder;
+import com.github.rholder.retry.StopStrategies;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
@@ -81,6 +90,7 @@ import org.apache.gobblin.util.FileUtils;
 import org.apache.gobblin.util.HadoopUtils;
 import org.apache.gobblin.util.JvmUtils;
 import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
+
 
 /**
  * The main class running in the containers managing services for running Gobblin
@@ -121,6 +131,7 @@ public class GobblinTaskRunner implements StandardMetricsBridge {
 
   private final String clusterName;
 
+  @Getter
   private HelixManager jobHelixManager;
 
   private Optional<HelixManager> taskDriverHelixManager = Optional.absent();
@@ -292,7 +303,7 @@ public class GobblinTaskRunner implements StandardMetricsBridge {
     // Add a shutdown hook so the task scheduler gets properly shutdown
     addShutdownHook();
 
-    connectHelixManager();
+    connectHelixManagerWithRetry();
 
     addInstanceTags();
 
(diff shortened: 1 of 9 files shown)


sv2000:
@autumnust @htran1 please review.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -82,6 +91,7 @@
 import org.apache.gobblin.util.JvmUtils;
 import org.apache.gobblin.util.reflection.GobblinConstructorUtils;
 
+
```
autumnust: Empty line.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -366,20 +377,74 @@ boolean isStopped() {
...
+   *   <ul>Disconnect from Helix cluster, which would close any open clients</ul>
+   *   <ul>Drop instance from Helix cluster, to remove any partial instance structure from Helix</ul>
+   *   <ul>Re-construct helix manager instances, used to re-join the cluster</ul>
+   * </li>
+   */
+  private void onClusterJoinFailure() {
+    logger.warn("Disconnecting Helix manager..");
+    disconnectHelixManager();
+
+    HelixAdmin admin = new ZKHelixAdmin(clusterConfig.getString(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY));
```
autumnust: I couldn't fully get the purpose of disconnect from existing helixManager, creating a HelixAdmin and use the newly-created admin to conduct instance-dropping. Isn't `helixManager.getClusterManagmentTool()` essentially a same HelixAdmin that we can use to drop instance ? What's the gain of disconnecting helixManager?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -366,20 +377,74 @@ boolean isStopped() {
...
+    };
+
+    Retryer<Void> retryer = RetryerBuilder.<Void>newBuilder()
+        .retryIfException()
+        .withStopStrategy(StopStrategies.stopAfterAttempt(5)).build();
+
+    try {
+      retryer.call(connectHelixManagerCallable);
+    } catch (ExecutionException | RetryException e) {
+      logger.error("Connecting to Helix manager failed", e);
```
autumnust: let's avoid log.error and having an unchecked exception propagated at the same time: It doubled stack traces which is not very necessary and confused reader on how many times the exception actually happen.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
```diff
@@ -366,20 +377,74 @@ boolean isStopped() {
...
+   * </li>
+   */
+  private void onClusterJoinFailure() {
+    logger.warn("Disconnecting Helix manager..");
+    disconnectHelixManager();
+
+    HelixAdmin admin = new ZKHelixAdmin(clusterConfig.getString(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY));
+    //Drop the helix Instance
+    logger.warn("Dropping instance: {} from cluster: {}", helixInstanceName, clusterName);
+    HelixUtils.dropInstanceIfExists(admin, clusterName, helixInstanceName);
```
autumnust: The return value of this method is ignored, is that intentional? Or shall we leave the handling of exception to caller of this static method ?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -625,7 +634,7 @@ private boolean shouldStickToTheSameNode(int containerExitStatus) {
    */
   protected void handleContainerCompletion(ContainerStatus containerStatus) {
     Map.Entry<Container, String> completedContainerEntry = this.containerMap.remove(containerStatus.getContainerId());
-    String completedInstanceName = completedContainerEntry == null? "unknown" : completedContainerEntry.getValue();
+    String completedInstanceName = completedContainerEntry == null?  UNKNOWN_HELIX_INSTANCE : completedContainerEntry.getValue();
```
autumnust: Just curious, What kind of situation result in a containerId is not in containerMap? can we also add javadoc here

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -635,10 +644,26 @@ protected void handleContainerCompletion(ContainerStatus containerStatus) {
...
+       // In this case, RM assumes that the containers are "lost", even though the container process may still be
+        // running on the node. We need to ensure that the Helix instances running on the orphaned containers
+        // are fenced off from the Helix cluster to avoid double publishing and state being committed by the
+        // instances.
+        if (!UNKNOWN_HELIX_INSTANCE.equals(completedInstanceName)) {
+          String clusterName = this.helixManager.getClusterName();
+          //Disable the orphaned instance.
+          if (HelixUtils.isInstanceLive(helixManager, completedInstanceName)) {
+            LOGGER.info("Disabling the Helix instance {}", completedInstanceName);
+            this.helixManager.getClusterManagmentTool().enableInstance(clusterName, completedInstanceName, false);
```
autumnust: enableInstance, This is a very good API :)

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
```diff
@@ -188,6 +193,7 @@
   // into the queue if the container running the instance completes. Unused Helix
   // instance names get picked up when replacement containers get allocated.
   private final ConcurrentLinkedQueue<String> unusedHelixInstanceNames = Queues.newConcurrentLinkedQueue();
+  private boolean reuseUnusedHelixInstanceNames = true;
```
autumnust: what's the usage of this?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -479,6 +488,26 @@ void connectHelixManager() {
...
+   * completion, this method should be removed.
+   */
+  void disableLiveHelixInstances() {
+    String clusterName = this.helixManager.getClusterName();
+    HelixAdmin helixAdmin = this.helixManager.getClusterManagmentTool();
+    List<String> liveInstances = HelixUtils.getLiveInstances(this.helixManager);
+    LOGGER.warn("Found {} live instances in the cluster.", liveInstances.size());
+    for (String instanceName: liveInstances) {
+      LOGGER.warn("Disabling instance: {}", instanceName);
+      helixAdmin.enableInstance(clusterName, instanceName, false);
```
autumnust: Are there any mechanism from Helix side to sort of retry-joining cluster after being disabled? Just want to make sure

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
sv2000: The intent was to avoid a state where the HelixManager is in an inconsistent state where the underlying ZkClient is connected but there is a failure later. By disconnecting before retrying, we ensure any partial state during the previous connect attempt is cleaned up. 

The reason we are instantiating a new HelixAdmin instance (separate from HelixManager) is that getClusterManagementTool() requires HelixManager to be connected first.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
sv2000: Thanks! Removed this variable.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
sv2000: No there is none. The instance is part of the cluster, but disabled. It needs to re-enabled to start getting task assignments from Helix.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
sv2000: Thanks! Will fix it.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
sv2000: Fixed.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnService.java
sv2000: Added a comment.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinTaskRunner.java
sv2000: Changed it to return void. The dropInstanceIfExists method is intended to swallow HelixException that can only occur due to instance path not being present i.e instance does not exist.

PR #2941, opened 2020-03-27
[GOBBLIN-1100] Set average fetch time in the KafkaExtractor even when…

htran1:
… metrics are disabled

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1100

### Description
The KafkaExtractor close method only calls the KafkaExtractorStatsTracker#emitTrackingEvents method when metrics instrumentation is enabled. This method calls KafkaExtractorStatsTracker#generateTagsForPartitions, which has the side effect of setting the average fetch time in the WorkUnitState. The average fetch time is required for the work unit packing to pack optimally.

Add a call to KafkaExtractorStatsTracker#generateTagsForPartitions in KafkaExtractor#close when metrics instrumentation is disabled to restore the behavior that existed prior to the KafkaExtractorStatsTracker refactoring.

### Tests
KafkaExtractorStatsTrackerTest

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 
 import lombok.Getter;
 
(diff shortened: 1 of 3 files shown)


htran1:
@autumnust, @sv2000, please review.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaExtractor.java
```diff
@@ -321,6 +322,11 @@ public void close() throws IOException {
     this.workUnitState.setActualHighWatermark(this.nextWatermark);
     if (isInstrumentationEnabled()) {
       this.statsTracker.emitTrackingEvents(getMetricContext(), this.lowWatermark, this.highWatermark, this.nextWatermark);
+    } else {
```
sv2000: Should we move the isInstrumentationEnabled() check inside emitTrackingEventsWithAdditionalTags() and emit tracking events if it is true?

htran1: I moved the generateTagsForPartitions() call out so that emitTrackingEvents is only for emitting the events.

PR #2942, opened 2020-03-30
GOBBLIN-1101(DSS-25241): Enhance bulk api retry for ExceedQuota

arekusuri:
### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1101

### Description
1. ExceedQuota exception
If the ExceedQuota exception happens, we should let the thread sleep 5 minutes and try again. There should not be a retryLimit for this exception.
2. Except stack in log file
For example we set up retryLimit to 10, we retried 10 times,  and failed; we need to print out exception stack in log file, there are 10 of them in the exception stack. We'd better skip all the retry exception, only keep the root cause exception.

### Tests
https://ltx1-holdemaz05.grid.linkedin.com:8443/executor?execid=25014730&job=salesforce_task_full&attempt=0

### DIFF
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
@@ -107,8 +107,9 @@ public class BulkResultIterator implements Iterator<JsonElement> {
   private void threadSleep(long millis) {
     try {
       Thread.sleep(millis);
-    } catch (Exception ee) {
-      log.warn("--sleep exception--");
+    } catch (Exception e) {
+      log.error("--Failed to sleep--", e);
+      throw new RuntimeException(e);
     }
   }
 
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
@@ -72,7 +72,7 @@ public class BulkResultIterator implements Iterator<JsonElement> {
     Throwable rootCause = null;
     int executeCount = 0;
     while (executeCount < retryLimit + 1) {
-      executeCount ++;
+      executeCount++;
       try {
         if (this.csvReader == null) {
           this.csvReader = openAndSeekCsvReader(rootCause);
@@ -89,7 +89,7 @@ public class BulkResultIterator implements Iterator<JsonElement> {
         if (e.isCurrentExceptionExceedQuota()) {
           log.warn("--Caught ExceededQuota: " + e.getMessage());
           threadSleep(retryExceedQuotaInterval);
-          executeCount --; // if the current exception is Quota Exceeded, keep trying forever
+          executeCount--; // if the current exception is Quota Exceeded, keep trying forever
         }
         log.info("***Retrying***1: {} - {}", fileIdVO, e.getMessage());
         this.csvReader = null; // in next loop, call openAndSeekCsvReader
@@ -158,13 +158,13 @@ public class BulkResultIterator implements Iterator<JsonElement> {
       if ((lastSkippedLine == null && preLoadedLine != null) || (lastSkippedLine != null && !lastSkippedLine.equals(
           preLoadedLine))) {
         // check if last skipped line is same as the line before error
-        String msg = rootCause == null? "null" : rootCause.getMessage();
+        String msg = rootCause == null ? "null" : rootCause.getMessage();
         throw new OpenAndSeekException("Failed to verify last skipped line - root cause [" + msg + "]", rootCause);
       }
       return csvReader;
 
     } catch (Exception currentException) { // failed to open reader and skip lineCount lines // ssl failures go here
-      Throwable cause = rootCause == null? currentException : rootCause;
+      Throwable cause = rootCause == null ? currentException : rootCause;
       throw new OpenAndSeekException("Failed to [" + cause.getMessage() + "]" , cause, currentException);
     }
   }
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
@@ -135,8 +135,8 @@ public class BulkResultIterator implements Iterator<JsonElement> {
   }
 
   private InputStreamCSVReader openAndSeekCsvReader(Throwable rootCause) throws OpenAndSeekException {
-    while (rootCause != null) {
-      rootCause = rootCause.getCause();
+    while (rootCause != null && rootCause.getCause() != null) {
+      rootCause = rootCause.getCause(); // find the root cause
     }
     String jobId = fileIdVO.getJobId();
     String batchId = fileIdVO.getBatchId();
@@ -154,12 +154,13 @@ public class BulkResultIterator implements Iterator<JsonElement> {
       if ((lastSkippedLine == null && preLoadedLine != null) || (lastSkippedLine != null && !lastSkippedLine.equals(
           preLoadedLine))) {
         // check if last skipped line is same as the line before error
-        throw new OpenAndSeekException("Failed to verify last skipped line - root cause [" + rootCause.getMessage() + "]", rootCause);
+        String msg = rootCause == null? "null" : rootCause.getMessage();
+        throw new OpenAndSeekException("Failed to verify last skipped line - root cause [" + msg + "]", rootCause);
       }
       return csvReader;
-    } catch (Exception e) { // failed to open reader and skip lineCount lines // ssl failures go here
-      rootCause = rootCause != null? rootCause : e;
-      throw new OpenAndSeekException("Failed to [" + rootCause.getMessage() + "]" , rootCause, e);
+    } catch (Exception currentException) { // failed to open reader and skip lineCount lines // ssl failures go here
+      Throwable cause = rootCause == null? currentException : rootCause;
+      throw new OpenAndSeekException("Failed to [" + cause.getMessage() + "]" , cause, currentException);
     }
   }
 
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
@@ -18,6 +18,8 @@
 package org.apache.gobblin.salesforce;
 
 import com.google.gson.JsonElement;
+import com.sforce.async.AsyncApiException;
+import com.sforce.async.AsyncExceptionCode;
 import com.sforce.async.BulkConnection;
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -29,6 +31,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.source.extractor.utils.InputStreamCSVReader;
 import org.apache.gobblin.source.extractor.utils.Utils;
+
 
 /**
  * Iterator for fetching result file of Bulk API.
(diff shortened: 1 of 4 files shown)


gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +65,51 @@ private void initHeader() {
...
-        exception = e;
-        log.info("***Retrying***: {} - {}", fileIdVO, e.getMessage());
-        this.csvReader = openAndSeekCsvReader(e);
+      } catch (OpenAndSeekException e) {
+        rootCause = e.getCause();
+        // Each organization is allowed 10 concurrent long-running requests. If the limit is reached,
+        // any new synchronous Apex request results in a runtime exception.
+        if (e.isCurrentExceptionExceedQuta()) {
+          log.warn("--Caught ExceededQuota: " + e.getMessage());
+          threadSleep(5 * 60 * 1000); // 5 minutes
```
htran1: Please make the timeout configurable. Default can be set to 5 minutes.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +65,51 @@ private void initHeader() {
...
+          executeCount --; // if the current exception is Quota Exceeded, keep trying forever
```
htran1: Whether to retry forever should be configurable, with the default as false to match the existing behavior.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +65,51 @@ private void initHeader() {
...
+        }
+        log.info("***Retrying***1: {} - {}", fileIdVO, e.getMessage());
+        this.csvReader = null; // in next loop, call openAndSeekCsvReader
+      } catch (Exception e) {
+        // Retry may resolve other exceptions.
+        rootCause = e;
+        threadSleep(1 * 60 * 1000); // 1 minute
```
htran1: This timeout should also be configurable.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +65,51 @@ private void initHeader() {
...
+    } else {
+      throw new RuntimeException("***Retried***: Failed, tried " + retryLimit + " times - ", rootCause);
+    }
+  }
+
+  private void threadSleep(long millis) {
+    try {
+      Thread.sleep(millis);
+    } catch (Exception ee) {
+      log.warn("--sleep exception--");
```
htran1: Add the exception to the log message. You should reraise InterruptedException or shutdown of the process may be blocked.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -144,3 +178,24 @@ private void closeCsvReader() {
...
+    }
+  }
+  public OpenAndSeekException(String msg, Throwable rootCause, Exception currentException) {
+    super(msg, rootCause);
+    if (currentException instanceof AsyncApiException &&
+        ((AsyncApiException) currentException).getExceptionCode() == AsyncExceptionCode.ExceededQuota) {
+      _isCurrentExceptionExceedQuota = true;
+    }
+  }
+  public boolean isCurrentExceptionExceedQuta() {
```
htran1: Quta -> Quota

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
```diff
@@ -25,7 +25,7 @@ private SalesforceConfigurationKeys() {
   }
   public static final String SOURCE_QUERYBASED_SALESFORCE_IS_SOFT_DELETES_PULL_DISABLED =
       "source.querybased.salesforce.is.soft.deletes.pull.disabled";
-  public static final int DEFAULT_FETCH_RETRY_LIMIT = 5;
+  public static final int DEFAULT_FETCH_RETRY_LIMIT = 1;
```
htran1: Why change the default here? It may affect existing users.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: This argument is not critical. Because of resource is not sufficient, we need to wait a random duration. 1 minute is OK, 5 minutes is OK too, for it fails, continue to wait until succeeds. 
User would not need to know a new concept about this.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: It is just for avoid request peak. In practice, it will never be forever.  We'd better not let user learn this concept.
It is a resource management strategy.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: This is meant to be random duration as well. 
This kind of failure will be retried only limited times.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: changed to 
`    try {
      Thread.sleep(millis);
    } catch (Exception e) {
      log.error("--Failed to sleep--", e);
    }`

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: good catch! thanks! fixed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceConfigurationKeys.java
arekusuri: OK, let's restore it.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
htran1: This affects resource utilization and latency, so I think users may want to control that based on their workload and priorities. For example, they may want the ETL job to fail immediately and not retry or to have a longer polling interval to reduce contention with higher priority jobs.

arekusuri:
> This affects resource utilization and latency, so I think users may want to control that based on their workload and priorities. For example, they may want the ETL job to fail immediately and not retry or to have a longer polling interval to reduce contention with higher priority jobs.
@htran1 thanks for your above comment.
Add sleep in the retry logic is to avoid resource consuming peak. The sleep duration should be a random number. Nobody knows what is the right number to set up. (to do a lot of experiments may help find out a optimal number) 
It is not depending anything like table size.
Flow job developer would not be able to figure out what number to set up. Leave users the key can only confuse them more.
> they may want the ETL job to fail immediately
Very good question. We have key "salesforce.fetchRetryLimit", if set to 0, there won't be retry. Default retry limit is 5.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
htran1: It is not only about Gobblin ETL jobs. There can be other consumers of the resource that is throttled. Having a configuration key with default value does not make the user experience any worse and allows tuning when required.

arekusuri:
@htran1 OK, I will add a key for the duration.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +69,52 @@ private void initHeader() {
   }
 
   private List<String> nextLineWithRetry() {
-    Exception exception = null;
-    for (int i = 0; i < retryLimit; i++) {
+    Throwable rootCause = null;
+    int executeCount = 0;
+    while (executeCount < retryLimit + 1) {
+      executeCount ++;
```
htran1, line 75: Please remove the space before ++.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -117,14 +155,17 @@ private InputStreamCSVReader openAndSeekCsvReader(Exception exceptionRetryFor) {
       for (int j = 0; j < lineCount; j++) {
         lastSkippedLine = csvReader.nextRecord(); // skip these records
       }
-      if ((lastSkippedLine == null && preLoadedLine != null) || (lastSkippedLine != null && !lastSkippedLine.equals(preLoadedLine))) {
+      if ((lastSkippedLine == null && preLoadedLine != null) || (lastSkippedLine != null && !lastSkippedLine.equals(
+          preLoadedLine))) {
         // check if last skipped line is same as the line before error
-        throw new RuntimeException("Failed to verify last skipped line - retrying for =>", exceptionRetryFor);
+        String msg = rootCause == null? "null" : rootCause.getMessage();
```
htran1, line 161: Add space between null and ?.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -117,14 +155,17 @@ private InputStreamCSVReader openAndSeekCsvReader(Exception exceptionRetryFor) {
...
+        throw new OpenAndSeekException("Failed to verify last skipped line - root cause [" + msg + "]", rootCause);
       }
       return csvReader;
-    } catch (Exception e) { // failed to open reader and skip lineCount lines
-      exceptionRetryFor = exceptionRetryFor != null? exceptionRetryFor : e;
-      throw new RuntimeException("Failed to [" + e.getMessage() + "] - retrying for => " , exceptionRetryFor);
+
+    } catch (Exception currentException) { // failed to open reader and skip lineCount lines // ssl failures go here
+      Throwable cause = rootCause == null? currentException : rootCause;
```
htran1, line 167: Add space between null and ?.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
```diff
@@ -62,24 +69,52 @@ private void initHeader() {
...
-        log.info("***Retrying***: {} - {}", fileIdVO, e.getMessage());
-        this.csvReader = openAndSeekCsvReader(e);
+      } catch (OpenAndSeekException e) {
+        rootCause = e.getCause();
+        // Each organization is allowed 10 concurrent long-running requests. If the limit is reached,
+        // any new synchronous Apex request results in a runtime exception.
+        if (e.isCurrentExceptionExceedQuota()) {
+          log.warn("--Caught ExceededQuota: " + e.getMessage());
+          threadSleep(retryExceedQuotaInterval);
+          executeCount --; // if the current exception is Quota Exceeded, keep trying forever
```
htran1, line 92: Remove space before --.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: fixed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: fixed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: fixed.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/BulkResultIterator.java
arekusuri: Fixed. They are all good catch! thanks

PR #2957, opened 2020-04-14
[GOBBLIN-1117]Enable record count verification for ORC format

autumnust:
Dear Gobblin maintainers,

### JIRA
-  https://issues.apache.org/jira/browse/[GOBBLIN-1117]

### Description
This PR fixed the hard-code suffix as "avro" in `InputRecordCountHelper` and add a unit test for re-compaction in ORC. 

There's formatting happening unfortunately in `CompactionThresholdVerifier.java` which might make review a little bit hard. But the only useful line out there is line 89 to check if `newCount` is larger than the `oldCount`. 

### Tests

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
@@ -85,11 +85,11 @@ public class CompactionThresholdVerifier implements CompactionVerifier<FileSyste
       if (oldRecords == 0) {
         return new Result(true, "");
       }
-      if ((newRecords - oldRecords) / oldRecords > threshold) {
-        if (newRecords < oldRecords) {
-          return new Result(false, "Illegal state: Current records count should old be smaller.");
-        }
+      if (newRecords < oldRecords) {
+        return new Result(false, "Illegal state: Current records count should old be smaller.");
+      }
 
+      if ((newRecords - oldRecords) / oldRecords > threshold) {
         log.debug("Dataset {} records exceeded the threshold {}", dataset.datasetURN(), threshold);
         return new Result(true, "");
       }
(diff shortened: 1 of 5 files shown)


autumnust:
@yukuai518  Can you help me take a look at this PR? Thanks.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
```diff
@@ -62,34 +62,41 @@ public CompactionThresholdVerifier(State state) {
...
       }
-      double oldRecords = helper.readRecordCount (new Path(result.getDstAbsoluteDir()));
+      double oldRecords = helper.readRecordCount(new Path(result.getDstAbsoluteDir()));
 
       if (oldRecords == 0) {
         return new Result(true, "");
       }
       if ((newRecords - oldRecords) / oldRecords > threshold) {
-        log.debug ("Dataset {} records exceeded the threshold {}", dataset.datasetURN(), threshold);
+        if (newRecords < oldRecords) {
```
zxcware, line 89: put outside? otherwise, `(newRecords - oldRecords) / oldRecords` is a negative value and will be always smaller than a positive threshold.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/InputRecordCountHelper.java
```diff
@@ -71,6 +75,7 @@ public InputRecordCountHelper(State state) {
     try {
       this.fs = getSourceFileSystem (state);
       this.state = state;
+      this.extensionName = state.getProp(COMPACTION_OUTPUT_EXTENSION, DEFAULT_COMPACTION_OUTPUT_EXTENSION);
```
zxcware, line 78: looks like it supports a list of extension as in line 97. How about using a new config key that indicates it supports a extensions list? It also helps decouple the configs between `InputRecordCountHelper` and `CompactorOutputCommitter`.

autumnust: I think making this configurable in the first place is a mistake - the extension should always stay consistent with the extension that you specified as part of compaction job. I get the point of decouple the configs, but they are needed in different places and that's why it is public static. ( If strict decoupling is applied then anything should be private or package-private static, which is not realistic

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
autumnust: Will address.

yukuai518:
+1 LGTM

PR #2960, opened 2020-04-16
GOBBLIN-1120: Reinitialize HelixManager when Helix participant check …

sv2000:
…throws an exception

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1120

### Description
When the underlying ZkClient is closed, Helix participant check throws:

Caused by: java.lang.IllegalStateException: ZkClient already closed!

In these cases, the HelixManager instance used to do the participant check needs to be re-initialized to obtain a new ZKClient. 

### Tests
Enhanced the unit tests in HelixAssignedParticipantCheckTest.

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
@@ -16,7 +16,6 @@
  */
 package org.apache.gobblin.cluster;
 
-import java.io.IOException;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -32,8 +31,10 @@ import com.github.rholder.retry.RetryException;
 import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.StopStrategies;
+import com.google.common.annotations.VisibleForTesting;
 import com.typesafe.config.Config;
 
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.gobblin.annotation.Alias;
@@ -52,6 +53,8 @@ import org.apache.gobblin.commit.CommitStepException;
 @Slf4j
 @Alias (value = "HelixParticipantCheck")
 public class HelixAssignedParticipantCheck implements CommitStep {
+  @Getter
+  @VisibleForTesting
   private static volatile HelixManager helixManager = null;
   private static volatile Retryer<Boolean> retryer = RetryerBuilder.<Boolean>newBuilder()
       .retryIfException()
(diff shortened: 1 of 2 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
```diff
@@ -52,6 +53,8 @@
 @Slf4j
 @Alias (value = "HelixParticipantCheck")
 public class HelixAssignedParticipantCheck implements CommitStep {
+  @Getter
+  @VisibleForTesting
```
autumnust, line 57: Is this variable visible ? ( private static

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
```diff
@@ -104,20 +122,23 @@ public boolean isCompleted()
...
         this.helixInstanceName, this.helixJob, this.partitionNum));
 
     //Query Helix to get the currently assigned participant for the Helix partitionNum
     Callable callable = () -> {
-      JobContext jobContext = taskDriver.getJobContext(helixJob);
+      JobContext jobContext;
+      try {
+        TaskDriver taskDriver = new TaskDriver(helixManager);
+        jobContext = taskDriver.getJobContext(helixJob);
+      } catch (Exception e) {
```
autumnust, line 134: What's the exact type of exception here? Shall we only catch that type

autumnust: and just curious, why will there be a zk connection lost in the first place ?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
sv2000: The variable is accessible via Getter in the Test class. The annotation is to indicate that the Getter is added for access inside the test class.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheck.java
sv2000: Helix throws an IllegalStateException (which is a RuntimeException) when the underlying ZkClient is closed. Since it is an unchecked exception, I added a clause to catch the generic Exception. 
It is hard to determine the exact cause why the zkclient is closed in the first place. One possibility is a temporary N/W glitch which might cause the container to loose connectivity with Zk. Another possibility is a Zk session expiry, which if not properly handled by the underlying client (i.e does not restablish a connection with server) may cause the client to be closed. The logic here is to ensure that we refresh the Helix manager in these scenarios.

PR #2969, opened 2020-04-27
[GOBBLIN-1132] move the logic of requester list verification to RequesterService implementation

arjun4084346:
codestyle changes

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1132

### Description

Currently, it checks “all current requesters must be in original list”, and this PR is changing it to be implemented in the implementation of RequesterService

made some codestyle changes and dependency upgrades

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsResource.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsResource.java
@@ -109,7 +109,7 @@ public class FlowConfigsResource extends ComplexKeyResourceTemplate<FlowId, Empt
    */
   @Override
   public UpdateResponse update(ComplexResourceKey<FlowId, EmptyRecord> key, FlowConfig flowConfig) {
-    checkRequester(get(key), this.requesterService.findRequesters(this));
+    checkRequester(this.requesterService, get(key), this.requesterService.findRequesters(this));
     String flowGroup = key.getKey().getFlowGroup();
     String flowName = key.getKey().getFlowName();
     FlowId flowId = new FlowId().setFlowGroup(flowGroup).setFlowName(flowName);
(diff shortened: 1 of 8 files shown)


gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsResource.java
```diff
@@ -140,7 +139,7 @@ public UpdateResponse delete(ComplexResourceKey<FlowId, EmptyRecord> key) {
    * @param originalFlowConfig original flow config to find original requester
    * @param requesterList list of requesters for this request
    */
-  public static void checkRequester(FlowConfig originalFlowConfig, List<ServiceRequester> requesterList) {
```
jack-moseley, line 143: Why not leave this as static to avoid code duplication? You could move it to FlowConfigsV2Resource if you're considering that FlowConfigsResource may be removed eventually

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/RequesterService.java
```diff
@@ -50,22 +55,27 @@ public RequesterService(Config config) {
...
-    return requesterList;
+    return objectMapper.readValue(jsonList, mapType);
   }
 
   protected abstract List<ServiceRequester> findRequesters(BaseResource resource);
+
+  /**
+   * This method should throw {@link FlowConfigLoggedException}, if the requester is not allowed to make this request.
+   */
+  protected abstract void requesterAllowed(List<ServiceRequester> originalRequesterList,
```
jack-moseley, line 79: Maybe call this isRequesterAllowed and make it return boolean?

I know I made the original checkRequester throw exception, but this one's only called from one place and is abstract, so it seems better to have a more clear contract of returning boolean.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsResource.java
arjun4084346: Had to change it, coz using non static object (this.requesterService) inside.
Will work actively towards deprecating V1

jack-moseley: You could pass the requesterservice as an argument. But sure, I guess having the duplication is fine if we're probably not going to make changes to v1 after this.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/RequesterService.java
arjun4084346: How about now?

PR #2981, opened 2020-05-06
[GOBBLIN-1144] remove specs from gobblin service job scheduler

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1144

### Description
implement option 4 mentioned in the doc https://docs.google.com/document/d/1OsImllAZRnJIp2NWEOdlfw0XtqY1b-ysyKEZYLHwVbQ/edit

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
@@ -21,6 +21,8 @@ import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.typesafe.config.Config;
+import com.typesafe.config.ConfigException;
+
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
@@ -33,6 +35,7 @@ import java.util.Map;
 import java.util.Properties;
 import javax.annotation.Nonnull;
 import org.apache.commons.lang3.reflect.ConstructorUtils;
+
 import org.apache.gobblin.instrumented.Instrumented;
 import org.apache.gobblin.metrics.MetricContext;
 import org.apache.gobblin.metrics.Tag;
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -293,30 +296,27 @@ public Spec getSpecWrapper(URI uri) {
...
+      long startTime = System.currentTimeMillis();
+      specStore.addSpec(flowSpec);
+      metrics.updatePutSpecTime(startTime);
+    } catch (IOException e) {
+      throw new RuntimeException("Cannot add Spec to Spec store: " + flowSpec, e);
+    }
 
-    log.info(String.format("Adding FlowSpec with URI: %s and Config: %s", spec.getUri(),
-        ((FlowSpec) spec).getConfigAsProperties()));
     if (triggerListener) {
```
enjoyear, line 312: Can you move this if section into the try block like before?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -293,30 +296,27 @@ public Spec getSpecWrapper(URI uri) {
...
-        if (!((FlowSpec) spec).isExplain()) {
-          specStore.addSpec(spec);
-        }
-      } catch (IOException e) {
-        throw new RuntimeException("Cannot add Spec to Spec store: " + spec, e);
-      }
+    if (isCompileSuccessful(responseMap)) {
       responseMap.put(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("true"));
     } else {
       responseMap.put(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("false"));
```
enjoyear, line 322: Where is your logic of removing the flow spec if compilation fails?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
arjun4084346: This if block is not throwing any exception, I do not understand why it was inside the try block earlier.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
arjun4084346: Removing of flow spec is moved to GobblinServiceJobScheduler.java, because it was anyway doing removal in some cases, see Line 314. [this is our option 3 from the doc]

enjoyear:
@arjun4084346 , LGTM. Please commit.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -325,12 +325,14 @@ public Spec getSpecWrapper(URI uri) {
...
+  public static boolean isCompileSuccessful(Map<String, AddSpecResponse> responseMap) {
     AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, new AddSpecResponse<>(""));
 
-    return addSpecResponse != null
-        && addSpecResponse.getValue() != null
-        && !addSpecResponse.getValue().contains("ConfigException");
+    return isCompileSuccessful(addSpecResponse.getValue());
+  }
+
+  public static boolean isCompileSuccessful(String dag) {
```
sv2000, line 334: Can we move this method into BaseFlowCompiler class and expose the method via compiler interface? It seems like a property of the compiler. Definitely, not that of FlowCatalog.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -293,19 +294,27 @@ public AddSpecResponse onAddSpec(Spec addedSpec) {
...
-              this.jobExecutor.execute(new NonScheduledJobRunner(flowSpec.getUri(), false, jobConfig, null));
+              this.jobExecutor.execute(new NonScheduledJobRunner(flowSpecUri, false, jobConfig, null));
             }
           } else {
             _log.info("No FlowSpec schedule found, so running FlowSpec: " + addedSpec);
-            this.jobExecutor.execute(new NonScheduledJobRunner(flowSpec.getUri(), true, jobConfig, null));
+            this.jobExecutor.execute(new NonScheduledJobRunner(flowSpecUri, true, jobConfig, null));
+          }
+        } else {
+          _log.info("Removing the flow spec: {}, since it is an EXPLAIN request or the flow compilation failed.", addedSpec);
```
sv2000, line 314: Any chance we can distinguish the log message based on whether it is an explain query of compile failed? Will help debugging.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
arjun4084346: Yes, we can move the method into gobblin-service module, but gobblin-runtime (FlowCatalog) cannot have dependency on gobblin-service. That means listeners have to return a map in AddSpecResponse if they want to pass multiple information ( a) isCompileSuccessful b) compiledDag) to the caller.
Let me know if you want me implement it that way.

sv2000: Understood. One option would be to relocate FlowCatalog inside gobblin-service module. Not sure, if it will require a significant refactoring effort, in which case, it is ok to leave it as-is and we can do the refactoring in a separate PR.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
arjun4084346: updated

PR #2986, opened 2020-05-13
[GOBBLIN-1147]Use one dfsClient in FsDataWriter to to rename and exists check to av…

ZihanLi58:
…oid inconsistency

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1147

### Description
Since the rename with overwrite option method in FileSystem is exposed as protected, so I created a FileSyatemAdapter in the package org.apache.hadoop.fs to call it and expose it as public method.

### Tests
Test in pipeline that the rename can succeed

### Commits


gobblin-utility/src/main/java/org/apache/gobblin/util/HadoopUtils.java
```diff
@@ -266,19 +268,25 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
-      throw new FileNotFoundException(String.format("Failed to rename %s to %s: src not found", oldName, newName));
-    }
-    if (fs.exists(newName)) {
-      if (overwrite) {
-        HadoopUtils.moveToTrash(fs, newName);
-      } else {
-        throw new FileAlreadyExistsException(
-            String.format("Failed to rename %s to %s: dst already exists", oldName, newName));
+    //In default implementation of rename with rewrite option in FileSystem, if the parent dir of dst does not exist, it will throw exception,
+    //Which will fail some of our job unintentionally. So we only call that method when fs is an instance of DistributedFileSystem to avoid inconsistency problem
```
autumnust, line 272: Are you referring to the fact that the implementation of rename in DistributedFileSystem is guaranteed to be atomic?  I am not getting this comment.

gobblin-utility/src/main/java/org/apache/hadoop/fs/FileSystemAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.fs;
+
+import java.io.IOException;
+
+
+public class FileSystemAdapter {
+  public static void rename(FileSystem fs, Path src, Path dst, Options.Rename options) throws IOException {
```
autumnust, line 24: `FileContext` is introduced to handle the protected problem of rename : http://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#rename-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Path-org.apache.hadoop.fs.Options.Rename...-

gobblin-utility/src/main/java/org/apache/gobblin/util/HadoopUtils.java
ZihanLi58: No, here I mean if we direct call FileSystem.rename, if it's not a DistributedFileSystem, it will try to use the default implementation in FileSystem.class, which will throw exception when parent of dst doe not exist.

gobblin-utility/src/main/java/org/apache/hadoop/fs/FileSystemAdapter.java
ZihanLi58: Yeah, but in FsDataWriter, when we try to use fc to do rename and fs to check the existence, underlying they are using different dfsClients, and we may face inconsistency issue. So we try to use the same dsfClient to do rename here.

gobblin-utility/src/main/java/org/apache/gobblin/util/HadoopUtils.java
```diff
@@ -266,19 +268,25 @@ public static void renamePath(FileSystem fs, Path oldName, Path newName) throws
...
+    if(fs instanceof DistributedFileSystem) {
+      Options.Rename renameOptions = (overwrite) ? Options.Rename.OVERWRITE : Options.Rename.NONE;
+      FileSystemAdapter.rename(fs, oldName, newName, renameOptions);
```
sv2000, line 275: Do we need the FileSystemAdapter class? Can't we directly call fs.rename(src,dest,options) here?

gobblin-core/src/main/java/org/apache/gobblin/writer/FsDataWriter.java
```diff
@@ -262,7 +262,7 @@ public void commit()
     LOG.info(String.format("Moving data from %s to %s", this.stagingFile, this.outputFile));
     // For the same reason as deleting the staging file if it already exists, overwrite
     // the output file if it already exists to prevent task retry from being blocked.
-    HadoopUtils.renamePath(this.fileContext, this.stagingFile, this.outputFile, true);
+    HadoopUtils.renamePath(this.fs, this.stagingFile, this.outputFile, true);
```
sv2000, line 265: Did you forget to remove the fileContext member variable from this class?

ZihanLi58: This is intentional, as we don't know if there is any subclass use this member. So I leave it to make backward consistency

gobblin-utility/src/main/java/org/apache/gobblin/util/HadoopUtils.java
ZihanLi58: Good catch, I just realize that DistributedFileSystem expose that method as public method.

PR #2988, opened 2020-05-14
[GOBBLIN-1150] spec catalog table schema change

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @jack-moseley  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1150

### Description
This PR make a transitional changes in spec store table, as described in https://docs.google.com/document/d/1POQXo9lHnpXfzE1MDYemdOdaUjhtBKhvuQxxsFlLIk8/edit
To use JSON data type in mysql, mysql version 5.7 was needed, which is not available in trusty distro of ubuntu where travis runs, so i had to bump that up to xenial. xenial does not accept oraclejdk8 or lower, so i had to change it to openjdk8

### Tests
added a test class which uses new table format for add/delete/get etc APIs

### Commits


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -65,31 +65,49 @@
 
   private static final String CREATE_TABLE_STATEMENT =
       "CREATE TABLE IF NOT EXISTS %s (spec_uri VARCHAR(128) NOT NULL, tag VARCHAR(128) NOT NULL, spec LONGBLOB, PRIMARY KEY (spec_uri))";
+  private static final String CREATE_TABLE_STATEMENT_V2 =
```
sv2000, line 68: Isn't this schema backwards compatible with the V1 schema i.e. you are simply adding new columns? We can retain the blob format for the "spec" column as before and define a new column "specV2" which stores the spec in JSON format. If so, do we need to support two different versions of the table?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -140,11 +172,22 @@ public boolean deleteSpec(Spec spec) throws IOException {
...
+        PreparedStatement statementV2 = connection.prepareStatement(String.format(DELETE_STATEMENT, this.tableNameV2))) {
+
       statement.setString(1, specUri.toString());
-      int result = statement.executeUpdate();
+      statementV2.setString(1, specUri.toString());
+
+      int resultV2 = statementV2.executeUpdate();
+      if (this.writeToOldTable) {
+        int result = statement.executeUpdate();
+        if (resultV2 != result) {
```
sv2000, line 184: Should we throw an exception and fail the deleteSpec, instead of logging and moving on?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -65,31 +65,49 @@
 
   private static final String CREATE_TABLE_STATEMENT =
       "CREATE TABLE IF NOT EXISTS %s (spec_uri VARCHAR(128) NOT NULL, tag VARCHAR(128) NOT NULL, spec LONGBLOB, PRIMARY KEY (spec_uri))";
+  private static final String CREATE_TABLE_STATEMENT_V2 =
+      "CREATE TABLE IF NOT EXISTS %s (spec_uri VARCHAR(128) NOT NULL, flow_group VARCHAR(128), flow_name VARCHAR(128), "
```
sv2000, line 69: I think the schema can be enhanced with a few additional fields such as requesterId, a boolean field "isRunImmediately" to indicate if the scheduled flow is also a run immediately flow, a futuristic field "owningGroup" for flows which are owned by a group, "timezone" to indicate the timezone the cron schedule applies to. Also, does user_to_proxy need to be a top level field? It applies only when the orchestrator is Azkaban and can probably be queried from the JSON spec.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
arjun4084346: That's right. will change

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
arjun4084346: will 'owningGroup' be a string, or a json having some more information like a requester list? if we are not sure, we can leave this column for implementation time, otherwise we might have to add additional column for 'owningGroup' if we choose a wrong type now?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
arjun4084346: yes

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -62,32 +62,34 @@
...
-  private static final String GET_STATEMENT = "SELECT spec FROM %s WHERE spec_uri = ?";
-  private static final String GET_ALL_STATEMENT = "SELECT spec_uri, spec FROM %s";
-  private static final String GET_ALL_STATEMENT_WITH_TAG = "SELECT spec_uri, spec FROM %s WHERE tag = ?";
-  static final String WRITE_TO_OLD_TABLE_KEY = "write.to.old.table";
-  static final String READ_FROM_OLD_TABLE_KEY = "read.from.old.table";
+  private static final String GET_STATEMENT = "SELECT %s FROM %s WHERE spec_uri = ?";
+  private static final String GET_ALL_STATEMENT = "SELECT spec_uri, %s FROM %s";
+  private static final String GET_ALL_URIS_STATEMENT = "SELECT spec_uri FROM %s";
+  private static final String GET_ALL_STATEMENT_WITH_TAG = "SELECT spec_uri FROM %s WHERE tag = ?";
+  static final String WRITE_TO_OLD_COLUMN = "write.to.old.column";
```
sv2000, line 82: Do you think we can mark the old column as deprecated and write to both old and new columns? Eventually, in a future release, the old column can be dropped. We won't need the user to provide configs to write to old/new columns.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -87,6 +101,9 @@ public MysqlSpecStore(Config config, SpecSerDe specSerDe) throws IOException {
     this.specStoreURI = URI.create(config.getString(ConfigurationKeys.STATE_STORE_DB_URL_KEY));
     this.specSerDe = specSerDe;
 
+    this.writeToOldColumn = ConfigUtils.getBoolean(config, WRITE_TO_OLD_COLUMN, true);
```
sv2000, line 104: See comment earlier about writing to both old and new columns.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -164,17 +182,17 @@ public Spec updateSpec(Spec spec) throws IOException, SpecNotFoundException {
...
       statement.setString(1, specUri.toString());
 
       try (ResultSet rs = statement.executeQuery()) {
         if (!rs.next()) {
           throw new SpecNotFoundException(specUri);
         }
-
-        Blob blob = rs.getBlob(1);
-        return this.specSerDe.deserialize(ByteStreams.toByteArray(blob.getBinaryStream()));
+        return this.readFromOldColumn
```
sv2000, line 193: Same comment as earlier.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -194,15 +212,18 @@ public Spec getSpec(URI specUri, String version) throws IOException, SpecNotFoun
...
+            String.format(GET_ALL_STATEMENT, this.readFromOldColumn ? "spec" : NEW_COLUMN, this.tableName))) {
       List<Spec> specs = new ArrayList<>();
 
       try (ResultSet rs = statement.executeQuery()) {
         while (rs.next()) {
           try {
-            Blob blob = rs.getBlob(2);
-            specs.add(this.specSerDe.deserialize(ByteStreams.toByteArray(blob.getBinaryStream())));
+            specs.add(
+                this.readFromOldColumn
```
sv2000, line 223: same comment as above.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
arjun4084346: Yea, these configs are only for transition period, will be removed later.

Do we want to always write to both the columns and one day stop writing to old column?

similarly do we want to start reading from the new column immediately?

sv2000: Sorry - my earlier comment could have been clearer. Yeah, writes to both columns is simpler. For read, we can check if the new column is not null and read from the new column else default to the old column. Would be nice to avoid having to pass configurations.

arjun4084346: Okay. Yea it will reduce complexity, but we would be reading 2 columns till we remove the old column.

.travis.yml
```diff
@@ -46,7 +44,7 @@ env:
   - RUN_TEST_GROUP=coverage
 
 jdk:
-  - oraclejdk8
+  - openjdk8
```
sv2000, line 47: Curious: can we make do without this change?

arjun4084346: xenial wanted oraclejdk9+
by chance it accepted openjdk8 though.
we should start supporting jdk9+ also, but i found a small portion of gobblin code is not compatible with jdk9

sv2000: Ok. I am not opinionated about this. Would like to hear what others think @htran1 ?

arjun4084346: Other option is to remain at trusty , but do not use pre installed mysql.
My opinion is that we should make efforts to support newer version of OS and java.

PR #2993, opened 2020-05-21
[GOBBLIN-1154] Improve gaas error messages

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1154

### Description

Variety of improvements to GaaS error returning:

- When creating flowConfig, in case of error throw it instead of returning 4xx along with flowConfig. Returning the `CreateKVResponse` means that the the status is only seen if using `--verbose`, and there is no clear error message
- Similarly return a more clear error message when calling get for a flowExecution that doesn't exist
- Instead of concatenating execution links in the message field of flowStatus (this is not useful since they are already in the jobStatus list), use the message field to return a flow level error message
- Added a `FlowCancelled` event so that flows that were SLA killed or manually killed will have a `CANCELLED` status instead of `FAILED`

### Tests

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
@@ -24,6 +24,7 @@ import com.linkedin.restli.common.ComplexResourceKey;
 import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.common.PatchRequest;
 import com.linkedin.restli.server.CreateKVResponse;
+import com.linkedin.restli.server.RestLiServiceException;
 import com.linkedin.restli.server.UpdateResponse;
 
 import lombok.extern.slf4j.Slf4j;
@@ -62,8 +63,9 @@ public class FlowConfigV2ResourceLocalHandler extends FlowConfigResourceLocalHan
 
     // Return conflict and take no action if flowSpec has already been created
     if (this.flowCatalog.exists(flowSpec.getUri())) {
-      log.warn("Flowspec with URI {} already exists, no action will be taken", flowSpec.getUri());
-      return new CreateKVResponse(new ComplexResourceKey<>(flowConfig.getId(), flowStatusId), flowConfig, HttpStatus.S_409_CONFLICT);
+      log.warn("FlowSpec with URI {} already exists, no action will be taken", flowSpec.getUri());
+      throw new RestLiServiceException(HttpStatus.S_409_CONFLICT,
+          "FlowSpec with URI " + flowSpec.getUri() + " already exists, no action will be taken");
     }
 
     Map<String, AddSpecResponse> responseMap = this.flowCatalog.put(flowSpec, triggerListener);
@@ -81,7 +83,11 @@ public class FlowConfigV2ResourceLocalHandler extends FlowConfigResourceLocalHan
     } else if (Boolean.parseBoolean(responseMap.getOrDefault(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("false")).getValue().toString())) {
       httpStatus = HttpStatus.S_201_CREATED;
     } else {
-      httpStatus = HttpStatus.S_400_BAD_REQUEST;
+      String message = "Flow was not compiled successfully. It may be due to no path being found";
+      if (!flowSpec.getCompilationErrors().isEmpty()) {
+        message = message + " or due to " + flowSpec.getCompilationErrors();
+      }
+      throw new RestLiServiceException(HttpStatus.S_400_BAD_REQUEST, message);
     }
 
     return new CreateKVResponse(new ComplexResourceKey<>(flowConfig.getId(), flowStatusId), flowConfig, httpStatus);
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
@@ -32,6 +32,7 @@ import com.linkedin.restli.common.ComplexResourceKey;
 import com.linkedin.restli.common.EmptyRecord;
 import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.server.PagingContext;
+import com.linkedin.restli.server.RestLiServiceException;
 import com.linkedin.restli.server.UpdateResponse;
 import com.linkedin.restli.server.annotations.Context;
 import com.linkedin.restli.server.annotations.Finder;
@@ -65,8 +66,12 @@ public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatus
    */
   @Override
   public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
-    // this returns null to raise a 404 error if flowStatus is null
-    return convertFlowStatus(getFlowStatusFromGenerator(key, this._flowStatusGenerator));
+    FlowExecution flowExecution = convertFlowStatus(getFlowStatusFromGenerator(key, this._flowStatusGenerator));
+    if (flowExecution == null) {
+      throw new RestLiServiceException(HttpStatus.S_404_NOT_FOUND, "No flow execution found for flowStatusId " + key.getKey()
+      + ". The flowStatusId may be incorrect, or the flow execution may have been cleaned up.");
+    }
+    return flowExecution;
   }
 
   @Finder("latestFlowExecution")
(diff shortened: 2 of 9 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -259,6 +259,7 @@ public void orchestrate(Spec spec) throws Exception {
         // In this case, the current time is used as the flow executionId.
         flowMetadata.putIfAbsent(TimingEvent.FlowEventConstants.FLOW_EXECUTION_ID_FIELD,
             Long.toString(System.currentTimeMillis()));
+        flowMetadata.put(TimingEvent.METADATA_MESSAGE, "Flow compilation failed");
```
arjun4084346, line 262: Can we add spec.getCompilationErrors() also? We will know which execution failed with which config exception.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -476,6 +476,9 @@ private void cancelDag(String dagToCancel) throws ExecutionException, Interrupte
         for (DagNode<JobExecutionPlan> dagNodeToCancel : dagNodesToCancel) {
           cancelDagNode(dagNodeToCancel);
         }
+
+        this.dags.get(dagToCancel).setFlowEvent(TimingEvent.FlowTimings.FLOW_CANCELLED);
```
arjun4084346, line 480: Should we setFlowEvent() inside cancelDagNode() ?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
```diff
@@ -81,7 +83,7 @@ public CreateKVResponse createFlowConfig(FlowConfig flowConfig, boolean triggerL
     } else if (Boolean.parseBoolean(responseMap.getOrDefault(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("false")).getValue().toString())) {
       httpStatus = HttpStatus.S_201_CREATED;
     } else {
-      httpStatus = HttpStatus.S_400_BAD_REQUEST;
+      throw new RestLiServiceException(HttpStatus.S_400_BAD_REQUEST, "Flow was not compiled successfully");
```
aplex, line 86: What are the next steps for the user here? Should he check the flow properties (we can point to the list of allowed properties), should he file a ticket and ask GaaS engineers to fix the issue? Also, do we have additional details on why the compilation was unsuccessful and what specific place had a problem?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
```diff
@@ -65,8 +66,11 @@ public FlowExecutionResource() {}
    */
   @Override
   public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
-    // this returns null to raise a 404 error if flowStatus is null
-    return convertFlowStatus(getFlowStatusFromGenerator(key, this._flowStatusGenerator));
+    FlowExecution flowExecution = convertFlowStatus(getFlowStatusFromGenerator(key, this._flowStatusGenerator));
+    if (flowExecution == null) {
+      throw new RestLiServiceException(HttpStatus.S_404_NOT_FOUND, "No flow execution found for flowStatusId " + key.getKey());
```
aplex, line 71: When can this happen? Was the wrong flow id provided and user needs to check it, or the flows can get cleaned up/deleted after some period of time? Can it be the problem with permissions that he does not see the flow?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: This. is setting the flow event on a dag level, not a dag node level, so it makes more sense to do it in cancelDag.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
jack-moseley: We could, the reason I didn't want to is it would be kind of misleading right now since sometimes that will have unrelated missing config messages. I guess for now I could just add it and say "it may be due to these errors".

arjun4084346: Right.

arjun4084346: So the next challenge for us is to find failures in resolving which flow/job template are the correct errors.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
jack-moseley: The challenge is that if a flow doesn't compile it's hard to know why because we don't know what was the intended path. It could be because of missing edges/nodes, or unresolved config keys. It's something that could be improved upon still, but most users of gaas self-serve should be using edges we've already defined so probably shouldn't have this problem anyway.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
jack-moseley: It would mean either wrong flow id or the flow status was cleaned up.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/TimingEventUtils.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.gobblin.runtime.api.FlowSpec;
 import org.apache.gobblin.runtime.api.JobSpec;
 import org.apache.gobblin.runtime.api.SpecExecutor;
+import org.apache.gobblin.service.modules.flowgraph.Dag;
```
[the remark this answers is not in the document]
aplex: These types of problems should be automatically caught by SpotBugs/CheckStyle. Another example that they should be enabled on all repos.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
aplex: Can you add it to the message? The point here is to provide detailed and helpful error messages to users, so they don't come back to us with follow up questions, and can troubleshoot themselves.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigV2ResourceLocalHandler.java
aplex: So this will be primarily used by SREs? I can think of a situation when they have a 2-page flow and a typo in it. How would they find where the typo is? Are there any helpful compilation exceptions we can provide? For example: "problem is in the line 123, variable 'sssschema' is not found", or "no path found between 'faro' and 'warr', check the names of the source and the target").

If we can't pinpoint a single error and the problems are all around, we can send users to the compilation logs, and provide them some unique token so that they can find the relevant lines easily. For example: "for more information, check the GaaS logs that have requestId=XXX"

jack-moseley: The reason could be a couple things:

1. No path found. This could be due to: the edges don't have the paths they are using, there are no edges between those nodes, the nodes or edges were not added on gaas startup due to an error, etc.

2. There is an unresolved config. We do have a field that stores whatever errors we encountered with unresolved config.

The problems are:
- If it fails due to 1, there isn't really any way to know exactly what went wrong since we don't know what the intended path was.
- Even if it fails due to 1, we may have encountered some unresolved config on a separate unintended path. So it would be pretty confusing to include an error that some config was unresolved when the real reason was that the path.

If anything maybe the best message for now would just be "This could be because no path is found, or because of X unresolved config"?

aplex: That can be a good start.

aplex:
LGTM, thanks for improving experience of our users!

jack-moseley:
@sv2000 please merge

PR #3022, opened 2020-06-02
Correct handling of 'special' dependencies in poms

mockitoguy:
Dependencies declared with target configurations need extra work to get them correctly mapped to pom files. Pom files are lossy when compared to Gradle or Ivy model of dependencies.

Added 'all' module that enables customers to declare a single dependency and get all directly consumable Gobblin artifacts.

Tested by running './gradlew publishToMavenLocal' and inspecting the build output and the generated pom files.

### DIFF
--- a/gobblin-modules/gobblin-parquet/build.gradle
+++ b/gobblin-modules/gobblin-parquet/build.gradle
@@ -24,7 +24,10 @@ dependencies {
   compile externalDependency.gson
   compile externalDependency.twitterParquet
   compile externalDependency.twitterParquetAvro
-  compile externalDependency.twitterParquetProto
+
+  //parquet-protobuf dependency is not resolvable from MavenCentral
+  //compileOnly ensures that gobblin-parquet can be cleanly resolved from MavenCentral
+  compileOnly externalDependency.twitterParquetProto
 
   testCompile externalDependency.testng
   testCompile externalDependency.mockito


mockitoguy:
Travis CI failure is due to test flakiness.

gobblin-all/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+
+apply plugin: "java"
+
+dependencies {
+    //TODO: list all dependencies that are directly consumable
+    compile project(":gobblin-api")
+    compile project(":gobblin-rest-service:gobblin-rest-api")
+    compile project(path: ':gobblin-rest-service:gobblin-rest-api', configuration: 'restClient')
+    compile project(path: ':gobblin-rest-service:gobblin-rest-api', configuration: 'dataTemplate')
+}
```
vikrambohra, line 12: Please add all gobblin modules provided on slack

gradle/scripts/bintrayPublishing.gradle
```diff
@@ -52,6 +52,42 @@ subprojects{
...
+                def newArtifactId = d.name + "-" + confToPom[d.targetConfiguration]
+                dependencyNode.appendNode('artifactId', newArtifactId)
+                dependencyNode.appendNode('version', d.version)
+                added.add(newArtifactId)
+              }
+            }
+            if (added || removed) {
+              logger.lifecycle("Updated pom dependencies in {}, added: {}, removed: {}", project.path, added, removed)
+            }
+          }
```
vikrambohra, line 90: Thanks for this. It looks clean. Can you verify if the generated pom also adds global dependency exclusions? (scripts/gradle/globalDependencies.gradle)

vikrambohra: Based on spot check. Looks like the dependency exclusion rules are only applied to the newly created (gobblin-rest-api-rest-client etc) dependencies in the POM. But does the initial POM have the dependency exclusion rules as well?

mockitoguy: They are applied to all dependencies: https://gist.github.com/mockitoguy/662dc4d0f444ee082c3ccc65d9526f35

PR #3029, opened 2020-06-04
Fixed resolution problems of gobblin-apache-parquet

mockitoguy:
'org.apache.parquet:parquet-protobuf:1.10.1' is not resolvable from MavenCentral because it's missing some of its transitive dependencies. This will cause 'gobblin-apache-parquet' module *also not be resolvable* from Central (in the future, when it will be published to Central). To fix this, I suggest using 'compileOnly'. This will ensure that 'gobblin-parquet-apache' is cleanly resolvable from Central.

This change unblocks automated consumption of gobblin external artifact at LinkedIn

Tested by running the build and inspecting the pom files


mockitoguy:
Actually, there is already a PR for this: https://github.com/apache/incubator-gobblin/pull/3028

PR #3038, opened 2020-06-11
[GOBBLIN-1190] Fallback to full schema if configured shuffle schema is not available

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1190

### Description

Check if the record schema contains configured shuffle-key schema: If not, keep using the whole record as shuffle key in the ORC-Compaction. 

### Tests

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/orc/OrcUtils.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/orc/OrcUtils.java
@@ -29,8 +29,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-import org.apache.gobblin.compaction.mapreduce.avro.MRCompactorAvroKeyDedupJobRunner;
-import org.apache.gobblin.util.FileListUtils;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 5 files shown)


gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/orc/OrcUtils.java
```diff
@@ -482,4 +483,32 @@ public static WritableComparable createValueRecursively(TypeDescription schema,
...
+
+  /**
+   * Check recursively if owning schema is eligible to be up-converted to targetSchema:
+   * 1. TargetSchema is a subset of originalSchema.
+   * 2. TargetSchema is larger than originalSchema.
+   */
+  public static boolean eligibleForUpConvert(TypeDescription originalSchema, TypeDescription targetSchema) {
+    if (!targetSchema.getCategory().isPrimitive()) {
+      if (originalSchema.getFieldNames().size() < targetSchema.getFieldNames().size()) {
+        return true;
```
sv2000, line 495: Why is targetSchema having more fields than original schema a sufficient condition for up conversion?

sv2000: Shouldn't we be checking if each field in the target Schema is present in the record schema?

PR #3042, opened 2020-06-15
[GOBBLIN-1196] search flow configs using flow properties and/or other parameters

arjun4084346:
Dear Gobblin maintainers,

Please accept this PR. I understand that it will not be reviewed until I have checked off all the steps below! @jack-moseley  @sv2000  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1196

### Description
This PR adds support to search flow configs using different parameters like flow group, flow name, schedule and also using flow properties

### Tests
added/updated unit tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
@@ -17,14 +17,6 @@
 
 package org.apache.gobblin.runtime.api;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Optional;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
-import com.typesafe.config.Config;
-import com.typesafe.config.ConfigFactory;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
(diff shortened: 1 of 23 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -267,7 +276,128 @@ public Spec getSpec(URI specUri, String version) throws IOException, SpecNotFoun
...
+      conditions.add("destination_identifier = ?");
+    }
+
+    if (flowSpecSearchObject.getSchedule() != null) {
+      conditions.add("schedule = ?");
+    }
+
+    if (flowSpecSearchObject.getSchedule() != null) {
+      conditions.add("modified_time = ?");
+    }
```
jack-moseley, line 318: There's two conditions added here for schedule and only one gets set in `setGetPreparedStatement`, is this a mistake?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
```diff
@@ -411,5 +419,43 @@ public static String getFlowGroup(URI flowUri) {
       }
       return uriTokens[EXPECTED_NUM_URI_PATH_TOKENS - 2];
     }
+
+    public static FlowConfig toFlowConfig(Spec spec) {
```
sv2000, line 423: Add javadoc here.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecSearchObject.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.runtime.api;
+
+/**
+ * This is an interface to package all the parameters that should be used to search {@link Spec} in a {@link SpecStore}
+ */
+public interface SpecSearchObject {
```
sv2000, line 23: Why define an empty interface?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -269,23 +270,40 @@ public boolean exists(URI uri) {
...
+    try {
+      return specStore.getSpecs(specSearchObject);
+    } catch (IOException e) {
+      throw new RuntimeException("Cannot retrieve Spec from Spec store for URI: " + specSearchObject, e);
+    }
+  }
+
+  public Collection<Spec> getAllSpecs() {
+    try {
+      return specStore.getSpecs();
```
sv2000, line 292: An option to consider: GobblinServiceScheduler already caches all the specs in memory. So we should be able to avoid going to the SpecStore to retrieve all the specs each time getAllSpecs() is invoked.

arjun4084346: Yes, good consideration. Also need to pay attention if slave nodes also have all the specs in their memory.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecSearchObject.java
arjun4084346: This is a "marker interface" to package parameters for search queries. So far, FlowSpecSearchObject implements this, which has some unique set of params suitable for a FlowSpec. In future, it can be implemented to package params suitable for TopologySpec s

PR #3049, opened 2020-06-19
GOBBLIN-1202 & DSS-26966: Add retry for REST API call

arekusuri:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1202

### Description
SFDC objects have index on their column - SystemModstamp
This index could be in disk. When we execute  
**Select count(systemmodstamp) from table_name group by day_only(systemmodstamp)**
If the index is in disk, it needs to load. It would be timeout.
Retry would result it.

### DIFF
--- a/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
+++ b/gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
@@ -424,9 +424,10 @@ public class SalesforceSource extends QueryBasedSource<JsonArray, JsonElement> {
   /**
    * Get a {@link JsonArray} containing the query results
    */
+  @SneakyThrows
   private JsonArray getRecordsForQuery(SalesforceConnector connector, String query) {
     RestApiProcessingException exception = null;
-    for (int i = 0; i < workUnitConf.restRetries; i++) {
+    for (int i = 0; i < workUnitConf.restApiRetryLimit; i++) {
       try {
         String soqlQuery = SalesforceExtractor.getSoqlUrl(query);
         List<Command> commands = RestApiConnector.constructGetCommand(connector.getFullUri(soqlQuery));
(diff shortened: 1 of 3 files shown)


gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -417,23 +425,31 @@ private int computeTargetPartitionSize(Histogram histogram, int minTargetPartiti
...
-      if (itr.hasNext()) {
-        output = itr.next();
-      } else {
-        throw new DataRecordException("Failed to get data from salesforce; REST response has no output");
+        return GSON.fromJson(output, JsonObject.class).getAsJsonArray("records");
+      } catch (RestApiClientException | DataRecordException e) {
+        throw new RuntimeException("Fail to get data from salesforce", e);
+      } catch (RestApiProcessingException e) {
+        exception = e;
+        log.info(String.format("Caught RestApiProcessingException, retry(%s) rest query: %s", i+1, query));
```
htran1, line 448: Please use {} logger formatting instead of String.format.

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
```diff
@@ -32,6 +32,7 @@
 import java.util.Set;
 
 import java.util.stream.Collectors;
+import lombok.SneakyThrows;
```
htran1, line 35: Unused import?

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: good catch, fixed!
# And added wait duration
# refactored variable name

gobblin-salesforce/src/main/java/org/apache/gobblin/salesforce/SalesforceSource.java
arekusuri: Thanks! It was unused import. 
We are using it now.

PR #3050, opened 2020-06-19
[GOBBLIN-1203] Adding configurations for staging directory in Distcp template

aryavaibhav93:
Dear Gobblin maintainers,

JIRA
https://jira01.corp.linkedin.com:8443/browse/ETL-10948

Description
Currently, Embedded distcp creates it's own temp staging directory. But, this PR adds the option to have a user defined staging directory which can be used in place of the temp staging directory.

Tests
My PR incorporates change to the following unit tests:
FileAwareInputStreamDataWriterTest.java

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -207,6 +207,8 @@ public class ConfigurationKeys {
   public static final String CLEANUP_OLD_JOBS_DATA = "cleanup.old.job.data";
   public static final boolean DEFAULT_CLEANUP_OLD_JOBS_DATA = false;
   public static final String MAXIMUM_JAR_COPY_RETRY_TIMES_KEY = JOB_JAR_FILES_KEY + ".uploading.retry.maximum";
+  public static final String USER_DEFINED_STATIC_STAGING_DIR = "user.defined.static.staging.dir";
+  public static final String USER_DEFINED_STAGING_DIR_FLAG = "user.defined.staging.dir.flag";
 
   public static final String QUEUED_TASK_TIME_MAX_SIZE = "taskexecutor.queued_task_time.history.max_size";
   public static final int DEFAULT_QUEUED_TASK_TIME_MAX_SIZE = 2048;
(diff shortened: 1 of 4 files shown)


gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -207,6 +207,8 @@
   public static final String CLEANUP_OLD_JOBS_DATA = "cleanup.old.job.data";
   public static final boolean DEFAULT_CLEANUP_OLD_JOBS_DATA = false;
   public static final String MAXIMUM_JAR_COPY_RETRY_TIMES_KEY = JOB_JAR_FILES_KEY + ".uploading.retry.maximum";
+  public static final String USER_DEFINED_STAGING_DIR = "user.defined.staging.dir";
```
autumnust, line 210: I'll just name it as "user.defined.static.staging.dir". Let's sync offline for why I suggest doing this.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -140,9 +140,13 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
     this.fs = FileSystem.get(uri, conf);
     this.fileContext = FileContext.getFileContext(uri, conf);
 
-    this.stagingDir = this.writerAttemptIdOptional.isPresent() ? WriterUtils
-        .getWriterStagingDir(state, numBranches, branchId, this.writerAttemptIdOptional.get())
-        : WriterUtils.getWriterStagingDir(state, numBranches, branchId);
+    if (state.getPropAsBoolean(ConfigurationKeys.USER_DEFINED_STAGING_DIR_FLAG)) {
```
autumnust, line 143: Think about maintaining the backward-compatibility for such kind of change.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
aryavaibhav93: I have renamed key: USER_DEFINED_STAGING_DIR as USER_DEFINED_STATIC_STAGING_DIR and value: user.defined.staging.dir as user.defined.static.staging.dir

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
aryavaibhav93: For backward-compatibility, I have added default value of USER_DEFINED_STAGING_DIR_FLAG as false, also, in disctp template the value for the above flag is false

PR #3054, opened 2020-07-01
Making dataset name as part of staging directory in gobblin-distcp (f…

aryavaibhav93:
…ile-based)

Dear Gobblin maintainers,

### JIRA
Making dataset name as part of staging directory in gobblin-distcp (file-based)

### Description
Currently, Embedded distcp creates it's own temp or user defined staging directory. But, this PR adds the option to have a staging directory which is being created within datasets.

### Tests
My PR incorporates change to the following unit tests:
FileAwareInputStreamDataWriterTest.java

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -349,6 +349,9 @@ public class ConfigurationKeys {
   public static final String DEFAULT_FORK_RECORD_QUEUE_TIMEOUT_UNIT = TimeUnit.MILLISECONDS.name();
   public static final String FORK_MAX_WAIT_MININUTES = "fork.max.wait.minutes";
   public static final long DEFAULT_FORK_MAX_WAIT_MININUTES = 60;
+  public static final String FORK_CLOSE_WRITER_ON_COMPLETION = "fork.closeWriterOnCompletion";
+  public static final boolean DEFAULT_FORK_CLOSE_WRITER_ON_COMPLETION = true;
+
 
   /**
    * Writer configuration properties.
--- a/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/filebased/FileBasedExtractor.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/filebased/FileBasedExtractor.java
@@ -160,6 +160,8 @@ public class FileBasedExtractor<S, D> extends InstrumentedExtractor<S, D> {
     if (this.currentFile != null && this.currentFileItr != null) {
       closeCurrentFile();
       incrementBytesReadCounter();
+      // release the reference to allow garbage collection
+      this.currentFileItr = null;
     }
 
     while (!this.hasNext && !this.filesToPull.isEmpty()) {
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/HiveWritableHdfsDataWriter.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/HiveWritableHdfsDataWriter.java
@@ -42,7 +42,7 @@ import org.apache.gobblin.configuration.State;
  */
 public class HiveWritableHdfsDataWriter extends FsDataWriter<Writable> {
 
-  protected final RecordWriter writer;
+  protected RecordWriter writer;
   protected final AtomicLong count = new AtomicLong(0);
   // the close method may be invoked multiple times, but the underlying writer only supports close being called once
   private boolean closed = false;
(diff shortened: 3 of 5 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -104,6 +106,7 @@
   private final Options.Rename renameOptions;
   private final FileContext fileContext;
 
+
```
autumnust, line 109: Can you remove this empty line if being added accidentally ?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -393,6 +393,7 @@
   public static final String SIMPLE_WRITER_DELIMITER = "simple.writer.delimiter";
   public static final String SIMPLE_WRITER_PREPEND_SIZE = "simple.writer.prepend.size";
   public static final String WRITER_ADD_TASK_TIMESTAMP = WRITER_PREFIX + ".addTaskTimestamp";
+  public static final String DATASET_DEFINED_STAGING_DIR_FLAG = "dataset.defined.staging.dir.flag";
```
autumnust, line 396: Shall we rename it ? since it is not clear to me what does "dataset.defined" mean here. Also, Let's avoid words like flag since it is usually ambiguous.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -139,17 +142,20 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
...
+      this.stagingDir = new Path(this.copyableDatasetMetadata.getDatasetURN() + STAGING_DIR_SUFFIX);
     } else {
       this.stagingDir = this.writerAttemptIdOptional.isPresent() ? WriterUtils.getWriterStagingDir(state, numBranches, branchId, this.writerAttemptIdOptional.get())
           : WriterUtils.getWriterStagingDir(state, numBranches, branchId);
     }
 
     this.outputDir = getOutputDir(state);
-    this.copyableDatasetMetadata =
-        CopyableDatasetMetadata.deserialize(state.getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
+
```
autumnust, line 158: same as above

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -139,17 +142,20 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
     URI uri = URI.create(uriStr);
     this.fs = FileSystem.get(uri, conf);
     this.fileContext = FileContext.getFileContext(uri, conf);
+    this.copyableDatasetMetadata =
+        CopyableDatasetMetadata.deserialize(state.getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
 
     if (state.getPropAsBoolean(ConfigurationKeys.USER_DEFINED_STAGING_DIR_FLAG,false)) {
       this.stagingDir = new Path(state.getProp(ConfigurationKeys.USER_DEFINED_STATIC_STAGING_DIR));
+    } else if (state.getPropAsBoolean(ConfigurationKeys.DATASET_DEFINED_STAGING_DIR_FLAG,false)) {
```
autumnust, line 150: Hmm I think there's something missing here.

Please take a look at `org.apache.gobblin.util.WriterUtils#getWriterStagingDir(org.apache.gobblin.configuration.State, int, int)` method, which is called below. The determination of a staging directory is not totally based on user-defined configuration, but some runtime behavior.


Specifically within the `getWriterStagingDir`, the root path of staging dir comes from `org.apache.gobblin.configuration.ConfigurationKeys#WRITER_STAGING_DIR` which is a conf set dynamically in the gobblin job. In the method it also contains logic to differentiate different staging directory for different forks. The current change you made will lost that feature. 

We can sync offline on this if that could help.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -139,17 +142,20 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
...
+      this.stagingDir = new Path(this.copyableDatasetMetadata.getDatasetURN() + STAGING_DIR_SUFFIX);
```
aplex, line 151: URN is a standard with a special format like "urn:linkedin:data". Looks like we'll have a path and not a URN here. I wonder if there are other methods that will return data set location, or whenever we can rename this method to something like getDatasetURI() or getDatasetURL(), depending on what we can get here. See also the discussion on the differences between URI/URL/URN - https://stackoverflow.com/a/28865728

PR #3065, opened 2020-07-21
[GOBBLIN-1217] start metrics reporting with a few map-reduce properties

arjun4084346:
Dear Gobblin maintainers,

@sv2000 @jack-moseley please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1217

### Description
1) Add a few map/reduce task related configs in building metric reporter. One of the use case can be that each mapper/reducer's metrics can be analyzed separately
2) made some minor codestyle changes intelliJ was complaining about

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -708,11 +708,6 @@ public class ConfigurationKeys {
   public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(false);
   public static final String METRICS_LOG_DIR_KEY = METRICS_CONFIGURATIONS_PREFIX + "log.dir";
   public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
-  public static final String MR_TYPE_KEY = METRICS_CONFIGURATIONS_PREFIX + "mr.type";
-  public static final String MAPPER_TASK_NUM_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.mapper.task.num";
-  public static final String MAPPER_TASK_ATTEMPT_NUM_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.mapper.task.attempt.num";
-  public static final String REDUCER_TASK_NUM_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.reducer.task.num";
-  public static final String REDUCER_TASK_ATTEMPT_NUM_KEY = METRICS_CONFIGURATIONS_PREFIX + "reporting.reducer.task.attempt.num";
   public static final String DEFAULT_METRICS_FILE_SUFFIX = "";
   public static final String FAILURE_REPORTING_FILE_ENABLED_KEY = "failure.reporting.file.enabled";
   public static final String DEFAULT_FAILURE_REPORTING_FILE_ENABLED = Boolean.toString(true);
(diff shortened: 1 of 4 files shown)


gobblin-metrics-libs/gobblin-metrics/src/main/java/org/apache/gobblin/metrics/GobblinMetrics.java
```diff
@@ -367,6 +374,20 @@ public void startMetricReportingWithFileSuffix(State state, String metricsFileSu
       oldMetricsFileSuffix += "." + metricsFileSuffix;
     }
     metricsReportingProps.setProperty(ConfigurationKeys.METRICS_FILE_SUFFIX, oldMetricsFileSuffix);
+
+    if (!Strings.isNullOrEmpty(metricsFileSuffix)) {
+      String[] tokens = metricsFileSuffix.split("_");
```
jack-moseley, line 379: I just googled and found that there are mapreduce classes for these such as `TaskAttemptID`. Would it be possible to make use of these to get the properties through APIs instead of parsing the string manually?

gobblin-metrics-libs/gobblin-metrics/src/main/java/org/apache/gobblin/metrics/GobblinMetrics.java
```diff
@@ -367,6 +374,20 @@ public void startMetricReportingWithFileSuffix(State state, String metricsFileSu
       oldMetricsFileSuffix += "." + metricsFileSuffix;
     }
     metricsReportingProps.setProperty(ConfigurationKeys.METRICS_FILE_SUFFIX, oldMetricsFileSuffix);
+
+    if (!Strings.isNullOrEmpty(metricsFileSuffix)) {
```
sv2000, line 378: I think GobblinMetrics should be agnostic to the execution mode of the job, and shouldn't have to worry about extracting mapper/reducer num from the task id. This block can be moved inside MRJobLauncherClass and the configurations such as MAPPER_TASK_NUM_KEY can be set there and added into the state object.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -708,6 +708,11 @@
   public static final String DEFAULT_METRICS_REPORTING_FILE_ENABLED = Boolean.toString(false);
   public static final String METRICS_LOG_DIR_KEY = METRICS_CONFIGURATIONS_PREFIX + "log.dir";
   public static final String METRICS_FILE_SUFFIX = METRICS_CONFIGURATIONS_PREFIX + "reporting.file.suffix";
+  public static final String MR_TYPE_KEY = METRICS_CONFIGURATIONS_PREFIX + "mr.type";
```
sv2000, line 711: Since these configuration keys are dynamically set and not user provided, its better to define them inside the MRJobLauncher class.

gobblin-metrics-libs/gobblin-metrics/src/main/java/org/apache/gobblin/metrics/GobblinMetrics.java
arjun4084346: I could not find any API which gives task number and attempt. getTaskAttemptID().getJobID(), getTaskAttemptID().getTaskID(), getTaskAttemptID().getId() did not quite give what I am trying to extract from taskAttemptID.
I'll use context.getTaskAttemptID().getTaskType()

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -771,6 +780,20 @@ protected void setup(Context context) {
...
       }
 
+      // add some more MR task related configs
+
+      String[] tokens = taskAttemptId.toString().split("_");
+      TaskType taskType = taskAttemptId.getTaskType();
+      gobblinJobState.setProp(MR_TYPE_KEY, taskType.name());
+
+      if (taskType.equals(TaskType.MAP)) {
+        gobblinJobState.setProp(MAPPER_TASK_NUM_KEY, tokens[4]);
```
sv2000, line 790: Add a check for the length of the "tokens" array or an assertion that tokens should be of length 6. Further, you may want to access the map/reduce task num as tokens[tokens.length - 1].

PR #3069, opened 2020-07-23
[GOBBLIN-1221] Preserve source file's ModTime by configuration

autumnust:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1221

### Description
This change enabled distcp to preserve the ModTime in destination. It refactored a code block in `CopyPublisher` which is used to preserve the file version. Note that this is required to be placed in publisher since the rename in the publisher is the last operation change the ModTime, while other attributes like Owner/Group will not change after write operation in writer. 

Add a unit test for both positive and negative case.

### Tests

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
@@ -186,8 +186,38 @@ public class CopyDataPublisher extends DataPublisher implements UnpublishedHandl
   }
 
   /**
+   * Unlike other preserving attributes of files (ownership, group, etc.), which is preserved in writer,
+   * some of the attributes have to be set during publish phase like ModTime,
+   * and versionStrategy (usually relevant to modTime as well), since they are subject to change with Publish(rename)
+   */
+  private void preserveFileAttrInPublisher(CopyableFile copyableFile) throws IOException {
+    // Preserving File ModTime, keep the accessing time of the target file.
+    if (copyableFile.getPreserve().preserve(PreserveAttributes.Option.MOD_TIME)) {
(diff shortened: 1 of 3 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -185,9 +185,39 @@ public void handleUnpublishedWorkUnits(Collection<? extends WorkUnitState> state
...
+  /**
+   * Unlike other preserving attributes of files (ownership, group, etc.), which is preserved in writer,
+   * some of the attributes have to be set during publish phase like ModTime,
+   * and versionStrategy (usually relevant to modTime as well), since they are subject to change with Publish(rename)
+   */
+  private void preserveFileAttrInPublisher(CopyableFile copyableFile) throws IOException {
+    // Preserving File ModTime, keep the accessing time of the target file.
+    if (copyableFile.getPreserve().preserve(PreserveAttributes.Option.MOD_TIME)) {
+      fs.setTimes(copyableFile.getDestination(), copyableFile.getOriginTimestamp(),
+          fs.getFileStatus(copyableFile.getDestination()).getAccessTime());
```
sv2000, line 197: Why not add accessTime as a field in CopyableFile and populate it when CopyableFiles are computed? We can avoid the fs.getFileStatus() RPC call here. Also, not sure whether we should preserve the file "accessTime" from the source. We might be better off setting this to -1.

autumnust: Actually I think a bit more into the accessTime, it might makes more sense to set this to -1 since the preservation of ModTime is actually breaking the semantics already (we preserve it for special use cases only) and it doesn't make sense to keep the original access time either.

PR #3070, opened 2020-07-27
[GOBBLIN-1222] Create right abstraction to assemble dataset staging dir for Hive dataset finder

aryavaibhav93:
…asetFinder

Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1222

### Description
- To incorporate the hive datasets as the part of staging directory. With the new configurations created with this PR: while copying hive datasets, staging directory will have dataset and database name as part of it. Therefore, it will help create staging directory within the destination shard while performing data copying onto cloud cluster.

### Tests

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
@@ -199,8 +199,6 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
           datasetFinder instanceof IterableDatasetFinder ? (IterableDatasetFinder<CopyableDatasetBase>) datasetFinder
               : new IterableDatasetFinderImpl<>(datasetFinder);
 
-
-
       Iterator<CopyableDatasetRequestor> requestorIteratorWithNulls = Iterators
           .transform(iterableDatasetFinder.getDatasetsIterator(),
               new CopyableDatasetRequestor.Factory(targetFs, copyConfiguration, log));
@@ -372,7 +370,7 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
-          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.DATASET_STAGING_DIR,false))) {
+          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.IS_DATASET_STAGING_DIR_USED,false))) {
             workUnit.setProp(DATASET_STAGING_DIR_PATH, ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH));
           }
           serializeCopyEntity(workUnit, copyEntity);
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
@@ -32,8 +32,6 @@ import javax.annotation.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
-import org.apache.gobblin.data.management.copy.hive.HiveDataset;
-import org.apache.gobblin.data.management.copy.hive.HiveDatasetFinder;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 
@@ -55,6 +53,8 @@ import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.data.management.copy.extractor.EmptyExtractor;
 import org.apache.gobblin.data.management.copy.extractor.FileAwareInputStreamExtractor;
+import org.apache.gobblin.data.management.copy.hive.HiveDataset;
+import org.apache.gobblin.data.management.copy.hive.HiveDatasetFinder;
 import org.apache.gobblin.data.management.copy.prioritization.FileSetComparator;
 import org.apache.gobblin.data.management.copy.publisher.CopyEventSubmitterHelper;
 import org.apache.gobblin.data.management.copy.replication.ConfigBasedDataset;
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
@@ -32,6 +32,8 @@ import javax.annotation.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+import org.apache.gobblin.data.management.copy.hive.HiveDataset;
+import org.apache.gobblin.data.management.copy.hive.HiveDatasetFinder;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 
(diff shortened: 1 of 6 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -94,6 +96,7 @@
 
   // Will not be serialized/de-serialized
   protected transient final Properties properties;
+  public Properties getProperties() { return properties; }
```
autumnust, line 99: Please move the method into the section below (methods should stay with method. Also you can use lombok annotation for boilerplate code. Check the "@Getter" annotation in the codebase for example)

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -139,17 +141,22 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
     URI uri = URI.create(uriStr);
     this.fs = FileSystem.get(uri, conf);
     this.fileContext = FileContext.getFileContext(uri, conf);
+    this.copyableDatasetMetadata =
+        CopyableDatasetMetadata.deserialize(state.getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
 
     if (state.getPropAsBoolean(ConfigurationKeys.USER_DEFINED_STAGING_DIR_FLAG,false)) {
       this.stagingDir = new Path(state.getProp(ConfigurationKeys.USER_DEFINED_STATIC_STAGING_DIR));
+    } else if ((state.getPropAsBoolean(ConfigurationKeys.DATASET_STAGING_DIR,false))) {
```
autumnust, line 149: Can you add comments on this code block ?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -195,6 +199,8 @@
           datasetFinder instanceof IterableDatasetFinder ? (IterableDatasetFinder<CopyableDatasetBase>) datasetFinder
               : new IterableDatasetFinderImpl<>(datasetFinder);
 
+
```
autumnust, line 202: Please remove the additional blank lines.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -85,6 +85,8 @@
   public static final String DATASET_NAME_PATTERN_KEY = "hive.datasetNamePattern";
   public static final String DATABASE = "Database";
   public static final String TABLE = "Table";
+  public static final String DATASET_STAGING_PATH = "dataset.staging.path";
+  public static final String DATASET_PREFIX_TOBEREPLACED = "hive.dataset.copy.target.table.prefixToBeReplaced";
```
autumnust, line 89: Isn't the same thing already defined somewhere else ?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -125,6 +128,8 @@ public HiveDataset(FileSystem fs, HiveMetastoreClientPool clientPool, Table tabl
         Optional.fromNullable(this.table.getDataLocation());
 
     this.tableIdentifier = this.table.getDbName() + "." + this.table.getTableName();
+    this.datasetStagingDir = properties.getProperty(DATASET_PREFIX_TOBEREPLACED) + "/" + this.table.getDbName() + "/" + this.table.getTableName();
```
autumnust, line 131: This one will not preserve the casing for the HDFS path.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -330,4 +335,8 @@ private boolean canCopyTable() {
     }
     return true;
   }
+
+  public Properties getProperties() {
```
autumnust, line 339: You can use lombok annotation to get rid of this method. Check "@Getter" annotation in the code base.

gobblin-utility/src/main/java/org/apache/gobblin/util/WriterUtils.java
```diff
@@ -100,6 +111,10 @@ public static Path getWriterStagingDir(State state, int numBranches, int branchI
     return new Path(getWriterStagingDir(state, numBranches, branchId), attemptId);
   }
 
+  public static Path getHiveDatasetWriterStagingDir(State state, int numBranches, int branchId, String attemptId) {
```
autumnust, line 114: why do you need this method given there's another `getHiveDatasetWriterStagingDir`?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -366,6 +372,9 @@ public Void call() {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
+          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.DATASET_STAGING_DIR,false))) {
```
autumnust, line 375: Sorry If we have discussed this before, but what's the reason that you couldn't set `DATASET_STAGING_DIR_PATH` in the Hive-specific constructs like `HiveDataset`?

gobblin-utility/src/main/java/org/apache/gobblin/util/WriterUtils.java
```diff
@@ -92,6 +92,17 @@ public static Path getWriterStagingDir(State state, int numBranches, int branchI
         WriterUtils.getWriterFilePath(state, numBranches, branchId));
   }
 
+  public static Path getDatasetWriterStagingDir(State state, int numBranches, int branchId) {
```
autumnust, line 95: You would like to reduce the code duplication here and re-use the method that is already implemented above.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriter.java
```diff
@@ -139,17 +141,28 @@ public FileAwareInputStreamDataWriter(State state, int numBranches, int branchId
     URI uri = URI.create(uriStr);
     this.fs = FileSystem.get(uri, conf);
     this.fileContext = FileContext.getFileContext(uri, conf);
+    this.copyableDatasetMetadata =
```
autumnust, line 144: Why are you moving this code block above ?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -102,6 +105,7 @@
 
   // Only set if table has exactly one location
   protected final Optional<Path> tableRootPath;
+  protected final Path tableLocation;
```
autumnust, line 108: This is just a local variable and don't have to be a field for this class

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -85,6 +85,8 @@
   public static final String DATASET_NAME_PATTERN_KEY = "hive.datasetNamePattern";
   public static final String DATABASE = "Database";
   public static final String TABLE = "Table";
+  public static final String DATASET_STAGING_PATH = "dataset.staging.path";
+  public static final String DATASET_PREFIX_REPLACEMENT = "hive.dataset.copy.target.table.prefixReplacement";
```
autumnust, line 89: Can you use the existing variable for this instead of inventing a new variable ?

PR #3072, opened 2020-08-06
[GOBBLIN-1226] remove lombok from runtime classpath

hkoosha:
### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1226

### Description

Lombok is a cool tool (although IMHO not worth it's troubles, considering generated code is harder to debug, incompatibilities in bytecode, line version mismatch in debugger, etc, etc) which promises to have zero runtime dependencies.

It is also strongly advised by its developers that it should be present on the classpath only during compilation/code generation and not at all, during runtime. (I don't have an exact reference at hand but it can be searched in the lombok google forum).

Lombok in runtime's classpath is considered not only a bad practice but wrong usage of lombok. 

The issue it caused me was that it brings an older version of guava into path.

I propose the `compile ...lombok` in `build.gradle` be changed to `compileOnly`. This has zero effects on any current project using gobblin.

### DIFF
--- a/gobblin-admin/build.gradle
+++ b/gobblin-admin/build.gradle
@@ -32,8 +32,10 @@ dependencies {
     compile externalDependency.commonsCli
     compile externalDependency.slf4j
     compile externalDependency.jodaTime
+    compileOnly externalDependency.lombok
 
     testCompile externalDependency.testng
+    testCompile externalDependency.lombok
 }
 
 ext.classification="library"
--- a/gobblin-admin/build.gradle
+++ b/gobblin-admin/build.gradle
@@ -32,10 +32,8 @@ dependencies {
     compile externalDependency.commonsCli
     compile externalDependency.slf4j
     compile externalDependency.jodaTime
-    compileOnly externalDependency.lombok
 
     testCompile externalDependency.testng
-    testCompile externalDependency.lombok
 }
 
 ext.classification="library"
--- a/gobblin-api/build.gradle
+++ b/gobblin-api/build.gradle
@@ -25,7 +25,6 @@ dependencies {
     compile externalDependency.commonsLang3
     compile externalDependency.slf4j
     compile externalDependency.commonsCli
-    compileOnly externalDependency.lombok
     compile externalDependency.commonsIo
     compile externalDependency.reflections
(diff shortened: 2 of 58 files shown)


jhsenjaliya:
+1 LGTM,
@hkoosha , It would be also good to place this dependencies at parent level, as a core feature available for all modules in Gobblin.
@sv2000 can u pls help review this.

hkoosha:
@jhsenjaliya where would it go? in globalDependencies.gradle?

jhsenjaliya:
@hkoosha you r right.  but if you want to do it. we can probably add `dependencies` section in globalDependencies.gradle that should apply/add dependecies for all modules.

hkoosha:
updated :heavy_check_mark:

jhsenjaliya:
+1, LGTM
@sv2000 , @shirshanka , can u pls take a look at this.

sv2000:
@hkoosha Thanks for this valuable contribution. Hope, this triggers a more comprehensive discussion about recommended usage of Lombok. To that end, it would be great if you can cite references/enumerate specific issues associated with usage of Lombok as a runtime dependency. This should be educational for the wider Gobblin developer community.

sv2000:
@autumnust @htran1 What are your thoughts/opinions on removing Lombok as a runtime dependency?

hkoosha:
Well, one problem could be that lombok may liberally use any dependency as they wish, no matter how outdated. Then including lombok in Gobblin's runtime means needlessly pulling those transitive dependencies into classpath.
 
Why needlessly? because one of lombok promises is: zero runtime dependency. No code path within lombok is accessed during runtime. No other third-party library is used within lombok's generated code either. The generated code is just pure vanilla java.
All those lombok annotations we add to the code, and lombok.* package references are all removed during AST transformation. 
So why include it's jar and its transitive dependencies at all?

autumnust:
+1 on this PR. @hkoosha thanks. 

@sv2000 I think that makes sense.  Reference:
https://projectlombok.org/mavenrepo/

sv2000:
+1. Thanks @hkoosha @jhsenjaliya for this PR!

PR #3081, opened 2020-08-14
[GOBBLIN-1235]Migrate Log4J to SLF4J to provide a clean log environment for downstream users

ZihanLi58:
… still use log4j as implementation

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1235

### Description
change gradle dependencies to migrate log4j to slf4j, and for unit test we still use log4j as log implementation
By applying this change, to make job can still running on hadoop or still use log4j as log implementation, user will need to exclude 'org.slf4j', module: 'log4j-over-slf4j' from their job to avoid endless loop between 'log4j-over-slf4j' and 'slf4j-log4j12'

### Tests
Test on hadoop that this can work, local deploy service and make sure they can use log4j2 as log implementation.

### Commits


gobblin-admin/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     compile externalDependency.jodaTime
 
     testCompile externalDependency.testng
+    testCompile externalDependency.slf4jToLog4j
```
autumnust, line 37: why this is only added for test-compile (isn't that the case we use slf4j all over the places?

autumnust: OK I need a bit more clarification here: 
I think the purpose of adding this line is to add the binding in these test classes to make the log4j being picked as the implementation of slf4j ?

gradle/scripts/dependencyDefinitions.gradle
```diff
@@ -180,13 +180,16 @@ ext.externalDependency = [
     'reactivex': 'io.reactivex.rxjava2:rxjava:2.1.0',
     "slf4j": [
         "org.slf4j:slf4j-api:" + slf4jVersion,
-        "org.slf4j:slf4j-log4j12:" + slf4jVersion,
+        "org.slf4j:log4j-over-slf4j:" + slf4jVersion,
```
autumnust, line 183: OK now I understand it through reading this [link](http://www.slf4j.org/legacy.html#:~:text=SLF4J%20ship%20with%20a%20module,simply%20by%20replacing%20the%20log4j.) !

gradle/scripts/dependencyDefinitions.gradle
```diff
@@ -180,13 +180,16 @@ ext.externalDependency = [
...
         "org.slf4j:jcl-over-slf4j:" + slf4jVersion
     ],
     "log4j": [
```
autumnust, line 186: is this block still needed if you removed references ?

ZihanLi58: We still use log4j in unit test, and I keep the dependency for log4j in gobblin-aws since we specifically use appender which slf4j does not support.

autumnust: Gotcha. You may also want to check the conflicts, after which I can merge this so it could be tested in larger scope.

PR #3083, opened 2020-08-17
[GOBBLIN-1241] Allow nodes/edges in HOCON format and delay resolution to allow overriding

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1241

### Description
1. Allow data nodes/edges to optionally use .conf files for HOCON format instead of only java format
2. Delay resolution of data nodes/edges to allow overriding, but temporarily resolve config when constructing data nodes to allow for example passing fs.uri validation check

### Tests
Tested GaaS locally with .conf files and overriding

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
@@ -68,7 +68,7 @@ public class GitFlowGraphMonitor extends GitMonitoringService {
   public static final String GIT_FLOWGRAPH_MONITOR_PREFIX = "gobblin.service.gitFlowGraphMonitor";
 
   private static final String PROPERTIES_EXTENSIONS = "properties";
-  private static final String CONF_EXTENSIONS = StringUtils.EMPTY;
+  private static final String CONF_EXTENSIONS = "conf";
   private static final String FLOW_EDGE_LABEL_JOINER_CHAR = "_";
   private static final String DEFAULT_GIT_FLOWGRAPH_MONITOR_REPO_DIR = "git-flowgraph";
   private static final String DEFAULT_GIT_FLOWGRAPH_MONITOR_FLOWGRAPH_DIR = "gobblin-flowgraph";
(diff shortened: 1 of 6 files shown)


jack-moseley:
@sv2000 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
```diff
@@ -184,10 +184,13 @@ private void addDataNode(DiffEntry change) {
     if (checkFilePath(change.getNewPath(), NODE_FILE_DEPTH)) {
       Path nodeFilePath = new Path(this.repositoryDir, change.getNewPath());
       try {
-        Config config = loadNodeFileWithOverrides(nodeFilePath);
-        Class dataNodeClass = Class.forName(ConfigUtils.getString(config, FlowGraphConfigurationKeys.DATA_NODE_CLASS,
+        // Temporarily resolve the data node config when constructing it, but leave it unresolved after to allow overriding
+        Config unresolvedConfig = loadNodeFileWithOverrides(nodeFilePath);
+        Config resolvedConfig = unresolvedConfig.resolve();
```
sv2000, line 189: Can the resolution be pushed into the BaseDataNode class?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/BaseDataNode.java
```diff
@@ -39,6 +40,7 @@
   @Getter
   private String id;
   @Getter
+  @Setter
```
sv2000, line 43: We probably do not need a setter, if we do the resolution inside BaseDataNode.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flowgraph/DataNode.java
```diff
@@ -38,6 +38,12 @@
    */
   Config getRawConfig();
 
+  /**
+   * Set raw config. This should only be used in cases where the data node's actual config should be different from the
+   * config it is constructed with (e.g. constructing with a resolved config).
+   */
+  void setRawConfig(Config config);
```
sv2000, line 45: Same comment as earlier.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GitFlowGraphMonitor.java
jack-moseley: I would have done that, but for example the fs URI check happens in `FileSystemDataNode`, not `BaseDataNode`. I did it this way to avoid needing to add this resolution into the constructor of every child data node.

sv2000: Can we add a new member variable resolvedConfig in addition to rawConfig and have resolvedConfig set inside BaseDataNode? the uri check can use the resolvedConfig?

jack-moseley: Sure, that seems like a good solution

sv2000:
@jack-moseley Can you check on travis failure?

PR #3091, opened 2020-08-25
[GOBBLIN-1248] Fix discrepancy between table schema and file schema

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1248

### Description
**Why this PR?**
Previously in streaming pipeline, to avoid race condition on metadata schema, when we do hive registration, we always fetch the latest schema from Kafka SchemaRegistry, since gobblin converter may change the schema, this introduce discrepancy between hive table schema and real file schema, so we need a better way to solve this problem
**Solution**
We leverage the schema creation time in the schema fetched from schema registry adn keep that information when converter the schema. If the schema creation time of writer's schema is the same as latest schema, we will update the table schema to the writer's schema, else we will keep the schema same as existing table's schema
**What's the change**
By default there is no change, to leverage this feature, the pipeline must be configured to use schema registry V2 to fetch schema which contains information about schema creation time.

### Tests
Run streaming pipeline to test hive can works well

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -216,7 +216,7 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
         }
         table.getSd().setSerdeInfo(HiveMetaStoreUtils.getSerDeInfo(spec.getTable()));
       } catch ( IOException e) {
-        log.error(String.format("Error when fetch latest schema for topic %s", topicName), e);
+        log.error(String.format("Error when update latest schema for topic %s", topicName), e);
         throw new IOException(e);
       }
     }
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -147,7 +147,7 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
   //for a partition is immutable
   private final boolean skipDiffComputation;
 
-  private Optional<KafkaSchemaRegistry> schemaRegistry = Optional.absent();
+  protected Optional<KafkaSchemaRegistry> schemaRegistry = Optional.absent();
   private String topicName = "";
   public HiveMetaStoreBasedRegister(State state, Optional<String> metastoreURI) throws IOException {
     super(state);
@@ -201,15 +201,14 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
    * existing table.
    * Note: If there is no schema specified in the table spec, we will directly update the schema to
    * the existing table schema
-   * Note: We treat the creation time as version number of schema, since according to Kafka team,
-   * schema registry allows "out of order registration" of schemas, this means chronological latest is
-   * NOT what the registry considers latest.
+   * Note: We cannot treat the creation time as version number of schema, since schema registry allows
+   * "out of order registration" of schemas, this means chronological latest is NOT what the registry considers latest.
    * @param spec
    * @param table
    * @param existingTable
    * @throws IOException
    */
-  private void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
+  protected void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
 
     if (this.schemaRegistry.isPresent()) {
       try (Timer.Context context = this.metricContext.timer(GET_AND_SET_LATEST_SCHEMA).time()) {
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -200,7 +200,7 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
    * table schema to the writer's schema, else we will keep the table schema the same as schema of
    * existing table.
    * Note: If there is no schema specified in the table spec, we will directly update the schema to
-   * the latest schema fetched from schema registry
+   * the existing table schema
    * Note: We treat the creation time as version number of schema, since according to Kafka team,
    * schema registry allows "out of order registration" of schemas, this means chronological latest is
    * NOT what the registry considers latest.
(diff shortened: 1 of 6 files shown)


ZihanLi58:
@sv2000 @autumnust Can you help review this PR? Thanks!

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +193,30 @@ protected void registerPath(HiveSpec spec) throws IOException {
       throw new IOException(e);
     }
   }
-  //TODO: We need to find a better to get the latest schema
-  private void updateSchema(HiveSpec spec, Table table) throws IOException{
+  private void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
```
sv2000, line 196: It would be good to add javadoc explaining the update behavior.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +193,30 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
 
     if (this.schemaRegistry.isPresent()) {
       try (Timer.Context context = this.metricContext.timer(GET_AND_SET_LATEST_SCHEMA).time()) {
-        String latestSchema = this.schemaRegistry.get().getLatestSchema(topicName).toString();
-        spec.getTable().getSerDeProps().setProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), latestSchema);
+        Schema latestSchema = (Schema) this.schemaRegistry.get().getLatestSchemaByTopic(topicName);
```
sv2000, line 200: It's confusing why we are fetching latest schema from schema registry. Can we not get the schema creation time from the hive table and update schema if the writer schema creation time > table schema creation time?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,47 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
+          spec.getTable()
+              .getSerDeProps()
+              .setProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(),
+                  existingTable.getSerDeProps().getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName()));
+        }
         table.getSd().setSerdeInfo(HiveMetaStoreUtils.getSerDeInfo(spec.getTable()));
-      } catch (SchemaRegistryException | IOException e) {
-        log.error(String.format("Error when fetch latest schema for topic %s", topicName), e);
+      } catch ( IOException e) {
+        log.error(String.format("Error when updating latest schema for topic %s", topicName), e);
```
autumnust, line 235: It is usually an anti-pattern to log.error with exception with IOE thrown at the same time. Please consider keeping one of them.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: According to kafka team, schema registry allows "out of order registration" of schemas - think of this as sorting schemas by compatibility instead of by timestamp. this means chronological latest is NOT what the registry considers latest.  I also include this information in comments to avoid confusing
In addition, I update the PR to first compare the creation time with existing schema, if we see difference, we will then fetch latest schema to get the latest schema creation time. In this way, we can avoid too many calls to schema registry

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: Thanks for pointing out, addressed.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,57 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
     }
   }
-  //TODO: We need to find a better to get the latest schema
-  private void updateSchema(HiveSpec spec, Table table) throws IOException{
+
+  /**
+   * This method is used to update the table schema to the latest schema
+   * It will fetch creation time of the latest schema from schema registry and compare that
+   * with the creation time of writer's schema. If they are the same, then we will update the
+   * table schema to the writer's schema, else we will keep the table schema the same as schema of
```
sv2000, line 200: It is not immediately clear why we should not update table schema if the latest schema from schema registry does not match the writer schema. Am I correct in understanding that we are delaying the schema registration in the expectation that eventually the writer schema will match the latest schema and the table schema is updated then?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,57 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
+            spec.getTable().getSerDeProps().getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), existingTableSchema.toString())));
+        String writerSchemaCreationTime = AvroUtils.getSchemaCreationTime(writerSchema);
+        if(existingSchemaCreationTime == null || existingSchemaCreationTime.equals(writerSchemaCreationTime)) {
+          spec.getTable()
+              .getSerDeProps()
+              .setProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), writerSchema);
+        } else {
+          // If creation time of writer schema does not equal to the existing schema, we compare with schema fetched from
+          // schema registry to determine whether to update the schema
+          Schema latestSchema = (Schema) this.schemaRegistry.get().getLatestSchemaByTopic(topicName);
```
sv2000, line 230: Do we need to get the latest schema from schema registry here? Can we have the converter set the latest schema and have it accessed here instead of fetching the schema twice from schema registry, once inside the converter and once here?

ZihanLi58: Synced offline and apply new solution

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,49 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
+
+  /**
+   * This method is used to update the table schema to the latest schema
+   * It will fetch creation time of the latest schema from schema registry and compare that
+   * with the creation time of writer's schema. If they are the same, then we will update the
+   * table schema to the writer's schema, else we will keep the table schema the same as schema of
+   * existing table.
+   * Note: If there is no schema specified in the table spec, we will directly update the schema to
+   * the existing table schema
+   * Note: We treat the creation time as version number of schema, since according to Kafka team,
```
sv2000, line 204: Is this comment accurate? Do you mean: We cannot use creation time as the schema version number?

Also: drop "according to Kafka team" from the comment.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,49 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
+        // If no schema set for the table spec, we fall back to existing schema
+        Schema writerSchema = new Schema.Parser().parse((
+            spec.getTable().getSerDeProps().getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), existingTableSchema.toString())));
+        String writerSchemaCreationTime = AvroUtils.getSchemaCreationTime(writerSchema);
+        if(existingSchemaCreationTime != null && !existingSchemaCreationTime.equals(writerSchemaCreationTime)) {
+          // If creation time of writer schema does not equal to the existing schema, we compare with schema fetched from
+          // schema registry to determine whether to update the schema
+          Schema latestSchema = (Schema) this.schemaRegistry.get().getLatestSchemaByTopic(topicName);
+          String latestSchemaCreationTime = AvroUtils.getSchemaCreationTime(latestSchema);
+          if (latestSchemaCreationTime != null && latestSchemaCreationTime.equals(existingSchemaCreationTime)) {
```
sv2000, line 228: Shouldn't there be an else block that updates the table schema if latestSchemaCreationTime does not equal existingSchemaCreationTime?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -192,16 +192,49 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
+   * the existing table schema
+   * Note: We treat the creation time as version number of schema, since according to Kafka team,
+   * schema registry allows "out of order registration" of schemas, this means chronological latest is
+   * NOT what the registry considers latest.
+   * @param spec
+   * @param table
+   * @param existingTable
+   * @throws IOException
+   */
+  private void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
```
sv2000, line 212: Is it possible to add unit tests for this method?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: By default, the table schema is writer schema, so we only need to update it if we don't want to change schema and set it to be the existing schema

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -201,15 +201,14 @@ protected void registerPath(HiveSpec spec) throws IOException {
...
-   * NOT what the registry considers latest.
+   * Note: We cannot treat the creation time as version number of schema, since schema registry allows
+   * "out of order registration" of schemas, this means chronological latest is NOT what the registry considers latest.
    * @param spec
    * @param table
    * @param existingTable
    * @throws IOException
    */
-  private void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
+  protected void updateSchema(HiveSpec spec, Table table, HiveTable existingTable) throws IOException{
```
sv2000, line 211: Nit: Would be annotate as VisibleForTesting.

PR #3097, opened 2020-09-04
GOBBLIN-1257: Fix the handling of collection field types during ORC s…

sv2000:
…chema up-conversion in compaction

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1257

### Description
Currently, on schema change while copying map/list entries from the original schema into the evolved schema, all the entries in the collection point to the same reference. This is inadvertently caused by re-using the same WritableComparable object for all the values. In addition to fixing this issue, this PR:
1. Refactors/enhances existing unit tests, and
2. Introduces a new convenience class for local testing of the compaction job and provides a simple way of reproducing compaction issues locally. This class requires users to provide input data records in json format along with an input ORC schema. It generates the corresponding input ORC files and runs the compaction job against the generated input. 

### Tests
Added tests to OrcUtilsTest class.

### Commits

### DIFF
--- a/gobblin-compaction/build.gradle
+++ b/gobblin-compaction/build.gradle
@@ -31,6 +31,7 @@ dependencies {
   // with hive-exec-core in older version(1.0.1), we need to shadow ord-mapreduce's transitive deps.
   // and include direct orc-mapreduce library just as a compileOnly dependency
   compileOnly externalDependency.orcMapreduce
+  compile externalDependency.orcTools
   testCompileOnly externalDependency.orcMapreduce
   compile project(path: ":gobblin-modules:gobblin-orc-dep", configuration:"shadow")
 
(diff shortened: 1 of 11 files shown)


sv2000:
@autumnust Please review

gobblin-compaction/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
   // with hive-exec-core in older version(1.0.1), we need to shadow ord-mapreduce's transitive deps.
   // and include direct orc-mapreduce library just as a compileOnly dependency
   compileOnly externalDependency.orcMapreduce
+  compile externalDependency.orcTools
```
autumnust, line 34: Should it be `testCompile`?

autumnust (changes requested):
Minor comment

gobblin-compaction/build.gradle
sv2000: Since the utility class that allows users to run compaction locally is not a unit test, I had it located in the src/main of gobblin-compaction module. Also, this dependency will be needed at runtime, hence I chose "compile".

autumnust: oh OK I didn't notice that ( utility class in src instead of test).

autumnust (approved):
LGTM

PR #3099, opened 2020-09-10
[GOBBLIN-1256]Exclude log4j related jars in compile but include those in test

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1256

### Description
This PR exclude all the log4j related jar from our published artifacts except gobblin-aws to provide a clean dependency for user. After this change, user who depend on gobblin will need to specify the log implementation in their project, i.e log4jV1 or log4jV2. 
Also this pr remain the log4j related dependency in unit test since we still use Log4j API to test. 

### Tests
Run unit test
Publish snapshot version and verify that we don't have log4j jars in the transitive dependency.

### Commits

### DIFF
--- a/gradle/scripts/globalDependencies.gradle
+++ b/gradle/scripts/globalDependencies.gradle
@@ -21,6 +21,7 @@ import javax.tools.ToolProvider
 subprojects {
   plugins.withType(JavaPlugin) {
     configurations {
+      customTestCompile
       compile
       dependencies {
         if (!project.name.contains('gobblin-elasticsearch-deps')) {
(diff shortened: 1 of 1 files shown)


ZihanLi58:
@autumnust Can you help take a look at this change? Thanks!

gradle/scripts/globalDependencies.gradle
```diff
@@ -35,20 +36,36 @@ subprojects {
...
+          //exclude this jar because we introduce log4j-over-slf4j, these two jars cannot present at the same time
+          it.exclude group: 'org.slf4j', module: 'slf4j-log4j12'
+          //exclude log4j related jars to provide a clean log environment
+          it.exclude group: 'log4j', module: 'log4j'
+          it.exclude group: 'log4j', module: 'apache-log4j-extras'
+        }
+      }
       all*.exclude group: 'org.apache.calcite', module: 'calcite-avatica' // replaced by org.apache.calcite.avatica:avatica-core
-      //exclude this jar because we introduce log4j-over-slf4j, these two jars cannot present at the same time
-      compileClasspath.exclude group: 'org.slf4j', module: 'slf4j-log4j12'
```
autumnust, line 48: I might miss something obvious here: why this line (used for exclusion) is no longer required?

autumnust (changes requested):
I don't see the part doing "excluding log4j related jar" in this PR but only the part of `customTestCompile` being added to ensure log4j being used in unit tests.

autumnust:
It is also worth noting that tests like `LoggingPusherTest` will not work if running in IntellIJ without removing bridging jar from external library directories. So it will be better to document this kind of "Gotcha" as well.

gradle/scripts/globalDependencies.gradle
ZihanLi58: On Line51 (configurations.compile.dependencies*.each ) I add the config to exclude slf4j-log4j12 and log4j related jar.  The reason for using this statement is that I find when we call compile.exclude to exclude some jar, the published artifacts will still contains that transient dependency. So I use this statement to make sure our published artifacts do not contains log4j related dependency

ZihanLi58:
> It is also worth noting that tests like `LoggingPusherTest` will not work if running in IntellIJ without removing bridging jar from external library directories. So it will be better to document this kind of "Gotcha" as well.

I haven't meet this error in intellij, maybe it randomly picks the right jar for me, will add this into the document

gradle/scripts/globalDependencies.gradle
autumnust: Got it. Missed exclamation mark ...

autumnust (approved):
LGTM.

PR #3100, opened 2020-09-10
[GOBBLIN-1260] add some logs in GobblinHelixTask

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1260

### Description
add some logs to debug if helix task cancel is throwing exceptions or hanging

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
@@ -220,7 +220,7 @@ public class GobblinHelixTask implements Task {
         log.info("Gobblin helix task cancellation completed for jobId {}.", jobId);
       } catch (Throwable t) {
         log.info("Gobblin helix task cancellation for jobId {} failed with exception.", jobId, t);
-        throw new RuntimeException("Gobblin helix task cancellation for jobId " + jobId + " failed with exception.", t);
+        Throwables.propagate(t);
       }
     } else {
       log.warn("Cancel called for an uninitialized Gobblin helix task for jobId {}.", jobId);
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
@@ -220,7 +220,7 @@ public class GobblinHelixTask implements Task {
         log.info("Gobblin helix task cancellation completed for jobId {}.", jobId);
       } catch (Throwable t) {
         log.info("Gobblin helix task cancellation for jobId {} failed with exception.", jobId, t);
-        throw t;
+        throw new RuntimeException("Gobblin helix task cancellation for jobId " + jobId + " failed with exception.", t);
       }
     } else {
       log.warn("Cancel called for an uninitialized Gobblin helix task for jobId {}.", jobId);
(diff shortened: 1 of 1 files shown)


zxcware (approved):
lgtm

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
```diff
@@ -213,9 +213,17 @@ private Integer getPartitionForHelixTask(TaskDriver taskDriver) {
...
   public void cancel() {
-    log.info("Gobblin helix task cancellation invoked.");
+    log.info("Gobblin helix task cancellation invoked for jobId {}.", jobId);
     if (this.task != null ) {
-      this.task.cancel();
+      try {
+        this.task.cancel();
+        log.info("Gobblin helix task cancellation completed for jobId {}.", jobId);
+      } catch (Throwable t) {
+        log.info("Gobblin helix task cancellation for jobId {} failed with exception.", jobId, t);
```
sv2000, line 222: As @autumnust has pointed out to me, the pattern of catching an exception to log and re-throw the same exception is an anti-pattern. I would recommend reading this SO post: https://softwareengineering.stackexchange.com/questions/365427/try-catch-log-rethrow-is-anti-pattern
In your case, I think you are catching the exception to log the jobId. In this case, per the SO post, it is recommended to wrap the original exception into another exception with the additional context and propagate it up the call stack.

sv2000, line 222: I don't think this log line is needed, given that the RuntimeException already has the context. This will pollute the logs with duplicate stacktraces.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
```diff
@@ -213,9 +213,17 @@ private Integer getPartitionForHelixTask(TaskDriver taskDriver) {
...
+        throw new RuntimeException("Gobblin helix task cancellation for jobId " + jobId + " failed with exception.", t);
```
sv2000, line 223: Why not wrap it into a Throwable instead of a RuntimeException?

sv2000 (changes requested):

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
arjun4084346: I do not intend to just catch and throw the Throwable. I wanted to log the exception in gobblin's log. This method is called from helix class and I do not want to depend on helix code on exception handling.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTask.java
arjun4084346: the method signature does not throw any exception. RuntimeException is Unchecked exception and can be thrown without changing the signature. Throwable is not Unchecked exception.

sv2000: Ah I see. Makes sense.

sv2000: Can we just do Throwables.propagate(t), which essentially wraps the throwable into a RuntimeException?

sv2000 (approved):

PR #3108, opened 2020-09-21
[GOBBLIN-1268] track WORK_UNITS_PREPARATION timer event also in gaas

arjun4084346:
Dear Gobblin maintainers,

@sv2000 please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1268

### Description
1) this PR moves the code block for setting number of workunits in job state at a more appropriate place
2) make gaas job status monitor track WORK_UNITS_PREPARATION timer event also. this event is emitted after the job start timer. the reason for tracking this is that WORK_UNITS_PREPARATION will have NUM_WORKUNITS set which are not set by the time job start event is emitted

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -446,10 +446,6 @@ public abstract class AbstractJobLauncher implements JobLauncher {
           return;
         }
 
-        // If it is a streaming source, workunits cannot be counted
-        this.jobContext.getJobState().setProp(NUM_WORKUNITS,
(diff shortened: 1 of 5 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -493,6 +489,10 @@ public void forWorkUnit(WorkUnit workUnit) {
             }
           });
 
+          // If it is a streaming source, workunits cannot be counted
```
sv2000, line 492: Is there a unit test we can add for this? It seems like the previous location of this block was incorrect. To prevent future regression, it would be good to add a unit test for this.

sv2000 (approved):
+1. LGTM.

PR #3109, opened 2020-09-21
[GOBBLIN-1269] add metrics field in JobStatus schema

arjun4084346:
Dear Gobblin maintainers,

@sv2000  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1269

### Description
add a field to JobStatus schema to add metrics related information for the job

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatus.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatus.java
@@ -40,7 +40,8 @@ public class JobStatus {
   private final long startTime;
   private final long endTime;
   @Setter
-  private String message;
+  private String metrics;
+  private final String message;
   private final long processedCount;
   private final String lowWatermark;
   private final String highWatermark;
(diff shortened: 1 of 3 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatus.java
```diff
@@ -40,7 +40,8 @@
   private final long startTime;
   private final long endTime;
   @Setter
-  private String message;
+  private String metrics;
```
sv2000, line 43: I think the Setter is needed for the message field too.

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatus.java
arjun4084346: Actually I added the setter in https://github.com/apache/incubator-gobblin/pull/3076 to add metrics in message field, so it can be set in internal projects. But now I am adding a separate metrics field for that purpose, which is to be set by custom job status retriever.

sv2000 (approved):
+1. LGTM.

PR #3112, opened 2020-09-25
[GOBBLIN-1272] Fixing bug in loading config-store when the entry is empty

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1272

### Description
A tiny PR that handle the case when there's empty string specified in an entry of Config-Store properly.

### Tests

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/ConfigStoreUtils.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/ConfigStoreUtils.java
@@ -20,13 +20,20 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.fs.Path;
+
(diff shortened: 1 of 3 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/ConfigStoreUtils.java
```diff
@@ -260,8 +262,11 @@ public static void setTopicsFromConfigStore(Properties properties, Set<String> b
   }
 
   public static List<String> getListOfValuesFromConfigStore(Config config, String keyValue) {
```
sv2000, line 264: Doesn't ConfigUtils#getStringList() achieve the same? If so, should we reuse the method defined there?

sv2000, line 264: Also - it seems like the more appropriate name of the 2nd argument should be key.

sv2000 (changes requested):

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/ConfigStoreUtils.java
autumnust: That's a good catch, will address.

sv2000 (approved):
+1. LGTM.

PR #3124, opened 2020-10-12
[GOBBLIN-1285] Add option to only accept requests to leader node and redirect requests in the client

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1285

### Description
- Modify helix instance name to have the current host name appended to it
- Add option where, if true, all requests sent to create/update/delete a flowConfig/flowExecution must be sent to the leader node. If they are not, an error will be thrown pointing to the leader URL.
- Add retry in clients for flowConfig/flowExecution, where if the above error is thrown, the client will retry by sending the same request directly to the leader node.

### Tests
Tested in gaas locally

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowClientUtils.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowClientUtils.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import com.linkedin.r2.RemoteInvocationException;
+import com.linkedin.r2.message.RequestContext;
+import com.linkedin.restli.client.Request;
+import com.linkedin.restli.client.Response;
+import com.linkedin.restli.client.RestClient;
+import com.linkedin.restli.client.RestLiResponseException;
+
+
+/**
+ * Utils to be used by clients
+ */
+public class FlowClientUtils {
+  /**
+   * Send a restli {@link Request} to the server through a {@link RestClient}, but if the request is rejected due to not
+   * being sent to a leader node, get the leader node from the errorDetails and retry the request with that node by setting
+   * the D2-Hint-TargetService attribute.
+   * @param restClient rest client to use to send the request
+   * @param request request to send
+   * @param primaryResource resource part of the request URL (e.g. flowconfigsV2, which can be taken from
+   *        {@link FlowconfigsV2RequestBuilders#getPrimaryResource()}
+   * @return {@link Response} returned from the request
+   * @throws RemoteInvocationException
+   */
+  public static Response<?> sendRequestWithRetry(RestClient restClient, Request<?> request, String primaryResource) throws RemoteInvocationException {
+    Response<?> response;
+    try {
+      response = restClient.sendRequest(request).getResponse();
+    } catch (RestLiResponseException exception) {
+      if (exception.hasErrorDetails() && exception.getErrorDetails().containsKey(ServiceConfigKeys.LEADER_URL)) {
+        String leaderUrl = exception.getErrorDetails().getString(ServiceConfigKeys.LEADER_URL);
+        RequestContext requestContext = new RequestContext();
+        try {
+          requestContext.putLocalAttr("D2-Hint-TargetService", new URI(leaderUrl + "/" + primaryResource));
+        } catch (URISyntaxException e) {
+          throw new RuntimeException("Could not build URI for for url " + leaderUrl, e);
+        }
+        response = restClient.sendRequest(request, requestContext).getResponse();
+      } else {
+        throw exception;
+      }
+    }
+    return response;
+  }
+}
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
@@ -17,22 +17,12 @@
 
 package org.apache.gobblin.service;
 
-import java.util.Comparator;
-import java.util.Iterator;
 import java.util.List;
-import java.util.stream.Collectors;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.common.base.Strings;
 import com.google.inject.Inject;
-import com.linkedin.data.template.SetMode;
 import com.linkedin.restli.common.ComplexResourceKey;
 import com.linkedin.restli.common.EmptyRecord;
-import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.server.PagingContext;
-import com.linkedin.restli.server.RestLiServiceException;
 import com.linkedin.restli.server.UpdateResponse;
 import com.linkedin.restli.server.annotations.Context;
 import com.linkedin.restli.server.annotations.Finder;
@@ -41,21 +31,16 @@ import com.linkedin.restli.server.annotations.QueryParam;
 import com.linkedin.restli.server.annotations.RestLiCollection;
 import com.linkedin.restli.server.resources.ComplexKeyResourceTemplate;
 
-import org.apache.gobblin.service.monitoring.FlowStatusGenerator;
-import org.apache.gobblin.service.monitoring.JobStatusRetriever;
-
 
 /**
  * Resource for handling flow execution requests
  */
 @RestLiCollection(name = "flowexecutions", namespace = "org.apache.gobblin.service", keyName = "id")
 public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatusId, EmptyRecord, FlowExecution> {
-  private static final Logger LOG = LoggerFactory.getLogger(FlowExecutionResource.class);
-  public static final String FLOW_STATUS_GENERATOR_INJECT_NAME = "FlowStatusGenerator";
-  public static final String MESSAGE_SEPARATOR = ", ";
+  public static final String FLOW_EXECUTION_GENERATOR_INJECT_NAME = "FlowExecutionResourceHandler";
 
-  @Inject @javax.inject.Inject @javax.inject.Named(FLOW_STATUS_GENERATOR_INJECT_NAME)
-  FlowStatusGenerator _flowStatusGenerator;
+  @Inject @javax.inject.Inject @javax.inject.Named(FLOW_EXECUTION_GENERATOR_INJECT_NAME)
+  FlowExecutionResourceHandler flowExecutionResourceHandler;
 
   public FlowExecutionResource() {}
 
@@ -66,25 +51,13 @@ public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatus
    */
   @Override
   public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
-    FlowExecution flowExecution = convertFlowStatus(getFlowStatusFromGenerator(key, this._flowStatusGenerator));
-    if (flowExecution == null) {
-      throw new RestLiServiceException(HttpStatus.S_404_NOT_FOUND, "No flow execution found for flowStatusId " + key.getKey()
-      + ". The flowStatusId may be incorrect, or the flow execution may have been cleaned up.");
-    }
-    return flowExecution;
+    return this.flowExecutionResourceHandler.get(key);
   }
 
   @Finder("latestFlowExecution")
   public List<FlowExecution> getLatestFlowExecution(@Context PagingContext context, @QueryParam("flowId") FlowId flowId,
       @Optional @QueryParam("count") Integer count, @Optional @QueryParam("tag") String tag, @Optional @QueryParam("executionStatus") String executionStatus) {
-    List<org.apache.gobblin.service.monitoring.FlowStatus> flowStatuses = getLatestFlowStatusesFromGenerator(flowId, count, tag, executionStatus, this._flowStatusGenerator);
-
-    if (flowStatuses != null) {
-      return flowStatuses.stream().map(FlowExecutionResource::convertFlowStatus).collect(Collectors.toList());
-    }
-
-    throw new RestLiServiceException(HttpStatus.S_404_NOT_FOUND, "No flow execution found for flowId " + flowId
-        + ". The flowId may be incorrect, or the flow execution may have been cleaned up.");
+    return this.flowExecutionResourceHandler.getLatestFlowExecution(context, flowId, count, tag, executionStatus);
   }
 
   /**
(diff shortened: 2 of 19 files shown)


gobblin-runtime/build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
   compile externalDependency.gson
   compile externalDependency.guava
   compile externalDependency.guice
+  compile externalDependency.helix
```
sv2000, line 69: Hmm. I would avoid Helix dependency in gobblin-runtime.

sv2000: Can we move FlowStatusGenerator to gobblin-service module?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.apache.gobblin.service.ServiceConfigKeys;
+import org.apache.gobblin.util.ConfigUtils;
+
+
+/**
+ * Utils for storing/parsing helix URL in the helix instance name
+ */
+@Slf4j
+public class HelixLeaderUtils {
```
sv2000, line 44: Can this be moved into gobblin-service?

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * If this host is not the leader, throw a {@link RestLiServiceException}, and include the URL of the leader host in
+   * the message and in the errorDetails under the key {@link ServiceConfigKeys#LEADER_URL}.
+   */
+  public static void throwErrorIfNotLeader(Optional<HelixManager> helixManager)  {
+    if (helixManager.isPresent() && !helixManager.get().isLeader()) {
+      String leaderUrl = getLeaderUrl(helixManager.get());
+      if (leaderUrl == null) {
+        return;
```
sv2000, line 71: Should we throw an error here?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowClientUtils.java
```diff
@@ -0,0 +1,66 @@
...
+   * being sent to a leader node, get the leader node from the errorDetails and retry the request with that node by setting
+   * the D2-Hint-TargetService attribute.
+   * @param restClient rest client to use to send the request
+   * @param request request to send
+   * @param primaryResource resource part of the request URL (e.g. flowconfigsV2, which can be taken from
+   *        {@link FlowconfigsV2RequestBuilders#getPrimaryResource()}
+   * @return {@link Response} returned from the request
+   * @throws RemoteInvocationException
+   */
+  public static Response<?> sendRequestWithRetry(RestClient restClient, Request<?> request, String primaryResource) throws RemoteInvocationException {
```
sv2000, line 46: Can we add unit tests? We should be able to add unit tests in GobblinServiceHATest.

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Build helix instance name by getting {@link org.apache.gobblin.service.ServiceConfigKeys#HELIX_INSTANCE_NAME_KEY}
+   * and appending the host, port, and service name with a separator
+   */
+  public static String buildHelixInstanceName(Config config, String defaultInstanceName) {
+    String helixInstanceName = ConfigUtils.getString(config, ServiceConfigKeys.HELIX_INSTANCE_NAME_KEY, defaultInstanceName);
+
+    String url = "";
+    try {
+      url = HELIX_INSTANCE_NAME_SEPARATOR + "https://" + InetAddress.getLocalHost().getHostName() + ":" + ConfigUtils.getString(config, ServiceConfigKeys.SERVICE_PORT, "")
```
sv2000, line 88: Also - should we always assume https? Can this be inferred from service configs i.e. http vs https.

gobblin-runtime/build.gradle
jack-moseley: I did try a lot of different locations for these things... the problem with doing that is that `FlowExecutionResource` needs `FlowStatusGenerator` as well, and I can't add gobblin-service as a dependency because gobblin-service depends on gobblin-restli.

Maybe I can try moving `FlowStatusGenerator` to gobblin-restli, that seems like it should be possible.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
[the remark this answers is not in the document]
jack-moseley: Do you think it's better to keep it backwards compatible? These configs were not required in the past, I didn't want block gaas from starting up if the version is bumped without adding these configs.

gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/FlowStatusGenerator.java
```diff
@@ -176,6 +182,9 @@ private boolean isJobRunning(JobStatus jobStatus) {
    * Send kill request for the given flow
    */
   public void killFlow(String flowGroup, String flowName, Long flowExecutionId) {
```
sv2000, line 184: Quick question: Why is there a killFlow() method in FlowStatusGenerator? Can this method be relocated into FlowExecutionResource which seems like the right location? Looks like you only want the eventBus instance in FlowExecutionResource, which can be injected the same way that FlowStatusGenerator is.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
jack-moseley: Moved to gobblin-restli

gobblin-runtime/build.gradle
jack-moseley: Moved the killflow logic to flowExecutionResource so it's not necessary.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/HelixLeaderUtils.java
jack-moseley: Added a config for this and default to https://

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
```diff
@@ -52,10 +55,19 @@
...
+  public static final String EVENTBUS_INJECT_NAME = "EventBus";
+  public static final String HELIX_MANAGER_INJECT_NAME = "HelixManager";
+  public static final String FORCE_LEADER_INJECT_NAME = "ForceLeader";
   public static final String MESSAGE_SEPARATOR = ", ";
 
   @Inject @javax.inject.Inject @javax.inject.Named(FLOW_STATUS_GENERATOR_INJECT_NAME)
   FlowStatusGenerator _flowStatusGenerator;
+  @Inject @javax.inject.Inject @javax.inject.Named(EVENTBUS_INJECT_NAME)
+  EventBus eventBus;
+  @Inject @javax.inject.Inject @javax.inject.Named(HELIX_MANAGER_INJECT_NAME)
```
sv2000, line 67: It seems like we don't really need a HelixManager instance here and the dependency on helix module can be removed. Can HelixLeaderUtils create a singleton HelixManager instance that it uses to check the leader status?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/HelixLeaderUtils.java
```diff
@@ -0,0 +1,105 @@
...
+
+
+/**
+ * Utils for storing/parsing helix URL in the helix instance name
+ */
+@Slf4j
+public class HelixLeaderUtils {
+  public static String HELIX_INSTANCE_NAME_SEPARATOR = "@";
+
+  /**
```
sv2000, line 49: Empty Javadoc here. You may want to remove it if not needed.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceRedirectTest.java
```diff
@@ -0,0 +1,301 @@
...
+    commonServiceCoreProperties.put(ServiceConfigKeys.SERVICE_NAME, SERVICE_NAME);
+
+    node1ServiceCoreProperties = new Properties();
+    node1ServiceCoreProperties.putAll(commonServiceCoreProperties);
+    node1ServiceCoreProperties.put(ConfigurationKeys.TOPOLOGYSPEC_STORE_DIR_KEY, NODE_1_TOPOLOGY_SPEC_STORE_DIR);
+    node1ServiceCoreProperties.put(FlowCatalog.FLOWSPEC_STORE_DIR_KEY, NODE_1_FLOW_SPEC_STORE_DIR);
+    node1ServiceCoreProperties.put(FsJobStatusRetriever.CONF_PREFIX + "." + ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY, NODE_1_JOB_STATUS_STATE_STORE_DIR);
+    node1ServiceCoreProperties.put(QUARTZ_INSTANCE_NAME, "QuartzScheduler1");
+    node1ServiceCoreProperties.put(QUARTZ_THREAD_POOL_COUNT, 3);
+    node1ServiceCoreProperties.put(ServiceConfigKeys.SERVICE_PORT, PORT1);
```
sv2000, line 163: Should this be restliServer.getPort() instead of PORT1?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceRedirectTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+    node2ServiceCoreProperties = new Properties();
+    node2ServiceCoreProperties.putAll(commonServiceCoreProperties);
+    node2ServiceCoreProperties.put(ConfigurationKeys.TOPOLOGYSPEC_STORE_DIR_KEY, NODE_2_TOPOLOGY_SPEC_STORE_DIR);
+    node2ServiceCoreProperties.put(FlowCatalog.FLOWSPEC_STORE_DIR_KEY, NODE_2_FLOW_SPEC_STORE_DIR);
+    node2ServiceCoreProperties.put(FsJobStatusRetriever.CONF_PREFIX + "." + ConfigurationKeys.STATE_STORE_ROOT_DIR_KEY, NODE_2_JOB_STATUS_STATE_STORE_DIR);
+    node2ServiceCoreProperties.put(QUARTZ_INSTANCE_NAME, "QuartzScheduler2");
+    node2ServiceCoreProperties.put(QUARTZ_THREAD_POOL_COUNT, 3);
+    node2ServiceCoreProperties.put(ServiceConfigKeys.SERVICE_PORT, PORT2);
```
sv2000, line 172: Same comment as above.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceRedirectTest.java
```diff
@@ -0,0 +1,301 @@
...
+public class GobblinServiceRedirectTest {
+
+  private static final Logger logger = LoggerFactory.getLogger(GobblinServiceRedirectTest.class);
+
+  private static final String QUARTZ_INSTANCE_NAME = "org.quartz.scheduler.instanceName";
+  private static final String QUARTZ_THREAD_POOL_COUNT = "org.quartz.threadPool.threadCount";
+
+  private static final String COMMON_SPEC_STORE_PARENT_DIR = "/tmp/serviceCoreCommon/";
+
+  private static final String NODE_1_SERVICE_WORK_DIR = "/tmp/serviceWorkDirNode1/";
```
sv2000, line 66: Can we dynamically generate tmp dirs using Guava Files.createTempDir() method? And then mark the file to be deleted using deleteOnExit(). Hardcoding the location to /tmp has the potential to create flaky tests due to left over state from previous builds in case of un-finished clean up. Also, it may end up interfering with identical locations used in other test classes.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
```diff
@@ -52,10 +55,19 @@
 public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatusId, EmptyRecord, FlowExecution> {
```
sv2000, line 55: One naive question: Can we create a FlowExecutionResourceHandler interface with impls inside gobblin-service module and inject the handler here? Essentially, replicating the pattern we used for FlowConfigs resource, and making the code uniform.

sv2000 (changes requested):
Some additional comments.

sv2000:
@jack-moseley Can you check on the travis build failures too?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
jack-moseley: I guess that is possible, at least it will be the same pattern between flow config and flow execution then, and we can pass the helixmanager/eventbus directly to it instead of injecting them.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceRedirectTest.java
jack-moseley: Can't really do it that way since the restliServer/port is not generated until gobblinservicemanager is instantiated (which already uses this property). Doing it this way just predefines port that will be used by each manager instead of it being random.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
jack-moseley: Added a change to:
- Add `FlowExecutionResourceHandler`
- Add `GobblinServiceFlowExecutionResourceHandler` which just calls the underlying one, except for delete where it does the leadership check
- Merge the `HelixLeaderUtils` methods into `HelixUtils` in gobblin-service and remove the extra helix dependencies added earlier
- Pass eventbus/helixmanager/flowstatusgenerator directly to flowExecutionResourceHandler instead of injecting them, and also switch to building flowstatusgenerator here instead of internally (since there is no reason for it to be internal)

sv2000 (approved):
+1. LGTM. Thanks for this PR! Well done.

jack-moseley:
Added commit to fix findbugs and tests.

Also checking in reordered restli snapshot.json files, I think there was a restli change where these are in alphabetical order when generated, so they will keep showing up as modified in git after building unless we check in the updated one.

sv2000:
@jack-moseley there seems to be some merge conflict here. Can you check?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceRedirectTest.java
```diff
@@ -90,8 +90,8 @@
   private static final String TEST_SOURCE_NAME = "testSource";
   private static final String TEST_SINK_NAME = "testSink";
 
-  private static final String PORT1 = "1000";
-  private static final String PORT2 = "2000";
+  private static final String PORT1 = "10000";
```
sv2000, line 93: You can checkout PortUtils class that provides a method for handing out random unused ports. Static port values run the risk of port conflicts.

jack-moseley:
Switched to PortUtils and squashed commits

shirshanka:
@jack-moseley : do you have the right JIRA on the PR?

jack-moseley:
@shirshanka good catch, I got the wrong jira somehow. Updated the title and description

PR #3128, opened 2020-10-16
[GOBBLIN-1290] Auto-tune ORC writer parameters

autumnust:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1290

### Description
Here is a brief intro for the memory consumption within the streaming ingestor: 

Physical Memory (Usually 4G)
- 50% for ORC writer (2G): This is not a hard cap, it is primarily used for scaling down the threshold for each writer registered in the memory manager. So it is possible to exceed this limit if the number of writers increases when, for example, it needs to deal with portions of late data. 
During normal traffic (non catching up/backfilling mode), having 2~3 writers running in parallel is quite normal due to hourly-boundary interval and late data arrival.
- Metrics objects and other Gobblin overhead : Usually around ~400MB, Typical consumer:
   - TaskExecutor
   - Kafka Extractor 
Non heap memory overhead (600MB, configurable) 
	
	
Given the approximate data points above, we should have no more than ~500MB raw buffer space for each writer. Given the default number of rows between memory check (5000), that means record size should be smaller than 100K. The record size we measured in the streaming pipeline comes from serialized key size plus serialized value size, when they are stored in TreeWriter, the value could be a bit more larger given the object overhead.  Therefore there's a scaling factor being used to indicate this. 

The equation for the auto-tuning is: 
#rowBetweenCheck = actualHeapSize * ratio (50%) / parallelsimOfWriters / (serializedRecordSize * scalingFactor) 

Tested with several heavy topics internally and it works well. 

Other than the major changes, also fix two small issues: 
- build.gradle in gobblin-service module hits a problem and resolved by referring to: https://stackoverflow.com/questions/36079404/errorcannot-configure-the-publishing-extension-after-it-has-been-accessed 
- update gitignore file

### Tests

### Commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 **/*.iws
 **/*.ipr
 .shelf/
+.ideaDataSources/
 
 # VS Code related 
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ out/
 */bin/
 **/mainGeneratedDataTemplate
 **/mainGeneratedRest
+**/main/snapshot
 gobblin-dist/
 gobblin-distribution-*.tar.gz
 gobblin.tar.gz
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -502,6 +502,10 @@ public class ConfigurationKeys {
   public static final boolean DEFAULT_SCHEMA_IN_SOURCE_DIR = false;
   public static final String SCHEMA_FILENAME = "schema.filename";
   public static final String DEFAULT_SCHEMA_FILENAME = "metadata.json";
+  // An optional configuration for extractor's specific implementation to set, which helps data writer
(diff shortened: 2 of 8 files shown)


sv2000 (changes requested):
Thanks for this PR! It highlights the challenges of dealing with memory-intensive file formats such as ORC and Parquet. The problem is definitely trickier than I would have originally imagined. One high-level concern I have is that with this PR we are replacing one static configuration i.e. rowBatchSize with two new static configurations e.g. PARALLEL_WRITERS and RECORD_SIZE_SCALE_FACTOR. It seems that each of these configs could themselves vary depending on the input data and configuration. E.g. PARALLEL_WRITERS could be higher if the threshold on late data is increased in the case of Gobblin Kafka. Also, the scale factor could be different based on narrow vs wide schemas. One question to consider: should the fix actually be implemented in the ORC core library to allow for flushing rowbatches based on memory usage. This way, we only need to set one config to limit the memory used by a single ORC writer.

sv2000 (approved):
+1. LGTM.

sv2000:
@lesun check the travis failure.

PR #3131, opened 2020-10-18
GOBBLIN-1292: Undo hardcoded tmp directory locations for unit tests

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1292

### Description
Several unit tests in Gobblin make use of hardcoded tmp directory locations. If these locations are not cleaned up, they have the potential to cause test failures in subsequent builds. Further, hardcoded locations may potentially collide with other tests which may accidentally use the same location. This task modifies a host of tests to use dynamically generated temp directories which are automatically cleaned up when tests complete.

### Tests
This PR modifies existing unit tests.

### Commits

### DIFF
--- a/gobblin-cluster/src/test/java/org/apache/gobblin/cluster/FsJobConfigurationManagerTest.java
+++ b/gobblin-cluster/src/test/java/org/apache/gobblin/cluster/FsJobConfigurationManagerTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.gobblin.cluster;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -58,9 +59,8 @@ import org.apache.gobblin.util.filters.HiddenFilter;
 public class FsJobConfigurationManagerTest {
   private MutableJobCatalog _jobCatalog;
   private FsJobConfigurationManager jobConfigurationManager;
-
-  private String jobConfDir = "/tmp/" + this.getClass().getSimpleName() + "/jobCatalog";
-  private String fsSpecConsumerPathString = "/tmp/fsJobConfigManagerTest";
+  private String jobConfDir;
+  private String fsSpecConsumerPathString;
   private String jobSpecUriString = "testJobSpec";
 
   private FileSystem fs;
@@ -91,6 +91,10 @@ public class FsJobConfigurationManagerTest {
       return null;
     }).when(this.eventBus).post(Mockito.anyObject());
 
+    File tmpDir = Files.createTempDir();
+    tmpDir.deleteOnExit();
+    this.jobConfDir = tmpDir.getAbsolutePath() + "/" + this.getClass().getSimpleName() + "/jobCatalog";
+    this.fsSpecConsumerPathString = tmpDir.getAbsolutePath() + "/fsJobConfigManagerTest";
     this.fs = FileSystem.getLocal(new Configuration(false));
     Path jobConfDirPath = new Path(jobConfDir);
     if (!this.fs.exists(jobConfDirPath)) {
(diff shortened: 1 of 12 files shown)


sv2000:
@abti @htran1 Please review.

shirshanka (commented):
@sv2000 : This works. 
A suggestion: Did you consider factoring out the temp directory generator into another class (perhaps in the module: gobblin-test-utils) and have every test use it directly? 

Usage might look like: 

TestDirectoryManager dirMgr = new TestDirectoryManager(this.getClass()); // this sets up the root path, attaches to the calling class. 
this.taskStateFile = dirMgr.getDirectoryPath("/taskState/_RUNNING");

sv2000:
> @sv2000 : This works.
> A suggestion: Did you consider factoring out the temp directory generator into another class (perhaps in the module: gobblin-test-utils) and have every test use it directly?
> 
> Usage might look like:
> 
> TestDirectoryManager dirMgr = new TestDirectoryManager(this.getClass()); // this sets up the root path, attaches to the calling class.
> this.taskStateFile = dirMgr.getDirectoryPath("/taskState/_RUNNING");

Thanks @shirshanka ! Yes, I did think of re-factoring the tmp dir creation. However, I did not see much of a benefit (in terms of avoiding code duplication) since the equivalent of the "TestDirectoryManager" needs to be duplicated whenever a tmp dir is created. Further, the tmp dir creation is really only 2 lines of code - one to create the directory itself and one for marking the dir for deletion.

shirshanka:
@sv2000 : I think the standardization of how temporary directories are created and how directories from within them are handed out might be useful refactoring for future test writers. They won't need to remember how to create a temp directory, how to combine a temp directory with inner directories etc. 
I know it seems like a nit, but I think it sets up the "testing harness" in a better way IMO.

sv2000:
@shirshanka Thanks for the comment. Will address.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/FsJobConfigurationManagerTest.java
```diff
@@ -91,6 +91,10 @@ public void setUp() throws IOException {
       return null;
     }).when(this.eventBus).post(Mockito.anyObject());
 
+    File tmpDir = Files.createTempDir();
+    tmpDir.deleteOnExit();
+    this.jobConfDir = tmpDir.getAbsolutePath() + "/" + this.getClass().getSimpleName() + "/jobCatalog";
```
aplex, line 96: Is it important to have a class name in this path? AFAIK, createTempDir will return a unique path on each call. (same in other test classes below)

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/suite/IntegrationJobCancelSuite.java
```diff
@@ -17,14 +17,22 @@
...
+import com.google.common.io.Files;
 import com.typesafe.config.Config;
 
 
 public class IntegrationJobCancelSuite extends IntegrationBasicSuite {
   public static final String JOB_ID = "job_HelloWorldTestJob_1234";
-  public static final String TASK_STATE_FILE = "/tmp/IntegrationJobCancelSuite/taskState/_RUNNING";
+  public static final String TASK_STATE_FILE;
+  static {
+    File tmpDir = Files.createTempDir();
```
aplex, line 32: Does this need to happen in static block? Can it happen for the instance of a class? Accessing file system from static block looks strange to me. Also TASK_STATE_FILE seems to be like a constant, but it changes on every application run.

aplex (commented):
I like the idea of creating unique temp folders that will get auto-deleted. Left some comments inline.

PR #3135, opened 2020-10-21
[GOBBLIN-1297] Register/unregister eventbus with dagManager on leadership change

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1297

### Description
To be able to kill a flow, the eventbus in GobblinServiceManager must be registered with the dagManager, but this currently only happens on startup. On leadership change, the eventbus of the new leader should be registered and the eventbus of the slave should be unregistered.

### Tests
Trivial

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
@@ -123,7 +123,7 @@ public class GobblinServiceManager implements ApplicationLauncher, StandardMetri
 
   // An EventBus used for communications between services running in the ApplicationMaster
   @Getter
-  protected final EventBus eventBus = new EventBus(GobblinServiceManager.class.getSimpleName());
+  protected EventBus eventBus = new EventBus(GobblinServiceManager.class.getSimpleName());
 
   protected final FileSystem fs;
   protected final Path serviceWorkDir;
@@ -136,7 +136,7 @@ public class GobblinServiceManager implements ApplicationLauncher, StandardMetri
   protected final boolean isRestLIServerEnabled;
   protected final boolean isTopologySpecFactoryEnabled;
   protected final boolean isGitConfigMonitorEnabled;
-  protected final boolean isDagManagerEnabled;
+  protected boolean isDagManagerEnabled;
   protected final boolean isJobStatusMonitorEnabled;
 
   protected TopologyCatalog topologyCatalog;
@@ -161,6 +161,7 @@ public class GobblinServiceManager implements ApplicationLauncher, StandardMetri
 
   protected GitConfigMonitor gitConfigMonitor;
 
+  @Getter
   protected DagManager dagManager;
 
   protected KafkaJobStatusMonitor jobStatusMonitor;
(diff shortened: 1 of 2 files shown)


jack-moseley:
@sv2000 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -417,6 +417,7 @@ private void  handleLeadershipChange(NotificationContext changeContext) {
         //Activate DagManager only if TopologyCatalog is initialized. If not; skip activation.
         if (this.topologyCatalog.getInitComplete().getCount() == 0) {
           this.dagManager.setActive(true);
+          this.eventBus.register(this.dagManager);
```
sv2000, line 420: Any chance we can update GobblinServiceHATest to verify that on leadership change the dagManager has been registered with the eventbus? e.g. construct a mock DagManager and pass it to GobblinServiceManager for it to register the mock dag manager?

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceHATest.java
```diff
@@ -546,6 +551,30 @@ public void testKillNode() throws Exception {
 
     Assert.assertTrue(assertSuccess, "New master should take over all old master jobs.");
 
+    // Check eventbus was registered with new leader
+    attempt = 0;
+    assertSuccess = false;
+    while (attempt < 10) {
```
sv2000, line 557: AssertWithBackoff instead of while(..) {..sleep(..)}?

sv2000 (changes requested):

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceHATest.java
```diff
@@ -546,6 +551,30 @@ public void testKillNode() throws Exception {
...
+      try {
+        Mockito.verify(secondary.getEventBus(), Mockito.atLeastOnce()).register(secondary.getDagManager());
+        assertSuccess = true;
+      } catch (MockitoAssertionError e) {
+        logger.info("Verification failed", e);
+      }
+      Thread.sleep(5);
```
aplex, line 564: The test makes 10 attempts with 5ms intervals. Is 50ms enough for the switch to happen on a busy test machine with high CPU usage? Looks like this can be a source of flakiness for the test.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceHATest.java
jack-moseley: How would you use AssertWithBackoff for a method that is throwing an exception instead of returning true/false? I guess only way is to wrap it in another method.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/core/GobblinServiceHATest.java
jack-moseley: Didn't notice it was ms, but changed to assertwithbackoff with 20 second timeout

sv2000 (approved):
+1. LGTM.

sv2000:
@jack-moseley There is a travis failure. Can you take a look?

jack-moseley:
Looks unrelated to my change, I will try rebuilding since the gradle changes have been reverted.

PR #3139, opened 2020-10-23
[GOBBLIN-1301] Add Azkaban OAuth token support

aplex:
Azkaban introduced a new job type with automatic authentication,
based on OAuth. This type of authentication is used in Azure and
with other cloud providers.

Previously we used Azkaban job type to determine the presence of
existing Hadoop token. This had a problem with new job types
that were not in the predefined list.

In the new implementation, we check the presence of the token
directly in the environment variable.

The author of the previous implementation (ibuenos) could not recall
the reason to use job types, and had no objections for this change.

https://issues.apache.org/jira/browse/GOBBLIN-1301

### DIFF
--- a/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobLauncher.java
+++ b/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobLauncher.java
@@ -113,10 +113,11 @@ public class AzkabanJobLauncher extends AbstractJob implements ApplicationLaunch
   private static final String DEFAULT_AZKABAN_GOBBLIN_JOB_SLA_IN_SECONDS = "-1"; // No SLA.
 
   private static final String HADOOP_JAVA_JOB = "hadoopJava";
+  private static final String HADOOP_JAVA_OAUTH_JOB = "hadoopJavaOAuth";
   private static final String JAVA_JOB = "java";
   private static final String GOBBLIN_JOB = "gobblin";
   private static final Set<String> JOB_TYPES_WITH_AUTOMATIC_TOKEN =
-      Sets.newHashSet(HADOOP_JAVA_JOB, JAVA_JOB, GOBBLIN_JOB);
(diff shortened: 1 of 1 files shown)


aplex:
@sv2000 , can you take a look?

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobLauncher.java
```diff
@@ -175,18 +169,14 @@ public AzkabanJobLauncher(String jobId, Properties props)
...
 
-    if (props.containsKey(JOB_TYPE) && JOB_TYPES_WITH_AUTOMATIC_TOKEN.contains(props.getProperty(JOB_TYPE))) {
-      // Necessary for compatibility with Azkaban's hadoopJava job type
-      // http://azkaban.github.io/azkaban/docs/2.5/#hadoopjava-type
-      LOG.info(
-          "Job type " + props.getProperty(JOB_TYPE) + " provides Hadoop tokens automatically. Using provided tokens.");
-      if (System.getenv(HADOOP_TOKEN_FILE_LOCATION) != null) {
-        this.props.setProperty(MAPREDUCE_JOB_CREDENTIALS_BINARY, System.getenv(HADOOP_TOKEN_FILE_LOCATION));
-      }
+    if (System.getenv(HADOOP_TOKEN_FILE_LOCATION) != null) {
```
sv2000, line 172: So is the idea that if the token file location env variable set, then we assume that Azkaban automatically provides tokens for the job? That should work.

li-ygerchikov: Yes, this should work. I just wander if there may be situations when the env var is set, but the file is either missing or obsolete or does not contain the right tokens... How much of this could (and is worth) checking? E.g. checking that the file exists should be fairly trivial.

aplex: If the token is expired/incorrect/missing due to a bug or system problem, the job will fail when it tries to use it. I don't think we need to switch to internal authentication workflow in this case, as it can mask the original problem.

sv2000 (approved):
+1.

PR #3141, opened 2020-10-28
[GOBBLIN-1303] Catch error when failing to clean staging data

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1303

### Description
Currently if there is a failure in cleanupStagingData, it is not caught, so the rest of launchJob (which includes sending GobblinTrackingEvents) will not be run

### Tests
trivial

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -559,43 +559,43 @@ public abstract class AbstractJobLauncher implements JobLauncher {
               }
             });
           }
-        }
-      }
 
-      for (JobState.DatasetState datasetState : this.jobContext.getDatasetStatesByUrns().values()) {
-        // Set the overall job state to FAILED if the job failed to process any dataset
-        if (datasetState.getState() == JobState.RunningState.FAILED) {
-          jobState.setState(JobState.RunningState.FAILED);
-          LOG.warn("At least one dataset state is FAILED. Setting job state to FAILED.");
-          break;
-        }
-      }
-
-      notifyListeners(this.jobContext, jobListener, TimingEvent.LauncherTimings.JOB_COMPLETE, new JobListenerAction() {
-        @Override
-        public void apply(JobListener jobListener, JobContext jobContext)
-            throws Exception {
-          jobListener.onJobCompletion(jobContext);
-        }
-      });
-
-      if (jobState.getState() == JobState.RunningState.FAILED) {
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -535,7 +535,11 @@ public WorkUnit apply(@Nullable WorkUnit input) {
       } finally {
         try {
           TimingEvent jobCleanupTimer = this.eventSubmitter.getTimingEvent(TimingEvent.LauncherTimings.JOB_CLEANUP);
-          cleanupStagingData(jobState);
+          try {
+            cleanupStagingData(jobState);
+          } catch (Throwable t) {
+            LOG.error("Failed to clean up staging data", t);
```
arjun4084346, line 541: But this change will swallow the exception and any errors in `cleanupStagingData` will not fail the job and hence will go undetected (even if there is log message). 
Should we rather move the code we want to run (sending GTEs) to the finally block at line 600?

jack-moseley: Sure that's a good point, will change that

jack-moseley: Actually not sure if that's the correct place for it either, because if there is an error when notifying listeners, then the cleanup code to close the metrics reporting wouldn't run. Maybe it should be in the finally block at line 542

jack-moseley: Actually isn't this kind of the normal behaviour for a failed job? Any exception in work unit creation is caught at line 530 and not rethrown, does the job not failed for those cases either?

arjun4084346: Got it. Yes 542 is the right place.

arjun4084346: Okay, so the current behavior is do not fail the job (just mark the gobblin job as FAILED) when workunit creation fails, or anything in the main try block fails. However, according to the current behavior any exception in `cleanupStagingData` and `notifyListeners` are thrown (i.e. making job fail).
So, if we want to preserve the same behavior, we should move the code form L565-L599 to finally block at 542, right?
Why the current behavior treats `cleanupStagingData` failures differently, and whether we should change this behavior is a different question that I am not fully sure about.

arjun4084346:
+1 LGTM

jack-moseley:
@sv2000 please merge

sv2000 (approved):
+1. LGTM.

PR #3142, opened 2020-10-29
[GOBBLIN-1304] Adds group ownership service

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1304

### Description

Utilizes the request option `owning_group` in GaaS so that users within the same group can modify and delete each others' flows. 

### Tests
`FlowConfigV2Test` updates
Tested locally with `PrincipalService.java`

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowconfigsV2.snapshot.json
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowconfigsV2.snapshot.json
@@ -68,6 +68,11 @@
       "doc" : "Return the compiled flow as a string. If enabled, the flow is not added.",
       "default" : false
     }, {
+      "name" : "owningGroup",
+      "type" : "string",
+      "doc" : "Optional string name of group that the requester belongs to for group ownership of flows.",
+      "optional" : true
+    }, {
       "name" : "properties",
       "type" : {
         "type" : "map",
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
@@ -18,11 +18,13 @@
 package org.apache.gobblin.service;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.mortbay.jetty.HttpStatus;
 import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
@@ -62,12 +64,18 @@ public class FlowConfigV2Test {
   private EmbeddedRestliServer _server;
   private File _testDirectory;
   private TestRequesterService _requesterService;
+  private GroupOwnershipService groupOwnershipService;
+  private File groupConfigFile;
 
   private static final String TEST_SPEC_STORE_DIR = "/tmp/flowConfigV2Test/";
   private static final String TEST_GROUP_NAME = "testGroup1";
   private static final String TEST_FLOW_NAME = "testFlow1";
   private static final String TEST_FLOW_NAME_2 = "testFlow2";
   private static final String TEST_FLOW_NAME_3 = "testFlow3";
+  private static final String TEST_FLOW_NAME_4 = "testFlow4";
+  private static final String TEST_FLOW_NAME_5 = "testFlow5";
+  private static final String TEST_FLOW_NAME_6 = "testFlow6";
+  private static final String TEST_FLOW_NAME_7 = "testFlow7";
   private static final String TEST_SCHEDULE = "0 1/0 * ? * *";
   private static final String TEST_TEMPLATE_URI = "FS:///templates/test.template";
 
@@ -90,6 +98,15 @@ public class FlowConfigV2Test {
 
     _requesterService = new TestRequesterService(ConfigFactory.empty());
 
+    this.groupConfigFile = new File(_testDirectory + "/TestGroups.json");
+    String groups ="{\"testGroup\": \"testName,testName2\"}";
+    Files.write(groups.getBytes(), this.groupConfigFile);
+    Config groupServiceConfig = ConfigBuilder.create()
+        .addPrimitive(LocalGroupOwnershipService.GROUP_MEMBER_LIST, this.groupConfigFile.getAbsolutePath())
+        .build();
+
+    groupOwnershipService = new LocalGroupOwnershipService(groupServiceConfig);
+
     Injector injector = Guice.createInjector(new Module() {
       @Override
       public void configure(Binder binder) {
@@ -98,6 +115,7 @@ public class FlowConfigV2Test {
         // been made
         binder.bindConstant().annotatedWith(Names.named(FlowConfigsV2Resource.INJECT_READY_TO_USE)).to(Boolean.TRUE);
         binder.bind(RequesterService.class).annotatedWith(Names.named(FlowConfigsV2Resource.INJECT_REQUESTER_SERVICE)).toInstance(_requesterService);
+        binder.bind(GroupOwnershipService.class).annotatedWith(Names.named(FlowConfigsV2Resource.INJECT_GROUP_OWNERSHIP_SERVICE)).toInstance(groupOwnershipService);
       }
     });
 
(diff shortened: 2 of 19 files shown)


gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
```diff
@@ -0,0 +1,72 @@
...
+  public boolean isMemberOfGroup(List<ServiceRequester> serviceRequesters, String group) {
+    if (this.groupOwnerships.has(group)) {
+      String member = extractRequesterUser(serviceRequesters);
+      List<String> groupMembers = Arrays.asList(groupOwnerships.get(group).getAsString().split(","));
+      return groupMembers.contains(member);
+    }
+    return false;
+  }
+
+  private static String extractRequesterUser(List<ServiceRequester> requesterList) {
```
sv2000, line 65: Probably this method can go into the abstract class and presumably, will be needed for other implementations too.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
```diff
@@ -0,0 +1,72 @@
...
+    return requesterList.stream()
+        .filter(serviceRequester -> serviceRequester.getType().equals(USER_TYPE))
```
sv2000, line 67: We shouldn't assume every request is made by a USER principal. There could be use cases where a SERVICE principal can make a request too.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
```diff
@@ -0,0 +1,72 @@
...
+        .collect(Collectors.toList())
+        .get(0)
```
sv2000, line 69: Potential NPE in case of requests made by non USER principals.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
```diff
@@ -0,0 +1,72 @@
...
+  public LocalGroupOwnershipService(Config config) {
+    Path groupOwnershipFilePath = new Path(config.getString(GROUP_MEMBER_LIST));
+    FSDataInputStream in = null;
+    try {
+      FileSystem fs = FileSystem.get(new Configuration());
+      in = fs.open(groupOwnershipFilePath);
+      String jsonString = IOUtils.toString(in, Charset.defaultCharset());
+      JsonParser parser = new JsonParser();
+      this.groupOwnerships = parser.parse(jsonString).getAsJsonObject();
+      in.close();
```
sv2000, line 49: try with resources to avoid having to call in.close()?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
```diff
@@ -0,0 +1,72 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public boolean isMemberOfGroup(List<ServiceRequester> serviceRequesters, String group) {
+    if (this.groupOwnerships.has(group)) {
+      String member = extractRequesterUser(serviceRequesters);
+      List<String> groupMembers = Arrays.asList(groupOwnerships.get(group).getAsString().split(","));
```
sv2000, line 59: You can use Guava's Splitter:
Splitter.on(',').trimResults().omitEmptyStrings().splitToList(groupOwnerships.get(group).getAsString())

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -200,4 +210,29 @@ private Properties getHeaders() {
     }
     return headerProperties;
   }
+
+  public void checkRequester(
```
sv2000, line 214: Add javadoc explaining the behavior of this method.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -200,4 +210,29 @@ private Properties getHeaders() {
...
+      return;
+    }
+
+    try {
+      String serializedOriginalRequesterList = originalFlowConfig.getProperties().get(RequesterService.REQUESTER_LIST);
+      if (serializedOriginalRequesterList != null) {
+        List<ServiceRequester> originalRequesterList = RequesterService.deserialize(serializedOriginalRequesterList);
+        String owningGroup = originalFlowConfig.getProperties().getOrDefault("owning_group", "");
+        // check that the requester is part of the owning group
+        if (!owningGroup.isEmpty()) {
```
sv2000, line 226: So is the expected behavior that if owningGroup is specified, only members of that group are allowed to perform the request? What if we have a group with a single member that created a flow and that member is no longer in the organization?

sv2000: We should accommodate an "admin" mode where a user with admin privileges can also make a request.

sv2000 (changes requested):
Thanks! The overall structure looks good to me. I think the checkRequester() method needs a little more thinking in terms of having overrides for admin users.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
Will-Lo: I like the idea of an admin mode! Would the admin mode just be another flow configuration parameter? How would we authenticate that?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -128,11 +132,17 @@ public FlowConfig get(ComplexResourceKey<FlowId, FlowStatusId> key) {
...
   public CreateKVResponse create(FlowConfig flowConfig) {
-    List<ServiceRequester> requestorList = this.requesterService.findRequesters(this);
+    List<ServiceRequester> requesterList = this.requesterService.findRequesters(this);
     try {
-      String serialized = RequesterService.serialize(requestorList);
+      String serialized = RequesterService.serialize(requesterList);
       flowConfig.getProperties().put(RequesterService.REQUESTER_LIST, serialized);
       LOG.info("Rest requester list is " + serialized);
+      String owningGroup = flowConfig.getProperties().getOrDefault("owning_group", "");
+      if (owningGroup != "") {
```
arjun4084346, line 141: Strings cannot be compared like this in java. This will always return false. Please use String's equals method. or guava's isNullOrEmpty()
I haven't thoroughly reviewed the unit test, but if unit test passes with this, maybe they need some correction too.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -200,4 +210,29 @@ private Properties getHeaders() {
...
+      RequesterService requesterService, FlowConfig originalFlowConfig, List<ServiceRequester> requesterList) {
+    if (requesterList == null) {
+      return;
+    }
+
+    try {
+      String serializedOriginalRequesterList = originalFlowConfig.getProperties().get(RequesterService.REQUESTER_LIST);
+      if (serializedOriginalRequesterList != null) {
+        List<ServiceRequester> originalRequesterList = RequesterService.deserialize(serializedOriginalRequesterList);
+        String owningGroup = originalFlowConfig.getProperties().getOrDefault("owning_group", "");
```
arjun4084346, line 224: Let's create a constant for "owning_group", it is being used at multiple places.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/GroupOwnershipService.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.apache.gobblin.service;
+
+import java.util.List;
+
+
+public abstract class GroupOwnershipService {
+   public static String USER_TYPE = "USER_PRINCIPAL";
+
+   abstract boolean isMemberOfGroup(List<ServiceRequester> serviceRequesters, String group);
```
arjun4084346, line 26: This should be either protected or public. Otherwise we will not be able to write implementations in li-gobblin (outside this package)

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipService.java
arjun4084346: +1 and not just to avoid calling in.close() but also to make sure it is called in case of exception.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/GroupOwnershipService.java
Will-Lo: By default it should be protected but I'll explicitly define it as such too

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowconfigsV2.snapshot.json
```diff
@@ -67,6 +67,11 @@
       "type" : "boolean",
       "doc" : "Return the compiled flow as a string. If enabled, the flow is not added.",
       "default" : false
+    }, {
+      "name" : "owningGroup",
```
arjun4084346, line 71: We may need to handle this in FlowSpecDeserializer/FlowSpecSerializer

arjun4084346: and also update mysql table schema with this field, and handle this field in MysqlSpecStore ?
Feel free to do this in other PR.

Will-Lo: I think the FlowSpecDeserializer/Serializer should be fine, since it reads the fields from `getConfigAsProperties` which I modified on `FlowSpec`. I'll take a look into seeing what needs to be changed in MySqlSpecStore though, good catch.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/NoopGroupOwnershipService.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gobblin.service;
+
+import com.typesafe.config.Config;
+import java.util.List;
+
+
+public class NoopGroupOwnershipService extends GroupOwnershipService{
```
sv2000, line 23: Can we use @Alias notations for different Group ownership service implementations?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
```diff
@@ -83,8 +83,8 @@
           + "spec LONGBLOB, " + NEW_COLUMN + " JSON, PRIMARY KEY (spec_uri))";
   private static final String EXISTS_STATEMENT = "SELECT EXISTS(SELECT * FROM %s WHERE spec_uri = ?)";
   protected static final String INSERT_STATEMENT = "INSERT INTO %s (spec_uri, flow_group, flow_name, template_uri, "
-      + "user_to_proxy, source_identifier, destination_identifier, schedule, tag, isRunImmediately, spec, " + NEW_COLUMN + ") "
-      + "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE spec = VALUES(spec), " + NEW_COLUMN + " = VALUES(" + NEW_COLUMN + ")";
+      + "user_to_proxy, source_identifier, destination_identifier, schedule, tag, isRunImmediately, owning_group, spec, " + NEW_COLUMN + ") "
```
sv2000, line 86: owning_group -> owningGroup?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -304,6 +306,19 @@ public GobblinServiceManager(String serviceName, String serviceId, Config config
...
         ServiceConfigKeys.GOBBLIN_SERVICE_RESTLI_SERVER_ENABLED_KEY, true);
 
+    ClassAliasResolver<GroupOwnershipService> groupOwnershipAliasResolver = new ClassAliasResolver<>(GroupOwnershipService.class);
+    String groupOwnershipServiceClass = ServiceConfigKeys.DEFAULT_GROUP_OWNERSHIP_SERVICE;
+    if (config.hasPath(ServiceConfigKeys.GROUP_OWNERSHIP_SERVICE_CLASS)) {
+      groupOwnershipServiceClass = config.getString(ServiceConfigKeys.GROUP_OWNERSHIP_SERVICE_CLASS);
+    }
+    GroupOwnershipService groupOwnershipService;
+    try {
+      groupOwnershipService = (GroupOwnershipService) ConstructorUtils.invokeConstructor(
```
sv2000, line 316: Use aliases for different group ownership service implementations and use ClassAliasResolver here.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/LocalGroupOwnershipPathAlterationListener.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.gobblin.util.filesystem.PathAlterationListenerAdaptor;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class LocalGroupOwnershipPathAlterationListener extends PathAlterationListenerAdaptor {
```
sv2000, line 35: +1. Well done.

sv2000 (changes requested):
Minor comments. The PR is looking good overall.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -185,20 +203,102 @@ public void testBadPartialUpdate() throws Exception {
...
+
+     _client.createFlowConfig(flowConfig);
 
     testRequester.setName("testName2");
-    _client.deleteFlowConfig(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME));
+    _client.deleteFlowConfig(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_5));
+  }
+
+  @Test
+  public void testGroupRequesterRejected() throws Exception {
```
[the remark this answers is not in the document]
Will-Lo: testDisallowedRequester doesn't test using owning groups, it was originally used to test the requester service. This tests for a group member that isn't part of a group that exists. They are pretty similar though.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/GroupOwnershipService.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+
+/**
+ * Service for handling group ownership of flows
+ */
+public abstract class GroupOwnershipService {
+
+   public static String GROUP_OWNERSHIP_PARAMETER = "owning_group";
```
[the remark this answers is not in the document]
Will-Lo: Whoops this variable isn't being used after owningGroup got moved to be top level, I'll just delete it.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_store/MysqlSpecStore.java
Will-Lo: Would we introduce backwards incompatible changes with that? The field was already initialized in sql tables as `owning_group` a few months back, so I stuck with that. Otherwise we would need to possibly rename the column names of existing flowspec stores stored in mysql.

sv2000 (approved):
+1. LGTM.

PR #3143, opened 2020-10-29
[GOBBLIN-1305] Fix ORC RecordReader processing splits with offset

autumnust:
Dear Gobblin maintainers,

### JIRA
- https://issues.apache.org/jira/browse/GOBBLIN-1305

### Description
The key change is inside `OrcValueCombineFileRecordReader.java` : It used to use `rows()`to get `RecordReader` from fileReader while it shouldn't be the case. The option it consumed will be able to set the range of records to be processed properly. 

Other changes are mostly coordinating with the unit tests.

### Tests

### Commits


gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/orc/OrcUtils.java
```diff
@@ -75,10 +75,10 @@ private OrcUtils() {
 
   public static TypeDescription getTypeDescriptionFromFile(Configuration conf, Path orcFilePath)
       throws IOException {
-    return getRecordReaderFromFile(conf, orcFilePath).getSchema();
+    return getFileReader(conf, orcFilePath).getSchema();
   }
 
-  public static Reader getRecordReaderFromFile(Configuration conf, Path orcFilePath)
+  public static Reader getFileReader(Configuration conf, Path orcFilePath)
```
ZihanLi58, line 81: Since this is a public static method, are we sure there is no other dependencies call this API?

ZihanLi58 (changes requested):
LGTM overall, just one comment

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/orc/OrcUtils.java
autumnust: Will put an umbrella method for backward compatibility.

sv2000 (approved):
+1. LGTM.

PR #3150, opened 2020-11-07
[GOBBLIN-1312] Bumping parquet lib to 1.11.1 to remove hadoop-lzo dependency

treff7es:
Bumping parquet lib to 1.11.1 to remove hadoop-lzo dependency which caused build error as twitter's maven repo is unreliable.

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1312

### Description
Bumping parquet lib to 1.11.1 to remove hadoop-lzo dependency which caused build error as twitter's maven repo is unreliable. -> https://github.com/apache/parquet-mr/commit/d1190abee3ff1f1183757f7e400d7b9b4025d95b?fbclid=IwAR3o3YfcKnQ06x7ZKsaKobCILlhaRn4Pa5aKIr9B6BVAgXR0JaW8D47M0Z8#diff-e1b56c6e57b54549dfbd8e36c064998921a7f1323773ab5f6cda891e580ffed1

### Tests

### Commits


autumnust (approved):
Thanks Tames for fixing this !

Will-Lo:
Thanks for doing this! I think there are some failing tests, when I checked out your repo there are some parquet tests that need to be modified, could you look into `JsonIntermediateToParquetGroupConverterTest` ? 
The error seems to be due to a decoding being interpreted as STRING instead of UTF8.

```
java.lang.AssertionError: expected  [messagetest_table{;optionalbinarysome_enum(UTF8);;};] but found [messagetest_table{;optionalbinarysome_enum(STRING);;};]
```

treff7es:
@Will-Lo thanks that one is fixed. The more problematic part is the parquetOutputFormatTest integration test which fails because the Parquet hive output format is deep in Hive.
The only way to being able to use the newer version of Parquet is to bump the Hive version as well. -> https://issues.apache.org/jira/browse/HIVE-11096

Will-Lo:
Upgrading Hive is quite risky for us as we've tried before and it failed in the past. Is it possible to add the upgraded version of hive as a shadow jar in the test module?

Will-Lo:
@treff7es I can't seem to replicate the problem you're facing, instead `parquetOutputFormatTest` is hanging for me when I run it with gradle, but not intellij. Could you give me more details so that I can try shadowing a hive jar?

Will-Lo:
After doing some investigation we think the best solution would be to disable the test and document in `HiveWritableHdfsDataWriter` that it currently does not support parquet. I made a ticket detailing the problem here: https://issues.apache.org/jira/browse/GOBBLIN-1318

shirshanka:
Would this break any production pipelines that are today using hivewriter (versus native writer) to write to parquet format data after they upgrade their version of Gobblin?

Will-Lo:
Looks like `JsonIntermediateToParquetGroupConverterTest` is failing due to the binary UTF-8 expected instead of string as well.

@shirshanka from what I can see we don't use hivewriter with parquet. @autumnust can you confirm?

autumnust:
> Looks like `JsonIntermediateToParquetGroupConverterTest` is still failing due to the binary UTF-8 expected instead of string as well.
> 
> @shirshanka from what I can see we don't use hivewriter with parquet. @autumnust can you confirm?

Concur that for no internal use case

autumnust (approved):

PR #3155, opened 2020-11-16
[GOBBLIN-1319] fix cancellation in gobblin cluster jobs

arjun4084346:
Dear Gobblin maintainers,

@yukuai518 @sv2000  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1319

### Description
- jobsMapping in GobblinHelixJobScheduler was using job name to store the mapping of job to planning/actual job. this pr is changing it to use job uri instead, because cancellation request just have a uri
- provide a clean separate path for cancellation on cluster side, so that 'delete' calls from within the cluster and from gobblin service can be treated differently
- storing non distributed mode jobs also in JobsMapping in order to provide a cleaner control over cancellation. Storing helix jobs' metadata in a mysql table will also let users cancel the flow if cluster-manager restarts after initial flow submission.

### Tests
verified with integration test

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.cluster;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -50,6 +51,7 @@ import com.google.common.util.concurrent.Striped;
 import com.typesafe.config.Config;
 
 import org.apache.gobblin.annotation.Alpha;
+import org.apache.gobblin.cluster.event.CancelJobConfigArrivalEvent;
 import org.apache.gobblin.cluster.event.DeleteJobConfigArrivalEvent;
 import org.apache.gobblin.cluster.event.NewJobConfigArrivalEvent;
 import org.apache.gobblin.cluster.event.UpdateJobConfigArrivalEvent;
@@ -84,7 +86,7 @@ import org.apache.gobblin.util.PropertiesUtils;
  * <p> More details can be found at {@link HelixRetriggeringJobCallable}.
  */
 @Alpha
-public class GobblinHelixJobScheduler extends JobScheduler implements StandardMetricsBridge{
+public class GobblinHelixJobScheduler extends JobScheduler implements StandardMetricsBridge {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(GobblinHelixJobScheduler.class);
   private static final String COMMON_JOB_PROPS = "gobblin.common.job.props";
@@ -205,9 +207,9 @@ public class GobblinHelixJobScheduler extends JobScheduler implements StandardMe
 
     if (cleanAllDistJobs) {
       for (org.apache.gobblin.configuration.State state : this.jobsMapping.getAllStates()) {
-        String jobName = state.getId();
-        LOGGER.info("Delete mapping for job " + jobName);
-        this.jobsMapping.deleteMapping(jobName);
+        String jobUri = state.getId();
+        LOGGER.info("Delete mapping for job " + jobUri);
+        this.jobsMapping.deleteMapping(jobUri);
       }
     }
   }
@@ -346,7 +348,7 @@ public class GobblinHelixJobScheduler extends JobScheduler implements StandardMe
   }
 
   private void waitForJobCompletion(String jobName) {
-    while (this.jobRunningMap.getOrDefault(jobName, false) != false) {
+    while (this.jobRunningMap.getOrDefault(jobName, false)) {
       LOGGER.info("Waiting for job {} to stop...", jobName);
       try {
         Thread.sleep(1000);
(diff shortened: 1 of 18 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -367,6 +369,33 @@ public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobAr
...
+
+    if (planningJob.isPresent()) {
+      LOGGER.info("Cancelling planning job helix workflow: {}", planningJob.get());
+      new TaskDriver(this.taskDriverHelixManager.get()).waitToStop(planningJob.get(), this.helixJobStopTimeoutMillis);
+    }
+
+    if (actualJob.isPresent()) {
+      LOGGER.info("Cancelling actual job helix workflow: {}", actualJob.get());
+      new TaskDriver(this.jobHelixManager).waitToStop(actualJob.get(), this.helixJobStopTimeoutMillis);
+    }
```
yukuai518, line 396: I have several questions:
1. What is the difference between this logic vs cancelJobIfRequired() method, can we reuse some part of the logic?
2. I don't know if it is safe to stop the actual job twice. When you cancel the planningJob, the GobblinHelixJobLauncher on the task driver node will receive a cancellation message and invoke its own executeCancellation() method, which will further cancel the actual job on the worker node. If we cancel the actual job directly here, do we have any conflict? Will there be any race condition needs some scrutify?
3. Can you provide some comments on which method will be used by GaaS and which one will be the opposite since you mentioned you want a separate path?

arjun4084346: 1) The problem is in `HelixUtils.getWorkflowIdsFromJobNames` method.
It uses ConfigurationKeys.JOB_NAME_KEY (job.name) to identify helix workflows for the provided jobs. However, for planning jobs' `job.name` is the name of original jobs and not what you get with `planningJob.get()`
The helix jobs' configs do have a field (`PLANNING_ID_KEY`) which stores `planningJob.get()` but I cannot use that field because I wanted to keep `cancelJobIfRequired()` method usable for rest of the helix jobs.

2) There is no side effect, but yes it can be removed from here.

3) HelixRetriggeringJobCallable.deleteJobSpec() calls `jobCatalog.remove()`, after it submits job to helix. I want this remove to be different than the DELETE request coming from GaaS, which is meant for cancellation (including deletion). added this comment in NonObservingFSJobCatalog

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -334,15 +351,44 @@ private void runJobExecutionLauncher() throws JobException {
...
       try {
         closer.close();
       } catch (IOException e) {
-        throw new JobException("Cannot properly close planning job " + jobName, e);
+        throw new JobException("Cannot properly close planning job for " + this.jobUri, e);
+      }
+    }
+  }
+
+  private boolean canRun(Optional<String> jobIdFromStore, HelixManager helixManager) throws JobException, InterruptedException {
```
sv2000, line 365: Can you document the behavior of canRun? Also, can this be unit tested?

sv2000 (changes requested):

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -367,6 +369,42 @@ public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobAr
...
+        actualJob = this.jobsMapping.getActualJobId(jobUri);
+      }
+    } catch (IOException e) {
+      LOGGER.warn("jobsMapping could not be retrieved for job {}", jobUri);
+      return;
+    }
+
+    if (planningJob.isPresent()) {
+      LOGGER.info("Cancelling planning job helix workflow: {}", planningJob.get());
+      new TaskDriver(this.taskDriverHelixManager.get()).waitToStop(planningJob.get(), this.helixJobStopTimeoutMillis);
```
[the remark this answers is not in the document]
arjun4084346: According to the javadoc, if this method does not throw any exception, the cancellation is successful.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -196,22 +202,57 @@ private void deleteJobSpec() throws JobException {
    * the job launcher determines it is safe to stop.
    */
   private void runJobLauncherLoop() throws JobException {
+    String actualJobId;
+    Lock jobLock = locks.get(this.jobUri);
```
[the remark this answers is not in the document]
arjun4084346: It is to avoid race condition in checking if the job with the same name is running. this is a new thing this method is doing.
reference https://github.com/apache/incubator-gobblin/blob/master/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java#L240

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
arjun4084346: I just refactored this code to avoid code duplication

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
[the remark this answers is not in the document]
arjun4084346: And also, there will be plenty of helix logs, cluster-worker and/or cluster-taskdriver logs to confirm the cancellation

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobScheduler.java
```diff
@@ -367,6 +369,42 @@ public void handleDeleteJobConfigArrival(DeleteJobConfigArrivalEvent deleteJobAr
...
+    Optional<String> distributedJobMode;
+    Optional<String> planningJob = Optional.empty();
+    Optional<String> actualJob = Optional.empty();
+
+    this.jobSchedulerMetrics.numCancellationStart.incrementAndGet();
+
+    try {
+      distributedJobMode = this.jobsMapping.getDistributedJobMode(jobUri);
+      if (distributedJobMode.isPresent() && Boolean.parseBoolean(distributedJobMode.get())) {
+        planningJob = this.jobsMapping.getPlanningJobId(jobUri);
```
sv2000, line 386: In distributed mode, do we only have to cancel the planning job but not the actual job? Why is that?

arjun4084346: Yes, because the planning job cancellation method cancels the actual job.
https://github.com/apache/incubator-gobblin/blob/master/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java#L256

sv2000 (approved):
+1. LGTM.

PR #3157, opened 2020-11-25
GOBBLIN-1308 ability to manage token for remote cluster

jhsenjaliya:
Dear Gobblin maintainers,

**Basic idea here is to hold all the tokens in `Credentials` instead of single `Token` variable.**

### JIRA
- [x] Issue: https://issues.apache.org/jira/browse/GOBBLIN-1308

### Description
- [x] description provided on the issue

### Tests

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
@@ -20,17 +20,8 @@ package org.apache.gobblin.util.hadoop;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
-import java.io.DataOutputStream;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
-import java.security.PrivilegedExceptionAction;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.regex.Pattern;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.gobblin.configuration.State;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
@@ -55,10 +46,20 @@ import org.apache.hadoop.security.token.Token;
 import org.apache.hadoop.yarn.factory.providers.RecordFactoryProvider;
 import org.apache.hadoop.yarn.ipc.YarnRPC;
 import org.apache.hadoop.yarn.util.ConverterUtils;
-import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.StringUtils;
-import java.util.ArrayList;
 import org.apache.thrift.TException;
+
+import java.io.DataOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Pattern;
 
 
 /**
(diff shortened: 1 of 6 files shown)


jhsenjaliya:
in the test class, I ideally wanted to mock up the `addDelegationToken` method for the `Filesystem` so that It can provide the mock'ed up tokens. but because Gobblin uses very old version of `mockito`, it can not mock up static methods. I tried `powerMockito` as well but it didnt work. so for now I have mock up the entire method `getNewDelegationTokenForLoginUser`.

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
```diff
@@ -132,6 +134,35 @@ public Void run() throws Exception {
     return ugi;
   }
 
+  public static void getHadoopFSTokens(final State state, Optional<File> tokenFile, final Credentials cred, final String renewer)
```
ZihanLi58, line 137: Should we also get renewer from the state here?

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
```diff
@@ -280,84 +314,122 @@ private static void getJhToken(Configuration conf, Credentials cred) throws IOEx
...
+  public static void getRemoteFSTokenFromURI(Configuration conf, Credentials cred, List<String> otherNamenodes, String renewer)
       throws IOException {
-    LOG.info(OTHER_NAMENODES + ": " + otherNamenodes);
+    log.debug("Getting tokens for other namenodes: " + otherNamenodes);
     Path[] ps = new Path[otherNamenodes.size()];
     for (int i = 0; i < ps.length; i++) {
       ps[i] = new Path(otherNamenodes.get(i).trim());
+      FileSystem otherNameNodeFS = ps[i].getFileSystem(conf);
+
+      if (StringUtils.isEmpty(renewer)) {
```
ZihanLi58, line 404: renewer here seems to be "getMRTokenRenewerInternal(new JobConf()).toString()" defined on line 372, will it be empty? this may end up with we forcing to obtain remote token from remote fs

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
```diff
@@ -280,84 +314,122 @@ private static void getJhToken(Configuration conf, Credentials cred) throws IOEx
...
+      } else {
+        final Token<?>[] tokens = otherNameNodeFS.addDelegationTokens(renewer, cred);
+        if (tokens != null) {
+          for (Token<?> token : tokens) {
+            log.info("Got dt token for " + otherNameNodeFS.getUri() + "; " + token);
+          }
+        }
+      }
     }
-    TokenCache.obtainTokensForNamenodes(cred, ps, conf);
```
ZihanLi58, line 342: I take a look at the implementation of `TokenCache.obtainTokensForNamenodes(cred, ps, conf);` Seems like the origin implementation do support get remote token by get it from the remote fs. Why we add this separately? To support use a non-default renewer?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/AbstractYarnAppSecurityManager.java
```diff
@@ -117,7 +118,11 @@ protected void startUp() throws Exception {
     this.loginExecutor.scheduleAtFixedRate(new Runnable() {
       @Override
       public void run() {
-        loginAndScheduleTokenRenewal();
+        try {
+          loginAndScheduleTokenRenewal();
+        }catch(Exception e){
+          LOGGER.error("Error login using keytab, will continue the thread and try next time.");
```
ZihanLi58, line 124: Change the message to "Error during login, will continue the thread and try next time." Since this is super class and we may not use keytab to login

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/YarnHelixUtils.java
```diff
@@ -60,15 +60,19 @@
...
-  public static void writeTokenToFile(Token<? extends TokenIdentifier> token, Path tokenFilePath,
-      Configuration configuration) throws IOException {
-    Credentials credentials = new Credentials();
-    credentials.addToken(token.getService(), token);
+  public static void writeTokenToFile(Path tokenFilePath, Credentials credentials, Configuration configuration) throws IOException {
+    if(credentials == null) {
+      LOGGER.warn("got empty credentials, creating default one as new.");
+      credentials = new Credentials();
+    }
+    // TODO: token needs to be removed from the function signature as well
```
ZihanLi58, line 73: Remove useless comment

ZihanLi58 (commented):
First round review, main question is why we need another approach of get remote token with specified renewer instead of the origin implementation.

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: You can. But this function should take renewer regardless.  The user of thefunction can implement to take renewed from config as override but we found that its better to use the process owner as  default renewer.

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: This is to maintain the backward. Omaptibility where if renewer is not explicitly defined it needs to take different approach.

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: No. The token cache implementation by default does not take renewer info so there is no way to specify it.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/AbstractYarnAppSecurityManager.java
jhsenjaliya: Sure

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
ZihanLi58: Yeah, previous approach renewer is null, but for this approach, even when renewer is null, it will be assign as "getMRTokenRenewerInternal(new JobConf()).toString()" defined on line 372 and will not fall into the origin implementation. If I misunderstand something here?

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
ZihanLi58: So the change is mainly to support define renewer, am I correct?

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: I see what you are saying, you r right about line 372, i will fix that, but the goal is to use `addDelegationTokens` method which allows us to explicitly specify renewer, whereas `TokenCache.obtainTokensForNamenodes` does not take renewer and will always use `Master.getMasterPrincipal(conf)`

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: yes, along with some refactoring

jhsenjaliya:
> First round review, main question is why we need another approach of get remote token with specified renewer instead of the origin implementation.

Using yarn or MR as renewer might not be allowed or possible, Also that is not the only change. 
The change is also to collect all the tokens under credentials, that can be renewed.

gobblin-utility/src/main/java/org/apache/gobblin/util/hadoop/TokenUtils.java
jhsenjaliya: relooking at this logic, earlier there were no way to specify renewer, now that I have added way to specify renewer, we have to fallback to default implementation ( which is to take renewer = `getMRTokenRenewerInternal(new JobConf()).toString();` ) so it will fall into original implementation if user does not specify renewer. I added an overloaded method `getAllFSTokens` that does not take renewer at all to make it look as earlier.

ZihanLi58: My concern is renewer on line 421 will never be null, since it's assigned as getMRTokenRenewerInternal(new JobConf()).toString(); So we for other name nodes, we never fall back to original implementation.

ZihanLi58:
@jhsenjaliya Can you check the travis build fail?

ZihanLi58 (approved):
LGTM, @sv2000 Can you help merge this change?

PR #3158, opened 2020-11-25
[GOBBLIN-1377] Adds functionality to create shards for target directories in hive distcp

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1377

### Description

This PR allows Gobblin to have an entrypoint to create shards on cloud environments before copying. Notable changes include
1) A configuration `gobblin.useDatasetLocalWorkDir` to enable dataset-local temporary directories for the writer and error files instead of a storing these on the job level.
2) Refactoring of `HiveCopyEntityHelper` for code cleanup
3) Hive Dataset unit tests for `CopySource`

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -210,7 +210,6 @@ public class ConfigurationKeys {
   public static final String MAXIMUM_JAR_COPY_RETRY_TIMES_KEY = JOB_JAR_FILES_KEY + ".uploading.retry.maximum";
   public static final String USER_DEFINED_STATIC_STAGING_DIR = "user.defined.static.staging.dir";
   public static final String USER_DEFINED_STAGING_DIR_FLAG = "user.defined.staging.dir.flag";
-  public static final String IS_DATASET_STAGING_DIR_USED = "dataset.staging.dir.used";
 
   public static final String QUEUED_TASK_TIME_MAX_SIZE = "taskexecutor.queued_task_time.history.max_size";
   public static final int DEFAULT_QUEUED_TASK_TIME_MAX_SIZE = 2048;
@@ -1062,4 +1061,14 @@ public class ConfigurationKeys {
    */
   public static final String TASK_EVENT_METADATA_GENERATOR_CLASS_KEY = "gobblin.task.event.metadata.generator.class";
   public static final String DEFAULT_TASK_EVENT_METADATA_GENERATOR_CLASS_KEY = "nooptask";
+
+  /**
+   * Configuration for sharded directory files
+   */
+  public static final String USE_DATASET_LOCAL_WORK_DIR = "gobblin.useDatasetLocalWorkDir";
+  public static final String DESTINATION_DATASET_HANDLER_CLASS = "gobblin.destination.datasetHandlerClass";
+  public static final String DATASET_DESTINATION_PATH = "gobblin.dataset.destination.path";
+  public static final String STAGING_DIR_DEFAULT_SUFFIX = "/.temp/taskStaging";
+  public static final String OUTPUT_DIR_DEFAULT_SUFFIX = "/.temp/taskOutput";
+  public static final String ROW_LEVEL_ERR_FILE_DEFAULT_SUFFIX = "/err";
 }
--- a/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandler.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandler.java
@@ -15,16 +15,27 @@
  * limitations under the License.
  */
 
-package org.apache.gobblin.data.management.copy;
+package org.apache.gobblin.destination;
 
-import org.apache.gobblin.dataset.Dataset;
-import org.apache.hadoop.fs.Path;
-
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Collection;
+import org.apache.gobblin.source.workunit.WorkUnit;
 
 /**
- * A common superinterface for {@link Dataset}s that can be operated on by distcp.
- * Concrete classes must implement a subinterface of this interface ({@link CopyableDataset} or {@link IterableCopyableDataset}).
+ * Performs work related to initializing the target environment before the files are written and published
  */
-public interface CopyableDatasetBase extends Dataset {
-  default String getDatasetPath() { return ""; }
+public interface DestinationDatasetHandler extends Closeable {
+
+  /**
+   * Handle destination setup before workunits are sent to writer and publisher
+   * @param workUnits
+   */
+  void handle(Collection<WorkUnit> workUnits) throws IOException;
+
+  /**
+   * Perform cleanup if needed
+   * @throws IOException
+   */
+  void close() throws IOException;
 }
--- a/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.destination;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.apache.gobblin.configuration.ConfigurationKeys;
+import org.apache.gobblin.configuration.SourceState;
+import org.apache.gobblin.metrics.event.EventSubmitter;
+import org.apache.gobblin.source.workunit.WorkUnit;
+import org.apache.gobblin.source.workunit.WorkUnitStream;
+import org.apache.gobblin.util.JobLauncherUtils;
+
+
+/**
+ * Initializes and runs handlers on workunits before writers are initialized
+ * Reads {@link ConfigurationKeys#DESTINATION_DATASET_HANDLER_CLASS} as a list
+ * of classes, separated by comma to initialize the handlers
+ */
+public class DestinationDatasetHandlerService implements Closeable {
+  List<DestinationDatasetHandler> handlers;
+
+  public DestinationDatasetHandlerService(SourceState jobState, Boolean canCleanUp, EventSubmitter eventSubmitter) {
+    this.handlers = new ArrayList<>();
+    if (jobState.contains(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)) {
+      String[] handlerList = jobState.getProp(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)
+          .replaceAll("\\s", "") // cleans out whitespace
+          .split(",");
+      for (String handlerClass : handlerList) {
+        this.handlers.add(DestinationDatasetHandlerFactory.newInstance(handlerClass, jobState, canCleanUp));
+      }
+    }
+  }
+
+  /**
+   * Executes handlers
+   * @param workUnitStream
+   */
+  public void executeHandlers(WorkUnitStream workUnitStream) {
+    if (handlers.size() > 0) {
+      if (workUnitStream.isSafeToMaterialize()) {
+        Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+          for (DestinationDatasetHandler handler : this.handlers) {
+            try {
+              handler.handle(workUnits);
+            } catch (IOException e) {
+              throw new RuntimeException(String.format("Handler %s failed to execute", handler.getClass().getName()), e);
+            }
+          }
+      } else {
+        throw new RuntimeException(DestinationDatasetHandlerService.class.getName() + " does not support work unit streams");
+      }
+    }
+  }
+
+
+  public void close() throws IOException {
+    for (DestinationDatasetHandler handler: this.handlers) {
+      handler.close();
+    }
+  }
+}
--- a/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
@@ -41,9 +41,7 @@ public class DestinationDatasetHandlerService implements Closeable {
   public DestinationDatasetHandlerService(SourceState jobState, Boolean canCleanUp, EventSubmitter eventSubmitter) {
     this.handlers = new ArrayList<>();
     if (jobState.contains(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)) {
-      String[] handlerList = jobState.getProp(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)
-          .replaceAll("\\s", "") // cleans out whitespace
-          .split(",");
+      List<String> handlerList = jobState.getPropAsList(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS);
       for (String handlerClass : handlerList) {
         this.handlers.add(DestinationDatasetHandlerFactory.newInstance(handlerClass, jobState, canCleanUp));
       }
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
@@ -53,8 +53,6 @@ import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.data.management.copy.extractor.EmptyExtractor;
 import org.apache.gobblin.data.management.copy.extractor.FileAwareInputStreamExtractor;
-import org.apache.gobblin.data.management.copy.hive.HiveDataset;
-import org.apache.gobblin.data.management.copy.hive.HiveDatasetFinder;
 import org.apache.gobblin.data.management.copy.prioritization.FileSetComparator;
 import org.apache.gobblin.data.management.copy.publisher.CopyEventSubmitterHelper;
 import org.apache.gobblin.data.management.copy.replication.ConfigBasedDataset;
@@ -132,8 +130,6 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
   public static final String FILESET_TOTAL_ENTITIES = "fileset.total.entities";
   public static final String FILESET_TOTAL_SIZE_IN_BYTES = "fileset.total.size";
   public static final String SCHEMA_CHECK_ENABLED = "shcema.check.enabled";
-  public static final String DATASET_STAGING_DIR_PATH = "dataset.staging.dir.path";
-  public static final String DATASET_STAGING_PATH = "dataset.staging.path";
   public final static boolean DEFAULT_SCHEMA_CHECK_ENABLED = false;
 
   private static final String WORK_UNIT_WEIGHT = CopyConfiguration.COPY_PREFIX + ".workUnitWeight";
@@ -370,9 +366,11 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
-          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.IS_DATASET_STAGING_DIR_USED,false))) {
-            workUnit.setProp(DATASET_STAGING_DIR_PATH, ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH));
-          }
+
+          // Ensure that the writer temporary directories are contained within the dataset shard
+          String datasetPath = this.copyableDataset.getDatasetPath();
+          workUnit.setProp(ConfigurationKeys.DATASET_DESTINATION_PATH, datasetPath);
+
           serializeCopyEntity(workUnit, copyEntity);
           serializeCopyableDataset(workUnit, metadata);
           GobblinMetrics.addCustomTagToState(workUnit,
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopyableDatasetBase.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopyableDatasetBase.java
@@ -18,7 +18,6 @@
 package org.apache.gobblin.data.management.copy;
 
 import org.apache.gobblin.dataset.Dataset;
-import org.apache.hadoop.fs.Path;
 
 
 /**
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -142,12 +142,6 @@ public class HiveCopyEntityHelper {
    * partitions with Path containing '/Hourly/' will be kept.
    */
   public static final String HIVE_PARTITION_EXTENDED_FILTER_TYPE = HiveDatasetFinder.HIVE_DATASET_PREFIX + ".extendedFilterType";
-
-
-  /**
-   * Config to specify class for ensuring that the target directory exists, generally for cloud providers
-   * */
-
   static final Gson gson = new Gson();
 
   private static final String source_client = "source_client";
@@ -315,9 +309,7 @@ public class HiveCopyEntityHelper {
         }
 
         Path targetPath = getTargetLocation(this.targetFs, this.dataset.table.getDataLocation(), Optional.<Partition>absent());
-        if (MapUtils.getBoolean(this.dataset.getProperties(), ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR, false)) {
-          this.dataset.datasetPath = targetPath.toUri().getRawPath();
-        }
+        this.dataset.setDatasetPath(targetPath.toUri().getRawPath());
 
         this.targetTable = getTargetTable(this.dataset.table, targetPath);
         HiveSpec tableHiveSpec = new SimpleHiveSpec.Builder<>(targetPath)
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
@@ -44,8 +44,10 @@ import lombok.Getter;
 import lombok.Singular;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gobblin.commit.CommitStep;
+import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.State;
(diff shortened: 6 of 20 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -351,11 +358,21 @@
         } else {
           this.existingTargetTable = Optional.absent();
         }
-
```
aplex, line 354: This constructor is too complex, and it's hard to follow what's going on here. You can look if it can be refactored and split into smaller functions. If the refactoring is big and touches a lot of code, you can send it as a separate PR, before you main change.

You can also install this IntelliJ plugin that will tell you if the function is too complex, right in the IDE: https://plugins.jetbrains.com/plugin/12159-codemetrics

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveTargetDirectoryClient.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.gobblin.data.management.copy.hive;
+
+import java.io.IOException;
+import java.util.Properties;
+import org.apache.hadoop.fs.Path;
+
+
+public class HiveTargetDirectoryClient {
```
aplex, line 8: You can add some docs all around, explaining how this can be used and extended

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveTargetDirectoryClient.java
```diff
@@ -0,0 +1,36 @@
...
+  public Path getOrCreateTargetPath(Path path) throws IOException {
+    return path;
+  }
+
+  /**
+   * Deletes references to the path from the directory client
+   * @param path path on target directory
+   * @throws IOException if deleting the path fails
+   */
+  public void deletePath(Path path) throws IOException {
```
aplex, line 32: This is a surprising behavior - we call delete and nothing happens by default. I would expect it to delete a thing.

Let's think about how this interface can be made more obvious.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveTargetDirectoryClient.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Properties;
+import org.apache.hadoop.fs.Path;
+
+
+/**
+ * Determines and manages target paths for {@link HiveCopyEntityHelper}
+ * An example implementation would create target directories on cloud environments before initiating the copy
+ * By default, returns the input path (do nothing)
+ */
+public class HiveTargetDirectoryClient {
```
aplex, line 13: This can be an interface and a default implementation class.

aplex (commented):
Things look better after refactoring.

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
```diff
@@ -197,6 +197,14 @@ public static void cleanTaskStagingData(State state, Logger logger) throws IOExc
           throw new IOException("Clean up output directory " + outputPath.toUri().getPath() + " failed");
         }
       }
+
+      Path errPath = new Path(state.getProp(ConfigurationKeys.ROW_LEVEL_ERR_FILE));
```
aplex, line 201: Is this cleanup related to sharding, or it is some separate improvement?

Also, is this error file not part of staging/output folder that are cleaned up earlier in this function?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -370,9 +366,25 @@ public Void call() {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
-          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.IS_DATASET_STAGING_DIR_USED,false))) {
-            workUnit.setProp(DATASET_STAGING_DIR_PATH, ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH));
+
+          // Ensure that the writer temporary directories are contained within the dataset shard
+          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.USE_SHARDED_WRITER_DIRS,false))) {
```
aplex, line 371: Would be good to add a bunch of unit tests, verifying that dataset-local-staging flag works as expect. This should help us iron out typos before doing end-to-end tests.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -210,7 +210,6 @@
   public static final String MAXIMUM_JAR_COPY_RETRY_TIMES_KEY = JOB_JAR_FILES_KEY + ".uploading.retry.maximum";
   public static final String USER_DEFINED_STATIC_STAGING_DIR = "user.defined.static.staging.dir";
   public static final String USER_DEFINED_STAGING_DIR_FLAG = "user.defined.staging.dir.flag";
-  public static final String IS_DATASET_STAGING_DIR_USED = "dataset.staging.dir.used";
```
aplex, line 213: Note: need to update the docs/wiki pages after this is merged, to reflect the new name.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -370,9 +368,16 @@ public Void call() {
...
+            String datasetPath = ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH);
+            workUnit.setProp(ConfigurationKeys.WRITER_STAGING_DIR, datasetPath + ConfigurationKeys.STAGING_DIR_DEFAULT_SUFFIX + "/" + state
```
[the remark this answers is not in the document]
Will-Lo: @aplex Yes I was thinking whether or not to allow both configs and have the `WRITER_STAGING_DIR` config work as specifying the suffix when `USE_SHARDED_WRITER_DIRS` is enabled, or to keep them fully mutually exclusive. Will ask SRE for their input on this as it'll cause quite a few configurations to be modified, but that's be unavoidable regardless what we go with.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -130,7 +130,9 @@ public HiveDataset(FileSystem fs, HiveMetastoreClientPool clientPool, Table tabl
     this.tableIdentifier = this.table.getDbName() + "." + this.table.getTableName();
     Path tableLocation = this.table.getPath();
     if (!(this.properties.isEmpty())) {
```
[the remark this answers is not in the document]
Will-Lo: Yes you're right I'll check specifically for the added flag in the state instead, not sure why it was implemented like this @autumnust if you have more context.

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
Will-Lo: The error file is specified as a separate path from the writer output/staging dirs, so I considered it as another temporary job-level directory that has to be modified to work on the dataset-level. Hence why it has to be added in cleanup here.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
Will-Lo: Yes I'll add the unit tests to the source file. Wasn't sure if it was going to be done on the source side or writer side but the tests would be similar regardless where we implement the logic in the end.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -1062,4 +1061,11 @@
    */
   public static final String TASK_EVENT_METADATA_GENERATOR_CLASS_KEY = "gobblin.task.event.metadata.generator.class";
   public static final String DEFAULT_TASK_EVENT_METADATA_GENERATOR_CLASS_KEY = "nooptask";
+
+  /**
+   * Configuration for sharded directory files
+   */
+  public static final String USE_SHARDED_WRITER_DIRS = "gobblin.dataset.sharded.work.dirs";
```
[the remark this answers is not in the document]
Will-Lo: I like `gobblin.useDatasetLocalWorkDir`, thanks for the suggestion :)

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -225,8 +226,13 @@ private void publishFileSet(CopyEntity.DatasetAndPartition datasetAndPartition,
...
-    CopyableDatasetMetadata metadata = CopyableDatasetMetadata
-        .deserialize(datasetWorkUnitStates.iterator().next().getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
+    CopyableDatasetMetadata metadata = CopyableDatasetMetadata.deserialize(
+        datasetWorkUnitStates.iterator().next().getProp(CopySource.SERIALIZED_COPYABLE_DATASET));
+
+    // Get the dataset specific sharded directory if configured
+    if (state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR)) {
+      this.writerOutputDir = new Path(datasetWorkUnitStates.iterator().next().getProp(ConfigurationKeys.WRITER_OUTPUT_DIR));
+    }
     Path datasetWriterOutputPath = new Path(this.writerOutputDir, datasetAndPartition.identifier());
```
autumnust, line 236: 1. Can we store `datasetWorkUnitStates.iterator().next()` in a temp variable like `sampledWUS` to avoid calling twice? 
2. `writerOutputDir` in this case should be a local variable since it will be specific to a workunitstate. Maybe converting the whole thing including line 236 into a ternary operator like `Path datasetWriterOutputPath = state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR) ? xxx : this.writerOutputDir` ? just leave the `writerOutputDir` as null otherwise might be safer: we can risk fail the job but not writing something that are not supposed to be in a dataset folder

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -79,7 +79,7 @@
 @Slf4j
 public class CopyDataPublisher extends DataPublisher implements UnpublishedHandling {
 
-  private final Path writerOutputDir;
+  private Path writerOutputDir;
```
autumnust, line 82: if adopting suggestion below this can stay as final I guess?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -370,9 +366,25 @@ public Void call() {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
-          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.IS_DATASET_STAGING_DIR_USED,false))) {
-            workUnit.setProp(DATASET_STAGING_DIR_PATH, ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH));
+
+          // Ensure that the writer temporary directories are contained within the dataset shard
+          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR,false))) {
```
autumnust, line 371: It is OK to do it temporarily but using `instanceOf` in `CopySource` definitely seems like a abstraction leaking ( I know it was there already). 

We don't have to address this in this PR, but do you think we should put something like `getDatasetPath` as part of `CopyableDatasetBase` interface?  So that we don't need to explicitly cast them

autumnust (changes requested):
Haven't gone through it fully. This is my first scan.

sv2000:
@Will-Lo Thanks for the PR. Going through the PR, I have a different proposal for how dataset-specific logic can be injected. We could invoke one or more configurable handler classes inside the AbstractJobLauncher right after the workunit creation step. For your use case, you could use the DATASET_URN property which is set in different source classes. You could provide an implementation of a handler that sets dataset-specific staging/output dirs. Please take a look at the TaskStateCollectorServiceHandler class as an example, which is invoked on task completion. What I am proposing here is an identical solution, but invoked before a job is launched.  

This high level approach is leverageable for future use cases e.g. where a Gobblin pipeline needs to write to a Table (e.g. Iceberg) and we could define a handler that ensures table is created before the job is started. 

Happy to discuss offline if you have questions.

gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
```diff
@@ -0,0 +1,72 @@
...
+ * Reads {@link ConfigurationKeys#DESTINATION_DATASET_HANDLER_CLASS} as a list
+ * of classes, separated by comma to initialize the handlers
+ */
+public class DestinationDatasetHandlerService implements Closeable {
+  List<DestinationDatasetHandler> handlers;
+
+  public DestinationDatasetHandlerService(SourceState jobState) {
+    this.handlers = new ArrayList<>();
+    if (jobState.contains(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)) {
+      String[] handlerList = jobState.getProp(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS).split(",");
```
aplex, line 42: I suggest to trim the whitespaces from splitted values . I've seen that people like to separate multiple value in configs with a coma and a space -"value1, value2, value3".

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -370,9 +366,13 @@ public Void call() {
               workUnit.setProp(ConfigurationKeys.COPY_EXPECTED_SCHEMA, ((ConfigBasedDataset) this.copyableDataset).getExpectedSchema());
             }
           }
-          if ((this.copyableDataset instanceof HiveDataset) && (state.getPropAsBoolean(ConfigurationKeys.IS_DATASET_STAGING_DIR_USED,false))) {
-            workUnit.setProp(DATASET_STAGING_DIR_PATH, ((HiveDataset) this.copyableDataset).getProperties().getProperty(DATASET_STAGING_PATH));
+
+          // Ensure that the writer temporary directories are contained within the dataset shard
+          if (state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR,false)) {
```
sv2000, line 371: Should we just set DATASET_DESTINATION_PATH regardless of whether we want to use DATASET_LOCAL_WORK_DIR?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -354,35 +313,24 @@
           this.existingTargetTable = Optional.absent();
         }
 
-        // Constructing CommitStep object for table registration
-        Path targetPath = getTargetLocation(this.dataset.fs, this.targetFs, this.dataset.table.getDataLocation(),
-            Optional.<Partition> absent());
+        Path targetPath = getTargetLocation(this.targetFs, this.dataset.table.getDataLocation(), Optional.<Partition>absent());
+        if (Boolean.parseBoolean(this.dataset.getProperties().getProperty(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR, "false"))) {
```
aplex, line 317: We have a bunch of utils to work with config/properties/maps - ConfigUtils.getBoolean() , State.getPropAsBoolean()
If none of them work here, you can use  MapUtils.getBooleanValue from Apache Commons.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
```diff
@@ -117,8 +117,9 @@ public CopyDataPublisher(State state) throws IOException {
     this.fs = FileSystem.get(URI.create(uri), WriterUtils.getFsConfiguration(state));
 
     FileAwareInputStreamDataWriterBuilder.setJobSpecificOutputPaths(state);
-
-    this.writerOutputDir = new Path(state.getProp(ConfigurationKeys.WRITER_OUTPUT_DIR));
+    // If directories are sharded by dataset, initialize writers using workunit state instead
+    this.writerOutputDir = state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR) ?
```
sv2000, line 121: Shouldn't WRITER_OUTPUT_DIR be set inside the handler? If so, we don't need these changes here, do we?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriterBuilder.java
```diff
@@ -52,14 +53,13 @@
...
     if (!StringUtils.containsIgnoreCase(state.getProp(ConfigurationKeys.WRITER_STAGING_DIR),
-        state.getProp(ConfigurationKeys.JOB_ID_KEY))) {
-
-      state.setProp(ConfigurationKeys.WRITER_STAGING_DIR, new Path(state.getProp(ConfigurationKeys.WRITER_STAGING_DIR),
-          state.getProp(ConfigurationKeys.JOB_ID_KEY)));
-      state.setProp(ConfigurationKeys.WRITER_OUTPUT_DIR, new Path(state.getProp(ConfigurationKeys.WRITER_OUTPUT_DIR),
-          state.getProp(ConfigurationKeys.JOB_ID_KEY)));
-
+        state.getProp(ConfigurationKeys.JOB_ID_KEY)) && !state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR)) {
+        state.setProp(ConfigurationKeys.WRITER_STAGING_DIR, new Path(state.getProp(ConfigurationKeys.WRITER_STAGING_DIR),
```
sv2000, line 59: Can all of this be moved inside the handler?

sv2000 (changes requested):
Is looking better. But I think we should aim mostly to leave source/writer/publishers unchanged and move dataset-specific logic inside the DatasetHandler impl.

gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
Will-Lo: This is a really good point, thanks for the catch

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/publisher/CopyDataPublisher.java
sv2000: I don't understand the explanation. What I am suggesting is to have the ConfigurationKeys#WRITER_OUTPUT_DIR being overwritten inside the DatasetHandler impl, which would avoid the need for this change.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriterBuilder.java
sv2000: Same comment as earlier. We should be able to set override the WRITER_STAGING_DIR config inside the DatasetHandler and leave the code here unchanged.

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -912,13 +915,14 @@ private void cleanLeftoverStagingData(WorkUnitStream workUnits, JobState jobStat
     }
 
     try {
-      if (this.jobContext.shouldCleanupStagingDataPerTask()) {
+      if (this.jobContext.shouldCleanupStagingDataPerTask() || jobState.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR)) {
```
sv2000, line 918: Why do we need to change the condition in the if (..) statement?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -972,7 +974,7 @@ private void cleanupStagingData(JobState jobState)
       throw new JobException("Failed to check unfinished commit sequences", e);
     }
 
-    if (this.jobContext.shouldCleanupStagingDataPerTask()) {
+    if (this.jobContext.shouldCleanupStagingDataPerTask() || jobState.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR)) {
```
sv2000, line 977: Same comment as above.

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
```diff
@@ -197,6 +197,16 @@ public static void cleanTaskStagingData(State state, Logger logger) throws IOExc
           throw new IOException("Clean up output directory " + outputPath.toUri().getPath() + " failed");
         }
       }
+
+      if (state.contains(ConfigurationKeys.ROW_LEVEL_ERR_FILE) && state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR, false)) {
```
sv2000, line 201: We should avoid leaking details about dataset-specific local work dir all over the place. This change seems unnecessary too.

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
```diff
@@ -235,6 +245,14 @@ public static void cleanTaskStagingData(State state, Logger logger, Closer close
         logger.info("Cleaning up output directory " + outputPath.toUri().getPath());
         parallelRunner.deletePath(outputPath, true);
       }
+
+      if (state.contains(ConfigurationKeys.ROW_LEVEL_ERR_FILE) && state.getPropAsBoolean(ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR, false)) {
```
sv2000, line 249: same comment as earlier.

sv2000 (changes requested):

gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandler.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Collection;
+import org.apache.gobblin.source.workunit.WorkUnit;
+
+/**
+ * Performs work related to initializing the target environment before the files are written and published
+ */
+public interface DestinationDatasetHandler extends Closeable {
+
+  /**
+   * Adds required properties to work units for determining writer staging and output paths
```
autumnust, line 31: The purpose for this class doesn't have to be very specific to the sharding thing.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/writer/FileAwareInputStreamDataWriterBuilder.java
autumnust: @Will-Lo  If your concern is about hive-distcp user with shard configuration: It should be fine to move this into handler only since I don't think there's any real usage existed for shard-aware distcp

autumnust (changes requested):

gobblin-core/src/main/java/org/apache/gobblin/destination/DestinationDatasetHandlerService.java
```diff
@@ -0,0 +1,80 @@
...
+ * Reads {@link ConfigurationKeys#DESTINATION_DATASET_HANDLER_CLASS} as a list
+ * of classes, separated by comma to initialize the handlers
+ */
+public class DestinationDatasetHandlerService implements Closeable {
+  List<DestinationDatasetHandler> handlers;
+
+  public DestinationDatasetHandlerService(SourceState jobState, Boolean canCleanUp, EventSubmitter eventSubmitter) {
+    this.handlers = new ArrayList<>();
+    if (jobState.contains(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)) {
+      String[] handlerList = jobState.getProp(ConfigurationKeys.DESTINATION_DATASET_HANDLER_CLASS)
```
sv2000, line 44: PropertiesUtils.getPropAsList()?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopyableDatasetBase.java
```diff
@@ -18,11 +18,13 @@
 package org.apache.gobblin.data.management.copy;
 
 import org.apache.gobblin.dataset.Dataset;
+import org.apache.hadoop.fs.Path;
```
sv2000, line 21: Unused import?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveCopyEntityHelper.java
```diff
@@ -354,35 +314,24 @@
           this.existingTargetTable = Optional.absent();
         }
 
-        // Constructing CommitStep object for table registration
-        Path targetPath = getTargetLocation(this.dataset.fs, this.targetFs, this.dataset.table.getDataLocation(),
-            Optional.<Partition> absent());
+        Path targetPath = getTargetLocation(this.targetFs, this.dataset.table.getDataLocation(), Optional.<Partition>absent());
+        if (MapUtils.getBoolean(this.dataset.getProperties(), ConfigurationKeys.USE_DATASET_LOCAL_WORK_DIR, false)) {
+          this.dataset.datasetPath = targetPath.toUri().getRawPath();
```
sv2000, line 319: Why does datasetPath depend on whether or not we are using dataset local working dir?

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
```diff
@@ -132,10 +132,9 @@ public static void cleanStagingData(List<? extends State> states, Logger logger)
    * @param logger a {@link Logger} used for logging
    */
   public static void cleanJobStagingData(State state, Logger logger) throws IOException {
-    Preconditions.checkArgument(state.contains(ConfigurationKeys.WRITER_STAGING_DIR),
-        "Missing required property " + ConfigurationKeys.WRITER_STAGING_DIR);
-    Preconditions.checkArgument(state.contains(ConfigurationKeys.WRITER_OUTPUT_DIR),
-        "Missing required property " + ConfigurationKeys.WRITER_OUTPUT_DIR);
+    if (!state.contains(ConfigurationKeys.WRITER_STAGING_DIR) || !state.contains(ConfigurationKeys.WRITER_OUTPUT_DIR)) {
```
sv2000, line 135: Shouldn't this be invoked after the workunit stream has been computed? If so, shouldn't we expect both these config keys to be set?

sv2000: This looks like a change in behavior from what existed earlier.

sv2000 (changes requested):

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/hive/HiveDataset.java
```diff
@@ -86,14 +86,15 @@
...
   public static final String TABLE = "Table";
-  public static final String DATASET_STAGING_PATH = "dataset.staging.path";
 
   public static final String DATABASE_TOKEN = "$DB";
   public static final String TABLE_TOKEN = "$TABLE";
 
   public static final String LOGICAL_DB_TOKEN = "$LOGICAL_DB";
   public static final String LOGICAL_TABLE_TOKEN = "$LOGICAL_TABLE";
 
+  public String datasetPath;
```
sv2000, line 96: Should this be private?

sv2000 (changes requested):

gobblin-utility/src/main/java/org/apache/gobblin/util/JobLauncherUtils.java
Will-Lo: This checks the job level state, not the workunit. So it would fail if we do not specify the temporary directories.

sv2000 (approved):
+1. LGTM.

autumnust (approved):

PR #3176, opened 2020-12-16
[GOBBLIN-1340] Extend Azkaban client

aplex:
We extend Azkaban client with API calls to check project existence,
and get list of the flows. Log retrieval API is refactored to stream
logs and use strongly-typed parameters.

Azkaban client tests can now run concurrently, and should be more stable.

These changes will be used in the end-to-end Gobblin testing framework.

### DIFF
--- a/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
+++ b/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
@@ -17,19 +17,17 @@
 
 package org.apache.gobblin.service.modules.orchestration;
 
-import java.io.Closeable;
-import java.io.File;
-import java.io.IOException;
-import java.time.Duration;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
+import com.github.rholder.retry.*;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import com.google.common.io.Closer;
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import lombok.Builder;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
@@ -46,24 +44,14 @@ import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.github.rholder.retry.AttemptTimeLimiters;
-import com.github.rholder.retry.RetryException;
-import com.github.rholder.retry.Retryer;
-import com.github.rholder.retry.RetryerBuilder;
-import com.github.rholder.retry.StopStrategies;
-import com.github.rholder.retry.WaitStrategies;
-import com.google.common.base.Optional;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
-import com.google.common.io.Closer;
-import com.google.common.util.concurrent.SimpleTimeLimiter;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
-
-import lombok.Builder;
-
-import org.apache.gobblin.util.ExecutorsUtils;
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.*;
 
 
 /**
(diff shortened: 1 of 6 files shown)


gobblin-modules/gobblin-azkaban/src/test/java/org/apache/gobblin/service/modules/orchestration/AzkabanClientTest.java
```diff
@@ -65,55 +63,73 @@ public void setup() throws Exception {
...
 
-  @Test(enabled = false)
-  public void testCreateProject() throws AzkabanClientException {
-    String projectName = "project-create";
-    String description = "This is a create project test.";
+    AzkabanExecuteFlowStatus execStatus = this.client.executeFlow(projectName, flowName, Maps.newHashMap());
+    String execId = execStatus.getResponse().getExecId();
+
+    ByteArrayOutputStream logStream = null;
+    int maxTries = 10;
```
autumnust, line 98: Considering smaller value?

gobblin-modules/gobblin-azkaban/src/test/java/org/apache/gobblin/service/modules/orchestration/AzkabanClientTest.java
```diff
@@ -48,9 +41,14 @@
  * Please check https://azkaban.github.io/azkaban/docs/latest/ for how to setup Azkaban-solo-server.
  */
 @Slf4j
+@Test(enabled = false)
```
autumnust, line 44: Intentional ?

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -306,6 +317,22 @@ public AzkabanClientStatus deleteProject(String projectName) throws AzkabanClien
...
+   * Checks if the project with specified name exists in Azkaban
+   */
+  public Boolean projectExists(String projectName) throws AzkabanClientException {
+    try {
+      fetchProjectFlows(projectName);
+      return true;
+    } catch (AzkabanClientException e) {
+      // Azkaban does not return a strongly typed error code, so we are checking the message
+      if (e.getCause().getMessage().contains("doesn't exist")) {
+        return false;
```
autumnust, line 330: Shall we print the error message anyway ? relying on a strict string matching here seems to be not reliable enough.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanProjectFlowsStatus.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.apache.gobblin.service.modules.orchestration;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+import java.util.List;
+
+public class AzkabanProjectFlowsStatus extends AzkabanClientStatus<AzkabanProjectFlowsStatus.Project> {
```
autumnust, line 24: I am not quite get the need for this class with two static inner class within it. Can you add a javadoc, and why this is needed ( Can the `Project` and `Flow` class themselves alone serve the same purpose? )

autumnust (changes requested):

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
```diff
@@ -97,7 +86,8 @@ protected AzkabanClient(String username,
                           long sessionExpireInMin,
                           CloseableHttpClient httpClient,
                           SessionManager sessionManager,
-                          ExecutorService executorService)
+                          ExecutorService executorService,
```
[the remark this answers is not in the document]
aplex: This is a builder constructor. I don't think it makes sense to have a backward-compatible overload every time we add a setting here, we'll end up wit a lot of overloads.

gobblin-modules/gobblin-azkaban/src/test/java/org/apache/gobblin/service/modules/orchestration/AzkabanClientTest.java
aplex: There is a comment explaining this later on, moved it higher for clarity.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanClient.java
aplex: This is an outcome of a poorly-designed REST API on Azkaban side. It does not have a direct way to check for project existence, and it does not return machine-readable error codes - https://azkaban.readthedocs.io/en/latest/ajaxApi.html#fetch-jobs-of-a-flow . This can be unreliable, if they change the messages. I'll add a feature request for them to have better error reporting in the API.

Since the error is an expected behavior here (when project does not exist), we shouldn't log it every time it happens - it will flood the logs when nothing bad has happened.

gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/service/modules/orchestration/AzkabanProjectFlowsStatus.java
aplex: Added a more extended comment in code. That's how all other API calls are designed for this Azkaban client.

autumnust (approved):

PR #3177, opened 2020-12-16
[GOBBLIN-1341] Make cluster name resolution more robust

aplex:
We extend cluster name conversion to handle names with
and without url schemes in config. In practice, scheme
and port numbers are irrelevant, as a single host usually
maps to a single cluster.

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
@@ -122,14 +122,22 @@ public class ClustersNames {
         }
 
         candidates.add(uri.getHost());
+      } else if (uri.getScheme() != null && uri.getPath() != null) {
+        // we have a scheme and a path, but not the host name
+        // assuming local host
+        candidates.add("localhost");
       } else {
-        candidates.add(clusterIdentifier.replaceAll("[/:]", " ").trim().replaceAll(" ", "_"));
+        candidates.add(getNormalizedName(clusterIdentifier));
       }
     } catch (URISyntaxException e) {
-      //leave ID as is
+      candidates.add(getNormalizedName(clusterIdentifier));
     }
 
     return candidates;
+  }
+
+  private static String getNormalizedName(String clusterIdentifier) {
+    return clusterIdentifier.replaceAll("[^\\w-\\.]", "_");
   }
 
   /**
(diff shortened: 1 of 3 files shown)


gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
```diff
@@ -97,21 +110,34 @@ public void addClusterMapping(URL clusterUrl, String clusterName) {
...
     } catch (URISyntaxException e) {
-      //leave ID as is
+      candidates.add(getNormalizedName(clusterIdentifier));
     }
 
-    return clusterIdentifier;
+    return candidates;
+  }
+
+  private static String getNormalizedName(String clusterIdentifier) {
```
autumnust, line 139: I am not quite following the purpose for this.

gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
```diff
@@ -97,21 +110,34 @@ public void addClusterMapping(URL clusterUrl, String clusterName) {
...
       URI uri = new URI(clusterIdentifier.trim());
-      // URIs without protocol prefix
-      if (!uri.isOpaque() && null != uri.getHost()) {
-        clusterIdentifier = uri.getHost();
+      if (uri.getHost() != null) {
+        if (uri.getPort() != -1) {
+          candidates.add(uri.getHost() + ":" + uri.getPort());
+        }
+
+        candidates.add(uri.getHost());
```
autumnust, line 124: I don't see any chance for host:port to be picked here since getHost will always be placed after that. Is this intentional ? Or can you give an unit test case to return host:port

autumnust (changes requested):

gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
aplex: Added a comment here. Also check out "testPortSpecificOverrides" test

gobblin-utility/src/main/java/org/apache/gobblin/util/ClustersNames.java
aplex: Added logic explanation to getClusterName(String). Let me know if it is sufficiently clear there.

autumnust (approved):

PR #3178, opened 2020-12-16
[GOBBLIN-1339] Add cluster name to dataset descriptor

aplex:
We use dataset descriptors to track lineage. Previously, it
only included the platform name (hive,hdfs) and path of the
dataset. As a result, we could not differentiate the data copy
between multiple production clusters, as the dataset descriptors
were the same for them. We add an optional storage system url to
address that.

This change will be used for data copy audit system.

Hive and file-based copy code is updated to include storage urls.

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
@@ -19,11 +19,11 @@ package org.apache.gobblin.dataset;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
-import lombok.Getter;
-
-import javax.annotation.Nullable;
+import java.net.URI;
 import java.util.Map;
 import java.util.Objects;
+import javax.annotation.Nullable;
+import lombok.Getter;
 
 
 /**
@@ -32,7 +32,7 @@ import java.util.Objects;
 public class DatasetDescriptor extends Descriptor {
   private static final String PLATFORM_KEY = "platform";
   private static final String NAME_KEY = "name";
-  private static final String CLUSTER_NAME_KEY = "clusterName";
+  private static final String STORAGE_URL = "storageUrl";
 
   /**
    * which platform the dataset is stored, for example: local, hdfs, oracle, mysql, kafka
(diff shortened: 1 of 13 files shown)


aplex:
This PR depends on https://github.com/apache/incubator-gobblin/pull/3177 , and should pass CI after the dependency is merged.

gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
```diff
@@ -17,34 +17,52 @@
...
   private final String platform;
 
+  /**
+   * Human-readeable cluster name.
+   *
+   * @see org.apache.gobblin.util.ClustersNames
+   */
+  @Getter
+  @Nullable
+  private final String clusterName;
```
sv2000, line 50: What exactly does clusterName mean? Is this the name of the storage instance that has an instance of a given dataset? If so, I think storageInstance or storageInstanceName would better capture what this field is.

gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
```diff
@@ -17,34 +17,52 @@
...
 
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.Objects;
+
 
 /**
  * A {@link Descriptor} identifies and provides metadata to describe a dataset
  */
 public class DatasetDescriptor extends Descriptor {
```
sv2000, line 32: One comment: it is not clear from this class if it was intended to capture the notion of a Dataset instance. If so, there might be advantages to extending this class to a new descriptor that defines a dataset instance where the "clusterName" is mandatory.

gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
sv2000: What is the expected format for the clusterName? Can it be any string? I think it would be better to define a new class called "StorageIdentifier" that includes among other things fields such as logicalName, address, port, availability zone, etc.  We should also define a URN representation for the storage instance.

sv2000 (changes requested):

gobblin-api/src/main/java/org/apache/gobblin/dataset/DatasetDescriptor.java
aplex: Initially, I was thinking about putting a well-known cluster name like "snap" into this field, and for unknown clusters put a DNS name like "datalake.corp.com". Then lineage events will include this name as-is, and it can be displayed to users directly.  We already use "cluster name" concept in org.apache.gobblin.util.ClustersNames, for getting nice names from yarn/hdfs urls.

However, this well-know name is not easily convertible back to domain name. Single well-known name can map to different urls for hdfs/hive. So, there could be a benefit in keeping a hostname instead of clustername, and converting it to human-readable name before reporting lineage. 

DatasetDescriptor already looks like a dataset URL with non-standard field names:
- platform is like url.protocol - "hdfs", "kafka", "mysql"
- name is like url.path - "table1" or "/data/tracking/test123"
- clusterName is like a url.host - "my-cluster" or "my-cluster.company.com"

Regarding the fields you suggested:
- I don't think we have availability zone information anywhere.
- We can get well-known nice name from hostname using org.apache.gobblin.util.ClustersNames , so don't have to store it separately.
- Keeping port information is discussable. I would assume that any production deployments will have separate DNS names for different clusters. However, for small scale, temporary deployments it's possible to have multiple Hives or namenodes on a single host, running on different ports.

sv2000 (approved):
+1. LGTM.

sv2000:
@aplex Check the travis build failure.

PR #3179, opened 2020-12-16
[GOBBLIN-1342] Add API to resume a flow

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1342

### Description

This PR adds flow resume functionality, which means after a flow fails, resume can be called on it, and the flow will be reattempted from the point before it failed (i.e. not rerunning the successful jobs).

How it works:
- Whenever a flow fails, on cleanup, it is now added to a `failedDagStateStore`, similar to the normal dag state store. (It is also kept in an in-memory `failedDags` map to avoid querying the store for every resume request).
- User triggers a resume request by calling partial update on the FlowExecutions endpoint, updating the status from `FAILED` to `RUNNING`. This is a little confusing, but as per [this discussion](https://softwareengineering.stackexchange.com/questions/261552/which-http-verb-should-i-use-to-trigger-an-action-in-a-rest-web-service) it seems to be the most correct way to represent a triggered action like this in rest.
- The request is forwarded to the `DagManager` as a `ResumeFlowEvent` through the eventbus, similar to how killing a flow works.
- `DagManager` adds it to a `resumeQueue` to be picked up by a thread
- The thread calls `beginResumingDag` on the dag, which sets the status of both the flow and each FAILED/CANCELLED job to `PENDING_RESUME` (and also sends events so this will be updated in the status store). The dag is then added to another map `resumingDags`
- In `finishResumingDags`, the dags in `resumingDags` are checked by querying the status store. If the dag has reached the correct state, then it is passed back to `initialize` to be launched. This is separated from `beginResumingDag` because it takes some time for the status to be updated in the status store.

Other related changes:
- Fixed a bug where `addDag` sets all jobs to `PENDING`, even if the dag is already running. This would cause their initial jobs to get relaunched on service restart.
- Added a `flowStartTime` variable in `JobExecutionPlan` which is set on flow resume so that flow SLA will not be calculated based on the original flow.
- Added `PENDING_RESUME` to execution status enum, it is a special status required because resuming flows are allowed to move "backwards" in status from `FAILED` to `PENDING`.

### Tests
Add unit test

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -540,22 +540,26 @@ public class DagManager extends AbstractIdleService {
     private void finishResumingDags() throws IOException {
       for (Map.Entry<String, Dag<JobExecutionPlan>> dag : this.resumingDags.entrySet()) {
         JobStatus flowStatus = pollFlowStatus(dag.getValue());
-        if (!flowStatus.getEventName().equals(PENDING_RESUME.name())) {
-          return;
+        if (flowStatus == null || !flowStatus.getEventName().equals(PENDING_RESUME.name())) {
+          continue;
         }
 
+        boolean dagReady = true;
         for (DagNode<JobExecutionPlan> node : dag.getValue().getNodes()) {
           JobStatus jobStatus = pollJobStatus(node);
           if (jobStatus == null || jobStatus.getEventName().equals(FAILED.name()) || jobStatus.getEventName().equals(CANCELLED.name())) {
-            return;
+            dagReady = false;
+            break;
           }
         }
 
-        this.dagStateStore.writeCheckpoint(dag.getValue());
-        this.failedDagStateStore.cleanUp(dag.getValue());
-        this.failedDags.remove(dag.getKey());
-        this.resumingDags.remove(dag.getKey());
-        initialize(dag.getValue());
+        if (dagReady) {
+          this.dagStateStore.writeCheckpoint(dag.getValue());
+          this.failedDagStateStore.cleanUp(dag.getValue());
+          this.failedDags.remove(dag.getKey());
+          this.resumingDags.remove(dag.getKey());
+          initialize(dag.getValue());
+        }
       }
     }
 
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -79,6 +79,7 @@ import org.apache.gobblin.service.monitoring.JobStatusRetriever;
 import org.apache.gobblin.service.monitoring.KafkaJobStatusMonitor;
 import org.apache.gobblin.service.monitoring.KafkaJobStatusMonitorFactory;
 import org.apache.gobblin.service.monitoring.KillFlowEvent;
+import org.apache.gobblin.service.monitoring.ResumeFlowEvent;
 import org.apache.gobblin.util.ConfigUtils;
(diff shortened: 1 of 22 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -462,6 +501,64 @@ public void run() {
...
+     * Finish resuming dags by first verifying the status is correct (flow should be {@link ExecutionStatus#PENDING_RESUME}
+     * and jobs should not be {@link ExecutionStatus#FAILED} or {@link ExecutionStatus#CANCELLED}) and then calling
+     * {@link #initialize}. This is separated from {@link #beginResumingDag} because it could take some time for the
+     * job status state store to reflect the updated status.
+     */
+    private void finishResumingDags() throws IOException {
+      for (Map.Entry<String, Dag<JobExecutionPlan>> dag : this.resumingDags.entrySet()) {
+        JobStatus flowStatus = pollFlowStatus(dag.getValue());
+        if (!flowStatus.getEventName().equals(PENDING_RESUME.name())) {
+          return;
```
sv2000, line 544: Should this be continue instead of return? Don't you want to process the remiaining dags?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -462,6 +501,64 @@ public void run() {
...
+        }
+
+        for (DagNode<JobExecutionPlan> node : dag.getValue().getNodes()) {
+          JobStatus jobStatus = pollJobStatus(node);
+          if (jobStatus == null || jobStatus.getEventName().equals(FAILED.name()) || jobStatus.getEventName().equals(CANCELLED.name())) {
+            return;
```
sv2000, line 550: Same comment as earlier.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -462,6 +501,64 @@ public void run() {
...
+    /**
+     * Finish resuming dags by first verifying the status is correct (flow should be {@link ExecutionStatus#PENDING_RESUME}
+     * and jobs should not be {@link ExecutionStatus#FAILED} or {@link ExecutionStatus#CANCELLED}) and then calling
+     * {@link #initialize}. This is separated from {@link #beginResumingDag} because it could take some time for the
+     * job status state store to reflect the updated status.
+     */
+    private void finishResumingDags() throws IOException {
+      for (Map.Entry<String, Dag<JobExecutionPlan>> dag : this.resumingDags.entrySet()) {
+        JobStatus flowStatus = pollFlowStatus(dag.getValue());
+        if (!flowStatus.getEventName().equals(PENDING_RESUME.name())) {
```
sv2000, line 543: Possible NPE here. Is there a possibility that flowStatus is null?

sv2000 (changes requested):
Thanks for this PR! LGTM overall. Couple of potential issues to look at.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: Good catch, yeah it should be continue.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: For this one continue doesn't work, since it is inside a second for loop, so I just added a dagReady flag

sv2000 (approved):
+1. LGTM.

PR #3180, opened 2020-12-17
Fix the data loss issue caused by the cache expiration in PartitionerDataWriter

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1343

### Description

Problem statement:

Previously, we maintain a cache in PartitionedDataWriter to avoid accumulate writer in memory in long running job. But when we expire the writer, we only close it without flush/commit, so it may cause data loss when there is a slowness happening on HDFS.

Potential solution:

In the removal logic, we can make sure the writer has been committed correctly, i.e. force it to commit before close.  But the issue here is we still remove the writer from cache, so next flush message will be handled and return without call commit for the right writer, and watermark will move without data being published to HDFS.
We calculate the time for the write operation, and if it takes a long time, we force to add the writer back to cache so that next flush message will be picked up by the writer. 
Here we use the second solution

### Tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
@@ -101,6 +101,7 @@ public class PartitionedDataWriter<S, D> extends WriterWrapper<D> implements Fin
   private final ControlMessageHandler controlMessageHandler;
   private boolean isSpeculativeAttemptSafe;
   private boolean isWatermarkCapable;
+  private long cacheExpiryInterval;
 
   private ScheduledExecutorService cacheCleanUpExecutor;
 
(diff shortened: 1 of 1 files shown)


gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.writer;
 
+import com.sun.javafx.binding.StringFormatter;
```
autumnust, line 20: `String.format` should serve the same purpose?

gobblin-core/src/main/java/org/apache/gobblin/writer/PartitionedDataWriter.java
```diff
@@ -235,18 +239,24 @@ private boolean isDataWriterWatermarkCapable(DataWriter<D> dataWriter) {
...
+      GenericRecord partition = getPartitionForRecord(recordEnvelope.getRecord());
+      DataWriter<D> writer = this.partitionWriters.get(partition);
+      long startTime = System.currentTimeMillis();
       writer.writeEnvelope(recordEnvelope);
-    } catch (ExecutionException ee) {
+      long timeForWriting = System.currentTimeMillis() - startTime;
+      // If the write take a long time, which is 1/3 of cache expiration time, we fail the writer to avoid data loss
+      // and further slowness on the same HDFS block
+      if (timeForWriting > this.writeTimeoutInterval ) {
+        throw new TimeoutException(StringFormatter.format("Write record took %s ms, but threshold is %s ms",
```
autumnust, line 250: If you will need to catch the `TimeoutException` later again and re-throw as IOE, why not directly throw IOE ?

autumnust (changes requested):

autumnust (approved):
+1

PR #3181, opened 2020-12-18
[GOBBLIN-1344] Enable configurable schema source db and table when registering hive schema

ZihanLi58:
…gistering hive schema

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1344

### Description
We need to set schema.literal for hive table, but in ORC compaction pipeline, there is no context for avro schema. So for de-duped DB which is registered by compaction pipeline, we need to be able to configure the schema source db/table to get the avro schema registered by data ingestion pipeline

### Tests

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -94,7 +94,6 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
   public static final String HIVE_REGISTER_METRICS_PREFIX = "hiveRegister.";
   public static final String ADD_PARTITION_TIMER = HIVE_REGISTER_METRICS_PREFIX + "addPartitionTimerTimer";
   public static final String SCHEMA_SOURCE_DB = HIVE_REGISTER_METRICS_PREFIX + "schema.source.dbName";
-  public static final String SCHEMA_SOURCE_TABLE = HIVE_REGISTER_METRICS_PREFIX + "schema.source.tableName";
   public static final String GET_HIVE_PARTITION = HIVE_REGISTER_METRICS_PREFIX + "getPartitionTimer";
   public static final String ALTER_PARTITION = HIVE_REGISTER_METRICS_PREFIX + "alterPartitionTimer";
   public static final String TABLE_EXISTS = HIVE_REGISTER_METRICS_PREFIX + "tableExistsTimer";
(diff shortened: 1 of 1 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -280,8 +291,15 @@ private boolean ensureHiveTableExistenceBeforeAlternation(String tableName, Stri
         try (Timer.Context context = this.metricContext.timer(GET_HIVE_TABLE).time()) {
           existingTable = HiveMetaStoreUtils.getHiveTable(client.getTable(dbName, tableName));
         }
+        HiveTable schemaSourceTable = existingTable;
+        if (state.contains(SCHEMA_SOURCE_DB) || state.contains(SCHEMA_SOURCE_TABLE)) {
+          try (Timer.Context context = this.metricContext.timer(GET_HIVE_TABLE).time()) {
```
autumnust, line 296: timer shouldn't be named after `GET_HIVE_TABLE` here.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -93,6 +93,8 @@
 
   public static final String HIVE_REGISTER_METRICS_PREFIX = "hiveRegister.";
   public static final String ADD_PARTITION_TIMER = HIVE_REGISTER_METRICS_PREFIX + "addPartitionTimerTimer";
+  public static final String SCHEMA_SOURCE_DB = HIVE_REGISTER_METRICS_PREFIX + "schema.source.dbName";
+  public static final String SCHEMA_SOURCE_TABLE = HIVE_REGISTER_METRICS_PREFIX + "schema.source.tableName";
```
autumnust, line 97: I would rather we getting rid of source table here for simplicity (and add javadoc indicating our assumption that the actual source table will be named the same with only database name being different). wdyt ?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -280,8 +291,15 @@ private boolean ensureHiveTableExistenceBeforeAlternation(String tableName, Stri
         try (Timer.Context context = this.metricContext.timer(GET_HIVE_TABLE).time()) {
           existingTable = HiveMetaStoreUtils.getHiveTable(client.getTable(dbName, tableName));
         }
+        HiveTable schemaSourceTable = existingTable;
+        if (state.contains(SCHEMA_SOURCE_DB) || state.contains(SCHEMA_SOURCE_TABLE)) {
```
autumnust, line 295: Should the information of source DB coming from HiveSpec instead of `state`?  IIRC `state` will be applied for all paths' registration here, it is more like a global state but not for a specific table/db

autumnust (changes requested):

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: Will address this.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
ZihanLi58: In both compaction pipeline and ingestion pipeline, this state is topic specific state. And for this case that db is registered as an additional db, the db name is specified by a job level config. So I think for all the tables in this db, it should has the same source db. That's why I do want to make this config a job level config also.

autumnust (approved):
+1

PR #3187, opened 2021-01-02
[GOBBLIN-1348] Uses own logic for generating tags instead of the docker meta action

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1348

### Description
- Removes use of non-verified third party github action to generate tags, and instead generates tags explicitly in a previous job
- Runs the action if the github action is modified or if docker files are modified, allowing earlier detection of bugs

### Tests

### Commits

### DIFF
--- a/.github/workflows/docker_build_publish.yaml
+++ b/.github/workflows/docker_build_publish.yaml
@@ -21,10 +21,14 @@ on:
     # Publish only on `master`
     branches:
       - master
-
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'gobblin-docker/**'
+      - '.github/workflows/docker_build_publish.yaml'
   release:
-    tags:
-      - 'v*'
+    types: [published, edited]
 
 env:
   IMAGE_NAME: apache/gobblin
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_build_publish.yaml
```diff
@@ -21,44 +21,60 @@ on:
     # Publish only on `master`
     branches:
       - master
-
+  pull_request:
```
abti, line 24: If I understand this correctly, it will trigger the build and push at each pull request creation rather than when a commit is pushed? That would be quite a few docker images with unreviewed code?

Will-Lo: It'll trigger the build whenever there is a push on master, when there is a new release or when there is a new pull request against master (and anytime that branch is updated after the PR is created)

In the build and push action I prevent pushing to the repository if the event is a pull request, so it should only be used to check if the docker file modifications/github actions modifications are valid without creating a new image on dockerhub.

Also it should only be triggering the CI if the PR includes a file that modifies the docker file or the github action

abti: Thanks, just saw the check added in push section. Makes sense.

abti (approved):

.github/workflows/docker_build_publish.yaml
johnfoconnor: There have been no release triggered in github for many versions
https://github.com/apache/gobblin/releases


Because the docker image build is now tied to a release, there are actually no docker images available 
https://hub.docker.com/r/apache/gobblin/tags

Please create a release for the latest (0.16.0) so we can get a proper container built

https://issues.apache.org/jira/browse/GOBBLIN-1555

PR #3196, opened 2021-01-09
[GOBBLIN-1357] Updates k8s and docker GaaS to use up-to-date images and non-deprecated ingress

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1357

### Description

 - Uses new images from apache/gobblin dockerhub. Currently there are no images in the repository so they must be generated from local and pushed to the minikube cache until dockerhub is populated with an image
 - Fixes bug where arguments with spaces was not properly passed in to docker through entrypoint script
 - Fixes minor bugs with k8s cluster including mysql version bump to support json syntax, bug with standalone not picking up jobs due to environment variable not being set
 - Adds flag for Gobblin instances to be cgroup memory aware when running on k8s, so that the java application will properly utilize the memory that's allocated to the k8s pod without having to change the runtime args by default.

### Tests

### Commits

### DIFF
--- a/gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
+++ b/gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
@@ -27,12 +27,14 @@ spec:
             name: flowconfig-templates
       containers:
         - name: gobblin-service
-          image: will97/gobblin-as-a-service:latest
+          image: apache/gobblin:latest
+          args: ["--mode", "gobblin-as-service", "--jvmopts", "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"]
           volumeMounts:
             - name: shared-jobs
-              mountPath: /tmp/gobblin-as-service/jobs
+              mountPath: /etc/gobblin-as-service/jobs
             - name: flowconfig-templates
-              mountPath: /tmp/templateCatalog
+              mountPath: /etc/templateCatalog
+          imagePullPolicy: Never # TODO: Remove this once docker images are deployed to Apache DockerHub post-graduation
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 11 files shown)


gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
```diff
@@ -57,7 +59,12 @@ spec:
...
         - name: gobblin-standalone
-          image: will97/gobblin-standalone:latest
+          image: apache/gobblin:latest
+          args: ["--mode", "standalone", "--jvmopts", "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"]
           volumeMounts:
             - name: shared-jobs
               mountPath: /tmp/gobblin-standalone/jobs
+          env:
+            - name: GOBBLIN_JOB_CONFIG_DIR
+              value: /tmp/gobblin-standalone/jobs
```
aplex, line 69: /tmp/ is not a good place to store configuration, as it can get cleaned up at any moment - https://serverfault.com/questions/377348/when-does-tmp-get-cleared

If those configs rarely change while container is running, you can put them in /etc/gobblin . For files that are constantly changing , use /var/... 

See https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard for explanation of linux file system structure

gobblin-kubernetes/gobblin-service/mysql-cluster/mysql-deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
           persistentVolumeClaim:
             claimName: mysql-pv-claim
       containers:
-        - image: mysql:5.6.45
+        - image: mysql:5.7.32
```
aplex, line 33: I think we migrated to 8.x this year

gobblin-kubernetes/gobblin-service/base-cluster/deployment.yaml
Will-Lo: Definitely learned something new today, will make the change. Thanks!

aplex (approved):
LGTM

sv2000 (approved):
+1.

PR #3206, opened 2021-01-20
[GOBBLIN-1365] Adds Github Action workflow for tests

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1365

### Description
- Travis CI is extremely slow and flaky recently, this PR aims to create a working PoC on Github Actions and disable the flaky tests for now

### Tests

### Commits

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -1,0 +1,164 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Build and Run Tests
+
+on:
+  push:
+    # Publish only on `master`
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  release:
+    types: [published, edited]
+
+jobs:
+  build:
+    name: Build repository
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      # Stores external dependencies, can be further improved with Gradle 6.1
+      - name: Cache Gradle Dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          # Only rebuild cache if build.gradle is changed
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build repository
+        run: |
+          ./gradlew --no-daemon clean build -x test -x javadoc -x findbugsMain -x findbugsTest -x checkstyleMain -x checkstyleJmh -x checkstyleTest -Dorg.gradle.parallel=true
+
+  test_coverage:
+    runs-on: ubuntu-latest
+    name: Generate test coverage
+    needs: build
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Cache Gradle Dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          # Only rebuild cache if build.gradle is changed
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Generate code coverage
+        run: |
+          ./gradlew -PskipTestGroup=disabledOnCI -Dorg.gradle.parallel=false -DjacocoBuild=1 $GOBBLIN_GRADLE_OPTS jacocoTestCoverage
+
+  static_checks:
+    name: Run static checks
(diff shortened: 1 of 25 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,162 @@
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          # Only rebuild cache if build.gradle is changed
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build repository
+        run: |
+          ./gradlew --no-daemon clean build -x test -x javadoc -x findbugsMain -x findbugsTest -x checkstyleMain -x checkstyleJmh -x checkstyleTest -Dorg.gradle.parallel=true
+
+  test_coverage:
```
autumnust, line 56: Other than run_tests, should other job also need to specify `needs:build` ?

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,162 @@
...
+        # Write retry logic as integration tests can fail due to timing out or network problems
+        run: |
+          ./gradlew getGroupedTests -PgroupName="${{matrix.test-group}}" > temp.txt
+          TASKS=$(sed -n 's/CI Task: //p' temp.txt)
+          echo $TASKS
+
+          n=0
+          until [ "$n" -ge 3 ]
+          do
+            ./gradlew -PskipTestGroup=disabledOnCI $GOBBLIN_GRADLE_OPTS $TASKS -Dorg.gradle.parallel=false && break
```
autumnust, line 154: Just to my understanding: we don't have to examine the gradle task status code here ? and if there's failure it just skip `break` when get into retry logic ?

autumnust (changes requested):
A couple of comments/questions.

.github/workflows/build_and_test.yaml
Will-Lo: Yup, the gradle task automatically returns exit code 1 here and aborts. So if it passes it'll call `break`, otherwise it'll be evaluated to false and keep looping until it attempts 3 times.

.github/workflows/build_and_test.yaml
Will-Lo: So interestingly enough none of the builds need to be dependent on each other, because gradle should be building the files needed if it hasn't been done so yet on the runner. But I can make the others dependent on `build` too, since I think it would make more logical sense

gobblin-metastore/src/test/java/org/apache/gobblin/metastore/testing/TestMetastoreDatabaseServer.java
```diff
@@ -110,7 +111,7 @@ static Config getDefaultConfig() {
            .put(EMBEDDED_MYSQL_ENABLED_FULL_KEY, true)
            .put(DBUSER_NAME_FULL_KEY, "testUser")
            .put(DBUSER_PASSWORD_FULL_KEY, "testPassword")
-           .put(DBHOST_FULL_KEY, "localhost")
+           .put(DBHOST_FULL_KEY, "127.0.0.1")
```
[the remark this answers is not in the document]
Will-Lo: Ah this was left in when I was trying to debug mysql connections I'll change it back to localhost.

gradle/scripts/testSetup.gradle
```diff
@@ -35,6 +35,7 @@ subprojects {
       }
       testLogging {
         events "started","skipped","passed","failed"
+        exceptionFormat "full"
```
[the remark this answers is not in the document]
Will-Lo: There isn't as far as I know, i needed to add this to gain some insight on failed tests since I don't think there's a way to access the gradle test html after tests.

.github/workflows/build_and_test.yaml
aplex: If every job checks out a clean repo, then the project will get rebuilt every time. This can increase the time of CI pipeline, as things will get rebuilt 3 times. This can also introduce surprising bugs. For example, if we use a wildcard dependency, the build job can pull one version, and test job can pull a newer one.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build repository
+        run: |
+          ./gradlew --no-daemon clean build -x test -x javadoc -x findbugsMain -x findbugsTest -x checkstyleMain -x checkstyleJmh -x checkstyleTest -Dorg.gradle.parallel=true
+
+  test_coverage:
+    runs-on: ubuntu-latest
+    name: Generate test coverage
+    needs: build
+    steps:
```
aplex, line 60: What is the benefit of having multiple jobs, compared to having multiple steps (build/test/coverage) in a single job?
I see that those jobs have a lot of repeated configs, and people might update it in one place, and forget to update it in other jobs.

Will-Lo: Jobs have the ability to be run in parallel, and steps do not. Since the test coverage and static checks are independent of the unit tests, it would be better to separate them from the normal build process to speed up the pipeline.

I agree that the configs are repeated a lot. It is possible to write composite steps in Github actions so that we can reuse them in multiple jobs though. https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action. I can implement it in this PR or put it in a different PR, depending on how high you want to prioritize it.

.github/workflows/build_and_test.yaml
Will-Lo: Yup that's a good point, I made the other jobs dependent on the first build job so now it should only download dependencies once. The dependencies should be stored by the gradle cache after `build` which is shared across jobs, but if it's wildcard then gradle might force a dependency download/check. I think in this scenario if it causes a problem it's acceptable as the later steps would have the newer dependency, and if it causes the tests to break then we should really be building against an exact version.

.github/workflows/build_and_test.yaml
Will-Lo: Responding to your first comment,

Yup that's a good point, I made the other jobs dependent on the first build job so now it should only download dependencies once. The dependencies should be stored by the gradle cache after `build` which is shared across jobs, but if it's wildcard then gradle might force a dependency download/check. I think in this scenario if it causes a problem it's acceptable as the later steps would have the newer dependency, and if it causes the tests to break then we should really be building against an exact version.

aplex: ok, we can have separate jobs for now, and later you can refactor them to use composite steps that hold common parts.

autumnust (approved):

PR #3211, opened 2021-01-21
[GOBBLIN-1370] Switch flow resume to be a restli action

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1370

### Description

Flow resume was previously represented as a partial update, which makes it a pretty confusing API. But restli actually supports an API for actions, which is a much simpler and more obvious way to represent a triggered action like resuming a flow.

With this method, the url to request resuming a flow looks like `https://localhost:6956/gobblinservice/flowexecutions/(flowGroup:testGroup1,flowName:testFlow1,flowExecutionId:1611270235422)?action=resume`

Also fixed a bug that all jobs were being set to `PENDING_RESUME` instead of only the failed/cancelled ones in `DagManager`.

### Tests
Tested locally

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/GobblinServiceFlowExecutionResourceHandler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/GobblinServiceFlowExecutionResourceHandler.java
@@ -71,7 +71,7 @@ public class GobblinServiceFlowExecutionResourceHandler implements FlowExecution
   }
 
   @Override
-  public UpdateResponse resume(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
+  public void resume(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
     String flowGroup = key.getKey().getFlowGroup();
     String flowName = key.getKey().getFlowName();
     Long flowExecutionId = key.getKey().getFlowExecutionId();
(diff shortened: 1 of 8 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/GobblinServiceFlowExecutionResourceHandler.java
```diff
@@ -71,15 +71,14 @@ public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
   }
 
   @Override
-  public UpdateResponse resume(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
+  public void resume(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
```
aplex, line 74: What's the reason to use ComplexResourceKey<FlowStatusId, EmptyRecord> instead of FlowStatusId directly?

jack-moseley: Not really a reason in particular, restli just gives us the ID as a `ComplexResourceKey` and at some point we have to call getKey to get the `FlowStatusId ` from it.

aplex: ok. I see that it's the same in other methods here. Let's keep it consistent then.

aplex (approved):
LGTM

htran1 (approved):
+1

PR #3213, opened 2021-01-26
[GOBBLIN-1372] Generalization of GobblinClusterUtils#setSystemProperties

autumnust:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1372

### Description
This PR mainly is for refactoring the method `GobblinClusterUtils#setSystemProperties`. The problem was: There are some unnecessary details being exposed: `JAVA_TMP_DIR_KEY` and resolution of `YARN_CACHE` are all very specific to Gobblin-on-YARN applications. These things shouldn't be set through if-branch but through configurations only. 

After refactoring: 
There's no specific key/value matching within `setSystemProperties` method, and we don't lose the capability for limiting resolution of `YARN_CACHE` only for a specific set of configuration with being backward-compatible. 

Also, the refactor also let the program to set additional configuration with value specified as `YARN_CACHE` using the following pairs: 
```
gobblin.cluster.systemPropertiesList.YARN_CACHE=grape.root
gobblin.cluster.sysProps.grape.root=YARN_CACHE
```

If trying to achieve the same purpose as `-Dgrape.root=YARN_CACHE`

Also, it adds a bunch of entries in the .gitignore file. 

### Tests
Unit tested with resolution cases. 

### Commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,13 @@ ligradle/checkstyle/suppressions.xml
 gobblin-core/src/test/resources/serde/output-staging/
 gobblin-integration-test-log-dir/
 gobblin-modules/gobblin-elasticsearch/test-elasticsearch/
+gobblin-modules/*/bin/
+gobblin-metrics-libs/*/bin/
+gobblin-config-management/*/bin/
+gobblin-cluster/GobblinHelixJobLauncherTest/
+gobblin-restli/gobblin-restli-utils/bin/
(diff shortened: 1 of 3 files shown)


autumnust:
@sv2000  @Will-Lo  Please help take a look, thanks

Will-Lo (approved):

sv2000 (approved):
+1. LGTM.

PR #3214, opened 2021-01-26
[GOBBLIN-1359] Fix the error that Exception thrown by writer be swallowed by Reactivex and cause jobs to hang

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1359

### Description

After the first change https://github.com/apache/incubator-gobblin/pull/3200, we still see the job hang if onNext throw error after a long processing time (i.e. 5 mins). 
And after investigating, we see the ConnectableFlowable.connect() which is supposed to return immediately does not return even the flow is finished and canceled. And the root cause is that in this case where consumer is super slow, extractor is sleeping on lastFlushAckable.waitForAck(), but the control message will never be processed, so the extractor just hang there.
So the solution will be 1. Make the call of connect in another thread, and let the sequence code get executed which is to periodically check whether the job finished. This is to make job respect fork timeout
2. In flow.doFinal(), call the shutdown again to make sure shutdown can be processed even extractor is sleep in readRecord. And in shutdown process, manually nack the Ackable to wake up the thread.  

### Tests

Run end to end test on Azkaban: 
1. For normal case, record get processed and write out correctly, also verify that the timeout for fork can take effect
2. For bad case where write is slow, we simulate the situation and make sure the exception thrown by writer can be catch and fail the job. Also I use jstack to make sure we do not have any thread hanging in this case (after the task restart)
3. For bad case where writer throw exception right away, make sure the exception will fail the task and cause task to retry

### Commits

### DIFF
--- a/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
+++ b/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
@@ -40,7 +40,6 @@ import org.apache.gobblin.commit.CommitStep;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.metrics.MetricContextUtils;
 import org.apache.gobblin.publisher.DataPublisher;
-import org.apache.gobblin.runtime.JobShutdownException;
 import org.apache.gobblin.runtime.StateStoreBasedWatermarkStorage;
 import org.apache.gobblin.source.extractor.CheckpointableWatermark;
 import org.apache.gobblin.source.extractor.DataRecordException;
--- a/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
+++ b/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
@@ -40,6 +40,7 @@ import org.apache.gobblin.commit.CommitStep;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.metrics.MetricContextUtils;
 import org.apache.gobblin.publisher.DataPublisher;
+import org.apache.gobblin.runtime.JobShutdownException;
 import org.apache.gobblin.runtime.StateStoreBasedWatermarkStorage;
 import org.apache.gobblin.source.extractor.CheckpointableWatermark;
 import org.apache.gobblin.source.extractor.DataRecordException;
@@ -108,8 +109,8 @@ public abstract class FlushingExtractor<S, D> extends EventBasedExtractor<S, D> 
   protected Long flushIntervalMillis;
 
   protected Long timeOfLastFlush = System.currentTimeMillis();
-  protected FlushAckable lastFlushAckable;
-  protected boolean hasOutstandingFlush = false;
+  private FlushAckable lastFlushAckable;
+  private boolean hasOutstandingFlush = false;
   private Optional<DataPublisher> flushPublisher = Optional.absent();
   protected WorkUnitState workUnitState;
 
@@ -298,6 +299,14 @@ public abstract class FlushingExtractor<S, D> extends EventBasedExtractor<S, D> 
     this.flushPublisher.get().publish(Collections.singletonList(workUnitState));
   }
 
+  @Override
+  public void shutdown() {
+    // In case hasOutstandingFlush, we need to manually nack the ackable to make sure the CountDownLatch not hang
+    if (this.hasOutstandingFlush) {
+      this.lastFlushAckable.nack(new IOException("Extractor already shutdown"));
+    }
+  }
+
   /**
    * Persist the watermarks in {@link WatermarkTracker#unacknowledgedWatermarks(Map)} to {@link WatermarkStorage}.
    * The method is called when after a {@link FlushControlMessage} has been acknowledged. To make retrieval of
@@ -332,7 +341,7 @@ public abstract class FlushingExtractor<S, D> extends EventBasedExtractor<S, D> 
   /**
    * {@link Ackable} for waiting for the flush control message to be processed
    */
-  protected static class FlushAckable implements Ackable {
+  private static class FlushAckable implements Ackable {
     private Throwable error;
     private final CountDownLatch processed;
 
--- a/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
+++ b/gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
@@ -108,8 +108,8 @@ public abstract class FlushingExtractor<S, D> extends EventBasedExtractor<S, D> 
   protected Long flushIntervalMillis;
 
   protected Long timeOfLastFlush = System.currentTimeMillis();
-  private FlushAckable lastFlushAckable;
-  private boolean hasOutstandingFlush = false;
+  protected FlushAckable lastFlushAckable;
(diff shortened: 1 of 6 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
```diff
@@ -17,6 +17,9 @@
 
 package org.apache.gobblin.source.extractor.extract.kafka;
 
+import io.reactivex.Emitter;
```
sv2000, line 20: There seem to be many unused imports.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
```diff
@@ -86,7 +87,10 @@ protected void run() throws Exception {
 
     // The cancel is not propagated to the extractor's record generator when it has been turned into a hot Flowable
     // by publish, so set the shutdownRequested flag on cancel to stop the extractor
-    Flowable streamWithShutdownOnCancel = connectableStream.doOnCancel(() -> this.shutdownRequested.set(true));
+    Flowable streamWithShutdownOnCancel = connectableStream.doOnCancel(() -> {
+      this.shutdownRequested.set(true);
+      this.extractor.shutdown();
```
sv2000, line 92: Shouldn't the extractor shutdown be called as a result of setting the shutdownRequestedFlag?

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -229,7 +229,7 @@ public void consumeRecordStream(RecordStreamWithMetadata<D, S> stream)
         }
       }, e -> {
           logger.error("Failed to process record.", e);
-          throw(new RuntimeException(e));
+          verifyAndSetForkState(ForkState.RUNNING, ForkState.FAILED);
```
sv2000, line 232: Does the writer need to be closed or will it be taken care of as part of shutdown sequence?

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
ZihanLi58: only when we request a new record in publisher,  the extractor will shutdown. But in some edge case, i.e. consumer is really slow and we have meet the buffer and not requests record anymore. So I add this to force/make sure executor.shutdown() is called.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
ZihanLi58: writer close is handled in the doFinal() part. It will be executed when the subscription is finished either by error or complete.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
```diff
@@ -139,6 +129,10 @@ public void shutdown() {
       log.error("Interrupted when attempting to shutdown metrics collection threads.");
     }
     this.shutdownRequested.set(true);
+    // In case hasOutstandingFlush, we need to manually nack the ackable to make sure the CountDownLatch not hang
```
sv2000, line 132: Can we move this block to the parent class inside its own shutdown() method and call super.shutdown() here?

gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
```diff
@@ -108,8 +108,8 @@
   protected Long flushIntervalMillis;
 
   protected Long timeOfLastFlush = System.currentTimeMillis();
-  private FlushAckable lastFlushAckable;
-  private boolean hasOutstandingFlush = false;
+  protected FlushAckable lastFlushAckable;
```
sv2000, line 111: See the comment below about creating a shutdown() method in FlushingExtractor class.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
```diff
@@ -89,6 +89,7 @@ protected void run() throws Exception {
     // by publish, so set the shutdownRequested flag on cancel to stop the extractor
     Flowable streamWithShutdownOnCancel = connectableStream.doOnCancel(() -> {
       this.shutdownRequested.set(true);
+      // In the case that extractor sucks in reading record when cancel get called, we call shutdown again to force it
```
sv2000, line 92: sucks -> stuck.

sv2000 (changes requested):

gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
```diff
@@ -40,6 +40,7 @@
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.metrics.MetricContextUtils;
 import org.apache.gobblin.publisher.DataPublisher;
+import org.apache.gobblin.runtime.JobShutdownException;
```
sv2000, line 43: Unused import.

sv2000 (changes requested):

sv2000 (approved):
+1. LGTM.

gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
```diff
@@ -298,6 +298,14 @@ protected void publishTaskOutput() throws IOException {
     this.flushPublisher.get().publish(Collections.singletonList(workUnitState));
   }
 
+  @Override
+  public void shutdown() {
```
autumnust, line 302: What's the reason super.shutdown is not called, is this intentional ?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
autumnust: I don't quite remember how was the failure propagated when the slow write happens, but feels like the right fix here is to add finally block in the place where `extractor.shutdown()` is called in `Task` (e.g. line 514) after catching failure, instead of forcing extractor closure here.

autumnust (changes requested):

gobblin-core-base/src/main/java/org/apache/gobblin/source/extractor/extract/FlushingExtractor.java
ZihanLi58: Yeah, it's a newly added method which will be called in kafkaStreamingExtractor.shutdown(), and the kafkaStreamingExtractor.shutdown() method did not call super.shutdown(), so I did not add it here as well. Also super.shutdown is meaning to check it's a decorator, which is not applied to this case as well.

autumnust (approved):
Sync'ed offline. LGTM

PR #3219, opened 2021-01-30
[GOBBLIN-1379] Fix retries in DataWriter

aplex:
When folder creation fails in FileAwareInputStreamDataWriter with permission error,
operation is retried. However, the original permission error was not logged or
shown anywhere. Instead users saw a misleading error about incorrect writer state.

Since errors can be different on each attempt, we're now logging all of them.

In addition, writer logic is updated to handle retries correctly when error
happens on the early stage.

https://issues.apache.org/jira/browse/GOBBLIN-1379


aplex:
@htran1 , can you take a look?

htran1 (approved):
+1

PR #3220, opened 2021-02-01
[GOBBLIN-1380] Add retention to failed dag state store

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1380

### Description
Add retention (defaulting to 7 days) to failed dag state store, since these dags will never be removed otherwise if flow resume is not called.

### Tests
Tested locally

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -178,7 +178,7 @@ public class DagManager extends AbstractIdleService {
   @Getter
   private final Integer numThreads;
   private final Integer pollingInterval;
-  private final Integer retentionInterval;
+  private final Integer retentionPollingInterval;
   @Getter
   private final JobStatusRetriever jobStatusRetriever;
   private final Config config;
@@ -198,7 +198,7 @@ public class DagManager extends AbstractIdleService {
     this.resumeQueue = initializeDagQueue(this.numThreads);
     this.scheduledExecutorPool = Executors.newScheduledThreadPool(numThreads);
     this.pollingInterval = ConfigUtils.getInt(config, JOB_STATUS_POLLING_INTERVAL_KEY, DEFAULT_JOB_STATUS_POLLING_INTERVAL);
-    this.retentionInterval = ConfigUtils.getInt(config, FAILED_DAG_POLLING_INTERVAL, DEFAULT_FAILED_DAG_POLLING_INTERVAL);
+    this.retentionPollingInterval = ConfigUtils.getInt(config, FAILED_DAG_POLLING_INTERVAL, DEFAULT_FAILED_DAG_POLLING_INTERVAL);
     this.instrumentationEnabled = instrumentationEnabled;
     if (instrumentationEnabled) {
       MetricContext metricContext = Instrumented.getMetricContext(ConfigUtils.configToState(ConfigFactory.empty()), getClass());
@@ -380,7 +380,7 @@ public class DagManager extends AbstractIdleService {
           this.scheduledExecutorPool.scheduleAtFixedRate(dagManagerThread, 0, this.pollingInterval, TimeUnit.SECONDS);
         }
         FailedDagRetentionThread failedDagRetentionThread = new FailedDagRetentionThread(failedDagStateStore, failedDags, failedDagRetentionTime);
-        this.scheduledExecutorPool.scheduleAtFixedRate(failedDagRetentionThread, 0, retentionInterval, TimeUnit.MINUTES);
+        this.scheduledExecutorPool.scheduleAtFixedRate(failedDagRetentionThread, 0, retentionPollingInterval, TimeUnit.MINUTES);
         List<Dag<JobExecutionPlan>> dags = dagStateStore.getDags();
         log.info("Loading " + dags.size() + " dags from dag state store");
         for (Dag<JobExecutionPlan> dag : dags) {
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -1193,7 +1193,7 @@ public class DagManager extends AbstractIdleService {
   }
 
   /**
-   * Thread that runs retention on failed dags based on their original start time (from flow execution ID).
+   * Thread that runs retention on failed dags based on their original start time (which is the flow execution ID).
    */
   public static class FailedDagRetentionThread implements Runnable {
     private final DagStateStore failedDagStateStore;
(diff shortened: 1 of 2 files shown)


jack-moseley:
@arjun4084346 please review

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -1144,6 +1154,15 @@ private void cleanUp() {
       for (String dagId: dagIdstoClean) {
         cleanUpDag(dagId);
       }
+
+      for (Iterator<Map.Entry<String, Dag<JobExecutionPlan>>> iter = this.failedDags.entrySet().iterator(); iter.hasNext();) {
```
arjun4084346, line 1158: This should work. But my only concern is that we are looping over all the failed dags to check if their clean up time has come at the end of every DagManager poll cycle. Should we do this in a separate cleaning thread in DagManager, which runs, say every 1 hour or so?

arjun4084346: The javadoc also says "Because this thread runs in a regular interval, we should avoid doing repetitive work inside it."

jack-moseley: Good point, moved it to a separate thread for retention that runs every 60 mins by default.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -194,6 +198,7 @@ public DagManager(Config config, boolean instrumentationEnabled) {
     this.resumeQueue = initializeDagQueue(this.numThreads);
     this.scheduledExecutorPool = Executors.newScheduledThreadPool(numThreads);
     this.pollingInterval = ConfigUtils.getInt(config, JOB_STATUS_POLLING_INTERVAL_KEY, DEFAULT_JOB_STATUS_POLLING_INTERVAL);
+    this.retentionInterval = ConfigUtils.getInt(config, FAILED_DAG_POLLING_INTERVAL, DEFAULT_FAILED_DAG_POLLING_INTERVAL);
```
arjun4084346, line 201: `retentionInterval` does not look like the best choice given it is actually a polling interval

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -172,12 +178,15 @@ public String toString() {
...
   private final Integer pollingInterval;
+  private final Integer retentionInterval;
   @Getter
   private final JobStatusRetriever jobStatusRetriever;
   private final Config config;
   private final Optional<EventSubmitter> eventSubmitter;
   private final int defaultQuota;
   private final Map<String, Integer> perUserQuota;
+  private final long failedDagRetentionTime;
+  private final Map<String, Dag<JobExecutionPlan>> failedDags = new ConcurrentHashMap<>();
```
arjun4084346, line 189: Why did we start creating this map here instead of inside DagManagerThread?

jack-moseley: The retention thread needs to delete from the same map as well, otherwise the DagManagerThread will think there is a failed dag but not find it in the failed dag state store. The failedDags map is basically a local version of the failed dag state store to avoid querying it often.

arjun4084346:
+1 LGTM

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -1176,6 +1192,39 @@ private synchronized void cleanUpDag(String dagId) {
...
+
+    @Override
+    public void run() {
+      try {
+        log.info("start clean");
+        for (Iterator<Map.Entry<String, Dag<JobExecutionPlan>>> iter = this.failedDags.entrySet().iterator(); iter.hasNext();) {
+          Map.Entry<String, Dag<JobExecutionPlan>> entry = iter.next();
+          if (this.failedDagRetentionTime > 0L
+              && System.currentTimeMillis() > DagManagerUtils.getFlowExecId(entry.getValue()) + this.failedDagRetentionTime) {
+            log.info("cleaning");
```
htran1, line 1217: Will this flood the log? How about keeping a count and only printing one line per execution of run()?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -1176,6 +1192,39 @@ private synchronized void cleanUpDag(String dagId) {
...
+    }
+
+    @Override
+    public void run() {
+      try {
+        log.info("start clean");
+        for (Iterator<Map.Entry<String, Dag<JobExecutionPlan>>> iter = this.failedDags.entrySet().iterator(); iter.hasNext();) {
+          Map.Entry<String, Dag<JobExecutionPlan>> entry = iter.next();
+          if (this.failedDagRetentionTime > 0L
+              && System.currentTimeMillis() > DagManagerUtils.getFlowExecId(entry.getValue()) + this.failedDagRetentionTime) {
```
htran1, line 1216: Can retrieve the time once outside the loop so that everything is cleaned up relative to the same time point.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -1176,6 +1192,39 @@ private synchronized void cleanUpDag(String dagId) {
...
+      this.failedDagStateStore = failedDagStateStore;
+      this.failedDags = failedDags;
+      this.failedDagRetentionTime = failedDagRetentionTime;
+    }
+
+    @Override
+    public void run() {
+      try {
+        log.info("start clean");
+        for (Iterator<Map.Entry<String, Dag<JobExecutionPlan>>> iter = this.failedDags.entrySet().iterator(); iter.hasNext();) {
```
htran1, line 1213: Can use the following for loop construct.
for (Map.Entry<String, Dag<JobExecutionPlan>> entry : this.failedDags.entrySet()) {

htran1 (changes requested):

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: Whoops, I added those logs for testing and meant to remove them.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: I think it is unsafe to remove elements from a collection that you are iterating right? In the for loop I am removing elements from `failedDags`.

I did it this way as per https://stackoverflow.com/questions/223918/iterating-through-a-collection-avoiding-concurrentmodificationexception-when-re, although I realize now I still did it incorrectly, I should be calling `iter.remove()` in the loop

jack-moseley: For simplicity, I just added the keys to a separate list and then removed them from the failed dag map instead of using an iterator.

htran1 (approved):
+1

PR #3228, opened 2021-02-17
[GOBBLIN-1389] Ensuring exception being propagated in Hive-Reg code path

autumnust:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1389

### Description

This small PR just ensures the exception occurred while hive-registration will be propagated as it is so that users/developers would be able to learn the root cause easily. 

### Tests

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
@@ -278,6 +278,7 @@ public class HiveMetaStoreBasedRegister extends HiveRegister {
           log.info(String.format("Created Hive table %s in db %s", tableName, dbName));
           return true;
         } catch (AlreadyExistsException e) {
+          log.debug("Table {}.{} already existed", table.getDbName(), table.getTableName());
         }
       }catch (TException e) {
         log.error(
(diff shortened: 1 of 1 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
```diff
@@ -377,7 +378,7 @@ public Boolean call() throws Exception {
           }
         });
       } catch (ExecutionException ee) {
-        throw new IOException("Database existence checking throwing execution exception.");
+        throw new IOException("Database existence checking throwing execution exception.", ee);
```
aplex, line 381: Do we know the typical reasons for the exception at this place, or it's hard to tell at this layer?
Ideally, exception messages should be in the format of "Problem happened because of A, do B to resolve it".

htran1 (approved):
+1

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/metastore/HiveMetaStoreBasedRegister.java
autumnust: It is hard to tell here. Mostly likely it could be Hive related issue, but maybe too vague to give users useful instructions to fix.

PR #3234, opened 2021-02-25
[GOBBLIN-1396]Enable HiveWriter to consume gmce and register into hive MetadataStore 

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1396

### Description
Add hiveWriter as one subtype of metadataWriter, so that it can consume gmce and register into hive metaStore. The writer should avoid trying to register the same HiveSpec several times also it should be able to handle the schema evolution. 

### Tests
Add unit test for HiveMetadataWriter

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegister.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegister.java
@@ -322,7 +322,7 @@ public abstract class HiveRegister implements Closeable {
     }
   }
 
-  protected boolean needToUpdateTable(HiveTable existingTable, HiveTable newTable) {
+  public boolean needToUpdateTable(HiveTable existingTable, HiveTable newTable) {
     return getTableComparator(existingTable, newTable).compareAll().result();
   }
 
@@ -339,7 +339,7 @@ public abstract class HiveRegister implements Closeable {
     }
   }
 
-  protected boolean needToUpdatePartition(HivePartition existingPartition, HivePartition newPartition) {
+  public boolean needToUpdatePartition(HivePartition existingPartition, HivePartition newPartition) {
     return getPartitionComparator(existingPartition, newPartition).compareAll().result();
   }
 
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
@@ -15,10 +15,9 @@
  * limitations under the License.
  */
 
-package org.apache.gobblin.iceberg.writer;
+package org.apache.gobblin.hive.writer;
 
 import com.google.common.base.Joiner;
-import com.google.common.base.Optional;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Lists;
@@ -49,7 +48,6 @@ import org.apache.gobblin.metrics.kafka.KafkaSchemaRegistry;
 import org.apache.gobblin.metrics.kafka.SchemaRegistryException;
 import org.apache.gobblin.stream.RecordEnvelope;
 import org.apache.gobblin.util.AvroUtils;
-import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.serde2.avro.AvroSerdeUtils;
 
 
@@ -94,9 +92,10 @@ public class HiveMetadataWriter implements MetadataWriter {
   public void flush(String dbName, String tableName) throws IOException {
     String tableKey = tableNameJoiner.join(dbName, tableName);
     log.info("start to flush table: " + tableKey);
-    HashMap<List<String>, ListenableFuture<Void>> executionMap = this.currentExecutionMap.computeIfAbsent(tableKey, s -> new HashMap<>());
+    HashMap<List<String>, ListenableFuture<Void>> executionMap =
+        this.currentExecutionMap.computeIfAbsent(tableKey, s -> new HashMap<>());
     //iterator all execution to get the result to make sure they all succeeded
-    for (HashMap.Entry<List<String>, ListenableFuture<Void>> execution: executionMap.entrySet()) {
+    for (HashMap.Entry<List<String>, ListenableFuture<Void>> execution : executionMap.entrySet()) {
       //todo: add timeout for the execution
       try {
         execution.getValue().get();
@@ -110,18 +109,17 @@ public class HiveMetadataWriter implements MetadataWriter {
 
   public void write(GobblinMetadataChangeEvent gmce, Map<String, Collection<HiveSpec>> newSpecsMap,
       Map<String, Collection<HiveSpec>> oldSpecsMap, HiveSpec tableSpec) throws IOException {
-    long startTime = System.currentTimeMillis();
     String dbName = tableSpec.getTable().getDbName();
     String tableName = tableSpec.getTable().getTableName();
     String tableKey = tableNameJoiner.join(dbName, tableName);
-    if( !specMaps.containsKey(tableKey) || specMaps.get(tableKey).size() == 0) {
+    if (!specMaps.containsKey(tableKey) || specMaps.get(tableKey).size() == 0) {
       //Try to create table first to make sure table existed
       this.hiveRegister.createTableIfNotExists(tableSpec.getTable());
     }
     if (!lastestSchemaMap.containsKey(tableKey)) {
       HiveTable existingTable = this.hiveRegister.getTable(dbName, tableName).get();
-      lastestSchemaMap.put(tableKey, existingTable.getSerDeProps().getProp(
-          AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName()));
+      lastestSchemaMap.put(tableKey,
+          existingTable.getSerDeProps().getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName()));
     }
 
     //Calculate the topic name from gmce, fall back to topic.name in hive spec which can also be null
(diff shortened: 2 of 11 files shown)


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  @Override
+  public void writeEnvelope(RecordEnvelope<GenericRecord> recordEnvelope, Map<String, Collection<HiveSpec>> newSpecsMap,
+      Map<String, Collection<HiveSpec>> oldSpecsMap, HiveSpec tableSpec) throws IOException {
+    GenericRecord genericRecord = recordEnvelope.getRecord();
+    GobblinMetadataChangeEvent gmce =
+        (GobblinMetadataChangeEvent) SpecificData.get().deepCopy(genericRecord.getSchema(), genericRecord);
+    if (whiteistBlacklist.acceptTable(tableSpec.getTable().getDbName(), tableSpec.getTable().getTableName())) {
+
```
autumnust, line 229: empty line

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,242 @@
...
+             //if existedSpec is not null, it means we already registered this partition, so check whether we need to update the table/partition
+             if (gmce.getSchemaSource() == SchemaSource.SCHEMAREGISTRY) {
+               // Only in this case, we need to update the schema to the latest schema in schemaRegistry
+               schemaUpdateHelper(spec, topicName);
+             }
+             if (this.hiveRegister.needToUpdateTable(existedSpec.getTable(), spec.getTable())) {
+                registerSpec(tableKey, partitionValue, spec, hiveSpecCache);
+                return;
+             }
+             if (spec.getPartition().isPresent() &&
```
autumnust, line 169: so if condition in line and the one in line 165 both satisfied, `registerSpec(tableKey, partitionValue, spec, hiveSpecCache);` will be called twice?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,242 @@
...
+    String tableKey = tableNameJoiner.join(dbName, tableName);
+    for (Collection<HiveSpec> specs : newSpecsMap.values()) {
+      for (HiveSpec spec: specs) {
+         if (spec.getTable().getDbName().equals(dbName) && spec.getTable().getTableName().equals(tableName)) {
+           List<String> partitionValue = spec.getPartition().isPresent()? spec.getPartition().get().getValues() : Lists.newArrayList();
+           Cache<List<String>, HiveSpec> hiveSpecCache = specMaps.computeIfAbsent(tableKey, s -> CacheBuilder.newBuilder()
+               .expireAfterAccess(state.getPropAsInt(GobblinMCEWriter.CACHE_EXPIRING_TIME,
+                   GobblinMCEWriter.DEFAULT_CACHE_EXPIRING_TIME), TimeUnit.HOURS)
+               .build());
+           HiveSpec existedSpec = hiveSpecCache.getIfPresent(partitionValue);
```
autumnust, line 158: What if `hiveSpec_A` contains partition`1,2,3`, and `hiveSpec_B` contains partition `2,3,4`, then `2,3` will be registered twice?

autumnust (changes requested):

autumnust:
Seems there's a github action failure above. Can you also check ? Thanks.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegister.java
```diff
@@ -322,7 +322,7 @@ public void addOrAlterPartition(HiveTable table, HivePartition partition)
     }
   }
 
-  protected boolean needToUpdateTable(HiveTable existingTable, HiveTable newTable) {
+  public boolean needToUpdateTable(HiveTable existingTable, HiveTable newTable) {
```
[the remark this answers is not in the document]
ZihanLi58: I move it under gobblin-hive-registration module but a new package writer for better management. So it still needs to be changed to public. Let me know if you have any concerns

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
ZihanLi58: on line 167, it will directly return the method

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
ZihanLi58: Hive spec is for one dir, so it should only contains one partition. PartitionValue is a list here since we support multi partition columns

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriter.java
autumnust: OK I missed that. Then why you need two branches?

autumnust (approved):
LGTM

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  @Override
+  public void writeEnvelope(RecordEnvelope<GenericRecord> recordEnvelope, Map<String, Collection<HiveSpec>> newSpecsMap,
+      Map<String, Collection<HiveSpec>> oldSpecsMap, HiveSpec tableSpec) throws IOException {
+    GenericRecord genericRecord = recordEnvelope.getRecord();
+    GobblinMetadataChangeEvent gmce =
+        (GobblinMetadataChangeEvent) SpecificData.get().deepCopy(genericRecord.getSchema(), genericRecord);
+    if (whiteistBlacklist.acceptTable(tableSpec.getTable().getDbName(), tableSpec.getTable().getTableName())) {
+
```
autumnust, line 275: Additional blank line

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,287 @@
...
+    hiveSpecCache.put(partitionValue, spec);
+  }
+
+  private void schemaUpdateHelper(HiveSpec spec, String topicName, String tableKey) throws SchemaRegistryException {
+    String newSchemaString =
+        spec.getTable().getSerDeProps().getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
+    if (newSchemaString != null) {
+      Schema newSchema = new Schema.Parser().parse(newSchemaString);
+      String newSchemaCreationTime = AvroUtils.getSchemaCreationTime(newSchema);
+      Set<String> existedSchemaCreationTimes = schemaCreationTimeMap.computeIfAbsent(tableKey, s -> new HashSet<>());
```
autumnust, line 242: I wonder why this has to be a set instead of a single value capturing the creation time of newestSchema that was determined before. What's the advantage to keep all ever-appearing schemaCreationTime ?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,287 @@
...
+              spec.getPartition().isPresent() ? spec.getPartition().get().getValues() : Lists.newArrayList();
+          Cache<List<String>, HiveSpec> hiveSpecCache = specMaps.computeIfAbsent(tableKey,
+              s -> CacheBuilder.newBuilder()
+                  .expireAfterAccess(state.getPropAsInt(MetadataWriter.CACHE_EXPIRING_TIME,
+                      MetadataWriter.DEFAULT_CACHE_EXPIRING_TIME), TimeUnit.HOURS)
+                  .build());
+          HiveSpec existedSpec = hiveSpecCache.getIfPresent(partitionValue);
+          if (existedSpec != null) {
+            //if existedSpec is not null, it means we already registered this partition, so check whether we need to update the table/partition
+            if (gmce.getSchemaSource() == SchemaSource.SCHEMAREGISTRY) {
```
autumnust, line 203: This block can be made more readable if we wrap the condition `if (gmce.getSchemaSource() == SchemaSource.SCHEMAREGISTRY)` as part of `schemaUpdateHelper`. So it becomes :
```
// Call it regardless. 
schemaUpdateHelper() 
if (existedSpec == null || tableOrPartitionNeedToBeUpdated() ) { 
  registerSpec() 
}
```

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,287 @@
...
+    for (Collection<HiveSpec> specs : oldSpecsMap.values()) {
+      for (HiveSpec spec : specs) {
+        if (spec.getTable().getDbName().equals(dbName) && spec.getTable().getTableName().equals(tableName)) {
+             if (spec.getPartition().isPresent()) {
+              deRegisterPartitionHelper(dbName, tableName, spec);
+             }
+        }
+      }
+    }
+    //TODO: De-register table if table location does not exist
```
autumnust, line 178: I would rather not to add it or at least make it configurable given the chance to accidentally dropping partitions. If possible this house-keeping responsibility should be done in a different job like retention.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+      }
+    }
+  }
+
+  private void registerSpec(String tableKey, List<String> partitionValue, HiveSpec spec,
+      Cache<List<String>, HiveSpec> hiveSpecCache) {
+    HashMap<List<String>, ListenableFuture<Void>> executionMap =
+        this.currentExecutionMap.computeIfAbsent(tableKey, s -> new HashMap<>());
+    if (executionMap.containsKey(partitionValue)) {
```
autumnust, line 224: I don't quite understand this part. Say we have an old `spec` to register ( meaning the executionMap contains it). We wait on the call to finish by calling future.get() in line 226, and then call it again in line 232 ?  

Also, when we find a branch new spec, we simply put the future object in the executionMap but why we want to ensure it finished by calling future.get() next time we see it ? Isn't this barrier being put in the flush method ?

autumnust (changes requested):

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: Since when schema change, we cannot make sure the gmce with latest always and gmce with old schema are coming in order, there are high chances that we see gmce with old schema and latest schema are interspersed with each other. In these case, we don't want to  query schemaRegistry every time we see a different schema creation time, so cache the creationTime we have seen. I will change Set into Cache to avoid memory accumulating

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: Will fix

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: I'll make it configurable to provide the function to user and by default it will be false

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: When we call registerSpec, it means we see a brand new spec: 
1. It's a new partition and we haven't see it before, in this case, we will skip 226 and direct register in 232. 
2. We have register this partition before and we see a updated spec for this partition (i.e. schema change), in this case, we want to make sure old one and the new one are registered in order so that there won't be situation that we try to register old partition spec A and it's pending in the queue, and then we submit another task to register new spec B, but during execution, B goes first and finally the partition in metaStore is A(the old one)
Since there should not be too many update in same partition during a flush interval, so this should not affect performance so much

autumnust: I see. It makes more sense, thanks !

autumnust (approved):
LGTM

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
autumnust: Thanks, this also helps to understand the comment around `HiveMetadataWriter.java # 223`

PR #3238, opened 2021-03-03
[GOBBLIN-1402] Allow flow's requester list/owner to be updated

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1402

### Description
Make it possible to update the owner/requester list in a FlowConfig in some circumstances, and also refactoring/fixing some other related issues.

- In `FlowConfigResourceLocalHandler`, don't overwrite the previous requester list property (to allow it to be updated if it was specified by a user)
- Move whitelisted requester checking to an API in `RequesterService` to allow it to be used by handlers
- Add some checks to deprecated "flowconfigs" endpoint, since it was possible to set owning group to a group you were not part of through that endpoint
- Add checks as well to "flowconfigsV2" endpoint to avoid a similar way of updating the owning group, and to include the logic for when requester list is allowed to be updated

### Tests
Updated unit tests and tested locally

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
@@ -55,6 +55,7 @@ import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.restli.EmbeddedRestliServer;
 import org.apache.gobblin.runtime.spec_catalog.FlowCatalog;
 import org.apache.gobblin.runtime.spec_store.FSSpecStore;
+import org.apache.gobblin.util.request_allocation.Request;
 
 
 @Test(groups = { "gobblin.service" }, singleThreaded = true)
@@ -77,6 +78,7 @@ public class FlowConfigV2Test {
   private static final String TEST_FLOW_NAME_7 = "testFlow7";
   private static final String TEST_FLOW_NAME_8 = "testFlow8";
   private static final String TEST_FLOW_NAME_9 = "testFlow9";
+  private static final String TEST_FLOW_NAME_10 = "testFlow10";
   private static final String TEST_SCHEDULE = "0 1/0 * ? * *";
   private static final String TEST_TEMPLATE_URI = "FS:///templates/test.template";
 
@@ -263,7 +265,7 @@ public class FlowConfigV2Test {
     }
   }
 
-  @Test
+  @Test(dependsOnMethods = {"testGroupRequesterRejected", "testGroupRequesterAllowed", "testGroupUpdateRejected", "testRequesterUpdate"})
   public void testLocalGroupOwnershipUpdates() throws Exception {
     try {
       ServiceRequester testRequester = new ServiceRequester("testName", "USER_PRINCIPAL", "testFrom");
@@ -303,7 +305,7 @@ public class FlowConfigV2Test {
   }
 
 
-  @Test (expectedExceptions = RestLiResponseException.class)
+  @Test
   public void testGroupUpdateRejected() throws Exception {
    ServiceRequester testRequester = new ServiceRequester("testName", "USER_PRINCIPAL", "testFrom");
    _requesterService.setRequester(testRequester);
@@ -316,26 +318,65 @@ public class FlowConfigV2Test {
 
    _client.createFlowConfig(flowConfig);
 
-   flowConfig.setOwningGroup("testGroup2");
-   _client.updateFlowConfig(flowConfig);
+   // Update should be rejected because testName is not part of dummyGroup
+   flowConfig.setOwningGroup("dummyGroup");
+   try {
+     _client.updateFlowConfig(flowConfig);
+     Assert.fail("Expected update to be rejected");
+   } catch (RestLiResponseException e) {
+     Assert.assertEquals(e.getStatus(), HttpStatus.ORDINAL_401_Unauthorized);
+   }
   }
 
-  @Test (expectedExceptions = RestLiResponseException.class)
+  @Test
+  public void testRequesterUpdate() throws Exception {
+    ServiceRequester testRequester = new ServiceRequester("testName", "USER_PRINCIPAL", "testFrom");
+    ServiceRequester testRequester2 = new ServiceRequester("testName2", "USER_PRINCIPAL", "testFrom");
+    _requesterService.setRequester(testRequester);
+    Map<String, String> flowProperties = Maps.newHashMap();
+
+    FlowId flowId = new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_9);
+    FlowConfig flowConfig = new FlowConfig().setId(flowId)
+        .setTemplateUris(TEST_TEMPLATE_URI)
+        .setProperties(new StringMap(flowProperties))
+        .setOwningGroup("testGroup");
+
+    _client.createFlowConfig(flowConfig);
+
+    // testName2 takes ownership of the flow
+    flowProperties.put(RequesterService.REQUESTER_LIST, RequesterService.serialize(Lists.newArrayList(testRequester2)));
+    flowConfig.setProperties(new StringMap(flowProperties));
+    _requesterService.setRequester(testRequester2);
+    _client.updateFlowConfig(flowConfig);
+
+    // Check that the requester list was actually updated
+    FlowConfig updatedFlowConfig = _client.getFlowConfig(flowId);
+    Assert.assertEquals(RequesterService.deserialize(updatedFlowConfig.getProperties().get(RequesterService.REQUESTER_LIST)),
+        Lists.newArrayList(testRequester2));
+  }
+
+  @Test
   public void testRequesterUpdateRejected() throws Exception {
     ServiceRequester testRequester = new ServiceRequester("testName", "USER_PRINCIPAL", "testFrom");
     ServiceRequester testRequester2 = new ServiceRequester("testName2", "USER_PRINCIPAL", "testFrom");
     _requesterService.setRequester(testRequester);
     Map<String, String> flowProperties = Maps.newHashMap();
 
-    FlowConfig flowConfig = new FlowConfig().setId(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_9))
+    FlowConfig flowConfig = new FlowConfig().setId(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_10))
         .setTemplateUris(TEST_TEMPLATE_URI)
         .setProperties(new StringMap(flowProperties));
 
     _client.createFlowConfig(flowConfig);
 
+    // Update should be rejected because testName is not allowed to update the owner to testName2
     flowProperties.put(RequesterService.REQUESTER_LIST, RequesterService.serialize(Lists.newArrayList(testRequester2)));
     flowConfig.setProperties(new StringMap(flowProperties));
-    _client.updateFlowConfig(flowConfig);
+    try {
+      _client.updateFlowConfig(flowConfig);
+      Assert.fail("Expected update to be rejected");
+    } catch (RestLiResponseException e) {
+      Assert.assertEquals(e.getStatus(), HttpStatus.ORDINAL_401_Unauthorized);
+    }
   }
 
   @AfterClass(alwaysRun = true)
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
@@ -18,7 +18,6 @@
 package org.apache.gobblin.service;
 
 import java.io.File;
-import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
@@ -76,6 +75,8 @@ public class FlowConfigV2Test {
   private static final String TEST_FLOW_NAME_5 = "testFlow5";
   private static final String TEST_FLOW_NAME_6 = "testFlow6";
   private static final String TEST_FLOW_NAME_7 = "testFlow7";
+  private static final String TEST_FLOW_NAME_8 = "testFlow8";
+  private static final String TEST_FLOW_NAME_9 = "testFlow9";
   private static final String TEST_SCHEDULE = "0 1/0 * ? * *";
   private static final String TEST_TEMPLATE_URI = "FS:///templates/test.template";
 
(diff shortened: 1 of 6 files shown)


jack-moseley:
@Will-Lo please review

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -301,6 +302,42 @@ public void testLocalGroupOwnershipUpdates() throws Exception {
...
+
+   FlowConfig flowConfig = new FlowConfig().setId(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_8))
+       .setTemplateUris(TEST_TEMPLATE_URI)
+       .setProperties(new StringMap(flowProperties))
+       .setOwningGroup("testGroup");
+
+   _client.createFlowConfig(flowConfig);
+
+   flowConfig.setOwningGroup("testGroup2");
+   _client.updateFlowConfig(flowConfig);
```
aplex, line 320: Why would we get exception here? From what I see, user creates a flow and then sets the owning group for the flow that he owns.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -301,6 +302,42 @@ public void testLocalGroupOwnershipUpdates() throws Exception {
     _client.deleteFlowConfig(new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_7));
   }
 
+
+  @Test (expectedExceptions = RestLiResponseException.class)
```
aplex, line 306: The design doc also mentioned that "A user that is part of the owning group can take control of the flow by updating the owner to themselves". Is this implemented/covered with tests?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -301,6 +302,42 @@ public void testLocalGroupOwnershipUpdates() throws Exception {
...
+       .setProperties(new StringMap(flowProperties))
+       .setOwningGroup("testGroup");
+
+   _client.createFlowConfig(flowConfig);
+
+   flowConfig.setOwningGroup("testGroup2");
+   _client.updateFlowConfig(flowConfig);
+  }
+
+  @Test (expectedExceptions = RestLiResponseException.class)
```
aplex, line 323: Can we also check that exception message or http result is the one that we expect, and it is about authorization/permissions? Currently if there is a typo in the test, and service rejects the request for any reason, the test will pass.

TestNG has expectedExceptionsMessageRegExp: https://howtodoinjava.com/testng/testng-expected-exception/

aplex (changes requested):

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
Will-Lo, line 306: I think it would be a better practice for us to try and catch the exception, and then assert that the error code is as expected. For example, with `testGroupRequesterRejected`
```
try {
...
} catch (RestLiResponseException e) {
  Assert.assertEquals(e.getStatus(), HttpStatus.ORDINAL_401_Unauthorized);
}
```

I was able to catch a bug before in this test suite as it actually returned a different error code than expected. Also these tests would mark 500 errors as passing tests which I think we should avoid.
Also it allows better granularity in the test too, so we can assert that only the second call here would fail

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: The point is to check that a user cannot update a flow's owningGroup to a group that that user is not a part of (the test user is not part of the group `testGroup2`). Previously we only checked this condition on flow creation, but you could still update it to a group you are not part of.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -208,18 +217,59 @@ private Properties getHeaders() {
...
    * {@link HttpStatus#S_400_BAD_REQUEST}.
-   * @param requesterService the {@link RequesterService} used to verify the requester
    * @param originalFlowConfig original flow config to find original requester
    * @param requesterList list of requesters for this request
    */
-  public void checkRequester(
-      RequesterService requesterService, FlowConfig originalFlowConfig, List<ServiceRequester> requesterList) {
-    if (requesterList == null) {
+  public void checkRequester(FlowConfig originalFlowConfig, List<ServiceRequester> requesterList) {
+    if (requesterList == null || this.requesterService.isRequesterWhitelisted(requesterList)) {
```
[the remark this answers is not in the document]
jack-moseley: We are using an internal implementation of `RequesterService` which will never return null. Although, it looks like even the open source ones (`TestRequesterService` and `NoopRequesterService`) never return null either, so the null check could probably be removed.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
Will-Lo: We should avoid using testGroup2 as a failing group. There's another test that writes testGroup2 to the file where the group ownership service reads from, to test if the groupownership service will update automatically on reads, and this could lead to flakiness

Will-Lo (changes requested):

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: Doesn't this have the issue that the test will pass if no exception is thrown? (Unless you do something like `throw new RuntimeException("Shouldn't reach here")` after the point you expect it to fail). Probably better to use the `expectedExceptionsMessageRegExp` mentioned by Alex above.

aplex: Try/catch will also work too. I usually add Assert.fail("Expected exception X") at the end of the try block.

Junit has this exception verification streamlined with assertThrows, but we're still on TestNg - https://stackoverflow.com/questions/40268446/junit-5-how-to-assert-an-exception-is-thrown

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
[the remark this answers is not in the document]
aplex: yep, let's remove it then (if we can't figure out the reason for it to exist from git blame). In security code it's safer to deny access in unknown or ambiguous situations.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -301,6 +302,42 @@ public void testLocalGroupOwnershipUpdates() throws Exception {
...
+  public void testRequesterUpdateRejected() throws Exception {
```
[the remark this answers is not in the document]
jack-moseley: Added a test

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: Yes, it's covered by the test I just added

jack-moseley: Actually I think I prefer it that way, because we can just check the HTTP status code instead, checking error messages seems a bit worse because if we ever want to change the message sent it will affect all the tests.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: Actually it looks like that test also deletes the original group (testGroup1), which was causing some confusing behaviour, and I think could affect the already existing tests as well depending on what order they run in. I made that one depend on all the ones that rely on that group being there.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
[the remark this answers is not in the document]
aplex: This is marked as resolved, but the null check is still there in the second commit.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsResource.java
```diff
@@ -142,7 +155,7 @@ public UpdateResponse delete(ComplexResourceKey<FlowId, EmptyRecord> key) {
    */
   public static void checkRequester(
       RequesterService requesterService, FlowConfig originalFlowConfig, List<ServiceRequester> requesterList) {
-    if (requesterList == null) {
+    if (requesterList == null || requesterService.isRequesterWhitelisted(requesterList)) {
```
aplex, line 158: Same comment about requester==null as for V2

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -208,18 +217,59 @@ private Properties getHeaders() {
...
+   *    it to themselves.
+   */
+  public void checkPropertyUpdatesAllowed(List<ServiceRequester> requesterList, FlowConfig updatedFlowConfig) {
+    if (requesterList == null || this.requesterService.isRequesterWhitelisted(requesterList)) {
+      return;
+    }
+
+    // Check that requester is part of owning group if owning group is being updated
+    if (updatedFlowConfig.hasOwningGroup() && !this.groupOwnershipService.isMemberOfGroup(requesterList, updatedFlowConfig.getOwningGroup())) {
+      throw new FlowConfigLoggedException(HttpStatus.S_401_UNAUTHORIZED, "Requester not part of owning group specified");
```
aplex, line 244: It would be good to add more details and an action part to the error message, so that we get fewer support requests asking us what to do about the error.

For example, we can append it with "User XXX should join the group YYY and try again".

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowConfigsV2Resource.java
```diff
@@ -208,18 +217,59 @@ private Properties getHeaders() {
...
+    }
+
+    if (updatedFlowConfig.hasProperties() && updatedFlowConfig.getProperties().containsKey(RequesterService.REQUESTER_LIST)) {
+      List<ServiceRequester> updatedRequesterList;
+      try {
+        updatedRequesterList = RequesterService.deserialize(updatedFlowConfig.getProperties().get(RequesterService.REQUESTER_LIST));
+      } catch (Exception e) {
+        throw new FlowConfigLoggedException(HttpStatus.S_400_BAD_REQUEST, RequesterService.REQUESTER_LIST + " property was "
+            + "provided but could not be deserialized", e);
```
aplex, line 253: It would be good to add an example of the value of this property to the error message, or add link to the doc showing how it should look like. 

One way to do this is to just serialize the sample user name and append json to the message.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -262,7 +265,7 @@ public void testGroupRequesterRejected() throws Exception {
     }
   }
 
-  @Test
+  @Test(dependsOnMethods = {"testGroupRequesterRejected", "testGroupRequesterAllowed", "testGroupUpdateRejected", "testRequesterUpdate"})
```
aplex, line 268: Do we have a way to run this without dependencies? Test dependencies bring a couple of problems:

- We can't run all tests in parallel, which affects build time
- To understand what goes on in the test, developer also need to read all dependent tests
- People that add new tests will likely be unaware about dependencies, and will just add their test to the bottom. This can break old tests in unexpected ways.

Usually, it's good to have tests isolated from each other. If you need a common setup logic, it can be moved to a separate method and called from all needed tests.

aplex (changes requested):

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: I couldn't think of a good way since it is modifying files that were used by other tests in the same class (this could have been a problem even before this PR).

I decided to refactor and move this test to a separate class. I'm not sure why it was even in this class in the first place, since it seems to be testing that `LocalGroupOwnershipService` is able to pick up changes to the file, which is unrelated to flowConfig client. (And in this class there is already tests that group ownership is working while calling flowConfig client).

aplex (approved):

Will-Lo (approved):

sv2000 (approved):
+1.

PR #3243, opened 2021-03-11
GOBBLIN-1409 Expose Kafka record timestamp

milimetric:
Dear Gobblin maintainers,

This is my first PR, please do point out any mistakes or omissions.

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1409

### Description
When consuming from Kafka, we need the kafka timestamp.  We weren't sure if there was another way to expose this, but figured it's a general enough need that it should be implemented.  And it didn't affect performance as far as we could see.

### Tests
    - I added a basic test, but I'm really not sure about it, I couldn't replicate the testing environment.  Some advice on that would be greatly appreciated.


ZihanLi58 (commented):
Can you check that the test does not pass?
ConsumerRecord allow different timestamp types: 
`NO_TIMESTAMP_TYPE(-1, "NoTimestampType"),CREATE_TIME(0, "CreateTime"), LOG_APPEND_TIME(1, "LogAppendTime");`
You may want another method to determine the timestamp type or overwrite the method  isTimestampLogAppend()

milimetric:
Thanks so much, I think that makes sense.  So now users of kafka-1 can check what type of timestamp they're going to expect and get it.  I guess the test setup consumes records with `NO_TIMESTAMP_TYPE`, but I couldn't see how that was done or how to change it.  Any guidance is appreciated.

PR #3250, opened 2021-03-19
GOBBLIN-1416: Fix a race condition caused by Helix task cancellation being invoked before Gobblin task creation

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1416

### Description
Invocation of Gobblin Helix task cancellation before underlying Gobblin tasks have been created or submitted to the task executor results in GobblinHelixTask#cancel() returning successfully, but without preventing future submission of Gobblin tasks to the executor. 

### Tests
Added unit test.

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
@@ -161,8 +161,10 @@ public class GobblinMultiTaskAttempt {
     Pair<List<Task>, Boolean> executionResult = runWorkUnits(countDownLatch);
     this.tasks = executionResult.getFirst();
 
+    // The task attempt has already been stopped and the task list is empty. This indicates that a cancel has been
+    // invoked prior to creation of underlying Gobblin tasks. In a normal scenario, where a cancel is invoked after
+    // successful task creation, the task list is guaranteed to be non-empty and we shouldn't enter the following block.
     if (this.tasks.isEmpty() && this.stopped.get()) {
-      //The task attempt has already been stopped.
       return;
     }
 
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
@@ -29,6 +29,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


sv2000:
@autumnust @ZihanLi58 Please review.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -384,10 +394,14 @@ private boolean taskSuccessfulInPriorAttempt(String taskId) {
...
    */
-  private Pair<List<Task>, Boolean> runWorkUnits(CountUpAndDownLatch countDownLatch) {
-
+  private synchronized Pair<List<Task>, Boolean> runWorkUnits(CountUpAndDownLatch countDownLatch) {
     List<Task> tasks = Lists.newArrayList();
-
+    //Has the task-attempt already been cancelled? This can happen for instance when a cancellation has been invoked on
+    // the GobblinMultiTaskAttempt instance (e.g. in the case of Helix task cancellation) before the Gobblin tasks
+    // have been submitted to the underlying task executor.
+    if (this.stopped.get()) {
```
autumnust, line 402: It doesn't seem to be enough for synchronization purpose. cancel could happen after this line and between line 407 and 408.

autumnust (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
sv2000: Note that shutdownTasks() is synchronized too. So if cancel has not been called yet (i.e. at line 402), the task creation is guaranteed to be completed and tasks will be submitted to the executor, before shutdownTasks() gets called.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/GobblinMultiTaskAttempt.java
```diff
@@ -157,6 +161,11 @@ public void run()
     Pair<List<Task>, Boolean> executionResult = runWorkUnits(countDownLatch);
     this.tasks = executionResult.getFirst();
 
+    if (this.tasks.isEmpty() && this.stopped.get()) {
```
autumnust, line 164: Let me know if this understanding is correct: whenever these two conditions hold at the same time, it indicates that the cancel happens before the task is materialized and executed in the executor. ( since the regular cancel will result in tasks being non empty). Can you add a comment if this is the case, just to distinguish between the regular cancel.

autumnust (changes requested):

autumnust (approved):
+1

PR #3252, opened 2021-03-24
[GOBBLIN-1413]-Emit-GMCE-as-long-as-watermark-moved

hanghangliu:
Dear Gobblin maintainers,

### JIRA
- [ ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-1413] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-1413

### Description
Emit GMCE(Gobblin Metadata Change Event) as long as watermark moved on streaming pipeline. 

Currently the GMCE won't be triggered within streaming pipeline if no new file being generated. This causes problem if watermarks moved, while no file being generated(for example, data been filtered out by quality checker), GMCE will be missed.

Change GMCE publisher to produce GMCE when no file generated. Change IcebergMetadataWriter to correctly update watermark. 

### Tests

### Commits

### DIFF
--- a/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/GobblinMCEProducer.java
+++ b/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/GobblinMCEProducer.java
@@ -198,6 +198,13 @@ public abstract class GobblinMCEProducer implements Closeable {
         }
         break;
       }
+      case change_property: {
+        if(oldFiles != null) {
+          log.warn("{} old files detected while no file alteration is performed", oldFiles.size());
+        }
+        log.info("Setting GMCE while no file changes need to be performed.");
+        break;
+      }
       default: {
         //unsupported operation
         log.error("Unsupported operation type {}", operationType);
@@ -209,18 +216,37 @@ public abstract class GobblinMCEProducer implements Closeable {
 
 
   private List<DataFile> toGobblinDataFileList(Map<Path, Metrics> files) {
-    return Lists.newArrayList(Iterables.transform(files.entrySet(), file -> DataFile.newBuilder()
-        .setFilePath(file.getKey().toString())
-        .setFileFormat(IcebergUtils.getIcebergFormat(state).toString())
-        .setFileMetrics(DataMetrics.newBuilder()
-            .setRecordCount(file.getValue().recordCount())
-            .setColumnSizes(getIntegerLongPairsFromMap(file.getValue().columnSizes()))
-            .setValueCounts(getIntegerLongPairsFromMap(file.getValue().valueCounts()))
-            .setNullValueCounts(getIntegerLongPairsFromMap(file.getValue().nullValueCounts()))
-            .setLowerBounds(getIntegerBytesPairsFromMap(file.getValue().lowerBounds()))
-            .setUpperBounds(getIntegerBytesPairsFromMap(file.getValue().upperBounds()))
-            .build())
-        .build()));
+    return Lists.newArrayList(Iterables.transform(files.entrySet(), file ->
+        {
+          DataFile.Builder builder = createBuilderWithFilePath(file.getKey());
+          addMetricsToFileBuilder(builder, file.getValue());
+          return builder.build();
+        }
+    ));
+  }
+
+  private DataFile.Builder createBuilderWithFilePath(Path filePath) {
+    return DataFile.newBuilder()
+        .setFilePath(filePath.toString())
+        .setFileFormat(IcebergUtils.getIcebergFormat(state).toString());
+  }
+
+  private void addMetricsToFileBuilder(DataFile.Builder builder, Metrics metrics) {
+    // If metrics is null or empty, set FileMetrics a dummy one
+    if(metrics == null || metrics.recordCount() == null) {
+      builder.setFileMetrics(DataMetrics.newBuilder().setRecordCount(0)
+          .build());
+      return;
+    }
+    // If metrics is concrete, fill all fields
+    builder.setFileMetrics(DataMetrics.newBuilder()
+        .setRecordCount(metrics.recordCount())
+        .setColumnSizes(getIntegerLongPairsFromMap(metrics.columnSizes()))
+        .setValueCounts(getIntegerLongPairsFromMap(metrics.valueCounts()))
+        .setNullValueCounts(getIntegerLongPairsFromMap(metrics.nullValueCounts()))
+        .setLowerBounds(getIntegerBytesPairsFromMap(metrics.lowerBounds()))
+        .setUpperBounds(getIntegerBytesPairsFromMap(metrics.upperBounds()))
+        .build());
   }
 
   private List<IntegerLongPair> getIntegerLongPairsFromMap(Map<Integer, Long> map) {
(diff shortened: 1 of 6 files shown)


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -132,6 +136,36 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
     return newFiles;
   }
 
+  /**
+   * Choose one file from the work unit state. There will be no modification to the file.
+   * It's used in GMCE writer {@link GobblinMCEWriter} merely for getting the DB and table name.
+   * @throws IOException
+   */
+  private Map<Path, Metrics> computeDummyFile (State state) throws IOException {
```
ZihanLi58, line 144: This method will end up with list all files under the dataset, and only pick up the oldest one. And the oldest file might be removed by retention pipeline in a short time. Can you modify the algorithm to get the latest available data file?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -229,7 +229,7 @@ private Long getCurrentWaterMark(TableIdentifier tid, String topicPartition) {
    * The logic of this function will be:
    * 1. Check whether a table exists, if not then create the iceberg table
    * 2. Compute schema from the gmce and update the cache for candidate schemas
-   * 3. Do the required operation of the gmce, i.e. addFile, rewriteFile or dropFile
+   * 3. Do the required operation of the gmce, i.e. addFile, rewriteFile or dropFile. change_property means no operation needed
```
ZihanLi58, line 232: change_property means we only update the table level property but not modify data?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/GobblinMCEProducer.java
```diff
@@ -198,6 +198,15 @@ private boolean verifyInput(Map<Path, Metrics> newFiles, List<String> oldFiles,
         }
         break;
       }
+      case change_property: {
+        if(newFiles != null || oldFiles != null) {
+          log.warn("{} new files and {} old files detected while no file alteration is performed",
```
ZihanLi58, line 203: Since we always set new files to a dummy file, do you think we need to modify the condition for the warn log?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -92,9 +93,12 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
       Map<Path, Metrics> newFiles = computeFileMetrics(state);
       Map<String, String> offsetRange = getPartitionOffsetRange(OFFSET_RANGE_KEY);
       if (newFiles.isEmpty()) {
-        return;
+        // There'll be only one dummy file here. This file is parsed for DB and table name calculation.
+        newFiles = computeDummyFile(state);
+        this.producer.sendGMCE(newFiles, null, null, offsetRange, OperationType.change_property, SchemaSource.SCHEMAREGISTRY);
```
ZihanLi58, line 98: You may want to set the schema source to be none in this case? since we don't want to update the schema?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -275,6 +275,14 @@ public void write(GobblinMetadataChangeEvent gmce, Map<String, Collection<HiveSp
         dropFiles(gmce, oldSpecsMap, table, tableMetadata, tid);
         break;
       }
+      case change_property: {
```
ZihanLi58, line 278: On Line 244, we may also want to skip processing the event if the operation type is change_property and table does not exist?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -132,6 +136,36 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
   }
 
+  /**
+   * Choose one file from the work unit state. There will be no modification to the file.
+   * It's used in GMCE writer {@link GobblinMCEWriter} merely for getting the DB and table name.
+   * @throws IOException
+   */
+  private Map<Path, Metrics> computeDummyFile (State state) throws IOException {
+    Map<Path, Metrics> newFiles = new HashMap<>();
+    NameMapping mapping = getNameMapping();
```
ZihanLi58, line 146: In this case, we don't need file metrics as it's just a dummy file. Also get nameMapping require the latest schema is set, but it's possible that we don't have this properties when no data written out.

ZihanLi58 (changes requested):
There is a merge conflict, can you check that?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
hanghangliu: Used a priority queue to chose file based on max modification_time to ensure the dummy file is the latest available file.

ZihanLi58 (approved):
+1 LGTM

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -132,6 +137,36 @@ public void publishData(Collection<? extends WorkUnitState> states) throws IOExc
...
 
+  /**
+   * Choose the latest file from the work unit state. There will be no modification to the file.
+   * It's used in GMCE writer {@link GobblinMCEWriter} merely for getting the DB and table name.
+   * @throws IOException
+   */
+  private Map<Path, Metrics> computeDummyFile (State state) throws IOException {
+    Map<Path, Metrics> newFiles = new HashMap<>();
+    FileSystem fs = FileSystem.get(conf);
+    for (final String pathString : state.getPropAsList(ConfigurationKeys.DATA_PUBLISHER_DATASET_DIR, "")) {
```
ZihanLi58, line 148: Oh one more thing I want to point out that if writer.partition.prefix is set, you may also want to include that value in the initial path, so that you won't end up with get the daily data which may cause the following operation to fail

hanghangliu: As we discussed and tested, this won't be an issue

autumnust (approved):
+1

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -18,11 +18,13 @@
 package org.apache.gobblin.iceberg.publisher;
 
 import com.google.common.io.Closer;
+import java.util.PriorityQueue;
```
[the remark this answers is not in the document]
hanghangliu: turned out to be the old imports were in wrong sequence. reformat the whole file's imports

PR #3254, opened 2021-03-29
[GOBBLIN-1418] Add config to make it possible to specify a default Authenticator for a job

jack-moseley:
… a job

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1418

### Description

Setting the default java Authenticator allows you to run a job and force all proxy/http calls to go through this Authenticator. This is useful for example for a job where going through a proxy (which requires user/password) is desired, but using APIs with underlying http clients that don't support providing user/password. (See [stackoverflow](https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java/43263674) discussion about this).  This PR makes this possible by:

- Adding a class extending Authenticator which provides a user/password from job properties (with decryption)
- Adding an optional key that will set the default Authenticator when the job launches (in AbstractJobLauncher)

### Tests
Successfully ran a job that was not working by providing user/password directly in jvm args

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/password/EncryptedPasswordAuthenticator.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/password/EncryptedPasswordAuthenticator.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.password;
+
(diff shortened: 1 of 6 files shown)


gobblin-api/src/main/java/org/apache/gobblin/password/EncryptedPasswordAuthenticator.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Properties;
+
+import com.google.common.base.Preconditions;
+
+
+/**
+ * {@link Authenticator} that uses a username and password from the provided {@link Properties} to authenticate, and
+ * also decrypts the password using {@link PasswordManager}
+ */
+public class EncryptedPasswordAuthenticator extends Authenticator {
```
sv2000, line 31: Can we add unit tests for this class?

jack-moseley: Is there anything worth testing here? The only actual logic is calling PasswordManager, and based on PasswordManagerTest, it seems like we can't even enable tests that use that due to flakiness.

jack-moseley:
Added this logic to the `MRJobLauncher` as well, since otherwise this would not apply to MR tasks launched by the job

gobblin-api/src/main/java/org/apache/gobblin/password/EncryptedPasswordAuthenticator.java
jack-moseley: Added a disabled test that checks an encrypted password, and an enabled test that just uses a password with no encryption.

sv2000 (approved):
+1. LGTM. For future PRs, please avoid doing a "git push -f", since it is hard to know the changes since the last review.

PR #3255, opened 2021-04-02
[GOBBLIN-1419]Error handling for compaction pipeline on GMCE emitted error

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1419

### Description
No compaction publish file and emit GMCE are happening separately. If file rename succeed but emit GMCE fails, we will end up with not emit GMCE for the compaction. So need  a strategy to do error handling for compaction that when emit GMCE fails, next job will treat the previous compaction as fail and re-compact the data.

### Tests
unit test

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
@@ -167,6 +167,10 @@ public class CompactionCompleteFileOperationAction implements CompactionComplete
         compactionState.setProp(DUPLICATE_COUNT_TOTAL + Long.toString(executionCount),
             compactionState.getProp(DUPLICATE_COUNT_TOTAL, "null"));
       }
+      if(state.getPropAsBoolean(ConfigurationKeys.GOBBLIN_METADATA_CHANGE_EVENT_ENABLED, false)) {
+        //GMCE enabled, set the key to be false to indicate that GMCE has not been sent yet
+        compactionState.setProp(CompactionGMCEPublishingAction.GMCE_EMITTED_KEY, false);
+      }
       compactionState.setProp(CompactionSlaEventHelper.RECORD_COUNT_TOTAL, Long.toString(newTotalRecords));
       compactionState.setProp(CompactionSlaEventHelper.EXEC_COUNT_TOTAL, Long.toString(executionCount + 1));
       compactionState.setProp(CompactionSlaEventHelper.MR_JOB_ID,
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
@@ -48,7 +48,7 @@ import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.mapreduce.Counter;
 import org.apache.hadoop.mapreduce.Job;
 
-import static org.apache.gobblin.compaction.event.CompactionSlaEventHelper.DUPLICATE_COUNT_TOTAL;
+import static org.apache.gobblin.compaction.event.CompactionSlaEventHelper.*;
 
 
 /**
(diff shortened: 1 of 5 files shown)


ZihanLi58:
@autumnust @sv2000 can you help take a look at this change? Thanks!

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionCompleteFileOperationAction.java
```diff
@@ -167,6 +167,10 @@ public void onCompactionJobComplete(FileSystemDataset dataset) throws IOExceptio
         compactionState.setProp(DUPLICATE_COUNT_TOTAL + Long.toString(executionCount),
             compactionState.getProp(DUPLICATE_COUNT_TOTAL, "null"));
       }
+      if(state.getPropAsBoolean(ConfigurationKeys.GOBBLIN_METADATA_CHANGE_EVENT_ENABLED, false)) {
```
autumnust, line 170: please consider using auto-formatting to clean the spaces between preserved keywords

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
```diff
@@ -81,10 +84,15 @@ public Result verify(FileSystemDataset dataset) {
         newRecords = helper.calculateRecordCount(Lists.newArrayList(new Path(dataset.datasetURN())));
       }
       double oldRecords = helper.readRecordCount(new Path(result.getDstAbsoluteDir()));
-
+      State datasetState = helper.loadState(new Path(result.getDstAbsoluteDir()));
       if (oldRecords == 0) {
         return new Result(true, "");
       }
+      if(state.getPropAsBoolean(ConfigurationKeys.GOBBLIN_METADATA_CHANGE_EVENT_ENABLED, false)) {
```
autumnust, line 91: these two branches could be merged.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
```diff
@@ -60,7 +62,8 @@ public CompactionThresholdVerifier(State state) {
    * dataset. To avoid scalability issue, we choose a stateless approach where each dataset tracks
    * record count by themselves and persist it in the file system)
    *
-   * @return true iff the difference exceeds the threshold or this is the first time compaction
+   * @return true if the difference exceeds the threshold or this is the first time compaction or
```
autumnust, line 65: I am not sure if it is a good idea to package the logic of checking gmce-emission inside `thresholder` verifier.  What's the cost of having a separate implementation of verifier ?

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionGMCEPublishingAction.java
```diff
@@ -57,9 +58,11 @@
 
   public static final String ICEBERG_ID_ATTRIBUTE = "iceberg.id";
   public static final String ICEBERG_REQUIRED_ATTRIBUTE = "iceberg.required";
+  public final static String GMCE_EMITTED_KEY = "GMCE.emitted";
   private final State state;
   private final CompactionJobConfigurator configurator;
   private final Configuration conf;
+  private InputRecordCountHelper helper;
```
autumnust, line 65: This class may deserve a better name given the fact that it is not only concerning for recordCount anymore.

autumnust, line 65: Second comment for this: Considering the wide usage of `InputRecordCountHelper` among different implementation of CompactionAction, does it make sense to share an instance of that instead of instantiating the same object in different implementations?

autumnust (changes requested):

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/verify/CompactionThresholdVerifier.java
ZihanLi58: So the logic of verifier is if any of the verifier fail the dataset, the compaction will not run. In this case, if gmce verifier say it needs to re compact but threshold verifier say it does not need to be compacted, then the dataset will be skipped. That's the reason I embedded the logic here.

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/action/CompactionGMCEPublishingAction.java
ZihanLi58: +1, but this will be a backward incompatible change, might be better when we release a new version?

autumnust: That's fair.

autumnust (approved):
+1

PR #3256, opened 2021-04-08
GOBBLIN-1420: Log uncaught exceptions from StreamModelTaskRunner

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1420

### Description
StreamModelTaskRunner submits a thread that consumes and processes records in the streaming mode. Currently, this thread can die silently in case of an uncaught exception (e.g. OOMEs), leaving no trace of the exception. 

### Tests
The proposed change only sets the UncaughtExceptionHandler for the thread created by StreamModelTaskRunner. Existing unit tests should cover the changes proposed in this PR.

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
@@ -39,6 +39,7 @@ import org.apache.gobblin.runtime.fork.Fork;
 import org.apache.gobblin.source.extractor.Extractor;
 import org.apache.gobblin.source.extractor.StreamingExtractor;
 import org.apache.gobblin.util.ExponentialBackoff;
+import org.apache.gobblin.util.LoggingUncaughtExceptionHandler;
 import org.apache.gobblin.writer.AcknowledgableWatermark;
 import org.apache.gobblin.writer.FineGrainedWatermarkTracker;
 import org.apache.gobblin.writer.WatermarkManager;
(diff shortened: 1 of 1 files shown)


sv2000:
@ZihanLi58 Please review.

ZihanLi58:
LGTM overall, can you check the failed tests?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
```diff
@@ -151,9 +152,11 @@ protected void run() throws Exception {
         this.task.configureStreamingFork(fork);
       }
     }
-    new Thread(() -> {
```
autumnust, line 154: Just curious: What was the motivation to put this line alone in a new thread? isn't the `connect()` an async call ?

ZihanLi58: Since we are single fork in this case, isForkAsync is false and we don't call observeOn method which cause the connect() method is not an async call and will hang forever until error/finished. It's changed in this PR https://github.com/apache/gobblin/pull/3214/files

ZihanLi58 (approved):
+1 LGTM

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/StreamModelTaskRunner.java
autumnust: If that's the case, why not create scheduler anyway (without looking at `isForkAsync`) ? Wrapping thread on top of rxjava doesn't seem to be right.

autumnust (approved):

PR #3268, opened 2021-04-22
[GOBBLIN-1437] Segregate FlowConfigs/Delete and FlowExecutions/Delete functions

arjun4084346:
Dear Gobblin maintainers,

@jack-moseley  @Will-Lo  @aplex @umustafi please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1437

### Description
This PR
1) make KafkaJobMonitor s parseJobSpec() method return the entire JobSpec, so that the caller method processMessage() can distinguish between a DELETE FLOW and CANCEL JOB requests and take appropriate actions. Actions are based on the doc https://docs.google.com/document/d/1lanG0R-w81MK5anpGBEE0hWF3rI8qArX4bixKzN0wEs/edit 
2) some minor refactoring

### Tests
Besides some existing unit tests, did following tests on LinkedIn test environment
1) flowConfig/Delete doing a) delete of state store b) cancelling a running job c) deleting specs from spec catalog
2) flowExecutions/Delete doing a) leaving state store intact b) cancelling a running job c ) leaving spec catalog intact

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
@@ -105,6 +105,14 @@ public class NonObservingFSJobCatalog extends FSJobCatalog {
       remove(jobURI, false);
   }
 
+  /**
+   * Removes a job from job catalog.
+   * If it is a flow delete request from service side, then alwaysTriggerListeners should be set to true so that a
+   * running job will also be cancelled. If it is a remove request because a job finishes or has been submitted to helix,
+   * then alwaysTriggerListeners should be set to false so that job cancellation is not triggered.
+   * @param jobURI job uri
+   * @param alwaysTriggerListeners if it is true, a running job will be cancelled.
+   */
   public synchronized void remove(URI jobURI, boolean alwaysTriggerListeners) {
     Preconditions.checkState(state() == State.RUNNING, String.format("%s is not running.", this.getClass().getName()));
     try {
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
@@ -102,18 +102,6 @@ public class NonObservingFSJobCatalog extends FSJobCatalog {
    */
   @Override
   public synchronized void remove(URI jobURI) {
-      remove(jobURI, false);
-  }
-
-  /**
-   * Removes a job from job catalog.
-   * If it is a flow delete request from service side, then alwaysTriggerListeners should be set to true so that a
-   * running job will also be cancelled. If it is a remove request because a job finishes or has been submitted to helix,
-   * then alwaysTriggerListeners should be set to false so that job cancellation is not triggered.
-   * @param jobURI job uri
-   * @param alwaysTriggerListeners if it is true, a running job will be cancelled.
-   */
-  public synchronized void remove(URI jobURI, boolean alwaysTriggerListeners) {
     Preconditions.checkState(state() == State.RUNNING, String.format("%s is not running.", this.getClass().getName()));
     try {
       long startTime = System.currentTimeMillis();
@@ -131,12 +119,6 @@ public class NonObservingFSJobCatalog extends FSJobCatalog {
       throw new RuntimeException("When removing a JobConf. file, issues unexpected happen:" + e.getMessage());
     } catch (SpecNotFoundException e) {
       LOGGER.warn("No file with URI:" + jobURI + " is found. Deletion failed.");
-    } finally {
-      // HelixRetriggeringJobCallable deletes the job file after submitting it to helix,
-      // so we will have to trigger listeners regardless the existence of job file to cancel the job
-      if (alwaysTriggerListeners) {
-        this.listeners.onCancelJob(jobURI);
-      }
     }
   }
 }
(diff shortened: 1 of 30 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -365,38 +365,42 @@ public void remove(Spec spec, Properties headers) throws IOException {
...
-            SpecProducer<Spec> producer = jobExecutionPlan.getSpecExecutor().getProducer().get();
-            _log.info(String.format("Going to delete JobSpec: %s on Executor: %s", jobSpec, producer));
-            producer.deleteSpec(jobSpec.getUri(), headers);
-          } catch (Exception e) {
-            _log.error(String.format("Could not delete JobSpec: %s for flow: %s", jobSpec, spec), e);
-          }
-        }
       }
+      // We need to recompile the flow to find the spec producer,
+      // If compilation results is different, it remove request can go to some different spec producer
```
umustafi, line 375: small nitpick here "it's"

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
```diff
@@ -105,6 +105,14 @@ public synchronized void remove(URI jobURI) {
       remove(jobURI, false);
   }
 
+  /**
+   * Removes a job from job catalog.
+   * If it is a flow delete request from service side, then alwaysTriggerListeners should be set to true so that a
+   * running job will also be cancelled. If it is a remove request because a job finishes or has been submitted to helix,
+   * then alwaysTriggerListeners should be set to false so that job cancellation is not triggered.
+   * @param jobURI job uri
+   * @param alwaysTriggerListeners if it is true, a running job will be cancelled.
```
aplex, line 114: This architecture looks a bit strange to me. And looks like this API ambiguity was the source of the original bug.

Since removal from the catalog does not always mean cancelling the job, should this job cancellation be moved out of the catalog code and be done explicitly by the callers? E.g. caller can first cancel the job, if they want, and then remove it from the catalog.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_monitor/KafkaJobMonitor.java
```diff
@@ -100,17 +100,29 @@ protected void shutdownMetrics()
...
+          case UNKNOWN: // unknown are considered as add request to maintain backward compatibility
+            this.newSpecs.inc();
+            this.jobCatalog.put(parsedMessage);
+            break;
+          case DELETE:
+            this.removedSpecs.inc();
+            URI jobSpecUri = parsedMessage.getUri();
+            this.jobCatalog.remove(jobSpecUri, true);
+            // Delete the job state if it is a delete spec request
+            deleteStateStore(jobSpecUri);
```
aplex, line 119: What will happen in the service process fails after removal of the job, but before state store is deleted? Will the state store eventually be deleted/cleaned up?

arjun4084346: kafka offsets are committed in HighLevelConsumer only after this method returns. So if state store deletion is not completed, both 'removal of job' and 'state store deletion' will be retried whenever the service restarts.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_monitor/KafkaJobMonitor.java
```diff
@@ -100,17 +100,29 @@ protected void shutdownMetrics()
...
-          // Delete the job state if it is a delete spec request
-          deleteStateStore(jobSpecUri);
+      Collection<JobSpec> parsedCollection = parseJobSpec(message.getValue());
+      for (JobSpec parsedMessage : parsedCollection) {
+        SpecExecutor.Verb verb = SpecExecutor.Verb.valueOf(parsedMessage.getMetadata().get(SpecExecutor.VERB_KEY));
+
+        switch (verb) {
+          case ADD:
+          case UPDATE:
+          case UNKNOWN: // unknown are considered as add request to maintain backward compatibility
```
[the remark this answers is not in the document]
arjun4084346: JobSpecs created in `SimpleKafkaSpecProducer` uses correct `verb` for all operations; and only those are consumed by this JobMonitor, unless implemented in other way in forked repositories. But some code flows, especially unit tests, do not set `verb`. When `verb` is not set, it is set to `unknown` by default. 
If the  old JobMonitor reads a new Verb, I have updated the PR to ignore new values.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/NonObservingFSJobCatalog.java
arjun4084346: called the cancel on spec catalog listeners directly from job monitor

arjun4084346: passing EventBus object from JobConfigurationManager to JobMonitor and making JobMonitor directly posting a cancel event

aplex (commented):
Overall looks good, but I think we still have some dead code left in JobCatalogListener - onCancelJob, CancelJobCallback and in surrounding classes. Can it be removed to prevent confusion for the next time, or it's still used for something?

arjun4084346:
> Overall looks good, but I think we still have some dead code left in JobCatalogListener - onCancelJob, CancelJobCallback and in surrounding classes. Can it be removed to prevent confusion for the next time, or it's still used for something?

Removed the dead code.

aplex (approved):
LGTM

sv2000 (changes requested):
IMO - this PR should be split into at least 2 PRs. It is very hard to understand what the implications of this PR will be - since it will clearly have an impact on Gobblin cluster and Gobblin-on-Yarn. My suggestion is to break it down into: 1. API changes on Gobblin service and 2. Changes on Gobblin cluster.

PR #3272, opened 2021-04-30
[GOBBLIN-1438] Only load failed dags at time of resume

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1438

### Description
This PR changes failed dags to only be loaded if resume is called (and only the dag that was resumed is loaded) to avoid keeping all failed dags in memory at all times, since this can cause memory issues.

Changes for this:
- In `DagManager`, keep a list of failed dagIds instead of a map of all failed dags. This list is still needed to know which dags to clean up, and also allows us to avoid calling the dag store for dags that aren't there.
- Add an API in `DagStateStore` to load only one dag, because previously the only option was to load all dags at once, which would still cause memory issues.
- Add an API  in `DagStateStore` to get all dag IDs

### Tests
Updated unit tests and tested resuming locally

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -20,6 +20,7 @@ package org.apache.gobblin.service.modules.orchestration;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -188,7 +189,6 @@ public class DagManager extends AbstractIdleService {
   private final int defaultQuota;
   private final Map<String, Integer> perUserQuota;
   private final long failedDagRetentionTime;
-  private final Map<String, Dag<JobExecutionPlan>> failedDags = new ConcurrentHashMap<>();
 
   private volatile boolean isActive = false;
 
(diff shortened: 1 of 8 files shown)


aplex:
LGTM

jack-moseley:
I realized that getting group/name from Dag ID (which is in the form `group_name_execid`) is not possible, because the group and name themselves could contain underscores, so had to make some changes to how the storename/tablename is parsed.

aplex (approved):

jack-moseley:
@sv2000 please merge

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -372,16 +372,17 @@ public synchronized void setActive(boolean active) {
         //Initializing state store for persisting Dags.
         this.dagStateStore = createDagStateStore(config, topologySpecMap);
         this.failedDagStateStore = createDagStateStore(ConfigUtils.getConfigOrEmpty(config, FAILED_DAG_STATESTORE_PREFIX).withFallback(config), topologySpecMap);
+        Set<String> failedDagIds = Collections.synchronizedSet(this.failedDagStateStore.getDagIds());
```
sv2000, line 375: Can we not look up a dag with a given dagId from the dagStateStore when a resume is called? (which should be infrequent), instead of loading dagIds on start up?

jack-moseley: That is what I am doing, see the changes in `beginResumingDag` just below this. The reason for still keeping this list is that it's used to know which dags are old enough to delete from the failed dag state store, and also it could avoid some calls to mysql (if calling resume with an invalid dag). The list of dagIds is only strings so it shouldn't be large in memory.

sv2000 (approved):
+1

PR #3273, opened 2021-05-01
[GOBBLIN-1439] Handle flow configs that cause compilation errors from being added and blocking scheduler

Will-Lo:
Dear Gobblin maintainers,

@aplex @arjun4084346  @jack-moseley  @umustafi  Please take a look

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1439

### Description
FlowSpecs that fail to compile due to unexpected exceptions (e.g. NullPointerException) can cause the Job Scheduler to silently fail adding jobs.
To fix the error we should first reject users from adding flowConfigs that cause compilation errors. Also, if encountering a flow that fails to compile, log it and continue scheduling the rest of the flows instead of silently failing with unscheduled jobs.

### Tests
Added unit tests mocking listener compilation errors

### Commits


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -338,6 +339,14 @@ public Spec getSpecWrapper(URI uri) {
 
     if (triggerListener) {
       AddSpecResponse<CallbacksDispatcher.CallbackResults<SpecCatalogListener, AddSpecResponse>> response = this.listeners.onAddSpec(flowSpec);
+      // If flow fails callbacks, need to prevent adding the flow to the catalog
+      if (!response.getValue().getFailures().isEmpty()) {
+        for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry: response.getValue().getFailures().entrySet()) {
+          flowSpec.getCompilationErrors().add(ExceptionUtils.getStackTrace(entry.getValue().getError()));
+        }
+        responseMap.put(ServiceConfigKeys.COMPILATION_SUCCESSFUL, new AddSpecResponse<>("false"));
```
arjun4084346, line 347: Can we reuse `if (isCompileSuccessful(responseMap)) {` block?

arjun4084346:
a minor comment. lgtm otherwise

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -366,11 +376,17 @@ public Spec getSpecWrapper(URI uri) {
...
 
-  public static boolean isCompileSuccessful(Map<String, AddSpecResponse> responseMap) {
-    AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(
-        ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, new AddSpecResponse<>(""));
-
-    return isCompileSuccessful(addSpecResponse.getValue());
+  public boolean isCompileSuccessful(Map<String, AddSpecResponse> responseMap) {
+    for (SpecCatalogListener listener: this.listeners.getListeners()) {
+      String listenerClass = listener.getClass().getCanonicalName();
+      AddSpecResponse<String> addSpecResponse =
```
aplex, line 382: Is it safe to call listeners from this place? I wonder if they have any side effects, and will call them multiple times for the same flow. It would be good to review what  they actually do and what is the contract/docs for the listener interface.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobSchedulerTest.java
```diff
@@ -129,6 +130,65 @@ public void testDisableFlowRunImmediatelyOnStart()
...
+    SpecCompiler mockCompiler = Mockito.mock(SpecCompiler.class);
+    Mockito.when(mockOrchestrator.getSpecCompiler()).thenReturn(mockCompiler);
+    Mockito.doAnswer((Answer<Void>) a -> {
+      scheduler.isCompilerHealthy = true;
+      return null;
+    }).when(mockCompiler).awaitHealthy();
+
+    scheduler.setActive(true);
+
+    AssertWithBackoff.create().timeoutMs(6000).maxSleepMs(2000).backoffFactor(2)
```
aplex, line 177: Timed waits in tests can introduce flakiness. Usually this happens when dozens of tests run concurrently on an overloaded CI machine, and even basic operation take way longer when developer expects.

The best way to address this it to have some kind of notification/event from a class, and react to it. It will have no delays, and can also tolerate the situation when it takes a long time for operation to happen.

If it's difficult or unclear how to do that in this case, I suggest to bump the timeout to 15-30 seconds.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/MockSpecCatalogListener.java
```diff
@@ -0,0 +1,45 @@
...
+
+package org.apache.gobblin.runtime.api;
+
+import java.net.URI;
+import java.util.Properties;
+import org.apache.gobblin.configuration.ConfigurationKeys;
+import org.apache.gobblin.runtime.spec_catalog.AddSpecResponse;
+
+
+public class MockSpecCatalogListener implements SpecCatalogListener {
```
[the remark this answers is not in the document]
Will-Lo: Thanks for the suggestion, going to use the Mockito mocks and clean it up

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
Will-Lo: I needed this originally for testing purposes but realized that I can mock the name of the expected listener instead

aplex (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -338,8 +340,16 @@ public Spec getSpecWrapper(URI uri) {
 
     if (triggerListener) {
       AddSpecResponse<CallbacksDispatcher.CallbackResults<SpecCatalogListener, AddSpecResponse>> response = this.listeners.onAddSpec(flowSpec);
-      for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry: response.getValue().getSuccesses().entrySet()) {
-        responseMap.put(entry.getKey().getName(), entry.getValue().getResult());
+      // If flow fails callbacks, need to prevent adding the flow to the catalog
+      if (!response.getValue().getFailures().isEmpty()) {
+        for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry: response.getValue().getFailures().entrySet()) {
+          flowSpec.getCompilationErrors().add(ExceptionUtils.getStackTrace(entry.getValue().getError()));
```
sv2000, line 346: Can we just use Throwables.getStackTraceAsString() to be consistent with the rest of the code base?

sv2000 (changes requested):

sv2000 (approved):
+1. LGTM.

PR #3281, opened 2021-05-14
[GOBBLIN-1444] Use Guice as DI framework in Gobblin service

aplex:
Previously, to initialize Gobblin service, we used a mixture
of dependency injection, direct class creation and config-based
class creation. In this change, we unify the service initialization
by moving towards using dependency injection(DI) with Guice everywhere.

Using DI will help with (1) unit testing; (2) overriding classes in the
middle of the dependency with company-specific implementations, and
(3) will improve code readability, as dependencies between classes
become visible from the outside and explicit.

We also move away from name-based injection for classes. Name-based
injection is useful when code needs several different implementations
of the same interface. In our use case, we had only one implementation
for each service that can be active. Name-based injection was used for
company-specific overrides, but there is a better way to do it - with
Guice module overrides.

There are still several improvements left to be done to have a full
Guice migration, but we'll make them in separate commits to limit the
PR size.

https://issues.apache.org/jira/browse/GOBBLIN-1444

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/util/ClassAliasResolver.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/util/ClassAliasResolver.java
@@ -19,8 +19,6 @@ package org.apache.gobblin.util;
 import java.util.List;
 import java.util.Map;
 
-import lombok.extern.slf4j.Slf4j;
-
 import org.reflections.Reflections;
 import org.reflections.util.ConfigurationBuilder;
 
@@ -28,6 +26,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
+import javax.annotation.concurrent.ThreadSafe;
+import lombok.extern.slf4j.Slf4j;
 
 import org.apache.gobblin.annotation.Alias;
 
@@ -49,16 +50,16 @@ import org.apache.gobblin.annotation.Alias;
  * </b>
  */
 @Slf4j
+@ThreadSafe
 public class ClassAliasResolver<T> {
 
   // Scan all packages in the classpath with prefix gobblin, com.linkedin.gobblin when class is loaded.
   // Since scan is expensive we do it only once when class is loaded.
-  private static final Reflections REFLECTIONS = new Reflections(new ConfigurationBuilder().forPackages("gobblin",
-      "com.linkedin.gobblin", "org.apache.gobblin"));
-
-  Map<String, Class<? extends T>> aliasToClassCache;
+  private static final Reflections REFLECTIONS =
+      new Reflections(new ConfigurationBuilder().forPackages("gobblin", "com.linkedin.gobblin", "org.apache.gobblin"));
   private final List<Alias> aliasObjects;
   private final Class<T> subtypeOf;
+  ImmutableMap<String, Class<? extends T>> aliasToClassCache;
 
   public ClassAliasResolver(Class<T> subTypeOf) {
     Map<String, Class<? extends T>> cache = Maps.newHashMap();
@@ -108,7 +109,8 @@ public class ClassAliasResolver<T> {
       return Class.forName(aliasOrClassName).asSubclass(this.subtypeOf);
     } catch (ClassCastException cce) {
       throw new ClassNotFoundException(
-          String.format("Found class %s but it cannot be cast to %s.", aliasOrClassName, this.subtypeOf.getName()), cce);
+          String.format("Found class %s but it cannot be cast to %s.", aliasOrClassName, this.subtypeOf.getName()),
+          cce);
     }
   }
 
--- a/gobblin-restli/gobblin-restli-utils/src/main/java/org/apache/gobblin/restli/EmbeddedRestliServer.java
+++ b/gobblin-restli/gobblin-restli-utils/src/main/java/org/apache/gobblin/restli/EmbeddedRestliServer.java
@@ -34,6 +34,7 @@ import com.google.inject.Binder;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
 import com.google.inject.Module;
+import com.google.inject.Singleton;
 import com.linkedin.r2.filter.FilterChain;
 import com.linkedin.r2.filter.FilterChains;
 import com.linkedin.r2.filter.compression.EncodingType;
@@ -69,6 +70,7 @@ import lombok.Setter;
  * * name - defaults to the name of the first resource in the resource collection.
  * * injector - an {@link Injector} to inject dependencies into the Rest.li resources.
  */
+@Singleton
 public class EmbeddedRestliServer extends AbstractIdleService {
 
   private static final int MAX_PORT = 65535;
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/ControllerUserDefinedMessageHandlerFactory.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/ControllerUserDefinedMessageHandlerFactory.java
@@ -37,10 +37,10 @@ import org.apache.gobblin.runtime.api.FlowSpec;
 import org.apache.gobblin.runtime.api.Spec;
 import org.apache.gobblin.service.FlowConfig;
 import org.apache.gobblin.service.FlowConfigResourceLocalHandler;
+import org.apache.gobblin.service.FlowConfigsResourceHandler;
 import org.apache.gobblin.service.FlowId;
 import org.apache.gobblin.service.ServiceConfigKeys;
 import org.apache.gobblin.service.modules.restli.FlowConfigUtils;
-import org.apache.gobblin.service.modules.restli.GobblinServiceFlowConfigResourceHandler;
 import org.apache.gobblin.service.modules.scheduler.GobblinServiceJobScheduler;
 
 /**
@@ -51,7 +51,7 @@ import org.apache.gobblin.service.modules.scheduler.GobblinServiceJobScheduler;
 class ControllerUserDefinedMessageHandlerFactory implements MessageHandlerFactory {
   private boolean flowCatalogLocalCommit;
   private GobblinServiceJobScheduler jobScheduler;
-  private GobblinServiceFlowConfigResourceHandler resourceHandler;
+  private FlowConfigsResourceHandler resourceHandler;
   private String serviceName;
 
   @Override
@@ -80,12 +80,12 @@ class ControllerUserDefinedMessageHandlerFactory implements MessageHandlerFactor
   private static class ControllerUserDefinedMessageHandler extends MessageHandler {
     private boolean flowCatalogLocalCommit;
     private GobblinServiceJobScheduler jobScheduler;
-    private GobblinServiceFlowConfigResourceHandler resourceHandler;
+    private FlowConfigsResourceHandler resourceHandler;
     private String serviceName;
 
     public ControllerUserDefinedMessageHandler(Message message, NotificationContext context, String serviceName,
         boolean flowCatalogLocalCommit, GobblinServiceJobScheduler scheduler,
-        GobblinServiceFlowConfigResourceHandler resourceHandler) {
+        FlowConfigsResourceHandler resourceHandler) {
       super(message, context);
       this.serviceName = serviceName;
       this.flowCatalogLocalCommit = flowCatalogLocalCommit;
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceConfiguration.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceConfiguration.java
@@ -1,0 +1,115 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.core;
+
+import java.util.Objects;
+
+import org.apache.hadoop.fs.Path;
+
+import com.typesafe.config.Config;
+
+import javax.annotation.Nullable;
+import lombok.Getter;
+import lombok.ToString;
+
+import org.apache.gobblin.service.ServiceConfigKeys;
+import org.apache.gobblin.util.ConfigUtils;
+
+
+@ToString
+public class GobblinServiceConfiguration {
+
+  @Getter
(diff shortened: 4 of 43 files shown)


aplex:
Tagging @sv2000 @jack-moseley and @arjun4084346 for review

jack-moseley:
Are you using the gobblin style from https://gobblin.readthedocs.io/en/latest/developer-guide/CodingStyle/#code-style-template-file? Seems like the imports are reordered incorrectly in many files.

aplex:
@jack-moseley , turns out if was a default import style. Reformatted it with Gobblin style.

gobblin-api/src/main/java/org/apache/gobblin/util/ClassAliasResolver.java
```diff
@@ -100,15 +101,17 @@ public String resolve(final String possibleAlias) {
    * and is also a subclass of {@link #subtypeOf}, if it fails it returns a class object for name
    * <code>aliasOrClassName</code>.
    */
-  public Class<? extends T> resolveClass(final String aliasOrClassName) throws ClassNotFoundException {
+  public Class<? extends T> resolveClass(final String aliasOrClassName)
+      throws ClassNotFoundException {
```
arjun4084346, line 105: Is this something controlled by the gobblin codestyle file or by the intellij settings?

aplex: I think it's part of Gobblin code style settings, and IntelliJ automatically reformats the code to this.

arjun4084346: Should we keep this style?

aplex:
Note for reviewers: this PR contains only refactoring, so there should not be any functionality changes. Many files contain only attribute changes, like specifying injection constructor.

A couple of big chunks of logic was moved out of GobblinServiceManager: 
1. Configuration reading and parsing is now in GobblinServiceConfiguration
2. Logic to glue classes together moved to GobblinServiceGuiceModule

We use several features of Guice that were not widely used in code base before:

1. Marking the class as "@Singleton" - this will make sure that DI framework creates only one instance of it. By default, Guice will create a new instance of the class for each http request that Gobblin service gets.
2. Marking constructors as @Inject to specify the one that should be used for dependency injection 
3. Using unique interfaces instead of having named instances (more context in the comments where that happens)
4. Using providers - those are like factories that generate an instance of the class for Guice
5. Using standard javax.Inject annotations instead of using Guice-specific ones. They both work in the same way, but using standard once opens us to use other DI frameworks if needed in the future.

More info is available in Guice docs: https://github.com/google/guice/wiki/MentalModel

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -146,26 +148,12 @@ public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Op
...
   }
 
-  public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Optional<DagManager> dagManager, Optional<Logger> log) {
-    this(config, topologyCatalog, dagManager, log, true);
-  }
-
-  public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Optional<DagManager> dagManager, Logger log) {
-    this(config, topologyCatalog, dagManager, Optional.of(log));
+  @Inject
+  public Orchestrator(Config config, FlowStatusGenerator flowStatusGenerator, Optional<TopologyCatalog> topologyCatalog,
```
arjun4084346, line 152: Why to place flowStatusGenerator at different positions in the constructors?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -146,26 +148,12 @@ public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Op
...
+      Optional<DagManager> dagManager, Optional<Logger> log) {
+    this(config, topologyCatalog, dagManager, log, flowStatusGenerator, true);
```
arjun4084346, line 154: Are we sure about removing these public constructors? Looks redundant to me though, but can be used by external users.

aplex: This class should be created using dependency injection, so other constructors should not be needed.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
aplex: Changed the other constructor to match the order, so that non-optional parameters come first in the constructor.

gobblin-restli/gobblin-restli-utils/src/main/java/org/apache/gobblin/restli/EmbeddedRestliServer.java
```diff
@@ -69,6 +70,7 @@
  * * name - defaults to the name of the first resource in the resource collection.
  * * injector - an {@link Injector} to inject dependencies into the Rest.li resources.
  */
+@Singleton
```
arjun4084346, line 73: So, only one instance of this class is needed? If we pass different injector (or other params), then also it reuses the existing instance?

aplex: Normally, if the class is marked with Singleton annotation , Guice will create only one instance of it. All other classes that depend on this one will get the same instance.

However, in this case the class is created by our provider, and not by Guice directly, so we need an extra step to make sure only one instance exists. I've updated Guice module like this:

` binder.bind(EmbeddedRestliServer.class).toProvider(EmbeddedRestliServerProvider.class).in(Singleton.class)`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java
```diff
@@ -0,0 +1,253 @@
...
+    String helixInstanceName = HelixUtils.buildHelixInstanceName(config, GobblinServiceManager.class.getSimpleName());
+
+    LOGGER.info(
+        "Creating Helix cluster if not already present [overwrite = false]: " + zkConnectionString);
+    HelixUtils.createGobblinHelixCluster(zkConnectionString, helixClusterName, false);
+
+    return HelixUtils.buildHelixManager(helixInstanceName, helixClusterName, zkConnectionString);
+  }
+
+  protected static <T> Class<? extends T> getClassByNameOrAlias(Class<T> baseClass, Config config,
```
arjun4084346, line 224: Should this method be moved to GobblinConstructorUtils ?

arjun4084346:
+1 LGTM

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -192,7 +193,7 @@ public String toString() {
 
   private volatile boolean isActive = false;
 
-  public DagManager(Config config, boolean instrumentationEnabled) {
+  public DagManager(Config config, JobStatusRetriever jobStatusRetriever, boolean instrumentationEnabled) {
```
autumnust, line 196: Is backward compatibility a concern here?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/utils/InjectionNames.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.utils;
+
+public final class InjectionNames {
```
autumnust, line 20: What's the reason for this additional class?

autumnust (changes requested):

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/ControllerUserDefinedMessageHandlerFactory.java
```diff
@@ -51,7 +51,7 @@
 class ControllerUserDefinedMessageHandlerFactory implements MessageHandlerFactory {
   private boolean flowCatalogLocalCommit;
   private GobblinServiceJobScheduler jobScheduler;
-  private GobblinServiceFlowConfigResourceHandler resourceHandler;
+  private FlowConfigsResourceHandler resourceHandler;
```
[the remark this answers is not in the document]
aplex: Instead of injecting a specific implementation, we inject an interface. GobblinServiceFlowConfigResourceHandler  implements this FlowConfigsResourceHandler.

It's generally a good practice to use interface-based dependency injection everywhere. That will allow us to replace implementations completely, and its more convenient to mock in unit tests. 

For example, some of our classes inherit from Guava service class. And that service class has final methods. Mockito by default cannot override final methods, so when we try to mock such dependency, part of the calls go to the mocked methods, while others go to real implementation, causing complicated bugs. If the dependency is an interface, we can mock it fully.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
aplex: It's not a part of the user-facing API, just an internal class. It should be instantiated by dependency injection framework, so even if new dependencies are added, they should be injected automatically.

So, I don't think backward compatibility is important here.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/utils/InjectionNames.java
aplex: These names are used for dependency injection, when we need to inject different instances of the same type, or inject constants. Since they are used in multiple classes, they don't fully belong to any of them, and that's why they landed in a separate class. FLOW_CATALOG_LOCAL_COMMIT for example is used in GobblinServiceFlowConfigResourceHandler and GobblinServiceFlowConfigV2ResourceHandler.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -471,21 +310,83 @@ private void  handleLeadershipChange(NotificationContext changeContext) {
...
+
+    registerServicesInLauncher();
+
+    // Register Scheduler to listen to changes in Flows
+    if (configuration.isSchedulerEnabled()) {
+      this.flowCatalog.addListener(this.scheduler);
+    }
+  }
+
+  void ensureInjected() {
```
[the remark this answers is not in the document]
aplex: updated

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+        ConfigUtils.getBoolean(config, ServiceConfigKeys.GOBBLIN_SERVICE_JOB_STATUS_MONITOR_ENABLED_KEY, true);
+    this.isSchedulerEnabled =
+        ConfigUtils.getBoolean(config, ServiceConfigKeys.GOBBLIN_SERVICE_SCHEDULER_ENABLED_KEY, true);
+    this.isRestLIServerEnabled =
+        ConfigUtils.getBoolean(config, ServiceConfigKeys.GOBBLIN_SERVICE_RESTLI_SERVER_ENABLED_KEY, true);
+    this.isTopologySpecFactoryEnabled =
+        ConfigUtils.getBoolean(config, ServiceConfigKeys.GOBBLIN_SERVICE_TOPOLOGY_SPEC_FACTORY_ENABLED_KEY, true);
+  }
+
+  public String getServiceName() {
```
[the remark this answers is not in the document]
aplex: changed to lombok properties

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java
```diff
@@ -0,0 +1,253 @@
...
+
+    binder.bind(GobblinInstanceEnvironment.class).toInstance(gobblinInstanceEnvironment);
+
+    binder.bind(EventBus.class)
+        .annotatedWith(Names.named(GobblinServiceManager.SERVICE_EVENT_BUS_NAME))
+        .toInstance(new EventBus(GobblinServiceManager.class.getSimpleName()));
+
+    binder.bindConstant().annotatedWith(Names.named(InjectionNames.SERVICE_NAME)).to(serviceConfig.getServiceName());
+
+    binder.bindConstant()
```
sv2000, line 122: Is the general recommendation then that all service wide constants which were getting passed around in config objects or as arguments use DI instead?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -193,7 +194,7 @@ public String toString() {
...
```
sv2000, line 197: @Singleton?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -105,9 +111,10 @@
    */
   public static final String DR_FILTER_TAG = "dr";
 
-  public GobblinServiceJobScheduler(String serviceName, Config config, Optional<HelixManager> helixManager,
-      Optional<FlowCatalog> flowCatalog, Optional<TopologyCatalog> topologyCatalog, Orchestrator orchestrator,
-      SchedulerService schedulerService, Optional<Logger> log) throws Exception {
+  @Inject
```
sv2000, line 114: Should this be a singleton too?

sv2000 (changes requested):
A few minor comments.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
aplex: It already is. "@singleton" annotation is placed on the class, and that controls Guice behavior. We also select the constructor that will be used for dependency injection with "@Inject" annotation.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
aplex: same as above

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java
aplex: If the class has many configuration settings, than it's better to have a config object and inject it. That's what we do for GobblinServiceManager that gets GobblinServiceConfiguration injected into it.

Service name is injected into a couple of classes that don't have much in common, and those classes don't have a lot of configuration of their own. In future PRs we can consider adding something like ServiceEnvironment class that will hold the service name, id and other properties that can be useful for the classes that run inside Gobblin service. 

There are still a bunch of places in the code that needs to be refactored for better DI support, but I left them out at the moment to reduce the PR size.

sv2000 (approved):
+1.

PR #3283, opened 2021-05-16
[PoC][NO-MERGE] Worker node on gobblin PoC

hkoosha:
This is how workers on k8s could look like. I followed mapreduce's approach: 
- Have a PersistentVolume shared among workers (same as HDFS).
- Copy jars and job.state there.
- Have a docker container available to k8s. The container startup scripts takes an argument: classpath. Then launches the JVM with a simple java application which in turn launches the main(String[]) method of the class passed as argument. 
- The KuberJobLauncher, uses k8s api to create a pod with the container described above (and copies the jars/files to PersistentVolume's location beforehand)
- Wait for the pod to exit.

Instead of pod I should have used k8s's job. But for some reason pv and pvc do not work with jobs on my machine.

k3d is used (https://github.com/rancher/k3d) to launch a cluster.

And finally the MRJobLaucnher is ab(used) to launch the a k8s-based job.

The steps to reproduce the build and jobs:
```bash
# This acts as the HDFS path, everything is put here! hardcoded a lot of places
sudo mkdir -p /a
# Spawn a cluster and have the docker image ready
cd gobblin-kubernetes-support/docker-java-runner/docker-local/
make k3d-reg k3d-cluster
make clean build push
cd -
# Create PV and PVC
cd gobblin-kubernetes-support/k8s/
make apply
# Launch gobblin
cd -
# build gobblin, last step (cp) fails, you can ignore that
make build
make run
make build
```


Will-Lo:
Hi,

I think this is really great stuff, especially as a PoC. Went over it a bit and a few thoughts, will look into this more too but wanted to brainstorm with you:

1) I think the user experience we want to submit jobs would be in some form of bash script (easy) or k8s operator (hard). This current work is closest to a script where we can submit jobs to k8s clusters with something like `./bin/gobblin-kubernetes.sh --cluster <cluster_id> --config <gobblin_config>` 

2) Passing the job properties to be read by the job launcher would be the ideal state. Since Gobblin is very configurable this will be a must. I'll take a closer look at the code too, but this is a good stepping stone to reaching that point.

3) There should be a new JobLauncher which is independent from MRJobLauncher that extends from the k8s launcher. Passing the configurations here helps as you are no longer constrained to what the CLI options are.

PR #3286, opened 2021-05-24
[GOBBLIN-1448] Send failed timer instead of success timer when job is cancelled

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1448

### Description
Currently AbstractJobLauncher will default to sending a jobSucceededTimer if job state is anything other than FAILED. It probably makes more sense to send a jobFailedTimer instead in the case of CANCELLED status.

### Tests
Trivial change

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -598,7 +598,7 @@ public abstract class AbstractJobLauncher implements JobLauncher {
               }
             });
 
-            if (jobState.getState() == JobState.RunningState.FAILED) {
+            if (jobState.getState() == JobState.RunningState.FAILED || jobState.getState() == JobState.RunningState.CANCELLED) {
               notifyListeners(this.jobContext, jobListener, TimingEvent.LauncherTimings.JOB_FAILED, new JobListenerAction() {
                 @Override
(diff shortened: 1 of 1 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -598,7 +598,7 @@ public void apply(JobListener jobListener, JobContext jobContext)
               }
             });
 
-            if (jobState.getState() == JobState.RunningState.FAILED) {
+            if (jobState.getState() == JobState.RunningState.FAILED || jobState.getState() == JobState.RunningState.CANCELLED) {
```
aplex, line 601: Is there another bug in "else" branch below? It creates a "JOB_SUCCEEDED" event, but the listener is calling "onJobFailure" event. Also the same logic is a coupe of blocks above.

jack-moseley: It's been like that for a long time, I'm not sure if that was originally intentional or not. But there is no `onJobSuccess` API, so I'm not sure what would replace it. Also would have to be careful changing something here, since changing logic could affect many gobblin jobs.

aplex (approved):

jack-moseley:
@autumnust could you merge if you get a chance? Thanks

PR #3288, opened 2021-05-26
[GOBBLIN-1450] Update test framework version for latest IntelliJ

aplex:
The recent version of IntelliJ are not fully compatible with old
versions of testng and junit that we use. As a result IDE does not show
test names in the UI correctly.

New version of test frameworks should solve this problem.

We also fix a couple of misconfigurations for TestNG. When the class
is annotated with "@test", TestNG will run all public methods as tests.
We convert a couple of helper methods to private to fix that.

https://issues.apache.org/jira/browse/GOBBLIN-1450


aplex:
@autumnust , can you take a look at merge?

autumnust (approved):
+1 LGTM

PR #3289, opened 2021-05-26
[GOBBLIN-1451] Log better error messages from mappers

aplex:
Previously, the root cause of the problems were wrapped in several
container exceptions that added little value in troubleshooting and
confused the users. They frequently reported that the job failed with
"RetryException" without looking at the root cause of it.

With this change, we'll unwrap the exceptions from tasks and show
the possible root cause at the top of the error message.

https://issues.apache.org/jira/browse/GOBBLIN-1451

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
@@ -56,6 +56,7 @@ import org.apache.gobblin.runtime.Task;
 import org.apache.gobblin.runtime.TaskContext;
 import org.apache.gobblin.runtime.TaskExecutor;
 import org.apache.gobblin.runtime.TaskState;
+import org.apache.gobblin.runtime.util.ExceptionCleanupUtils;
 import org.apache.gobblin.runtime.util.ForkMetrics;
 import org.apache.gobblin.state.ConstructState;
 import org.apache.gobblin.stream.ControlMessage;
@@ -265,11 +266,13 @@ public class Fork<S, D> implements Closeable, FinalState, RecordStreamConsumer<S
 
       compareAndSetForkState(ForkState.RUNNING, ForkState.SUCCEEDED);
     } catch (Throwable t) {
+      Throwable cleanedUpException = ExceptionCleanupUtils.removeEmptyWrappers(t);
+
       // Set throwable to holder first because AsynchronousFork::putRecord can pull the throwable when it detects ForkState.FAILED status.
       ForkThrowableHolder holder = Task.getForkThrowableHolder(this.broker);
-      holder.setThrowable(this.getIndex(), t);
+      holder.setThrowable(this.getIndex(), cleanedUpException);
       this.forkState.set(ForkState.FAILED);
-      this.logger.error(String.format("Fork %d of task %s failed to process data records. Set throwable in holder %s", this.index, this.taskId, holder), t);
+      this.logger.error(String.format("Fork %d of task %s failed.", this.index, this.taskId), cleanedUpException);
     } finally {
       this.cleanup();
     }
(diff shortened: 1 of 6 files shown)


aplex:
tagging @jack-moseley , @sv2000

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -265,11 +268,15 @@ public void run() {
 
       compareAndSetForkState(ForkState.RUNNING, ForkState.SUCCEEDED);
     } catch (Throwable t) {
+      Throwable rootCauseException =
```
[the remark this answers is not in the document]
sv2000: The other concern I have is if Fork is the right layer at which we should be unwrapping exceptions. Why not do this inside the Task class?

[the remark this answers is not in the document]
sv2000: Finally, we may want to preserve the wrapped exceptions as it shows the sequence in which the exceptions are propagated up, which is useful for debugging. The "likely" root cause exception can be logged in addition to the wrapped exception.

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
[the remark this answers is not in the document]
aplex: Refactored the code to clean up exceptions at lower layers.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -24,6 +24,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.rholder.retry.RetryException;
```
sv2000, line 27: Unused import.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
```diff
@@ -50,12 +51,14 @@
 import org.apache.gobblin.records.RecordStreamWithMetadata;
 import org.apache.gobblin.runtime.BoundedBlockingRecordQueue;
 import org.apache.gobblin.runtime.ExecutionModel;
+import org.apache.gobblin.runtime.ForkException;
```
sv2000, line 54: Unused import.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ExceptionCleanupUtils.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+  /**
+   * The method will return the bottom of the exception stack, up to one of the provided wrappers.
+   *
+   * It is useful when exception can be wrapped multiple times in the codebase, and only some of those wrappings
+   * can contain useful information, while others will just confuse the user.
+   *
+   * Examples: if the chain is wrapper1 -> wrapper2 -> real1 -> real2, the method will return real1 -> real2 chain.
+   * */
+  public static Throwable getRootCauseUpToWrapper(Throwable exception, Class<?>... wrapperExceptionTypes) {
```
sv2000, line 53: Is this method used anywhere other than test class?

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/fork/Fork.java
aplex: That should have been caught by checkstyle plugin, but apparently it is configured to check only a very small subset of things. I'll send separate PRs to enable more checks.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/util/ExceptionCleanupUtils.java
aplex: It was used before, but is not used now. Removed to have less code.

sv2000 (approved):
+1.

PR #3290, opened 2021-05-26
[GOBBLIN-1452] Add meters for successful/failed dags in total and by flowGroup

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1452

### Description
Add meters to track the number of successful and failed flows run by gaas. There is one service level meter that tracks all flows, and a meter for each flowGroup.

### Tests
Tested locally by printing value of meter over time

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
@@ -59,6 +59,7 @@ import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.instrumented.Instrumented;
 import org.apache.gobblin.metrics.ContextAwareCounter;
 import org.apache.gobblin.metrics.ContextAwareGauge;
+import org.apache.gobblin.metrics.ContextAwareMeter;
 import org.apache.gobblin.metrics.MetricContext;
 import org.apache.gobblin.metrics.RootMetricContext;
 import org.apache.gobblin.metrics.ServiceMetricNames;
@@ -431,6 +432,8 @@ public class DagManager extends AbstractIdleService {
     private final Map<String, Integer> perUserQuota;
     private final AtomicLong orchestrationDelay = new AtomicLong(0);
     private static Map<String, FlowState> flowGauges = Maps.newHashMap();
+    private ContextAwareMeter allSuccessfulMeter;
+    private ContextAwareMeter allFailedMeter;
 
     private JobStatusRetriever jobStatusRetriever;
     private DagStateStore dagStateStore;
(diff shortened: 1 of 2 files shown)


jack-moseley:
@arjun4084346 @Will-Lo please review

arjun4084346:
Why are we not counting flow status by flow name?
LGTM otherwise

jack-moseley:
There's already a RunningStatus metric that can track flow status on flow level, the purpose of this one is to aggregate on group level.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -460,6 +463,10 @@ public synchronized void setActive(boolean active) {
         this.jobStatusPolledTimer = Optional.of(this.metricContext.timer(ServiceMetricNames.JOB_STATUS_POLLED_TIMER));
         ContextAwareGauge<Long> orchestrationDelayMetric = metricContext.newContextAwareGauge(ServiceMetricNames.FLOW_ORCHESTRATION_DELAY,
             () -> orchestrationDelay.get());
+        this.allSuccessfulMeter = metricContext.contextAwareMeter(
```
aplex, line 466: How will this meter work? Looks like the number of successful flows will start at 0 when service restarts, and then increment with each flow completion.

Will this work with our monitoring system to get an aggregate metric like "Number of successful flows in the past X minutes"?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
```diff
@@ -460,6 +463,10 @@ public synchronized void setActive(boolean active) {
...
+            MetricRegistry.name(ServiceMetricNames.GOBBLIN_SERVICE_PREFIX, ServiceMetricNames.SUCCESSFUL_FLOW_METER));
+        this.allFailedMeter = metricContext.contextAwareMeter(
```
aplex, line 468: What will happen with the metrics when the failover from master to slave occurs? Are we going to see a sudden spike in metrics that can trigger alerts?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: See https://javadoc.io/doc/io.dropwizard.metrics/metrics-core/3.2.1/com/codahale/metrics/Meter.html

It can return an "exponentially-weighted moving average rate" of the past 5/10/15 minutes. Which is not exactly a "number of flows", but instead each time an event occurs there is an increase in the meter, then it gradually drops to 0 over the course of the window.

I thought it makes more sense to use the existing meter concept rather than try to make our own implementation of a meter by having a gauge that we reset ourselves. And if we count number of flows as we thought of before, I think it is confusing (like if you see the number 20 on a graph, does that mean 20 failures in the past 5 minutes from that point? 20 failures in a fixed 5 minute interval?).

With this I think we can still look for spikes in failures on the graph, or look at the ratio of success to failure meters to measure the health of the system.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/DagManager.java
jack-moseley: The meters would drop to 0, then return to normal after the length of the window time (a few minutes). It will be a slight blip in the graph, but I think it would not be enough to impact alerts as long as we do alerts on the right things (either looking for spikes in failures, or looking at ratio of success vs failure over a sufficiently large window).

aplex (approved):

Will-Lo (approved):

sv2000 (approved):
+1

PR #3291, opened 2021-05-27
[GOBBLIN-1453] Improve error reporting on failed flow compilations and fix bugs wher…

Will-Lo:
…e failed flow compilations can be saved

Dear Gobblin maintainers,

### JIRA

### Description

- Flowconfigs that fail to compile and return 400 errors should report why
- Fix bug where templates that fail to compile would reject flow configs (even if they are not relevant)

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
@@ -368,8 +368,9 @@ public class FlowCatalog extends AbstractIdleService implements SpecCatalog, Mut
   }
 
   public static boolean isCompileSuccessful(Map<String, AddSpecResponse> responseMap) {
+    // If we cannot get the response from the scheduler, assume that the flow failed compilation
     AddSpecResponse<String> addSpecResponse = responseMap.getOrDefault(
-        ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, new AddSpecResponse<>(""));
+        ServiceConfigKeys.GOBBLIN_SERVICE_JOB_SCHEDULER_LISTENER_CLASS, new AddSpecResponse<>(null));
     return isCompileSuccessful(addSpecResponse.getValue());
   }
 
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
@@ -17,7 +17,6 @@
 
 package org.apache.gobblin.runtime.spec_catalog;
 
-import com.google.common.base.Throwables;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
@@ -29,7 +28,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
-import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.commons.lang3.reflect.ConstructorUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -42,7 +40,6 @@ import com.typesafe.config.ConfigException;
 
 import javax.annotation.Nonnull;
 import lombok.Getter;
-
 import org.apache.gobblin.instrumented.Instrumented;
 import org.apache.gobblin.metrics.MetricContext;
 import org.apache.gobblin.metrics.Tag;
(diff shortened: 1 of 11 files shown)


Will-Lo:
@jack-moseley @arjun4084346 @aplex Please take a look

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
```diff
@@ -201,13 +202,19 @@ public void awaitHealthy() throws InterruptedException {
...
         for (DataNode destNode : destNodes) {
           long authStartTime = System.nanoTime();
-          boolean authorized = this.dataMovementAuthorizer.isMovementAuthorized(flowSpec, sourceNode, destNode);
-          Instrumented.updateTimer(dataAuthorizationTimer, System.nanoTime() - authStartTime, TimeUnit.NANOSECONDS);
-          if (!authorized) {
-            String message = String.format("Data movement is not authorized for flow: %s, source: %s, destination: %s",
-                flowSpec.getUri().toString(), source, destination);
-            log.error(message);
-            datasetFlowSpec.getCompilationErrors().add(message);
+          try {
```
jack-moseley, line 205: So for the NPE we experienced previously, it happened inside this block?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -289,10 +288,9 @@ public AddSpecResponse onAddSpec(Spec addedSpec) {
 
     // always try to compile the flow to verify if it is compilable
     Dag<JobExecutionPlan> dag = this.orchestrator.getSpecCompiler().compileFlow(flowSpec);
+    // If dag is null then a compilation error has occurred
```
arjun4084346, line 291: Let's updated the javadoc that response is null if compilation fails

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -341,16 +338,9 @@ public Spec getSpecWrapper(URI uri) {
...
-      if (!response.getValue().getFailures().isEmpty()) {
-        for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry: response.getValue().getFailures().entrySet()) {
-          flowSpec.getCompilationErrors().add(Throwables.getStackTraceAsString(entry.getValue().getError()));
-          responseMap.put(entry.getKey().getName(), new AddSpecResponse(entry.getValue().getResult()));
-        }
-      } else {
-        for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry : response.getValue().getSuccesses().entrySet()) {
-          responseMap.put(entry.getKey().getName(), entry.getValue().getResult());
-        }
+      // If flow fails compilation, the result will have a non-empty string with the error
```
arjun4084346, line 341: Should we update teh TopologyCatalog also?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
Will-Lo: Yes it happened in the DMASAuthorizer

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
```diff
@@ -341,16 +338,9 @@ public Spec getSpecWrapper(URI uri) {
...
+      for (Map.Entry<SpecCatalogListener, CallbackResult<AddSpecResponse>> entry : response.getValue().getSuccesses().entrySet()) {
+        responseMap.put(entry.getKey().getName(), entry.getValue().getResult());
```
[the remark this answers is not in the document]
Will-Lo: We do not need to look at response.getValue().getFailures() as long as no uncaught exceptions occur. It might be better if we set the default response from the listener to be a failure, so I'll go ahead and make those changes. So in the following scenarios:
1. If flow compilation from the listener is successful, return a dag in the AddSpecResponse
2. If flow compilation from the listener is unsuccessful, return null.
3. If flow compilation from the listener cannot be found, response should be null. 

I think (3) is what caused an error with saving uncompilable flow specs before.

Or we throw an exception on every compilation error that occurs, that can work too. Except that the template resolution exceptions for flow edges can be unrelated to flow so those shouldn't be thrown.

umustafi (commented):
LGTM except for the other tests that are failing. They look like they mostly have to do with missing configs in the spec store (perhaps because of the change in behavior for bad specs)?

jack-moseley (approved):

arjun4084346:
+1 LGTM

arjun4084346 (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalog.java
Will-Lo: We don't need to update topology catalog I think here as it doesn't look into failures

Will-Lo:
@sv2000 Could you review/merge?

PR #3296, opened 2021-06-02
Add initial version of automatic troubleshooter

aplex:
Gobblin users, developers and SREs spend a significant amount of time
troubleshooting issues with their jobs and flows. Previously,
troubleshooting required looking into logs across multiple systems,
like Hadoop mappers, Azkaban jobs and Gobblin service. Log messages
required knowledge of how Gobblin works internally, and users frequently
had to involve Gobblin developers to understand them.

This is an initial commit for automatic Gobblin troubleshooter. The
current implementation will intercept logs from Mappers and Azkaban job
and compile a list of issues based on them. This list will be filtered
and prioritized. Then it will be displayed to the user in logs, and
forwarded as an event for consumption by Gobblin service.

Future commits will add issue reporting in Gobblin service, as well
as improve on issue filtering & refining.

For more information, check the docs on AutomaticTroubleshooter class,
and review the design doc: https://docs.google.com/document/d/1BAYr-dHtdauX6Uf13VP3-IlHkyUQ5V64PjFIg9MtpaA/edit#

### DIFF
--- a/gobblin-modules/gobblin-azkaban/build.gradle
+++ b/gobblin-modules/gobblin-azkaban/build.gradle
@@ -38,6 +38,7 @@ dependencies {
   compile externalDependency.httpclient
   compile externalDependency.httpmime
   compile externalDependency.jodaTime
+  compile externalDependency.log4j
   compile externalDependency.slf4j
   compile externalDependency.typesafeConfig
   compile externalDependency.hadoopYarnApi
--- a/gobblin-modules/gobblin-azkaban/build.gradle
+++ b/gobblin-modules/gobblin-azkaban/build.gradle
@@ -38,7 +38,6 @@ dependencies {
   compile externalDependency.httpclient
   compile externalDependency.httpmime
   compile externalDependency.jodaTime
-  compile externalDependency.log4j
   compile externalDependency.slf4j
   compile externalDependency.typesafeConfig
   compile externalDependency.hadoopYarnApi
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -87,6 +87,7 @@ import org.apache.gobblin.runtime.locks.JobLockEventListener;
 import org.apache.gobblin.runtime.locks.JobLockException;
 import org.apache.gobblin.runtime.locks.LegacyJobLockFactoryManager;
 import org.apache.gobblin.runtime.troubleshooter.AutomaticTroubleshooter;
+import org.apache.gobblin.runtime.troubleshooter.AutomaticTroubleshooterFactory;
 import org.apache.gobblin.runtime.troubleshooter.IssueRepository;
 import org.apache.gobblin.runtime.util.JobMetrics;
 import org.apache.gobblin.source.Source;
@@ -192,7 +193,7 @@ public abstract class AbstractJobLauncher implements JobLauncher {
     clusterNameTags.addAll(Tag.fromMap(ClusterNameTags.getClusterNameTags()));
     GobblinMetrics.addCustomTagsToProperties(jobProps, clusterNameTags);
 
-    troubleshooter = AutomaticTroubleshooter.createForJob(ConfigUtils.propertiesToConfig(jobProps));
+    troubleshooter = AutomaticTroubleshooterFactory.createForJob(ConfigUtils.propertiesToConfig(jobProps));
     troubleshooter.start();
 
     // Make a copy for both the system and job configuration properties and resolve the job-template if any.
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -31,8 +31,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.gobblin.destination.DestinationDatasetHandlerService;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -64,6 +62,7 @@ import org.apache.gobblin.commit.DeliverySemantics;
 import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.WorkUnitState;
 import org.apache.gobblin.converter.initializer.ConverterInitializerFactory;
+import org.apache.gobblin.destination.DestinationDatasetHandlerService;
 import org.apache.gobblin.metrics.ContextAwareGauge;
 import org.apache.gobblin.metrics.GobblinMetrics;
 import org.apache.gobblin.metrics.GobblinMetricsRegistry;
@@ -87,6 +86,8 @@ import org.apache.gobblin.runtime.locks.JobLock;
 import org.apache.gobblin.runtime.locks.JobLockEventListener;
 import org.apache.gobblin.runtime.locks.JobLockException;
 import org.apache.gobblin.runtime.locks.LegacyJobLockFactoryManager;
+import org.apache.gobblin.runtime.troubleshooter.AutomaticTroubleshooter;
+import org.apache.gobblin.runtime.troubleshooter.IssueRepository;
 import org.apache.gobblin.runtime.util.JobMetrics;
 import org.apache.gobblin.source.Source;
 import org.apache.gobblin.source.WorkUnitStreamSource;
@@ -173,6 +174,8 @@ public abstract class AbstractJobLauncher implements JobLauncher {
   // Used to generate additional metadata to emit in timing events
   private final MultiEventMetadataGenerator multiEventMetadataGenerator;
 
+  private final AutomaticTroubleshooter troubleshooter;
+
   public AbstractJobLauncher(Properties jobProps, List<? extends Tag<?>> metadataTags)
       throws Exception {
     this(jobProps, metadataTags, null);
@@ -188,6 +191,9 @@ public abstract class AbstractJobLauncher implements JobLauncher {
     List<Tag<?>> clusterNameTags = Lists.newArrayList();
     clusterNameTags.addAll(Tag.fromMap(ClusterNameTags.getClusterNameTags()));
     GobblinMetrics.addCustomTagsToProperties(jobProps, clusterNameTags);
+
+    troubleshooter = AutomaticTroubleshooter.createForJob(ConfigUtils.propertiesToConfig(jobProps));
+    troubleshooter.start();
 
     // Make a copy for both the system and job configuration properties and resolve the job-template if any.
     this.jobProps = new Properties();
@@ -206,7 +212,7 @@ public abstract class AbstractJobLauncher implements JobLauncher {
         instanceBroker = createDefaultInstanceBroker(jobProps);
       }
 
-      this.jobContext = new JobContext(this.jobProps, LOG, instanceBroker);
+      this.jobContext = new JobContext(this.jobProps, LOG, instanceBroker, troubleshooter.getIssueRepository());
       this.eventBus.register(this.jobContext);
 
       this.cancellationExecutor = Executors.newSingleThreadExecutor(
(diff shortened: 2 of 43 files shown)


aplex:
Tagging @jack-moseley , @Will-Lo  and @sv2000  for review. Take a look folks.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
```diff
@@ -0,0 +1,187 @@
...
+  protected void startUp()
+      throws Exception {
+    setupLogAppender();
+  }
+
+  @Override
+  protected void shutDown()
+      throws Exception {
+    removeLogAppender();
+  }
```
jack-moseley, line 75: I'm not sure I understand why `AutomaticTroubleshooter` is an async service, couldn't we just directly call the methods to setup and remove the log appender? It doesn't actually do anything asynchronously does it?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/ThrowableWithErrorCode.java
```diff
@@ -0,0 +1,28 @@
...
+package org.apache.gobblin.runtime;
+
+public interface ThrowableWithErrorCode {
+
+  /**
+   * Returns an error code that uniquely identifies the problem in the given context.
+   *
+   * The error code will be used programmatically, to take different recovery actions.
+   * */
+  String getErrorCode();
```
jack-moseley, line 27: Maybe in the javadoc include a couple of example error codes, from this it's not quite clear if an error code would be just a number, a single word, a short description of the issue, etc. I guess any of those would work the same, but would probably be good to have a standard.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
```diff
@@ -737,11 +738,19 @@ private static FileSystem buildFileSystem(Properties jobProps, Configuration con
...
+
     @Override
     protected void setup(Context context) {
       final State gobblinJobState = HadoopUtils.getStateFromConf(context.getConfiguration());
       TaskAttemptID taskAttemptID = context.getTaskAttemptID();
 
+      issueRepository = new InMemoryIssueRepository();
+      troubleshooter = new AutomaticTroubleshooter(issueRepository);
+
+      troubleshooter.startAsync().awaitRunning();
```
jack-moseley, line 752: This one should probably have a 1 minute timeout as well to be consistent with the other starts/stops of the troubleshooter.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -189,6 +197,10 @@ public AbstractJobLauncher(Properties jobProps, List<? extends Tag<?>> metadataT
     clusterNameTags.addAll(Tag.fromMap(ClusterNameTags.getClusterNameTags()));
     GobblinMetrics.addCustomTagsToProperties(jobProps, clusterNameTags);
 
+    issueRepository = new InMemoryIssueRepository();
+    troubleshooter = new AutomaticTroubleshooter(issueRepository);
```
jack-moseley, line 201: Instead of enabling this in general for all gobblin jobs, should we consider making this troubleshooter opt-in to start with, or at least have a config to disable it? Just in case there are unexpected performance issues, it floods logs too much, it's sending too many events, etc.

jack-moseley (commented):
Mostly LGTM, several comments/questions

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
```diff
@@ -0,0 +1,187 @@
...
+    log.info("Removed logger for automatic troubleshooting. Processed {} events.",
+             troubleshooterLogger.getProcessedEventCount());
+  }
+
+  /**
+   * This method will sort, filter and enhance the list of issues to make it more meaningful for the user.
+   */
+  public void refineIssues()
+      throws TroubleshooterException {
+    IssueRefinery refinery = new DefaultIssueRefinery();
```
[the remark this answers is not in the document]
aplex: This refinery logic will be significantly rebuilt in future PRs. What we have today is just a stub.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
```diff
@@ -0,0 +1,187 @@
...
+
+    List<Issue> issues = issueRepository.getAll();
+    issues.sort(Comparator.comparing(Issue::getSeverity).reversed().thenComparing(Issue::getTime));
```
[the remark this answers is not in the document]
aplex: moved to refinery

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
```diff
@@ -0,0 +1,187 @@
...
+    removeLogAppender();
+  }
+
+  private void setupLogAppender() {
+    org.apache.log4j.Logger rootLogger = LogManager.getRootLogger();
+
+    troubleshooterLogger = new AutoTroubleshooterLogAppender(issueRepository);
+    troubleshooterLogger.setThreshold(Level.WARN);
+    troubleshooterLogger.activateOptions();
+    rootLogger.addAppender(troubleshooterLogger);
```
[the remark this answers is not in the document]
aplex: Yes. Components can log various warnings that will also be summarized and displayed to the user in Azkaban/Gobblin service. This is an opportunity for them to provide suggestion to user for non-fatal situations.

Issue repository will do deduplications of all issues, and has a limit of 100 issues per job, so if we have runaway code that just writes thousands of warnings, it shouldn't impact the troubleshooter or repository.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
aplex: yeah, most of the services in Gobblin codebase inherit from AbstractIdleService, but in this case it's not really needed. Converted to plain start/stop methods.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/ThrowableWithErrorCode.java
aplex: Added an example.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/mapreduce/MRJobLauncher.java
aplex: Replaced with synchronous calls.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
aplex: Added configs to disable event reporting and all troubleshooter activity in case of unexpected problems

jack-moseley (approved):
LGTM

gradle/scripts/dependencyDefinitions.gradle
```diff
@@ -50,6 +50,7 @@ ext.externalDependency = [
     "guava": "com.google.guava:guava:15.0",
     "groovy": "org.codehaus.groovy:groovy:2.4.8",
     "gson": "com.google.code.gson:gson:2.6.2",
+    "gsonJavatimeSerialisers": "com.fatboyindustrial.gson-javatime-serialisers:gson-javatime-serialisers:1.1.1",
```
Will-Lo, line 53: does this dependency require us to add a repository?

Will-Lo (commented):
Would it make sense for AbstractJobLauncher to use the troubleshooter to report errors rather than setting the jobContext error on its own after it catches an exception? I believe there's some other tools with some automation that parses the JobException() as the root cause.

gradle/scripts/dependencyDefinitions.gradle
aplex: nope, it's in the standard maven repo - https://search.maven.org/artifact/com.fatboyindustrial.gson-javatime-serialisers/gson-javatime-serialisers/1.1.1/jar

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooterConfig.java
```diff
@@ -0,0 +1,58 @@
...
+  private final boolean disableEventReporting;
+
+  @Builder.Default
+  private int inMemoryRepositoryMaxSize = ConfigurationKeys.DEFAULT_TROUBLESHOOTER_IN_MEMORY_ISSUE_REPOSITORY_MAX_SIZE;
+
+  @Inject
+  public AutomaticTroubleshooterConfig(Config config) {
+    Objects.requireNonNull(config, "Config cannot be null");
+
+    disabled = ConfigUtils.getBoolean(config, ConfigurationKeys.TROUBLESHOOTER_DISABLED, false);
```
sv2000, line 50: Can we set the default to true? Let pipelines explicitly enable the trouble shooter if they desire.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooterConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    disableEventReporting =
```
sv2000, line 51: Same comment as above.

sv2000 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooter.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public void stop() {
+    if (config.isDisabled()) {
+      return;
+    }
+    removeLogAppender();
+  }
+
+  private void setupLogAppender() {
+    org.apache.log4j.Logger rootLogger = LogManager.getRootLogger();
```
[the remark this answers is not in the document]
aplex: Looks like SLF4j does not have an ability to configure appenders. From their docs: 

> SLF4J is only a facade, meaning that it does not provide a complete logging solution. Operations such as configuring appenders or setting logging levels cannot be performed with SLF4J. Thus, at some point in time, any non-trivial application will need to directly invoke the underlying logging system. 

http://www.slf4j.org/faq.html#when

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooterConfig.java
aplex: Switched to "disabled by default" for initial rollout, to reduce the risks, but the intention is to enable it by default in the future. People that would need troubleshooter most, wouldn't know that it exists or how to turn it on.

Will-Lo (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutomaticTroubleshooterConfig.java
sv2000: Have we switched the default behavior? I am still seeing it enabled by default.

sv2000 (changes requested):

gobblin-modules/gobblin-azkaban/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
   compile externalDependency.httpclient
   compile externalDependency.httpmime
   compile externalDependency.jodaTime
+  compile externalDependency.log4j
```
sv2000, line 41: I think we should avoid this. Why do we need to add this dependency?

sv2000: The log4j dependency of this module was explicitly removed in https://github.com/apache/gobblin/pull/3081/files. This module depends on log4j-over-slf4j module, and you should be able to use the log4j APIs via that module.

aplex: That was unintentional, removed this dependency.

sv2000 (approved):
+1.

PR #3297, opened 2021-06-04
[GOBBLIN-1458] Add endpoint to trigger a flow in GaaS

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1458

### Description

Previously the way to manually trigger an execution of a scheduled flow is to send a partial update that sets the flow's `runImmediately` property to true. This PR improves the API by adding a restli action that sends the same partial update. With this, flows can be triggered through the URL: `/flowconfigsV2/(flowGroup:test1,flowName:test1)?action=trigger`

Also copied the implementation of `partialUpdateFlowConfig` to `FlowConfigV2ResourceLocalHandler` for testing convenience (normally when running the service `GobblinServiceFlowConfigResourceHandler` would be used, but for tests it is `FlowConfigV2ResourceLocalHandler`).

### Tests
Updated unit test and tested in local gaas

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowconfigsV2.restspec.json
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowconfigsV2.restspec.json
@@ -79,7 +79,12 @@
       } ]
     } ],
     "entity" : {
-      "path" : "/flowconfigsV2/{id}"
+      "path" : "/flowconfigsV2/{id}",
+      "actions" : [ {
+        "name" : "runImmediately",
+        "doc" : "Trigger a new execution of an existing flow",
+        "returns" : "string"
+      } ]
     }
   }
 }
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowConfigV2Client.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowConfigV2Client.java
@@ -36,6 +36,7 @@ import com.linkedin.r2.RemoteInvocationException;
 import com.linkedin.r2.transport.common.Client;
 import com.linkedin.r2.transport.common.bridge.client.TransportClientAdapter;
 import com.linkedin.r2.transport.http.client.HttpClientFactory;
+import com.linkedin.restli.client.ActionRequest;
 import com.linkedin.restli.client.CreateIdEntityRequest;
 import com.linkedin.restli.client.DeleteRequest;
 import com.linkedin.restli.client.FindRequest;
@@ -253,6 +254,19 @@ public class FlowConfigV2Client implements Closeable {
     ResponseFuture<EmptyRecord> response = _restClient.get().sendRequest(deleteRequest);
 
     response.getResponse();
+  }
+
+  public String runImmediately(FlowId flowId)
+      throws RemoteInvocationException {
+    LOG.debug("runImmediately with groupName " + flowId.getFlowGroup() + " flowName " + flowId.getFlowName());
+
+    ActionRequest<String> runImmediatelyRequest = _flowconfigsV2RequestBuilders.actionRunImmediately()
+        .id(new ComplexResourceKey<>(flowId, new FlowStatusId())).build();
+
+    Response<String> response = (Response<String>) FlowClientUtils.sendRequestWithRetry(_restClient.get(), runImmediatelyRequest,
+        FlowconfigsV2RequestBuilders.getPrimaryResource());
+
+    return response.getEntity();
   }
 
   @Override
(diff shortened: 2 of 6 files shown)


jack-moseley:
@umustafi @Will-Lo please review

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowconfigsV2.restspec.json
```diff
@@ -79,7 +79,12 @@
       } ]
     } ],
     "entity" : {
-      "path" : "/flowconfigsV2/{id}"
+      "path" : "/flowconfigsV2/{id}",
+      "actions" : [ {
+        "name" : "trigger",
```
arjun4084346, line 84: should we name it trigger-execution ?

arjun4084346 (approved):

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
```diff
@@ -160,6 +160,8 @@ public void testCheckFlowExecutionId() throws Exception {
   @Test
   public void testPartialUpdate() throws Exception {
     FlowId flowId = new FlowId().setFlowGroup(TEST_GROUP_NAME).setFlowName(TEST_FLOW_NAME_3);
+    ServiceRequester testRequester = new ServiceRequester("testName", "USER_PRINCIPAL", "testFrom");
```
umustafi, line 163: nitpick but do we want to add these as constants if they are reused?

umustafi (commented):
Otherwise LGTM

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowconfigsV2.restspec.json
jack-moseley: Changed it to `triggerExecution`

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/test/java/org/apache/gobblin/service/FlowConfigV2Test.java
jack-moseley: Good point, updated the tests that don't need to modify the requester to use a constant

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowconfigsV2.restspec.json
```diff
@@ -79,7 +79,12 @@
       } ]
     } ],
     "entity" : {
-      "path" : "/flowconfigsV2/{id}"
+      "path" : "/flowconfigsV2/{id}",
+      "actions" : [ {
+        "name" : "triggerExecution",
```
aplex, line 84: The action name is a bit strange. What options do we have here? run/start/execute/startFlow/startExecution ... ?

"trigger" is an uncommon way to describe an action in APIs.

jack-moseley: Is it? I can think of a couple LI tools that use "trigger" to describe an API action, and for example Azkaban uses "trigger" when referring to flows launched through data landing or kafka events: https://azkaban.readthedocs.io/en/latest/eventTrigger.html

To me, trigger seems like a good way to describe launching a flow that exists through a one-time request instead of due to the schedule or because the flowConfig was just sent.

aplex: Looks like Azkaban use "trigger" as a noun, and there are noun-triggers in database world as well. 
Google translate says that as a verb, trigger means "cause (an event or situation) to happen or exist." . I think it's more common to use this verb to describe something in the past, like "flow was triggered by a schedule".

https://translate.google.com/?sl=auto&tl=en&text=trigger&op=translate

Also, this action will set "runImmediately:true", so we might just call it run or runImmediately for consistency.

jack-moseley: Oh changed it to `executeFlow`, but just noticed you suggested `runImmediately`, I think that makes sense, I will change it to that.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-client/src/main/java/org/apache/gobblin/service/FlowConfigV2Client.java
```diff
@@ -256,6 +257,19 @@ public void deleteFlowConfigWithStateStore(FlowId flowId)
     response.getResponse();
   }
 
+  public String triggerFlowConfig(FlowId flowId)
```
aplex, line 260: Should this also be updated to reflect a new name? Also in other places.

jack-moseley: Yeah it should have been, updated that as well.

aplex (approved):

PR #3298, opened 2021-06-04
[GOBBLIN-1459] correct column length in MysqlSpecStore, MysqlDagStore, MysqlJobStatu…

arjun4084346:
…sStore

Dear Gobblin maintainers,

@Will-Lo @jack-moseley  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1459

### Description
  -some column length in MysqlSpecStore, MysqlDagStore, MysqlJobStatusStore are incorrect. limit on these column lengths is decided by the limit on flow group and flow name limits.
  - MysqlDagStateStore uses an internal MysqlStateStore. This PR will change it to use MysqlDagStore which just extends MysqlStateStore so that it can use override create table statement

### Tests
added a few tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
@@ -135,6 +135,8 @@ public class ServiceConfigKeys {
 
   public static final int MAX_FLOW_NAME_LENGTH = 128; // defined in FlowId.pdl
   public static final int MAX_FLOW_GROUP_LENGTH = 128; // defined in FlowId.pdl
+  public static final int MAX_JOB_NAME_LENGTH = 374;
+  public static final int MAX_JOB_GROUP_LENGTH = 374;
   public static final String STATE_STORE_TABLE_SUFFIX = "gst";
   public static final String STATE_STORE_KEY_SEPARATION_CHARACTER = ".";
   public static final String DAG_STORE_KEY_SEPARATION_CHARACTER = "_";
--- a/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
@@ -132,4 +132,11 @@ public class ServiceConfigKeys {
   // Group Membership authentication service
   public static final String GROUP_OWNERSHIP_SERVICE_CLASS = GOBBLIN_SERVICE_PREFIX + "groupOwnershipService.class";
   public static final String DEFAULT_GROUP_OWNERSHIP_SERVICE = "org.apache.gobblin.service.NoopGroupOwnershipService";
+
+  public static final int MAX_FLOW_NAME = 128; // defined in FlowId.pdl
+  public static final int MAX_FLOW_GROUP = 128; // defined in FlowId.pdl
+  public static final String STATE_STORE_TABLE_SUFFIX = "gst";
+  public static final String STATE_STORE_KEY_SEPARATION_CHARACTER = ".";
+  public static final String DAG_STORE_KEY_SEPARATION_CHARACTER = "_";
+
 }
--- a/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
@@ -133,8 +133,8 @@ public class ServiceConfigKeys {
   public static final String GROUP_OWNERSHIP_SERVICE_CLASS = GOBBLIN_SERVICE_PREFIX + "groupOwnershipService.class";
   public static final String DEFAULT_GROUP_OWNERSHIP_SERVICE = "org.apache.gobblin.service.NoopGroupOwnershipService";
 
-  public static final int MAX_FLOW_NAME = 128; // defined in FlowId.pdl
-  public static final int MAX_FLOW_GROUP = 128; // defined in FlowId.pdl
+  public static final int MAX_FLOW_NAME_LENGTH = 128; // defined in FlowId.pdl
+  public static final int MAX_FLOW_GROUP_LENGTH = 128; // defined in FlowId.pdl
   public static final String STATE_STORE_TABLE_SUFFIX = "gst";
   public static final String STATE_STORE_KEY_SEPARATION_CHARACTER = ".";
   public static final String DAG_STORE_KEY_SEPARATION_CHARACTER = "_";
--- a/gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
+++ b/gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
@@ -50,8 +50,8 @@ public class MysqlDagStore<T extends State> extends MysqlStateStore<T> {
 
   @Override
   protected String getCreateJobStateTableTemplate() {
-    int maxStoreName = ServiceConfigKeys.MAX_FLOW_NAME + ServiceConfigKeys.STATE_STORE_KEY_SEPARATION_CHARACTER.length()
-        + ServiceConfigKeys.MAX_FLOW_GROUP;
+    int maxStoreName = ServiceConfigKeys.MAX_FLOW_NAME_LENGTH + ServiceConfigKeys.STATE_STORE_KEY_SEPARATION_CHARACTER.length()
+        + ServiceConfigKeys.MAX_FLOW_GROUP_LENGTH;
     int maxTableName = 13; // length of flowExecutionId which is epoch timestamp
 
     return "CREATE TABLE IF NOT EXISTS $TABLE$ (store_name varchar(" + maxStoreName + ") CHARACTER SET latin1 COLLATE latin1_bin not null,"
--- a/gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
+++ b/gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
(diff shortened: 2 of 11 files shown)


gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
```diff
@@ -132,4 +132,11 @@
   // Group Membership authentication service
   public static final String GROUP_OWNERSHIP_SERVICE_CLASS = GOBBLIN_SERVICE_PREFIX + "groupOwnershipService.class";
   public static final String DEFAULT_GROUP_OWNERSHIP_SERVICE = "org.apache.gobblin.service.NoopGroupOwnershipService";
+
+  public static final int MAX_FLOW_NAME = 128; // defined in FlowId.pdl
```
aplex, line 136: maybe MAX_FLOW_NAME_LENGTH? and same for the next config.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FlowSpec.java
```diff
@@ -467,5 +468,10 @@ public static FlowConfig toFlowConfig(Spec spec) {
           .setFlowName(flowProps.getProperty(ConfigurationKeys.FLOW_NAME_KEY)))
           .setProperties(flowPropsAsStringMap);
     }
+
+    public static int maxFlowSpecUri() {
```
aplex, line 472: we can end the name with "length" or "size" to make it more clear.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.gobblin.service.ServiceConfigKeys;
+
+
+@Slf4j
+/**
+ * An implementation of {@link MysqlStateStore} backed by MySQL to store Dag.
+ *
+ * @param <T> state object type
+ **/
+public class MysqlDagStore<T extends State> extends MysqlStateStore<T> {
```
aplex, line 36: Do we need a separate MysqlDagStore, or we can just bump the lengths in the base class?

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
```diff
@@ -0,0 +1,62 @@
...
+    super(dataSource, stateStoreTableName, compressedValues, stateClass);
+  }
+
+  @Override
+  protected String getCreateJobStateTableTemplate() {
+    int maxStoreName = ServiceConfigKeys.MAX_FLOW_NAME + ServiceConfigKeys.STATE_STORE_KEY_SEPARATION_CHARACTER.length()
+        + ServiceConfigKeys.MAX_FLOW_GROUP;
+    int maxTableName = 13; // length of flowExecutionId which is epoch timestamp
+
+    return "CREATE TABLE IF NOT EXISTS $TABLE$ (store_name varchar(" + maxStoreName + ") CHARACTER SET latin1 COLLATE latin1_bin not null,"
```
aplex, line 57: Looks like we'll use longer column names when new state stores are created, but what is our story for the existing DBs? Are we going to ask SREs to increase the limit manually, or we'll include some "alter table" in the codebase?

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlJobStatusStateStore.java
```diff
@@ -70,6 +71,21 @@ public MysqlJobStatusStateStore(DataSource dataSource, String stateStoreTableNam
     return getAll(storeName, flowExecutionId + "%", true);
   }
 
+  @Override
+  protected String getCreateJobStateTableTemplate() {
```
aplex, line 75: Would be good to add a couple of tests that will verify the state store behavior with long names. Older version of MySql with a certain db layout types would limit the total length of the primary key to about 700-800 bytes. At the same time, it was possible to define any limits in table schema, even more than 800 total bytes. But, when you try to insert a long key, a runtime exception will occur.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
arjun4084346: Let's aks SREs to increase the limit manually as it is a one time thing?

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
arjun4084346: The base class MysqlStateStore is a base class and is shared by a lot of implementations. I found it cleaner this way.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
aplex: ok, what would the the actual value they need to set? Looks like its is computed

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
aplex: So is this length problem only present for dag store? I would expect that if the flow name is part of the key, it will affect other state stores as well. I see that we have at least a "job_status" state store in shared Gobblin service. And if the new state store is created, it will also hit this problem.

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlJobStatusStateStore.java
arjun4084346: Ok, so what should a test verify?

aplex: The test can write the state for the flow with max length, and max expected table length, and then try to read it.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
arjun4084346: Yes, we have job_status store and flow spec store besides dag store.
job_status store is also a MysqlStateStore like this dag store. Spec store is different in its own.
So I have created a new method getCreateJobStateTableTemplate that new stores should override to enforce their own limits.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
arjun4084346: Yes, it is computed. I will work with SREs to do a one time manual update.

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/MysqlJobStatusStateStore.java
arjun4084346: FYI, I added a test in other module to test long names when creating spec with invalid names.

gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
```diff
@@ -132,4 +132,13 @@
   // Group Membership authentication service
   public static final String GROUP_OWNERSHIP_SERVICE_CLASS = GOBBLIN_SERVICE_PREFIX + "groupOwnershipService.class";
   public static final String DEFAULT_GROUP_OWNERSHIP_SERVICE = "org.apache.gobblin.service.NoopGroupOwnershipService";
+
+  public static final int MAX_FLOW_NAME_LENGTH = 128; // defined in FlowId.pdl
+  public static final int MAX_FLOW_GROUP_LENGTH = 128; // defined in FlowId.pdl
+  public static final int MAX_JOB_NAME_LENGTH = 374;
```
arjun4084346, line 138: @aplex @sv2000 will add a check in compiler to not produce job names more than this limit.

aplex (approved):

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import org.apache.gobblin.configuration.State;
+import org.apache.gobblin.metastore.MysqlStateStore;
+import org.apache.gobblin.service.ServiceConfigKeys;
+
+
+@Slf4j
+/**
+ * An implementation of {@link MysqlStateStore} backed by MySQL to store Dag.
```
sv2000, line 32: doesn't the MySqlDagStateStore exist for this reason? Why are we creating more classes which seem to have similar APIs?

arjun4084346: In `MySqlDagStateStore` we cannot change the `createTable` statement, which now I am able to do by overriding `getCreateJobStateTableTemplate`. Check out the other comment and replies for similar discussion.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
Will-Lo: Is there ever a scenario where we want to use the old dag state store?

Will-Lo (commented):
Would this cause any issues for flows that are not persisted in the dag state store? i.e. services that do not use the DagManager.

gobblin-metastore/src/main/java/org/apache/gobblin/MysqlDagStore.java
arjun4084346: Both the dag stores are actually the same. New store extends the old store overriding one method here https://github.com/apache/gobblin/pull/3298/files/d45e2c83eab8b6420fc90162ef27f1e3f8d0a9c9#diff-e44b7b6bb150dd599827d36ca60c5a08e31e406013cfb22d3b363f794899f4e8R52

arjun4084346:
> Would this cause any issues for flows that are not persisted in the dag state store? i.e. services that do not use the DagManager.

No, this change is equivalent to overriding a method in the old dag store. The services which do not use DagManager are not effected. Those which use DagManager are also not effected

arjun4084346:
@sv2000  I updated the PR description, hope it helps.

Will-Lo (approved):

arjun4084346:
@sv2000  please review/merge

sv2000 (commented):
+1.

PR #3299, opened 2021-06-05
[GOBBLIN-1457] Add automatic troubleshooter to Gobblin service

aplex:
In previous commits, we've added automatic troubleshooting to Gobblin
Azkaban jobs, and here we will collect and expose discovered issues
in Gobblin service.

Initial implementation will store issues for a limited number of jobs in
memory, and future commits will add persistence.


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/IssueEventBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+    });
+    return issueEventBuilder;
+  }
+
+  public static Issue getIssueFromEvent(GobblinTrackingEvent event) {
+    String serializedIssue = event.getMetadata().get(METADATA_ISSUE);
+    return GsonUtils.GSON_WITH_DATE_HANDLING.fromJson(serializedIssue, Issue.class);
+  }
+
+  public GobblinTrackingEvent build() {
```
umustafi, line 89: previously we never returned GobblinTrackingEvents for exception/error propogation right?

gobblin-runtime/src/test/java/org/apache/gobblin/runtime/troubleshooter/InMemoryIssueRepositoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    repository.put(getTestIssue("first", "code1"));
+    repository.put(getTestIssue("second", "code2"));
+    repository.put(getTestIssue("second-2", "code2"));
+    repository.put(getTestIssue("second-3", "code2"));
+
+    List<Issue> issues = repository.getAll();
+    assertEquals(2, issues.size());
+    assertTrue(issues.stream().anyMatch(i -> i.getCode().equals("code1")));
+    assertTrue(issues.stream().anyMatch(i -> i.getCode().equals("code2")));
```
umustafi, line 92: Does this account for a case where we have multiple issues with the same code that originated from two different places or does this usually map to propagated errors?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/IssueEventBuilder.java
aplex: We have a "job failed" GTE, but it only contains one latest exception. GTE with issues will also capture warnings and other errors that happened in Azkaban Job or Mapper.

gobblin-runtime/src/test/java/org/apache/gobblin/runtime/troubleshooter/InMemoryIssueRepositoryTest.java
aplex: For each job, we store only the first issue with a specific code. So, if we had 1000 "AccessDenied" errors, we'll only capture the first one and discard the rest. This should ensure that user is not overwhelmed with repeated errors.

aplex:
@arjun4084346 , @Will-Lo , can you take a look?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
```diff
@@ -168,6 +174,24 @@ public static FlowExecution convertFlowStatus(org.apache.gobblin.service.monitor
         .setJobStatuses(jobStatusArray);
   }
 
+  private static org.apache.gobblin.service.Issue convertIssue(Issue issues) {
```
Will-Lo, line 177: Add a javadoc that this is to convert a runtime issue to a gobblin service reported issue?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/InMemoryMultiContextIssueRepository.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Stores issues from multiple jobs, flows or other contexts in memory.
+ *
+ * To limit the memory consumption, it will keep only the last {@link #MAX_CONTEXT_COUNT} contexts,
+ * and older ones will be discarded.
+ * */
+@Singleton
+public class InMemoryMultiContextIssueRepository implements MultiContextIssueRepository {
+  public static final int DEFAULT_MAX_CONTEXT_COUNT = 100;
```
Will-Lo, line 41: If there is a long running job that emits a large number of failures, is it possible that this number is too high?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/orchestration/Orchestrator.java
```diff
@@ -104,8 +104,8 @@
   private Map<String, FlowCompiledState> flowGauges = Maps.newHashMap();
 
 
-  public Orchestrator(Config config, FlowStatusGenerator flowStatusGenerator, Optional<TopologyCatalog> topologyCatalog,
-      Optional<DagManager> dagManager, Optional<Logger> log, boolean instrumentationEnabled) {
+  public Orchestrator(Config config, Optional<TopologyCatalog> topologyCatalog, Optional<DagManager> dagManager, Optional<Logger> log,
```
Will-Lo, line 107: Was this file modified by Intellij autoformatter?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/InMemoryMultiContextIssueRepository.java
jack-moseley: So just curious, are we planning to use a much larger value than 100? Because I can imagine if one job of a flow runs, the flow takes several hours to finish, then a user queries the status, it's possible that 100 other GaaS jobs have already run in that time and then the issue information will be missing for the early jobs of the flow.

aplex: we'll use persistent issue storage in the next PR - https://github.com/apache/gobblin/pull/3327 that will store more issues in DB. In-memory is a fallback implementation in case we want to reduce the DB load for some reason.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/Issue.pdl
```diff
@@ -0,0 +1,46 @@
...
+ * Issue describes a specific unique problem in the job or application.
+ *
+ * Issue can be generated from log entries, health checks, and other places.
+ */
+record Issue {
+
+  /**
+   * Time when the issue have occured
+   */
+  time: Timestamp
```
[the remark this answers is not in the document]
aplex: It's actually unclear what "long" means for Time. In the most common interpretation, unix timestamp is a number of seconds since 1970, but in our API it is milliseconds. Having a type reference with a description of what it means should be more clear. Timestamp type comment says that it is in milliseconds.

Personally, I think we should have used ISO format like "2015-04-14T11:07:36.639Z" that is both human- and computer- readable, but it's too late to change it since other parts of API use timestamps.  https://stackoverflow.com/a/29626123

jack-moseley (approved):

sv2000 (approved):
+1.

PR #3300, opened 2021-06-06
[GOBBLIN-1460] Fix NPE in DagManager cleanup

aplex:
Under certain conditions, we can get a null list when checking if
dag has running job. This is a quickfix to address that.

https://issues.apache.org/jira/browse/GOBBLIN-1460


sv2000:
@aplex List<DagNode>> instead of LinkedList?

sv2000 (approved):
+1.

PR #3302, opened 2021-06-09
GOBBLIN-1462: Ensure FsSpecConsumer handles config keys which are prefixes of other keys

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1462

### Description
Currently, the FsSpecConsumer fails when building a JobSpec if the configuration contains keys which are prefixes of other keys, which is not allowed in typesafe config. In practice, this constraint may be violated particularly when loading system properties which are outside application's control. To ensure these properties are safely handled, the FsSpecConsumer should use the proertiesToConfig() method in PropertiesUtils which adds a suffix to the root properties.  

### Tests
Modified unit tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
@@ -45,6 +45,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.apache.gobblin.runtime.job_spec.AvroJobSpec;
 import org.apache.gobblin.util.CompletedFuture;
+import org.apache.gobblin.util.ConfigUtils;
 import org.apache.gobblin.util.filters.HiddenFilter;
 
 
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
```diff
@@ -112,8 +113,11 @@ public FsSpecConsumer(@Nullable FileSystem fs, Config config) {
         JobSpec.Builder jobSpecBuilder = new JobSpec.Builder(avroJobSpec.getUri());
         Properties props = new Properties();
         props.putAll(avroJobSpec.getProperties());
-        jobSpecBuilder.withJobCatalogURI(avroJobSpec.getUri()).withVersion(avroJobSpec.getVersion())
-            .withDescription(avroJobSpec.getDescription()).withConfigAsProperties(props);
+        jobSpecBuilder.withJobCatalogURI(avroJobSpec.getUri())
+            .withVersion(avroJobSpec.getVersion())
+            .withDescription(avroJobSpec.getDescription())
+            .withConfigAsProperties(props)
```
ZihanLi58, line 119: Do we still need this line?

sv2000: Yes, we will still need this line. The only change is we are now explicit about how the config member variable of JobSpec is set. Previously, it defaults to using ConfigUtils#propertiesToTypedConfig() which throws an exception when it encounters prefixed keys.

ZihanLi58 (approved):
+1, LGTM

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/FsSpecConsumer.java
autumnust: This API is quite confusing ... But I get the point why both will be needed.

autumnust (approved):
+1 LGTM

PR #3312, opened 2021-06-14
[GOBBLIN-1473] Fix ConcurrentModificationException when handling leadership change

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1473

### Description
Prevously, `ConcurrentModificationException` could happen when switching from active to inactive in the scheduler, because specs were removed from `scheduledFlowSpecs` while iterating through it. This fixes it in this scenario by not removing during iteration and just clearing the map at the end (since every entry should get removed anyway).

### Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
@@ -19,7 +19,9 @@ package org.apache.gobblin.service.modules.scheduler;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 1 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -165,8 +165,9 @@ public void run() {
     } else {
       // Since we are going to change status to isActive=false, unschedule all flows
       for (Spec spec : this.scheduledFlowSpecs.values()) {
-        onDeleteSpec(spec.getUri(), spec.getVersion());
+        onDeleteSpec(spec.getUri(), spec.getVersion(), new Properties(), false);
```
aplex, line 168: Can we clone the values list instead? Then we will iterate on a copy of it that would not change when we remove entries from the map.

Overload with "removeFromMap" looks a bit strange, and future maintainers will have to spend some brain cycles to understand why it was added.

jack-moseley: Yeah that's true, that is much simpler. I was thinking that it may be expensive to clone all the specs, but we actually only need to clone the list, and it is okay if they are referencing the same Spec objects since we are not modifying those.

aplex (approved):

PR #3319, opened 2021-06-23
[GOBBLIN-1479] Add support to convert zeroDateTime to null when configured

lchik:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1479

### Description
In MySQL Connector 8, calling ResultSet.getString(int) on a record with a zero timestamp
returns the string "0000-00-00 00:00:00", even when the connection property
`zeroDateTimeBehavior=CONVERT_TO_NULL` is set.
This change is a workaround to return null for these timestamps when the property is set.

### Tests

### Commits

### DIFF
--- a/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/source/jdbc/JdbcExtractor.java
+++ b/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/source/jdbc/JdbcExtractor.java
@@ -959,6 +959,10 @@ public abstract class JdbcExtractor extends QueryBasedExtractor<JsonArray, JsonE
       int batchSize = this.workUnitState.getPropAsInt(ConfigurationKeys.SOURCE_QUERYBASED_FETCH_SIZE, 0);
       batchSize = (batchSize == 0 ? ConfigurationKeys.DEFAULT_SOURCE_FETCH_SIZE : batchSize);
 
+      String sourceConnProps = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_PROPERTIES);
+      boolean convertZeroDateTimeToNull = sourceConnProps != null
+          && sourceConnProps.contains("zeroDateTimeBehavior=CONVERT_TO_NULL");
+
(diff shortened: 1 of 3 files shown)


gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/source/jdbc/JdbcExtractor.java
```diff
@@ -959,6 +959,10 @@ public long getCount(CommandOutput<?, ?> response) throws RecordCountException {
       int batchSize = this.workUnitState.getPropAsInt(ConfigurationKeys.SOURCE_QUERYBASED_FETCH_SIZE, 0);
       batchSize = (batchSize == 0 ? ConfigurationKeys.DEFAULT_SOURCE_FETCH_SIZE : batchSize);
 
+      String sourceConnProps = this.workUnitState.getProp(ConfigurationKeys.SOURCE_CONN_PROPERTIES);
+      boolean convertZeroDateTimeToNull = sourceConnProps != null
```
Will-Lo, line 963: I think here we can generalize this a bit more rather than keep it only to the zeroDateTime behavior, to support the other zeroDateTime properties if there are any users. We can check that sourceConnProps contains zeroDateTimeBehavior, then parse it as timestamp and return the result. This way it can handle the `ROUND` case

Will-Lo (changes requested):

Will-Lo (approved):

Will-Lo:
Thanks for the fix! LGTM

autumnust (changes requested):
Thanks for fixing, the change itself LGTM. 

Since we have `org.apache.gobblin.source.jdbc.JdbcExtractorTest`, can we add a unit test for the change we made here?

autumnust (approved):
LGTM !

autumnust:
Is the failing tests relevant ? 
Seems this is failing: ```:gobblin-rest-service:gobblin-rest-api:generateDataTemplate```

PR #3323, opened 2021-06-25
GOBBLIN-1483: Handle null valued ConsumerRecords in Kafka Streaming Extractor

sv2000:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1483

### Description
Currently, null-valued ConsumerRecords returned by the Kafka consumer are passed down the construct chain and can result in irrecoverable failures. Null valued records should be filtered out early in the pipeline so as to avoid such errors.

### Tests
Added unit tests

### Commits

### DIFF
--- a/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
+++ b/gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
@@ -87,7 +87,8 @@ import static org.apache.gobblin.source.extractor.extract.kafka.workunit.packer.
 public class KafkaStreamingExtractor<S> extends FlushingExtractor<S, DecodeableKafkaRecord> {
   public static final String DATASET_KEY = "dataset";
   public static final String DATASET_PARTITION_KEY = "datasetPartition";
-  private static final Long MAX_LOG_DECODING_ERRORS = 5L;
+  private static final Long MAX_LOG_ERRORS = 100L;
+
   private static final String KAFKA_EXTRACTOR_STATS_REPORTING_INTERVAL_MINUTES_KEY =
       "gobblin.kafka.extractor.statsReportingIntervalMinutes";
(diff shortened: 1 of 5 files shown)


gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
```diff
@@ -350,20 +351,33 @@ public S getSchema() {
     this.readStartTime = System.nanoTime();
     long fetchStartTime = System.nanoTime();
     try {
-      while (this.messageIterator == null || !this.messageIterator.hasNext()) {
-        Long currentTime = System.currentTimeMillis();
-        //it's time to flush, so break the while loop and directly return null
-        if ((currentTime - timeOfLastFlush) > this.flushIntervalMillis) {
-          return new FlushRecordEnvelope();
+      DecodeableKafkaRecord kafkaConsumerRecord;
+      while(true) {
```
autumnust, line 355: Why do we need this additional loop given there's already a `while` to check if `messageIterator.hasNext()` returns true?  Should it be enough to examine `.next()` each time and verify if that's actually a `null` value?

autumnust (changes requested):

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
sv2000: Each call to consume() returns an iterator over a new batch of records. Let's say the returned records are R1, R2, R3, and further, only R2 is null-valued. If we use a single while loop, we would end up calling consume() the moment we encounter R2, resulting in R3 being skipped. With the current implementation, R2 will be skipped because it is null and the next iteration of the while loop will correctly return R3, causing the outer while loop to be exited.

In theory, we can make it work with a single while loop and if conditions inside the while loop to handle the null-valued records as a special case. I am not sure if it would add more clarity than what the current implementation does.

autumnust (approved):
+1

PR #3324, opened 2021-06-28
[GOBBLIN-1485]Enable feature to get schema from writer schema when do hive registration

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1485

### Description
1. enable feature to get schema from writer schema instead of listing directory to get latest schema. 
2. When determine whether there is schema change, use schema version to compare to avoid special character issue. 

### Tests
Run test on parallel pipeline

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
@@ -149,9 +149,9 @@ public class HiveRegistrationUnitComparator<T extends HiveRegistrationUnitCompar
   private State extractSchemaVersion(State state) {
     State newState = new State(state);
     String schemaFromState = state.getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
-    if (Strings.isNullOrEmpty(schemaFromState)) {
+    if (!Strings.isNullOrEmpty(schemaFromState)) {
       String schemaVersion = AvroUtils.getSchemaCreationTime(new Schema.Parser().parse(schemaFromState));
-      if (Strings.isNullOrEmpty(schemaVersion)) {
+      if (!Strings.isNullOrEmpty(schemaVersion)) {
          newState.removeProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
          newState.setProp("schema.creationTime", schemaVersion);
       }
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
@@ -70,6 +70,7 @@ import org.apache.hadoop.hive.serde2.avro.AvroSerdeUtils;
  */
 @Alpha
 public class HiveRegistrationUnitComparator<T extends HiveRegistrationUnitComparator<?>> {
+  private static String SCHEMA_CREATION_TIME = "schema.creationTime";
 
   protected final HiveRegistrationUnit existingUnit;
   protected final HiveRegistrationUnit newUnit;
@@ -147,13 +148,15 @@ public class HiveRegistrationUnitComparator<T extends HiveRegistrationUnitCompar
   }
 
   private State extractSchemaVersion(State state) {
+    //FIXME: This is a temp fix for special character in schema string, need to investigate the root
+    //cause of why we see different encoding here and have a permanent fix for this
     State newState = new State(state);
     String schemaFromState = state.getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
     if (!Strings.isNullOrEmpty(schemaFromState)) {
       String schemaVersion = AvroUtils.getSchemaCreationTime(new Schema.Parser().parse(schemaFromState));
       if (!Strings.isNullOrEmpty(schemaVersion)) {
          newState.removeProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
-         newState.setProp("schema.creationTime", schemaVersion);
+         newState.setProp(SCHEMA_CREATION_TIME, schemaVersion);
       }
     }
     return newState;
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.hive;
 
+import com.google.common.base.Strings;
 import java.util.Set;
 
 import org.apache.avro.Schema;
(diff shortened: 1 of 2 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
```diff
@@ -142,12 +145,24 @@ public T compareIsStoredAsSubDirs() {
     return (T) this;
   }
 
+  private State extractSchemaVersion(State state) {
+    State newState = new State(state);
+    String schemaFromState = state.getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
+    if (schemaFromState != null && !schemaFromState.isEmpty()) {
+      String schemaVersion = AvroUtils.getSchemaCreationTime(new Schema.Parser().parse(schemaFromState));
+      if (schemaVersion != null && !schemaVersion.isEmpty()) {
```
sv2000, line 153: can we re-written using Strings.isNullOrEmpty().

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
```diff
@@ -142,12 +145,24 @@ public T compareIsStoredAsSubDirs() {
     return (T) this;
   }
 
+  private State extractSchemaVersion(State state) {
+    State newState = new State(state);
+    String schemaFromState = state.getProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
+    if (schemaFromState != null && !schemaFromState.isEmpty()) {
```
sv2000, line 151: can be re-written using Strings.isNullOrEmpty()

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
```diff
@@ -142,12 +145,24 @@ public T compareIsStoredAsSubDirs() {
...
+      String schemaVersion = AvroUtils.getSchemaCreationTime(new Schema.Parser().parse(schemaFromState));
+      if (schemaVersion != null && !schemaVersion.isEmpty()) {
+         newState.removeProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
```
sv2000, line 154: Why do we remove this property?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/orc/HiveOrcSerDeManager.java
```diff
@@ -75,12 +78,17 @@
...
   public static final String INPUT_FORMAT_CLASS_KEY = "hiveOrcSerdeManager.inputFormatClass";
   public static final String DEFAULT_INPUT_FORMAT_CLASS = OrcInputFormat.class.getName();
+  public static final String WRITER_LATEST_SCHEMA = "writer.latest.schema";
 
   public static final String OUTPUT_FORMAT_CLASS_KEY = "hiveOrcSerdeManager.outputFormatClass";
   public static final String DEFAULT_OUTPUT_FORMAT_CLASS = OrcOutputFormat.class.getName();
 
   public static final String HIVE_SPEC_SCHEMA_READING_TIMER = "hiveOrcSerdeManager.schemaReadTimer";
 
+  public static final String HIVE_SPEC_SCHEMA_FROM_WRITER = "hiveOrcSerdeManager.getSchemaFromWriterSchema";
```
sv2000, line 88: Do we need to introduce a config? Can we always look for writer schema first and then default to file schema if it doesn't exist?

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/orc/HiveOrcSerDeManager.java
```diff
@@ -264,7 +272,18 @@ private void addSchemaProperties(Path path, HiveRegistrationUnit hiveUnit)
    *
    */
   protected void addSchemaPropertiesHelper(Path path, HiveRegistrationUnit hiveUnit) throws IOException {
-    TypeInfo schema = getSchemaFromLatestFile(path, this.fs);
+    TypeInfo schema;
+    if(props.getPropAsBoolean(HIVE_SPEC_SCHEMA_FROM_WRITER, DEFAULT_HIVE_SPEC_SCHEMA_FROM_WRITER)) {
+      try {
+        Preconditions.checkArgument(props.contains(WRITER_LATEST_SCHEMA));
+        Schema avroSchema = new Schema.Parser().parse(props.getProp(WRITER_LATEST_SCHEMA));
+        schema = TypeInfoUtils.getTypeInfoFromObjectInspector(new AvroObjectInspectorGenerator(avroSchema).getObjectInspector());
```
sv2000, line 280: There is a Avro->Orc schema converter utility method in AvroOrcSchemaConverter class. Can we use that, given that it is being used by GobblinOrcWriter to create the ORC file?

sv2000: The schema converter returns a TypeDescription object. Not sure if it is easy to go from TypeDescription to TypeInfo.

sv2000 (changes requested):

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/HiveRegistrationUnitComparator.java
ZihanLi58: The whole schema may contains special character, so when comparing whether we need to update the table, we remove this prop, but it's a new state object, so we are not removing this prop from the hive unit.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/orc/HiveOrcSerDeManager.java
ZihanLi58: There are some use case that we have writer schema set but it's not the schema for this topic. i.e. in metadata pipeline, writer schema is the schema for GMCE. If we enable this feature for all, we'll hit issue in this case.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/orc/HiveOrcSerDeManager.java
ZihanLi58: Yeah I was trying to do that. Several reasons here: 
1. AvroOrcSchemaConverter is now defined in gobblin-orc module, I don't think it make sense for us to introduce new dependency for hive registration module.
2.  It's doable to transfer TypeDescription to TypeInfo, but it's the same way that we need to use OrcUtils to create one objectInspector and get typeInfo there. As we are using the writer schema to get orcSchema, I think the two results should be the same? I verified one table and it looks good to me.
What do you think?

sv2000 (changes requested):

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/orc/HiveOrcSerDeManager.java
sv2000: Makes sense. Thanks for the explanation!

sv2000 (approved):
+1.

PR #3327, opened 2021-07-02
[GOBBLIN-1457] Add persistence for troubleshooter in Gobblin service

aplex:
Previously, Gobblin service kept the last few job issues in memory.
In this commit, we add MySql-based persistence for job issues.

We introduce Flyway-based migrations to Gobblin service DB, so that
we can adjust the table schemas in the future, and add new tables
in a consistent way.


gobblin-service/src/main/java/org/apache/gobblin/service/modules/db/ServiceDatabaseProviderImpl.java
```diff
@@ -0,0 +1,117 @@
...
+  public ServiceDatabaseProviderImpl(Configuration configuration) {
+    this.configuration = configuration;
+  }
+
+  public DataSource getDatasource() {
+    ensureDataSource();
+    return dataSource;
+  }
+
+  private synchronized void ensureDataSource() {
```
jack-moseley, line 54: For all other usages of mysql in gobblin and gaas code, we use `MysqlDataSourceFactory` (which calls `MysqlStateStore.newDataSource`) to create the datasource. Is the configuration for this one unique enough that it needs to be a separate thing, or that these configurations would not also be useful in that place?

If we do want to keep these separate, maybe we could at least reuse the validation query (which I just added to `MysqlStateStore.newDataSource`) so that we don't have the same query in two places.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/db/ServiceDatabaseManager.java
```diff
@@ -0,0 +1,60 @@
...
+  protected void startUp()
+      throws Exception {
+
+    Flyway flyway =
+        Flyway.configure().locations("classpath:org/apache/gobblin/service/db/migration").failOnMissingLocations(true)
+            .dataSource(databaseProvider.getDatasource()).load();
+
+    log.info("Ensuring service database is migrated to latest schema");
+    // Start the migration
+    flyway.migrate();
```
jack-moseley, line 52: Just curious, by using flyway, does this table have to be in a separate db from our other tables, or it's fine to have a flyway-managed table in the same db as other regular tables?

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/util/DatabaseJobHistoryStoreSchemaManager.java
```diff
@@ -54,9 +55,7 @@
   private final Flyway flyway;
 
   private DatabaseJobHistoryStoreSchemaManager(Properties properties) {
-    flyway = new Flyway();
-    flyway.configure(properties);
-    flyway.setClassLoader(this.getClass().getClassLoader());
+    flyway = Flyway.configure(this.getClass().getClassLoader()).configuration(properties).load();
```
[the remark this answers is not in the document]
aplex: Yes. We upgraded from version 3 to 7, and there were some API changes.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/db/ServiceDatabaseManager.java
aplex: It's ok to have other tables in flyway-managed DB as long as they don't collide in names. Flyway will basically execute the sql config scripts one by one, and that's it.

Although, a specific Flyway app shouldn't use the same database as other Flyway apps, as Flyway creates it's own state table that has a fixed name.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java
```diff
@@ -205,12 +209,22 @@ public void configure(Binder binder) {
...
 
-    binder.bind(MultiContextIssueRepository.class).to(InMemoryMultiContextIssueRepository.class);
+    binder.bind(ServiceDatabaseProvider.class).to(ServiceDatabaseProviderImpl.class);
+    binder.bind(ServiceDatabaseProviderImpl.Configuration.class);
+
+    binder.bind(ServiceDatabaseManager.class);
+
+    binder.bind(MultiContextIssueRepository.class)
+        .to(getClassByNameOrAlias(MultiContextIssueRepository.class, serviceConfig.getInnerConfig(),
+                                  ServiceConfigKeys.ISSUE_REPO_CLASS, MySqlMultiContextIssueRepository.class.getName()));
```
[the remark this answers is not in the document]
aplex: Updated to in-memory by default

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/troubleshooter/AutoTroubleshooterLogAppender.java
```diff
@@ -0,0 +1,169 @@
...
+
+import org.apache.gobblin.runtime.ThrowableWithErrorCode;
+
+
+/**
+ * Collects messages from log4j and converts them into issues that are used in {@link AutomaticTroubleshooter}.
+ */
+@Slf4j
+@ThreadSafe
+public class AutoTroubleshooterLogAppender extends AppenderSkeleton {
```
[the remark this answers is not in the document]
aplex: It was moved to a different module in one of the previous commits, and after rebase we got two of them - in the old and new place. I'll remove it here.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/db/ServiceDatabaseProviderImpl.java
aplex: I used it initially, but there are problems with configuring it. It expects settings like "state.store.db.user" and "state.store.db.table" which don't really make sense for the issues and other tables. What we have in service db is not limited to one table, and it's not a state store.

I'll put the query in utility class to be reused.

jack-moseley: Just noting, for other GaaS tables we separate them by using a config prefix like `gobblin.service.mysqlDagStateStore.state.store.db.table`. And we do also use `state.store.db.` prefixed configs even for the `MySqlSpecStore`, which is not an implementation of a state store. But yeah that's not really the best pattern, so I think it's fine to have this one separate. If anything maybe we should change the spec store to use this datasource instead at some point.

jack-moseley (approved):
LGTM

PR #3328, opened 2021-07-06
Initial investigation and changes for Avro 1.9.2 upgrade.

abhisheknath2011:
Avro 1.9 upgrade HI: https://jira01.corp.linkedin.com:8443/browse/DEPEND-15444
Avro upgrade wiki: https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Avro+Initiative+2021
Highlights from Avro 1.9 upgrade: https://godatadriven.com/blog/highlights-from-the-new-apache-avro-1-9-0-release/

Below are some of the high level changes in Avro 1.9 (based on above highlight and initial investigation)
1. Deprecation of Joda time librarty and usage of Java 8 library.
2. Move from Jackson 1.x to 2.9. As part of this Jackson classes are removed from the public API. All the public method dealing with older version of Jackson are removed. 
   Below are some of methods no longer available
   - getJsonProps of Scheam, Schema.Field is no longer public.
   - getJsonProp of Schema, Scheam.Field is no longer available
3. Changes related to Schema.Field default value. The "defaultValue" method of Schema.field is no longer available and is replaced by "defaultVal" method
4. Avro 1.9 is leaner now and all avro shaded google guava libararies are removed.
5. There are bunch of tests failing. Some of the tests are fixed and for some TODO list is added in the test itself.
6. There are still lot more unidentified failed tests.
7. To fix some of the tests is required to fix some of the existing Gobblin classes.
8. Some of the tests are failing due to Helix, Hadoop, Hive libraries are using older Jackson version.
9. Some of the tests are failing due to handling of NULL values in the schema.
10. This is a bigger change and breaking change and more analysis and discussion is required.


gobblin-core/src/main/java/org/apache/gobblin/converter/avro/JsonElementConversionFactory.java
```diff
@@ -617,9 +618,12 @@ private Schema buildRecordSchema(JsonSchema schema, WorkUnitState workUnit, Stri
         } catch (UnsupportedDateTypeException e) {
           throw new UnsupportedOperationException(e);
         }
-
-        Schema.Field fld = new Schema.Field(map.getColumnName(), fldSchema, map.getComment(),
-            map.isNullable() ? JsonNodeFactory.instance.nullNode() : null);
+        // TODO - Fix this. The "map" variable is of type JsonSchema (Gobblin internal
+        // implementation). For map.isNullable, AvroCompatibilityUtils.NULL_VALUE is passed. For
+        // non null, identify the exact change. As curently null is passed,
+        // JsonElementConversionFactoryTest is failing.
```
srramach, line 624: > As curently null is passed, JsonElementConversionFactoryTest is failing.

When map.isNullable() is false, even the previous code was passing null, so there doesn't seem to be any change in that respect. So why does the test fail now, but didn't before?

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/conversion/hive/avro/AvroSchemaManager.java
```diff
@@ -161,6 +161,12 @@ else if (serdeParameters != null && serdeParameters.containsKey(HiveAvroSerDeMan
         List<TypeInfo> typeInfos = fields.stream().map(fs -> TypeInfoUtils.getTypeInfoFromTypeString(fs.getType()))
             .collect(Collectors.toList());
         List<String> comments = fields.stream().map(fs -> fs.getComment()).collect(Collectors.toList());
+        // TODO: Fix this code. Error - the below method is deprecated in Schema.Field in avro 1.9
+        // java.lang.NoSuchMethodError: org.apache.avro.Schema$Field.<init>(Ljava/lang/String;Lorg/apache/avro/Schema;Ljava/lang/String;Lorg/codehaus/jackson/JsonNode;)V
+        // at org.apache.hadoop.hive.serde2.avro.TypeInfoToSchema.createAvroField(TypeInfoToSchema.java:76)
+        // at org.apache.hadoop.hive.serde2.avro.TypeInfoToSchema.convert(TypeInfoToSchema.java:61)
+        // TypeInfoToSchema is from hive-serde-1.0.1 and uses avro 1.7. The latest version 3.1.2
+        // still uses avro 1.7
```
srramach, line 169: This is a problem, but maybe we can copy the TypeInfoToSchema class and customize it so that it uses the compat helper. Not great, but I think the blast radius will be small and manageable.

gobblin-utility/src/main/java/org/apache/gobblin/util/AvroCompatibilityUtils.java
```diff
@@ -0,0 +1,148 @@
...
+import org.codehaus.jackson.node.LongNode;
+import org.codehaus.jackson.node.NullNode;
+import org.codehaus.jackson.node.TextNode;
+
+
+/**
+ * The purpose of this helper class is to make some avro API portable across difference versions of avro.
+ * Once all the consumers of Gobblin core library is migrated to Avro 1.9.2, this class can be
+ * removed.
+ * See also {@link com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper}
```
srramach, line 40: Why not use the AvroComaptibilityHelper directly? Why make a wrapper around it? I am sure you have good reasons; could you please explain some of them?

PR #3329, opened 2021-07-14
[GOBBLIN-1484]Make Gobblin metadata writer be able to support schema source DB

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1484

### Description
Sometime we need the avro schema in place even for ORC tables, we have that information in ingestion job but not for compaction job. And it's hard to get the avro schema from orc file itself, so we want to support schema source db, so that we can fetch the schema from source db where ingestion job registers to.

### Tests
unit test

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
@@ -18,6 +18,7 @@
 package org.apache.gobblin.hive.writer;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Lists;
@@ -40,6 +41,7 @@ import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.data.management.copy.hive.WhitelistBlacklist;
 import org.apache.gobblin.hive.HiveRegister;
 import org.apache.gobblin.hive.HiveTable;
+import org.apache.gobblin.hive.metastore.HiveMetaStoreBasedRegister;
 import org.apache.gobblin.hive.spec.HiveSpec;
 import org.apache.gobblin.metadata.GobblinMetadataChangeEvent;
 import org.apache.gobblin.metadata.SchemaSource;
@@ -197,9 +199,9 @@ public class HiveMetadataWriter implements MetadataWriter {
                       MetadataWriter.DEFAULT_CACHE_EXPIRING_TIME), TimeUnit.HOURS)
                   .build());
           HiveSpec existedSpec = hiveSpecCache.getIfPresent(partitionValue);
+          schemaUpdateHelper(gmce, spec, topicName, tableKey);
           if (existedSpec != null) {
             //if existedSpec is not null, it means we already registered this partition, so check whether we need to update the table/partition
-            schemaUpdateHelper(gmce, spec, topicName, tableKey);
             if ((this.hiveRegister.needToUpdateTable(existedSpec.getTable(), spec.getTable())) || (
                 spec.getPartition().isPresent() && this.hiveRegister.needToUpdatePartition(
                     existedSpec.getPartition().get(), spec.getPartition().get()))) {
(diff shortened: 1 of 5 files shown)


gobblin-iceberg/src/test/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriterTest.java
```diff
@@ -278,9 +297,9 @@ public TestHiveRegistrationPolicy(State props) throws IOException {
       }
       return tables;
     }
-    protected Iterable<String> getDatabaseNames(Path path) {
+    /*protected Iterable<String> getDatabaseNames(Path path) {
```
sv2000, line 300: Remove the commented out block.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -262,13 +264,34 @@ private void schemaUpdateHelper(GobblinMetadataChangeEvent gmce, HiveSpec spec,
           }
         }
       }
+    } else if (gmce.getRegistrationProperties().containsKey(HiveMetaStoreBasedRegister.SCHEMA_SOURCE_DB)
+        && !gmce.getRegistrationProperties().get(HiveMetaStoreBasedRegister.SCHEMA_SOURCE_DB).equals(spec.getTable().getDbName())) {
+      // If schema source is NONE and schema source db is set, we will directly update the schema to source db schema
+      String schemaSourceDb = gmce.getRegistrationProperties().get(HiveMetaStoreBasedRegister.SCHEMA_SOURCE_DB);
+      try {
+        String sourceSchema = fetchSchemaFromTable(schemaSourceDb, spec.getTable().getTableName());
```
sv2000, line 272: Should we cache the sourceSchema to avoid repeated lookups?

sv2000 (changes requested):

autumnust:
No additional comments beyond what @sv2000  mentioned above. But a broader question to deal with this kind of feature is: What should be the right way to specify "lineage" of schema between different tables? Is setting source.db in GMCE a right approach (which means you need to set this in a specific application's GMCE if you expect the application itself doesn't carry the schema during runtime, for example compaction), or is there something broader missing in the overall picture.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: Update the code to try to use the existing schema map to get the schema. But if there is no schema in schema map, which means we don't register the source table and not maintain the latest schema, we will still fetch from hive to make sure we can always have the latest schema

gobblin-iceberg/src/test/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriterTest.java
ZihanLi58: fixed.

ZihanLi58:
> No additional comments beyond what @sv2000 mentioned above. But a broader question to deal with this kind of feature is: What should be the right way to specify "lineage" of schema between different tables? Is setting source.db in GMCE a right approach (which means you need to set this in a specific application's GMCE if you expect the application itself doesn't carry the schema during runtime, for example compaction), or is there something broader missing in the overall picture.

ZihanLi58:
> No additional comments beyond what @sv2000 mentioned above. But a broader question to deal with this kind of feature is: What should be the right way to specify "lineage" of schema between different tables? Is setting source.db in GMCE a right approach (which means you need to set this in a specific application's GMCE if you expect the application itself doesn't carry the schema during runtime, for example compaction), or is there something broader missing in the overall picture.

ZihanLi58:
> No additional comments beyond what @sv2000 mentioned above. But a broader question to deal with this kind of feature is: What should be the right way to specify "lineage" of schema between different tables? Is setting source.db in GMCE a right approach (which means you need to set this in a specific application's GMCE if you expect the application itself doesn't carry the schema during runtime, for example compaction), or is there something broader missing in the overall picture.
Yeah I do think there is something miss broader. Ideally, we should have a source of truth relationship graph between each table, so that when we see schema update, we can modify all tables using that schema. Leveraging config store is doable, but will introduce more complexity in manage the relationships. One better way is that we can use datahub for this case, but this will need more design. As for now, I would like to support source db in the GMCE itself to make it feasible for OSS user as well

sv2000 (approved):
+1.

PR #3330, opened 2021-07-14
[GOBBLIN-1489] Add Azkaban Server Name and non transient error config keys

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1489

### Description
define AZKABAN_SERVER_NAME in ConfigurationKeys.java which are otherwise defined in various places. 
add NON_TRANSIENT_ERROR_MESSAGES which can be used to separate retry-able exceptions from fatal ones

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -946,6 +946,7 @@ public class ConfigurationKeys {
   public static final String AZKABAN_JOB_URL = "azkaban.link.job.url";
   public static final String AZKABAN_JOB_EXEC_URL = "azkaban.link.jobexec.url";
   public static final String AZKABAN_WEBSERVERHOST = "azkaban.webserverhost";
+  public static final String AZKABAN_SERVER_NAME = "azkaban.server.name";
 
   /**
    * Hive registration properties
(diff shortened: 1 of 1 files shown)


gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -969,6 +970,7 @@
   public static final Charset DEFAULT_CHARSET_ENCODING = Charsets.UTF_8;
   public static final String TEST_HARNESS_LAUNCHER_IMPL = "gobblin.testharness.launcher.impl";
   public static final int PERMISSION_PARSING_RADIX = 8;
+  public static final String NON_TRANSIENT_ERROR_MESSAGES = "non.transient.error.messages";
```
aplex, line 973: Is this a new setting, or this property comes from the external system?

If it's a new one, I suggest to prefix it with "gobblin." , and consider a more descriptive name. Thing separate by dots usually mean class/property names. So in this case there will be class "non", class "transient", "error" and it will have a property "message". Something like "gobblin.errorMessages.logNonTransientErrors" could be more meaningful

aplex: Also, you can add a purpose of this setting as a javadoc comment. I see that you included it in the commit message, but having it in code can be beneficial too, if people look at the list of settings and try to find out what they mean.

arjun4084346: It is a new one and describes a comma separated list of non transient error messages that can come in a gobblin job.

aplex: Are those like exception classes or parts of the text in exception message? Would be good to add an example in javadoc alongside with intended use case.

arjun4084346: It can be either. Sometimes, the exception class is very generic, e.g. IOException, in that case this config can include error message also.
It all depends how to value of this config is being used. in li-gobblin, I compare it with exception.getMessage() which contains both class and message.

aplex (approved):

PR #3331, opened 2021-07-15
[GOBBLIN-1490] Make metadata pipeline to support consume GMCE emitted from different cluster

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1490

### Description
In some use case, we may run data pipeline on different clusters but those clusters share the same file system. And in those case, we will want the metadata pipeline to be able to consume all the records even they are running on different clusters. Also make the cluster name for GMCE be configurable so that we don't need to change the accepted cluster names when cluster name itself changes 

### Tests
unit test

### Commits


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/GobblinMCEWriter.java
```diff
@@ -84,6 +86,7 @@
   List<MetadataWriter> metadataWriters;
   Map<String, OperationType> tableOperationTypeMap;
   Map<String, OperationType> datasetOperationTypeMap;
+  Set<String> acceptClusters;
```
autumnust, line 89: acceptedClusters?

autumnust (changes requested):

PR #3333, opened 2021-07-16
[GOBBLIN-1491] Add option to mark up/down D2 servers for gobblin service on leadership change

jack-moseley:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1491

### Description

Currently, D2 servers are started on all hosts at the time of service start. This change will allow using D2 delayed start, then marking up/down the servers when a host becomes the master/slave respectively. This means only the master will be announced to D2, so all requests will be routed to it.

The logic for markup is linkedin specific so must be implemented in internal code.

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
@@ -35,6 +35,8 @@ public class ServiceConfigKeys {
   public static final String GOBBLIN_SERVICE_GIT_CONFIG_MONITOR_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "gitConfigMonitor.enabled";
   public static final String GOBBLIN_SERVICE_DAG_MANAGER_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "dagManager.enabled";
   public static final String GOBBLIN_SERVICE_JOB_STATUS_MONITOR_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "jobStatusMonitor.enabled";
+  // If true, will mark up/down d2 servers on leadership so that all requests will be routed to the leader node
+  public static final String GOBBLIN_SERVICE_D2_ONLY_ANNOUNCE_LEADER = GOBBLIN_SERVICE_PREFIX + "d2.onlyAnnounceLeader";
 
   // Helix / ServiceScheduler Keys
   public static final String HELIX_CLUSTER_NAME_KEY = GOBBLIN_SERVICE_PREFIX + "helix.cluster.name";
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/D2Announcer.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/D2Announcer.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.service.modules.core;
+
+
+/**
+ * Interface for marking up/down D2 servers on gobblin service startup. This is only required if using delayed announcement.
+ */
+public interface D2Announcer {
+  /**
+   * Mark up this host's D2 server
+   */
+  void markUpServer();
+
+  /**
+   * Mark down this host's D2 server
+   */
+  void markDownServer();
+}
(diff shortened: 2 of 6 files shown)


gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
```diff
@@ -35,6 +35,7 @@
   public static final String GOBBLIN_SERVICE_GIT_CONFIG_MONITOR_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "gitConfigMonitor.enabled";
   public static final String GOBBLIN_SERVICE_DAG_MANAGER_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "dagManager.enabled";
   public static final String GOBBLIN_SERVICE_JOB_STATUS_MONITOR_ENABLED_KEY = GOBBLIN_SERVICE_PREFIX + "jobStatusMonitor.enabled";
+  public static final String GOBBLIN_SERVICE_DELAYED_START = GOBBLIN_SERVICE_PREFIX + "d2.delayedStart";
```
aplex, line 38: Can we name this after intention, rather than implementation? Looks like what we want to configure here is "send d2 requests only to master node". You can also add a comment when this is needed - if d2 randomly balances requests across all servers, some of the write requests will fail.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
```diff
@@ -191,6 +191,7 @@
 
   protected Optional<HelixLeaderState> helixLeaderGauges;
 
+  public D2Announcer d2Announcer;
```
aplex, line 194: How is it initialized? If it is optional, and is going to be set through dependency injection, it needs  @Inject(optional = true) attribute.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/D2Announcer.java
```diff
@@ -0,0 +1,17 @@
...
+
+
+/**
+ * Interface for marking up/down D2 servers on gobblin service startup. This is only required if using delayed announcement.
+ */
+public interface D2Announcer {
+  /**
+   * Mark up D2 servers
+   */
+  void markUpServers();
```
aplex, line 11: Is it marking all servers as "up" or just the current one?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
jack-moseley: Right now I am just directly setting this field after the `GobblinServiceManager` is initialized.

How does the dependency injection work exactly? Like I see for the `FlowStatusGenerator` we call `binder.bind(FlowStatusGenerator.class);`, does that construct a `FlowStatusGenerator` and inject it, and if so how does it provide arguments to the constructor?

I needed to, in our internal code, construct a subclass of D2Announcer and provide it to `GobblinServiceManager`, I couldn't figure out how to do it in a different way.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/D2Announcer.java
jack-moseley: The API from D2 we will use gives a server list, even though I think there will just be one server per host in this case.

aplex: I suggest to double-check with D2 docs/code that it will actually only mark the current server. If it is going to mark all servers as down, that wouldn't help us with routing data to master node.

If it indeed just works with the current server, then it looks like a confusing API on D2 side. We don't have to carry it inside our code base, and can make it a bit more correct and meaningful in our context.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
aplex: For li-specific implementations, you can change LiGobblinServiceGuiceModule to override the defaults and inject private implementations. If your class depends on other interfaces registered in Guice, they will be injected into its constructor. Search for "@Inject" annotation on constructors in other classes to see how it works.

if the subclass of D2Announcer depends on some instances that you don't create yourself,  you can register such instances in Guice like this - https://github.com/apache/gobblin/blob/b5e32b158bd713c45caa5b4b8866f8a4c4fc3a29/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java#L104 , or create a provider for them/D2Announcer like this - https://github.com/apache/gobblin/blob/b5e32b158bd713c45caa5b4b8866f8a4c4fc3a29/gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceGuiceModule.java#L184

Here is more info on how Guice works - https://github.com/google/guice/wiki/Motivation

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/D2Announcer.java
jack-moseley: Well if I understand correctly it would be the D2 servers associated with the current host, I don't think it would be possible to markup/markdown D2 servers for other hosts.

But sure, since I think it should just be one server in our case I can change it.

gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
jack-moseley: Changed to `d2.onlyAnnounceLeader`

gobblin-service/src/main/java/org/apache/gobblin/service/modules/core/GobblinServiceManager.java
jack-moseley: Thanks for the explanation, was able to get it working through injection.

aplex (approved):
LGTM. Can you fix the build issues? I'll merge it after.

PR #3334, opened 2021-07-16
[GOBBLIN-1488] Added option to set perm group at table level

vikrambohra:
Dear Gobblin maintainers,

### JIRA
- [ ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title. For example, "[GOBBLIN-1488] My Gobblin PR"
    - https://issues.apache.org/jira/browse/GOBBLIN-1488

### Description
Allow option to change perm group at publisher dir leaf level. 
if publisherdir = /db/table then the perm group is set for path /table

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -494,7 +494,10 @@ public class ConfigurationKeys {
   public static final String DATA_PUBLISHER_REPLACE_FINAL_DIR = DATA_PUBLISHER_PREFIX + ".replace.final.dir";
   public static final String DATA_PUBLISHER_FINAL_NAME = DATA_PUBLISHER_PREFIX + ".final.name";
   public static final String DATA_PUBLISHER_OVERWRITE_ENABLED = DATA_PUBLISHER_PREFIX + ".overwrite.enabled";
-  // This property is used to specify the owner group of the data publisher final output directory
+  // @DATA_PUBLISHER_FINAL_DIR is the final publishing root directory
+  // @DATA_PUBLISHER_FINAL_DIR_GROUP is set at the leaf level (DATA_PUBLISHER_FINAL_DIR/EXTRACT/file.xxx) which is incorrect
+  // Use @DATA_PUBLISHER_OUTPUT_DIR_GROUP to set group at output dir level @DATA_PUBLISHER_FINAL_DIR/EXTRACT
+  @Deprecated
   public static final String DATA_PUBLISHER_FINAL_DIR_GROUP = DATA_PUBLISHER_PREFIX + ".final.dir.group";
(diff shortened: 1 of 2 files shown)


gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -496,6 +496,7 @@
   public static final String DATA_PUBLISHER_OVERWRITE_ENABLED = DATA_PUBLISHER_PREFIX + ".overwrite.enabled";
   // This property is used to specify the owner group of the data publisher final output directory
   public static final String DATA_PUBLISHER_FINAL_DIR_GROUP = DATA_PUBLISHER_PREFIX + ".final.dir.group";
+  public static final String DATA_PUBLISHER_OUTPUT_DIR_GROUP = DATA_PUBLISHER_PREFIX + ".output.dir.group";
```
aplex, line 499: What's the difference between "final" and "output" dir for publisher?

aplex (commented):
We also have "writer.group.name" and "publishe.final.dir.group". Do they not cover the needed use case, and we need extra flexibility here?

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
vikrambohra: Although the FINAL DIR and OUTPUT DIR are essentially the same dir (/db/table in case of databases) but the DATA_PUBLISHER_FINAL_DIR_GROUP is applied at the leaf level (/db/table/yyyy/mm/dd/hh) while DATA_PUBLISHER_OUTPUT_DIR_GROUP is applied at the table level (/db/table which is what we want)

vikrambohra: "writer.group.name" is applied to the actual data file (avro, orc) while "publisher.final.dir.group" only applied the group at leaf level (DATA_PUBLISHER_FINAL_DIR_GROUP)

aplex: This is confusing. If FINAL DIR and OUTPUT DIR are the same, why does the group apply to different folders?

Also, what group will apply to folder at intermediate level, like "/db/table/yyyy/mm" ?

aplex: Discussed over the chat. The summary is that the old setting works incorrectly, but even if it would work correctly, it would not be what we need.

From vbohra:
so the publisherOutputDir can be the following based on use case
- data.publisher.final.dir ( if data.publisher.appendExtractToFinalDir is set to false)
- data.publisher.final.dir/db/table ( if data.publisher.appendExtractToFinalDir is set to true and writer.file.path.type = namespace_table) [ THIS IS THE CASE FOR brooklin-etl ]
- data.publisher.final.dir/table ( if data.publisher.appendExtractToFinalDir is set to true and writer.file.path.type = tablename)
- and a default

aplex (approved):

PR #3335, opened 2021-07-21
[GOBBLIN-1492] Optimize flowspec keys on configToProperties

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1492

### Description
- Use String.intern() on FlowSpec keys so that duplicate strings are not created. Duplicate keys in configToProperties can lead about to high heap usage (5% of Gobblin as a Service)

### Tests

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
@@ -303,7 +303,11 @@ public class ConfigUtils {
    * values Strings. This implementation will try to recognize booleans and numbers. All keys are
    * treated as strings.*/
   private static Map<String, Object> guessPropertiesTypes(Map<Object, Object> srcProperties) {
+<<<<<<< HEAD
     Map<String, Object> res = Maps.newHashMapWithExpectedSize(srcProperties.size());
+=======
+    Map<String, Object> res = new HashMap<>(srcProperties.size());
+>>>>>>> ff249aa9c (Restore accidentally deleted line)
     for (Map.Entry<Object, Object> prop : srcProperties.entrySet()) {
       Object value = prop.getValue();
       if (null != value && value instanceof String && !Strings.isNullOrEmpty(value.toString())) {
@@ -562,6 +566,7 @@ public class ConfigUtils {
 
     PasswordManager passwordManager = PasswordManager.getInstance(configToProperties(config));
     Map<String, String> tmpMap = Maps.newHashMapWithExpectedSize(encryptedConfig.entrySet().size());
+
     for (Map.Entry<String, ConfigValue> entry : encryptedConfig.entrySet()) {
       String val = entry.getValue().unwrapped().toString();
       val = passwordManager.readPassword(val);
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
@@ -303,11 +303,7 @@ public class ConfigUtils {
    * values Strings. This implementation will try to recognize booleans and numbers. All keys are
    * treated as strings.*/
   private static Map<String, Object> guessPropertiesTypes(Map<Object, Object> srcProperties) {
-<<<<<<< HEAD
     Map<String, Object> res = Maps.newHashMapWithExpectedSize(srcProperties.size());
-=======
-    Map<String, Object> res = new HashMap<>(srcProperties.size());
->>>>>>> ff249aa9c (Restore accidentally deleted line)
     for (Map.Entry<Object, Object> prop : srcProperties.entrySet()) {
       Object value = prop.getValue();
       if (null != value && value instanceof String && !Strings.isNullOrEmpty(value.toString())) {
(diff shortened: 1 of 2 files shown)


Will-Lo:
@aplex @arjun4084346 Please take a look

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
```diff
@@ -561,7 +562,6 @@ public static Config resolveEncrypted(Config config, Optional<String> encConfigP
     Config encryptedConfig = config.getConfig(encConfigPath.get());
 
     PasswordManager passwordManager = PasswordManager.getInstance(configToProperties(config));
-    Map<String, String> tmpMap = Maps.newHashMap();
```
arjun4084346, line 564: deleted by mistake?

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
```diff
@@ -116,7 +116,8 @@ public static Properties configToProperties(Config config, Optional<String> pref
       Config resolvedConfig = config.resolve();
       for (Map.Entry<String, ConfigValue> entry : resolvedConfig.entrySet()) {
         if (!prefix.isPresent() || entry.getKey().startsWith(prefix.get())) {
-          String propKey = desanitizeKey(entry.getKey());
+          // Intern the string so that constant keys are not duplicated to save memory
+          String propKey = desanitizeKey(entry.getKey()).intern();
```
arjun4084346, line 120: So is this the place where we populate configToProperties in FlowSpec? Should we `intern()` the values as well.Fliptop datapacks has duplicate values as well. I can give examples on slack because these are internal configs

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
Will-Lo: Oops yeah, that was supposed to be in the next PR but I missed the glob, will fix

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
Will-Lo: I would be scared to intern() the values, if unless we can guarantee that the values are acting immutable here. Otherwise they will point to the same object, and any config value that is changed will change across all uses.

aplex:
@Will-Lo , did you find that a lot of memory is consumed by duplicated strings created at this place?

By the way, you can also check out string de-duplication feature of GC. It's benefit is that it will work on all strings automatically, but the downside is that it happens only after several collections and introduces overhead by itself.

https://sauravomar01.medium.com/string-deduplication-of-g1-garbage-collector-in-java-8-cb8914a139f4

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
arjun4084346: propVal is a String and Strings are immutable, right? When a config value is changed, it does not change the value, it creates a new value in the underlying map and set the new value to that key, right?

Will-Lo: Ah you're right it would create a new one, I misunderstood some of the underlying risks of intern() 😄 . Doing some more reading, it does have some risks if there are a number of unique values though. Given that there are many values that are unique in key vals for configs, it might lead to other issues moreso than keys, like affecting the permanent memory pool size.

aplex: The exact behavior of intern() is not well document, and changed across java versions. Here is an article where a guy reverse-engineers it - http://java-performance.info/string-intern-in-java-6-7-8

umustafi: The article Alex cited says to "Use -XX:StringTableSize JVM parameter in Java 7 and 8 to set the string pool map size." Is that not needed here?

umustafi (approved):
If intern() feature is working as expected then LGTM

gobblin-utility/src/main/java/org/apache/gobblin/util/ConfigUtils.java
Will-Lo: @umustafi thanks for the callout, it defaults to ~60k strings which is likely more than enough for keys (as we have a fixed number of configuration keys that should not exceed that number) but I will add the argument `-XX:+PrintStringTableStatistics` so that I can see how many we're actually using. These arguments would be in the service deployable runtime args, so it wouldn't be included in this repository.

sv2000: @Will-Lo This method is invoked in a number of places and not limited to Gobblin service. One concern is that the number of keys may not be as small as you imagine. In production settings, Gobblin jobs could load system configs set by AZ/Hadoop or even the host where a job/container is running. These can be non-trivial in number. 60k may be sufficient. But what happens if the number of config keys exceed this value? Will the JVM just crash?

sv2000: It may be safer to introduce an overloaded method with an additional boolean flag as an argument, which when enabled interns the keys/values. FlowSpec builder can then call this overloaded method and the rest of the usage can remain as-is.

PR #3336, opened 2021-07-22
[GOBBLIN-1493] Report data copy's progress

umustafi:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1493

### Description
Progress reporting for a data copy will provide users with quantitative feedback on the progress of a data copy job as a percentage as well as an estimate of the time remaining for completion. This will update the existing job status endpoint to include the progress percentage and estimate of time left. 

This PR decorates each work unit with its size in bytes and uses that information to determine the progress of a copy while individual tasks are completed. It emits GobblinTrackingEvents containing the progress as a percentage intermittently and calculates the remaining time for the copy, storing both in the job status. 

### Tests

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -681,6 +681,9 @@ public class ConfigurationKeys {
   public static final String TASK_STATE_COLLECTOR_INTERVAL_SECONDS = "task.state.collector.interval.secs";
   public static final int DEFAULT_TASK_STATE_COLLECTOR_INTERVAL_SECONDS = 60;
   public static final String TASK_STATE_COLLECTOR_HANDLER_CLASS = "task.state.collector.handler.class";
+  public static final String REPORT_JOB_PROGRESS = "report.job.progress";
+  public static final boolean DEFAULT_REPORT_JOB_PROGRESS = false;
+  public static final double DEFAULT_PROGRESS_REPORTING_THRESHOLD = 0.05;
 
   /**
    * Set to true so that job still proceed if TaskStateCollectorService failed.
--- a/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
@@ -58,6 +58,10 @@ public class ServiceConfigKeys {
   public static final String GOBBLIN_SERVICE_FLOW_CATALOG_LOCAL_COMMIT = GOBBLIN_SERVICE_PREFIX + "flowCatalog.localCommit";
   public static final boolean DEFAULT_GOBBLIN_SERVICE_FLOW_CATALOG_LOCAL_COMMIT = true;
 
+  // Job Level Keys
+  public static final String WORK_UNIT_SIZE = GOBBLIN_SERVICE_PREFIX + "work.unit.size";
+  public static final String TOTAL_WORK_UNIT_SIZE = GOBBLIN_SERVICE_PREFIX + "total.work.unit.size";
+  public static final String TOTAL_WORK_UNIT_COUNT = GOBBLIN_SERVICE_PREFIX + "total.work.unit.count";
   /**
    * Directly use canonical class name here to avoid introducing additional dependency here.
    */
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
@@ -98,6 +98,7 @@ import org.apache.gobblin.util.request_allocation.PriorityIterableBasedRequestAl
 import org.apache.gobblin.util.request_allocation.RequestAllocator;
 import org.apache.gobblin.util.request_allocation.RequestAllocatorConfig;
 import org.apache.gobblin.util.request_allocation.RequestAllocatorUtils;
+import org.apache.gobblin.service.ServiceConfigKeys;
 
 
 /**
@@ -353,8 +354,9 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
 
         Extract extract = new Extract(Extract.TableType.SNAPSHOT_ONLY, CopyConfiguration.COPY_PREFIX, extractId);
         List<WorkUnit> workUnitsForPartition = Lists.newArrayList();
-        for (CopyEntity copyEntity : fileSet.getFiles()) {
 
+        long fileSize;
+        for (CopyEntity copyEntity : fileSet.getFiles()) {
           CopyableDatasetMetadata metadata = new CopyableDatasetMetadata(this.copyableDataset);
           CopyEntity.DatasetAndPartition datasetAndPartition = copyEntity.getDatasetAndPartition(metadata);
 
@@ -382,9 +384,18 @@ public class CopySource extends AbstractSource<String, FileAwareInputStream> {
           setWorkUnitWatermark(workUnit, watermarkGenerator, copyEntity);
           computeAndSetWorkUnitGuid(workUnit);
           addLineageInfo(copyEntity, workUnit);
-          if (copyEntity instanceof CopyableFile && DistcpFileSplitter.allowSplit(this.state, this.targetFs)) {
-            workUnitsForPartition.addAll(DistcpFileSplitter.splitFile((CopyableFile) copyEntity, workUnit, this.targetFs));
+          if (copyEntity instanceof CopyableFile) {
+            CopyableFile castedCopyEntity = (CopyableFile) copyEntity;
+            fileSize = castedCopyEntity.getFileStatus().getLen();
+            workUnit.setProp(ServiceConfigKeys.WORK_UNIT_SIZE, fileSize);
+            if (DistcpFileSplitter.allowSplit(this.state, this.targetFs)) {
+              workUnitsForPartition.addAll(DistcpFileSplitter.splitFile((CopyableFile) copyEntity, workUnit, this.targetFs));
+            } else {
+              workUnitsForPartition.add(workUnit);
+            }
           } else {
+            // Occurs when a job has local state files or post publishing steps that are not associated with a byte size
+            workUnit.setProp(ServiceConfigKeys.WORK_UNIT_SIZE, 0);
             workUnitsForPartition.add(workUnit);
           }
         }
--- a/gobblin-data-management/src/test/java/org/apache/gobblin/data/management/copy/TestCopyableDataset.java
+++ b/gobblin-data-management/src/test/java/org/apache/gobblin/data/management/copy/TestCopyableDataset.java
@@ -38,6 +38,7 @@ import org.apache.gobblin.dataset.FileSystemDataset;
 public class TestCopyableDataset implements CopyableDataset, FileSystemDataset {
 
   public static final int FILE_COUNT = 10;
+  public static final int FILE_LENGTH = 15;
   public static final String ORIGIN_PREFIX = "/test";
   public static final String DESTINATION_PREFIX = "/destination";
   public static final String RELATIVE_PREFIX = "/relative";
@@ -61,7 +62,7 @@ public class TestCopyableDataset implements CopyableDataset, FileSystemDataset {
     List<CopyEntity> files = Lists.newArrayList();
 
     for (int i = 0; i < FILE_COUNT; i++) {
-      FileStatus origin = new FileStatus(10, false, 0, 0, 0, new Path(this.datasetRoot, Integer.toString(i)));
+      FileStatus origin = new FileStatus(FILE_LENGTH, false, 0, 0, 0, new Path(this.datasetRoot, Integer.toString(i)));
       CopyableFile.Builder builder = CopyableFile
           .builder(FileSystem.getLocal(new Configuration()), origin, datasetRoot(), configuration).destinationOwnerAndPermission(OWNER_AND_PERMISSION).
           ancestorsOwnerAndPermission(Lists.newArrayList(OWNER_AND_PERMISSION)).checksum("checksum".getBytes());
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
@@ -19,4 +19,14 @@ record JobStatistics {
    * number of records processed in the last job execution
    */
   processedCount: long
+
+  /**
+   * job progress as a percentage (0-100)
+   */
+  jobProgress: int
+
+  /**
+   * estimate of time left until job completion
+   */
+  estimatedSecondsToCompletion: long
 }
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowstatuses.snapshot.json
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowstatuses.snapshot.json
@@ -180,6 +180,14 @@
                 "name" : "processedCount",
                 "type" : "long",
                 "doc" : "number of records processed in the last job execution"
+              }, {
+                "name" : "jobProgress",
+                "type" : "int",
+                "doc" : "job progress as a percentage (0-100)"
+              }, {
+                "name" : "estimatedSecondsToCompletion",
+                "type" : "long",
+                "doc" : "estimate of time left until job completion"
               } ]
             },
             "doc" : "Statistics from the job execution. The values may be updated during the run of a job."
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
@@ -16,6 +16,8 @@
  */
 package org.apache.gobblin.service;
 
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
@@ -140,6 +142,9 @@ public class FlowExecutionResourceLocalHandler implements FlowExecutionResourceH
 
       JobStatus jobStatus = new JobStatus();
 
+      Long timeLeft = estimateCopyTimeLeft(queriedJobStatus.getLastProgressEventTime(), queriedJobStatus.getStartTime(),
+          queriedJobStatus.getProgressPercentage());
+
       jobStatus.setFlowId(flowId)
           .setJobId(new JobId().setJobName(queriedJobStatus.getJobName())
               .setJobGroup(queriedJobStatus.getJobGroup()))
@@ -147,7 +152,9 @@ public class FlowExecutionResourceLocalHandler implements FlowExecutionResourceH
           .setExecutionStatistics(new JobStatistics()
               .setExecutionStartTime(queriedJobStatus.getStartTime())
               .setExecutionEndTime(queriedJobStatus.getEndTime())
-              .setProcessedCount(queriedJobStatus.getProcessedCount()))
+              .setProcessedCount(queriedJobStatus.getProcessedCount())
+              .setJobProgress(queriedJobStatus.getProgressPercentage())
+              .setEstimatedSecondsToCompletion(timeLeft))
           .setExecutionStatus(ExecutionStatus.valueOf(queriedJobStatus.getEventName()))
           .setMessage(queriedJobStatus.getMessage())
           .setJobState(new JobState().setLowWatermark(queriedJobStatus.getLowWatermark()).
@@ -198,5 +205,25 @@ public class FlowExecutionResourceLocalHandler implements FlowExecutionResourceH
    */
   private static long getFlowStartTime(org.apache.gobblin.service.monitoring.FlowStatus flowStatus) {
     return flowStatus.getFlowExecutionId();
+  }
+
+  /**
+   * Estimate the time left to complete the copy based on the following formula -
+   *  timeLeft = (100/completionPercentage - 1) * timeElapsed
+   * @param currentTime as an epoch
+   * @param startTime as an epoch
+   * @param completionPercentage of the job
+   * @return time left in seconds
+   */
+  public static long estimateCopyTimeLeft(Long currentTime, Long startTime, int completionPercentage) {
+    if (completionPercentage == 0) {
+      return 0;
+    }
+
+    Instant current = Instant.ofEpochMilli(currentTime);
+    Instant start = Instant.ofEpochMilli(startTime);
+    Long timeElapsed = Duration.between(start, current).getSeconds();
+    Long timeLeft = (long) (timeElapsed * (100.0 / Double.valueOf(completionPercentage) - 1));
+    return timeLeft;
   }
 }
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -17,11 +17,13 @@
 
 package org.apache.gobblin.runtime;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.net.Authenticator;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -31,6 +33,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang.StringUtils;
+
+import org.apache.gobblin.service.ServiceConfigKeys;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -463,6 +467,18 @@ public abstract class AbstractJobLauncher implements JobLauncher {
           return;
         }
 
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, ConfigurationKeys.DEFAULT_REPORT_JOB_PROGRESS)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnitStream);
+            this.jobContext.getJobState().setProp(ServiceConfigKeys.TOTAL_WORK_UNIT_SIZE, totalSizeInBytes);
+          } else {
+            LOG.warn("Property " + ConfigurationKeys.REPORT_JOB_PROGRESS + " is turned on, but "
+                + "progress cannot be reported for infinite work unit streams. Turn off property "
+                + ConfigurationKeys.REPORT_JOB_PROGRESS);
+          }
+        }
+
         // Perform work needed before writing is done
         Boolean canCleanUp = this.canCleanStagingData(this.jobContext.getJobState());
         closer.register(new DestinationDatasetHandlerService(jobState, canCleanUp, this.eventSubmitter))
@@ -642,6 +658,13 @@ public abstract class AbstractJobLauncher implements JobLauncher {
       MDC.remove(ConfigurationKeys.JOB_NAME_KEY);
       MDC.remove(ConfigurationKeys.JOB_KEY_KEY);
     }
+  }
+
+  @VisibleForTesting
+  public static long sumWorkUnitsSizes (WorkUnitStream workUnitStream) {
+    Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+    long totalSizeInBytes = workUnits.stream().mapToLong(wu -> wu.getPropAsLong(ServiceConfigKeys.WORK_UNIT_SIZE)).sum();
+    return totalSizeInBytes;
   }
 
   private void executeUnfinishedCommitSequences(String jobName)
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
@@ -18,7 +18,9 @@
 package org.apache.gobblin.runtime;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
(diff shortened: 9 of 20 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -206,12 +222,53 @@ public Void call() throws Exception {
...
+    Long taskByteSize;
+    String stringSize;
+    Double newPercentageCopied;
     // Add the TaskStates of completed tasks to the JobState so when the control
     // returns to the launcher, it sees the TaskStates of all completed tasks.
     for (TaskState taskState : taskStateQueue) {
       consumeTaskIssues(taskState);
       taskState.setJobState(this.jobState);
       this.jobState.addTaskState(taskState);
+
```
aplex, line 234: This method is already quite complicated. Generally, we should try to have no more a screens of code in each method, to make it easier to understand and read. You can try moving the logic to a separate method or class, whatever makes more sense.

Fields like "lastPercentageReported" can be more meaningful in separate class, rather than in task state collector. That will better match single responsibility principle - https://en.wikipedia.org/wiki/SOLID

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -206,12 +222,53 @@ public Void call() throws Exception {
...
+          break;
+        }
+
+        this.totalSizeToCopy = Long.parseLong(strTotalSizeToCopy);
+        // avoid flooding Kafka message queue by sending GobblinTrackingEvents only when threshold is passed
+        this.bytesCopiedSoFar += taskByteSize;
+        newPercentageCopied = this.bytesCopiedSoFar / this.totalSizeToCopy;
+        if (newPercentageCopied - this.lastPercentageReported > ConfigurationKeys.DEFAULT_PROGRESS_REPORTING_THRESHOLD) {
+          this.lastPercentageReported = newPercentageCopied;
+          Long percentageToReport = (long) (this.lastPercentageReported * 100);
```
aplex, line 256: Should we do a Math.Round() here? I think by default the value will be truncated when you convert double to long. so 99.999 will be 99 , not 100. And we can end up with jobs that are fully complete, but still show 99%  progress.

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaAvroJobStatusMonitor.java
```diff
@@ -163,6 +163,9 @@ public GobblinTrackingEvent deserializeEvent(DecodeableKafkaRecord<byte[],byte[]
         properties.put(JobStatusRetriever.EVENT_NAME_FIELD, ExecutionStatus.CANCELLED.name());
         properties.put(TimingEvent.JOB_END_TIME, properties.getProperty(TimingEvent.METADATA_END_TIME));
         break;
+      case TimingEvent.JOB_COMPLETION_PERCENTAGE:
+        properties.put(TimingEvent.JOB_LAST_PROGRESS_PERCENT_TIME, properties.getProperty(TimingEvent.METADATA_END_TIME));
```
aplex, line 167: Is METADATA_END_TIME the correct value here?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
```diff
@@ -19,4 +19,14 @@ record JobStatistics {
...
+
+  /**
+   * data copy progress as a percentage
+   */
+  copyProgress: long
+
+  /**
+   * estimate of time left until copy completion
+   */
+  estimatedTimeLeft: long
```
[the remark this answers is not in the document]
umustafi: I have it in milliseconds since that's how the executionStart/EndTimes are set.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +468,17 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
         }
 
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_COPY_PROGRESS, false)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnits);
+            this.jobContext.getJobState().setProp(TOTAL_BYTES_TO_COPY, totalSizeInBytes);
+          } else {
+            throw new RuntimeException(DestinationDatasetHandlerService.class.getName() + " does not support work unit streams");
```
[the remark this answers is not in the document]
umustafi: Great point, the error message you suggested will be more helpful!

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
umustafi: Moving to a separate method for clarity because the state of the progress related class is tied to one instance of the taskCollectorService. If it still makes sense to separate the two into classes then I can attempt that.

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaAvroJobStatusMonitor.java
umustafi: I believe METADATA_END_TIME is set when event.stop() is called (which is what I use to obtain the timestamp of when the progress event was set)

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
```diff
@@ -199,4 +206,22 @@ public static FlowExecution convertFlowStatus(org.apache.gobblin.service.monitor
...
+  public static long estimateCopyTimeLeft(Long currentTime, Long startTime, int completionPercentage) {
+    if (completionPercentage == 0) {
+      return 0;
+    }
+
+    Instant current = Instant.ofEpochMilli(currentTime);
+    Instant start = Instant.ofEpochMilli(startTime);
+    Long timeBetween = Duration.between(current, start).toMillis();
+    Long timeElapsed = currentTime - startTime;
+    log.info("mylogs: timeBetween is " + timeBetween + " and elapsed is " + timeElapsed);
```
umustafi, line 223: will remove this log statement

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
```diff
@@ -382,9 +384,17 @@ public Void call() {
...
+            CopyableFile castedCopyEntity = (CopyableFile) copyEntity;
+            fileSize = castedCopyEntity.getFileStatus().getLen();
+            workUnit.setProp(ServiceConfigKeys.WORK_UNIT_BYTE_SIZE, fileSize);
+            if (DistcpFileSplitter.allowSplit(this.state, this.targetFs)) {
+              workUnitsForPartition.addAll(DistcpFileSplitter.splitFile((CopyableFile) copyEntity, workUnit, this.targetFs));
+            } else {
+              workUnitsForPartition.add(workUnit);
+            }
           } else {
+            workUnit.setProp(ServiceConfigKeys.WORK_UNIT_BYTE_SIZE, 0);
```
Will-Lo, line 397: Can we leave a comment for when this case is true? From what I remember it had to do with local state files/ post publishing steps also existing in the directory.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +469,20 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, false)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnits);
+            this.jobContext.getJobState().setProp(TOTAL_BYTES_TO_COPY, totalSizeInBytes);
+          } else {
+            throw new RuntimeException("Property " + ConfigurationKeys.REPORT_JOB_PROGRESS + " is turned on, but "
+                + "progress cannot be collected because " + DestinationDatasetHandlerService.class.getName()
+                + " does not support work unit streams. Turn off property " + ConfigurationKeys.REPORT_JOB_PROGRESS
+                + " and rerun job.");
```
Will-Lo, line 482: I think it's not necessarily true that it's due to `DestinationDatasetHandler` not supporting workunitstreams, but that this feature also does not support work unit streams (as total size is determined only once at the beginning).

Will-Lo: We can put a better exception message here to let the user know that copy progression doesn't work for infinite streams (which make sense, since how would we determine that the work is completed?)

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -644,6 +660,15 @@ public void apply(JobListener jobListener, JobContext jobContext)
     }
   }
 
+  public static long sumWorkUnitsSizes (Collection<WorkUnit> workUnits) {
```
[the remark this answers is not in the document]
Will-Lo: If you mark @VisibleForTesting you should make the function private/protected (probably protected in this scenario)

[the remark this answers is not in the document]
Will-Lo: If you add the annotation the function should be protected/private, in this case I think it's good to make protected.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/copy/CopySource.java
Will-Lo, line 397: Can you leave a comment for when this condition would apply? If I recall correctly it's for some state files/post publish states

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -235,6 +255,49 @@ public Void call() throws Exception {
...
+  private void reportJobProgress(TaskState taskState) {
+    String stringSize = taskState.getProp(ServiceConfigKeys.WORK_UNIT_BYTE_SIZE);
+    if (stringSize == null) {
+      LOGGER.warn("Expected to report job progress but work unit byte size property null");
+      return;
+    }
+
+    Long taskByteSize = Long.parseLong(stringSize);
+    // if progress reporting is enabled, value should be present
+    String strTotalSizeToCopy = this.jobState.getProp(AbstractJobLauncher.TOTAL_BYTES_TO_COPY);
```
Will-Lo, line 272: You can use `PropertiesUtils.getPropAsLong` here (and for task byte size) instead of pulling the String. Instead of checking for null, check that the property exists in the state.

gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatusRetriever.java
```diff
@@ -105,6 +105,8 @@ protected JobStatus getJobStatus(State jobState) {
     int maxAttempts = Integer.parseInt(jobState.getProp(TimingEvent.FlowEventConstants.MAX_ATTEMPTS_FIELD, "1"));
     int currentAttempts = Integer.parseInt(jobState.getProp(TimingEvent.FlowEventConstants.CURRENT_ATTEMPTS_FIELD, "1"));
     boolean shouldRetry = Boolean.parseBoolean(jobState.getProp(TimingEvent.FlowEventConstants.SHOULD_RETRY_FIELD, "false"));
+    int progressPercentage = Integer.parseInt(jobState.getProp(TimingEvent.JOB_COMPLETION_PERCENTAGE, "0"));
+    long lastProgressEventTime = Long.parseLong(jobState.getProp(TimingEvent.JOB_LAST_PROGRESS_PERCENT_TIME, "0"));
```
Will-Lo, line 109: Similar comment here, you can use PropertiesUtils to parse ints/longs

Will-Lo (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
Will-Lo: I think a separate method for now is good here, until we can determine more use cases/extensions to the progress reporting

gobblin-data-management/src/test/java/org/apache/gobblin/data/management/copy/TestCopyableDataset.java
```diff
@@ -72,6 +73,7 @@ public TestCopyableDataset() {
     return files;
   }
 
+
```
umustafi, line 76: removing

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -235,6 +255,49 @@ public Void call() throws Exception {
...
+      int percentageToReport = (int) Math.round(this.lastPercentageReported * 100);
+
+      Map<String, String> progress = new HashMap<>();
+      progress.put(TimingEvent.JOB_COMPLETION_PERCENTAGE, String.valueOf(percentageToReport));
+      progress.put(TimingEvent.FlowEventConstants.FLOW_GROUP_FIELD,
+          this.jobState.getProp(ConfigurationKeys.FLOW_GROUP_KEY));
+      progress.put(TimingEvent.FlowEventConstants.FLOW_NAME_FIELD,
+          this.jobState.getProp(ConfigurationKeys.FLOW_NAME_KEY));
+      progress.put(TimingEvent.FlowEventConstants.FLOW_EXECUTION_ID_FIELD,
+          this.jobState.getProp(ConfigurationKeys.FLOW_EXECUTION_ID_KEY));
```
jack-moseley, line 294: I think adding these properties is not actually required. You are using the `eventSubmitter` passed from `MRJobLauncher`, and that already includes a bunch of additional metadata including all three of these properties that are added by default.

If you look at other job level events sent by GaaS such as `WorkUnitsCreated`, they already contain these despite not explicitly adding them.

jack-moseley (changes requested):

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -681,6 +681,8 @@
   public static final String TASK_STATE_COLLECTOR_INTERVAL_SECONDS = "task.state.collector.interval.secs";
   public static final int DEFAULT_TASK_STATE_COLLECTOR_INTERVAL_SECONDS = 60;
   public static final String TASK_STATE_COLLECTOR_HANDLER_CLASS = "task.state.collector.handler.class";
+  public static final String REPORT_JOB_PROGRESS = "job.report.progress";
```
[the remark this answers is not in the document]
arjun4084346: Let's not jumble the words in key and value

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -17,11 +17,13 @@
 
 package org.apache.gobblin.runtime;
 
+import avro.shaded.com.google.common.annotations.VisibleForTesting;
```
arjun4084346, line 20: just keep it com.google.common.annotations.VisibleForTesting

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +469,19 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
         }
 
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, false)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnits);
+            this.jobContext.getJobState().setProp(TOTAL_BYTES_TO_COPY, totalSizeInBytes);
+          } else {
+            throw new RuntimeException("Property " + ConfigurationKeys.REPORT_JOB_PROGRESS + " is turned on, but "
```
arjun4084346, line 479: UnsupportedOperationException is more appropriate, IMO

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +469,19 @@ public void apply(JobListener jobListener, JobContext jobContext)
           return;
         }
 
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, false)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
```
arjun4084346, line 475: I think workUnits can go inside sumWorkUnitsSizes because it is not used here.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -644,6 +663,16 @@ public void apply(JobListener jobListener, JobContext jobContext)
     }
   }
 
+  @VisibleForTesting
+  public static long sumWorkUnitsSizes (Collection<WorkUnit> workUnits) {
+    long totalSizeInBytes = 0;
+    for (WorkUnit workUnit : workUnits) {
```
arjun4084346, line 669: This is okay too. can be written with lamba expressions too
`long totalSizeInBytes = workUnits.stream().mapToLong(wu -> Long.parseLong(wu.getProp(ServiceConfigKeys.WORK_UNIT_BYTE_SIZE))).sum();`

gobblin-runtime/src/main/java/org/apache/gobblin/service/monitoring/JobStatusRetriever.java
```diff
@@ -105,6 +105,8 @@ protected JobStatus getJobStatus(State jobState) {
     int maxAttempts = Integer.parseInt(jobState.getProp(TimingEvent.FlowEventConstants.MAX_ATTEMPTS_FIELD, "1"));
     int currentAttempts = Integer.parseInt(jobState.getProp(TimingEvent.FlowEventConstants.CURRENT_ATTEMPTS_FIELD, "1"));
     boolean shouldRetry = Boolean.parseBoolean(jobState.getProp(TimingEvent.FlowEventConstants.SHOULD_RETRY_FIELD, "false"));
+    int progressPercentage = jobState.getPropAsInt(TimingEvent.JOB_COMPLETION_PERCENTAGE, 0);
+    long lastProgressEventTime = jobState.getPropAsLong(TimingEvent.JOB_LAST_PROGRESS_PERCENT_TIME, 0);
```
arjun4084346, line 109: It took me a little while to figure this out. IMO it could be named JOB_LAST_PROGRESS_EVENT_TIME

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
```diff
@@ -199,4 +206,20 @@ public static FlowExecution convertFlowStatus(org.apache.gobblin.service.monitor
   private static long getFlowStartTime(org.apache.gobblin.service.monitoring.FlowStatus flowStatus) {
     return flowStatus.getFlowExecutionId();
   }
+
+  /**
+   * Estimate the time left to complete the copy based on the following formula -
+   * timeLeft = (100/completionPercentage) * timeElapsed
```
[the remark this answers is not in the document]
umustafi: Good catch, it should be timeElapsed * (100/completionPercentage - 1)

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +469,19 @@ public void apply(JobListener jobListener, JobContext jobContext)
...
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, false)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnits);
+            this.jobContext.getJobState().setProp(TOTAL_BYTES_TO_COPY, totalSizeInBytes);
+          } else {
+            throw new RuntimeException("Property " + ConfigurationKeys.REPORT_JOB_PROGRESS + " is turned on, but "
+                + "progress cannot be reported for infinite work unit streams. Turn off property "
+                + ConfigurationKeys.REPORT_JOB_PROGRESS + " and rerun job.");
```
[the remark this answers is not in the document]
umustafi: good point, job failure may not make sense if we just expect the user to turn off the property and retry it. Changing it to an error message.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
umustafi: Yes, I had noticed that other events do not need these three properties because they are not processed by the `KafkaAvroStatusMonitor` (ie: in the case of `WorkUnitsCreated`) or are explicitly adding them. From printing the metadata it seems it only has the following and yet needs the flow level props:
`{jobName, jobId, azkabanJobId=autotest-hive-copy_hive-copy, azkabanFlowURL, azkabanJobExecURL, azkabanJobURL, azkabanProjectName, azkabanFlowId, azkabanExecId, azkabanURL, clusterIdentifier, metricContextID, metricContextName}`

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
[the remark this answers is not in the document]
umustafi: also added a test with this case

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
```diff
@@ -199,4 +206,20 @@ public static FlowExecution convertFlowStatus(org.apache.gobblin.service.monitor
...
+   */
+  public static long estimateCopyTimeLeft(Long currentTime, Long startTime, int completionPercentage) {
+    if (completionPercentage == 0) {
+      return 0;
+    }
+
+    Instant current = Instant.ofEpochMilli(currentTime);
+    Instant start = Instant.ofEpochMilli(startTime);
+    Long timeBetween = Duration.between(start, current).toMillis();
```
arjun4084346, line 221: Is not in simply `current-start` ? Or did I miss something?

umustafi: it is however alex had recommended using the Instant/Duration classes earlier to make it clear what units the times were in. I agree that this may not add much clarity and perhaps a comment can do the same.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/snapshot/org.apache.gobblin.service.flowstatuses.snapshot.json
```diff
@@ -180,6 +180,14 @@
                 "name" : "processedCount",
                 "type" : "long",
                 "doc" : "number of records processed in the last job execution"
+              }, {
+                "name" : "jobProgress",
+                "type" : "int",
+                "doc" : "job progress as a percentage (0-100)"
+              }, {
+                "name" : "estimatedMillisecondsToCompletion",
```
arjun4084346, line 188: Sorry for the late comment. Do you think we should have `estimatedSecondsToCompletion` instead of Milliseconds?

umustafi: the `executionStart/EndTimes` are in milliseconds, that's why I left it in the same format. I agree we won't pay attention to milliseconds but I was planning on rounding down to appropriate units of time before formatting the result to user in the jobStatus endpoint.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
jack-moseley: If we do need to add these, let's check for existence in the jobState before adding them, because this feature could be turned on for a non-gaas distcp job and then we would be setting those properties to null.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -235,6 +255,49 @@ public Void call() throws Exception {
...
+
+    // if progress reporting is enabled, value should be present
+    if (!this.jobState.contains(AbstractJobLauncher.TOTAL_BYTES_TO_COPY)) {
+      LOGGER.warn("Expected to report job progress but total bytes to copy property null");
+      return;
+    }
+    this.totalSizeToCopy = this.jobState.getPropAsLong(AbstractJobLauncher.TOTAL_BYTES_TO_COPY);
+
+    // avoid flooding Kafka message queue by sending GobblinTrackingEvents only when threshold is passed
+    this.bytesCopiedSoFar += taskByteSize;
```
arjun4084346, line 280: so the bytesCopiedSoFar will be updated once the first mapper completes? Is this true?
If yes, is it okay to show timeLeft = 0 before any mapper finished? Should we return "-1"  or something? Usually how long will it take for first mapper to complete?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
arjun4084346: @umustafi these params are added in IvyJobLauncher. Let's remove them from here and if it does not work, we will debug.

umustafi: removed them because they should be present

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
umustafi: Yes it will be 0 until the first mapper completes but the percentage of progress will also be 0 - one will not be set without the other and we hope this is understandable enough to the user. In the usual case we expect the first mapper to complete in seconds or 1-2 min.

arjun4084346:
+1 LGTM

arjun4084346 (approved):

gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
```diff
@@ -58,6 +58,10 @@
   public static final String GOBBLIN_SERVICE_FLOW_CATALOG_LOCAL_COMMIT = GOBBLIN_SERVICE_PREFIX + "flowCatalog.localCommit";
   public static final boolean DEFAULT_GOBBLIN_SERVICE_FLOW_CATALOG_LOCAL_COMMIT = true;
 
+  // Job Level Keys
+  public static final String WORK_UNIT_BYTE_SIZE = GOBBLIN_SERVICE_PREFIX + ".work.unit.byte.size";
```
sv2000, line 62: WORK_UNIT_BYTE_SIZE is confusing to read. Can we just call it work_unit_size? Also, is this config only intended for File-based sources? How about other sources e.g. Kafka Source, where this config could be used to mean number of records?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -463,6 +469,18 @@ public void apply(JobListener jobListener, JobContext jobContext)
           return;
         }
 
+        // calculation of total bytes to copy in a job used to track a job's copy progress
+        if (jobState.getPropAsBoolean(ConfigurationKeys.REPORT_JOB_PROGRESS, ConfigurationKeys.DEFAULT_REPORT_JOB_PROGRESS)) {
+          if (workUnitStream.isSafeToMaterialize()) {
+            long totalSizeInBytes = sumWorkUnitsSizes(workUnitStream);
+            this.jobContext.getJobState().setProp(TOTAL_BYTES_TO_COPY, totalSizeInBytes);
```
sv2000, line 476: TOTAL_BYTES_TO_COPY seems very Distcp-centric. We may want to think more broadly and just call this TOTAL_WORK_UNIT_SIZE. In case when individual sources do not provide workunit sizes, this should sum to total number of workunits.

sv2000 (changes requested):

gobblin-api/src/main/java/org/apache/gobblin/service/ServiceConfigKeys.java
umustafi: This isn't meant for only file-based sources. It will also be used for record-based sources so it will make more sense to call it `work_unit_size` like you're suggesting.

jack-moseley (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
umustafi: right now we have this for jobEndTime as well I noticed in the jobStatistics (everything starts at 0)

Will-Lo (approved):

sv2000 (approved):
+1.

PR #3340, opened 2021-07-26
[GOBBLIN-1497] Reduce the number of calls on FlowSpec initialization where possible,…

Will-Lo:
… and configToProperties/vice versa

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1497

### Description

FlowSpecs contain both configs and property fields that duplicate the same key/value information, however there are slight differences with properties being mutable, configs being type-safe, and legacy code being compatible with properties but not configs.

As such, FlowSpec creation and configToProperties are memory expensive operations at scale, and we should aim to minimize them as much as possible.

### Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/FlowConfigUtils.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/FlowConfigUtils.java
@@ -22,12 +22,10 @@ import java.util.Properties;
 
 import com.google.common.collect.Maps;
 import com.linkedin.data.template.StringMap;
-import com.typesafe.config.ConfigFactory;
 
 import org.apache.gobblin.service.FlowConfig;
 import org.apache.gobblin.service.FlowId;
 import org.apache.gobblin.service.Schedule;
-import org.apache.gobblin.util.ConfigUtils;
 import org.apache.gobblin.util.PropertiesUtils;
 
 
@@ -60,7 +58,7 @@ public class FlowConfigUtils {
   }
 
   public static String serializeFlowConfig(FlowConfig flowConfig) throws IOException {
-    Properties properties = ConfigUtils.configToProperties(ConfigFactory.parseMap(flowConfig.getProperties()));
+    Properties properties = PropertiesUtils.stringMapToProperties(flowConfig.getProperties());
     properties.setProperty(FLOWCONFIG_ID_NAME, flowConfig.getId().getFlowName());
     properties.setProperty(FLOWCONFIG_ID_GROUP, flowConfig.getId().getFlowGroup());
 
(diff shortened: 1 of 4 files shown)


Will-Lo:
@arjun4084346 @jack-moseley @umustafi could you take a look?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/restli/FlowConfigUtils.java
```diff
@@ -60,7 +58,7 @@ public static FlowId deserializeFlowId(String serialized) throws IOException {
   }
 
   public static String serializeFlowConfig(FlowConfig flowConfig) throws IOException {
-    Properties properties = ConfigUtils.configToProperties(ConfigFactory.parseMap(flowConfig.getProperties()));
+    Properties properties = PropertiesUtils.stringMapToProperties(flowConfig.getProperties());
```
arjun4084346, line 61: `Properties properties = new Properties().putAll(flowConfig.getProperties());` works

gobblin-utility/src/main/java/org/apache/gobblin/util/PropertiesUtils.java
```diff
@@ -152,4 +152,12 @@ public static String prettyPrintProperties(Properties properties) {
         .map(entry -> "\"" + entry.getKey() + "\"" + " : " + "\"" + entry.getValue() + "\"")
         .collect(Collectors.joining(",\n"));
   }
+
+  public static Properties stringMapToProperties(Map<String, String> values) {
+    Properties properties = new Properties();
```
arjun4084346, line 157: equivalent to putAll()

arjun4084346:
minor comments to simply things. LGTM otherwise. good changes. maybe we should get rid of configAsProperties entirely. but that may need more changes.

Will-Lo:
@arjun4084346 yea we should look into configAsProperties especially at the rate that GaaS is growing at. Would require some refactoring around some legacy classes and probably more conversions from propertiesToConfig. However since those would be short-lived compared to FlowSpecs it would still be an overall memory improvement

umustafi (approved):
These replacements look good!

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -202,8 +202,9 @@ private void scheduleSpecsFromCatalog() {
     while (specUris.hasNext()) {
       Spec spec = this.flowCatalog.get().getSpecWrapper(specUris.next());
       try {
-        // Disable FLOW_RUN_IMMEDIATELY on service startup or leadership change
-        if (spec instanceof FlowSpec) {
+        // Disable FLOW_RUN_IMMEDIATELY on service startup or leadership change if the property is set to true
+        if (spec instanceof FlowSpec && ((FlowSpec) spec).getConfigAsProperties()
+            .getProperty(ConfigurationKeys.FLOW_RUN_IMMEDIATELY, "false").equals("true")) {
```
sv2000, line 207: Use PropertiesUtils.getPropAsBoolean() here and check for the boolean instead of the string value.

sv2000 (changes requested):
A minor comment.

sv2000 (approved):
+1.

PR #3343, opened 2021-07-28
[GOBBLIN-1499] fix the bug that compaction job hang when there is no fix the bug that compaction job hang when there is no path match the datasets patten

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1499

### Description
fix the bug that compaction job hang when there is no path match the datasets patten. When get the compaction workunit, it reach the end only when isDone set to true and workunit size is zero. But if there is no match of pattern, isDone is always false, so job hang.

### Tests

### Commits

### DIFF
--- a/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
+++ b/gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
@@ -295,7 +295,6 @@ public class HiveMetadataWriter implements MetadataWriter {
     spec.getTable()
         .getSerDeProps()
         .setProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), latestSchemaMap.get(tableKey));
-
   }
 
   private String fetchSchemaFromTable(String dbName, String tableName) throws IOException {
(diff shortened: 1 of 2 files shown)


gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
```diff
@@ -292,9 +292,9 @@ private void schemaUpdateHelper(GobblinMetadataChangeEvent gmce, HiveSpec spec,
       return;
     }
     //Force to set the schema even there is no schema literal defined in the spec
-    spec.getTable()
-        .getSerDeProps()
-        .setProp(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName(), latestSchemaMap.get(tableKey));
+    if(latestSchemaMap.containsKey(tableKey)) {
```
sv2000, line 295: This change looks unrelated to the PR.

sv2000 (changes requested):
Looks like there is an unrelated change that is being pulled in.

gobblin-hive-registration/src/main/java/org/apache/gobblin/hive/writer/HiveMetadataWriter.java
ZihanLi58: Thanks for catching that. just revert the change, will create a separate pr for that fix.

sv2000 (approved):
+1.

PR #3345, opened 2021-07-30
[GOBBLIN-1500]Support gobblin on yarn to be able to run on clusters with federated Yarn cluster enabled 

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1500

### Description
Previously we always check with one default yarn client, but if the cluster is enabled with Robin and can connect to different RM, it will ends up we cannot find existing application and submit duplicate tasks. So now we need to check against different potential yarn cluster, and also make sure when submitting yarn application, we only have one RM token.

Options

### Tests
Test start a new application can work, test reconnect can work 

### Commits

### DIFF
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
@@ -266,7 +266,7 @@ public class GobblinYarnAppLauncher {
     for (String rmAddress : potentialRMAddresses) {
       YarnClient tmpYarnClient = YarnClient.createYarnClient();
       this.yarnConfiguration.set(GobblinYarnConfigurationKeys.YARN_RESOURCE_MANAGER_ADDRESS, rmAddress);
-      tmpYarnClient.init(this.yarnConfiguration);
+      tmpYarnClient.init(new YarnConfiguration(this.yarnConfiguration));
       potentialYarnClients.put(rmAddress, tmpYarnClient);
     }
 
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
@@ -264,7 +264,7 @@ public class GobblinYarnAppLauncher {
     Set<String> potentialRMAddresses = new HashSet<>(ConfigUtils.getStringList(config, GobblinYarnConfigurationKeys.OTHER_YARN_RESOURCE_MANAGER_ADDRESSES));
     potentialRMAddresses.add(originalYarnRMAddress);
     for (String rmAddress : potentialRMAddresses) {
-       YarnClient tmpYarnClient = YarnClient.createYarnClient();
+      YarnClient tmpYarnClient = YarnClient.createYarnClient();
       this.yarnConfiguration.set(GobblinYarnConfigurationKeys.YARN_RESOURCE_MANAGER_ADDRESS, rmAddress);
       tmpYarnClient.init(this.yarnConfiguration);
       potentialYarnClients.put(rmAddress, tmpYarnClient);
--- a/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
+++ b/gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
@@ -259,13 +259,13 @@ public class GobblinYarnAppLauncher {
     YarnHelixUtils.setYarnClassPath(config, this.yarnConfiguration);
     YarnHelixUtils.setAdditionalYarnClassPath(config, this.yarnConfiguration);
     this.yarnConfiguration.set("fs.automatic.close", "false");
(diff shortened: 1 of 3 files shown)


gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnConfigurationKeys.java
```diff
@@ -37,6 +37,8 @@
   public static final int DEFAULT_RELEASED_CONTAINERS_CACHE_EXPIRY_SECS = 300;
   public static final String APP_VIEW_ACL = GOBBLIN_YARN_PREFIX + "appViewAcl";
   public static final String DEFAULT_APP_VIEW_ACL = "*";
+  public static final String YARN_RESOURCEMANAGER_ADDRESS= "yarn.resourcemanager.address";
+  public static final String POTENTIAL_YARN_RESOURCEMANAGER_ADDRESSES= "potential.yarn.resourcemanager.addresses";
```
sv2000, line 41: OTHER_YARN_RESOURCE_MANAGER_ADDRESSES instead of POTENTIAL_

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -574,19 +592,21 @@ private String sanitizeApplicationId(String applicationId) {
...
 
-    // Try to find an application with a matching application name
-    for (ApplicationReport applicationReport : applicationReports) {
-      if (this.applicationName.equals(applicationReport.getName())) {
-        String applicationId = sanitizeApplicationId(applicationReport.getApplicationId().toString());
-        LOGGER.info("Found reconnectable application with application ID: " + applicationId);
-        LOGGER.info("Application Tracking URL: " + applicationReport.getTrackingUrl());
-        return Optional.of(applicationReport.getApplicationId());
+      // Try to find an application with a matching application name
+      for (ApplicationReport applicationReport : applicationReports) {
```
sv2000, line 602: What happens if applicationReports is null? We won't enter this loop. Shouldn't we return Optional.absent() for that case?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -835,14 +855,22 @@ private void setupSecurityTokens(ContainerLaunchContext containerLaunchContext)
 
     TokenUtils.getAllFSTokens(new Configuration(), credentials, renewerName,
         Optional.absent(), ConfigUtils.getStringList(this.config, TokenUtils.OTHER_NAMENODES));
-
+    // Only pass token here and no secrets. (since there is no simple way to remove single token/ get secrets)
+    // For RM token, only pass the RM token for the current RM, or the RM will fail to update the token
+    Credentials finalCredentials = new Credentials();
+    for ( Token<? extends TokenIdentifier> token: credentials.getAllTokens()) {
```
sv2000, line 861: Nit: Remove unnecessary whitespaces after "for" and "(".

sv2000 (changes requested):
A few comments. One question: How does token refresh work in a federated Yarn cluster for keytab based authentication? Will it work as is after the changes in this PR?

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
ZihanLi58: yeah, in that case, we will continue to check next potential clusters, and if we found nothing in all clusters, we will return Optional.absent on line 613

ZihanLi58:
> A few comments. One question: How does token refresh work in a federated Yarn cluster for keytab based authentication? Will it work as is after the changes in this PR?

@sv2000 For original implementation, i.e. there is only one yarn cluster, it should still work after this change. But to support the federated clusters, we may need to implement the method to get delegation token from remote yarn clusters and I don't see anything related to rm token in our keytab based authentication.

gobblin-yarn/src/main/java/org/apache/gobblin/yarn/GobblinYarnAppLauncher.java
```diff
@@ -253,8 +259,16 @@ public GobblinYarnAppLauncher(Config config, YarnConfiguration yarnConfiguration
...
     YarnHelixUtils.setAdditionalYarnClassPath(config, this.yarnConfiguration);
     this.yarnConfiguration.set("fs.automatic.close", "false");
-    this.yarnClient = YarnClient.createYarnClient();
-    this.yarnClient.init(this.yarnConfiguration);
+    this.originalYarnRMAddress = this.yarnConfiguration.get(GobblinYarnConfigurationKeys.YARN_RESOURCE_MANAGER_ADDRESS);
+    this.potentialYarnClients = new HashMap();
+    Set<String> potentialRMAddresses = new HashSet<>(ConfigUtils.getStringList(config, GobblinYarnConfigurationKeys.OTHER_YARN_RESOURCE_MANAGER_ADDRESSES));
+    potentialRMAddresses.add(originalYarnRMAddress);
+    for (String rmAddress : potentialRMAddresses) {
+       YarnClient tmpYarnClient = YarnClient.createYarnClient();
```
umustafi, line 267: nit but i think the spacing is off here - Alex told me to run "silent code cleanup" in Intellij on blocks of code we change to auto correct these.

sv2000 (approved):
+1.

PR #3353, opened 2021-08-05
[GOBBLIN-1507] Prevent orchestrator removals when unscheduling flows

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1507

### Description

When demoted from leader, GaaS removes all flows from orchestrator. We need to ensure that fail over events are not disruptive or destructive to currently running flows, we only want the new leader node to handle scheduling responsibilities from that point forward.

### Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
@@ -71,7 +71,7 @@ import org.quartz.UnableToInterruptJobException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static org.apache.gobblin.service.ServiceConfigKeys.*;
+import static org.apache.gobblin.service.ServiceConfigKeys.GOBBLIN_SERVICE_PREFIX;
 
 
 /**
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
@@ -17,6 +17,11 @@
 
 package org.apache.gobblin.service.modules.scheduler;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
+import com.google.common.collect.Maps;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -235,7 +231,7 @@ private void scheduleSpecsFromCatalog() {
   private void clearRunningFlowState(Iterator<URI> drUris) {
     while (drUris.hasNext()) {
       // TODO: Instead of simply call onDeleteSpec, a callback when FlowSpec is deleted from FlowCatalog, should also kill Azkaban Flow from AzkabanSpecProducer.
-      onDeleteSpec(drUris.next(), FlowSpec.Builder.DEFAULT_VERSION);
+      unscheduleSpec(drUris.next(), FlowSpec.Builder.DEFAULT_VERSION);
```
jack-moseley, line 234: We're not using this DR code anyway, but based on comments it seems like it's intended that it kills running flows that are marked as DR, so maybe let's leave this method as is.

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -347,6 +343,30 @@ public AddSpecResponse onAddSpec(Spec addedSpec) {
...
   }
 
+  /**
+   * Remove a flowSpec from schedule due to another leader being elected
+   * Unlike onDeleteSpec, we want to avoid deleting the flowSpec on the executor
+   * and we still want to unschedule if we cannot connect to zookeeper as the current node cannot be the master
+   * @param specURI
+   * @param specVersion
+   */
+  private void unscheduleSpec(URI specURI, String specVersion) {
```
jack-moseley, line 353: Could we do this instead by just adding an overloaded onDeleteSpec with a boolean option to remove from the orchestrator to avoid the code duplication?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
```diff
@@ -176,7 +172,7 @@ public void run() {
       // Since we are going to change status to isActive=false, unschedule all flows
       List<Spec> specs = new ArrayList<>(this.scheduledFlowSpecs.values());
       for (Spec spec : specs) {
-        onDeleteSpec(spec.getUri(), spec.getVersion());
+        unscheduleSpec(spec.getUri(), spec.getVersion());
```
arjun4084346, line 175: why not have ` this.scheduledFlowSpecs.remove(deletedSpecURI.toString());` also?

gobblin-service/src/main/java/org/apache/gobblin/service/modules/scheduler/GobblinServiceJobScheduler.java
arjun4084346: or call this from inside onDeleteSpec()

Will-Lo: onDeleteSpec also checks for connection to zookeeper though, and I think that's a bad check if the host is demoted. If a host that used to be master gets demoted, it's likely struggling with connectivity so unscheduleSpec handles that.

@arjun4084346 can do that too

jack-moseley (approved):

aplex (approved):

PR #3358, opened 2021-08-09
[GOBBLIN-1510] Fix Progress Reporting Behavior & Cleanup Logs

umustafi:
* fix bug where progress appears "stuck" between 96-99%
 * decrease log statements
 * format estimated time left to be human readable

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1510

### Description

### Tests
It updates the FlowExecutionResourceLocalHandlerTests 

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
@@ -27,6 +27,7 @@ record JobStatistics {
 
   /**
    * estimate of time left until job completion
+   * "X hours Y minutes Z seconds"
    */
-  estimatedSecondsToCompletion: long
+  estimatedTimeToCompletion: string
 }
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
@@ -27,7 +27,6 @@ record JobStatistics {
 
   /**
    * estimate of time left until job completion
-   * "X hours Y minutes Z seconds"
    */
-  estimatedTimeToCompletion: string
+  estimatedSecondsToCompletion: long
 }
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
@@ -23,6 +23,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.time.DurationFormatUtils;
 import org.apache.commons.lang3.ObjectUtils;
 
 import com.google.common.base.Strings;
@@ -142,7 +143,7 @@ public class FlowExecutionResourceLocalHandler implements FlowExecutionResourceH
 
       JobStatus jobStatus = new JobStatus();
 
-      Long timeLeft = estimateCopyTimeLeft(queriedJobStatus.getLastProgressEventTime(), queriedJobStatus.getStartTime(),
+      String timeLeft = estimateCopyTimeLeft(queriedJobStatus.getLastProgressEventTime(), queriedJobStatus.getStartTime(),
           queriedJobStatus.getProgressPercentage());
 
       jobStatus.setFlowId(flowId)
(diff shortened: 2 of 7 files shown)


umustafi:
@Will-Lo @arjun4084346 @jack-moseley can you review?

arjun4084346 (approved):

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceLocalHandler.java
```diff
@@ -215,15 +216,18 @@ private static long getFlowStartTime(org.apache.gobblin.service.monitoring.FlowS
...
+      return "Not calculated yet";
     }
 
     Instant current = Instant.ofEpochMilli(currentTime);
     Instant start = Instant.ofEpochMilli(startTime);
     Long timeElapsed = Duration.between(start, current).getSeconds();
     Long timeLeft = (long) (timeElapsed * (100.0 / Double.valueOf(completionPercentage) - 1));
-    return timeLeft;
+
+    Long timeLeftInMilli = Instant.ofEpochSecond(timeLeft).toEpochMilli();
```
Will-Lo, line 229: I'm wondering if it makes sense to change the time to String here, or if we want to change it on the API level right before outputting the number to the user. A few minor caveats for doing it here is that strings take more space and that it'll be harder to run queries through the metrics with strings as opposed to a number for timestamps. Thoughts?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/TaskStateCollectorService.java
```diff
@@ -281,7 +280,8 @@ private void reportJobProgress(TaskState taskState) {
     this.bytesCopiedSoFar += taskByteSize;
     Double newPercentageCopied = this.bytesCopiedSoFar / this.totalSizeToCopy;
 
-    if (newPercentageCopied - this.lastPercentageReported > ConfigurationKeys.DEFAULT_PROGRESS_REPORTING_THRESHOLD) {
+    if (newPercentageCopied - this.lastPercentageReported >= ConfigurationKeys.DEFAULT_PROGRESS_REPORTING_THRESHOLD ||
+        (newPercentageCopied > 1 - ConfigurationKeys.DEFAULT_PROGRESS_REPORTING_THRESHOLD && Math.abs(newPercentageCopied - this.lastPercentageReported) > 0.001)) {
```
Will-Lo, line 284: Let's leave a comment here for what this second scenario covers. Even though it's when the job is going from 95-100%, would 0.001 be a bit low of a threshold? It'll send up to 50 events reporting from 95-100 for each job, which seems a quite different from the previous reporting thresholds.

jack-moseley: Yeah seems to me like we should just have a check for if the percentage has reached 100 instead of reporting more events in the last interval of the job.

umustafi: Agree with the points above. It's a choice essentially between if we want to report 96, 97, 98, 99, and 100 potentially or just report 95 and then 100. I did 0.001 as the threshold in case the 0.1% difference corresponds to the number being rounded up, but I did think it could be excessive like you are saying in the worst case. Any preferences?

jack-moseley: Since we have a config for progress reporting threshold set to 5%, I think it makes the most sense if that still applies even in the last interval instead of suddenly sending more events at the end of the job (other than the fix of always sending the final 100% completion event).

Will-Lo (approved):

Will-Lo (approved):

umustafi:
@sv2000 can you merge?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/pegasus/org/apache/gobblin/service/JobStatistics.pdl
```diff
@@ -27,6 +27,7 @@ record JobStatistics {
 
   /**
    * estimate of time left until job completion
+   * "X hours Y minutes Z seconds"
    */
-  estimatedSecondsToCompletion: long
+  estimatedTimeToCompletion: string
```
aplex, line 32: Rest APIs can be consumed by applications written in any language. Java has  DurationFormatUtils.formatDurationWords, but I guess there is no counterpart in Python/Javascript that will be able to convert that string back to internal duration format.

So, for API fields we should stick to formats that any programming language can handle. To make the progress reporting more nice, we can convert those seconds to human readable-message in the client app.

For dates, there is actually an ISO standard https://en.wikipedia.org/wiki/ISO_8601 that is both human- and machine- readable. Unfortunately, the current rest of the API uses milliseconds to represent dates, and it's important to keep the API surface consistent. So, I don't think we should migrate to it at this point, but we can do in the next major API version. It also has a standard for duration, but it looks pretty wired and I haven't seen it in other Rest APIs.

umustafi: Reverted back, I can do the conversion from seconds to human readable message in the client as you are suggesting.

umustafi:
@aplex can you review again and merge

aplex (approved):

PR #3360, opened 2021-08-10
[GOBBLIN-1512] add a path filter to filter  out only files starting with a dot

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1512

### Description
used HiddenFilter to filters out files starting with .

### Tests
added a unit test

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/filters/DotFileFilter.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/filters/DotFileFilter.java
@@ -17,29 +17,15 @@
 
 package org.apache.gobblin.util.filters;
 
-import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 
-
 /**
- * A {@link PathFilter} that filters out hidden files (those starting with '_' or '.').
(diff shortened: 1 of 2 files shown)


gobblin-utility/src/main/java/org/apache/gobblin/util/filters/DotFileFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.gobblin.util.filters;
+
+import org.apache.hadoop.fs.PathFilter;
+
+/**
+ * A {@link PathFilter} similar to {@link HiddenFilter}, but does not filter files starting with '_'.
+ */
+public class DotFileFilter extends HiddenFilter {
```
aplex, line 25: What do you think is better - to have a separate class for dot, or a config setting, where users can override default ignored files in HiddenFilter?

I'm thinking about a couple use cases:
1. People want to copy everything, including hidden files
2. There are other types of files to ignore, like "~something"

Also, I found that we have RegexPathFilter, can users just use that to override filter behaviour?

arjun4084346: Yea I realized that too. closing it.

PR #3361, opened 2021-08-10
[GOBBLIN-1513] fixed the construction of regex filters

arjun4084346:
Dear Gobblin maintainers,

@aplex  please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1513

### Description
Path filters are supposed to be provided by key "gobblin.dataset.path.filter.class", but `instantiatePathFilter` only calls the empty constructors for the path filter classes. Right now only HiddenFilter can be created with empty constructor.
This PR will make a constructor for RegexPathFilter so that it can be used with config "gobblin.dataset.path.filter.class"

### Tests
added a unit test to test initialization and basic working of RegexPathFilter

### Commits

### DIFF
--- a/gobblin-utility/src/main/java/org/apache/gobblin/util/filters/RegexPathFilter.java
+++ b/gobblin-utility/src/main/java/org/apache/gobblin/util/filters/RegexPathFilter.java
@@ -17,6 +17,7 @@
 
 package org.apache.gobblin.util.filters;
 
+import java.util.Properties;
 import java.util.regex.Pattern;
 
 import org.apache.hadoop.fs.Path;
@@ -34,6 +35,11 @@ public class RegexPathFilter implements PathFilter {
 
   private final Pattern regex;
   private final boolean include;
+  public static final String REGEX = "path.filter.regex";
+
+  public RegexPathFilter(Properties props) {
+    this(props.getProperty(REGEX));
+  }
 
   public RegexPathFilter(String regex) {
     this(regex, true);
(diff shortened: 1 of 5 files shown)


gobblin-utility/src/main/java/org/apache/gobblin/util/filters/RegexPathFilter.java
```diff
@@ -34,6 +35,11 @@
 
   private final Pattern regex;
   private final boolean include;
+  public static final String REGEX = "path.filter.regex";
```
aplex, line 38: So, is this path filter only used to filter source datasets?
I can see that we specify the class using gobblin.dataset.path.filter.class . It could make sense to have the same "gobblin.dataset" prefix for this setting for consistency.

Also, it could be better to keep those config settings together, so that they would be easily discovered.

Usually, they land up in ConfigKeys class, but I can see that bunch of keys are defined in DatasetUtils. You can try to figure out if there is any reason to keep them there.

arjun4084346: imo, this and other path filters can be used to filter any path. They are not tied up with filtering only gobblin datasets.
In past, there have been discussion about where to keep configs, a) in ConfigurationKeys or somewhere else along with other configs or b) in the class which actually use them.
For now, I will move it to DatasetUtils

arjun4084346: Ah, I cannot move it to DatasetUtils because that would add a circular dependency between gobblin-utility and gobblin-data-management

aplex (approved):

PR #3365, opened 2021-08-13
[GOBBLIN-1516] Only schedule TaskMetricsUpdateder when metric-reporting is enabled by config

autumnust:
Dear Gobblin maintainers,

### JIRA
https://issues.apache.org/jira/browse/GOBBLIN-1516

### Description
This PR contains two parts: 
- Rename `Optional<ScheduledThreadPoolExecutor> writerMetricsUpdater` to be `writerMetricsUpdateExecutor` which is more accurate. The original name collide with `WriterMetricsUpdater` class which is the actual Runnable to be scheduled. 
- Within `registerNewTask`, only schedule the `TaskMetricsUpdater` if `metrics.enabled` is true. The `TaskMetricsUpdater` holds a reference to `Task` object and `TaskMetricsUpdater` is only terminated/removed after commit. As a result, when there are a large number of tasks being executed, all of them will only be GC'ed after commit phase. This is particularly not friendly to `AsyncDataWriter` when there are data being buffer until commit. For such application, we might consider disable the writer metric reporting to exchange for more efficient heap usage.

### Tests

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTaskStateTracker.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTaskStateTracker.java
@@ -61,7 +61,9 @@ public class GobblinHelixTaskStateTracker extends AbstractTaskStateTracker {
   @Override
   public void registerNewTask(Task task) {
     try {
-      this.scheduledReporters.put(task.getTaskId(), scheduleTaskMetricsUpdater(new TaskMetricsUpdater(task), task));
+      if (GobblinMetrics.isEnabled(task.getTaskState().getWorkunit())) {
+        this.scheduledReporters.put(task.getTaskId(), scheduleTaskMetricsUpdater(new TaskMetricsUpdater(task), task));
+      }
     } catch (RejectedExecutionException ree) {
       // Propagate the exception to caller that has full control of the life-cycle of a helix task.
       log.error(String.format("Scheduling of task state reporter for task %s was rejected", task.getTaskId()));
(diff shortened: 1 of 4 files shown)


autumnust:
@sv2000 @ZihanLi58   Can you take a look ? Thanks.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTaskStateTracker.java
```diff
@@ -61,7 +61,9 @@ public GobblinHelixTaskStateTracker(Properties properties) {
   @Override
   public void registerNewTask(Task task) {
     try {
-      this.scheduledReporters.put(task.getTaskId(), scheduleTaskMetricsUpdater(new TaskMetricsUpdater(task), task));
+      if (GobblinMetrics.isEnabled(task.getTaskState().getWorkunit())) {
+        this.scheduledReporters.put(task.getTaskId(), scheduleTaskMetricsUpdater(new TaskMetricsUpdater(task), task));
```
sv2000, line 65: Using the global "metrics.enabled" key to turn off task metrics reporting seems too heavy. This global config will turn off other reporting as well, e.g. GTEs, and other metrics.

sv2000: I think a better approach might be to not create one metrics updater per task, and create a single updater instead, which iterates over all the tasks and updates records/bytes written.

sv2000 (changes requested):

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixTaskStateTracker.java
autumnust: the additional thing being placed within the condition of `metrics.enabled` to be true, a.k.a `TaskMetricsUpdater` seems to be pretty metrics related and lightweight as it calls `updateRecordMetrics` and `updateByteMetrics`. Both of them are included if instrumented writer is being used, so I believe it is fair to gate these behavior with this config. 

If it is reversed, i.e. the `metrics.enabled` is controlling lightweight metrics-reporting behavior while we are trying to let it control the emission of GTE then I will be concerned as well. Let me know if it makes sense.

sv2000: Agreed. TaskMetrics collection seems un-necessary if metrics.enabled = false.

sv2000 (approved):
+1.

PR #3367, opened 2021-08-16
[GOBBLIN-1518] create getSize api for spec store which is faster than doing getAll a…

arjun4084346:
…nd then returning its size

Dear Gobblin maintainers,

@jack-moseley  @phet @aplex please review

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1518

### Description
During a memory profiling we noticed that a lot of CPU cycles are used in calling getSpecs().size() from metric reporting thread which runs in every n seconds. Right now there is no API to get size of the spec store.
Doing getAllSpecs and then calling size() over it, does a lot of unnecessary work viz reading the content of spec.
This PR will provide a method to direct get the size of the spec store without doing any unnecessary work.

Options

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/InstrumentedSpecStore.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/InstrumentedSpecStore.java
@@ -164,6 +164,18 @@ public abstract class InstrumentedSpecStore implements SpecStore {
     }
   }
 
+  @Override
+  public int getSize() throws IOException {
+    if (!instrumentationEnabled) {
+      return getSizeImpl();
+    } else {
+      long startTimeMillis = System.currentTimeMillis();
+      int size = getSizeImpl();
+      Instrumented.updateTimer(this.getAllTimer, System.currentTimeMillis() - startTimeMillis, TimeUnit.MILLISECONDS);
+      return size;
+    }
+  }
+
   public abstract void addSpecImpl(Spec spec) throws IOException;
   public abstract Spec updateSpecImpl(Spec spec) throws IOException, SpecNotFoundException;
   public abstract boolean existsImpl(URI specUri) throws IOException;
@@ -171,6 +183,7 @@ public abstract class InstrumentedSpecStore implements SpecStore {
   public abstract boolean deleteSpecImpl(URI specUri) throws IOException;
   public abstract Collection<Spec> getSpecsImpl() throws IOException;
   public abstract Iterator<URI> getSpecURIsImpl() throws IOException;
+  public abstract int getSizeImpl() throws IOException;
 
   /** child classes can implement this if they want to get specs using {@link SpecSearchObject} */
   public Collection<Spec> getSpecsImpl(SpecSearchObject specUri) throws IOException {
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecCatalog.java
@@ -51,6 +51,8 @@ public interface SpecCatalog extends SpecCatalogListenersContainer, Instrumentab
    * */
   Collection<Spec> getSpecs();
 
+  int getSize();
+
   /** Metrics for the spec catalog; null if
    * ({@link #isInstrumentationEnabled()}) is false. */
   SpecCatalog.StandardMetrics getMetrics();
@@ -110,7 +112,7 @@ public interface SpecCatalog extends SpecCatalogListenersContainer, Instrumentab
       this.totalUpdatedSpecs = new AtomicLong(0);
       this.numActiveSpecs = metricsContext.newContextAwareGauge(NUM_ACTIVE_SPECS_NAME,  ()->{
           long startTime = System.currentTimeMillis();
-          int size = specCatalog.getSpecs().size();
+          int size = specCatalog.getSize();
           updateGetSpecTime(startTime);
           return size;
       });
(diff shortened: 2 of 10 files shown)


gobblin-runtime/src/test/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalogTest.java
```diff
@@ -161,6 +161,7 @@ public void createFlowSpec() {
       logger.info("[After Create] Spec " + i++ + ": " + gson.toJson(flowSpec));
     }
     Assert.assertTrue(specs.size() == 1, "Spec store should contain 1 Spec after addition");
```
umustafi, line 163: nit: for code quality should we mirror these assertions to use either assertEquals or assertTrue? I prefer the former (the way you did it)

gobblin-runtime/src/test/java/org/apache/gobblin/runtime/spec_catalog/FlowCatalogTest.java
```diff
@@ -191,6 +192,7 @@ public void deleteFlowSpec() throws SpecNotFoundException {
       logger.info("[After Delete] Spec " + i++ + ": " + gson.toJson(flowSpec));
     }
     Assert.assertTrue(specs.size() == 0, "Spec store should be empty after deletion");
+    Assert.assertEquals(flowCatalog.getSize(), 0, "Spec store should be empty after deletion");
```
umustafi, line 195: in these places we can change them to assertEquals as well

umustafi (changes requested):

aplex (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecCatalog.java
```diff
@@ -110,7 +112,7 @@ public StandardMetrics(final SpecCatalog specCatalog, Optional<Config> sysConfig
       this.totalUpdatedSpecs = new AtomicLong(0);
       this.numActiveSpecs = metricsContext.newContextAwareGauge(NUM_ACTIVE_SPECS_NAME,  ()->{
           long startTime = System.currentTimeMillis();
-          int size = specCatalog.getSpecs().size();
+          int size = specCatalog.getSize();
```
phet, line 115: incredible!  when we spoke, I had imagined the copious deserialization to originate in the need to  peek inside the `Spec` (e.g. to selectively count by filtering in java-space).

I'm absolutely floored that sampling showed this (teeny-tiny) invocation to occupy 42% of CPU time (on the leader) while accounting for 77.5% of memory allocations

excellent work, arjun!

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/SpecCatalog.java
```diff
@@ -51,6 +51,8 @@
    * */
```
phet, line 51: prophetic comment, if there ever was one...

now that you've reworked the metrics gauge, the only remaining use seems to be in managing listeners.  AFAICT (in my budding understanding) that appears reasonable, but might be worth you confirming.

(presuming this optimization bears the copious fruit we hope for) shall we return to update this comment, to scare off casually-considered use?  although perhaps a slight abusage of `@Deprecated` we could shout that caution as a build-time warning.  what's your take?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_catalog/TopologyCatalog.java
```diff
@@ -232,6 +232,15 @@ public void switchMetricContext(MetricContext context) {
...
   }
 
+  @Override
+  public int getSize() {
+    try {
+      return specStore.getSize();
+    } catch (IOException e) {
+      throw new RuntimeException("Cannot retrieve number of specs from Spec store", e);
+    }
+  }
```
phet, line 242: [thought for later] literal duplication (in each containing class):  perhaps explore a `CheckedExceptionWrappingSpecStore`, basically a Decorator for centralizing this?  doing so lends similar benefit as the `InstrumentedSpecStore` by encapsulating timing in one place (although based on composition rather than impl inheritance, as ISS is).

phet (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/api/InstrumentedSpecStore.java
```diff
@@ -164,13 +164,26 @@ public Spec updateSpec(Spec spec) throws IOException, SpecNotFoundException {
...
   }
 
+  @Override
+  public int getSize() throws IOException {
+    if (!instrumentationEnabled) {
+      return getSizeImpl();
+    } else {
+      long startTimeMillis = System.currentTimeMillis();
+      int size = getSizeImpl();
+      Instrumented.updateTimer(this.getAllTimer, System.currentTimeMillis() - startTimeMillis, TimeUnit.MILLISECONDS);
```
phet, line 174: looking at something else in this same file, I noticed this (post-merge)...

wouldn't we want a separate `Timer`, to avoid mixing durations from two different operations?  I can include that change in what I'm working on now, if you agree @arjun4084346

arjun4084346: Ah. will fix in the next PR. thank you for catching

phet: cool, no worries.   I added it along with my recent PR already in progress - https://github.com/apache/gobblin/pull/3369

PR #3370, opened 2021-08-19
[GOBBLIN-1520] Standardize FsSpec's protocol in file name

autumnust:
Dear Gobblin maintainers,

### JIRA
 https://issues.apache.org/jira/browse/GOBBLIN-1520

### Description
This PR standardize the file format used in `FsSpecProducer` and `FsSpecConsumer`, by persisting all `Specs` in a file with ".avro" as extension file and only load files with the same extension in consumer side. 
Also added related unit tests. 

### Tests

### Commits

### DIFF
--- a/gobblin-runtime/src/test/java/org/apache/gobblin/runtime/api/FsSpecProducerTest.java
+++ b/gobblin-runtime/src/test/java/org/apache/gobblin/runtime/api/FsSpecProducerTest.java
@@ -42,13 +42,15 @@ public class FsSpecProducerTest {
   private FsSpecProducer _fsSpecProducer;
   private FsSpecConsumer _fsSpecConsumer;
   private Config _config;
+  private File workDir;
 
   @BeforeMethod
   public void setUp()
       throws IOException {
-    File tmpDir = Files.createTempDir();
+    this.workDir = Files.createTempDir();
+    this.workDir.deleteOnExit();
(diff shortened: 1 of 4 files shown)


gobblin-runtime/src/test/java/org/apache/gobblin/runtime/api/FsSpecProducerTest.java
```diff
@@ -106,4 +131,19 @@ public void testDeleteSpec() throws URISyntaxException, ExecutionException, Inte
     Assert.assertEquals(jobSpecs.get(0).getLeft(), SpecExecutor.Verb.DELETE);
     Assert.assertEquals(jobSpecs.get(0).getRight().getUri().toString(), "testDeleteJob");
   }
+
+  class TestingFsSpecProducer extends FsSpecProducer {
```
sv2000, line 135: Can we simply write a dummy file (e.g. using File API) into the spec path without having to extend FsSpecProducer class? We can avoid having to make the annotateSpecFilename method package-private.

sv2000 (changes requested):

gobblin-runtime/src/test/java/org/apache/gobblin/runtime/api/FsSpecProducerTest.java
autumnust: Sure, I just realize `tmpDir` is available in the test. Addressed.

sv2000 (approved):
+1.

PR #3371, opened 2021-08-19
[GOBBLIN-1457] Fix service DB initialization when DB is not empty

aplex:
By default, Flyway will throw an exception if the database is not
empty and Flyway's own table is missing. Production Gobblin service
DBs already have state tables in them, so migration didn't work.

With this change, we ask Flyway to ignore existing tables when
initializing new db.


sv2000 (approved):

sv2000:
+1.

PR #3372, opened 2021-08-20
[GOBBLIN-1521] Create local mode of streaming kafka job to help user quickly onboard

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1521

### Description
Using embedded gobblin to run streaming job, and create a guid to help user quickly onboard
* Change kafka09ConsumerClient to make get earliest and latest offset won't change the current position of the consumer, and support assignAndSeek() method
* Change FsStateStore rename method to enable overwrite so that we can write to existing state store
* Change reference in kafka-hdfs-streaming-avro.job to OSS class to give user better idea about how the config looks like if they want to ingest as avro, although we don't use that config in this local job since we don't have schema registry setup.  
* change KafkaStreamingExtractor to support the scenario where schemaRegistry is not setup

### Tests
Test we can run the job locally

### Commits

### DIFF
--- a/gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
+++ b/gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
@@ -1,0 +1,30 @@
+# Table of Contents
+
+[TOC]
+
+# Introduction
+
+Gobblin streaming enable user to ingest data from Kafka to HDFS in streaming manner. Normally it's running on yarn, to get the full function, 
+you need have helix cluster, yarn cluster, kafka server and schema registry set up. But we also have one local mode of streaming task using EmbeddedGobblin
+for user to easily start and test out the job. Here is the steps:
+
+# Setup local kafka cluster 
+
+Just follow [kafka quick start](https://kafka.apache.org/quickstart) to set up your kafka cluster, and create test topic "testEvents"
+
+# Run EmbeddedGobblin to start the job
+
+We are using configuration: /gobblin-modules/gobblin-kafka-09/src/test/resources/kafkaHDFSStreaming.conf to execute the job.
+
+To run the job, in your intellij, you can run the test in /gobblin-modules/gobblin-kafka-09/src/test/java/org/apache/gobblin/kafka/KafkaStreamingLocalTest
+by remove the line '(enabled=false)'. In order to run the test in IDE, you may need to manually delete log4j-over-slf4j jar in IDE 
+
+Under your kafka dir, you can run following command to produce data into your kafka topic
+
+`bin/kafka-console-producer.sh --topic testEvents --bootstrap-server localhost:9092`
+
+The job will continually consume from testEvents and write out data as txt file onto your local fileSystem (/tmp/gobblin/kafka/publish). It will write put data every 60 seconds, and will never end until
+you manually kill it.
+
+If you want the job ingest data as avro/orc, you will need to have schema registry as schema source and change the job configuration to control the behavior
+
--- a/gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
+++ b/gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
@@ -26,5 +26,5 @@ Under your kafka dir, you can run following command to produce data into your ka
 The job will continually consume from testEvents and write out data as txt file onto your local fileSystem (/tmp/gobblin/kafka/publish). It will write put data every 60 seconds, and will never end until
 you manually kill it.
 
-If you want the job ingest data as avro/orc, you will need to have schema registry as schema source and change the job configuration to control the behavior
+If you want the job ingest data as avro/orc, you will need to have schema registry as schema source and change the job configuration to control the behavior, a sample configuration can be found [here](https://github.com/apache/gobblin/blob/master/gobblin-modules/gobblin-azkaban/src/main/resources/conf/gobblin_jobs/kafka-hdfs-streaming-avro.conf)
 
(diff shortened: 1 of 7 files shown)


ZihanLi58:
@sv2000 @autumnust can you help take a look at the change when you have time? Thanks

gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
```diff
@@ -0,0 +1,30 @@
+# Table of Contents
+
+[TOC]
+
+# Introduction
+
+Gobblin streaming enable user to ingest data from Kafka to HDFS in streaming manner. Normally it's running on yarn, to get the full function,
```
sv2000, line 7: Reword to: "Gobblin supports streaming mode that allows continuous ingestion of data from Kafka to HDFS. The streaming mode has been deployed in production at LinkedIn as a Gobblin cluster that uses Yarn for container allocation and Helix for task coordination. Here, we describe how to set up a Kafka -> HDFS pipeline in local mode for users to easily start and test out a streaming ingestion pipeline. "

gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Introduction
+
+Gobblin streaming enable user to ingest data from Kafka to HDFS in streaming manner. Normally it's running on yarn, to get the full function, 
+you need have helix cluster, yarn cluster, kafka server and schema registry set up. But we also have one local mode of streaming task using EmbeddedGobblin
+for user to easily start and test out the job. Here is the steps:
+
+# Setup local kafka cluster 
+
+Just follow [kafka quick start](https://kafka.apache.org/quickstart) to set up your kafka cluster, and create test topic "testEvents"
```
sv2000, line 13: Drop "Just".

gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Run EmbeddedGobblin to start the job
+
+We are using configuration: /gobblin-modules/gobblin-kafka-09/src/test/resources/kafkaHDFSStreaming.conf to execute the job.
```
sv2000, line 17: "We are using" -> "We use the"

gobblin-docs/user-guide/Run-Gobblin-Streaming-kafka-hdfs-Locally.md
```diff
@@ -0,0 +1,30 @@
...
+
+To run the job, in your intellij, you can run the test in /gobblin-modules/gobblin-kafka-09/src/test/java/org/apache/gobblin/kafka/KafkaStreamingLocalTest
+by remove the line '(enabled=false)'. In order to run the test in IDE, you may need to manually delete log4j-over-slf4j jar in IDE
```
sv2000, line 20: remove -> removing.

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
```diff
@@ -323,8 +323,10 @@ public KafkaStreamingExtractor(WorkUnitState state) {
   @Override
   public S getSchema() {
     try {
-      Schema schema = (Schema) this._schemaRegistry.get().getLatestSchemaByTopic(this.topicPartitions.get(0).getTopicName());
-      return (S) schema;
+      if(this._schemaRegistry.isPresent()) {
+        return (S)(Schema) this._schemaRegistry.get().getLatestSchemaByTopic(this.topicPartitions.get(0).getTopicName());
+      }
+      return (S) this.topicPartitions.iterator().next().getTopicName();
```
sv2000, line 329: return null? Don't understand why we are returning topic name.

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/FsStateStore.java
```diff
@@ -217,7 +217,7 @@ public void putAll(String storeName, String tableName, Collection<T> states) thr
   }
 
   protected void renamePath(Path tmpTablePath, Path tablePath) throws IOException {
-    HadoopUtils.renamePath(this.fs, tmpTablePath, tablePath);
+    HadoopUtils.renamePath(this.fs, tmpTablePath, tablePath, true);
```
sv2000, line 220: We are changing the behavior of renamePath(). I am assuming you need this to overwrite the watermarks.

sv2000 (changes requested):

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
ZihanLi58: It's because many converter does not accept null in convertSchema() method. And simply return null will cause issue there. I just use the same behavior as KafkaSimpleStreamingExtractor where if the schema registry is null, we just return topic name as schema

gobblin-metastore/src/main/java/org/apache/gobblin/metastore/FsStateStore.java
ZihanLi58: Yes, Fs state store does not accept add/update watermark into an existing store now, but in our case, we need to continually publish new file and update watermark so I change the behavior to make it able to overwrite existing state file.  Do you have any concern for this change? i.e we have pipeline that rely on this feature? If so, I can also make this behavior configurable

PR #3381, opened 2021-09-01
[GOBBLIN-1530] Add error reporting for missing source nodes that may not exist

Will-Lo:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1530

### Description

When copying data from a source node that does not exist/fails to be added to GaaS FlowGraph, a very cryptic error message pops up where DmasAuthorizer shows a NullPointerException.

This PR guards against source nodes that do not exist so that the error is much easier for developers and users to understand.

### Tests

### Commits

### DIFF
--- a/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
+++ b/gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
@@ -190,6 +190,11 @@ public class MultiHopFlowCompiler extends BaseFlowToJobSpecCompiler {
     String destination = ConfigUtils.getString(flowSpec.getConfig(), ServiceConfigKeys.FLOW_DESTINATION_IDENTIFIER_KEY, "");
 
     DataNode sourceNode = this.flowGraph.getNode(source);
+    if (sourceNode == null) {
+      flowSpec.getCompilationErrors()
+          .add(String.format("Flowgraph does not have a node with id %s", source));
+      return null;
+    }
     List<String> destNodeIds = ConfigUtils.getStringList(flowSpec.getConfig(), ServiceConfigKeys.FLOW_DESTINATION_IDENTIFIER_KEY);
     List<DataNode> destNodes = destNodeIds.stream().map(this.flowGraph::getNode).collect(Collectors.toList());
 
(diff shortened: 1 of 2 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompiler.java
```diff
@@ -190,6 +190,11 @@ public void awaitHealthy() throws InterruptedException {
     String destination = ConfigUtils.getString(flowSpec.getConfig(), ServiceConfigKeys.FLOW_DESTINATION_IDENTIFIER_KEY, "");
 
     DataNode sourceNode = this.flowGraph.getNode(source);
+    if (sourceNode == null) {
+      flowSpec.getCompilationErrors()
+          .add(String.format("Flowgraph does not have a node with id %s", source));
+      return null;
+    }
     List<String> destNodeIds = ConfigUtils.getStringList(flowSpec.getConfig(), ServiceConfigKeys.FLOW_DESTINATION_IDENTIFIER_KEY);
     List<DataNode> destNodes = destNodeIds.stream().map(this.flowGraph::getNode).collect(Collectors.toList());
```
jack-moseley, line 199: The same error happens if the destination node is invalid, so we should have validation on this one as well.

gobblin-service/src/test/java/org/apache/gobblin/service/modules/flow/MultiHopFlowCompilerTest.java
```diff
@@ -670,6 +670,28 @@ public void testUnresolvedFlow() throws Exception {
...
+
+    Dag<JobExecutionPlan> dag = specCompiler.compileFlow(spec);
+
+    Assert.assertEquals(dag, null);
+    Assert.assertEquals(spec.getCompilationErrors().size(), 1);
+    spec.getCompilationErrors().stream().anyMatch(s -> s.contains("Flowgraph does not have a node with id"));
+  }
+
+  @Test (dependsOnMethods = "testMissingSourceNodeError")
+  public void testMissingDestinationNodeError() throws Exception {
```
jack-moseley, line 685: Am I missing something, or shouldn't this require validation on the `List<DataNode> destNodes` object? Right now it's only validating the `DataNode sourceNode` right?

Will-Lo: Ah my bad I added that way too fast, good catch

Will-Lo: @jack-moseley you're right, it reports the error correctly on local which is what I tested, but when we add our dmas authorizer it runs into the nullpointer exception before it does another check when calculating the path, I'll add the error check now.

jack-moseley (approved):

Will-Lo:
@aplex Could you merge?

PR #3382, opened 2021-09-02
[GOBBLIN-1527] Add finder `latestFlowGroupExecutions` to `FlowExecutions` endpoint.

phet:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1527
    "Extend REST `flowexecutions` endpoint with finder `latestFlowGroupExecutions`"

### Description

This adds the new endpoint while augmenting both (primary) forms of `JobStatusRetriever` (Mysql and FS) to support querying flow executions across a flow group.

It leverages the recent refactoring of https://github.com/apache/gobblin/pull/3378

### Tests

Included enhancements of existing unit tests cover added code paths.

### Commits

### DIFF
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
@@ -63,6 +63,12 @@ public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatus
     return this.flowExecutionResourceHandler.getLatestFlowExecution(context, flowId, count, tag, executionStatus);
   }
 
+  @Finder("latestFlowGroupExecutions")
+  public List<FlowExecution> getLatestFlowGroupExecutions(@Context PagingContext context, @QueryParam("flowGroup") String flowGroup,
+      @Optional @QueryParam("count") Integer count, @Optional @QueryParam("tag") String tag) {
+    return this.flowExecutionResourceHandler.getLatestFlowGroupExecutions(context, flowGroup, count, tag);
+  }
+
   /**
    * Resume a failed {@link FlowExecution} from the point before failure.
    * @param pathKeys key of {@link FlowExecution} specified in path
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
@@ -65,8 +65,8 @@ public class FlowExecutionResource extends ComplexKeyResourceTemplate<FlowStatus
 
   @Finder("latestFlowGroupExecutions")
   public List<FlowExecution> getLatestFlowGroupExecutions(@Context PagingContext context, @QueryParam("flowGroup") String flowGroup,
-      @Optional @QueryParam("count") Integer count, @Optional @QueryParam("tag") String tag) {
-    return this.flowExecutionResourceHandler.getLatestFlowGroupExecutions(context, flowGroup, count, tag);
+      @Optional @QueryParam("countPerFlow") Integer countPerFlow, @Optional @QueryParam("tag") String tag) {
+    return this.flowExecutionResourceHandler.getLatestFlowGroupExecutions(context, flowGroup, countPerFlow, tag);
   }
 
   /**
--- a/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceHandler.java
+++ b/gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceHandler.java
@@ -37,6 +37,14 @@ public interface FlowExecutionResourceHandler {
   public List<FlowExecution> getLatestFlowExecution(PagingContext context, FlowId flowId, Integer count, String tag, String executionStatus);
 
   /**
+   * Get latest {@link FlowExecution} for every flow in `flowGroup`
+   *
+   * NOTE: `executionStatus` param not provided yet, without justifying use case, due to complexity of interaction with `count`
+   * and resulting efficiency concern of performing across many flows sharing the single named group.
+   */
+  public List<FlowExecution> getLatestFlowGroupExecutions(PagingContext context, String flowGroup, Integer count, String tag);
+
+  /**
    * Resume a failed {@link FlowExecution} from the point before failure
    */
   public void resume(ComplexResourceKey<FlowStatusId, EmptyRecord> key);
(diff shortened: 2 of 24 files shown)


gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/FsJobStatusRetriever.java
```diff
@@ -83,7 +84,7 @@ public FsJobStatusRetriever(Config config, MultiContextIssueRepository issueRepo
       }
       return jobStatuses.iterator();
     } catch (IOException e) {
-      log.error("IOException encountered when retrieving job statuses for flow: {},{},{}", flowGroup, flowName, flowExecutionId, e);
+      log.error(String.format("IOException encountered when retrieving job statuses for flow: %s,%s,%s", flowGroup, flowName, flowExecutionId), e);
```
phet, line 87: I changed this because my understanding of signature, `(String, Object...)`, is that it treats all trailing args as format string interpolations, not recognizing that the final one might be `? extends Throwable`.  thus it won't print the stacktrace, as `(String, Throwable)` would.  if that's incorrect, or I misinterpreted the intent here, let's revert these lines.

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResourceHandler.java
```diff
@@ -36,6 +36,14 @@
    */
   public List<FlowExecution> getLatestFlowExecution(PagingContext context, FlowId flowId, Integer count, String tag, String executionStatus);
 
+  /**
+   * Get latest {@link FlowExecution} for every flow in `flowGroup`
+   *
+   * NOTE: `executionStatus` param not provided yet, without justifying use case, due to complexity of interaction with `count`
+   * and resulting efficiency concern of performing across many flows sharing the single named group.
+   */
+  public List<FlowExecution> getLatestFlowGroupExecutions(PagingContext context, String flowGroup, Integer count, String tag);
```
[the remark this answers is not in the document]
phet: this count is semantically the maximum number of executions for each flow found within the group, not the total number of `FlowExecution`s to return.  that said, following the lead of `getLatestFlowExecution` (directly above), which also ignores it, I never inspect or utilize the `PagingContext`.  I presumed it to be a mandated param, but if it would be legal to omit it, that would be a clearer statement of intent (to ignore).  do you know if I could?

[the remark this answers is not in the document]
aplex: How many results can we return in this response? There could be 100k flows, and if we return all of them in one response, it might take too long to generate it and the request will time out.

If this count is not about the number of flowExecutions, then can we name it differently to make this more clear? Like "maxExecutionsPerFlow" or something like that?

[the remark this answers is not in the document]
phet: as we agreed offline, I've renamed `count` to `countPerFlow` (on the `latestFlowGroupExecutions` finder).

AFA potential to time out from a large number of historical flow executions (within the group), there is the `DatasetCleaner`, scheduled by `KafkaJobStatusMonitor` to run in the background.  that can constrain historical flow executions from accumulating forever.  when the config property named by `MultiCleanableDatasetFinder.datasetFinderClassKey` is set to
`org.apache.gobblin.data.management.retention.dataset.finder.TimeBasedDatasetStoreDatasetFinder`, the latter will scan the `MysqlJobStatusStateStore` to offer the contents up for retention/expiration.

so while there could still be lots of prior executions, there are constraints on accumulation.  I propose we gain operational experience to determine whether there's a need to truncate or page through the results: in practice we may not be in danger of timing out (but I agree with you that it's worth keeping an eye out for).

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
```diff
@@ -63,6 +63,12 @@ public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
     return this.flowExecutionResourceHandler.getLatestFlowExecution(context, flowId, count, tag, executionStatus);
   }
 
+  @Finder("latestFlowGroupExecutions")
+  public List<FlowExecution> getLatestFlowGroupExecutions(@Context PagingContext context, @QueryParam("flowGroup") String flowGroup,
```
arjun4084346, line 67: Should we rather just add an @Optional @QueryParam to the existing getLatestFlowExecution finder method?

phet: I wanted to differentiate them semantically with a separate name for each finder.  the other issue with making the flow name optional on the existing finder is that it is not top-level, but a component of the `flowId` compound param.

arjun4084346: I see, then should we have changed getLatestFlowExecution to accept two separate params flowGroup and flowName , instead of accepting FlowId ?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
```diff
@@ -63,6 +63,12 @@ public FlowExecution get(ComplexResourceKey<FlowStatusId, EmptyRecord> key) {
...
+      @Optional @QueryParam("countPerFlow") Integer countPerFlow, @Optional @QueryParam("tag") String tag) {
```
arjun4084346, line 68: Should this new finder also accept 'executionStatus' as another param ?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
phet: that change would break b/w compatibility for existing clients.  also, the existing name is misleading as `latestFlowExecution`, if it would return multiple executions across all flows of the group.

plus, after my latest update to this PR, at alex's suggestion, naming of corresponding params has diverged across the two (i.e. `count` and `countPerFlow`).

do you have a particular concern about adding a new finder?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-server/src/main/java/org/apache/gobblin/service/FlowExecutionResource.java
phet: I punted on that... and was unsure with all the layering where to put javadoc to that effect.  finding more on the `FlowExecutionResourceHandler` than here, I added this note in that location - 

```
   * NOTE: `executionStatus` param not provided yet, without justifying use case, due to complexity of interaction with `count`
   * and resulting efficiency concern of performing across many flows sharing the single named group.
```
so basically I was waiting for a use case.  I do note that the `executionStatus` on `latestFlowExecution` remains unexercised by our existing CLI client, so I definitely didn't want to pursue replicating something that itself may see little usage.

do note that if we decide to add this param here later, it should preserve b/w compatibility on this finder.

arjun4084346: Okay

aplex (approved):

PR #3385, opened 2021-09-03
[GOBBLIN-1533] Add completeness watermark to iceberg tables

vikrambohra:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1533

### Description
- Added a new module "gobblin-completeness" and moved audit count APIs from gobblin-compaction here.
- Added KafkaAuditCountVerifier that can be used to compare audit counts regardless of whether its a compaction or ingestion pipeline
- Added logic to compute late partition for iceberg data files in IcebergMetadataWriter and logic to update the completeness watermark iceberg table property

### Tests
- Added test case testWriteAddFileGMCECompleteness to IcebergMetadataWriterTest 

### Commits


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -616,23 +689,75 @@ protected void addFiles(GobblinMetadataChangeEvent gmce, Map<String, Collection<
...
     Set<DataFile> dataFiles = new HashSet<>();
     for (org.apache.gobblin.metadata.DataFile file : files) {
       try {
-        StructLike partition = getIcebergPartitionVal(newSpecsMap.get(new Path(file.getFilePath()).getParent().toString()),
-            file.getFilePath(), partitionSpec);
-        dataFiles.add(IcebergUtils.getIcebergDataFileWithMetric(file, partitionSpec, partition, conf, schemaIdMap));
+        Collection<HiveSpec> hiveSpecs = newSpecsMap.get(new Path(file.getFilePath()).getParent().toString());
+        StructLike partition = getIcebergPartitionVal(hiveSpecs, file.getFilePath(), partitionSpec);
+
+        if(tableMetadata.completenessEnabled && gmce.getOperationType() == OperationType.add_files) {
```
ZihanLi58, line 704: We only have late partition for hourly data? and for daily data, we won't have this partition? Will this cause some problem?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -616,23 +689,75 @@ protected void addFiles(GobblinMetadataChangeEvent gmce, Map<String, Collection<
...
-  private Set<DataFile> getIcebergDataFilesToBeAdded(List<org.apache.gobblin.metadata.DataFile> files,
+  private Set<DataFile> getIcebergDataFilesToBeAdded(Table table, TableMetadata tableMetadata, GobblinMetadataChangeEvent gmce, List<org.apache.gobblin.metadata.DataFile> files,
       PartitionSpec partitionSpec, Map<String, Collection<HiveSpec>> newSpecsMap, Map<Integer, Integer> schemaIdMap) {
     Set<DataFile> dataFiles = new HashSet<>();
     for (org.apache.gobblin.metadata.DataFile file : files) {
       try {
-        StructLike partition = getIcebergPartitionVal(newSpecsMap.get(new Path(file.getFilePath()).getParent().toString()),
-            file.getFilePath(), partitionSpec);
-        dataFiles.add(IcebergUtils.getIcebergDataFileWithMetric(file, partitionSpec, partition, conf, schemaIdMap));
+        Collection<HiveSpec> hiveSpecs = newSpecsMap.get(new Path(file.getFilePath()).getParent().toString());
```
ZihanLi58, line 701: For this part, I feel we have many assumptions, i.e. we only have one partition column, and it's datepartition. But it's not always hold true and GMIP was supposed to support all generic use cases. Do you think it's make sense to have another icebergWriterWithCompletenessCheck for this new feature?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -616,23 +689,75 @@ protected void addFiles(GobblinMetadataChangeEvent gmce, Map<String, Collection<
...
+   * 1. Add "late" partition column to iceberg table if not exists
+   * 2. compute "late" partition value based on datepartition and completion watermark
+   * @param table
+   * @param tableMetadata
+   * @param hivePartition
+   * @param datepartition
+   * @return new iceberg partition value for file
+   */
+  private StructLike addLatePartitionValueToIcebergTable(Table table, TableMetadata tableMetadata, HivePartition hivePartition, String datepartition) {
+    table = addPartitionToIcebergTable(table, newPartitionColumn, newPartitionColumnType);
```
ZihanLi58, line 730: Why we call this again? shouldn't this be handled at very beginning of the write?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -616,23 +689,75 @@ protected void addFiles(GobblinMetadataChangeEvent gmce, Map<String, Collection<
...
+  /**
+   * 1. Add "late" partition column to iceberg table if not exists
+   * 2. compute "late" partition value based on datepartition and completion watermark
+   * @param table
+   * @param tableMetadata
+   * @param hivePartition
+   * @param datepartition
+   * @return new iceberg partition value for file
+   */
+  private StructLike addLatePartitionValueToIcebergTable(Table table, TableMetadata tableMetadata, HivePartition hivePartition, String datepartition) {
```
ZihanLi58, line 729: Can we move these functions to Util class?

ZihanLi58 (changes requested):
First round of review, have several comments there

gobblin-iceberg/src/test/java/org/apache/gobblin/iceberg/writer/HiveMetadataWriterTest.java
```diff
@@ -178,7 +178,7 @@ public void setUp() throws Exception {
     gobblinMCEWriter = new GobblinMCEWriter(new GobblinMCEWriterBuilder(), state);
   }
 
-  @Test( priority = 3 )
+  @Test( priority = 4 )
```
sv2000, line 181: If this test case depends on another test case, we should use "dependsOnMethod", for better readability.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -148,6 +158,29 @@
   /* one of the fields in DataFile entry to describe the location URI of a data file with FS Scheme */
   private static final String ICEBERG_FILE_PATH_COLUMN = DataFile.FILE_PATH.name();
 
+  public static final String ICEBERG_COMPLETENESS_ENABLED = "iceberg.completeness.enabled";
+  private static final boolean DEFAULT_ICEBERG_COMPLETENESS = false;
+  private final boolean completenessEnabled;
+  public static final String ICEBERG_COMPLETENESS_WHITELIST = "iceberg.completeness.whitelist";
```
sv2000, line 164: Can we refactor the code and move all the config keys for the metadata writer to a separate class?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -422,6 +480,21 @@ private void computeCandidateSchema(GobblinMetadataChangeEvent gmce, TableIdenti
...
+  /**
+   * Add a partition column to the schema and partition spec
+   * @param table incoming iceberg table
+   * @param fieldName name of partition column
+   * @param type datatype of partition column
+   * @return table with updated schema and partition spec
+   */
+  private Table addPartitionToIcebergTable(Table table, String fieldName, String type) {
+    if(!table.schema().columns().stream().anyMatch(x -> x.name().equalsIgnoreCase(fieldName))) {
+      table.updateSchema().addColumn(fieldName, Types.fromPrimitiveString(type)).commit();
```
sv2000, line 492: Are we enforcing string type for all partition columns? So late values will be string type as opposed to int?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -731,6 +871,34 @@ public void flush(String dbName, String tableName) throws IOException {
...
+        if (timestamp > previousWatermark) {
+          if(auditCountVerifier.get().isComplete(table, timestamp, timestamp + TimeUnit.HOURS.toMillis(1))) {
+            completionWatermark = timestamp;
+          }
+        } else {
+          break;
+        }
+      }
+    } catch (IOException e) {
+      log.error(ExceptionUtils.getFullStackTrace(e));
```
sv2000, line 897: Is it intentional to swallow the exception?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -731,6 +871,34 @@ public void flush(String dbName, String tableName) throws IOException {
...
+   * @param table
+   * @param timestamps
+   * @param previousWatermark
+   * @return updated completion watermark
+   */
+  private long computeCompletenessWatermark(String table, Collection<Long> timestamps, long previousWatermark) {
+    log.info(String.format("Compute completion watermark for %s and timestamps %s with previous watermark %s", table, timestamps, previousWatermark));
+    long completionWatermark = previousWatermark;
+    try {
+      for(long timestamp : timestamps) {
```
sv2000, line 887: Can we sort the timestamp in decreasing order and check for the first timestamp for which the audit counts match? Would make the code simpler and easier to read.
Also: I think this should be done as a binary search to minimize calls to audit system.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -148,6 +158,29 @@
   /* one of the fields in DataFile entry to describe the location URI of a data file with FS Scheme */
   private static final String ICEBERG_FILE_PATH_COLUMN = DataFile.FILE_PATH.name();
 
+  public static final String ICEBERG_COMPLETENESS_ENABLED = "iceberg.completeness.enabled";
+  private static final boolean DEFAULT_ICEBERG_COMPLETENESS = false;
+  private final boolean completenessEnabled;
+  public static final String ICEBERG_COMPLETENESS_WHITELIST = "iceberg.completeness.whitelist";
+  public static final String ICEBERG_COMPLETENESS_BLACKLIST = "iceberg.completeness.blacklist";
+  private final WhitelistBlacklist completenessWhitelistBlacklist;
+  public static final String COMPLETION_WATERMARK_KEY = "completion.watermark";
```
sv2000, line 167: "completionWatermark" instead of completion.watermark? At LI, we already have a precedence of using the former.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -148,6 +158,29 @@
...
+  public static final String COMPLETION_WATERMARK_TIMEZONE_KEY = "completion.watermark.timezone";
```
sv2000, line 168: Similarly, completionWatermarkTimezone

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -671,14 +796,29 @@ public void flush(String dbName, String tableName) throws IOException {
...
           tableMetadata.appendFiles.get().commit();
+          if(tableMetadata.completenessEnabled) {
+            String topicName = props.get(TOPIC_NAME_KEY);
+            if(topicName == null) {
+              log.error(String.format("Not performing audit check. %s is null. Please set as table property of %s.%s",
+                  TOPIC_NAME_KEY, dbName, tableName));
+            } else {
+              long newCompletenessWatermark =
+                  computeCompletenessWatermark(topicName, tableMetadata.datePartitions, tableMetadata.prevCompletenessWatermark);
+              if(newCompletenessWatermark != tableMetadata.prevCompletenessWatermark) {
```
sv2000, line 811: I think we should assert that the new watermark is always >= previous watermark to ensure monotonicity and avoid accidentally moving watermarks backwards.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -731,6 +871,34 @@ public void flush(String dbName, String tableName) throws IOException {
...
+   * @param previousWatermark
+   * @return updated completion watermark
+   */
+  private long computeCompletenessWatermark(String table, Collection<Long> timestamps, long previousWatermark) {
+    log.info(String.format("Compute completion watermark for %s and timestamps %s with previous watermark %s", table, timestamps, previousWatermark));
+    long completionWatermark = previousWatermark;
+    try {
+      for(long timestamp : timestamps) {
+        if (timestamp > previousWatermark) {
+          if(auditCountVerifier.get().isComplete(table, timestamp, timestamp + TimeUnit.HOURS.toMillis(1))) {
```
sv2000, line 889: Let's not assume hourly completeness checks and make the window granularity configurable with a default of 1 hour.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -671,14 +796,29 @@ public void flush(String dbName, String tableName) throws IOException {
...
+            Maps.newHashMap(tableMetadata.lastProperties.or(getIcebergTable(tid).properties())));
         if (tableMetadata.appendFiles.isPresent()) {
           tableMetadata.appendFiles.get().commit();
+          if(tableMetadata.completenessEnabled) {
+            String topicName = props.get(TOPIC_NAME_KEY);
+            if(topicName == null) {
+              log.error(String.format("Not performing audit check. %s is null. Please set as table property of %s.%s",
+                  TOPIC_NAME_KEY, dbName, tableName));
+            } else {
+              long newCompletenessWatermark =
```
sv2000, line 809: Calling this on each flush may be excessive. If the table is complete upto the most recent hour (or whatever the window granularity is), then we should skip the check until the end of the hour.

sv2000 (changes requested):
First round of comments.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -148,6 +158,28 @@
   /* one of the fields in DataFile entry to describe the location URI of a data file with FS Scheme */
   private static final String ICEBERG_FILE_PATH_COLUMN = DataFile.FILE_PATH.name();
 
+  public static final String ICEBERG_COMPLETENESS_ENABLED = "iceberg.completeness.enabled";
+  private static final boolean DEFAULT_ICEBERG_COMPLETENESS = false;
+  private final boolean completenessEnabled;
```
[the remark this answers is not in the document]
vikrambohra: I don't understand this comment. Do you mean to move this to a separate class?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: Agree with the assumptions here but this logic is only triggered when completeness is enabled. Creating a new writer extending this writer will need a bigger refactor which should probably be part of another PR.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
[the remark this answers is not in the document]
ZihanLi58: I mean put all constant value definitions at the beginning of the class, and then start the normal field definition.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: I think it makes sense for it to be here. I will remove the one during table creation.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -265,6 +313,9 @@ public void write(GobblinMetadataChangeEvent gmce, Map<String, Collection<HiveSp
           return;
         }
         table = createTable(gmce, tableSpec);
+        if(tableMetadata.completenessEnabled) {
```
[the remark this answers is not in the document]
vikrambohra: Will remove this as the new column will be added during computation of late column

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
ZihanLi58: Yeah, just want to avoid confusing people when reading through the code. Can you create a follow up ticket to refactor the class?

gobblin-completeness/src/testFixtures/java/org/apache/gobblin/completeness/audit/TestAuditClient.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.gobblin.completeness.audit;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.gobblin.configuration.State;
+
+
+public class TestAuditClient implements AuditCountClient {
```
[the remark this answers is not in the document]
vikrambohra: This is a new and cleaner way of using test classes from one module in another introduced in gradle 5. These classes will not show up in the compile classpath

gobblin-completeness/src/testFixtures/java/org/apache/gobblin/completeness/audit/TestAuditClientFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gobblin.completeness.audit;
+
+import org.apache.gobblin.configuration.State;
+
+public class TestAuditClientFactory implements AuditCountClientFactory {
```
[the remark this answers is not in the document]
vikrambohra: same as above

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: No. We accept the type of new column via config "iceberg.completeness.add.partition.type"

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: Yes. intention is to keep going even if audit count system is down.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: Had a discussion with @omalley and it made more sense to be conservative and move in ascending order in case GMIP is lagging by a few hours

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: I think it makes sense for it to be here since its specific to this class. We should move more generic methods to util class.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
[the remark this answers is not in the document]
vikrambohra: Moved the config keys to a separate class

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
sv2000: Don't understand. Why should we be conservative? The goal should be to minimize calls to audit system.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: This cannot be done without using the current hour of the day.
So you are suggesting something like if(datepartition > previousWatermark && currentHour > datepartitionHour)
if prevWatermark = 9am
datepartition = 10am
currentHour = 10am (do not check audit system) vs 11 am (check audit system because GMIP may be lagging)

sv2000: My main idea is to optimize for the common case, which is that pipeline is caught up and data is current. If hour(currentTime) == prevWatermark: skip audit check, else (we are lagging) and hence, do the check.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: https://issues.apache.org/jira/browse/GOBBLIN-1540

sv2000 (changes requested):
One comment on implicit assumption of hourly granularity for completion watermarks.

sv2000 (approved):
+1. LGTM.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -422,6 +465,21 @@ private void computeCandidateSchema(GobblinMetadataChangeEvent gmce, TableIdenti
...
+   * Add a partition column to the schema and partition spec
+   * @param table incoming iceberg table
+   * @param fieldName name of partition column
+   * @param type datatype of partition column
+   * @return table with updated schema and partition spec
+   */
+  private Table addPartitionToIcebergTable(Table table, String fieldName, String type) {
+    if(!table.schema().columns().stream().anyMatch(x -> x.name().equalsIgnoreCase(fieldName))) {
+      table.updateSchema().addColumn(fieldName, Types.fromPrimitiveString(type)).commit();
+      table.updateSpec().addField(fieldName).commit();
```
ZihanLi58, line 478: Can you update this to include the change we talk offline before we merge it?

ZihanLi58 (approved):
+1, LGTM

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: done

PR #3387, opened 2021-09-06
[GOBBLIN-1537] provide a config to provide custom Trash class

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1537

### Description
- Provide a config for users to specify their custom trash class
- some minor refactoring to remove duplicate code

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/trash/Trash.java
+++ b/gobblin-data-management/src/main/java/org/apache/gobblin/data/management/trash/Trash.java
@@ -24,6 +24,9 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Properties;
 
+import org.apache.gobblin.configuration.State;
+import org.apache.gobblin.destination.DestinationDatasetHandlerService;
+import org.apache.gobblin.source.workunit.BasicWorkUnitStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -94,6 +97,8 @@ public class Trash implements GobblinTrash {
    */
   protected Path createTrashLocation(FileSystem fs, Properties props, String user) throws IOException {
     Path trashLocation;
+    new DestinationDatasetHandlerService(new State(props), false, null)
+        .executeHandlers(new BasicWorkUnitStream.Builder(Collections.EMPTY_LIST).build());
     if (props.containsKey(TRASH_LOCATION_KEY)) {
       trashLocation = new Path(props.getProperty(TRASH_LOCATION_KEY).replaceAll("\\$USER", user));
     } else {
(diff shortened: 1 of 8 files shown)


gobblin-data-management/src/main/java/org/apache/gobblin/data/management/trash/Trash.java
```diff
@@ -94,6 +97,8 @@ public Path getTrashLocation() throws IOException {
    */
   protected Path createTrashLocation(FileSystem fs, Properties props, String user) throws IOException {
     Path trashLocation;
+    new DestinationDatasetHandlerService(new State(props), false, null)
```
aplex, line 100: What will this do? Add something to props? Can we get the trash folder or dataset root folder directly, through the statically-typed APIs, instead of a global collection of props?

arjun4084346: Yes, is supposed to add configs to props. I thought about having a direct method to determine trash location, but we then will need to add a method in interface DestinationDatasetHandler. Should i do it that way?
Another way is to create a custom trash factory.

aplex: You can take a look on how this can be refactored better, to make the logic more obvious.

gobblin-data-management/src/main/java/org/apache/gobblin/data/management/trash/TrashFactory.java
```diff
@@ -59,19 +59,12 @@ public static Trash createTrash(FileSystem fs, Properties props) throws IOExcept
...
       throws IOException {
-    if(props.containsKey(TRASH_TEST) && Boolean.parseBoolean(props.getProperty(TRASH_TEST))) {
-      LOG.info("Creating a test trash. Nothing will actually be deleted.");
-      return new TestTrash(fs, props, user);
-    }
-    if(props.containsKey(SIMULATE) && Boolean.parseBoolean(props.getProperty(SIMULATE))) {
-      LOG.info("Creating a simulate trash. Nothing will actually be deleted.");
-      return new MockTrash(fs, props, user);
-    }
-    if(props.containsKey(SKIP_TRASH) && Boolean.parseBoolean(props.getProperty(SKIP_TRASH))) {
```
aplex, line 70: Looks like we no longer handle SKIP_TRASH and SIMULATE, and replace them with custom trash classes. Is there any effect on existing jobs/backward compatibility?

arjun4084346: Can you please reframe your question? Why do we no longer handle SKIP_TRASH and SIMULATE? I moved this piece of code to createTestMockOrImmediateDeletionTrash()

aplex: oh, I see. it was already checking it there.

aplex (approved):

PR #3388, opened 2021-09-07
[GOBBLIN-1536] Fix issue code generation

aplex:
Issue code should depend only on exception stack trace and type,
but previously there was a bug when exception message was included
in generation of the code. Since the message can have variable
parts, like paths and dataset names, some issue got multiple
codes assigned to them.

https://issues.apache.org/jira/browse/GOBBLIN-1536


autumnust (approved):
+1

PR #3393, opened 2021-09-10
[GOBBLIN-1542] Integrate with Helix API to add/remove task from a running helix job

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1542

### Description
Integrate with Helix API to add/remove task from a running helix job. This ticket also contains the API change that Source need to contact with JobLauncher to notify that there is some change in workunit. This will enable us to dynamically add/delete task from running job and save resources. 

### Tests

### Commits


gobblin-yarn/src/test/java/org/apache/gobblin/yarn/YarnServiceTestWithExpiration.java
```diff
@@ -204,7 +204,7 @@ public void testStartError() throws Exception{
     Assert.assertEquals(this.expiredYarnService.getNumRequestedContainers(), 10);
 
     try {
-      Thread.sleep(20000);
+      Thread.sleep(60000);
```
autumnust, line 207: Usually it is not a good practice to use thread.sleep in the unit test if not necessary. (and 60 seconds is really long if you considering the number of unit tests we have). 

why is this needed ? is there any other way to achieve the same? There's a test utilities called `AssertWithBackoff` you might want to look into, if satisfying certain condition is what you want after the sleep.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -259,6 +263,25 @@ public static void setDefaultAuthenticator(Properties properties) {
     }
   }
 
+  /**
+   * Handle {@link WorkUnitChangeEvent}, by default it will donothing
```
autumnust, line 267: donothing -> do nothing

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -259,6 +263,25 @@ public static void setDefaultAuthenticator(Properties properties) {
...
+    try {
+      this.removeTaskFromRunningHelixJob(workUnitChangeEvent.getOldTaskIds());
+      this.addTaskToRunningHelixJob(workUnitChangeEvent.getNewWorkUnits());
+    } catch (Exception e) {
+      //todo: emit some event to indicate there is an error handling this event that may cause starvation
+      throw new InvocationTargetException(e);
+    }
+  }
+
+  protected void removeTaskFromRunningHelixJob(List<String> taskIdsToRemove) throws IOException {}
```
autumnust, line 282: `removeTasksFromHelixJob`?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -259,6 +263,25 @@ public static void setDefaultAuthenticator(Properties properties) {
...
+  protected void addTaskToRunningHelixJob(List<WorkUnit> workUnitsToAdd) throws IOException {}
```
autumnust, line 283: shall it be tasks or there will only be one task ?

gobblin-restli/gobblin-flow-config-service/gobblin-flow-config-service-api/src/main/idl/org.apache.gobblin.service.flowexecutions.restspec.json
```diff
@@ -22,7 +22,6 @@
       "name" : "latestFlowExecution",
       "parameters" : [ {
         "name" : "flowId",
-        "doc" : "Retrieve the most recent matching FlowExecution(s) of the identified FlowId",
```
autumnust, line 25: is this change relevant?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaStreamingExtractor.java
```diff
@@ -302,7 +302,7 @@ public KafkaStreamingExtractor(WorkUnitState state) {
 
     for (int i = 0; i < numOfPartitions; ++i) {
       if (workUnit.getProp(topicNameProp, null) == null) {
-        log.warn("There's no topic.name property being set in workunt which could be an illegal state");
+        //log.warn("There's no topic.name property being set in workunt which could be an illegal state");
```
autumnust, line 305: if it is not appropriate let's remove this line

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -269,6 +271,62 @@ protected void executeCancellation() {
...
+              throw new IOException(
+                  String.format("Cannot remove task %s from helix job %s:%s after 3 times", workUnitId, this.helixWorkFlowName, jobName));
+            }
+          }
+        }
+      }
+    }
+  }
+
+  protected void addTaskToRunningHelixJob(List<WorkUnit> workUnitsToAdd) throws IOException {
```
autumnust, line 303: there's a lot of logic duplication between these two method. What they do seem to be as simple as: 
```
for workunit: workunits: 
    helixTask = getHelixTask (workunit) 
   retryer.execute(helixTask, helixOps)
```

Can you try to refactor them a bit ? There's retry object you could use, or create your own and pass-in the lambda expression

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -145,6 +146,31 @@ static void waitJobInitialization(
     log.info("Work flow {} initialized", workFlowName);
   }
 
+  public static boolean deleteTaskFromHelixJob(String workFlowName,
+      String jobName, String taskID, TaskDriver helixTaskDriver) {
+    try {
+      log.info(String.format("try to delete task %s from workflow %s, job %s", taskID, workFlowName, jobName));
+      helixTaskDriver.deleteTask(workFlowName, jobName, taskID);
+    } catch (Exception e) {
+      e.printStackTrace();
```
autumnust, line 155: why handle in this way ? why not propagated to caller and let them determine since this is just a util method ?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -145,6 +146,31 @@ static void waitJobInitialization(
...
+    }
+    return !helixTaskDriver.getJobConfig(TaskUtil.getNamespacedJobName(workFlowName, jobName)).getMapConfigs().containsKey(taskID);
```
autumnust, line 157: Do you know what kind of consistency guarantee offered by deleteTask and addTask from helix ? Or rather, what's the intention of read-after-write pattern here ?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
```diff
@@ -145,6 +146,31 @@ static void waitJobInitialization(
     log.info("Work flow {} initialized", workFlowName);
   }
 
+  public static boolean deleteTaskFromHelixJob(String workFlowName,
```
autumnust, line 149: package-private static is good enough?

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -427,6 +450,13 @@ public void apply(JobListener jobListener, JobContext jobContext)
         TimingEvent workUnitsCreationTimer =
             this.eventSubmitter.getTimingEvent(TimingEvent.LauncherTimings.WORK_UNITS_CREATION);
         Source<?, ?> source = this.jobContext.getSource();
+        if (source instanceof KafkaSource) {
```
autumnust, line 453: I think this is leaking the abstraction if we just use `instanceof` here ... AbstractJobLauncher is supposed to contain no implementation-specific constructs and I would argue even importing stuff from `import org.apache.gobblin.source.extractor.extract.kafka` package should not happen here. 

Would you reconsider this design, do you really need the abstractJobLauncher to be registered?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -375,6 +378,12 @@ private void createEmptyWorkUnitsForSkippedPartitions(Map<String, List<WorkUnit>
     }
   }
 
+  private void onWorkUnitUpdate (List<String> oldTaskIds, List<WorkUnit> newWorkUnits) {
```
autumnust, line 381: Where is this method called? should it really be Source's responsibility to detect it ?

gobblin-modules/gobblin-kafka-common/src/main/java/org/apache/gobblin/source/extractor/extract/kafka/KafkaSource.java
```diff
@@ -168,6 +169,8 @@
   private MetricContext metricContext;
 
   protected Optional<LineageInfo> lineageInfo;
+  @Setter
+  protected EventBus eventBus;
```
autumnust, line 173: it sounds awkward to set this object from external.

autumnust (changes requested):
Can you actually add more description in terms of how you want the workunit-udpates to be captured, propagated to some handler, and how it be handled ?  I feel those details were needed for discussion before we continue digging into code.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobLauncher.java
```diff
@@ -269,6 +271,62 @@ protected void executeCancellation() {
     }
   }
 
+  protected void removeTaskFromRunningHelixJob(List<String> workUnitIdsToRemove) throws IOException {
```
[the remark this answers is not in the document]
ZihanLi58: I run reformat to clean the whole class, this should be good now.

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/ClusterIntegrationTest.java
```diff
@@ -89,8 +89,8 @@ void testJobShouldGetCancelled() throws Exception {
         IntegrationJobCancelSuite.TASK_STATE_FILE)
         .withValue(SleepingTask.SLEEP_TIME_IN_SECONDS, ConfigValueFactory.fromAnyRef(100));
     this.suite = new IntegrationJobCancelSuite(jobConfigOverrides);
-    HelixManager helixManager = getHelixManager();
     suite.startCluster();
+    HelixManager helixManager = getHelixManager();
```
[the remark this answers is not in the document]
ZihanLi58: This is relevant, since I see this task fail locally after I bump helix version, and this is to fix the test

gobblin-cluster/src/test/java/org/apache/gobblin/cluster/HelixAssignedParticipantCheckTest.java
```diff
@@ -50,16 +50,16 @@ public void setUp()
...
-    String zkConnectString = helixConfig.getString(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY);
-    helixManager = HelixManagerFactory.getZKHelixManager(clusterName, "TestManager",
-        InstanceType.SPECTATOR, zkConnectString);
   }
 
   @Test (groups = {"disabledOnCI"})
   //Test disabled on Travis because cluster integration tests are generally flaky on Travis.
   public void testExecute() throws Exception {
     suite.startCluster();
+    String clusterName = helixConfig.getString(GobblinClusterConfigurationKeys.HELIX_CLUSTER_NAME_KEY);
```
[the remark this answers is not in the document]
ZihanLi58: Same as the comment above, it's after bump helix version, seems we have to call suite.startCluster first before we get a new helix manager instance

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
ZihanLi58: oh that's a good question, I haven't check that part, just follow the instruction on the doc that when error happens it's caller's responsibility to check whether the update successful. Let me only add this part when exception happens to not rely on the read/write consistency in normal case

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixUtils.java
ZihanLi58: Just want to wrap all the delete / add logic in the util method, and as the doc mentioned, if there is exception it's caller's responsibility to determine whether it succeed.

autumnust:
Is this ready for review ? I see there's still a test failure. @ZihanLi58

PR #3394, opened 2021-09-10
[GOBBLIN-1543] Set default for work unit size

umustafi:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1543

### Description
User encountered exception when summing work units for retention jobs that do not have size associated with it, so I am defaulting to progress reporting in terms of work units for cases that do not have size associated with it. 

### Tests
Tested manually with retention job with progress reporting turned on. 

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
@@ -663,7 +663,7 @@ public abstract class AbstractJobLauncher implements JobLauncher {
   @VisibleForTesting
   public static long sumWorkUnitsSizes (WorkUnitStream workUnitStream) {
     Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
-    long totalSizeInBytes = workUnits.stream().mapToLong(wu -> wu.getPropAsLong(ServiceConfigKeys.WORK_UNIT_SIZE, 1)).sum();
+    long totalSizeInBytes = workUnits.stream().mapToLong(wu -> wu.getPropAsLong(ServiceConfigKeys.WORK_UNIT_SIZE, 0)).sum();
     return totalSizeInBytes;
   }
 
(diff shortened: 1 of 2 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/AbstractJobLauncher.java
```diff
@@ -663,7 +663,7 @@ public void apply(JobListener jobListener, JobContext jobContext)
   @VisibleForTesting
   public static long sumWorkUnitsSizes (WorkUnitStream workUnitStream) {
     Collection<WorkUnit> workUnits = JobLauncherUtils.flattenWorkUnits(workUnitStream.getMaterializedWorkUnitCollection());
-    long totalSizeInBytes = workUnits.stream().mapToLong(wu -> wu.getPropAsLong(ServiceConfigKeys.WORK_UNIT_SIZE)).sum();
+    long totalSizeInBytes = workUnits.stream().mapToLong(wu -> wu.getPropAsLong(ServiceConfigKeys.WORK_UNIT_SIZE, 1)).sum();
```
aplex, line 666: If it does not have the size, I would assume that size is 0. Any reason to use 1 here?

umustafi: We had a case where a user started a retention job with progress reporting that failed since the Retention Source does not set a value for this property. However, I also noticed that if all the workunits have size 0 and progress reporting is turned on then there will be a divide by 0 error [here](https://jarvis.corp.linkedin.com/codesearch/result/?name=TaskStateCollectorService.java&path=gobblin-elr%2Fgobblin-runtime%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fgobblin%2Fruntime&reponame=linkedin%2Fgobblin-elr#281). I defaulted to using "1" for the size so progress reporting will default to being based on # work units. Although retention jobs will occur very quickly the progress will appear "100%" instead of "0%" or "Not calculated" if I handled the error within the TaskStateCollectorService.

aplex: I think we should fix this division by zero where it occurs. It's still possible for the job to copy 0 bytes. For example, when it copies empty files.

In the place where we calculate progress, can we fallback to counting the work units if total size is zero? So the variable says total size in bytes, and it will confuse future developers if we put work unit count there.

aplex (approved):

PR #3402, opened 2021-09-24
[GOBBLIN-1552] determine flow status based on the fact if dag manager is enabled

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1552

### Description
flow status is determined by looking at the flow level events. but flow level events are not emitted outside of dag manager. so currently flow status is not being determined correctly when dag manager is disabled.
this pr adds a field flowStatus to FlowStatus class, which is determined by JobStatus'es when DagManager is not enabled. in a way it used to be done prior to 59f3bee#diff-141f30f7aa658e9e18316e3e5ecea3b92865dcc88a4c0809b6d5c0ef1598e410

Another refactoring this PR does is to make getJobStatusesForFlowExecution in jobStatusRetriever return a list instead of an Iterator, this is done because flow status determination needs to iterate through this iterator and the iterator cannot be re-parsed so it become unusable afterwards.

### Tests
added a couple of unit tests and updated the existing ones

### Commits


phet (commented):
my comments on https://github.com/apache/gobblin/pull/3403#pullrequestreview-764664143https://github.com/apache/gobblin/pull/3403
(almost all) still apply here, but overall, I am definitely in favor of this change, replacing the pitfall-prone `Iterator` w/ `List`.

PR #3405, opened 2021-09-27
[GOBBLIN-1553] Catch exception when iceberg does not support get metrics for non-optional union type

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1553
    - this is one hotfix

### Description
There is one update in upstream that support get iceberg schema for non-optinal union type, but get metrics will throw exception. So we need to catch this kind of exception when get metrics as well

### Tests

### Commits

### DIFF
--- a/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
+++ b/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
@@ -198,7 +198,12 @@ public class GobblinMCEPublisher extends DataPublisher {
           //This means the table is not compatible with iceberg, so return a dummy metric
           return new Metrics(100000000L, null, null, null);
         }
-        return OrcMetrics.fromInputFile(HadoopInputFile.fromPath(path, conf), MetricsConfig.getDefault(), mapping);
+        try {
+          return OrcMetrics.fromInputFile(HadoopInputFile.fromPath(path, conf), MetricsConfig.getDefault(), mapping);
+        } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
```diff
@@ -198,7 +198,12 @@ public static Metrics getMetrics(State state, Path path, Configuration conf, Nam
           //This means the table is not compatible with iceberg, so return a dummy metric
           return new Metrics(100000000L, null, null, null);
         }
-        return OrcMetrics.fromInputFile(HadoopInputFile.fromPath(path, conf), MetricsConfig.getDefault(), mapping);
+        try {
+          return OrcMetrics.fromInputFile(HadoopInputFile.fromPath(path, conf), MetricsConfig.getDefault(), mapping);
+        } catch (Exception e) {
+          //This means the table is not compatible with iceberg, so return a dummy metric
+          return new Metrics(100000000L, null, null, null);
```
vikrambohra, line 205: Can we have a common static instance called DUMMY_METRIC = new Metrics(100000000L, null, null, null); and return in both places

vikrambohra (changes requested):

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/publisher/GobblinMCEPublisher.java
ZihanLi58: addressed your comment

autumnust (approved):
+1

PR #3407, opened 2021-10-04
[GOBBLIN-1556]Add shutdown logic in FsJobConfigurationManager

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1556

### Description
We don't have shutdown logic in FsJobConfigurationManager, so when AM stopped, it still tries to fetch job status.

### Tests
Test when job finish, this service can exit cleanly

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsJobConfigurationManager.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsJobConfigurationManager.java
@@ -88,6 +88,12 @@ public class FsJobConfigurationManager extends JobConfigurationManager {
     }, 0, this.refreshIntervalInSeconds, TimeUnit.SECONDS);
   }
 
+  @Override
+  protected void shutDown() throws Exception {
+    ExecutorsUtils.shutdownExecutorService(this.fetchJobSpecExecutor, Optional.of(log));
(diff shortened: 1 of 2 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/FsJobConfigurationManager.java
```diff
@@ -88,6 +88,12 @@ public void run() {
     }, 0, this.refreshIntervalInSeconds, TimeUnit.SECONDS);
   }
 
+  @Override
+  protected void shutDown() throws Exception {
```
sv2000, line 92: Interesting. What happens in the absence of shutdown() method?  It was a little confusing from the PR description - if AM is stopped and the JobConfigManager continues to check for job specs, does it mean that the AM is prevented from exiting? Is there any impact on the Gobblin-on-Yarn application because of this?

ZihanLi58: No, it does not affect the exiting. It will stop after the whole container is existed. Just in another issue where AM stuck after existing, I notice this thread also not been exited, so make this change.

PR #3409, opened 2021-10-06
[GOBBLIN-1558] Update BaseDataPublisher

jobar:
Dear Gobblin maintainers,

This is my second PR, I hope I'm doing it in a way that works for you folks :)

### JIRA
- [ ] My PR addresses the following [Gobblin JIRA](https://issues.apache.org/jira/browse/GOBBLIN/) issues and references them in the PR title.
    - https://issues.apache.org/jira/browse/GOBBLIN-1558

### Description
- [ ] Make BaseDataPublisher publish to new folders in a way that preserve detailed subfolder information in PUBLISHER_DIRS state value. In previous version, when publishing to non-existing parent directory (first run of job type for instance), the BaseDataPublisher wouldn't recurse into written folders but moves the top written folder at once, leading to the state's PUBLISHER_DIRS value containing the top-level written folder, not the contain the detailed published folder.

### Tests
- [ ] Not yet done, will come soon.

### Commits


gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -487,6 +480,15 @@ protected void addSingleTaskWriterOutputToExistingDir(Path writerOutputDir, Path
     }
   }
 
+  protected void addWriterOutputToNewDir(Path writerOutputDir, Path publisherOutputDir,
+      WorkUnitState workUnitState, int branchId, ParallelRunner parallelRunner)
+      throws IOException {
+    // Create the parent directory of the final output directory if it does not exist
+    WriterUtils.mkdirsWithRecursivePermissionWithRetry(this.publisherFileSystemByBranches.get(branchId),
+            publisherOutputDir.getParent(), this.permissions.get(branchId), retrierConfig);
```
Will-Lo, line 488: Is there a reason why we're omitting the set output dir groups? I believe it's needed for permissions if configured
```
if(this.publisherOutputDirOwnerGroupByBranches.get(branchId).isPresent()) {
          LOG.info(String.format("Setting path %s group to %s", publisherOutputDir.toString(), this.publisherOutputDirOwnerGroupByBranches.get(branchId).get()));
          HadoopUtils.setGroup(this.publisherFileSystemByBranches.get(branchId), publisherOutputDir, this.publisherOutputDirOwnerGroupByBranches.get(branchId).get());
 ```

Will-Lo (commented):
Left a comment @jobar , sorry I got to this pretty late 😓

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
jobar: wow thank you for the catch, I didn't notice and should have been more careful. I just added the missing lines to the new functions.

jobar:
Thank you for the review @Will-Lo  - I added a commit corrected the problem you pointed out in comment.

PR #3417, opened 2021-10-21
GOBBLIN-1566:fix resource leak due to Files.walk and replace guava's …

lujiefsi:
https://issues.apache.org/jira/browse/GOBBLIN-1566

### DIFF
--- a/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobRunner.java
+++ b/gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobRunner.java
@@ -61,7 +61,7 @@ public class AzkabanJobRunner {
     runner.run();
   }
 
-  public static String getTempDirectory() {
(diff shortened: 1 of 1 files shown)


gobblin-modules/gobblin-azkaban/src/main/java/org/apache/gobblin/azkaban/AzkabanJobRunner.java
```diff
@@ -62,8 +61,10 @@ static void doMain(Class<? extends AzkabanJobRunner> klazz, String[] args)
     runner.run();
   }
 
-  public static String getTempDirectory() {
-    File tmpDirectory = Files.createTempDir();
+  public static String getTempDirectory() throws IOException {
+    File tempDirBase = new File(System.getProperty("java.io.tmpdir"));
+    File tmpDirectory = Files.createTempDirectory(
+         tempDirBase.toPath(), System.currentTimeMillis() + "-").toFile();
```
ZihanLi58, line 67: Why do we still want add "-" there? maybe change it to "-Gobblin"?

PR #3421, opened 2021-10-29
[GOBBLIN-1569] Add RDBMS-backed `MysqlJobCatalog`, as alternative to file system storage

phet:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1569

### Description

Add RDBMS-backed `MysqlJobCatalog`, as alternative to file system storage

Present storage bindings for `JobCatalog` are in-memory or file system.  While the latter is durable, it requires an NFS-like capability for cross-instance sharing.

Support alternative multi-instance backing through MySQL, precluding the need for NFS.

### Tests

New unit tests

### Commits

### DIFF
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/MysqlJobCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/MysqlJobCatalog.java
@@ -39,7 +39,8 @@ import org.slf4j.LoggerFactory;
 
 
 /**
- * MySQL-backed Job Catalog for persisting (`JobSpec`) job configuration information.  Fully support (mutation) listeners and metrics.
+ * MySQL-backed Job Catalog for persisting (`JobSpec`) job configuration information, which fully supports (mutation)
+ * listeners and metrics.
  */
 public class MysqlJobCatalog extends JobCatalogBase implements MutableJobCatalog {
 
--- a/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/MysqlJobCatalog.java
+++ b/gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/MysqlJobCatalog.java
@@ -1,0 +1,214 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gobblin.runtime.job_catalog;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Iterators;
+import com.typesafe.config.Config;
+import java.io.IOException;
+import java.net.URI;
+import java.util.Iterator;
+import java.util.List;
+import org.apache.gobblin.metrics.GobblinMetrics;
+import org.apache.gobblin.metrics.MetricContext;
+import org.apache.gobblin.runtime.api.GobblinInstanceEnvironment;
+import org.apache.gobblin.runtime.api.JobCatalog;
+import org.apache.gobblin.runtime.api.JobSpec;
+import org.apache.gobblin.runtime.api.JobSpecNotFoundException;
+import org.apache.gobblin.runtime.api.MutableJobCatalog;
+import org.apache.gobblin.runtime.api.SpecNotFoundException;
+import org.apache.gobblin.runtime.spec_serde.GsonJobSpecSerDe;
+import org.apache.gobblin.runtime.spec_store.MysqlBaseSpecStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * MySQL-backed Job Catalog for persisting (`JobSpec`) job configuration information.  Fully support (mutation) listeners and metrics.
+ */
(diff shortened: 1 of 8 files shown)


gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/JobSpecDeserializer.java
```diff
@@ -0,0 +1,83 @@
...
+          return new URI(s);
+        } catch (URISyntaxException e) {
+          LOGGER.warn(String.format("error deserializing '%s' as a URI",
+              jsonObject.get(JobSpecSerializer.JOB_SPEC_TEMPLATE_URI_KEY).getAsString()), e);
+          throw new RuntimeException(e);
+        }
+      });
+    } catch (RuntimeException e) {
+      templateURI = Optional.empty();
+    }
```
jack-moseley, line 61: This map function seems unnecessarily complicated, couldn't we just do:

```
    try {
      templateURI = Optional.ofNullable(new URI(jsonObject.get(JobSpecSerializer.JOB_SPEC_TEMPLATE_URI_KEY).getAsString()));
    } catch (URISyntaxException e) {
       LOGGER.warn(String.format("error deserializing '%s' as a URI",
              jsonObject.get(JobSpecSerializer.JOB_SPEC_TEMPLATE_URI_KEY).getAsString()), e);
      templateURI = Optional.empty();
    }
```

phet: first off, I agree it's unsatisfying.  (I welcome any tips toward finessing java's checked exceptions!)

the transformation you suggest doesn't protect against passing `null` to the `URI` ctor.  that's why I was working (via `map`) *within* the `Optional`.  still, you did helped spot a NPE bug from the parens closed in the wrong place.  we need:
```
Optional.ofNullable(jsonObj.get(...)).map
```
not the potentially explosive:
```
Optional.ofNullable(jsonObj.get(...).getAsString()).map
```
in addition, I'll use `Optional.flatMap` to eliminate the second level of exception handling.  I'll push those corrections, and you please tell me if you see any other potential improvement.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/GsonJobSpecSerDe.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.gobblin.runtime.api.Spec;
+import org.apache.gobblin.runtime.api.SpecSerDe;
+import org.apache.gobblin.runtime.api.SpecSerDeException;
+
+
+/**
+ * {@link SpecSerDe} that serializes as Json using {@link Gson}. Note that currently only {@link JobSpec}s are supported.
+ */
+public class GsonJobSpecSerDe implements SpecSerDe {
+  private GsonSerDe<JobSpec> gsonSerDe;
```
arjun4084346, line 34: could be `final`

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/GsonJobSpecSerDe.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.gobblin.runtime.api.JobSpec;
+import org.apache.gobblin.runtime.api.Spec;
+import org.apache.gobblin.runtime.api.SpecSerDe;
+import org.apache.gobblin.runtime.api.SpecSerDeException;
+
+
+/**
+ * {@link SpecSerDe} that serializes as Json using {@link Gson}. Note that currently only {@link JobSpec}s are supported.
+ */
+public class GsonJobSpecSerDe implements SpecSerDe {
```
arjun4084346, line 33: There must be a way to reuse `GsonFlowSpecSerDe` , can we use generics in the interface?

arjun4084346 (changes requested):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/GsonJobSpecSerDe.java
phet: yes, true

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/GsonJobSpecSerDe.java
phet: good suggestion... I guess I took the prior impl at face value, but I was able to re-work w/ generics into an abstract base class doing the heavy lifting.

since java lacks a type class resolution capability, we still need to create a specific, named derivation to supply the serializer and the deserializer.  handily, that non-generic class name also imparts the ability to refer to the type within HOCON or even simple properties-based config.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/job_catalog/MysqlJobCatalog.java
```diff
@@ -0,0 +1,214 @@
...
+  @Override
+  public synchronized void remove(URI jobURI, boolean alwaysTriggerListeners) {
+    Preconditions.checkState(state() == State.RUNNING, String.format("%s is not running.", this.getClass().getName()));
+    Preconditions.checkNotNull(jobURI);
+    try {
+      long startTime = System.currentTimeMillis();
+      JobSpec jobSpec = (JobSpec) jobSpecStore.getSpec(jobURI);
+      jobSpecStore.deleteSpec(jobURI);
+      this.mutableMetrics.updateRemoveJobTime(startTime);
+      this.listeners.onDeleteJob(jobURI, jobSpec.getVersion());
```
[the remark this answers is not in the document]
phet: I lack the history on this one.  I'm also not following what additional spec store calls arise as a result.
(overall I'm not opposed to streamlining the functionality on offer...)
please either provide more background context, or, if that effort would be a separate follow-on, we could save the discussion for later.

arjun4084346 (approved):

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/spec_serde/JobSpecDeserializer.java
jack-moseley: Okay I see, yeah I think the current way looks fine, it is good to not have the two levels of exception handling.

jack-moseley (approved):

ZihanLi58 (approved):
+1

PR #3426, opened 2021-11-16
[GOBBLIN-1574] Added whitelist for iceberg tables to add new partitio…

vikrambohra:
…n column

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1574

### Description
- Added whitelist for iceberg tables to which to add new partition column to schema and pspec
- completenessCheck config controls only if audit system need to be checked
- streamlined code to store db and table name to TableMetadata.

### Tests

### Commits

### DIFF
--- a/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
+++ b/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
@@ -1032,7 +1032,7 @@ public class IcebergMetadataWriter implements MetadataWriter {
           }
           tableMetadataMap.get(tid).setDatasetName(gmce.getDatasetIdentifier().getNativeName());
           tableMetadataMap.get(tid).setDbName(dbName);
-          tableMetadataMap.get(tid).setTableName(dbName);
+          tableMetadataMap.get(tid).setTableName(tableName);
           if(this.completenessEnabled && this.completenessWhitelistBlacklist.acceptTable(dbName, tableName)) {
             tableMetadataMap.get(tid).setCompletenessEnabled(true);
           }
--- a/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
+++ b/gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
@@ -165,6 +165,7 @@ public class IcebergMetadataWriter implements MetadataWriter {
   private final DateTimeFormatter HOURLY_DATEPARTITION_FORMAT;
   private final String newPartitionColumn;
   private final String newPartitionColumnType;
+  private final boolean newPartitionEnabled;
   private final WhitelistBlacklist newPartitionTableWhitelistBlacklist;
   private Optional<KafkaAuditCountVerifier> auditCountVerifier;
   private final String auditCheckGranularity;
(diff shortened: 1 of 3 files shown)


gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -1078,6 +1086,10 @@ public void close() throws IOException {
     long newCompletenessWatermark = DEFAULT_COMPLETION_WATERMARK;
     SortedSet<ZonedDateTime> datePartitions = new TreeSet<>(Collections.reverseOrder());
 
+    @Setter
+    String dbName;
```
ZihanLi58, line 1090: Why this two arg needed here?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -284,9 +287,8 @@ private Long getAndPersistCurrentWatermark(TableIdentifier tid, String topicPart
    * information increases the memory footprints, therefore we would like to flush them eagerly).
    */
   public void write(GobblinMetadataChangeEvent gmce, Map<String, Collection<HiveSpec>> newSpecsMap,
-      Map<String, Collection<HiveSpec>> oldSpecsMap, HiveSpec tableSpec) throws IOException {
-    TableIdentifier tid = TableIdentifier.of(tableSpec.getTable().getDbName(), tableSpec.getTable().getTableName());
-    TableMetadata tableMetadata = tableMetadataMap.computeIfAbsent(tid, t -> new TableMetadata());
+      Map<String, Collection<HiveSpec>> oldSpecsMap, HiveSpec tableSpec, TableMetadata tableMetadata) throws IOException {
```
ZihanLi58, line 290: We can directly parse tid instead of tableMetadata here?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -1025,11 +1031,13 @@ public void writeEnvelope(RecordEnvelope<GenericRecord> recordEnvelope, Map<Stri
                 Optional.of(currentOffset - 1);
           }
           tableMetadataMap.get(tid).setDatasetName(gmce.getDatasetIdentifier().getNativeName());
```
ZihanLi58, line 1033: We want to set those only when low water mark is not set (which means this is the first time we compute this table since last failure/ restart), so we can move those lines inside the block of line 1028?

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
```diff
@@ -684,7 +687,8 @@ protected void addFiles(GobblinMetadataChangeEvent gmce, Map<String, Collection<
         Collection<HiveSpec> hiveSpecs = newSpecsMap.get(new Path(file.getFilePath()).getParent().toString());
         StructLike partition = getIcebergPartitionVal(hiveSpecs, file.getFilePath(), partitionSpec);
 
-        if(tableMetadata.completenessEnabled && gmce.getOperationType() == OperationType.add_files) {
+        if(this.newPartitionTableWhitelistBlacklist.acceptTable(tableMetadata.dbName, tableMetadata.tableName)
+            && gmce.getOperationType() == OperationType.add_files) {
           tableMetadata.prevCompletenessWatermark = Long.parseLong(table.properties().getOrDefault(COMPLETION_WATERMARK_KEY,
```
ZihanLi58, line 692: What will happen if it's not add_files operation? i.e. what will happen for compaction?

ZihanLi58 (changes requested):

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: tid is an iceberg concept with namespace and name rather than db and table. Namespace can be dbName or a multilevel name like hivedb.dbname which could lead to errors. Other option is to pass tableSpec (which has db and table name) to addFiles method. Parsing db and table name everytime we need is not ideal.

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: same comment as below

gobblin-iceberg/src/main/java/org/apache/gobblin/iceberg/writer/IcebergMetadataWriter.java
vikrambohra: we will only white list kafka_streaming_hourly tables and re_write files will never occur for them.

ZihanLi58: Do we want to make this more general? we can just add one place holder here and address in another PR

PR #3427, opened 2021-11-16
[GOBBLIN-1575] use reference count in helix manager, so that connect/disconnect are called once and at the right time

arjun4084346:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1575

### Description
GobblinTaskRunner creates a HelixManager and passes it to HelixJobTask. When the GobblinTaskRunner is stopped, it disconnects to HelixManager, but because the HelixJobTask is also using the same HelixManager object, it cannot use it do its own shutdown operation. This PR will make a copy of HelixManager to be used by HelixJobTask 's.

Same argument with GobblinClusterManager passing the same HelixManager object to HelixRetriggeringJobCallable

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobTask.java
@@ -37,6 +37,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Closer;
 import com.typesafe.config.Config;
 
+import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.gobblin.configuration.ConfigurationKeys;
@@ -143,13 +144,14 @@ class GobblinHelixJobTask implements Task {
   /**
    * Launch the actual {@link GobblinHelixJobLauncher}.
    */
+  @SneakyThrows
   @Override
   public TaskResult run() {
     log.info("Running planning job {} [{} {}]", this.planningJobId, this.applicationName, this.instanceName);
     this.jobTaskMetrics.updateTimeBetweenJobSubmissionAndExecution(this.jobPlusSysConfig);
+    this.jobHelixManager.connect();
 
     try (Closer closer = Closer.create()) {
-      this.jobHelixManager.connect();
       Optional<String> planningIdFromStateStore = this.jobsMapping.getPlanningJobId(jobUri);
 
       long timeOut = PropertiesUtils.getPropAsLong(jobPlusSysConfig,
--- a/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobTask.java
+++ b/gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobTask.java
@@ -84,8 +84,13 @@ class GobblinHelixJobTask implements Task {
     this.helixMetrics = helixMetrics;
     this.taskConfig = context.getTaskConfig();
     this.sysConfig = builder.getConfig();
-    this.jobHelixManager = builder.getJobHelixManager();
     this.jobPlusSysConfig = ConfigUtils.configToProperties(sysConfig);
+
+    // make a copy of helix manager, so that each job can keep using it without worrying about disconnect from the GobblinTaskRunner where they are created
+    this.jobHelixManager =  GobblinHelixMultiManager.buildHelixManager(ConfigUtils.propertiesToConfig(jobPlusSysConfig),
+        jobPlusSysConfig.getProperty(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY),
+        GobblinClusterConfigurationKeys.HELIX_CLUSTER_NAME_KEY, builder.getJobHelixManager().getInstanceType());
+
     this.jobLauncherListener = new GobblinHelixJobLauncherListener(launcherMetrics);
 
     Map<String, String> configMap = this.taskConfig.getConfigMap();
@@ -149,6 +154,7 @@ class GobblinHelixJobTask implements Task {
     this.jobTaskMetrics.updateTimeBetweenJobSubmissionAndExecution(this.jobPlusSysConfig);
 
     try (Closer closer = Closer.create()) {
+      this.jobHelixManager.connect();
       Optional<String> planningIdFromStateStore = this.jobsMapping.getPlanningJobId(jobUri);
 
       long timeOut = PropertiesUtils.getPropAsLong(jobPlusSysConfig,
@@ -203,6 +209,7 @@ class GobblinHelixJobTask implements Task {
       return new TaskResult(TaskResult.Status.FAILED, "Exception occurred for job " + planningJobId + ":" + ExceptionUtils
           .getFullStackTrace(e));
     } finally {
+      this.jobHelixManager.disconnect();
       // always cleanup the job mapping for current job name.
       try {
         this.jobsMapping.deleteMapping(jobUri);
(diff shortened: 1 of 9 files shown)


gobblin-cluster/src/main/java/org/apache/gobblin/cluster/GobblinHelixJobTask.java
```diff
@@ -203,6 +209,7 @@ public TaskResult run() {
       return new TaskResult(TaskResult.Status.FAILED, "Exception occurred for job " + planningJobId + ":" + ExceptionUtils
           .getFullStackTrace(e));
     } finally {
+      this.jobHelixManager.disconnect();
```
phet, line 212: formatting is a bit hard to discern (in the PR viewer), but is this the `finally` of the `try` block where the `.connect()` call is the first statement?  if so, shouldn't we either move the `.connect()` directly prior to the `try`, or else begin another `try` right after the `connect()`?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -161,6 +170,10 @@ public Void call() throws JobException {
     }
 
     try {
+      this.jobHelixManager.connect();
+      if (this.taskDriverHelixManager.isPresent()) {
+        this.taskDriverHelixManager.get().connect();
+      }
```
phet, line 176: again... don't these belong prior to and outside the `try` (technically even w/ each its own separate, nested `try` block)?

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -172,7 +185,10 @@ public Void call() throws JobException {
       if (deleteJobWhenException) {
         deleteJobSpec();
       }
-      throw e;
+      throw new JobException(e.getMessage());
+    } finally {
+      this.jobHelixManager.disconnect();
+      this.taskDriverHelixManager.ifPresent(HelixManager::disconnect);
```
phet, line 191: I believe these belong in separate `try ... finally` blocks (and, per nesting, `disconnect()` (i.e. closing/finalization) should be in *reverse* order of `connect()` (initialization)

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -126,8 +126,17 @@ public HelixRetriggeringJobCallable(
...
+    // make a copy of helix managers, so that each job can keep using it without worrying about disconnect from the GobblinClusterManager where they are created
+    this.jobHelixManager = GobblinHelixMultiManager.buildHelixManager(ConfigUtils.propertiesToConfig(sysProps),
+        sysProps.getProperty(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY),
+        GobblinClusterConfigurationKeys.HELIX_CLUSTER_NAME_KEY, jobHelixManager.getInstanceType());
+
+    this.taskDriverHelixManager = taskDriverHelixManager.map(
+        helixManager -> GobblinHelixMultiManager.buildHelixManager(ConfigUtils.propertiesToConfig(sysProps),
+            sysProps.getProperty(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY),
+            GobblinClusterConfigurationKeys.TASK_DRIVER_CLUSTER_NAME_KEY, helixManager.getInstanceType()));
+
```
phet, line 139: when I saw this above, I was wondering whether it shouldn't be abstracted into a utility... now seeing it again (twice!), I really think so.  any impediment to doing so?

phet (commented):
couple smaller things to clarify about resource management...

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
arjun4084346: The thing I considered is that I was trying to copy a HelixManager, not a ZKHelixManager. That was all the reasoing behind using the utility method.

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -249,8 +260,9 @@ private void runJobExecutionLauncher() throws JobException {
     String newPlanningId;
     Closer closer = Closer.create();
     try {
-      HelixManager planningJobManager = this.taskDriverHelixManager.isPresent()?
+      HelixManager planningJobHelixManager = this.taskDriverHelixManager.isPresent()?
           this.taskDriverHelixManager.get() : this.jobHelixManager;
```
phet, line 264: `getOrElse`?

...actually thought I earlier saw a similar replacement above...

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
```diff
@@ -263,7 +275,7 @@ private void runJobExecutionLauncher() throws JobException {
       jobLock.lock();
 
       try {
-        if (planningJobIdFromStore.isPresent() && !canRun(planningJobIdFromStore.get(), planningJobManager)) {
+        if (planningJobIdFromStore.isPresent() && !canRun(planningJobIdFromStore.get(), planningJobHelixManager)) {
```
phet, line 278: not pressing, but equivalent:
```
if (planningJobIdFromStore.map(id -> !canRun(id, planningJobHelixManager)).getOrElse(false)) {
```

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
phet: yeah, I suppose `GobblinHelixMultiManager.buildHelixManager` is already a utility, probably excessive to abstract too much beyond.  what I noticed is that (before calling) you're pulling a property out of the same `Config` you provide to the call.  usage shows `ZK_CONNECTION_STRING_KEY` to be pretty widely referenced... what about a 3-param overload:
```
protected static HelixManager buildHelixManager(Config config, String clusterName, InstanceType type) {
  return buildHelixManager(config, config.getProperty(GobblinClusterConfigurationKeys.ZK_CONNECTION_STRING_KEY), clusterName, type);
}
```
?
  

is

arjun4084346: I agree, the existing code did not look good to me as well. will change

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
arjun4084346: Yes, i removed it in this PR. 1st change of the PR

gobblin-cluster/src/main/java/org/apache/gobblin/cluster/HelixRetriggeringJobCallable.java
arjun4084346: Actually I tried that, here and at other place. But because the map Function does not expect to throw an exception, it does not end up as clean as this.

phet: aah, darn, yes, frustrating...

phet (approved):
looks good!

phet (commented):
very close... just some fine-tuning

phet (approved):
looks great, arjun!

ZihanLi58 (approved):
+1

PR #3430, opened 2021-11-17
[GOBBLIN-1577] change the multiplier used in ExponentialWaitStrategy to a reasonable…

arjun4084346:
… number 2 seconds

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1577

### Description
The retryer multiplier for exponential wait strategy was 2ms which makes some of our jobs retries too fast. It works essentially same as instant retry when we do not want it. changing it to 1s

Also changed a part in JobStatusMonitor where it can accept retryer configs via config object that was passed to it.

also did some codestyle fixes which I came across while navigating the code

### Tests
trivial changes

### Commits

### DIFF
--- a/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
+++ b/gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
@@ -17,7 +17,6 @@
 
 package org.apache.gobblin.compaction.mapreduce;
 
-import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
@@ -55,6 +54,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
@@ -186,7 +186,7 @@ public abstract class MRCompactorJobRunner implements Runnable, Comparable<MRCom
         ImmutableMap.<String, Object>builder()
             .put(RETRY_TIME_OUT_MS, TimeUnit.MINUTES.toMillis(2L))   //Overall retry for 2 minutes
             .put(RETRY_INTERVAL_MS, TimeUnit.SECONDS.toMillis(5L)) //Try to retry 5 seconds
-            .put(RETRY_MULTIPLIER, 2L) // Muliply by 2 every attempt
+            .put(RETRY_MULTIPLIER, 2L) // Multiply by 2 every attempt
             .put(RETRY_TYPE, RetryType.EXPONENTIAL.name())
             .build();
     COMPACTION_RETRY_DEFAULTS = ConfigFactory.parseMap(configMap);
--- a/gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
@@ -17,16 +17,6 @@
 
 package org.apache.gobblin.publisher;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
-import com.google.common.io.Closer;
-import com.typesafe.config.Config;
-import com.typesafe.config.ConfigFactory;
-import com.typesafe.config.ConfigRenderOptions;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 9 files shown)


gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
```diff
@@ -186,7 +186,7 @@
         ImmutableMap.<String, Object>builder()
             .put(RETRY_TIME_OUT_MS, TimeUnit.MINUTES.toMillis(2L))   //Overall retry for 2 minutes
             .put(RETRY_INTERVAL_MS, TimeUnit.SECONDS.toMillis(5L)) //Try to retry 5 seconds
-            .put(RETRY_MULTIPLIER, 2L) // Muliply by 2 every attempt
+            .put(RETRY_MULTIPLIER, TimeUnit.SECONDS.toMillis(2L))
```
phet, line 189: `RETRY_INTERVAL_MS` appears to be a ceiling on retry pause duration, so a multiplier of 2s, gives the sequence: [2s, 4s, 5s, 5s, 5s, ...] -
from com.github.rholder.retry.WaitStrategies.ExponentialWaitStrategy:
```
        public long computeSleepTime(Attempt failedAttempt) {
            double exp = Math.pow(2, failedAttempt.getAttemptNumber());
            long result = Math.round(multiplier * exp);
            if (result > maximumWait) {
                result = maximumWait;
            }
```
is that what we want or does it merit increase too?

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
```diff
@@ -122,14 +125,14 @@
   static final String PUBLISH_RETRY_ENABLED = DATA_PUBLISHER_RETRY_PREFIX + "enabled";
 
   static final Config PUBLISH_RETRY_DEFAULTS;
-  protected final Config retrierConfig;
+  protected final Config retryerConfig;
```
phet, line 128: wow, really hard to know how to spell this!  I might prefer slightly the form you just chose, although I do find this in the AWS SDK - https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/stepfunctions/builder/states/Retrier.html

I guess we just want it consistent in the codebase

Note though that the guava lib has a `RetryerBuilder` and the file at the end of this PR is `RetryerFactory`, so seemingly where the momentum's going

gobblin-compaction/src/main/java/org/apache/gobblin/compaction/mapreduce/MRCompactorJobRunner.java
arjun4084346: god catch. increased maximumWait

gobblin-core/src/main/java/org/apache/gobblin/publisher/BaseDataPublisher.java
arjun4084346: haha. I think we need to follow https://github.com/rholder/guava-retrying/blob/master/src/main/java/com/github/rholder/retry/Retryer.java name

phet: fo' sho'!

gobblin-utility/src/main/java/org/apache/gobblin/util/retry/RetryerFactory.java
```diff
@@ -127,7 +124,7 @@ public boolean apply(Throwable t) {
     return RetryerBuilder.<T> newBuilder()
         .retryIfException(RETRY_EXCEPTION_PREDICATE)
         .withWaitStrategy(WaitStrategies.exponentialWait(config.getLong(RETRY_MULTIPLIER),
-                                                         config.getLong(RETRY_INTERVAL_MS),
+                                                         config.getLong(RETRY_MAX_INTERVAL_MS),
```
phet, line 127: doesn't this break backward compatibility?

first off, I prefer your new name to the existing one... but existing code out there is setting `RETRY_MULTIPLIER` and that no longer means anything for EB--which is the default strategy.  I'd say it's too late to change naming.

less severely, but since this is a general library, is it possible certain existing clients aren't on such a large time-scale and would instead consider seconds (rather than ms) way too long?

...or do you think I'm being too conservative?

gobblin-modules/gobblin-kafka-09/src/test/java/org/apache/gobblin/runtime/KafkaAvroJobStatusMonitorTest.java
```diff
@@ -356,7 +357,8 @@ public void testProcessingRetriedForApparentlyTransientErrors() throws IOExcepti
     }, 2, 2, TimeUnit.SECONDS);
     Thread mainThread = Thread.currentThread();
     // guardrail against excessive retries (befitting this unit test):
-    toggleManagementExecutor.scheduleAtFixedRate(mainThread::interrupt, 60, 5, TimeUnit.SECONDS);
+    toggleManagementExecutor.scheduleAtFixedRate(mainThread::interrupt, 20, 5, TimeUnit.SECONDS);
```
phet, line 360: so is the present EB series - [2s, 4s, 8s, 16s, 20s, 20s, 20s, ... ]?  if so only 20s may not be long enough to allow for 5 "fake exception" failures.  or am I getting my numbers wrong?

phet: actually, now I see the 1ms above, so sequence of [1ms, 2ms, 4ms, 8ms, 16ms, 32ms, ... ]

all good!

phet (approved):
looks good, arjun!

gobblin-service/src/main/java/org/apache/gobblin/service/monitoring/KafkaJobStatusMonitor.java
```diff
@@ -131,10 +131,8 @@ public KafkaJobStatusMonitor(String topic, Config config, int numThreads, JobIss
 
     this.flowNameGroupToWorkUnitCount = new ConcurrentHashMap<>();
 
-    Config retryerOverridesConfig = config.hasPath(ConfigurationKeys.KAFKA_JOB_STATUS_MONITOR_RETRY_TIME_OUT_MINUTES)
-        ? ConfigFactory.parseMap(ImmutableMap.of(
-                RETRY_TIME_OUT_MS,
-                config.getDuration(ConfigurationKeys.KAFKA_JOB_STATUS_MONITOR_RETRY_TIME_OUT_MINUTES, TimeUnit.MINUTES)))
```
ZihanLi58, line 137: Do we want to remove this definition from ConfigurationKeys as well?

ZihanLi58 (changes requested):
one minor comment

PR #3436, opened 2021-11-23
[GOBBLIN-1582] Fill low/high watermark info in SourceState for QueryBasedSource

ZihanLi58:
Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1582

### Description
Fill low/high watermark info in SourceState for QueryBasedSource
Foe stress test class, add function to make the job always fail

### Tests
unit tests

### Commits

### DIFF
--- a/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
@@ -234,8 +234,8 @@ public abstract class QueryBasedSource<S, D> extends AbstractSource<S, D> {
       extract.setFullTrue(System.currentTimeMillis());
     }
 
-    long highestWaterMark = -1;
-    long lowestWaterMark = Long.MAX_VALUE;
+    Optional<Long> highestWaterMark = Optional.absent();
+    Optional<Long> lowestWaterMark = Optional.absent();
     for (Partition partition : partitions) {
       WorkUnit workunit = WorkUnit.create(extract);
       workunit.setProp(ConfigurationKeys.SOURCE_ENTITY, sourceEntity.getSourceEntityName());
@@ -244,13 +244,15 @@ public abstract class QueryBasedSource<S, D> extends AbstractSource<S, D> {
       addLineageSourceInfo(state, sourceEntity, workunit);
       partition.serialize(workunit);
       workUnits.add(workunit);
-      highestWaterMark = Math.max(highestWaterMark, partition.getHighWatermark());
-      lowestWaterMark = Math.min(lowestWaterMark, partition.getLowWatermark());
+      highestWaterMark = highestWaterMark.isPresent() ?
+          Optional.of(Math.max(highestWaterMark.get(), partition.getHighWatermark())) : Optional.of(partition.getHighWatermark());
+      lowestWaterMark = lowestWaterMark.isPresent() ?
+          Optional.of(Math.min(lowestWaterMark.get(), partition.getLowWatermark())) : Optional.of(partition.getLowWatermark());
     }
-    state.appendToListProp(TimingEvent.FlowEventConstants.HIGH_WATERMARK_FIELD,
-        String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, highestWaterMark));
-    state.appendToListProp(TimingEvent.FlowEventConstants.LOW_WATERMARK_FIELD,
-        String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, lowestWaterMark));
+    if(highestWaterMark.isPresent() && lowestWaterMark.isPresent()) {
+      state.appendToListProp(TimingEvent.FlowEventConstants.HIGH_WATERMARK_FIELD, String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, highestWaterMark));
+      state.appendToListProp(TimingEvent.FlowEventConstants.LOW_WATERMARK_FIELD, String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, lowestWaterMark));
+    }
 
     return workUnits;
   }
--- a/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
+++ b/gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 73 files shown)


gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -241,7 +244,13 @@ public int hashCode() {
       addLineageSourceInfo(state, sourceEntity, workunit);
       partition.serialize(workunit);
       workUnits.add(workunit);
+      highestWaterMark = Math.max(highestWaterMark, partition.getHighWatermark());
+      lowestWaterMark = Math.min(lowestWaterMark, partition.getLowWatermark());
     }
+    state.appendToListProp(TimingEvent.FlowEventConstants.HIGH_WATERMARK_FIELD,
+        String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, highestWaterMark));
+    state.appendToListProp(TimingEvent.FlowEventConstants.LOW_WATERMARK_FIELD,
+        String.format("%s.%s: %s", sourceEntity.getDatasetName(), sourceEntity.destTableName, lowestWaterMark));
```
phet, line 253: could `partitions` ever be empty?  if so, I recommend lowest, highest to be `Optional` and only write conditionally, rather than setting the properties to `MAX_VALUE` and `-1`.

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/JobExecutionEventSubmitter.java
```diff
@@ -77,6 +78,8 @@ private void submitJobStateEvent(JobState jobState) {
     jobMetadataBuilder.put(METADATA_JOB_COMPLETED_TASKS, Integer.toString(jobState.getCompletedTasks()));
     jobMetadataBuilder.put(METADATA_JOB_LAUNCHER_TYPE, jobState.getLauncherType().toString());
     jobMetadataBuilder.put(METADATA_JOB_TRACKING_URL, jobState.getTrackingURL().or(UNKNOWN_VALUE));
+    jobMetadataBuilder.put(TimingEvent.FlowEventConstants.HIGH_WATERMARK_FIELD, jobState.getProp(TimingEvent.FlowEventConstants.HIGH_WATERMARK_FIELD, ""));
+    jobMetadataBuilder.put(TimingEvent.FlowEventConstants.LOW_WATERMARK_FIELD, jobState.getProp(TimingEvent.FlowEventConstants.LOW_WATERMARK_FIELD, ""));
```
phet, line 82: may not be an issue... just curious: when do we use `""` and when instead `UNKNOWN_VALUE`?

gobblin-utility/src/test/java/org/apache/gobblin/util/test/TestStressTestingSource.java
```diff
@@ -168,4 +168,37 @@ public void testRunDuration() throws DataRecordException, IOException {
...
+    Assert.assertEquals(wus.size(), NUM_WORK_UNITS);
+
+    WorkUnit wu = wus.get(0);
+    WorkUnitState wuState = new WorkUnitState(wu, state);
+    Extractor<String, byte[]> extractor = source.getExtractor(wuState);
+
+    Assert.expectThrows(IOException.class, () -> {
+      byte[] record;
+      while ((record = extractor.readRecord(null)) != null) {
+        Assert.assertEquals(record.length, 100);
```
phet, line 200: minor, but for clarity I might replace with: `Assert.fail("should have thrown!")`

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/JobExecutionEventSubmitter.java
ZihanLi58: For me I think UNKONW_VALUE is not expected in most cases and indicate there is something wrong, but "" just indicate we don't set it.

gobblin-utility/src/test/java/org/apache/gobblin/util/test/TestStressTestingSource.java
ZihanLi58: In this case I want to test whether the job can fail as expected. But assert.fail will not test anything and directly fail the test which is not what I want. Please let me know if I miss something here.

gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -241,6 +244,14 @@ public int hashCode() {
       addLineageSourceInfo(state, sourceEntity, workunit);
       partition.serialize(workunit);
       workUnits.add(workunit);
+      highestWaterMark = highestWaterMark.isPresent() ?
+          Optional.of(Math.max(highestWaterMark.get(), partition.getHighWatermark())) : Optional.of(partition.getHighWatermark());
```
phet, line 248: `Optional.map` would be more idiomatic

gobblin-runtime/src/main/java/org/apache/gobblin/runtime/JobExecutionEventSubmitter.java
phet: sounds reasonable

gobblin-utility/src/test/java/org/apache/gobblin/util/test/TestStressTestingSource.java
phet: I figured with `THROW_EXCEPTION == true` it would throw in the first `extractor.readRecord(null)` call.  therefore the `Assert.fail` immediately alerts us to begin debugging if that doesn't happen.  it's also self-documenting, so future maintainers know, "that loop body should never run".

phet: actually... looks like it throws just prior to the (final) `extractor.readRecord(null)` call that would first return `null`.  so in that case, yes, leave as is and ignore my `Assert.fail` recommendation.

gobblin-core/src/main/java/org/apache/gobblin/source/extractor/extract/QueryBasedSource.java
```diff
@@ -241,6 +244,14 @@ public int hashCode() {
       addLineageSourceInfo(state, sourceEntity, workunit);
       partition.serialize(workunit);
       workUnits.add(workunit);
+      highestWaterMark = highestWaterMark.isPresent() ?
+          highestWaterMark.transform(l -> Math.max(l, partition.getHighWatermark())) : Optional.of(partition.getHighWatermark());
```
arjun4084346, line 248: *nit maybe `l` should be renamed to hw ?
and in next line `l` should be `lw ?

arjun4084346 (approved):

phet (approved):
nice work, zihan!

PR #3438, opened 2021-12-03
[GOBBLIN-1584] Add replace record logic for Mysql writer

umustafi:
* Allows user to configure a mysql ingestion job to allow replacement of the record's value for an existing record in the table.
* Also replaces a backward iteration of ResultSet() with forward iteration in MySqlWriterCommands

Dear Gobblin maintainers,

### JIRA
    - https://issues.apache.org/jira/browse/GOBBLIN-1584 

### Description
We only support insertion of new values into tables, not update/upsert. If you run an ingestion job with a record containing a record with the same primary key the whole job will fail because of the duplicate entry. We don't expect ingestion jobs to always contain only new records, so we should handle duplicate entries, ingestion of data already in the table or updates to old values. The insert vs. replace logic should be configurable to the user as well.

### Tests
It adds a unit test that utilizes replace logic in the JdbcPublisherTest.
I have also tested manually by copying duplicate data and repeated primary keys into a Mysql table and verified the values were updated.

### Commits

### DIFF
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -219,7 +219,7 @@ public class ConfigurationKeys {
   /**
    * Optional property to specify whether existing data in databases can be overwritten during ingestion jobs
    */
-  public static final String ALLOW_DATA_OVERWRITE = "allow.data.overwrite";
+  public static final String ALLOW_JDBC_RECORD_OVERWRITE = "allow.jdbc.record.overwrite";
 
   /**
    * Optional property to specify a default Authenticator class for a job
--- a/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
+++ b/gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
@@ -217,6 +217,11 @@ public class ConfigurationKeys {
   public static final long DEFAULT_QUEUED_TASK_TIME_MAX_AGE = TimeUnit.HOURS.toMillis(1);
 
   /**
+   * Optional property to specify whether existing data in databases can be overwritten during ingestion jobs
+   */
+  public static final String ALLOW_DATA_OVERWRITE = "allow.data.overwrite";
+
+  /**
    * Optional property to specify a default Authenticator class for a job
    */
   public static final String DEFAULT_AUTHENTICATOR_CLASS = "job.default.authenticator.class";
--- a/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
+++ b/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
@@ -183,7 +183,6 @@ public abstract class BaseJdbcBufferedInserter implements JdbcBufferedInserter {
     return String.format(INSERT_STATEMENT_PREFIX_FORMAT, databaseName, table, JOINER_ON_COMMA.join(this.columnNames));
   }
 
-
   @Override
   public void flush() throws SQLException {
     if (this.pendingInserts == null || this.pendingInserts.isEmpty()) {
--- a/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
+++ b/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
@@ -34,10 +34,12 @@ import com.github.rholder.retry.WaitStrategies;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 
+import lombok.ToString;
+
+import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.converter.jdbc.JdbcEntryData;
 import org.apache.gobblin.converter.jdbc.JdbcEntryDatum;
-import lombok.ToString;
 
 
 /**
@@ -65,6 +67,9 @@ public abstract class BaseJdbcBufferedInserter implements JdbcBufferedInserter {
   protected PreparedStatement insertPstmtForFixedBatch;
   private final Retryer<Boolean> retryer;
 
+  // If this config is true, the inserter can insert duplicate primary records according to the specific language
+  protected final boolean replaceExistingValues;
+
   public BaseJdbcBufferedInserter(State state, Connection conn) {
     this.conn = conn;
     this.batchSize = state.getPropAsInt(WRITER_JDBC_INSERT_BATCH_SIZE, DEFAULT_WRITER_JDBC_INSERT_BATCH_SIZE);
@@ -80,6 +85,8 @@ public abstract class BaseJdbcBufferedInserter implements JdbcBufferedInserter {
     this.retryer = RetryerBuilder.<Boolean> newBuilder().retryIfException()
         .withWaitStrategy(WaitStrategies.exponentialWait(1000, maxWait, TimeUnit.SECONDS))
         .withStopStrategy(StopStrategies.stopAfterAttempt(maxAttempts)).build();
+
+    this.replaceExistingValues = state.getPropAsBoolean(ConfigurationKeys.ALLOW_DATA_OVERWRITE);
   }
 
   /**
@@ -175,6 +182,7 @@ public abstract class BaseJdbcBufferedInserter implements JdbcBufferedInserter {
   protected String createInsertStatementStr(String databaseName, String table) {
     return String.format(INSERT_STATEMENT_PREFIX_FORMAT, databaseName, table, JOINER_ON_COMMA.join(this.columnNames));
   }
+
 
   @Override
   public void flush() throws SQLException {
--- a/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
+++ b/gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
@@ -36,7 +36,6 @@ import com.google.common.collect.Lists;
 
 import lombok.ToString;
 
-import org.apache.gobblin.configuration.ConfigurationKeys;
 import org.apache.gobblin.configuration.State;
 import org.apache.gobblin.converter.jdbc.JdbcEntryData;
 import org.apache.gobblin.converter.jdbc.JdbcEntryDatum;
@@ -67,9 +66,6 @@ public abstract class BaseJdbcBufferedInserter implements JdbcBufferedInserter {
   protected PreparedStatement insertPstmtForFixedBatch;
   private final Retryer<Boolean> retryer;
 
-  // If this config is true, the inserter can insert duplicate primary records according to the specific language
-  protected final boolean replaceExistingValues;
-
   public BaseJdbcBufferedInserter(State state, Connection conn) {
     this.conn = conn;
     this.batchSize = state.getPropAsInt(WRITER_JDBC_INSERT_BATCH_SIZE, DEFAULT_WRITER_JDBC_INSERT_BATCH_SIZE);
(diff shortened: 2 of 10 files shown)


gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/JdbcPublisherTest.java
```diff
@@ -87,6 +87,22 @@ public void testPublish() throws IOException, SQLException {
     inOrder.verify(workUnitState, times(1)).setWorkingState(WorkUnitState.WorkingState.COMMITTED);
     inOrder.verify(conn, times(1)).commit();
     inOrder.verify(conn, times(1)).close();
+  }
+
+  @Test(dependsOnMethods = { "testPublish" })
```
umustafi, line 92: Let me know if you have suggestions for improving this unit test

umustafi:
@Will-Lo @phet @arjun4084346 can you review?

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
```diff
@@ -51,6 +51,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(BaseJdbcBufferedInserter.class);
 
   protected static final String INSERT_STATEMENT_PREFIX_FORMAT = "INSERT INTO %s.%s (%s) VALUES ";
+  protected static final String REPLACE_STATEMENT_PREFIX_FORMAT = "REPLACE INTO %s.%s (%s) VALUES ";
```
phet, line 54: `REPLACE` is non-standard-SQL, specific to Mysql, so it shouldn't live in the base.  other DBs may support a `MERGE` statement... or have their own extensions.

gobblin-api/src/main/java/org/apache/gobblin/configuration/ConfigurationKeys.java
```diff
@@ -216,6 +216,12 @@
   public static final String QUEUED_TASK_TIME_MAX_AGE = "taskexecutor.queued_task_time.history.max_age";
   public static final long DEFAULT_QUEUED_TASK_TIME_MAX_AGE = TimeUnit.HOURS.toMillis(1);
 
+  /**
+   * Optional property to specify for Mysql ingestion whether insertions should do replace logic for records
+   * with the same PRIMARY KEY or unique identifier.
+   */
+  public static final String REPLACE_FOR_MYSQL_INGESTION = "replace.allowed.for.mysql.ingestion";
```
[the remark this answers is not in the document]
phet: also, be sure to note in the documentation that replacement granularity is 'record-level', meaning that if there are duplicates in the input (by primary key) they'd even replace one another.  hence which record ultimately gets loaded is undefined (since they may be in separate WUs, which are not processed in a specific order).

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/MySqlWriterCommands.java
```diff
@@ -47,14 +48,17 @@
...
   private static final String DELETE_STATEMENT_FORMAT = "DELETE FROM %s.%s";
 
   private final JdbcBufferedInserter jdbcBufferedWriter;
   private final Connection conn;
+  private final boolean replaceExistingValues;
 
   public MySqlWriterCommands(State state, Connection conn) {
     this.conn = conn;
     this.jdbcBufferedWriter = new MySqlBufferedInserter(state, conn);
+    this.replaceExistingValues = state.getPropAsBoolean(ConfigurationKeys.REPLACE_FOR_MYSQL_INGESTION);
```
phet, line 61: I hold a bias toward ctors taking typed params, rather than sifting through config for special keys.  up to you, but I'd read config earlier on the call stack and translate to a `boolean` param, ultimately used here.

gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/JdbcPublisherTest.java
phet: I can't confirm that it would actually exercise any of the mysql-specific code you modified... does it?

I didn't see you update the `JdbcPublisher` class (above), so perhaps the better unit test to augment would be `MySqlBufferedInserterTest`?

umustafi: It's a bit hard to tell but these tests use `copyTable(database, stagingTable, destinationTable)` which I modified above. However adding a unit test in `MySqlBufferedInserterTest` would be a more direct approach to achieving this.

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/MySqlWriterCommands.java
umustafi: The place I could provide the boolean param for the ctor looks to be in the constructor for `JdbcWriterCommands`. It does not have much logic other than choosing which class to instantiate. There doesn't seem to be a large benefit from translating the config to a boolean one step earlier so I'll leave as is.

gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/JdbcPublisherTest.java
umustafi: update: added a better test in `MySqlBufferedInserterTest ` that checks for a `REPLACE` statement versus `INSERT`

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
```diff
@@ -80,6 +85,8 @@ public BaseJdbcBufferedInserter(State state, Connection conn) {
     this.retryer = RetryerBuilder.<Boolean> newBuilder().retryIfException()
         .withWaitStrategy(WaitStrategies.exponentialWait(1000, maxWait, TimeUnit.SECONDS))
         .withStopStrategy(StopStrategies.stopAfterAttempt(maxAttempts)).build();
+
+    this.replaceExistingValues = state.getPropAsBoolean(ConfigurationKeys.ALLOW_DATA_OVERWRITE);
```
umustafi, line 89: I haven't added the "fail fast" logic in the other child classes that do not have any functionality implemented for this. If the config is true should the job fail in the constructor or at the time of writing `(initialize batch)` or should I add documentation that the logic does not exist and allow the job to succeed?

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/MySqlWriterCommands.java
```diff
@@ -145,7 +149,13 @@ public void drop(String database, String table) throws SQLException {
...
   @Override
   public void copyTable(String databaseName, String from, String to) throws SQLException {
-    String sql = String.format(COPY_INSERT_STATEMENT_FORMAT, databaseName, to, databaseName, from);
+    // Chooses between INSERT and REPLACE logic based on the job configurations
+    String sql;
+    if (this.replaceExistingValues) {
+      sql = String.format(COPY_REPLACE_STATEMENT_FORMAT, databaseName, to, databaseName, from);
+    } else {
+      sql = String.format(COPY_INSERT_STATEMENT_FORMAT, databaseName, to, databaseName, from);
+    }
```
phet, line 158: minor: the ternary op `c ? x : y` is nice for avoiding the need to initialize empty.  here it could even be tucked within the first arg expr to `String.format`.

gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/MySqlBufferedInserterTest.java
```diff
@@ -63,7 +93,7 @@ public void testMySqlBufferedInsert() throws SQLException {
     }
     inserter.flush();
 
-    verify(conn, times(2)).prepareStatement(anyString());
+    verify(conn, times(2)).prepareStatement(matches("REPLACE INTO .*"));
```
phet, line 96: nice--clear and direct.  could we do similar for the `MySqlWriterCommands`?  (does it have a unit test yet for you to build on?)

phet (changes requested):
nice work, urmi.  this is getting close... just a further iterative step in the same evolution

gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/MySqlBufferedInserterTest.java
umustafi: The MySQL writer commands does not have a similar test, which is why I had modified the `JdbcPublisherTest` initially. There is a `JdbcWriterCommandsTest` that mocks a `preparedStatement` to submit and `execute` but it does not actually test insertions into the database. It only tries to read from a mock database (see `testMySqlDateTypeRetrieval`) without testing the other writer commands.

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/BaseJdbcBufferedInserter.java
```diff
@@ -65,6 +67,9 @@
   protected PreparedStatement insertPstmtForFixedBatch;
   private final Retryer<Boolean> retryer;
 
+  // If this config is true, the inserter can insert duplicate primary records according to the specific language
+  protected final boolean replaceExistingValues;
```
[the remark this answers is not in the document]
umustafi: fair argument, I moved this field to the base class of `MySqlBufferedInserter` but not the other child classes because it is the only one that implements this logic so far. As per your comment below I have added it as a field to all the `*WriterCommands` that will fail fast if the config is turned on before creating the `*Inserter`

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/MySqlBufferedInserter.java
```diff
@@ -84,6 +93,26 @@ protected void initializeBatch(String databaseName, String table)
           + " due to # of params limitation " + this.maxParamSize + " , # of columns: " + this.columnNames.size());
     }
     this.batchSize = actualBatchSize;
-    super.initializeBatch(databaseName, table);
+
+    // Use separate insertion statement if replacements are allowed
+    if (this.replaceExistingValues) {
+        this.insertStmtPrefix = createReplaceStatementStr(databaseName, table);
```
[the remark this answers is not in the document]
umustafi: Great idea, modified to override `createInsertStatementStr` to avoid reassigning base class fields in the child class

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/MySqlBufferedInserter.java
```diff
@@ -84,6 +96,13 @@ protected void initializeBatch(String databaseName, String table)
           + " due to # of params limitation " + this.maxParamSize + " , # of columns: " + this.columnNames.size());
     }
     this.batchSize = actualBatchSize;
+
+    // Use separate insertion statement if replacements are allowed
```
phet, line 100: comment could potentially live in `createInsertStatementStr` javadoc

gobblin-modules/gobblin-sql/src/main/java/org/apache/gobblin/writer/commands/PostgresWriterCommands.java
```diff
@@ -51,7 +51,11 @@
   private final JdbcBufferedInserter jdbcBufferedWriter;
   private final Connection conn;
 
-  public PostgresWriterCommands(State state, Connection conn) {
+  public PostgresWriterCommands(State state, Connection conn, boolean overwriteRecords) throws IllegalArgumentException {
+    if (overwriteRecords) {
+      throw new IllegalArgumentException("Replace existing records is not supported in PostgresWriterCommands");
```
phet, line 56: somewhat esoteric, but canonical practice might recommend `UnsupportedOperationException` in this situation - see, e.g.: https://stackoverflow.com/a/49749857

gobblin-modules/gobblin-sql/src/test/java/org/apache/gobblin/writer/MySqlBufferedInserterTest.java
phet: ideally we'd have one, but pragmatically I understand the imperative to ship these changes, so up to you whether we leave this as tech debt IoU to return to

phet (approved):
excellent urmi -- looking very good!

phet (approved):
these changes look good... but `TerradataWriterCommands` need to transition to `UnsupportedOperationException` too.

ZihanLi58 (approved):
+1

